PR #684, opened 2025-01-21
chore(clp-s): Adjust directory creation failure log message during compression.

gibber9809:
# Description
This PR slightly changes the directory creation failure log message during compression in clp-s to print the error code before the error message in order to match common conventions for error logs.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated error logging format for archive directory creation to improve readability of error messages.

### DIFF
--- a/components/core/src/clp_s/ArchiveWriter.cpp
+++ b/components/core/src/clp_s/ArchiveWriter.cpp
@@ -33,10 +33,10 @@ void ArchiveWriter::open(ArchiveWriterOption const& option) {
     m_archive_path = archive_path.string();
     if (false == std::filesystem::create_directory(m_archive_path, ec)) {
         SPDLOG_ERROR(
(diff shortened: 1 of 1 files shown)


components/core/src/clp_s/ArchiveWriter.cpp
```diff
@@ -33,10 +33,10 @@ void ArchiveWriter::open(ArchiveWriterOption const& option) {
     m_archive_path = archive_path.string();
     if (false == std::filesystem::create_directory(m_archive_path, ec)) {
         SPDLOG_ERROR(
-                "Failed to create archive directory \"{}\" - ({}) {}",
+                "Failed to create archive directory \"{}\" - {} ({})",
```
kirkrodrigues, line 36: Can you undo this line? I only meant we should change the order of the args. So the output should be (error value) error message.

kirkrodrigues (approved):

PR #687, opened 2025-01-22
fix(core-clp): Handle 0-byte reads when `BufferReader`'s underlying buffer is fully consumed.

LinZhihao-723:
# Description

The current `BufferReader` implementation doesn't handle cases when reading 0-byte from a fully consumed buffer. This error was triggered when using it in Python ffi library to deserialize an empty string value at the end of the buffer. The size of string is zero, meaning that we don't have to read any bytes from the underlying buffer, but the buffer still returns EndOfFile, and thus causing an incomplete IR exception.
This PR fixes this problem by adding a check for 0-byte reads.

# Validation performed

- Ensure the problem in Python ffi has been solved after applying this PR.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of zero-byte read operations in the buffer reading mechanism.
	- Prevents unnecessary processing when no bytes are requested for reading.

### DIFF
--- a/components/core/src/clp/BufferReader.cpp
+++ b/components/core/src/clp/BufferReader.cpp
@@ -60,13 +60,12 @@ auto BufferReader::try_read(char* buf, size_t num_bytes_to_read, size_t& num_byt
         throw OperationFailed(ErrorCode_BadParam, __FILENAME__, __LINE__);
     }
 
-    if (0 == num_bytes_to_read) {
-        num_bytes_read = 0;
-        return ErrorCode_Success;
-    }
-
     auto remaining_data_size = get_remaining_data_size();
     if (0 == remaining_data_size) {
+        if (0 == num_bytes_to_read) {
+            num_bytes_read = 0;
+            return ErrorCode_Success;
+        }
         return ErrorCode_EndOfFile;
     }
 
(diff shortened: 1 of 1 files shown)


components/core/src/clp/BufferReader.cpp
```diff
@@ -60,6 +60,11 @@ auto BufferReader::try_read(char* buf, size_t num_bytes_to_read, size_t& num_byt
         throw OperationFailed(ErrorCode_BadParam, __FILENAME__, __LINE__);
     }
 
+    if (0 == num_bytes_to_read) {
```
haiqi96, line 63: One concern, when only I read this code, I thought the copy doesn't handle 0 bytes copy.

Then after reading the case description, I realized that the issue is when (0 == remaining_data_size).

If I were to write this code, I might put this if-clause under (0 == remaining_data_size). That said, I feel it's also ok to handle the 0 bytes read earlier so the function doesn't do unnecessary work.

Up to you which you choose.

LinZhihao-723: Yeah I put it this way mainly for efficiency. I could move it inside the if block

haiqi96 (approved):
Looks good to me. The title looks ok as well.

PR #689, opened 2025-01-23
feat(log-viewer-webui): Refactor S3Manager into a Fastify plugin.

haiqi96:
# Description
This PR refactors the S3Manager into a fastify plugin, which allows as to avoid initializing S3 Manager as a global variable.

# Validation performed
Manually tested FS and S3 stream viewing.

## Release Notes

- **New Features**
  - Enhanced S3 file management capabilities with improved configuration and pre-signed URL generation.
  - Added conditional S3 manager initialization for more flexible file streaming.

- **Improvements**
  - Streamlined S3 integration in the application's server configuration.
  - Updated route handling to support more robust S3 file access, directly utilizing the Fastify context for S3 operations.

These updates provide more flexible and reliable file management when working with S3 storage.

### DIFF
--- a/components/log-viewer-webui/server/src/S3Manager.js
+++ b/components/log-viewer-webui/server/src/S3Manager.js
@@ -58,8 +58,8 @@ class S3Manager {
 export default fastifyPlugin(async (app, options) => {
     const {region} = options;
     if (null === region) {
-        return
+        return;
     }
-    console.log(`s3Manager initialized with region ${region}`)
+    console.log(`s3Manager initialized with region ${region}`);
     await app.decorate("s3Manager", new S3Manager(region));
 });
--- a/components/log-viewer-webui/server/src/S3Manager.js
+++ b/components/log-viewer-webui/server/src/S3Manager.js
@@ -16,21 +16,15 @@ const PRE_SIGNED_URL_EXPIRY_TIME_SECONDS = 3600;
  * Class to manage Simple Storage Service (S3) objects.
  */
 class S3Manager {
-    #s3Client;
-
-    #enabled;
+    #s3Client = null;
 
     /**
-     * @param {string|null} region
+     * @param {object} props
+     * @param {string | null} props.region
      */
-    constructor (region) {
-        console.log(`Region's length is ${region.length}`)
-        console.log(`Region is ${region}`)
-        console.log(region)
-        this.#enabled = null !== region;
-        console.log(`Enabled is ${this.#enabled}`)
-        if (true === this.#enabled) {
-            console.log(`Initializing S3 client`)
+    constructor (props) {
+        const region = props.region;
+        if (region !== null) {
             this.#s3Client = new S3Client({
                 region: region,
             });
@@ -38,7 +32,7 @@ class S3Manager {
     }
 
     isEnabled () {
-        return this.#enabled;
+        return null !== this.#s3Client;
     }
 
     /**
@@ -49,6 +43,9 @@ class S3Manager {
      * @throws {Error} If a pre-signed URL couldn't be generated.
      */
     async getPreSignedUrl (s3UriString) {
+        if (false === this.isEnabled()) {
+            throw new Error(`pre-signed URL cannot be generated when S3Manager is not enabled`);
+        }
         const s3Uri = new URL(s3UriString);
         const command = new GetObjectCommand({
             Bucket: s3Uri.hostname,
@@ -69,8 +66,6 @@ class S3Manager {
     }
 }
 
-export default fastifyPlugin(async (app, region) => {
-    console.log(`Region in Plugin init is ${region}`)
-    console.log(region)
-    await app.decorate("s3Manager", new S3Manager(region));
+export default fastifyPlugin(async (app, options) => {
+    await app.decorate("s3Manager", new S3Manager(options));
 });
--- a/components/log-viewer-webui/server/src/S3Manager.js
+++ b/components/log-viewer-webui/server/src/S3Manager.js
@@ -16,22 +16,15 @@ const PRE_SIGNED_URL_EXPIRY_TIME_SECONDS = 3600;
  * Class to manage Simple Storage Service (S3) objects.
  */
 class S3Manager {
-    #s3Client = null;
+    #s3Client;
 
     /**
-     * @param {object} props
-     * @param {string | null} props.region
+     * @param {string} region
      */
-    constructor ({region}) {
-        if (null !== region) {
-            this.#s3Client = new S3Client({
-                region: region,
-            });
-        }
-    }
-
-    isEnabled () {
-        return null !== this.#s3Client;
+    constructor (region) {
+        this.#s3Client = new S3Client({
+            region: region,
+        });
     }
 
     /**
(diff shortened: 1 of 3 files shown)


components/log-viewer-webui/server/src/S3Manager.js
```diff
@@ -55,4 +69,8 @@ class S3Manager {
     }
 }
 
-export default S3Manager;
+export default fastifyPlugin(async (app, region) => {
```
junhaoliao, line 72: `fastifyPlugin` requires the second argument in the callback to be an `options` object: See https://fastify.dev/docs/latest/Reference/Plugins/#plugin-options 

because Fastify specific options like `logLevel`, `logSerializers`, `prefix` are expected to show up in the options object if they exist.

```suggestion
export default fastifyPlugin(async (app, options) => {
```

components/log-viewer-webui/server/src/S3Manager.js
```diff
@@ -55,4 +69,8 @@ class S3Manager {
...
+    console.log(`Region in Plugin init is ${region}`)
+    console.log(region)
+    await app.decorate("s3Manager", new S3Manager(region));
```
junhaoliao, line 75: I think we can directly pass all options to the `S3Manager` constructor, assuming we will be adding more options in the future.
```suggestion
    await app.decorate("s3Manager", new S3Manager(options));
```

components/log-viewer-webui/server/src/app.js
```diff
@@ -41,6 +42,8 @@ const app = async ({
                 port: settings.MongoDbPort,
             },
         });
+        console.log(settings.StreamFilesS3Region)
+        await server.register(S3Manager, settings.StreamFilesS3Region);
```
junhaoliao, line 46: Then we will be calling this plugin registration with an option object:
```suggestion
        await server.register(S3Manager, {region: settings.StreamFilesS3Region});
```

components/log-viewer-webui/server/src/S3Manager.js
```diff
@@ -16,15 +18,27 @@ const PRE_SIGNED_URL_EXPIRY_TIME_SECONDS = 3600;
...
-     * @param {string} region
+     * @param {string|null} region
      */
-    constructor (
-        region,
-    ) {
-        this.#s3Client = new S3Client({
-            region: region,
-        });
+    constructor (region) {
```
junhaoliao, line 26: Then this can take an options object instead

```suggestion
    constructor ({region}) {
```

components/log-viewer-webui/server/src/S3Manager.js
```diff
@@ -16,15 +18,27 @@ const PRE_SIGNED_URL_EXPIRY_TIME_SECONDS = 3600;
 class S3Manager {
     #s3Client;
 
+    #enabled;
+
     /**
-     * @param {string} region
+     * @param {string|null} region
```
junhaoliao, line 24:
```suggestion
     * @param {object} props
     * @param {string | null} props.region
```

junhaoliao (changes requested):
The overall structure is good. I made some comment about how to solve the issue of the `region` option being inaccessible.

components/log-viewer-webui/server/src/S3Manager.js
haiqi96: I used props instead, since I feel its neater.

haiqi96: Nvm, linter hates it. I will use what you suggested.

components/log-viewer-webui/server/src/S3Manager.js
```diff
@@ -55,4 +65,6 @@ class S3Manager {
     }
 }
 
-export default S3Manager;
+export default fastifyPlugin(async (app, options) => {
+    await app.decorate("s3Manager", new S3Manager(options));
```
junhaoliao, line 69: Sorry for overlooking this enablement check mechanism when we discussed this offline. Some changes can be made to make the code even cleaner, if you also agree:
1. Let's treat this callback provided to `fastifyPlugin()` as a factory function of class S3Manager and perform all argument checks inside.
2. If any such checks fail, we don't decorate the Fastify app. Consider adding an error log / throwing an error so developers become aware of this.
3. That way we can (hopefully) ensure all argument passed to S3Manager are valid and do not cause the constructor to crash, and we can also guarantee the `s3Client` is always valid within the scope of the class (no need to do `isEnabled` checks).
4. Most importantly, we can make the query route code unaware of any S3Manager's exact implementation, by checking `fastify.hasDecorator('s3Manager')` instead of `fastify.s3Manager.isEnabled()`

junhaoliao (changes requested):

components/log-viewer-webui/server/src/S3Manager.js
haiqi96: For 1, can you give a more concrete example?

Do you mean instead of writing `new S3Manager()` as the second argument, pass in a function that
1. Takes in `options` as argument
2. If argument verification passes, return calls S3Manager constructor and return a `new S3Manager()` instance
3. If fails, throw an error, and does it need to return anythinig?

junhaoliao: right, let's do something like
```ts
export default fastifyPlugin(async (app, options) => {
    const {xxx} = options;
    if (xxx is invalid) {
        // if throwing halts the server (please help confirm), we shall require users (developers who register this plugin) to validate the args.
        // or simply, instead, `console.error("error message")` then `return`
        throw new Error("error message");
    }

    await app.decorate("s3Manager", new S3Manager(options));
}
```

components/log-viewer-webui/server/src/S3Manager.js
```diff
@@ -55,4 +55,11 @@ class S3Manager {
     }
 }
 
-export default S3Manager;
+export default fastifyPlugin(async (app, options) => {
+    const {region} = options;
+    if (null === region) {
+        return;
+    }
+    console.log(`s3Manager initialized with region ${region}`);
```
junhaoliao, line 63:
```suggestion

    console.log(`Initializing S3Manager with region="${region}"...`);
```

junhaoliao (approved):

junhaoliao:
I modified the PR title directly.

components/log-viewer-webui/server/src/S3Manager.js
```diff
@@ -55,4 +55,11 @@ class S3Manager {
     }
 }
 
-export default S3Manager;
+export default fastifyPlugin(async (app, options) => {
```
junhaoliao, line 58: /**
 * Initializes a Fastify plugin, which decorates the application with an S3 manager at the
 * "s3Manager" property when all plugin options are valid.
 */
```suggestion
export default fastifyPlugin(async (app, options) => {
```

components/log-viewer-webui/server/src/S3Manager.js
```diff
@@ -55,4 +55,16 @@ class S3Manager {
...
+ * "s3Manager" property when all plugin options are valid.
+ */
+export default fastifyPlugin(async (app, options) => {
+    const {region} = options;
+    if (null === region) {
+        return;
+    }
+
+    console.log(`Initializing S3Manager with region="${region}"...`);
+    await app.decorate("s3Manager", new S3Manager(region));
```
haiqi96, line 69: Should we add another log after the decorate is done? Just for debugging purpose.
@junhaoliao

junhaoliao: i think for now it's fine not to log after init completion, since we don't expect constructor to hang. In cases where the constructor throws, we shall be notified for any exception messages.

junhaoliao (approved):

PR #690, opened 2025-01-23
fix(core): Update `DictionaryReader::get_entry_matching_value` to handle case-insensitive searches (fixes #648).

### LINKED ISSUE #648: get_entry_matching_value does not handle ignore-case properly.
### Bughttps://github.com/y-scope/clp/issues/new/choose

Our current implementation of [get_entry_matching_value()](https://github.com/y-scope/clp/blob/main/components/core/src/clp/DictionaryReader.hpp#L236C65-L236C89) assumes only 1 value in the dictionary will match the input string.

This assumption is correct when match is case_sensitive (since the the input value won't contain wildcard). 
However, when ignore_case is switched on, it's possible more than one dictioanry variables will match the input value. (for example, all of "VAR1", "var1" and "vAr1") matches input "var1", but only one of them will be returned. 
As a result, CLP won't return matching messages with the other variables.

I believe the proper fix to this issue is to ether:
1. update the return type of the method, so it returns a vector of entry. For case-sensitive case, the vector will always contain only one entry (or 0 if no matches). For ignore-case, the vector could have more than one elements.
2. Create two different methods, one for case-senstive match, one for non-case sensitive match so they can have their dedicated interface.

### CLP version

329edf66e8d368de3070ba266fcf653060fd1541

### Environment

unrelated to OS version.

### Reproduction steps

1. Compress the attached log.
2. run clg search with `./cmake-build-release/clg -i output/ "My custom log two var1 num"`

[test.log](https://github.com/user-attachments/files/18292386/test.log)

aestriplex:
This pull request implements a fix for the issue #648 

Here's a short recap of the main changes:
1. `get_entry_matching_value` now returns a vector
2. `encode_and_search_dictionary` adds imprecise dict var when there are more matching entries

There are four commits because I worked on it taking advantage of @haiqi96's advices (see the discussion on #648). Should I squash them?

## Summary by CodeRabbit

- **New Features**
	- Enhanced dictionary search functionality to support multiple matching entries.
	- Improved variable matching with more flexible search capabilities.

- **Bug Fixes**
	- Updated error handling for empty variable strings.
	- Refined dictionary entry retrieval process.

- **Tests**
	- Added comprehensive test cases for multiple matching dictionary entries.
	- Verified case-sensitive and case-insensitive search behaviours.

The changes improve the system's ability to handle complex dictionary searches, allowing for more nuanced and precise variable matching across different components.


components/core/src/clp/EncodedVariableInterpreter.cpp
```diff
@@ -386,15 +386,31 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
...
+        }
+
+        std::unordered_set<encoded_variable_t> encoded_vars;
+        std::unordered_set<clp::VariableDictionaryEntry const*> entries_set(
+                entries.begin(),
+                entries.end()
+        );
+        for (auto entry : entries) {
+            encoded_vars.insert(encode_var_dict_id(entry->get_id()));
+        }
```
haiqi96, line 412: Since we are duplicating this code at two places, what do you think about adding a new api for sub_query.add_imprecise_dict_var that takes in a vector of entires as argument?

@LinZhihao-723

components/core/src/clp/EncodedVariableInterpreter.cpp
```diff
@@ -386,15 +386,31 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
...
+            sub_query.add_dict_var(encoded_var, entry);
+            return true;
+        }
+
+        std::unordered_set<encoded_variable_t> encoded_vars;
+        std::unordered_set<clp::VariableDictionaryEntry const*> entries_set(
+                entries.begin(),
+                entries.end()
+        );
+        for (auto entry : entries) {
```
haiqi96, line 410:
```suggestion
        for (auto const& entry : entries) {
```

components/core/src/clp_s/search/Output.cpp
```diff
@@ -932,12 +932,12 @@ void Output::populate_string_queries(std::shared_ptr<Expression> const& expr) {
                     }
                 }
 
-                auto const* entry = m_var_dict->get_entry_matching_value(
+                auto entries = m_var_dict->get_entry_matching_value(
```
haiqi96, line 935:
```suggestion
                auto const entries = m_var_dict->get_entry_matching_value(
```

components/core/src/clp_s/search/Output.cpp
```diff
@@ -932,12 +932,12 @@ void Output::populate_string_queries(std::shared_ptr<Expression> const& expr) {
...
                         unescaped_query_string,
                         m_ignore_case
                 );
 
-                if (entry != nullptr) {
+                for (auto entry : entries) {
```
haiqi96, line 940:
```suggestion
                for (auto const& entry : entries) {
```

components/core/src/clp_s/search/clp_search/EncodedVariableInterpreter.cpp
```diff
@@ -34,15 +34,31 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
         LogTypeDictionaryEntry::add_double_var(logtype);
         sub_query.add_non_dict_var(encoded_var);
     } else {
-        auto entry = var_dict.get_entry_matching_value(var_str, ignore_case);
-        if (nullptr == entry) {
+        auto entries = var_dict.get_entry_matching_value(var_str, ignore_case);
```
haiqi96, line 37: We weren't using const in the previous code, but since you are changing the code, let's adpat to our latest coding standard.
```suggestion
        auto const entries = var_dict.get_entry_matching_value(var_str, ignore_case);
```

components/core/src/clp_s/DictionaryReader.hpp
haiqi96, line 1: @gibber9809 The change in CLP-S looks reasonable to me, but can you take another look?

components/core/tests/test-EncodedVariableInterpreter.cpp
```diff
@@ -376,6 +376,43 @@ TEST_CASE("EncodedVariableInterpreter", "[EncodedVariableInterpreter]") {
         ));
     }
 
+    SECTION("Test multiple metching values") {
+        char const cVarDictPath[] = "var.dict";
+        char const cVarSegmentIndexPath[] = "var.segindex";
+        vector<string> var_strs = {"python2.7.3", "Python2.7.3", "PyThOn2.7.3", "PYTHON2.7.3"};
+        clp::VariableDictionaryWriter var_dict_writer;
```
haiqi96, line 383: Nit: let's add some new lines so it's easier to read.
```suggestion

        clp::VariableDictionaryWriter var_dict_writer;
```

components/core/tests/test-EncodedVariableInterpreter.cpp
```diff
@@ -376,6 +376,43 @@ TEST_CASE("EncodedVariableInterpreter", "[EncodedVariableInterpreter]") {
...
+        var_dict_writer.open(cVarDictPath, cVarSegmentIndexPath, cVariableDictionaryIdMax);
+        vector<encoded_variable_t> encoded_vars;
```
haiqi96, line 385:
```suggestion

        vector<encoded_variable_t> encoded_vars;
```

components/core/tests/test-EncodedVariableInterpreter.cpp
```diff
@@ -376,6 +376,43 @@ TEST_CASE("EncodedVariableInterpreter", "[EncodedVariableInterpreter]") {
...
+        vector<clp::variable_dictionary_id_t> var_ids;
+        clp::LogTypeDictionaryEntry logtype_dict_entry;
+        for (auto var_str : var_strs) {
+            EncodedVariableInterpreter::encode_and_add_to_dictionary(
```
haiqi96, line 389: Instead of directly using the var_strs as msg argument for encode_and_add_to_dictionary, can we construct a message with a dummy log type?

I guess what you can do can be similar to

```
string const msg_template  = "here is a string with a dictionary var: ".

for var_str in var_strs:
    auto const msg = msg_template + var_id
    ...
```

components/core/tests/test-EncodedVariableInterpreter.cpp
```diff
@@ -376,6 +376,43 @@ TEST_CASE("EncodedVariableInterpreter", "[EncodedVariableInterpreter]") {
...
+    SECTION("Test multiple metching values") {
+        char const cVarDictPath[] = "var.dict";
+        char const cVarSegmentIndexPath[] = "var.segindex";
+        vector<string> var_strs = {"python2.7.3", "Python2.7.3", "PyThOn2.7.3", "PYTHON2.7.3"};
+        clp::VariableDictionaryWriter var_dict_writer;
+        var_dict_writer.open(cVarDictPath, cVarSegmentIndexPath, cVariableDictionaryIdMax);
+        vector<encoded_variable_t> encoded_vars;
+        vector<clp::variable_dictionary_id_t> var_ids;
+        clp::LogTypeDictionaryEntry logtype_dict_entry;
+        for (auto var_str : var_strs) {
```
haiqi96, line 388:
```suggestion
        for (auto const& var_str : var_strs) {
```

components/core/tests/test-EncodedVariableInterpreter.cpp
```diff
@@ -376,6 +376,43 @@ TEST_CASE("EncodedVariableInterpreter", "[EncodedVariableInterpreter]") {
...
+                    var_ids
+            );
+        }
+        var_dict_writer.close();
+
+        clp::VariableDictionaryReader var_dict_reader;
+        var_dict_reader.open(cVarDictPath, cVarSegmentIndexPath);
+        var_dict_reader.read_new_entries();
+
+        REQUIRE(var_dict_reader.get_entry_matching_value(var_strs[0], true).size()
```
haiqi96, line 403: let's use at() instead

haiqi96:
Thanks for opening the PR. When the PR is approved. you will need to squash the changes so it looks like a single commit. We can discuss after we get there. 

I have left a batch of comments.
Note that you will also need to change the title of the PR. you can refer to the past merged PRs to get an idea of how it should look like.

components/core/src/clp/DictionaryReader.hpp
```diff
@@ -233,26 +233,29 @@ std::string const& DictionaryReader<DictionaryIdType, EntryType>::get_value(Dict
...
         std::string const& search_string,
         bool ignore_case
 ) const {
+    std::vector<EntryType const*> entries;
     if (false == ignore_case) {
         for (auto const& entry : m_entries) {
             if (entry.get_value() == search_string) {
-                return &entry;
+                entries.push_back(&entry);
+                return entries; /* early exit for case sensitive branch */
```
haiqi96, line 246: We usually don't use /* */ unless it's docstring.

Let's move the comment on the line above and use // instead
```suggestion
                // early exit for case sensitive branch
                return entries;
```

components/core/src/clp/EncodedVariableInterpreter.cpp
aestriplex: What if we added an API that takes an itarator? That way we could potentially deal with all collections (that support iterators) without much overhead, Something like this:
```cpp
template <typename ConstIterator>
void SubQuery::add_imprecise_dict_var(ConstIterator begin, ConstIterator end) {
	for(; begin != end; ++begin) {
		...
	}
}
```

components/core/src/clp/EncodedVariableInterpreter.cpp
```diff
@@ -386,15 +386,31 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
...
             return false;
         }
-        encoded_var = encode_var_dict_id(entry->get_id());
 
         LogTypeDictionaryEntry::add_dict_var(logtype);
-        sub_query.add_dict_var(encoded_var, entry);
+
+        if (entries.size() == 1) {
+            clp::VariableDictionaryEntry const* entry = entries[0];
+            auto encoded_var = encode_var_dict_id(entry->get_id());
```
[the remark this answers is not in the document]
aestriplex: Yes, it was an oversight. It seems to me that remove the "globally declared" encoded_var here may make the first two cases a little more verbose. At the end of the day, this seems clear enough:
```cpp
encoded_variable_t encoded_var;
if (convert_string_to_representable_integer_var(var_str, encoded_var)) {
    LogTypeDictionaryEntry::add_int_var(logtype);
    sub_query.add_non_dict_var(encoded_var);
} else if (convert_string_to_representable_float_var(var_str, encoded_var)) {
    LogTypeDictionaryEntry::add_float_var(logtype);
    sub_query.add_non_dict_var(encoded_var);
} else {
    auto const entries = var_dict.get_entry_matching_value(var_str, ignore_case);
    ...
}
```

components/core/src/clp/DictionaryReader.hpp
```diff
@@ -233,26 +233,29 @@ std::string const& DictionaryReader<DictionaryIdType, EntryType>::get_value(Dict
...
+DictionaryReader<DictionaryIdType, EntryType>::get_entry_matching_value(
         std::string const& search_string,
         bool ignore_case
 ) const {
+    std::vector<EntryType const*> entries;
     if (false == ignore_case) {
         for (auto const& entry : m_entries) {
             if (entry.get_value() == search_string) {
-                return &entry;
+                entries.push_back(&entry);
```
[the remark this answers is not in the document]
LinZhihao-723: It's ok to use `push_back` for pointer types. Emplace is only required if the underlying object requires a constructor.

components/core/src/clp_s/search/clp_search/EncodedVariableInterpreter.cpp
```diff
@@ -34,15 +34,31 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
         LogTypeDictionaryEntry::add_double_var(logtype);
         sub_query.add_non_dict_var(encoded_var);
     } else {
-        auto entry = var_dict.get_entry_matching_value(var_str, ignore_case);
-        if (nullptr == entry) {
+        auto const entries = var_dict.get_entry_matching_value(var_str, ignore_case);
+
```
haiqi96, line 38:
```suggestion
```

components/core/src/clp_s/search/clp_search/EncodedVariableInterpreter.cpp
```diff
@@ -34,15 +34,31 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
...
             // Not in dictionary
             return false;
         }
-        encoded_var = VariableEncoder::encode_var_dict_id(entry->get_id());
 
         LogTypeDictionaryEntry::add_non_double_var(logtype);
-        sub_query.add_dict_var(encoded_var, entry);
+
+        if (entries.size() == 1) {
+            VariableDictionaryEntry const* entry = entries[0];
```
haiqi96, line 47:
```suggestion
            auto const* entry = entries.at(0);
```

components/core/src/clp_s/search/clp_search/EncodedVariableInterpreter.cpp
```diff
@@ -34,15 +34,31 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
...
+            auto encoded_var = VariableEncoder::encode_var_dict_id(entry->get_id());
```
haiqi96, line 48: shadowing variables outside this scope?

components/core/src/clp/EncodedVariableInterpreter.cpp
```diff
@@ -386,15 +386,31 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
...
+
+        std::unordered_set<encoded_variable_t> encoded_vars;
+        std::unordered_set<clp::VariableDictionaryEntry const*> const entries_set(
+                entries.begin(),
+                entries.end()
+        );
+        for (auto const& entry : entries) {
+            encoded_vars.insert(encode_var_dict_id(entry->get_id()));
+        }
+        sub_query.add_imprecise_dict_var(encoded_vars, entries_set);
```
haiqi96, line 413: Do you see any value turning this into a new method for SubQuery that adds imprecise_dict_var by taking in a ```vector<VariableDictionaryEntry*>```?

i.e.
```
void SubQuery::add_imprecise_dict_var(
        vector<VariableDictionaryEntry*> const& possible_var_dict_entries,
);
```

haiqi96:
@aestriplex In the email, I saw on of your comment mentioning about adding a ```add_imprecise_dict_var``` method that takes in a template, but can't find the comment in the Pull request. 

Just want to double check if you intended to delete the comment.

components/core/src/clp_s/search/clp_search/EncodedVariableInterpreter.cpp
```diff
@@ -34,15 +34,31 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
...
+            sub_query.add_dict_var(encoded_var, entry);
+            return true;
+        }
+
+        std::unordered_set<encoded_variable_t> encoded_vars;
+        std::unordered_set<VariableDictionaryEntry const*> entries_set(
+                entries.begin(),
+                entries.end()
+        );
+        for (auto entry : entries) {
```
haiqi96, line 58:
```suggestion
        for (auto const entry : entries) {
```

components/core/src/clp_s/DictionaryReader.hpp
```diff
@@ -64,9 +64,9 @@ class DictionaryReader {
      * Gets the entry exactly matching the given search string
```
haiqi96, line 64: @gibber9809 Can you take a look at the changes in clp-s part?

components/core/src/clp_s/search/clp_search/EncodedVariableInterpreter.cpp
aestriplex: It was the same mistake as in `components/core/src/clp/EncodedVariableInterpreter.cpp`. I didn't fix it in clp_s in the first place. Done

components/core/src/clp/EncodedVariableInterpreter.cpp
haiqi96: > What if we added an API that takes an iterator? That way we could potentially deal with all collections (that support iterators) without much overhead, Something like this:
> 
> ```c++
> template <typename ConstIterator>
> void SubQuery::add_imprecise_dict_var(ConstIterator begin, ConstIterator end) {
> 	for(; begin != end; ++begin) {
> 		...
> 	}
> }
> ```

Hmm, how would it change our existing code? Note the current API takes two parameters.
Do you mind giving an example of converting the current API into this template?

aestriplex: @haiqi96 I was thinking about something like this
```cpp
template <typename DictVarIterator, typename EntiresIterator>
void add_imprecise_dict_var(
        DictVarIterator possible_dict_vars_begin,
        DictVarIterator possible_dict_vars_end,
        EntiresIterator possible_var_dict_entries_begin,
        EntiresIterator possible_var_dict_entries_end
){
    QueryVar queryVar(
        std::unordered_set<encoded_variable_t>(
                possible_dict_vars_begin,
                possible_dict_vars_end
        ),
        std::unordered_set<VariableDictionaryEntry const*>(
                possible_var_dict_entries_begin,
                possible_var_dict_entries_end
        )
    );
    m_vars.push_back(queryVar);
}
```
You can also get rid of the temporary queryVar by using emplace_back.
I wouldn't change the current API (the one that takes two `std::unordered_set`). I would add this one instead. It would be used in the same contexts as a hypothetical API that would take an std::vector, but it would be more general and could also be used with std::array, std::set, or any other custom collection that supports iterators.

haiqi96: Would it fit in the current code change? without preconstructing encoded_vars, there doesn't seem to be an easy way to get the the iterator of possible_dict_vars_begin (I assume it's the iterator for ID).

aestriplex: Yes, we can use it (e.g. in `EncodedVariableInterpreter.cpp`) like this
```cpp
std::unordered_set<encoded_variable_t> encoded_vars;
for (auto const entry : entries) {
     encoded_vars.insert(encode_var_dict_id(entry->get_id()));
}
sub_query.add_imprecise_dict_var(encoded_vars.cbegin(), encoded_vars.cend(), entries.cbegin(),entries.cend());
```
We would basically just get rid of these instruction
```cpp
std::unordered_set<clp::VariableDictionaryEntry const*> entries_set(
      entries.begin(),
      entries.end()
);
```
because we can use the entries vector directly.
I compiled it and I executed the unit tests and everything works.

Anyway, probably it wouldn't be a good choice for at least two reasons
1. It takes four parameters instead of the original two. That's not a bad thing in itself, but it seems to me that it makes the cose less readable
2. There's no convenient way to unify it with the original API (the one that takes two unordered sets).

haiqi96:
Btw, please try to avoid using force push. It seems some comments are not displayed properly once there is a forced push

components/core/src/clp/EncodedVariableInterpreter.cpp
```diff
@@ -386,15 +386,30 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
...
+
+        std::unordered_set<encoded_variable_t> encoded_vars;
+        std::unordered_set<clp::VariableDictionaryEntry const*> const entries_set(
+                entries.begin(),
+                entries.end()
+        );
+        for (auto const entry : entries) {
+            encoded_vars.insert(encode_var_dict_id(entry->get_id()));
+        }
+        sub_query.add_imprecise_dict_var(encoded_vars, entries_set);
```
haiqi96, line 412: what do you think about adding a new signature for sub_query.add_imprecise_dict_var that takes in a vector of entires as argument?. So we can replace this piece of code witha a single call to the new signature @LinZhihao-723

Note this won't solve the code duplication problem, because CLP-S duplicates the entire Query class

aestriplex:
> Btw, please try to avoid using force push. It seems some comments are not displayed properly once there is a forced push

@haiqi96 You're right, my bad. I wasn't that used to GitHub pull requests, and I was trying to keep the branch up to date with man

components/core/src/clp/EncodedVariableInterpreter.cpp
aestriplex: No problem. Thanks a lot for your review

haiqi96:
> > Btw, please try to avoid using force push. It seems some comments are not displayed properly once there is a forced push
> 
> @haiqi96 You're right, my bad. I wasn't that used to GitHub pull requests, and I was trying to keep the branch up to date with man

No problem. I also did a lot of forced push when I started working with PRs. To keep the branch up to date, you and always run `git merge main`. Note it would make your commit history looks ugly, but since eventually we will squash all changes into one commit when PRing into main, the commit history wouldn't matter.

There might be cases where when you do a rebase (I assume you are rebasing to main?), there are mutliple commits causing conflicts which could cause a lot of headache since you need to resolve conflict for each commits. Using git merge will make the things easier

components/core/src/clp/EncodedVariableInterpreter.cpp
```diff
@@ -386,15 +386,30 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
...
         }
-        encoded_var = encode_var_dict_id(entry->get_id());
 
         LogTypeDictionaryEntry::add_dict_var(logtype);
-        sub_query.add_dict_var(encoded_var, entry);
+
+        if (entries.size() == 1) {
+            auto const* entry = entries.at(0);
+            encoded_var = encode_var_dict_id(entry->get_id());
+            sub_query.add_dict_var(encoded_var, entry);
```
LinZhihao-723, line 400:
```suggestion
            sub_query.add_dict_var(encode_var_dict_id(entry->get_id()), entry);
```

components/core/src/clp/EncodedVariableInterpreter.cpp
```diff
@@ -386,15 +386,30 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
...
+            return true;
+        }
+
+        std::unordered_set<encoded_variable_t> encoded_vars;
```
LinZhihao-723, line 404: Missing include of `<unordered_set>` (this is recently added to our contributing guideline)

components/core/src/clp/EncodedVariableInterpreter.cpp
LinZhihao-723: Instead of returning a vector, can we directly return an unordered set?

components/core/src/clp/EncodedVariableInterpreter.cpp
```diff
@@ -386,15 +386,30 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
...
+        std::unordered_set<clp::VariableDictionaryEntry const*> const entries_set(
+                entries.begin(),
+                entries.end()
+        );
+        for (auto const entry : entries) {
+            encoded_vars.insert(encode_var_dict_id(entry->get_id()));
+        }
```
LinZhihao-723, line 411:
```suggestion
        std::unordered_set<clp::VariableDictionaryEntry const*> const entries_set{
                entries.cbegin(),
                entries.cend()
        };

        std::unordered_set<encoded_variable_t> encoded_vars;
        for (auto const* entry : entries) {
            encoded_vars.emplace(encode_var_dict_id(entry->get_id()));
        }
        sub_query.add_imprecise_dict_var(encoded_vars, entries_set);
```
Some coding style related comments. If we choose to go with the design as discussed in another comment we might not need to apply this change in the end, but I still leave it here just to get u on board to our C++ coding guidelines for future development :D

- We should order `entries_set` before `encoded_vars` to make the declaration of `encoded_vars` as close to its first use as possible
- We use `{}` instead of `()` for explicitly typed variable initialization. It's a bit tricky to use `{}` in std containers sometimes (e.g., using `{1, 2}` to initialize a vector would not be equivalent to `(1, 2)`), but in this case the inputs are iterators so we should use `{}`.
- Let's use the const version of iterators whenever possible for const correctness.
- When using `auto` to declare a variable, we should add `*` for ptrs and `&` for references. This is actually checked by our clang-tidy rules, which is to improve the readability.
- In most cases we should use `emplace` rather than `insert` (same as `emplace_back` rather than `push_back`) for in-place object construct. Since the underlying type is a fundamental type, so in this case using `insert` would be fine and there's no rule to not strictly enforce u to use `emplace`. If u don't want to think about `emplace` vs. `insert` everytime, u could always use `emplace`.

LinZhihao-723 (changes requested):
Sorry for the late reply. Thanks for your contribution!
I gave a short review for the changes in clp. The overall fix looks good to me, the only issue is that we might need to look into this discussion and come with an agreement: https://github.com/y-scope/clp/pull/690/files#r1932918266
I also left some comments about some style changes to get u on board with with with with our general coding guidelines. Let me know if you have any questions on these comments.
I will proceed my review for clp-s and unit tests after we resolve the clp component.

components/core/src/clp/EncodedVariableInterpreter.cpp
aestriplex: Thank you @LinZhihao-723 for this feedback. I implemented your suggestions both in clp and in clp_s namespaces

aestriplex:
@LinZhihao-723
> Instead of returning a vector, can we directly return an unordered set?

It was considered in one of the comments [here](https://github.com/y-scope/clp/issues/648#issuecomment-2605297953). Looks good to me

LinZhihao-723:
> @LinZhihao-723
> 
> > Instead of returning a vector, can we directly return an unordered set?
> 
> It was considered in one of the comments [here](https://github.com/y-scope/clp/issues/648#issuecomment-2605297953). Looks good to me

Sure, make sense to me. Shall we create a new issue to keep track of this vector-to-set inefficiency?

components/core/src/clp/EncodedVariableInterpreter.cpp
```diff
@@ -386,15 +387,29 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
...
+            auto const* entry = entries.at(0);
+            sub_query.add_dict_var(encode_var_dict_id(entry->get_id()), entry);
+            return true;
+        }
+
+        std::unordered_set<clp::VariableDictionaryEntry const*> const entries_set{
+                entries.cbegin(),
+                entries.cend()
+        };
+        std::unordered_set<encoded_variable_t> encoded_vars;
```
LinZhihao-723, line 408: Sorry for missing this in the previous review. We can reserve its size to avoid rehashing:
```suggestion
        std::unordered_set<encoded_variable_t> encoded_vars;
        encoded_vars.reserve(entries.size());
```

LinZhihao-723 (changes requested):
The clp-s changes lgtm.
@gibber9809 Do you want to take a look for a sanity check?
NTS: I will still need to go through the unit test changes

components/core/src/clp/EncodedVariableInterpreter.cpp
aestriplex: Added it bot in clp and in clp_s

aestriplex:
@LinZhihao-723 
> Sure, make sense to me. Shall we create a new issue to keep track of this vector-to-set inefficiency?

Yes, I'll open a new issue to keep track of this, and I'll link this pr

gibber9809:
> The clp-s changes lgtm. @gibber9809 Do you want to take a look for a sanity check? NTS: I will still need to go through the unit test changes

yes, clp-s changes look good to me as well

components/core/tests/test-EncodedVariableInterpreter.cpp
```diff
@@ -376,6 +376,47 @@ TEST_CASE("EncodedVariableInterpreter", "[EncodedVariableInterpreter]") {
         ));
     }
 
+    SECTION("Test multiple metching values") {
+        char const cVarDictPath[] = "var.dict";
+        char const cVarSegmentIndexPath[] = "var.segindex";
```
LinZhihao-723, line 381: according to our latest coding guideline, we should use `constexpr std::string_view` for compile time static strings.

components/core/tests/test-EncodedVariableInterpreter.cpp
```diff
@@ -376,6 +376,47 @@ TEST_CASE("EncodedVariableInterpreter", "[EncodedVariableInterpreter]") {
...
+        vector<string> var_strs = {"python2.7.3", "Python2.7.3", "PyThOn2.7.3", "PYTHON2.7.3"};
```
LinZhihao-723, line 382: How about using `constexpr std::array<std::string_view>` instead?

components/core/tests/test-EncodedVariableInterpreter.cpp
```diff
@@ -376,6 +376,47 @@ TEST_CASE("EncodedVariableInterpreter", "[EncodedVariableInterpreter]") {
...
+        clp::VariableDictionaryWriter var_dict_writer;
+
+        var_dict_writer.open(cVarDictPath, cVarSegmentIndexPath, cVariableDictionaryIdMax);
+
+        vector<encoded_variable_t> encoded_vars;
+        vector<clp::variable_dictionary_id_t> var_ids;
+        clp::LogTypeDictionaryEntry logtype_dict_entry;
+        string const msg_template = "here is a string with a dictionary var: ";
```
LinZhihao-723, line 390:
```suggestion
        string const msg_template{"here is a string with a dictionary var: "};
```
Let's use `{}` initializer whenever possible.

components/core/tests/test-EncodedVariableInterpreter.cpp
```diff
@@ -376,6 +376,47 @@ TEST_CASE("EncodedVariableInterpreter", "[EncodedVariableInterpreter]") {
...
+                == var_strs.size());
+        REQUIRE(var_dict_reader.get_entry_matching_value(var_strs.at(0), false).size() == 1);
+
+        var_dict_reader.close();
+
+        // Clean-up
+        int retval = unlink(cVarDictPath);
+        REQUIRE(0 == retval);
+        retval = unlink(cVarSegmentIndexPath);
+        REQUIRE(0 == retval);
```
LinZhihao-723, line 417:
```suggestion
        REQUIRE(0 == unlink(cVarDictPath));
        REQUIRE(0 == unlink(cVarSegmentIndexPath));
```

LinZhihao-723 (changes requested):
Some comments on the unit tests: all of them are just style changes

aestriplex:
@LinZhihao-723 I applied your suggestions. I also added a couple of missing `std::` on my tests

LinZhihao-723:
> @LinZhihao-723
> 
> > Instead of returning a vector, can we directly return an unordered set?
> 
> It was considered in one of the comments [here](https://github.com/y-scope/clp/issues/648#issuecomment-2605297953). Looks good to me

lgtm! Let's create the issue and I will approve :D

aestriplex:
@LinZhihao-723 I created the issue. Should I merge main into this branch and squash all the commits before you approve?

aestriplex:
@LinZhihao-723 Ok, I merged main, thank you

LinZhihao-723 (approved):
For the PR title, how about:
```
fix(core): Update `DictionaryReader::get_entry_matching_value` to handle case-insensitive searches (fixes #648).
```

aestriplex:
@LinZhihao-723 LGTM, updated. Thanks

gibber9809 (approved):
LGTM!

PR #677, opened 2025-01-18
feat(ffi): Add support for serializing/deserializing user-defined stream-level metadata in the new IR format.

LinZhihao-723:
# Description

This PR adds support for serializing and deserializing user-defined stream-level metadata.
To support user-defined metadata serialization, we update serializer's factory function to allow users to give the metadata as an optional JSON object.
We already have stream-level deserialization support before this PR. However, we don't store the deserialized metadata. This PR updates the deserializer to store the stream-level metadata as a private member and exposes APIs to get an immutable view of the metadata through the deserializer.

# Validation performed

- Ensure all workflows passed.
- Update unit tests to test metadata serder.

## Summary by CodeRabbit

- **New Features**
	- Added support for user-defined metadata in serialization and deserialization processes
	- Introduced a new method to retrieve metadata from deserialized streams

- **Improvements**
	- Enhanced serializer to optionally include custom metadata
	- Added a new constant for user-defined metadata key

- **Technical Updates**
	- Updated method signatures in serializer and deserializer classes to accommodate metadata handling

PR #691, opened 2025-01-24
feat(ffi): Add type checking to only accept JSON objects as user-defined stream-level metadata for the new IR format.

LinZhihao-723:
# Description

As introduced in #677, we added support for serializing/deserializing user-defined stream-level metadata. However, we didn't add any type checking in this implementation, meaning that user can give any valid `nlohmann::json` instances, including arrays and primitive type values.

After an offline discussion with @kirkrodrigues, we decided to constrain the metadata type to only accept JSON objects (key-value pair dictionaries) to avoid undefined behaviors in our ffi libraries. This PR implements this new requirement.

Ideally, the way to implement this PR is to constrain the type of the input to be `nlohmann::json::object_t`, so that other types of inputs can't be compiled. However, after testing I realized the compiler doesn't prevent an implicit conversion from `nlohmann::json` to `nlohmann::json::object_t`, meaning that you could still pass in any `nlohmann::json` object to a `nlohmann::json::object_t` typed parameter, and the error will be raised happens during the runtime through a nlohmann::json exception. To prevent the use of exceptions, we add explicit type checking before serializing the user-defined metadata, and return `std::errc::protocol_not_supported` when seeing a non-JSON-object input.

# Validation performed

- Added new unit tests to ensure invalid types of values can't be serialized.

## Release Notes

- **New Features**
  - Enhanced metadata validation for serialization process.

- **Bug Fixes**
  - Added robust error handling for invalid user-defined metadata inputs.

- **Tests**
  - Introduced comprehensive test coverage for invalid user-defined metadata scenarios.

The update improves the reliability of metadata processing by ensuring only valid JSON objects are accepted during serialization, preventing potential data integrity issues.

### DIFF
--- a/components/core/src/clp/ffi/ir_stream/Serializer.hpp
+++ b/components/core/src/clp/ffi/ir_stream/Serializer.hpp
@@ -44,7 +44,7 @@ public:
      * @param optional_user_defined_metadata Stream-level user-defined metadata, given as a JSON
      * object.
      * @return A result containing the serializer or an error code indicating the failure:
-     * - std::errc::protocol_error if failed to serialize the preamble.
+     * - std::errc::protocol_error if the stream's metadata couldn't be serialized.
      * - std::errc::protocol_not_supported if the given user-defined metadata is not a JSON object.
      */
     [[nodiscard]] static auto create(
--- a/components/core/src/clp/ffi/ir_stream/Serializer.hpp
+++ b/components/core/src/clp/ffi/ir_stream/Serializer.hpp
@@ -41,9 +41,11 @@ public:
     // Factory functions
     /**
      * Creates an IR serializer and serializes the stream's preamble.
-     * @param optional_user_defined_metadata Stream-level user-defined metadata.
+     * @param optional_user_defined_metadata Stream-level user-defined metadata, given as a JSON
+     * object.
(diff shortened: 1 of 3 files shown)


components/core/src/clp/ffi/ir_stream/Serializer.hpp
```diff
@@ -41,9 +41,11 @@ class Serializer {
     // Factory functions
     /**
      * Creates an IR serializer and serializes the stream's preamble.
-     * @param optional_user_defined_metadata Stream-level user-defined metadata.
+     * @param optional_user_defined_metadata Stream-level user-defined metadata, given as a JSON
+     * object.
      * @return A result containing the serializer or an error code indicating the failure:
-     * - std::errc::protocol_error on failure to serialize the preamble.
+     * - std::errc::protocol_error if failed to serialize the preamble.
```
kirkrodrigues, line 47:
```suggestion
     * - std::errc::protocol_error if the stream's metadata couldn't be serialized.
```

* The grammar of the original is correct.
  * If you want to write it as "value if ..." then you can write "if the preamble could not be serialized."
* That said, I think my suggestion above is more accurate?

kirkrodrigues (approved):

PR #697, opened 2025-01-27
feat(clp-package): Add package support for the indexer process.

wraymo:
# Description

This PR adds the indexer process as a step for compressing JSON logs using the package. After compression, it runs the indexer process to populate all json paths and corresponding types into a mysql table. This table can then be used by any SQL query engine to resolve metadata for CLP archives.

# Validation performed

+ Built and package
+ Modified `clp_config.yml` to output archives to a test S3 bucket
+ Compressed part of [MongoDB logs](https://zenodo.org/records/11075361) using the package
+ Archives were successfully uploaded to S3
+ Verified that the rows in the clp_column_metadata_default table match the JSON paths and types from the logs.

## Summary by CodeRabbit

- **New Features**
	- The build and packaging process now integrates a new component, streamlining overall workflows.
	- A new SQL table, `column_metadata_default`, has been added to enhance metadata management.
	- Enhanced metadata handling with an additional storage mechanism to support improved data management.
	- Updated the file compression workflow with an optimized metadata upload step for smoother performance.

### DIFF
--- a/components/clp-py-utils/clp_py_utils/initialize-clp-metadata-db.py
+++ b/components/clp-py-utils/clp_py_utils/initialize-clp-metadata-db.py
@@ -99,7 +99,7 @@ def main(argv):
                 f"""
                 CREATE TABLE IF NOT EXISTS `{table_prefix}column_metadata_default` (
                     `name` VARCHAR(512) NOT NULL,
-                    `type` INT NOT NULL,
+                    `type` TINYINT NOT NULL,
                     PRIMARY KEY (`name`, `type`)
                 )
                 """
--- a/components/clp-py-utils/clp_py_utils/initialize-clp-metadata-db.py
+++ b/components/clp-py-utils/clp_py_utils/initialize-clp-metadata-db.py
@@ -95,6 +95,17 @@ def main(argv):
                 """
             )
 
+            metadata_db_cursor.execute(
+                f"""
+            CREATE TABLE IF NOT EXISTS `{table_prefix}column_metadata_default` (
+                `name` VARCHAR(512) NOT NULL,
+                `type` BIGINT NOT NULL,
+                PRIMARY KEY (`name`, `type`)
+                PRIMARY KEY (`file_id`,`tag_id`)
+            )
+            """
+            )
+
             metadata_db.commit()
     except:
         logger.exception("Failed to create clp metadata tables.")
--- a/components/clp-py-utils/clp_py_utils/initialize-clp-metadata-db.py
+++ b/components/clp-py-utils/clp_py_utils/initialize-clp-metadata-db.py
@@ -97,13 +97,12 @@ def main(argv):
 
             metadata_db_cursor.execute(
                 f"""
-            CREATE TABLE IF NOT EXISTS `{table_prefix}column_metadata_default` (
-                `name` VARCHAR(512) NOT NULL,
-                `type` BIGINT NOT NULL,
-                PRIMARY KEY (`name`, `type`)
-                PRIMARY KEY (`file_id`,`tag_id`)
-            )
-            """
+                CREATE TABLE IF NOT EXISTS `{table_prefix}column_metadata_default` (
+                    `name` VARCHAR(512) NOT NULL,
+                    `type` BIGINT NOT NULL,
+                    PRIMARY KEY (`name`, `type`)
+                )
+                """
             )
 
             metadata_db.commit()
--- a/components/clp-py-utils/clp_py_utils/initialize-clp-metadata-db.py
+++ b/components/clp-py-utils/clp_py_utils/initialize-clp-metadata-db.py
@@ -99,7 +99,7 @@ def main(argv):
                 f"""
                 CREATE TABLE IF NOT EXISTS `{table_prefix}column_metadata_default` (
                     `name` VARCHAR(512) NOT NULL,
-                    `type` BIGINT NOT NULL,
+                    `type` INT NOT NULL,
                     PRIMARY KEY (`name`, `type`)
                 )
                 """
--- a/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
+++ b/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
@@ -353,15 +353,27 @@ def run_clp(
                         last_archive_stats,
                     )
                     db_conn.commit()
+
                 if StorageEngine.CLP_S == clp_storage_engine:
+                    # TODO: Since CLP doesn't currently support datasets but users of the index
+                    # require a dataset name, we hardcode a name for now.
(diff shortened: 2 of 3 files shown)


components/clp-py-utils/clp_py_utils/initialize-clp-metadata-db.py
```diff
@@ -95,6 +95,16 @@ def main(argv):
                 """
             )
 
+            metadata_db_cursor.execute(
+                f"""
+                CREATE TABLE IF NOT EXISTS `{table_prefix}column_metadata_default` (
+                    `name` VARCHAR(512) NOT NULL,
+                    `type` BIGINT NOT NULL,
```
kirkrodrigues, line 102: Why `BIGINT`? Do we expect that many types?

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -322,10 +322,21 @@ def run_clp(
...
+                    str(clp_home / "bin" / "indexer"),
+                    "--db-config-file",
+                    str(db_config_file_path),
+                    "default",  # hardcode the table name for now
+                    archive_path,
+                ]
+                proc = subprocess.Popen(
+                    metadata_uploader_cmd, stdout=subprocess.DEVNULL, stderr=stderr_log_file
+                )
+                proc.wait()
```
kirkrodrigues, line 338: We should update the index *after* the archive has been uploaded, right?

kirkrodrigues, line 338: This is replacing the previous value of `proc`, right? It also seems like you could use `subprocess.run` instead?

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -322,10 +322,21 @@ def run_clp(
...
             None is stats or stats["id"] != last_archive_stats["id"]
         ):
+            archive_id = last_archive_stats["id"]
+            archive_path = archive_output_dir / archive_id
+            if StorageEngine.CLP_S == clp_storage_engine:
+                metadata_uploader_cmd = [
+                    str(clp_home / "bin" / "indexer"),
+                    "--db-config-file",
+                    str(db_config_file_path),
+                    "default",  # hardcode the table name for now
```
kirkrodrigues, line 332: I see why you want to have support for multiple tables but I don't think it makes sense to enforce that within the package since the package *doesn't* currently support multiple tables. Instead, I would suggest we:

* leave the column metadata table name as `{table_prefix}column_metadata`.
  * When we eventually add support for multiple tables/datasets in the clp package, 
we can modify this.
* modify the indexer process to take in the *full* table name rather than just the part after "column_metadata_". That way, we can easily control what the table should be called and there's less hard-coding of constants.

kirkrodrigues (changes requested):

components/clp-py-utils/clp_py_utils/initialize-clp-metadata-db.py
```diff
@@ -95,6 +95,16 @@ def main(argv):
                 """
             )
 
+            metadata_db_cursor.execute(
+                f"""
+                CREATE TABLE IF NOT EXISTS `{table_prefix}column_metadata_default` (
+                    `name` VARCHAR(512) NOT NULL,
+                    `type` INT NOT NULL,
```
kirkrodrigues, line 102: We need to make the same change in `indexer`, right?

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -356,6 +353,18 @@ def run_clp(
                         last_archive_stats,
                     )
                     db_conn.commit()
+                if StorageEngine.CLP_S == clp_storage_engine:
```
kirkrodrigues, line 356:
```suggestion

                if StorageEngine.CLP_S == clp_storage_engine:
```

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -356,6 +353,18 @@ def run_clp(
...
+                    indexer_cmd = [
```
kirkrodrigues, line 357:
```suggestion
                    # TODO: Since CLP doesn't currently support datasets but users of the index
                    # require a dataset name, we hardcode a name for now.
                    dataset_name = "default"
                    indexer_cmd = [
```

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -356,6 +353,18 @@ def run_clp(
...
+                        str(clp_home / "bin" / "indexer"),
+                        "--db-config-file",
+                        str(db_config_file_path),
+                        "default",  # hardcode the table name for now
```
kirkrodrigues, line 361:
```suggestion
                        dataset_name,
```

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -356,6 +353,18 @@ def run_clp(
...
+                        archive_path,
+                    ]
+                    subprocess.run(indexer_cmd, stdout=subprocess.DEVNULL, stderr=stderr_log_file)
```
kirkrodrigues, line 364:
```suggestion
                    try:
                        subprocess.run(
                            indexer_cmd,
                            stdout=subprocess.DEVNULL,
                            stderr=stderr_log_file,
                            check=True,
                        )
                    except subprocess.CalledProcessError:
                        logger.exception("Failed to index archive.")
```

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
kirkrodrigues: Based on our discussion offline, we decided not to do this but comment about why we need a table name.

kirkrodrigues (changes requested):

components/clp-py-utils/clp_py_utils/initialize-clp-metadata-db.py
wraymo: Sure, I will submit a PR later.

wraymo: What if we use TINYINT?

kirkrodrigues: I'm fine with TINYINT. Yeah, let's add the necessary binding in MySQLParamBindings.

kirkrodrigues (approved):
For the PR title, how about:

```
feat(clp-json): Index each field's name and type in the metadata DB after generating an archive.
```

PR #698, opened 2025-01-28
feat(core-clp)!: Change archive format version from a counter to a SemVer value.

davemarco:
# Description
This PR changes the archive version format from a uint_16 counter to a uint_32 semver. **It is breaking**

The major version is allocated 8 bits,  and the minor version is allocated 8 bits, the patch version is allocated 16 bits.

The main purpose of this PR is to bring the OSS main branch into alignment with a private branch, to make the reading each others archives possible. 

I removed the development version flag since it isn't really compatible with semver. Moreover, it seemed unused as all released versions were "development".

### Version

The new starting version will be 0.1.0, and will be independent from the existing private branch version 0.1.1.  The following describes the new versioning system for public and private branches.

We can now support the following cases:

|                       | Private archive format | Public archive format  |
|-----------------------|------------------------|------------------------|
| Private code can read | Yes                    | No                     |
| Public code can read  | Yes                    | Yes                    |

Achieved using two version numbers as follows.

| Public archive version | Private archive version | Meaning                                                                     |
|------------------------|-------------------------|-----------------------------------------------------------------------------|
| 0.0.0                  | x.y.z                   | A private archive version that's incompatible with any public version.      |
| a.b.c                  | x.y.z                   | A private archive version that's compatible with a specific public version. |
| a.b.c                  | 0.0.0                   | A public archive version. The private archive version is ignored.           |

Note how we treat `0.0.0` as an invalid version number and instead use it to indicate incompatibility.

The private version number will be stored as new field in the archive metadata(To be added in PR #700); however, the public code can ignore it entirely, while the private code can use it to determine the private archive format
version, without interfering with the public archive format version.

# Validation performed
Compressed and decompressed an archive successfully. Verified, as expected, that a failure occurs when reading older archives. 

## Summary by CodeRabbit

- **New Features**
	- Enhanced archive format versioning with explicit major, minor, and patch version support.
	- Expanded archive format version type to support a larger range of version numbers.

- **Refactor**
	- Updated archive format version constant structure for clarity.
	- Modified type definition for archive format version to use a 32-bit integer.
	- Changed initialization of archive format version in metadata to the default version.

- **Technical Improvement**
	- Improved version tracking mechanism with a more granular versioning approach.
	- Updated compatibility checks for archive format version.

### DIFF
--- a/components/core/src/clp/streaming_archive/Constants.hpp
+++ b/components/core/src/clp/streaming_archive/Constants.hpp
@@ -19,6 +19,7 @@ constexpr uint8_t VersionMajor{0};
 constexpr uint8_t VersionMinor{1};
 constexpr uint16_t VersionPatch{0};
 constexpr archive_format_version_t Version{VersionMajor << 24 | VersionMinor << 16 | VersionPatch};
+}  // namespace cArchiveFormatVersion
 
 namespace cMetadataDB {
 constexpr char ArchivesTableName[] = "archives";


haiqi96:
Looks reasonable. Does @LinZhihao-723 agree with the version number 0.1.2?

LinZhihao-723:
> Looks reasonable. Does @LinZhihao-723 agree with the version number 0.1.2?

The private branch is using `0.1.1`. Considering we don't port all archive format changes from the private branch and also have new changes in the OSS archive format, I can see there will be problems if we use 0.1.2 here. Can we pick a different minor version? Either `0.2.x` or `0.0.x`?
@kirkrodrigues what do you think?

LinZhihao-723:
> > Can we pick a different minor version? Either `0.2.x` or `0.0.x`? 
> 
> 
> 
> Note that current behaviour is to throw an exception when the version does not match exactly. If versions aren't the same, we will need some other behaviour. Maybe checking only if major/minor versions match? @LinZhihao-723 
> 
> 
> 
> It potentially possible to add an extra field at the end in open source metadata like "branch": "OSS" (default to "branch":"private"). To see if the version is for open source or private. However, the behaviour is not well defined in msgpack  docs so would prefer to avoid. It also only seems to work for last serialized field. 
> 
> 

Yeah we might some special handling to check if a version number is supported. We have the same problem in IR version. I think the solution should be integrated a third-party SemVer checker to handle it in future PRs.
For the extra field, I prefer to not include it and rely on the given version.

kirkrodrigues:
It's somewhat inevitable that as long as we deploy private, customized versions of CLP, then at some point, the private and public formats will diverge (although, we should try to avoid that for as long as possible). In these cases, using a single set of version numbers for both private and public formats will be too cumbersome and confusing, so I think we should try to support some way of using namespaces to separate private and public versions.

I think the conventional solution namespacing versions is to use the prerelease part of a semantic version, which in the case of the archive format, would be stored as a field in the archive metadata. Since we're making a breaking change to the public archive format, it's probably a good time to add this field to the archive metadata and then figure out some way to support compatibility for the private archive format. In the worst case, we would probably need to maintain a fork of the code for private and public formats, but since our current goal is to unify the formats, we can hopefully avoid that soon.

If we go with the above proposal, then perhaps we can start our version numbers at 0.2.0? Ideally, we'd pick 0.1.0, but 0.2.0 is probably a good compromise to avoid overlapping with the private format version and to indicate that the first public format using semver isn't really the first public format in existence.

LinZhihao-723:
I agree with @kirkrodrigues' proposal. Can we add this into the single file archive design doc?

haiqi96 (dismissed):
Looks good to me.

Just need to figure out the title.

components/core/src/clp/streaming_archive/Constants.hpp
```diff
@@ -15,6 +14,40 @@ constexpr char cMetadataFileName[] = "metadata";
...
+// | 0.0.0           | x.y.z            |
+//
+constexpr uint8_t cPublicVersionMajor{0};
+constexpr uint8_t cPublicVersionMinor{1};
+constexpr uint16_t cPublicVersionPatch{0};
+constexpr archive_format_version_t cPublicVersion{
+        cPublicVersionMajor << 24 | cPublicVersionMinor << 16 | cPublicVersionPatch
+};
+constexpr archive_format_version_t cNullVersion{0};
+constexpr archive_format_version_t cPrivateVersion{cNullVersion};
```
kirkrodrigues, line 48: Based on my original comment:

> a field that's unused and meaningless to the public code but can be used by the private code

I was expecting we'd implement the private version as a reserved field in the archive metadata. But I guess we can't do that once we switch to MessagePack since each field in the archive metadata needs a key; is that why you're doing this here?

If so, how about this instead:

1. Let's call the field something like "ArchiveFormatCustomVersion" instead of "PrivateVersion".
2. Let's remove all of the discussion about public/private versions. As my original comment mentions, the public code should not need to care about private versions and so I don't even think it needs to document it. Strictly speaking, the private code or some other fork of the code can use the custom version field (or the combination of the public version and custom version) however it wants. It should be the job of the downstream code to document it.
3. Let's call the "public" version simply "ArchiveFormatVersion".

components/core/src/clp/streaming_archive/Constants.hpp
```diff
@@ -15,6 +14,40 @@ constexpr char cMetadataFileName[] = "metadata";
 constexpr char cMetadataDBFileName[] = "metadata.db";
 constexpr char cSchemaFileName[] = "schema.txt";
 
+namespace version {
```
kirkrodrigues, line 17: The current convention for constants in this file is to name the namespace as `cNamespace` and then the constants inside as `ConstantXxx`. I know that disagrees with our clang-tidy rules, but in this case, I would stick with the existing conventions to avoid confusion until we refactor this file to address the clang-tidy warnings (especially the refactoring will propagate into a lot of files).

If you really want to namespace the version fields (which is fair), how about cArchiveFormatVersion with internal constants "Default" and "Custom"? "Default" isn't the greatest name, but I can't think of something better.

kirkrodrigues (changes requested):

components/core/src/clp/streaming_archive/Constants.hpp
```diff
@@ -15,6 +14,40 @@ constexpr char cMetadataFileName[] = "metadata";
...
+// | Public  version | Private  version |
+// | --------------- | ---------------- |
+// | a.b.c           | x.y.z            |
+//
+// Private branch with version x.y.z incompatible with any public version
+//
+// | Public  version | Private  version |
+// | --------------- | ---------------- |
+// | 0.0.0           | x.y.z            |
+//
```
LinZhihao-723, line 40: Shall we consider to use `/**/` style docstrings instead of `//` which is for inline comments?

components/core/src/clp/streaming_archive/Constants.hpp
kirkrodrigues: > Also in general it seems like people are recommending to me to move to new clang-tidy even in old files. But maybe this is not our recommendation?

It's our recommendation in other cases. In this case, I'm suggesting an exception until we refactor all other constants / make a more long-term decision about how to refactor them.

> Also, I can just default initialize the custom version to 0 in the metadata class. Then we don't need to specify default or custom here. And can just use cVersion. Let me know

True. I like that idea.

components/core/src/clp/streaming_archive/Constants.hpp
```diff
@@ -15,6 +14,16 @@ constexpr char cMetadataFileName[] = "metadata";
 constexpr char cMetadataDBFileName[] = "metadata.db";
 constexpr char cSchemaFileName[] = "schema.txt";
 
+namespace cArchiveFormatVersion {
+constexpr uint8_t cDefaultVersionMajor{0};
+constexpr uint8_t cDefaultVersionMinor{1};
+constexpr uint16_t cDefaultVersionPatch{0};
+constexpr archive_format_version_t cDefaultVersion{
```
kirkrodrigues, line 21: If we keep these inside the namespace, we should remove the prefix `c` to follow the convention of the other constants in this file (again, this is an exception to our clang-tidy rules).

kirkrodrigues (changes requested):

kirkrodrigues (commented):
Can you update the PR description to reflect the final decisions/code?

For the PR title, how about:

```
feat(core-clp)!: Change archive format version from a counter to a SemVer value.
```

kirkrodrigues (approved):

PR #699, opened 2025-01-28
refactor(clp::streaming_compression): Use `clp::Array` in zstd decompressor to replace C-style arrays.

LinZhihao-723:
# Description

This PR uses `clp::Array` to replace C-style array (backed using a unique pointer) to modernize the implementation so that it matches our latest coding standard.

# Validation performed

- Unfortunately, we don't have unit tests to test the touched code path. I tested locally to ensure that:
  - Decompression by reading from a file works -> the decompressed results matched the original input before compression
  - Seeking from the beginning works as expected.

## Summary by CodeRabbit

- **Refactor**
	- Updated memory management for decompression buffers
	- Modernized buffer handling using standard library types
	- Improved error checking and initialization processes

- **Bug Fixes**
	- Enhanced validation for read buffer operations
	- Simplified buffer management to reduce potential memory-related issues

The changes focus on improving the internal implementation of the Zstandard decompression mechanism, making the code more robust and maintainable.

### DIFF
--- a/components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
+++ b/components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
@@ -3,21 +3,20 @@
 #include <algorithm>
 
 #include "../../Defs.h"
+#include "../../ErrorCode.hpp"
 #include "../../ReadOnlyMemoryMappedFile.hpp"
 #include "../../spdlog_with_specializations.hpp"
+#include "../../TraceableException.hpp"
 
 namespace clp::streaming_compression::zstd {
-Decompressor::Decompressor() : ::clp::streaming_compression::Decompressor{CompressorType::ZSTD} {
-    m_decompression_stream = ZSTD_createDStream();
+Decompressor::Decompressor()
+        : ::clp::streaming_compression::Decompressor{CompressorType::ZSTD},
+          m_decompression_stream{ZSTD_createDStream()},
+          m_unused_decompressed_stream_block_buffer{ZSTD_DStreamOutSize()} {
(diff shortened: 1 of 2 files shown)


components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
```diff
@@ -50,9 +49,13 @@ ErrorCode Decompressor::try_read(char* buf, size_t num_bytes_to_read, size_t& nu
                     }
                     break;
                 case InputType::ReaderInterface: {
+                    if (false == m_read_buffer.has_value()) {
+                        throw OperationFailed(ErrorCode_Corrupt, __FILENAME__, __LINE__);
+                    }
+                    auto& read_buffer{m_read_buffer.value()};
```
gibber9809, line 55: Is this check necessary? Seems like the open/close flow will force m_read_buffer.has_value() to always be true here.

gibber9809 (changes requested):
Minor comment, but otherwise looks good.

components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
LinZhihao-723: This is to:
- Prevent future developers from breaking the assumption about the current `open`/`close` flow (especially we don't have unit tests for this code path atm)
- Silent clang-tidy warnings

If any later benchmark shows this becomes the bottleneck, we can remove it by manually silent clang-tidy for line 55.

gibber9809 (approved):
LGTM. PR title seems fine as well.

PR #700, opened 2025-01-29
feat(core-clp)!: Migrate archive metadata file format to MessagePack.

davemarco:
# Description
**This PR is breaking**

MIgrates the archive metadata format from a byte stream to MessagePack Map (same as sfa spec). Using MessagePack allows us to more easily store variable length metadata like strings. In previous format, a size specifier was required for any variable length field. 

Note the metadata file will take up slightly more space, but likely small relative to archive. 

Lastly the **metadata size is now excluded from the compressed size.** Now that the metadata is variable length, it is not trivial to incorporate the metadata size in the metadata itself (there is a circularity problem). The simplest solution is just to remove it from compressed size value.

This PR is sfa related like: 
https://github.com/y-scope/clp/pull/698

# Validation performed
Tested compression/decompression of archive. 

## Summary by CodeRabbit

- **Refactor**
	- Updated the `ArchiveMetadata` class to use MessagePack serialization for more efficient file reading and writing.
	- Replaced the constructor with a static `create_from_file` method.
	- Simplified the object creation process for `ArchiveMetadata` in the `Archive::open` method.

- **New Features**
	- Added a default constructor for `ArchiveMetadata`.
	- Enhanced error handling for file reading and metadata creation.
	- Introduced a new method for writing `ArchiveMetadata` to a file.

- **Improvements**
	- Streamlined metadata serialization process.
	- Added a new member variable for improved metadata handling.

### DIFF
--- a/components/core/src/clp/streaming_archive/ArchiveMetadata.cpp
+++ b/components/core/src/clp/streaming_archive/ArchiveMetadata.cpp
@@ -49,15 +49,10 @@ void ArchiveMetadata::expand_time_range(epochtime_t begin_timestamp, epochtime_t
     }
 }
 
-void ArchiveMetadata::write_to_file(FileWriter& file_writer) {
+void ArchiveMetadata::write_to_file(FileWriter& file_writer) const {
     std::ostringstream buf;
     msgpack::pack(buf, *this);
     auto const& string_buf = buf.str();
-
-    size_t previous_metadata_size = m_metadata_size;
-    m_metadata_size = string_buf.size();
-    file_writer.write(string_buf.data(), m_metadata_size);
-
-    m_compressed_size = m_compressed_size - previous_metadata_size + m_metadata_size;
+    file_writer.write(string_buf.data(), string_buf.size());
 }
 }  // namespace clp::streaming_archive
--- a/components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
+++ b/components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
@@ -104,7 +104,7 @@ public:
      *
      * @param file_writer
      */
-    void write_to_file(FileWriter& file_writer);
+    void write_to_file(FileWriter& file_writer) const;
 
     MSGPACK_DEFINE_MAP(
             MSGPACK_NVP("archive_format_version", m_archive_format_version),
@@ -128,7 +128,6 @@ private:
     uint64_t m_dynamic_uncompressed_size{0};
     // The size of the archive
     uint64_t m_compressed_size{0};
-    uint64_t m_metadata_size{0};
     uint64_t m_dynamic_compressed_size{0};
 };
 }  // namespace clp::streaming_archive


components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -79,15 +92,45 @@ class ArchiveMetadata {
...
+            MSGPACK_NVP("archive_format_version", m_archive_format_version),
+            MSGPACK_NVP("variable_encoding_methods_version", m_variable_encoding_methods_version),
+            MSGPACK_NVP("variables_schema_version", m_variables_schema_version),
+            MSGPACK_NVP("compression_type", m_compression_type),
+            MSGPACK_NVP("creator_id", m_creator_id),
+            MSGPACK_NVP("begin_timestamp", m_begin_timestamp),
+            MSGPACK_NVP("end_timestamp", m_end_timestamp),
+            MSGPACK_NVP("uncompressed_size", m_uncompressed_size),
+            MSGPACK_NVP("compressed_size", m_compressed_size)
+    );
```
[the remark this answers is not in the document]
haiqi96: @davemarco is this an issue?

components/core/src/clp/streaming_archive/reader/Archive.cpp
```diff
@@ -38,7 +38,7 @@ void Archive::open(string const& path) {
     archive_format_version_t format_version{};
     try {
         FileReader file_reader{metadata_file_path};
-        ArchiveMetadata const metadata{file_reader};
+        auto const metadata = ArchiveMetadata::create_from_file_reader(file_reader);
```
haiqi96, line 41: since the FileReader goes out of scope right after the metadata is created, can the method takes metadata_file_path as its parameter and create a FileReader internally?

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -102,6 +145,9 @@ class ArchiveMetadata {
     // The size of the archive
     uint64_t m_compressed_size{0};
     uint64_t m_dynamic_compressed_size{0};
+    std::string m_variable_encoding_methods_version{ffi::cVariableEncodingMethodsVersion};
```
haiqi96, line 148: The new coding standard encourages using std::string_view where possible. is it possible to use string_view instead?

components/core/src/clp/streaming_archive/ArchiveMetadata.cpp
```diff
@@ -22,15 +26,24 @@ ArchiveMetadata::ArchiveMetadata(
...
-    file_reader.read_numeric_value(m_begin_timestamp, false);
-    file_reader.read_numeric_value(m_end_timestamp, false);
+auto ArchiveMetadata::create_from_file_reader(FileReader& file_reader) -> ArchiveMetadata {
+    struct stat file_stat{};
+    auto clp_rc = file_reader.try_fstat(file_stat);
+    if (clp::ErrorCode::ErrorCode_Success != clp_rc) {
+        throw OperationFailed(ErrorCode_Failure, __FILENAME__, __LINE__);
+    }
+
+    std::vector<char> buf(file_stat.st_size);
```
haiqi96, line 36: @LinZhihao-723 should we use clp::array instead?

components/core/src/clp/streaming_archive/ArchiveMetadata.cpp
```diff
@@ -22,15 +26,24 @@ ArchiveMetadata::ArchiveMetadata(
...
-    file_reader.read_numeric_value(m_uncompressed_size, false);
-    file_reader.read_numeric_value(m_compressed_size, false);
-    file_reader.read_numeric_value(m_begin_timestamp, false);
-    file_reader.read_numeric_value(m_end_timestamp, false);
+auto ArchiveMetadata::create_from_file_reader(FileReader& file_reader) -> ArchiveMetadata {
+    struct stat file_stat{};
+    auto clp_rc = file_reader.try_fstat(file_stat);
+    if (clp::ErrorCode::ErrorCode_Success != clp_rc) {
+        throw OperationFailed(ErrorCode_Failure, __FILENAME__, __LINE__);
+    }
```
haiqi96, line 34: another way to write this code is:
Do we prefer one way to another? @LinZhihao-723 

```suggestion
    auto clp_rc = file_reader.try_fstat(file_stat);
    if (auto const clp_rc = file_reader.try_fstat(file_stat);
        clp::ErrorCode::ErrorCode_Success != clp_rc) {
        throw OperationFailed(ErrorCode_Failure, __FILENAME__, __LINE__);
    }
```

components/core/src/clp/streaming_archive/ArchiveMetadata.cpp
```diff
@@ -22,15 +26,24 @@ ArchiveMetadata::ArchiveMetadata(
...
+
+    std::vector<char> buf(file_stat.st_size);
+    clp_rc = file_reader.try_read_exact_length(buf.data(), buf.size());
+    if (clp::ErrorCode::ErrorCode_Success != clp_rc) {
+        throw OperationFailed(ErrorCode_Failure, __FILENAME__, __LINE__);
+    }
```
haiqi96, line 40: ditto, maybe can be rewritten as
```suggestion
    clp_rc = file_reader.try_read_exact_length(buf.data(), buf.size());
    if (auto const clp_rc = file_reader.try_read_exact_length(buf.data(), buf.size());
        clp::ErrorCode::ErrorCode_Success != clp_rc) {
        throw OperationFailed(ErrorCode_Failure, __FILENAME__, __LINE__);
    }
```

components/core/src/clp/streaming_archive/ArchiveMetadata.cpp
```diff
@@ -22,15 +26,24 @@ ArchiveMetadata::ArchiveMetadata(
...
+
+    ArchiveMetadata metadata;
+    msgpack::object_handle oh = msgpack::unpack(buf.data(), buf.size());
+    msgpack::object obj = oh.get();
+    obj.convert(metadata);
+    return metadata;
```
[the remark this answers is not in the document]
haiqi96: I kind of feel throwing the returned RC is better since it gives user more info, thoughts?

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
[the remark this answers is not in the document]
haiqi96: Hm, I think either works. I guess we can omit creation_idx for now, but add a comment to explain it's for format consistency. Then we can open another issue tracking the effort to handle this field properly in private branch

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -40,13 +46,20 @@ class ArchiveMetadata {
...
-     * @param file_reader
+     * Reads serialized MessagePack data from open file, unpacks it into an
+     * `ArchiveMetadata` instance.
+     *
+     * @param file_reader Reader for the file containing archive metadata.
+     * @return The created instance.
+     * @throw `ArchiveMetadata::OperationFailed` if stat or read operation on metadata file fails.
+     * @throw `msgpack::unpack_error` if data cannot be unpacked into MessagePack object.
+     * @throw `msgpack::type_error` if MessagePack object can't be converted to
+     * `ArchiveMetadata`.
```
haiqi96, line 59: can it be at end of the preivious line?

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -79,15 +92,45 @@ class ArchiveMetadata {
...
+
     /**
      * Expands the archive's time range based to encompass the given time range
      * @param begin_timestamp
      * @param end_timestamp
      */
     void expand_time_range(epochtime_t begin_timestamp, epochtime_t end_timestamp);
 
+    /**
+     * Packs `ArchiveMetadata` to MessagePack and writes to open file.
```
haiqi96, line 115:
```suggestion
     * Packs `ArchiveMetadata` to MessagePack and writes to the open file.
```

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -79,15 +92,45 @@ class ArchiveMetadata {
...
+     *
+     * @param file_writer Writer for archive metadata file.
+     * @throw FileWriter::OperationFailed if failed to write.
+     */
     void write_to_file(FileWriter& file_writer) const;
 
+    MSGPACK_DEFINE_MAP(
```
haiqi96, line 122: Does order matter?

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -28,6 +33,7 @@ class ArchiveMetadata {
     };
 
     // Constructors
+    ArchiveMetadata() = default;
```
haiqi96, line 36: Are we using the default constructor anywhere?

haiqi96: Thanks! a comment would definitely help.

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
haiqi96: Thanks for the clarification.

components/core/src/clp/streaming_archive/reader/Archive.cpp
haiqi96: >It will move some of the code (the read and stat) into archive reader, but then this constructor can also be used for sfa.

Is this planned in a future PR? I think it makes sense, but the change you propose might be easier to reason about when we are actually making the changes that include sfa.

haiqi96 (dismissed):
Looks good to me. as long as you ensure that the PR is tested.

for title, how about

feat(core-clp)!: Migrate archive metadata to MessagePack format and add new archive encoding fields.

components/core/src/clp/streaming_archive/ArchiveMetadata.cpp
```diff
@@ -22,15 +29,27 @@ ArchiveMetadata::ArchiveMetadata(
...
+        clp::ErrorCode::ErrorCode_Success != clp_rc)
+    {
+        throw OperationFailed(clp_rc, __FILENAME__, __LINE__);
+    }
+
+    clp::Array<char> buf(file_stat.st_size);
+    if (auto const clp_rc = file_reader.try_read_exact_length(buf.data(), buf.size());
+        clp::ErrorCode::ErrorCode_Success != clp_rc)
+    {
+        throw OperationFailed(clp_rc, __FILENAME__, __LINE__);
```
kirkrodrigues, line 45: Generally, exceptions should only be used for exceptional circumstances. I don't think corrupt data is exceptional since it can occur under normal circumstances (e.g., an interrupted download, a drive failure, etc.). Thus, I think this method should return these errors as error codes. To support a return value that can be an error code OR an object, you can use outcome. See `clp::ffi::get_schema_subtree_bitmap` for an example.

components/core/src/clp/streaming_archive/ArchiveMetadata.cpp
```diff
@@ -22,15 +29,27 @@ ArchiveMetadata::ArchiveMetadata(
...
-    file_reader.read_numeric_value(m_archive_format_version, false);
-    file_reader.read_numeric_value(m_creator_id_len, false);
-    file_reader.read_string(m_creator_id_len, m_creator_id, false);
-    file_reader.read_numeric_value(m_creation_idx, false);
-    file_reader.read_numeric_value(m_uncompressed_size, false);
-    file_reader.read_numeric_value(m_compressed_size, false);
-    file_reader.read_numeric_value(m_begin_timestamp, false);
-    file_reader.read_numeric_value(m_end_timestamp, false);
+auto ArchiveMetadata::create_from_file(std::string_view path) -> ArchiveMetadata {
+    auto file_reader = FileReader(std::string(path));
```
kirkrodrigues, line 33:
```suggestion
     FileReader file_reader{std::string(path)};
```

* Using `auto` in this case is more verbose and unnecessary since we need to explicitly specify the type anyway.
* We should also use brace init.

components/core/src/clp/streaming_archive/ArchiveMetadata.cpp
```diff
@@ -22,15 +29,27 @@ ArchiveMetadata::ArchiveMetadata(
...
+    struct stat file_stat{};
+    if (auto const clp_rc = file_reader.try_fstat(file_stat);
+        clp::ErrorCode::ErrorCode_Success != clp_rc)
+    {
+        throw OperationFailed(clp_rc, __FILENAME__, __LINE__);
+    }
+
+    clp::Array<char> buf(file_stat.st_size);
```
kirkrodrigues, line 41:
```suggestion
    clp::Array<char> buf{static_cast<size_t>(file_stat.st_size)};
```

* We should use brace init.
* Brace init results in a compilation error due to the implicit conversion (precisely what brace init is meant to catch) forcing us to explicit convert the value.

components/core/src/clp/streaming_archive/ArchiveMetadata.cpp
```diff
@@ -22,15 +29,27 @@ ArchiveMetadata::ArchiveMetadata(
...
+    if (auto const clp_rc = file_reader.try_read_exact_length(buf.data(), buf.size());
+        clp::ErrorCode::ErrorCode_Success != clp_rc)
+    {
+        throw OperationFailed(clp_rc, __FILENAME__, __LINE__);
+    }
+
+    ArchiveMetadata metadata;
+    msgpack::object_handle oh = msgpack::unpack(buf.data(), buf.size());
```
kirkrodrigues, line 49: Nit: How about `obj_handle`?

components/core/src/clp/streaming_archive/ArchiveMetadata.cpp
```diff
@@ -22,15 +29,27 @@ ArchiveMetadata::ArchiveMetadata(
...
+    msgpack::object obj = oh.get();
```
kirkrodrigues, line 50:
```suggestion
    msgpack::object_handle const oh = msgpack::unpack(buf.data(), buf.size());
    msgpack::object const obj = oh.get();
```

Resolves clang-tidy warnings.

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -102,6 +150,9 @@ class ArchiveMetadata {
     // The size of the archive
     uint64_t m_compressed_size{0};
     uint64_t m_dynamic_compressed_size{0};
+    std::string_view m_variable_encoding_methods_version{ffi::cVariableEncodingMethodsVersion};
+    std::string_view m_variables_schema_version{ffi::cVariablesSchemaVersion};
+    std::string_view m_compression_type{cCompressionTypeZstd};
```
kirkrodrigues, line 155: Can we add these in another PR? Switching to MessagePack and adding the new fields are kind of unrelated changes.

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -2,13 +2,19 @@
 #define CLP_STREAMING_ARCHIVE_ARCHIVEMETADATA_HPP
 
 #include <cstdint>
+#include <string_view>
 
 #include "../Defs.h"
+#include "../ffi/encoding_methods.hpp"
 #include "../FileReader.hpp"
```
kirkrodrigues, line 9:
```suggestion
```

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -2,13 +2,19 @@
...
 #include "../FileWriter.hpp"
 #include "Constants.hpp"
+#include "msgpack.hpp"
```
kirkrodrigues, line 12:
```suggestion
#include <msgpack.hpp>
```

Should use `<>` for library includes.

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -28,6 +34,11 @@ class ArchiveMetadata {
     };
 
     // Constructors
+    // The class must be default constructible to convert from msgpack::object in
+    // `create_from_file` method.
```
kirkrodrigues, line 38:
```suggestion
    // We need a default constructor to convert from a msgpack::object in `create_from_file`. See
```

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -40,13 +51,19 @@ class ArchiveMetadata {
             uint64_t creation_idx
     );
 
+    // Methods
     /**
-     * Constructs a metadata object and initializes it from the given file reader
-     * @param file_reader
+     * Reads serialized MessagePack data from open file, unpacks it into an
+     * `ArchiveMetadata` instance.
```
kirkrodrigues, line 57:
```suggestion
     * Reads serialized MessagePack data from a file and unpacks it into an `ArchiveMetadata`
     * instance.
```

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -40,13 +51,19 @@ class ArchiveMetadata {
...
+     *
+     * @param path Path to archive metadata file.
```
kirkrodrigues, line 59:
```suggestion
     * @param path
```

kirkrodrigues, line 59: How about `file_path`?

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -79,15 +96,46 @@ class ArchiveMetadata {
...
      * Expands the archive's time range based to encompass the given time range
      * @param begin_timestamp
      * @param end_timestamp
      */
     void expand_time_range(epochtime_t begin_timestamp, epochtime_t end_timestamp);
 
+    /**
+     * Packs `ArchiveMetadata` to MessagePack and writes to the open file.
+     *
+     * @param file_writer Writer for archive metadata file.
```
kirkrodrigues, line 121:
```suggestion
     * Packs this instance into a MessagePack object and writes it to the open file.
     *
     * @param file_writer
```

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -79,15 +96,46 @@ class ArchiveMetadata {
...
+     * @throw FileWriter::OperationFailed if failed to write.
```
kirkrodrigues, line 122:
```suggestion
```

Generally we don't document thrown exceptions, except at the boundary of a library, for the following reasons:

* C++ exceptions propagate through functions, so tracing every exception that can be thrown to document it is impractical. The best you could do is catch in every method and then re-throw (this could be the strategy to use at the boundary of a library's interface).
* As mentioned in the other comment, C++ exceptions should be used in exceptional circumstances---this is so that callers don't need to be expected to gracefully handle every exception (instead, they can simply log the error and exit).

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -2,13 +2,19 @@
...
 #include "../Defs.h"
+#include "../ffi/encoding_methods.hpp"
 #include "../FileReader.hpp"
 #include "../FileWriter.hpp"
 #include "Constants.hpp"
+#include "msgpack.hpp"
 
 namespace clp::streaming_archive {
+
+static constexpr std::string_view cCompressionTypeZstd = "ZSTD";
```
kirkrodrigues, line 16:
```suggestion
constexpr std::string_view cCompressionTypeZstd{"ZSTD"};
```

kirkrodrigues (changes requested):

components/core/src/clp/streaming_archive/ArchiveMetadata.cpp
kirkrodrigues: Gotcha. So from a guideline perspective, we should:

1. Catch the MessagePack exceptions and return an error
    1. I think the FileReader exceptions should be truly exception, so we can leave them as exceptions for the top-layer of code to catch; but if not, then we should ideally use calls that return an error.
2. Remove the external try-catch and change it into an error check
3. Change Archive::open to return an error

I think (1) & (2) is doable in this PR (but feel free to disagree if it seems like too much code to change). (3) will probably be larger than the scope of this PR. So if we just do (1) & (2), to be able to propagate the error to the caller, I'd throw an exception.

Let's just make sure we're on the same page with the above before we go ahead with the change (don't want you to have to change too much).

kirkrodrigues: Uh, we usually use the generic error codes (I gave a bad example, lol), but I guess since it isn't clean, let's leave it for now. I'll review the rest of the PR when I get a chance today.

components/core/src/clp/streaming_archive/ArchiveMetadata.cpp
```diff
@@ -1,5 +1,12 @@
 #include "ArchiveMetadata.hpp"
 
+#include <sys/stat.h>
+
+#include <fmt/core.h>
+
+#include "../Array.hpp"
+#include "FileReader.hpp"
```
kirkrodrigues, line 8:
```suggestion
#include "../FileReader.hpp"
```

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -2,13 +2,16 @@
...
+
+#include <msgpack.hpp>
 
 #include "../Defs.h"
-#include "../FileReader.hpp"
 #include "../FileWriter.hpp"
 #include "Constants.hpp"
 
 namespace clp::streaming_archive {
+
```
kirkrodrigues, line 14:
```suggestion
```

components/core/src/clp/streaming_archive/ArchiveMetadata.hpp
```diff
@@ -40,13 +47,19 @@ class ArchiveMetadata {
...
-     * Constructs a metadata object and initializes it from the given file reader
-     * @param file_reader
+     * Reads serialized MessagePack data from a file and unpacks it into an `ArchiveMetadata`
+     * instance.
+     *
+     * @param file_path
+     * @return The created instance.
+     * @throw `ArchiveMetadata::OperationFailed` if stat or read operation on metadata file fails.
+     * @throw `msgpack::unpack_error` if data cannot be unpacked into MessagePack object.
+     * @throw `msgpack::type_error` if MessagePack object can't be converted to `ArchiveMetadata`.
```
kirkrodrigues, line 59:
```suggestion
     * @throw `ArchiveMetadata::OperationFailed` if `stat` fails or the file couldn't be read.
     * @throw `msgpack::unpack_error` if the data cannot be unpacked into a MessagePack object.
     * @throw `msgpack::type_error` if the MessagePack object can't be converted to an
     * `ArchiveMetadata` instance.
```

kirkrodrigues (changes requested):

kirkrodrigues (approved):
For the PR title, how about:

```
feat(core-clp)!: Migrate archive metadata file format to MessagePack.
```

kirkrodrigues:
How about let's only add the metadata file's size to `m_compressed_size` inside `write_to_file`?

kirkrodrigues:
> > How about let's only add the metadata file's size to `m_compressed_size` inside `write_to_file`?
> 
> write_to_file is actually [called](https://github.com/y-scope/clp/blob/121b717696124006beb9ae4cdde0075c472eca51/components/core/src/clp/streaming_archive/writer/Archive.cpp#L636) whenever a segment is closed, so they would accumulate.

Can we do the following?
* Add a new member that just holds the size of the metadata file and which is set only in `write_to_file`.
* Update `get_compressed_size_bytes` to include this new member.

> Even if called only once, or some behavior to handle that issue, there could also be issues with global_metadatadb update. Say we update global metadataDB before actually writing to file.

* I think it would be fine for the size in the global metadata database to only reflect the size on disk, i.e., post write?
* We should generally ensure that the global metadata is updated after writes to file. If we're updating the metadata before something is on disk, that is a bug and should be fixed.

kirkrodrigues (dismissed):

kirkrodrigues (approved):
Approved pending Haiqi's final approval.

haiqi96 (approved):
Quickly took a look. The final version looks reasonable.

PR #701, opened 2025-01-29
chore(core): Add utfcpp library as a submodule.

gibber9809:
# Description
This PR adds the utfcpp library to this project as a submodule. This library provides the capability to replace invalid sequences in utf-8 strings with a replacement character; this functionality will be used in a follow-up PR to create a validating UTF-8 parser class that aims to efficiently validate UTF-8 and allow users to control how illegal UTF-8 is handled (e.g. by returning an error, or replacing invalid code sequences with a replacement character).

This library was chosen because it seems mature, and the main alternative (the [official icu library](https://github.com/unicode-org/icu)) seems a bit heavy-weight to pull in for this use-case. simdjson is not viable here because its invalid sequence replacment logic is tied to JSON unescaping, and simdutf can not be used because (surprisingly) it only provides UTF validation and conversion not invalid UTF-8 replacement.

# Validation performed
* Validated that updated `deps:core` task passes

## Summary by CodeRabbit

- **New Features**
  - Added UTF-8 C++ library (utfcpp) as a project dependency
  - Integrated new submodule for UTF-8 processing support

- **Chores**
  - Updated dependency management configuration
  - Added checksum validation for the new library
  - Configured automatic download of utfcpp library version 4.0.6

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -29,3 +29,9 @@
 [submodule "components/log-viewer-webui/yscope-log-viewer"]
 	path = components/log-viewer-webui/yscope-log-viewer
 	url = https://github.com/y-scope/yscope-log-viewer.git
+[submodule "components/core/submodules/--force"]
+	path = components/core/submodules/--force
+	url = https://github.com/nemtrif/utfcpp.git
+[submodule "components/core/submodules/utfcpp"]
+	path = components/core/submodules/utfcpp
+	url = https://github.com/nemtrif/utfcpp.git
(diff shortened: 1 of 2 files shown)


LinZhihao-723 (commented):
The changes look good to me. A few things to comment:
- I understand that we're having discussions offline about whether we should enforce the UTF8 encoding, but I think it is still good to have utils that can efficiently validate UTF8 byte sequence with error handling options (such as replacing invalid chars by the dedicated placeholder). Therefore, I think this PR should be merged and we should add the planned UTF8 validation features into CLP core.
- That said, can we create a feature request with a description of our plan? That would make it easier to track the implementation status. 

I will approve the PR once the feature request is ready

LinZhihao-723 (approved):
The PR title lgtm.

PR #705, opened 2025-01-30
feat(core-clp): Postpone GlobalMetadataDB updates until archive writing finishes (fixes #685).

### LINKED ISSUE #685: Postpone all updates to the GlobalMetadataDB until after archive writing is complete.
### Task Overview
I believe the following enhancement is required for unstructured single-file archives implementation. @kirkrodrigues asked if I could submit this proposal to see if the change makes sense to @jackluo923 @LinZhihao-723 @haiqi96

### High Level Background
In unstructured CLP, the GlobalMetadataDB stores archive-level metadata and file data (see schemas in later comment). Archive-level metadata is created when the archive is opened, while file entries are written as segments are closed. According to @kirkrodrigues, GlobalMetadataDB is read by clg, allowing files to be searchable immediately after their entries are written to the database. As a result, files that have been written can be searched and read, even while CLP continues compressing the remainder of the archive.

### Incompatibility with S3
While concurrent reading and compressing is attractive, it is not compatible with single file archive written to an object store. In such cases, the archive is only written to the object store once the whole archive is complete.  However, entries are written to GlobalMetadataDB when the archive is created and when segments are closed. Thus there is an discrepancy in the GlobalMetadataDB and the data accessible by clg (search workers), potentially leading to errors. This is not a problem in the package currently since we are using a FUSE layer, and segments are written as they are created. 

### Why is this not an issue in private branch?
The GlobalMetadataDB was removed in private branch. While we considered this option as well for open source, the GlobalMetadataDB is required for clp package, and thus keeping it seems more desirable for eventually using the single-file format in the package as well (like clp-s). 

### Possible implementation

Only update the GlobalMetadataDB once the archive is complete. **We will lose concurrent writing/reading feature**. It was also considered to make two paths for multi-file archive, and single-file archive, but was disregarded since it just adds additional complexity.

davemarco:
# Description

See #685 for high level PR goals.

### re: implementation

Removed all writes during serialization to Global metadata DB.

Instead, file metadata is maintained in a new list (previously was deleted when a segment was closed), and all file metadata is written to Global metadata DB when archive finished. 

I explored other options, but in the end, this seemed to be the cleanest. 

# Validation performed
- Compared contents of global metadata DB on compressed loghub/Hadoop (lots of files) for previous commit and this PR, and validated contents were the same.
- Tested compression/decompression of archive
- Ran unit test
- Ran clg
- Added assertion (deleted now in PR) to test that file data was actually deallocated before adding to list. 

## Summary by CodeRabbit

- **New Features**  
  - Introduced a unified approach to handling metadata updates during file archiving, enhancing overall consistency and reliability.
  - Added a new method for updating global metadata related to archived files.

- **Refactor**  
  - Streamlined the archival workflow by consolidating metadata operations, resulting in a clearer and more efficient process for updating file information.  
  - Enhanced resource management by ensuring proper deletion of written file metadata after updates.

### DIFF
--- a/components/core/src/clp/streaming_archive/writer/Archive.cpp
+++ b/components/core/src/clp/streaming_archive/writer/Archive.cpp
@@ -239,26 +239,15 @@ void Archive::close() {
         throw OperationFailed(ErrorCode_Failure, __FILENAME__, __LINE__);
     }
     m_global_metadata_db->add_archive(m_id_as_string, m_local_metadata.value());
-
-    auto const file_it = archive_metadata_db.get_file_iterator(
-        cEpochTimeMin,
-        cEpochTimeMax,
-        "",
-        "",
-        false,
-        cInvalidSegmentId,
-        false
-    );
-
-        while (file_it->has_next()) {
-        file_it->next();
-
-
-    m_global_metadata_db->add_metadata_for_files_from_archive_metadata_db(
+    m_global_metadata_db->update_metadata_for_files(
             m_id_as_string,
-            m_metadata_db
+            m_clean_files
     );
+    for (auto file : m_clean_files) {
+        delete file;
+    }
     m_global_metadata_db->close();
+
     m_global_metadata_db = nullptr;
 
     m_metadata_db.close();
@@ -581,6 +570,7 @@ void Archive::persist_file_metadata(vector<File*> const& files) {
     // Mark files' metadata as clean
     for (auto file : files) {
         file->mark_metadata_as_clean();
+        m_clean_files.emplace_back(file)
     }
 }
 
@@ -618,7 +608,7 @@ void Archive::close_segment_and_persist_file_metadata(
     update_metadata();
 
     for (auto file : files) {
-        delete file;
+        m_clean_files.emplace_back(std::move(file));
     }
     files.clear();
 }
--- a/components/core/src/clp/streaming_archive/writer/Archive.cpp
+++ b/components/core/src/clp/streaming_archive/writer/Archive.cpp
@@ -644,7 +644,7 @@ auto Archive::update_global_metadata() -> void {
     }
     m_global_metadata_db->add_archive(m_id_as_string, m_local_metadata.value());
     m_global_metadata_db->update_metadata_for_files(m_id_as_string, m_files_written);
-    for (auto *file : m_files_written) {
+    for (auto* file : m_files_written) {
         delete file;
     }
     m_global_metadata_db->close();
--- a/components/core/src/clp/streaming_archive/writer/Archive.cpp
+++ b/components/core/src/clp/streaming_archive/writer/Archive.cpp
@@ -237,10 +237,10 @@ void Archive::close() {
     update_global_metadata();
     m_global_metadata_db = nullptr;
 
-    for (auto* file : m_files_written) {
+    for (auto* file : m_file_metadata_for_global_update) {
         delete file;
     }
-    m_files_written.clear();
+    m_file_metadata_for_global_update.clear();
 
     m_metadata_db.close();
 
@@ -593,7 +593,7 @@ void Archive::close_segment_and_persist_file_metadata(
 
     for (auto file : files) {
         file->mark_as_in_committed_segment();
-        m_files_written.emplace_back(file);
+        m_file_metadata_for_global_update.emplace_back(file);
     }
 
     persist_file_metadata(files);
@@ -647,7 +647,7 @@ auto Archive::update_global_metadata() -> void {
         throw OperationFailed(ErrorCode_Failure, __FILENAME__, __LINE__);
     }
     m_global_metadata_db->add_archive(m_id_as_string, m_local_metadata.value());
-    m_global_metadata_db->update_metadata_for_files(m_id_as_string, m_files_written);
+    m_global_metadata_db->update_metadata_for_files(m_id_as_string, m_file_metadata_for_global_update);
     m_global_metadata_db->close();
 }
 
--- a/components/core/src/clp/streaming_archive/writer/Archive.cpp
+++ b/components/core/src/clp/streaming_archive/writer/Archive.cpp
@@ -157,10 +157,6 @@ void Archive::open(UserConfig const& user_config) {
 
     m_global_metadata_db = user_config.global_metadata_db;
 
-    m_global_metadata_db->open();
-    m_global_metadata_db->add_archive(m_id_as_string, *m_local_metadata);
-    m_global_metadata_db->close();
-
     m_file = nullptr;
 
     // Open log-type dictionary
@@ -238,6 +234,13 @@ void Archive::close() {
 
     m_metadata_file_writer.close();
 
+    m_global_metadata_db->open();
+    if (false == m_local_metadata.has_value()) {
+        throw OperationFailed(ErrorCode_Failure, __FILENAME__, __LINE__);
+    }
+    m_global_metadata_db->add_archive(m_id_as_string, m_local_metadata.value());
+    m_global_metadata_db->copy_metadata_for_files_from_archive_metadata_db(m_id_as_string, m_metadata_db);
+    m_global_metadata_db->close();
     m_global_metadata_db = nullptr;
 
     m_metadata_db.close();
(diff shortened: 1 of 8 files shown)


components/core/src/clp/streaming_archive/writer/Archive.cpp
```diff
@@ -593,16 +589,12 @@ void Archive::close_segment_and_persist_file_metadata(
 
     for (auto file : files) {
         file->mark_as_in_committed_segment();
+        m_files_written.emplace_back(file);
```
haiqi96, line 592: just wonder, can we use something like m_files_written.insert(files.begin(), files.end())?

components/core/src/clp/streaming_archive/writer/Archive.cpp
```diff
@@ -647,6 +637,19 @@ void Archive::update_metadata() {
...
+auto Archive::update_global_metadata() -> void {
+    m_global_metadata_db->open();
+    if (false == m_local_metadata.has_value()) {
+        throw OperationFailed(ErrorCode_Failure, __FILENAME__, __LINE__);
+    }
+    m_global_metadata_db->add_archive(m_id_as_string, m_local_metadata.value());
+    m_global_metadata_db->update_metadata_for_files(m_id_as_string, m_files_written);
+    for (auto* file : m_files_written) {
+        delete file;
+    }
```
haiqi96, line 649: Just looking at the function name, one may not expect the function to call delete on m_files_written.

I feel it would be more clear to put these 3 lines out side of the function, or maybe update the funciton name to include the "delete" part

components/core/src/clp/streaming_archive/writer/Archive.hpp
```diff
@@ -279,6 +279,11 @@ class Archive {
      */
     void update_metadata();
 
+    /**
+     * Updates metadata in the global metadata database.
```
haiqi96, line 283: ditto comment for deleting m_files_written.

You might also need to mention about exceptions if m_global_metadata_db's api throws any.

components/core/src/clp/streaming_archive/writer/Archive.cpp
haiqi96: Can't say for sure which one is better, so I am fine to keep the emplace.
I do notice that emplace construct the object in place and insert copies the object. but in our case, I guess emplace will also somehow have to copy via the copy constructor.
Anyway, I don't think the performance really matter here. 

@LinZhihao-723 do you have any insights?

components/core/src/clp/streaming_archive/writer/Archive.hpp
```diff
@@ -316,6 +321,10 @@ class Archive {
     std::vector<File*> m_files_with_timestamps_in_segment;
     std::vector<File*> m_files_without_timestamps_in_segment;
 
+    // Data for all files in this collection has been deallocated, and should only
+    // contain metadata.
```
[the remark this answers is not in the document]
haiqi96: The comment makes sense to me. maybe we can rename the variables?

something like m_file_metadata_for_global_update.

components/core/src/clp/streaming_archive/writer/Archive.cpp
haiqi96: Ture, you are right. my bad

haiqi96:
> `m_is_written_out`

I see

components/core/src/clp/streaming_archive/writer/Archive.hpp
```diff
@@ -316,6 +321,11 @@ class Archive {
     std::vector<File*> m_files_with_timestamps_in_segment;
     std::vector<File*> m_files_without_timestamps_in_segment;
 
+    // Files in this collection only hold metadata. Files are added to this collection after
```
haiqi96, line 324: maybe move this comment to where the m_file_metadata is updated?
```suggestion
    // Files in this collection only hold metadata. Files are added to this collection after
```

components/core/src/clp/streaming_archive/writer/Archive.cpp
```diff
@@ -593,16 +593,12 @@ void Archive::close_segment_and_persist_file_metadata(
 
     for (auto file : files) {
         file->mark_as_in_committed_segment();
+        m_file_metadata_for_global_update.emplace_back(file);
```
haiqi96, line 596:
```suggestion
        // Files in this collection only hold metadata. Files are added to this collection after
        // `File::append_to_segment()` is called, which deallocates memory for timestamp,
        // logtype, and variable fields.
        m_file_metadata_for_global_update.emplace_back(file);
```

components/core/src/clp/streaming_archive/writer/Archive.hpp
kirkrodrigues, line 283:
```suggestion
     * Updates the archive's metadata in the global metadata database.
```

And we should probably update `update_metadata`'s docstring to something similar but replacing "global" with "local".

components/core/src/clp/streaming_archive/writer/Archive.cpp
```diff
@@ -593,16 +593,15 @@ void Archive::close_segment_and_persist_file_metadata(
 
     for (auto file : files) {
         file->mark_as_in_committed_segment();
+        // Files in this collection only hold metadata. Emplaced files have called method
+        // `File::append_to_segment()` which deallocates memory for timestamp,
+        // logtype, and variable fields.
+        m_file_metadata_for_global_update.emplace_back(file);
```
kirkrodrigues, line 599: This comment confused me a bit since this method doesn't seem to have anything to do with managing the lifecycle of the files (and the grammar is a bit odd). I think it would make more sense to add a comment above `m_file_metadata_for_global_update` in `Archive.hpp` indicating that the collection contains Files that have been written to the local archive and deallocated but whose metadata has not been persisted globally (rephrase it as appropriate).

components/core/src/clp/streaming_archive/writer/Archive.cpp
```diff
@@ -647,6 +644,19 @@ void Archive::update_metadata() {
     }
 }
 
+auto Archive::update_global_metadata() -> void {
```
kirkrodrigues, line 647: How about renaming `update_metadata` to `update_local_metadata` so that there's congruency?

components/core/src/clp/streaming_archive/writer/Archive.cpp
```diff
@@ -635,8 +634,6 @@ void Archive::update_metadata() {
     m_metadata_file_writer.seek_from_begin(0);
     m_local_metadata->write_to_file(m_metadata_file_writer);
 
-    m_global_metadata_db->update_archive_metadata(m_id_as_string, *m_local_metadata);
-
     if (m_print_archive_stats_progress) {
```
kirkrodrigues, line 637: This block should probably move into `Archive::close` since it may mislead the external caller if we indicate that an archive's stats have been updated, but that archive doesn't exist in the global metadata database (whereas it did before this change).

kirkrodrigues (changes requested):

components/core/src/clp/streaming_archive/writer/Archive.cpp
kirkrodrigues: The progress output is being [used by the package](https://github.com/y-scope/clp/blob/main/components/job-orchestration/job_orchestration/executor/compress/compression_task.py#L322) to keep track of when new archives have been generated. The package expects that the archive exists in the global metadata database when it receives the progress update. Progress is probably a poor name for this option at this point since its use is not really for humans per se.

kirkrodrigues: True, it does use only the last update. For the sake of safety, can you add a comment indicating that the stats are printed even though the archive doesn't exist in the global metadata database?

kirkrodrigues: > If you want to be more safe. I could potential add a duplicate stat printing when the archive is closed (maybe with a new field like complete) lmk

Good idea. Sgtm.

kirkrodrigues (approved):
For the PR title, how about:

```
feat(core-clp): Defer each archive's global metadata updates until it has been completed (resolves #685).
```

PR #707, opened 2025-01-30
docs(IRv2): Add documentation for current IRv2 implementation

AVMatthews:
# Description

Draft of documentation for IRv2 Implementation. 

Pulled Materials from IRv2 Design Doc and draft IRv2 Blog Post

Still need many images converted to Mermaid and some of the Mermaid visuals need to be cleaned up to be more readable.

Some visuals need auto generated keys to be integrated.

# Checklist

* [x ] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [ x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

---
  - https://app.asana.com/0/0/1212851383737245


LinZhihao-723 (changes requested):
Some high-level comments before we proceed:
- In the md file, let's wrap the line to ensure each line doesn't exceed 100-char limit. This helps increase the readability when directly editing the md file.
- iiuc, "IRv2" is the internal name we used. For the formal release, we should use "key-value pair IR stream" (like we did here: https://github.com/y-scope/clp-ffi-py?tab=readme-ov-file#using-key-value-pair-ir-streams)
- @kirkrodrigues do we need spec of how packets/IR units are serialized in this doc?

docs/src/dev-guide/design-key-value-pair-ir-stream.md
```diff
@@ -0,0 +1,503 @@
...
+CLP key-value pair IR format categorizes the kv-pairs of a log event into two categories:
+- **Auto-generated kv-pairs**: KV-pairs (e.g., timestamps, log levels, other metadata) that are
+  automatically generated by the logging library.
+- **User-generated kv-pairs**: Custom kv-pairs (e.g., log messages).
+
+This design prevents conflicts between auto-generated and user-generated keys by keeping them
+separate. Each IR stream maintains auto-generated keys and user-generated keys in two independent
+schema trees, ensuring clear differentiation and avoiding unintended interference.
+
+#### Example
```
LinZhihao-723, line 56: We need to refactor this example to showcase auto-gen and user-gen kv-pairs

kirkrodrigues (changes requested):
Ideally, these docs should explain the motivation behind the KV-pair IR format both at the high level as well as the individual pieces. Think about it from the perspective of a developer who wants to modify/work with the format---what is the minimum amount of information they need to understand why the format is what it is and how to extend it.

Although the PR already contains a lot of the pieces necessary for the docs, they're:

* missing some info (e.g., the schema tree diagram doesn't actually explain what the nodes are); remember that developers are time-constrained, so they likely won't read the clp-s paper.
* going bottom-up rather than top-down.
* missing the motivation for why we designed things in the way we did.

I would suggest structuring things roughly as follows:

* A concise but sufficient intro to clp-s' compression algorithm
  * This should explain how to convert the example JSON records into clp-s' format.
* Overview of the KV-pair IR format
* Motivation and specification of each part of the format (if this ends up being too convoluted, we could defer the specification to another page that we link to).

In the specification of each part, I would highlight the differences between the format and clp-s (e.g., the type differences).

Since this is a big task, I would suggest going piece by piece (e.g., update the intro to clp-s' compression algorithm first) and then we can review that before moving onto to the next section.

docs/src/dev-guide/design-key-value-pair-ir-stream.md
```diff
@@ -0,0 +1,984 @@
...
+log event down into a series of IR units: schema tree growth IR units describing changes in the 
+schema tree and log event IR units which consists of key and value packets. Although we will use 
+the terminology IR unit to describe the individual IR stream elements, all of the IR units 
+corresponding to a single log event would be sent in a single network packet to preserve the order 
+of the individual elements, since the order and location of the IR Units in the stream are used 
+inherently by the future log events. 
+
+So the overall format of the IR stream consists of 4 sections; a [magic number](#magic-number) 
+which indicates the stream is a KV Pair Stream and how many byte encoding is being used, a 
+[metadata IR unit](#metadata) which provides information necessary to properly deserailize the
```
LinZhihao-723, line 105: Metadata is not an IR unit.

docs/src/dev-guide/design-key-value-pair-ir-stream.md
```diff
@@ -0,0 +1,984 @@
...
+corresponding to a single log event would be sent in a single network packet to preserve the order 
+of the individual elements, since the order and location of the IR Units in the stream are used 
+inherently by the future log events. 
+
+So the overall format of the IR stream consists of 4 sections; a [magic number](#magic-number) 
+which indicates the stream is a KV Pair Stream and how many byte encoding is being used, a 
+[metadata IR unit](#metadata) which provides information necessary to properly deserailize the 
+stream later, a alternating set of [schema tree growth IR units](#schema-tree-node-insertion-unit) 
+and [log event IR units](#log-event-unit), and finally a 
+[end of stream IR unit](#end-of-stream-unit) to indicate you've reached the end of a valid KV Pair
```
LinZhihao-723, line 108: "end-of-stream IR unit" instead of "end of stream IR unit"

docs/src/dev-guide/design-key-value-pair-ir-stream.md
```diff
@@ -0,0 +1,984 @@
...
+schema tree and log event IR units which consists of key and value packets. Although we will use 
+the terminology IR unit to describe the individual IR stream elements, all of the IR units 
+corresponding to a single log event would be sent in a single network packet to preserve the order 
+of the individual elements, since the order and location of the IR Units in the stream are used 
+inherently by the future log events. 
+
+So the overall format of the IR stream consists of 4 sections; a [magic number](#magic-number) 
+which indicates the stream is a KV Pair Stream and how many byte encoding is being used, a 
+[metadata IR unit](#metadata) which provides information necessary to properly deserailize the 
+stream later, a alternating set of [schema tree growth IR units](#schema-tree-node-insertion-unit)
```
LinZhihao-723, line 106: "schema tree node insertion IR units" instead of "schema tree growth IR units"

docs/src/dev-guide/design-key-value-pair-ir-stream.md
```diff
@@ -0,0 +1,984 @@
...
+generated schema nodes, and then a set of 
+[key IR packets and corresponding value IR packets](#user-generated-schema-tree-node-id-value-pairs)
+ for each user generated key value pair in the log entry (???). To differnetiate between the auto 
+ and user generated node ids we use the 1's complement to encode the auto generated node ids. 
+ Rather than organizing the IR units so that each value comes directly after its key, we group the 
+ keys and values together. We found that in practice the similarities between keys and between 
+ values allows for better compression ratios when stored together. If a log record does not result 
+ in any new schema tree nodes, there will be no new schema tree growth IR units for that record. 
+ While we stream the IR units we keep the schema tree we’ve built in memory for easy reference, so 
+ all packets in the stream can reference the same set of schema tree nodes.
```
LinZhihao-723, line 121: As an overview, we might not need to cover these details.

docs/src/dev-guide/design-key-value-pair-ir-stream.md
```diff
@@ -0,0 +1,984 @@
...
+which typically consist of only a timestamp field and a log message field. Consequently,
+structured logs, such as JSON logs, cannot be efficiently serialized using this format.
+
+[CLP-S][clp-s-osdi] is an extension of the CLP's design principle but applied to semi-structured logs such as 
+JSON. It uses the key value structure of JSON entries so that a schema is a unique set of keys and 
+the values associated with those keys are the variables. Each schema can be represented by a tree. 
+When a unique key name and primitive type pair is encountered, we add a new leaf node to the tree. 
+The internal nodes of the tree are any surrounding objects. We merge these individual trees 
+together to form a merged parse tree which we’ll refer to as the schema tree. If two keys in 
+different schemas share a key name, but not a type, we add that new node to the tree. Since a set of key and type pairs make a schema, more than one log event can share the same schema and just have seperate values correpsonding to the keys. We collect the unique schemas forming a schema map. Like in CLP these
```
LinZhihao-723, line 26:
```suggestion
different schemas share a key name, but not a type, we add that new node to the tree. Since a set of key and type pairs make a schema, more than one log event can share the same schema and just have separate values corresponding to the keys. We collect the unique schemas forming a schema map. Like in CLP these
```

LinZhihao-723 (changes requested):
- Left some comments on the intro part. Will need Kirk to give more detailed feedback on the CLP-S intro section.
- Also reviewed other modified parts to correct some terms/concepts.
- It would be helpful if u could run a spelling check before asking for review comments.

kirkrodrigues:
@AVMatthews Could you do the following?

1. Re-add the png files from docs/src/dev-guide/images as follows:

    * Install git-lfs if you don't have it already.
    * Copy the images to a temporary location.
    * `git rm -f` all the images.
    * Commit the change.
    * Move the images back from the temporary location.
    * Add the images again.
    * Commit the change.

2. Review the new docs I added.

docs/src/dev-guide/kv-ir/background.md
```diff
@@ -0,0 +1,352 @@
...
+
+The kv-pairs in a log event can typically be divided into two types:
+
+1. *auto-generated* kv-pairs - those inserted automatically by the logging framework.
+2. *user-generated* kv-pairs - user-defined kv-pairs inserted explicitly by the user.
+
+For instance, consider the structured Go log printing statement (LPS) in Figure 6 (that uses the
+[Zap](https://github.com/uber-go/zap) logging library) and its output in Figure 7.
+
+:::{card}
```
AVMatthews, line 309: Generally I think this is a good example, but do auto generated keys always end up in the message like this? How do we distinguish them in the log message and so if you have two timestamps for example do they both show up in the JSON message.

AVMatthews (commented):
Content looks pretty good to me. Most of my comments are about the details of how it looks or reads. It's hard to say if I think we should add anything until I see the draft of the rest of the content.

PR #721, opened 2025-02-14
feat(core): Add ValidatingUtf8Parser to string_utils.

gibber9809:
# Description

This PR adds the `ValidatingUtf8Parser` class to string_utils. This class accepts raw, potentially corrupted, UTF-8 strings and returns valid UTF-8 strings or an error according to a configurable invalid UTF-8 handling policy. The return signature of the validate method is `outcome<std::string_view>`, and the `ValidatingUtf8Parser` class maintains an internal buffer so that we can support error handling policies that require copying and manipulating the original string. Note that as a result of this choice the lifetime of the returned  `string_view` is up to the next call to `validate` or the end of the lifetime of the `string_view `passed as input, whichever is lesser.

`ValidatingUtf8Parser` currently has two flavours of validate signatures `validate(raw, policy)` and templated `validate<policy>(raw)`. It should be trivial to add another family of policies that accept a lambda to handle invalid UTF-8 if needed.

# Checklist

* [ x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [ x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [ x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* Added appropriate unit tests for all `validate` signatures and error handling policies

## Summary by CodeRabbit

- **New Features**
  - Introduced a robust UTF-8 validation capability that allows configurable error handling—either substituting replacement characters or returning clear error notifications for invalid strings.
  - Integrated improvements to external dependency management to enhance overall string processing performance.

- **Tests**
  - Added comprehensive test cases to verify the accuracy of UTF-8 validation under various scenarios, ensuring reliable handling of both valid and invalid string inputs.

---
  - https://app.asana.com/0/0/1212888452014173

### DIFF
--- a/components/core/src/clp/string_utils/string_utils.cpp
+++ b/components/core/src/clp/string_utils/string_utils.cpp
@@ -302,12 +302,12 @@ bool wildcard_match_unsafe_case_sensitive(string_view tame, string_view wild) {
     }
 }
 
-auto ValidatingUtf8Parser::validate(std::string_view raw)
+auto ValidatingUtf8Parser::validate(std::string_view raw, InvalidUtf8Policy policy)
         -> OUTCOME_V2_NAMESPACE::std_result<std::string_view> {
     if (simdjson::validate_utf8(raw.data(), raw.size())) {
         return raw;
     }
-    switch (m_policy) {
+    switch (policy) {
         case InvalidUtf8Policy::SubstituteReplacementCharacter:
             m_buffer.clear();
             utf8::replace_invalid(raw.begin(), raw.end(), std::back_inserter(m_buffer));
@@ -317,5 +317,28 @@ auto ValidatingUtf8Parser::validate(std::string_view raw)
         default:
             return std::errc::invalid_argument;
     }
+}
+
+template <>
+auto ValidatingUtf8Parser::validate<ValidatingUtf8Parser::InvalidUtf8Policy::ReturnError>(
+        std::string_view raw
+) -> OUTCOME_V2_NAMESPACE::std_result<std::string_view> {
+    if (simdjson::validate_utf8(raw.data(), raw.size())) {
+        return raw;
+    }
+    return std::errc::illegal_byte_sequence;
+}
+
+template <>
+auto ValidatingUtf8Parser::validate<
+        ValidatingUtf8Parser::InvalidUtf8Policy::SubstituteReplacementCharacter>(
+        std::string_view raw
+) -> OUTCOME_V2_NAMESPACE::std_result<std::string_view> {
+    if (simdjson::validate_utf8(raw.data(), raw.size())) {
+        return raw;
+    }
+    m_buffer.clear();
+    utf8::replace_invalid(raw.begin(), raw.end(), std::back_inserter(m_buffer));
+    return m_buffer;
 }
 }  // namespace clp::string_utils
--- a/components/core/src/clp/string_utils/string_utils.cpp
+++ b/components/core/src/clp/string_utils/string_utils.cpp
@@ -1,8 +1,15 @@
 #include "string_utils/string_utils.hpp"
 
+#include <utfcpp/source/utf8.h>
+
 #include <algorithm>
 #include <charconv>
 #include <cstring>
+#include <iterator>
+#include <string_view>
+#include <system_error>
+
+#include <simdjson.h>
 
 using std::string;
 using std::string_view;
@@ -292,6 +299,23 @@ bool wildcard_match_unsafe_case_sensitive(string_view tame, string_view wild) {
                 }
             }
         }
+    }
+}
+
+auto ValidatingUtf8Parser::validate(std::string_view raw)
+        -> OUTCOME_V2_NAMESPACE::std_result<std::string_view> {
+    if (simdjson::validate_utf8(raw.data(), raw.size())) {
+        return raw;
+    }
+    switch (m_policy) {
+        case InvalidUtf8Policy::SubstituteReplacementCharacter:
+            m_buffer.clear();
+            utf8::replace_invalid(raw.begin(), raw.end(), std::back_inserter(m_buffer));
+            return m_buffer;
+        case InvalidUtf8Policy::ReturnError:
+            return std::errc::illegal_byte_sequence;
+        default:
+            return std::errc::invalid_argument;
     }
 }
 }  // namespace clp::string_utils
--- a/components/core/src/clp/string_utils/string_utils.hpp
+++ b/components/core/src/clp/string_utils/string_utils.hpp
@@ -2,6 +2,7 @@
 #define CLP_STRING_UTILS_STRING_UTILS_HPP
 
 #include <charconv>
+#include <cstdint>
 #include <string>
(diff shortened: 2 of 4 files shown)


components/core/src/clp/string_utils/string_utils.hpp
```diff
@@ -140,6 +144,53 @@ bool convert_string_to_int(std::string_view raw, integer_t& converted) {
         return result.ec == std::errc();
     }
 }
+
+/**
+ * This class implements a validating UTF-8 parser which takes as input valid or nearly-valid UTF-8
+ * and outputs valid UTF-8, handling invalid UTF-8 characters according to some policy.
+ */
+class ValidatingUtf8Parser {
```
LinZhihao-723, line 152: It might be worth creating a new header under `string_utils` for this utility class.

components/core/src/clp/string_utils/string_utils.cpp
```diff
@@ -294,4 +301,44 @@ bool wildcard_match_unsafe_case_sensitive(string_view tame, string_view wild) {
...
+}
+
+template <>
+auto ValidatingUtf8Parser::validate<ValidatingUtf8Parser::InvalidUtf8Policy::ReturnError>(
+        std::string_view raw
+) -> OUTCOME_V2_NAMESPACE::std_result<std::string_view> {
+    if (simdjson::validate_utf8(raw.data(), raw.size())) {
+        return raw;
+    }
+    return std::errc::illegal_byte_sequence;
```
LinZhihao-723, line 329: I think it would be more straightforward to define our own error code for UTF8 validation, like in `ffi::ir_stream::IrErrorCode`. You can refer to my L&L slides for more details.

components/core/src/clp/string_utils/string_utils.cpp
```diff
@@ -294,4 +301,44 @@ bool wildcard_match_unsafe_case_sensitive(string_view tame, string_view wild) {
...
+            return m_buffer;
+        case InvalidUtf8Policy::ReturnError:
+            return std::errc::illegal_byte_sequence;
+        default:
+            return std::errc::invalid_argument;
+    }
+}
+
+template <>
+auto ValidatingUtf8Parser::validate<ValidatingUtf8Parser::InvalidUtf8Policy::ReturnError>(
```
LinZhihao-723, line 323: I'd prefer to move the template implementation into the header file and use `constexpr` if-else to determine which policy to use.

components/core/tests/test-string_utils.cpp
```diff
@@ -560,3 +562,69 @@ TEST_CASE("convert_string_to_int", "[convert_string_to_int]") {
...
+                                               "\x80"
+                                               "\x80"
+                                               "\xC0"
+                                               "4\u1234"};
+    constexpr std::string_view cExpectedReplacementForInvalidString2{"abcd123\uFFFD\uFFFD4\u1234"};
+    ValidatingUtf8Parser parser;
+
+    SECTION("Error on invalid UTF-8 policy") {
+        constexpr auto cPolicy{ValidatingUtf8Parser::InvalidUtf8Policy::ReturnError};
+        auto outcome = parser.validate(cValidString, cPolicy);
```
LinZhihao-723, line 583: Our convention is to name a `std_result` retval `xxx_result`, so how about `outcome` -> `result` or `validation_result`?

components/core/src/clp/string_utils/string_utils.hpp
LinZhihao-723, line 152: In terms of naming, how about `Utf8Validator`?

components/core/src/clp/string_utils/string_utils.hpp
```diff
@@ -140,6 +144,53 @@ bool convert_string_to_int(std::string_view raw, integer_t& converted) {
...
+     * or until the input string_view passed to this function becomes invalid, whichever is sooner.
+     *
+     * @param raw
+     * @param policy The error handling policy to apply when encountering invalid UTF-8 characters.
+     * @return A result containing a valid UTF-8 string or an error code indicating the failure:
+     * - std::errc::illegal_byte_sequence if the input contains invalid UTF-8 and the policy is
+     *   ReturnError.
+     * - std::errc::invalid_argument if the configured policy is unknown.
+     */
+    auto validate(std::string_view raw, InvalidUtf8Policy policy)
```
LinZhihao-723, line 172: What do u think if we move the implementation of this function into the header so that the compiler can optimize and inline this call without a run-time branching if the policy is a compile-time constant? In that way we might not need a templated version anymore.

LinZhihao-723, line 172: If we use my suggested class name, how about renaming this function to `parse`?

LinZhihao-723, line 172:
```suggestion
    [[nodiscard]] auto validate(std::string_view raw, InvalidUtf8Policy policy)
```

components/core/src/clp/string_utils/string_utils.cpp
```diff
@@ -294,4 +301,44 @@ bool wildcard_match_unsafe_case_sensitive(string_view tame, string_view wild) {
...
+    switch (policy) {
+        case InvalidUtf8Policy::SubstituteReplacementCharacter:
+            m_buffer.clear();
+            utf8::replace_invalid(raw.begin(), raw.end(), std::back_inserter(m_buffer));
+            return m_buffer;
+        case InvalidUtf8Policy::ReturnError:
+            return std::errc::illegal_byte_sequence;
+        default:
+            return std::errc::invalid_argument;
+    }
```
LinZhihao-723, line 319: NTS: If we decide to keep both the non-template and the template version, we should call the template one in this call.

components/core/src/clp/string_utils/string_utils.hpp
```diff
@@ -140,6 +144,53 @@ bool convert_string_to_int(std::string_view raw, integer_t& converted) {
...
+/**
+ * This class implements a validating UTF-8 parser which takes as input valid or nearly-valid UTF-8
+ * and outputs valid UTF-8, handling invalid UTF-8 characters according to some policy.
+ */
+class ValidatingUtf8Parser {
+public:
+    enum class InvalidUtf8Policy : uint8_t {
+        SubstituteReplacementCharacter,
+        ReturnError
+    };
```
LinZhihao-723, line 157: How about:
```suggestion
    enum class ErrorHandlingPolicy : uint8_t {
        Strict,
        Replace
    };
```
- Since this class is inside this utf8 validation class, we don't have to duplicate "Utf8" in its name
- `SubstituteReplacementCharacter` -> `Replace`, `ReturnError` -> `Strict` to mirror Python's error handling policy names ([here](https://docs.python.org/3/library/codecs.html#error-handlers)), which might be more concise.

Open to other naming plans if u have any

LinZhihao-723 (changes requested):
Reviewed the overall structure and API naming. Haven't looked into the docstrings in detail yet. Will do that after we finalize the API design.

PR #733, opened 2025-02-26
feat(clp-s): Change behaviour for kv-ir ingestion to warn and continue on error

gibber9809:
# Description

Properly handling kv-ir requires us to support ingesting truncated kv-ir streams, which can come about when workers producing those streams crash.

This PR modifies ingestion behaviour to handle most kv-ir deserialization errors by logging a warning and continuing on to the next stream. The only exceptions are that
1) If we can not open a stream at all then we treat that as an error and fail ingestion
2) If we encounter a deserialization error but we are ingesting data from a network source whose reader has encountered a network error then the network error takes precedence and ingestion fails

# Checklist

  breaking change.

# Validation performed

* Validated that ingestion succeeds with a warning when removing the zero byte at the end of an IR stream
* Validated that ingestion succeeds with a warning when ingesting an IR stream that has been truncated in a random location (and that records that were not part of the truncated data appear in the archive)

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced error handling during log processing, ensuring that less critical issues generate warnings instead of errors.
  - Improved response to incomplete data, allowing for more graceful adjustment during unexpected stream interruptions.

### DIFF
--- a/components/core/src/clp_s/JsonParser.cpp
+++ b/components/core/src/clp_s/JsonParser.cpp
@@ -930,15 +930,21 @@ auto JsonParser::parse_from_ir() -> bool {
 
             if (kv_log_event_result.has_error()) {
                 auto err = kv_log_event_result.error();
-                SPDLOG_ERROR(
-                        "Encountered error while deserializing kv-ir log event: ({}) - {}",
+                SPDLOG_WARN(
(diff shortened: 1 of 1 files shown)


wraymo (approved):
Looks good to me

PR #743, opened 2025-03-04
feat(clp-package): Add support for more AWS authentication methods.

Eden-D-Zhang:
# Description

This PR introduces significant improvements to AWS authentication handling and other improvements to the CLP package. These changes include:
1. **Flexible AWS Authentication**
    - Introduced the AwsAuthentication class within the existing S3Config to support multiple authentication methods:
      - EC2 instance metadata (`ec2`)
      - Environment variables (`env_vars`)
      - AWS configuration profile (`profile`)
      - Explicit credentials (`credentials`)
    - Added the `aws_config_directory` field to `CLPConfig` to specify the location of the `.aws` where profiles are configured (defaults to `~/.aws`).
    - Added functionality to provide credential sources to components depending on authentication type:
      - `aws_config_dir` Docker mount for authentication through configured profiles.
      - Modified the `S3Manager` class in the `log_viewer_webui` component to support authentication through profiles.
      - Environmental variable handling for tasks that cannot authenticate dynamically.
    - Refactored existing authentication control flow to eliminate redundant code.
2. **Other Improvements**
    - Refactored the `S3InputConfig` class to inherit `S3Config` class to allow reusing similar code.
    - Refactored handling of docker mounts and environmental variables in `start_clp.py`:
        - Created `append_docker_mounts` and `append_docker_env_vars` functions to simplify construction of Docker container start commands.
        - Removed hardcoded environmental variables in Docker start commands to make use of aforementioned `append_docker_env_vars` function and improve maintainabiility.
        - These changes help with method-specific and component-specific AWS authentication.

# Checklist

  breaking change.

# Validation performed

Manually verified compression, search, and log viewer both locally and on EC2 using all combinations of authentication methods.

## Summary by CodeRabbit

- **New Features**
  - Added support for multiple AWS authentication methods including EC2, environment variables, profiles, and credentials with session tokens.
  - Enabled mounting of host AWS configuration directories into Docker containers for seamless AWS credential access.
  - Log Viewer Web UI now supports AWS profile specification for S3 access.

- **Improvements**
  - Centralized AWS authentication management across components, simplifying credential handling.
  - Streamlined Docker container startup by consolidating environment variable and mount option configuration.
  - Enhanced configuration files and UI to accommodate new AWS authentication options.

- **Documentation**
  - Updated configuration templates and settings to incorporate new AWS authentication structures and directory options.
  - Revised user guides to explain expanded AWS authentication methods and updated IAM policy requirements.


components/package-template/src/etc/clp-config.yml
```diff
@@ -99,3 +99,6 @@
 #
 ## Location where logs are stored. It will be created if it doesn't exist.
 #logs_directory: "var/log"
+#
+## Location where the `.aws` directory should be read from
+#aws_config_directory: "$HOME/.aws"
```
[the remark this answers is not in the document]
Eden-D-Zhang: we agreed that it makes sense to keep it at the top level.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -332,7 +333,8 @@ class S3Config(BaseModel):
     bucket: str
     key_prefix: str
 
-    credentials: S3Credentials
+    profile: Optional[str] = None
```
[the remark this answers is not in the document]
Eden-D-Zhang: I believe that if no profile is specified to boto3, it automatically tries to use the default profile, but we could include it specifically for clarity

components/job-orchestration/job_orchestration/executor/query/fs_search_task.py
```diff
@@ -71,12 +77,29 @@ def _make_core_clp_s_command_and_env_vars(
...
         # fmt: on
-        aws_access_key_id, aws_secret_access_key = s3_config.get_credentials()
-        env_vars = {
-            **os.environ,
-            "AWS_ACCESS_KEY_ID": aws_access_key_id,
-            "AWS_SECRET_ACCESS_KEY": aws_secret_access_key,
-        }
+        env_vars: Dict[str, str] = None
+
+        if s3_config.credentials is not None:
```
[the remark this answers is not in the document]
Eden-D-Zhang: yeah, I was thinking it would be a good idea

components/clp-package-utils/clp_package_utils/scripts/compress.py
```diff
@@ -124,16 +90,15 @@ def _generate_compress_cmd(
...
-            aws_access_key_id, aws_secret_access_key = _parse_aws_credentials_file(
-                pathlib.Path(parsed_args.aws_credentials_file), default_credentials_user
-            )
+        aws_profile = parsed_args.aws_profile
         if bool(aws_access_key_id) and bool(aws_secret_access_key):
             compress_cmd.append("--aws-access-key-id")
             compress_cmd.append(aws_access_key_id)
             compress_cmd.append("--aws-secret-access-key")
             compress_cmd.append(aws_secret_access_key)
+        if bool(aws_profile):
```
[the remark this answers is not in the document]
Eden-D-Zhang: handled at line 150

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -657,7 +672,11 @@ def start_query_worker(
     celery_route = f"{QueueName.QUERY}"
 
     query_worker_mounts = [mounts.stream_output_dir]
-    if clp_config.archive_output.storage.type == StorageType.FS:
+    if StorageType.S3 == clp_config.stream_output.storage.type:
+        query_worker_mounts.append(mounts.aws_config_dir)
+    elif StorageType.S3 == clp_config.archive_output.storage.type:
+        query_worker_mounts.append(mounts.aws_config_dir)
```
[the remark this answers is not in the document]
Eden-D-Zhang: I have it this way because I only want to append `aws_config_dir` once if both `stream_output` and `archive_output` are of type S3, but still append both `aws_config_dir` and `archives_output_dir` if `stream_output` is S3 and `archive_output` is FS. Feel free to suggest a better way to do it

components/job-orchestration/job_orchestration/scheduler/job_config.py
```diff
@@ -35,7 +35,8 @@ class S3InputConfig(BaseModel):
     bucket: str
     key_prefix: str
 
-    credentials: S3Credentials
+    profile: typing.Optional[str] = None
```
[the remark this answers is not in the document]
Eden-D-Zhang: would it make sense to handle credentials/profile as a `Union[str, S3Credentials]`?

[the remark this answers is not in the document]
haiqi96: I see two disadvantage:
1. Once we use a union, the credentials and profile needs to have the same name in the config.
2. With a union, the error message can be confusing.

I think for now, it's ok to keep them separate as optional, unless you have a different opinion.

components/clp-py-utils/clp_py_utils/clp_config.py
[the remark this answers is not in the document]
haiqi96: a benefit of having default as profile is that then we can have it in our example config file, which will also serve an example for users.

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,31 @@
...
+def get_session_credentials(aws_profile: str = "default") -> Optional[S3Credentials]:
+    aws_session = boto3.Session(profile_name=aws_profile)
+    credentials = aws_session.get_credentials()
+    if credentials is None:
+        return None
+
+    return S3Credentials(
+        access_key_id=credentials.access_key,
+        secret_access_key=credentials.secret_key,
+        session_token=credentials.token,
```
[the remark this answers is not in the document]
haiqi96: >If the use chooses to use a profile containing short-term credentials, boto3 would treat it like any other credentials and try to use it, and fail if it can't authenticate.

In that case, will the returned credentials.token be the same as the one specified in the .aws, or will boto3 generate a new temporary token?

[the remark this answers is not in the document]
Eden-D-Zhang: I believe it would be the same as specified

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
haiqi96, line 678: can we combine these two lines into a single if statement?

components/clp-py-utils/clp_py_utils/s3_utils.py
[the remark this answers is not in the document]
haiqi96: Would you please test the behavior?

Ideally, we should prevent user from using temporary credentials in the profile to start clp package, since eventually it will expire.. so maybe we should return an error in this case.

However, i believe if user just configures their profile through sso, we will always get a temporary credentials (which will be updated whenever we query the credentials from the session)?

components/clp-py-utils/clp_py_utils/clp_config.py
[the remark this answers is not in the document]
Eden-D-Zhang: if profile has a default value, how can we check for mutual exclusion?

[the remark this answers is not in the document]
haiqi96: Ture, I remember having a similar conversation with Kirk a while ago about cmdline option. 
Basically an alternative is to let clp allows the following input case:
- aws profile is default
- aws_credentials is specified.

But we finally didn't go with this method because it could also cause confusions.. I guess we will have to ask user to explicilty specify it.

@kirkrodrigues let me know if you have any different thoughts.

Eden-D-Zhang:
Ignore latest commit doesn't work atm

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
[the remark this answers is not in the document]
haiqi96: I thought I commented on this but seems I didn't.

What if we just combine line 676~679 into a single if statement? Looks like the body of the if clause are the same?

components/clp-package-utils/clp_package_utils/scripts/compress.py
```diff
@@ -263,7 +225,13 @@ def main(argv):
         container_clp_config, clp_config, container_name
     )
 
-    necessary_mounts = [mounts.clp_home, mounts.input_logs_dir, mounts.data_dir, mounts.logs_dir]
+    necessary_mounts = [
+        mounts.clp_home,
+        mounts.input_logs_dir,
+        mounts.data_dir,
+        mounts.logs_dir,
+        mounts.aws_config_dir,
```
haiqi96, line 233: we only need to mount this if an --aws-profile is provided?

components/clp-package-utils/clp_package_utils/scripts/compress.py
```diff
@@ -231,7 +193,7 @@ def main(argv):
         "--aws-secret-access-key", type=str, default=None, help="AWS secret access key."
     )
     s3_compressor_parser.add_argument(
-        "--aws-credentials-file", type=str, default=None, help="Path to AWS credentials file."
+        "--aws-profile", type=str, default=None, help="AWS configuration profile in ~/.aws"
```
haiqi96, line 196: maybe something like:
```suggestion
        "--aws-profile", type=str, default=None, help="AWS configuration profile to use for authentication"
```

feel free to change it.

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -220,6 +232,9 @@ def main(argv):
     s3_compressor_parser.add_argument(
         "--aws-secret-access-key", type=str, default=None, help="AWS secret access key."
     )
+    s3_compressor_parser.add_argument(
+        "--aws-profile", type=str, default=None, help="AWS config profile."
```
haiqi96, line 236: nit: just to be consistent.
```suggestion
        "--aws-profile", type=str, default=None, help="AWS configuration profile to use for authentication"
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -313,6 +313,7 @@ class Queue(BaseModel):
 class S3Credentials(BaseModel):
     access_key_id: str
     secret_access_key: str
+    session_token: Optional[str]
```
haiqi96, line 316: unused?

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -74,11 +130,21 @@ def s3_get_object_metadata(s3_input_config: S3InputConfig) -> List[FileMetadata]
     :raises: Propagates `boto3.paginator`'s exceptions.
     """
 
-    s3_client = boto3.client(
+    aws_session = None
+    if s3_input_config.profile is not None:
+        aws_session = boto3.Session(profile_name=s3_input_config.get_profile())
```
haiqi96, line 135: missing region_name=s3_input_config.region_code,?

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -124,13 +190,22 @@ def s3_put(
...
-    my_s3_client = boto3.client(
+    aws_session = None
+    if aws_profile is not None:
+        aws_session = boto3.Session(profile_name=aws_profile)
+    else:
+        aws_session = boto3.Session(
+            aws_access_key_id=aws_access_key_id,
+            aws_secret_access_key=aws_secret_access_key,
+            region_name=s3_config.region_code,
+        )
```
haiqi96, line 205: this is similar to line 133 ~ 144 but written in a slightly different way. can we factor this out as a helper function that eventually returns a s3_client?

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,74 @@
...
+        return None
+
+    return S3Credentials(
+        access_key_id=credentials.access_key,
+        secret_access_key=credentials.secret_key,
+        session_token=credentials.token,
+    )
+
+
+def make_s3_env_vars(config: Union[S3Config, S3InputConfig]) -> Dict[str, str]:
```
haiqi96, line 35: The method's docstrings need to be updated.

I also don't completely agree with the name of the method and where to put it.
For name, technically this is not s3 env vars, but rather env vars for aws credentials. also feel that the function can simply return a dictionary of only new env-vars, instead of appending to the `os.environ` directly.

Need to think about it for a bit.

components/job-orchestration/job_orchestration/executor/query/fs_search_task.py
```diff
@@ -5,9 +5,15 @@
 
 from celery.app.task import Task
 from celery.utils.log import get_task_logger
-from clp_py_utils.clp_config import Database, StorageEngine, StorageType, WorkerConfig
+from clp_py_utils.clp_config import (
+    Database,
+    S3Credentials,
```
haiqi96, line 10: unused

components/job-orchestration/job_orchestration/executor/query/extract_stream_task.py
```diff
@@ -6,9 +6,20 @@
 
 from celery.app.task import Task
 from celery.utils.log import get_task_logger
-from clp_py_utils.clp_config import Database, S3Config, StorageEngine, StorageType, WorkerConfig
+from clp_py_utils.clp_config import (
+    Database,
+    S3Config,
+    S3Credentials,
```
haiqi96, line 12: ditto

components/job-orchestration/job_orchestration/executor/query/extract_stream_task.py
```diff
@@ -6,9 +6,20 @@
...
+    StorageEngine,
+    StorageType,
+    WorkerConfig,
+)
 from clp_py_utils.clp_logging import set_logging_level
-from clp_py_utils.s3_utils import generate_s3_virtual_hosted_style_url, s3_put
+from clp_py_utils.s3_utils import (
+    generate_s3_virtual_hosted_style_url,
+    make_s3_env_vars,
+    s3_put,
```
haiqi96, line 21: some are unused. ditto for other places

components/job-orchestration/job_orchestration/scheduler/job_config.py
```diff
@@ -35,7 +35,16 @@ class S3InputConfig(BaseModel):
...
+    credentials: typing.Optional[S3Credentials] = None
+
+    def get_profile(self) -> typing.Optional[str]:
+        return self.profile
+
+    def get_credentials(self) -> typing.Tuple[typing.Optional[str], typing.Optional[str]]:
+        if self.credentials is None:
+            return None, None
+        return self.credentials.access_key_id, self.credentials.secret_access_key
```
haiqi96, line 49: please add a check to ensure profile and credentials are mutually exclusive

components/log-viewer-webui/server/src/S3Manager.js
```diff
@@ -60,11 +62,11 @@ class S3Manager {
  * "s3Manager" property when all plugin options are valid.
  */
 export default fastifyPlugin(async (app, options) => {
-    const {region} = options;
-    if (null === region) {
+    const {region, profile} = options;
+    if (!region || !profile) {
```
haiqi96, line 66: I feel this will not work on ec2? I believe that on ec2, we don't provide a profile, wonder if you have a plan to update this.

haiqi96 (commented):
Haven't looked at start_clp.py carefully,. but let me leave the first batch of comments

components/clp-py-utils/clp_py_utils/clp_config.py
Eden-D-Zhang: it's used in the function that gets temporary credentials

haiqi96: True, you are right. sorry for my oversight

components/log-viewer-webui/server/src/S3Manager.js
haiqi96: Right, it's not decided yet. Just to make sure, it will work on ec2 if we create S3Client without profile param?

My guess is maybe uf we assign null to profile, it will work, but not 100% sure.

Eden-D-Zhang: yeah we'd pass no params for EC2. I'm not sure about passing a null value for profile but when I tried running the package on EC2 yesterday it seemed to work without changes anyway? Will have to look into it more.

components/clp-package-utils/clp_package_utils/scripts/compress.py
Eden-D-Zhang: Right, I'll change that

components/clp-py-utils/clp_py_utils/s3_utils.py
haiqi96: Can we:
1. Rename this to `get_credential_env_vars`.
2. Remove **os.environ from the return, and let callers combine the return value and **os.environ.

I feel this will make the method clearer

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,101 @@
...
+            "AWS_ACCESS_KEY_ID": aws_access_key_id,
+            "AWS_SECRET_ACCESS_KEY": aws_secret_access_key,
+        }
+        return env_vars
+
+    else:
+        raise ValueError("No AWS credential source found")
+
+
+def _generate_s3_client(s3_config: Union[S3Config, S3InputConfig]) -> boto3.client:
```
haiqi96, line 76: nit: how about `_create_s3_client`

haiqi96, line 76: I personally feel the union here looks weird..
Looking at the S3Config and S3InputConfig, they are very similar, except that S3InputConfig contains an extra timestamp_key. the rest are just duplicated

I see two possible ways to reduce the duplications:
1. Let S3InputConfig inherit the S3Config, so we only need to define the extra member variables
2. Let S3InputConfig internally stores an S3Config. with this, you might also need to change how we interface with S3InputConfig  in the othe scripts.
I would say this is low priority compare to the other part of the review, but when you get a chance, can you let me know which version you prefer, and try implememting it?

components/clp-py-utils/clp_py_utils/clp_config.py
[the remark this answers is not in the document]
Eden-D-Zhang: Yes, that's in pr #788, preferably you'd take a look at that one first since it will change a few things here.

components/clp-py-utils/clp_py_utils/s3_utils.py
Eden-D-Zhang: I think option 1 makes more sense, especially since #788 will make a base S3Config class without the staging directory.

components/clp-py-utils/clp_py_utils/s3_utils.py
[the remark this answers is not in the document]
Eden-D-Zhang: yeah I don't think we can check if a profile is configured with a temporary token, since we just pass the profile name to boto3.
For SSO, boto3 authenticates through ./aws/sso metadata and generates temporary credentials when queried.

components/clp-py-utils/clp_py_utils/s3_utils.py
haiqi96: I see, what's the current plan? Do we first PR in 788 before this PR?

components/clp-py-utils/clp_py_utils/s3_utils.py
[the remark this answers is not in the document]
haiqi96: I see, then Iet's not worry about it. The current implementation should be good enough.

components/job-orchestration/job_orchestration/scheduler/job_config.py
```diff
@@ -35,7 +35,27 @@ class S3InputConfig(BaseModel):
...
+
+    @root_validator(pre=True)
+    def validate_profile_and_credentials(cls, values):
+        profile = values.get("profile")
+        credentials = values.get("credentials")
+
+        if profile and credentials:
+            raise ValueError("profile and credentials cannot be set simultaneously")
+        elif not profile and not credentials:
+            raise ValueError("one of either profile or credentials must be set")
```
haiqi96, line 49: It looks like we are enforcing a profile even for EC2 support. I remember that you tested it once, does profile=default work, or do we need to explicit to profile=None?

components/clp-py-utils/clp_py_utils/s3_utils.py
Eden-D-Zhang: Yes, 788 should be merged first, then this PR will add profile/ec2/env_vars support to the changes in 788 as well as everything else.

haiqi96: Gotcha, can you remind me what's the proposal to support ec2? (since I assume the current config.py wouldn' support it, but please correct me if wrong).

I will need to think about how to port it to cloud.

components/log-viewer-webui/server/src/plugins/S3Manager.ts
```diff
@@ -63,14 +71,14 @@ class S3Manager {
  * Initializes a Fastify plugin, which decorates the application with an S3 manager at the
  * "s3Manager" property only when all plugin options are valid.
  */
-export default fastifyPlugin(async (app, options: {region: Nullable<string>}) => {
-    const {region} = options;
+export default fastifyPlugin(async (app, options: {region: Nullable<string>, profile: string}) => {
+    const {region, profile} = options;
     if (null === region) {
```
junhaoliao, line 76: @coderabbitai In a separate PR, we should add an `isEnabled` option and check that instead. create an issue to track this.

components/job-orchestration/job_orchestration/scheduler/job_config.py
Eden-D-Zhang: EC2 changes aren't in yet, I'll get them pushed soon.

Eden-D-Zhang:
I should've clarified, the changes here right now are incomplete since #788 changes a few things relating to this PR and I've yet to finish the EC2/env var implementation. I'll make sure the current sugestions are addressed.

components/clp-py-utils/clp_py_utils/s3_utils.py
haiqi96: I see, this will for the cloud purpose. thanks for clarificaton.

components/clp-package-utils/clp_package_utils/scripts/compress.py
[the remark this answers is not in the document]
Eden-D-Zhang: 788 will make most of the changes to the compress.py scripts here obsolete

components/log-viewer-webui/server/src/plugins/S3Manager.ts
```diff
@@ -20,12 +20,20 @@ class S3Manager {
     readonly #s3Client;
 
     /**
-     * @param region
+     * @param {string} region
+     * @param {string} [profile]
```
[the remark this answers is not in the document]
Eden-D-Zhang: To specify that it's optional

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -32,6 +32,7 @@
     StorageType,
     WEBUI_COMPONENT_NAME,
 )
+from clp_py_utils.s3_utils import get_container_authentication
```
haiqi96, line 35: maybe keep an empty line?

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -74,6 +75,20 @@ def container_exists(container_name):
...
+        if mount:
+            cmd.append("--mount")
+            cmd.append(str(mount))
+
+
+def append_docker_env_vars(cmd: List[str], env_vars: List[str]):
+    for env_var in env_vars:
+        if "" != env_var:
+            cmd.append("-e")
+            cmd.append(env_var)
```
haiqi96, line 89: wonder, can we make those two functions into a single one, with an optional(env_vars)?

components/job-orchestration/job_orchestration/scheduler/job_config.py
```diff
@@ -3,7 +3,7 @@
 import typing
 from enum import auto
 
-from clp_py_utils.clp_config import S3Credentials
+from clp_py_utils.clp_config import AwsAuthentication, S3Config
```
haiqi96, line 6: nit:
AwsAuthentication seems to be unused

components/package-template/src/etc/clp-config.yml
```diff
@@ -105,3 +105,6 @@
...
```
haiqi96, line 110: one small concern, if user types "$HOME/.aws" in the config.yml, will it be intepreted as plain text "$HOME"?

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
Eden-D-Zhang: It looks like other files (like `general.py`) tend not to have empty lines here

components/package-template/src/etc/clp-config.yml
Eden-D-Zhang: yeah I wasn't sure what the right way to handle this since the home directory depends on the user If they leave it commented then it will work properly, but I think it won't work if they uncomment and leave it as is.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -314,52 +314,73 @@ class Queue(BaseModel):
...
     @validator("secret_access_key")
     def validate_secret_access_key(cls, field):
-        if field == "":
+        if "" == field:
             raise ValueError("secret_access_key cannot be empty")
         return field
 
 
+class AwsAuthentication(BaseModel):
+    type: Literal["ec2", "env_vars", "profile", "credentials"]
```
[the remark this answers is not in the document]
Eden-D-Zhang: YAML is having issues representing these when they're set as `LowercaseStrEnum` in the same way as StorageType. Should they just be string literals instead?

components/package-template/src/etc/clp-config.yml
haiqi96: gotcha, I think we just need to add a note to remind user that if they want to uncomment it, they must use the real path.

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
haiqi96: I see, thanks for catching this. Let's remove the empty line then.

components/package-template/src/etc/clp-config.yml
Eden-D-Zhang: sounds good

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -314,52 +314,73 @@ class Queue(BaseModel):
...
+        if "profile" == auth_type:
+            if not profile:
+                raise ValueError("profile must be set when type is 'profile'")
+            if credentials:
+                raise ValueError("credentials must not be set when type is 'profile'")
+        elif "credentials" == auth_type:
+            if not credentials:
+                raise ValueError("credentials must be set when type is 'credentials'")
+            if profile:
+                raise ValueError("profile must not be set when type is 'credentials'")
```
haiqi96, line 352: nit: how about first check that credentials and profile are not both specfied outside the if-clause. then within the if-clause you only need to check what the auth-type cares.

components/package-template/src/etc/clp-config.yml
haiqi96: or maybe try using ~ instead of $HOME (as Kirk suggested)

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,177 @@
...
+    :return: An S3Credentials object with access key pair and session token if applicable.
+    """
+    aws_session = None
+    if aws_profile is not None:
+        aws_session = boto3.Session(profile_name=aws_profile)
+    else:
+        aws_session = boto3.Session()
+    credentials = aws_session.get_credentials()
+    if credentials is None:
+        return None
```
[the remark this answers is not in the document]
haiqi96: nvm, we can not, my bad.

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,177 @@
...
+        if aws_credentials is None:
+            raise ValueError(f"Failed to authenticate with profile {auth.profile}")
+
+        env_vars = {
+            "AWS_ACCESS_KEY_ID": aws_credentials.access_key_id,
+            "AWS_SECRET_ACCESS_KEY": aws_credentials.secret_access_key,
+        }
+        aws_session_token = aws_credentials.session_token
+        if aws_session_token is not None:
+            env_vars["AWS_SESSION_TOKEN"] = aws_session_token
```
haiqi96, line 59: this part of code is duplicated in 3 clauses. Can we reorganize the code by doing the following:
1. Init credentials: S3Credentials;
2. Handle "env_vars" == auth.type: with early return.
3. For credentials, ec2 and profile, update the "credentials" in their own if-statement.
4. use the common logic to convert it into env_vars format.

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,177 @@
...
+
+    :param clp_config: CLPConfig containing storage configurations
+    :param type: Type of calling container (compression, log_viewer, or query)
+    :return: Tuple of (whether aws config mount is needed, credential env_vars to set)
+    :raises: ValueError if environment variables are not set correctly or if type is invalid
+    """
+    if type not in ["compression", "log_viewer", "query"]:
+        raise ValueError(f"Unsupported authentication type: {type}")
+
+    storages = {
```
haiqi96, line 107: how about "storages_by_component_type"

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,177 @@
...
+    """
+    Generates Docker container authentication options for AWS S3 access based on the given type.
+    Handles authentication methods that require extra configuration (profile, env_vars).
+
+    :param clp_config: CLPConfig containing storage configurations
+    :param type: Type of calling container (compression, log_viewer, or query)
+    :return: Tuple of (whether aws config mount is needed, credential env_vars to set)
+    :raises: ValueError if environment variables are not set correctly or if type is invalid
+    """
+    if type not in ["compression", "log_viewer", "query"]:
```
haiqi96, line 104: type -> component, or component_type, component_name

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
haiqi96: that said, can't think of a good name. @kirkrodrigues do you have any recommendation?

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -263,15 +276,13 @@ def create_results_cache_indices(
         "--rm",
         "--name", container_name,
         "--log-driver", "local",
-        "-e", f"PYTHONPATH={clp_site_packages_dir}",
         "-u", f"{os.getuid()}:{os.getgid()}",
     ]
     # fmt: on
+    necessary_env_vars = [f"PYTHONPATH={clp_site_packages_dir}"]
```
haiqi96, line 282: let's not name it as "necessary" unless there's an optional counter part for the same component.
I understand that you copied the "necessary" from necessary_mounts. Tbh, I feel the name for mount is also confusing but let's not touch it in this PR.

components/clp-py-utils/clp_py_utils/s3_utils.py
haiqi96, line 104: instead of defining your own types, I think you can reuse things like "COMPRESSION_WORKER_COMPONENT_NAME"

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,177 @@
...
+        }
+        if aws_credentials.session_token:
+            env_vars["AWS_SESSION_TOKEN"] = aws_credentials.session_token
+        return env_vars
+
+    else:
+        raise ValueError(f"Unsupported authentication type: {auth.type}")
+
+
+def get_container_authentication(
```
haiqi96, line 92: Similar to your doc strings, maybe rename it to something like "generate_container_auth_options/configs"?

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -538,13 +555,22 @@ def start_query_scheduler(
     mounts: CLPDockerMounts,
 ):
     module_name = "job_orchestration.scheduler.query.query_scheduler"
+    query_scheduler_mount = None
```
haiqi96, line 558: once you use vars like QUERY_SCHEDULER_COMPONENT_NAME as 2nd argument, maybe you can then refactor this logic into `generic_start_scheduler`.

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -580,33 +608,37 @@ def generic_start_scheduler(
         "-w", str(CONTAINER_CLP_HOME),
         "--name", container_name,
         "--log-driver", "local",
-        "-e", f"PYTHONPATH={clp_site_packages_dir}",
-        "-e", (
+        "-u", f"{os.getuid()}:{os.getgid()}",
+    ]
+    # fmt: on
+
+    necessary_env_vars = [
```
haiqi96, line 615: Small nit: what if we turn this into a dictionary with EnvVarName:Value structure?

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -965,25 +1010,36 @@ def start_log_viewer_webui(
...
+        auth = stream_storage.s3_config.aws_authentication
+        if "credentials" == auth.type:
+            credentials = auth.credentials
+            necessary_env_vars.append(f"AWS_ACCESS_KEY_ID={credentials.access_key_id}")
+            necessary_env_vars.append(f"AWS_SECRET_ACCESS_KEY={credentials.secret_access_key}")
+        aws_mount, aws_env_vars = get_container_authentication(clp_config, "log_viewer")
+        if aws_mount:
+            necessary_mounts.append(mounts.aws_config_dir)
+        if aws_env_vars:
+            necessary_env_vars.extend(aws_env_vars)
```
haiqi96, line 1040: this can be under an "else"?

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
Eden-D-Zhang: I have append_docker_options for now

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -698,6 +754,7 @@ def dump_to_primitive_dict(self):
         d["credentials_file_path"] = str(self.credentials_file_path)
         d["data_directory"] = str(self.data_directory)
         d["logs_directory"] = str(self.logs_directory)
+        d["aws_config_directory"] = str(self.aws_config_directory)
```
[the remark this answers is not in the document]
haiqi96: @Eden-D-Zhang wonder could this be an issue?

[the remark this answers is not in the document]
Eden-D-Zhang: yeah I wasn't sure, but looking into it I think we should take the suggestion

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -74,6 +76,29 @@ def container_exists(container_name):
...
+    """
+    Appends Docker mount and environment variable options to a command list.
+
+    :param cmd: The command list to append options to.
+    :param mounts: Optional list of DockerMount objects to add as --mount options.
+    :param env_vars: Optional list of environment variables to add as -e options.
+    """
+    if mounts:
+        for mount in mounts:
+            if mount:
```
haiqi96, line 91: nit: I don't see why you need an `if mount` line here. Do you mean that input type should be `Optional[List[Optional[DockerMount]]]` ?

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -314,52 +321,76 @@ class Queue(BaseModel):
...
+        credentials = values.get("credentials")
+
+        if profile and credentials:
+            raise ValueError("profile and credentials cannot be set simultaneously.")
+        if AwsAuthType.profile == auth_type:
+            if not profile:
+                raise ValueError(f"profile must be set when type is '{auth_type}.'")
+        elif AwsAuthType.credentials == auth_type:
+            if not credentials:
+                raise ValueError(f"credentials must be set when type is '{auth_type}.'")
```
haiqi96, line 362:
```suggestion
        elif AwsAuthType.credentials == auth_type and not credentials:
            raise ValueError(f"credentials must be set when type is '{auth_type}.'")
```

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,208 @@
...
+    """
+    if component_type not in [
+        COMPRESSION_SCHEDULER_COMPONENT_NAME,
+        COMPRESSION_WORKER_COMPONENT_NAME,
+        LOG_VIEWER_WEBUI_COMPONENT_NAME,
+        QUERY_SCHEDULER_COMPONENT_NAME,
+        QUERY_WORKER_COMPONENT_NAME,
+    ]:
+        raise ValueError(f"Component type {component_type} is not valid.")
+
```
haiqi96, line 124:
```suggestion
```

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,208 @@
...
+    elif (
+        QUERY_SCHEDULER_COMPONENT_NAME == component_type
+        or QUERY_WORKER_COMPONENT_NAME == component_type
+    ):
+        storages_by_component_type = [
+            clp_config.logs_input,
+            clp_config.archive_output.storage,
+            clp_config.stream_output.storage,
+        ]
+
```
haiqi96, line 142:
```suggestion
        ]
    else:
        raise ValueError(f"Component type {component_type} is not valid.")
```

components/clp-py-utils/clp_py_utils/s3_utils.py
haiqi96: let's move this raise to under if-elif-else clause

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,208 @@
...
+        QUERY_SCHEDULER_COMPONENT_NAME,
+        QUERY_WORKER_COMPONENT_NAME,
+    ]:
+        raise ValueError(f"Component type {component_type} is not valid.")
+
+    storages_by_component_type = []
+    if (
+        COMPRESSION_SCHEDULER_COMPONENT_NAME == component_type
+        or COMPRESSION_WORKER_COMPONENT_NAME == component_type
+    ):
```
haiqi96, line 129: nit: Let's use what's currently used in start-clp.py
```suggestion
    if component_type in (
        COMPRESSION_SCHEDULER_COMPONENT_NAME, 
        COMPRESSION_WORKER_COMPONENT_NAME
    ):
```

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,208 @@
...
+        storages_by_component_type = [clp_config.logs_input, clp_config.archive_output.storage]
+    elif LOG_VIEWER_WEBUI_COMPONENT_NAME == component_type:
```
haiqi96, line 131:
```suggestion
    elif component_type in (LOG_VIEWER_WEBUI_COMPONENT_NAME):
```

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,208 @@
...
+        storages_by_component_type = [clp_config.stream_output.storage]
+    elif (
+        QUERY_SCHEDULER_COMPONENT_NAME == component_type
+        or QUERY_WORKER_COMPONENT_NAME == component_type
+    ):
```
haiqi96, line 136:
```suggestion
    elif component_type in (
        QUERY_SCHEDULER_COMPONENT_NAME,
        QUERY_WORKER_COMPONENT_NAME
    ):
```

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,208 @@
...
+        if os.getenv("AWS_SESSION_TOKEN"):
+            raise ValueError(
+                "AWS_SESSION_TOKEN not supported for environmental variable credentials."
+            )
+
+    return (config_mount, credentials_env_vars)
+
+
+def _create_s3_client(s3_config: S3Config) -> boto3.client:
+    config = Config(retries=dict(total_max_attempts=3, mode="adaptive"))
```
haiqi96, line 179: let's make config an optional argument (with None as default), and let caller passes it in.
For now, I don't see issue applying `total_max_attempts=3` to ever client, but it's possible there will be some configs we want to apply case-by-case instead of to all clients

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -637,6 +669,30 @@ def validate_logs_dir(self):
...
+                profile_auth_used = True
+                break
+
+        if profile_auth_used and self.aws_config_directory is None:
+            raise ValueError("aws_config_directory must be set when using profile authentication")
+        elif not profile_auth_used and self.aws_config_directory is not None:
+            raise ValueError(
+                "aws_config_directory should not be set when profile authentication is not used"
+            )
+
```
haiqi96, line 695: we should also verify that the directory exists if `profile_auth_used=true`, right?
In order to implement the check, you may need to resturcture the condition of the if-else statement

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -268,6 +269,14 @@ def generate_container_config(
             container_clp_config.stream_output.get_directory(),
         )
 
+    # Only create the mount if the directory exists
+    if clp_config.aws_config_directory.exists():
```
haiqi96, line 273: Note: after `validate_aws_config_dir` gets updated, this should be updated to 
```if clp_config.aws_config_directory is not None```

**Review by haiqi96 — COMMENTED** [review]
some nits.

File: components/clp-package-utils/clp_package_utils/scripts/start_clp.py (hunk 40 above)
Comment (Eden-D-Zhang) [reply components/clp-package-utils/clp_package_utils/scripts/start_clp.py:91]: Maybe? The original code usually checked for each mount.

Comment (haiqi96) [reply components/clp-package-utils/clp_package_utils/scripts/start_clp.py:91]: Trace up the code a bit, can you please update the input type to `Optional[List[Optional[DockerMount]]]`?

File: components/clp-py-utils/clp_py_utils/s3_utils.py [hunk 50]
```diff
@@ -1,18 +1,200 @@
...
+    :return: Tuple of (whether aws config mount is needed, credential env_vars to set).
+    :raises: ValueError if environment variables are not set correctly.
+    """
+    storages_by_component_type = []
+    if component_type in (
+        COMPRESSION_SCHEDULER_COMPONENT_NAME, 
+        COMPRESSION_WORKER_COMPONENT_NAME
+    ):
+        storages_by_component_type = [clp_config.logs_input, clp_config.archive_output.storage]
+    elif component_type in (LOG_VIEWER_WEBUI_COMPONENT_NAME):
```
haiqi96, line 122: Sorry, need to add ',' at the end
```suggestion
    elif component_type in (LOG_VIEWER_WEBUI_COMPONENT_NAME,):
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -314,52 +321,76 @@ class Queue(BaseModel):
...
+        if profile and credentials:
+            raise ValueError("profile and credentials cannot be set simultaneously.")
+        if AwsAuthType.profile == auth_type:
+            if not profile:
+                raise ValueError(f"profile must be set when type is '{auth_type}.'")
+        elif AwsAuthType.credentials == auth_type:
+            if not credentials:
+                raise ValueError(f"credentials must be set when type is '{auth_type}.'")
+        elif auth_type in [AwsAuthType.ec2, AwsAuthType.env_vars] and (profile or credentials):
+            raise ValueError(f"profile and credentials must not be set when type is '{auth_type}.'")
```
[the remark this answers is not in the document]
Eden-D-Zhang: I don't think we can use an else here because the values need to be returned

docs/src/user-guide/guides-using-object-storage/clp-config.md
```diff
@@ -19,17 +19,27 @@ logs_input:
...
   * `<region-code>` is the AWS region [code][aws-region-codes] for the bucket.
   * `<bucket-name>` is the bucket's name.
   * `<key-prefix>` is the prefix of all logs you wish to compress and should be the same as the
     `<all-logs-prefix>` value from the [compression IAM policy][compression-iam-policy].
-  * `credentials` contains the CLP IAM user's credentials.
+  * `aws_authentication` configures the type of authentication used and additional information if
+    needed.
+    * `type` must be one of `credentials`, `profile`, `env_vars`, and `ec2`.
+    * If using long-term IAM credentials, they should be provided under the `credentials` field.
+    * If using a named profile, the name should be provided under the `profile` field.
```
haiqi96, line 40: nit; what would be a unnamed profile? My first interpretion is that "default" is a unnamed profile, but i remember that even for default, you will need to explicitly provide the name? If that's the case, let's remove "named"

components/clp-py-utils/clp_py_utils/clp_config.py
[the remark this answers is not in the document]
haiqi96: True, forgot that the previous if-elif are checking not only the auth_type. in that case, can you change the `elif` to `if`? since in each if you will early return with raise, we don't need to use elif.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -314,52 +321,76 @@
...
+            raise ValueError("profile and credentials cannot be set simultaneously.")
+        if AwsAuthType.profile == auth_type and not profile:
+            raise ValueError(f"profile must be set when type is '{auth_type}.'")
+        elif AwsAuthType.credentials == auth_type and not credentials:
+            raise ValueError(f"credentials must be set when type is '{auth_type}.'")
+        elif auth_type in [AwsAuthType.ec2, AwsAuthType.env_vars] and (profile or credentials):
+            raise ValueError(f"profile and credentials must not be set when type is '{auth_type}.'")
+        else:
+            raise ValueError(f"Unsupported authentication type '{auth_type}.'")
+        return values
```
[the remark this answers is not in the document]
haiqi96: I think this is a valid suggestion.

components/clp-py-utils/clp_py_utils/clp_config.py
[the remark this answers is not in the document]
Eden-D-Zhang: Sure

docs/src/user-guide/guides-using-object-storage/index.md
```diff
@@ -27,21 +27,30 @@ will be added in a future release.
...
+        with the corresponding credential keys.
+    * An IAM role attached to an EC2 instance (that CLP is hosted on).
+
+    
     :::{note}
-    CLP currently requires IAM user (long-term) credentials to access the relevant S3 buckets.
-    Support for other authentication methods (e.g., temporary credentials) will be added in a future
-    release.
+    You may use a single authentication method for all use cases, or a separate one for each.
+    Short-term STS credentials (that include a Session Token) are unsupported for all methods.
```
haiqi96, line 53:
```suggestion
    Short-term STS credentials (that include a Session Token) are not supported for any method.
```

components/clp-py-utils/clp_py_utils/clp_config.py
Eden-D-Zhang: That's the first if statement of the block

components/clp-package-utils/clp_package_utils/general.py
Eden-D-Zhang: Is that right? The `validate_aws_config_dir` function passes if the directory doesn't exist, which we don't want here. Also it doesn't return anything.

docs/src/user-guide/guides-using-object-storage/clp-config.md
Eden-D-Zhang: named profile is just what they're called in the AWS docs.

haiqi96: I see, sorry for my oversight. let's keep the name

components/clp-py-utils/clp_py_utils/clp_config.py
haiqi96: I meant we should verify if the directory actgually exists on the file system.

components/clp-package-utils/clp_package_utils/general.py
haiqi96: iirc, our intension is to stop and return an error to user if the given directory doesn't exist. Note because the default of aws_config_directory is None, it must be provided by user if it's not None. Then there wouldn't be a case where the aws_config_directory "validly" doesn't exist. 

My proposal is to validate that the directory exist in ```validate_aws_config_dir```.

Eden-D-Zhang: Right, but the validate function checks two cases (profile enabled and directory exists, no profile and no directory), but here we only want the first case.

haiqi96:
Tested on my end and didn't see any issue. Eden, can you fix the conflicts?

Also pending @junhaoliao's review on Webui components and Kirk's review on docs.

haiqi96 (approved):
Thanks for the hardwork. The change in package's python scripts look reasonable to me.
Waiting for the other two reviewers

haiqi96:
While testing, I have noticed one issue for S3 ingestion:
1. It seems we can't specify a bucket with out a prefix: (i.e. if I want to ingest all files under <bucket>). Currently, we are not able to leave the prefix as empty. 
2. The CLP can't ingest a single file, since the s3 config will require the input path to end with a "/".

I have missed those in https://github.com/y-scope/clp/pull/788, so they are unrelated to this PR. Just document it here.

Eden-D-Zhang:
> While testing, I have noticed one issue for S3 ingestion:
> 
> 1. It seems we can't specify a bucket with out a prefix: (i.e. if I want to ingest all files under ). Currently, we are not able to leave the prefix as empty.
> 2. The CLP can't ingest a single file, since the s3 config will require the input path to end with a "/".
> 
> I have missed those in #788, so they are unrelated to this PR. Just document it here.

For point 2, you can ingest a single file under the s3_config prefix. For example, if the input prefix in config is "logs/", you should be able to specify the file name (i.e "log1.json") to compress ingest the file with path "logs/log1.json".

Eden-D-Zhang:
> > While testing, I have noticed one issue for S3 ingestion:
> > 
> > 1. It seems we can't specify a bucket with out a prefix: (i.e. if I want to ingest all files under ). Currently, we are not able to leave the prefix as empty.
> > 2. The CLP can't ingest a single file, since the s3 config will require the input path to end with a "/".
> > 
> > I have missed those in #788, so they are unrelated to this PR. Just document it here.
> 
> For point 2, you can ingest a single file under the s3_config prefix. For example, if the input prefix in config is "logs/", you should be able to specify the file name (i.e "log1.json") to compress ingest the file with path "logs/log1.json".

Actually I think there's a problem with doing this right now, I'll look into it.

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -863,26 +887,24 @@ def start_webui(instance_id: str, clp_config: CLPConfig, mounts: CLPDockerMounts
...
     # fmt: on
+    env_vars = [
+        f"NODE_PATH={node_path}",
+        f"MONGO_URL={clp_config.results_cache.get_uri()}",
+        f"PORT={clp_config.webui.port}",
+        f"ROOT_URL=http://{clp_config.webui.host}",
+        f"METEOR_SETTINGS={json.dumps(meteor_settings)}",
+        f"CLP_DB_USER={clp_config.database.username}",
+        f"CLP_DB_PASS={clp_config.database.password}",
+        f"WEBUI_LOGS_DIR={container_webui_logs_dir}",
```
junhaoliao, line 901: if the removal isn't intentional
```suggestion
        f"WEBUI_LOGS_DIR={container_webui_logs_dir}",
        f"WEBUI_LOGGING_LEVEL={clp_config.webui.logging_level}",
```

components/log-viewer-webui/server/src/plugins/S3Manager.ts
```diff
@@ -63,15 +70,21 @@ class S3Manager {
...
+        }
 
-    console.log(`Initializing S3Manager with region="${region}"...`);
-    app.decorate("s3Manager", new S3Manager(region));
-});
+        const logMessage = null === profile ?
+            `Initializing S3Manager with region="${region}"` :
+            `Initializing S3Manager with region="${region}" and profile="${profile}"...`;
+
+        console.log(logMessage);
```
junhaoliao, line 84: In my opinion, it's fine to log `profile=null` since only developers / service maintainers would read the logs. It's even easier for debugging when we always log `profile=xxx`.
```suggestion
        console.log(`Initializing S3Manager with region="${region}" and profile="${profile}"...`);
```

components/log-viewer-webui/server/src/plugins/S3Manager.ts
```diff
@@ -21,11 +21,18 @@ class S3Manager {
 
     /**
      * @param region
+     * @param [profile]
      */
-    constructor (region: string) {
-        this.#s3Client = new S3Client({
+    constructor (region: string, profile: Nullable<string>) {
+        const s3ClientArgs: {region: string; profile?: string} = {
```
junhaoliao, line 27: syntactic sugar + we want to explicitly check Nullable values against `null`:
```
this.#s3Client = new S3Client({
    region,
    ...((null !== profile) && {profile}),
});
```

junhaoliao (changes requested):
The webui-side changes look nicely done. I'll approve again once some nit picking comments are addressed.

docs/src/user-guide/guides-using-object-storage/index.md
```diff
@@ -27,21 +27,31 @@ will be added in a future release.
...
-      * Choose the "Other" use case to generate long-term credentials.
-
+    * A named profile configured by the AWS CLI in your AWS config directory (typically `~/.aws`).
+      * Profiles may make use of a variety of AWS authentication mechanisms, including:
+        * IAM Identity Center.
+        * Assuming an IAM role.
+        * Long-term IAM user credentials.
+      * Follow [this guide][aws-configure-profiles] for more information on
+        configuring profiles with the AWS CLI.
+    * Long-term AWS credentials set as environmental variables.
```
kirkrodrigues, line 45:
```suggestion
    * Long-term AWS credentials set as environment variables.
```

The computer science convention is to call them "environment variables." Unfortunately, I don't know the origin of the phrase, so I can't really explain why they chose "environment" over "environmental".

docs/src/user-guide/guides-using-object-storage/index.md
```diff
@@ -27,21 +27,31 @@ will be added in a future release.
...
+      * Set the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environmental variables
+        with the corresponding credential keys.
+    * An IAM role attached to an EC2 instance (that CLP is hosted on).
+
+
```
kirkrodrigues, line 50:
```suggestion
```

docs/src/user-guide/guides-using-object-storage/index.md
```diff
@@ -27,21 +27,31 @@ will be added in a future release.
...
+        with the corresponding credential keys.
+    * An IAM role attached to an EC2 instance (that CLP is hosted on).
+
+    
     :::{note}
-    CLP currently requires IAM user (long-term) credentials to access the relevant S3 buckets.
-    Support for other authentication methods (e.g., temporary credentials) will be added in a future
-    release.
+    You may use a single authentication method for all use cases, or a separate one for each.
+    Short-term STS credentials (that include a Session Token) are not supported directly. To use
```
kirkrodrigues, line 53: * Can we split this up into its own note/warning?
* Can we link to some AWS docs about what STS credentials are?

components/package-template/src/etc/clp-config.yml
```diff
@@ -105,3 +105,6 @@
 #
 ## Location where logs are stored. It will be created if it doesn't exist.
 #logs_directory: "var/log"
+#
+## Location where the `.aws` directory should be read from
```
kirkrodrigues, line 109:
```suggestion
## Location of the AWS tools' config files (e.g., `~/.aws`)
```

kirkrodrigues (changes requested):
Docs review

docs/src/user-guide/guides-using-object-storage/clp-config.md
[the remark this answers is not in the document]
Eden-D-Zhang: We don't want `aws_config_directory` to have duplicates.

components/package-template/src/etc/clp-config.yml
```diff
@@ -105,3 +105,6 @@
 #
 ## Location where logs are stored. It will be created if it doesn't exist.
 #logs_directory: "var/log"
+#
+## Location of the AWS tools' config files (e.g., `~/.aws`)
+#aws_config_directory: None
```
kirkrodrigues, line 110: Does `None` work in YAML? I thought the value should be `null`, no?

kirkrodrigues:
@Eden-D-Zhang Can you review the changes I made?

components/package-template/src/etc/clp-config.yml
Eden-D-Zhang: Yeah looks like it should be null

docs/src/user-guide/guides-using-object-storage/clp-config.md
```diff
@@ -135,14 +95,91 @@ stream_output:
...
+For each use case above, you can configure its AWS authentication method through the
+`aws_authentication` config object, which includes the authentication method type to use and any
+additional settings necessary for the chosen authentication type.
+
+Settings for each type are described below:
+
+* [credentials](#credentials)
+* [profile](#profile)
+* [env_vars](#env_vars)
+* [ec_2](#ec_2)
```
Eden-D-Zhang: There shouldn't be an underscore for the EC2 type since the implementation uses `ec2`.

Eden-D-Zhang (commented):
Small inconsistency, otherwise LGTM

docs/src/user-guide/guides-using-object-storage/clp-config.md
Eden-D-Zhang: Just noticed this was fixed, ignore this comment.

docs/src/user-guide/guides-using-object-storage/clp-config.md
```diff
@@ -93,12 +95,91 @@ stream_output:
...
+  profile: "<profile-name>"
+```
+
+* `<profile-name>` should be the name of an existing [AWS CLI profile](index.md#named-profiles).
+
+In addition, `aws_config_directory` must be set to the directory containing the profile
+configurations (typically `~/.aws`):
+
+```yaml
+aws_config_directory: "<aws-config-dir>"
```
Eden-D-Zhang, line 152: I agree with the CodeRabbit nitpick suggestion about clarifying this is a top-level suggestion, especially since it's buried at the bottom of the template.

docs/src/user-guide/guides-using-object-storage/clp-config.md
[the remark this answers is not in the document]
haiqi96: I think the high-level argument is that:
1. it's unnecessary if user only uses a single config directory for all components
2. We don't want to support using multiple config directories. 

For 2, it would become tricky if user specifies different `aws_config_directory ` for different component. by default, boto3 only uses config under $HOME/.aws (which where container mounts the config to).
While it is possible to let different boto3 clients load from different config files, it would unnecessarily complicates the code. 

In addition, I don't see much value supporting 2. User can have a fine-grained permission control by simply using different profiles for different components.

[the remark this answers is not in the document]
kirkrodrigues: Yeah, I remembered after Eden replied.

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,186 @@
...
+    """
+    Generates Docker container authentication options for AWS S3 access based on the given type.
+    Handles authentication methods that require extra configuration (profile, env_vars).
+
+    :param clp_config: CLPConfig containing storage configurations.
+    :param component_type: Type of calling container (compression, log_viewer, or query).
+    :return: Tuple of (whether aws config mount is needed, credential env_vars to set).
+    :raises: ValueError if environment variables are not set correctly.
+    """
+    storages_by_component_type = []
```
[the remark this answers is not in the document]
Eden-D-Zhang: Those are the right classes, but I'm not sure if the syntax is right.

junhaoliao (dismissed):

junhaoliao (dismissed):

components/clp-py-utils/clp_py_utils/s3_utils.py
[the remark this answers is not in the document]
haiqi96: True, forgot the Union.

junhaoliao:
The webui changes lgtm

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -1,18 +1,185 @@
...
+    elif AwsAuthType.ec2 == auth.type:
+        aws_credentials = _get_session_credentials()
+        if aws_credentials is None:
+            raise ValueError(f"Failed to authenticate with EC2 metadata.")
+    else:
+        raise ValueError(f"Unsupported authentication type: {auth.type}")
+
+    env_vars = {
+        "AWS_ACCESS_KEY_ID": aws_credentials.access_key_id,
+        "AWS_SECRET_ACCESS_KEY": aws_credentials.secret_access_key,
```
kirkrodrigues, line 82: Can we replace instances of AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY with string constants? Just to avoid typos causing issues in the future.

Eden-D-Zhang: Do you have a preference for what they should be named

kirkrodrigues: How about AWS_ENV_VAR_ACCESS_KEY_ID and AWS_ENV_VAR_SECRET_ACCESS_KEY?

docs/src/user-guide/guides-using-object-storage/index.md
```diff
@@ -27,22 +27,60 @@ will be added in a future release.
...
+
+### Supported AWS authentication methods
+
+clp-json currently supports the AWS authentication methods described below.
+
+:::{caution}
+Short-term [STS credentials][aws-sts-credentials] (which include a Session Token) are not supported
+directly. To use temporary credentials, use [named profiles](#named-profiles) (with
+IAM Identity Center authentication or IAM role assumption) to generate them dynamically instead.
+:::
```
haiqi96, line 46: this part is inaccurate, because while using named profiles, the process of generating temporary credentials is hidden from the user. How about changing it to something like

```suggestion
:::{caution}
Short-term [STS credentials][aws-sts-credentials] (which include a Session Token) are not supported
directly. Instead, use [named profiles](#named-profiles) (with
IAM Identity Center authentication or IAM role assumption) which provide the required permissions without manual credential handling.
:::
```

haiqi96 (commented):
The code change looks good to me. Added one more suggestion to the doc.

haiqi96 (dismissed):
Looks good to me.

kirkrodrigues (dismissed):
I edited the PR title slightly. Let me know if y'all disagree.

kirkrodrigues (approved):

PR #745, opened 2025-03-05
build(cmake): Require compiler versions that support the C++20 features necessary for compiling `ystdlib-cpp`; Add toolchain file for LLVM Clang 16 on macOS 13 & 14.

Bill-hbrhbr:
# Description

To meet the `c++20` feature requirements of [`ystdlib-cpp`](https://github.com/y-scope/ystdlib-cpp/pull/29), we must use Clang/AppleClang version 16 or later.

On macOS 13 and 14, a good solution is to install LLVM Clang v16 compilers. However, by default CMake only picks up the new toolchain's `llvm-ranlib`, while still using `/usr/bin/ar` for AR, which creates a mismatch in tool versions. For instance, after running `brew install llvm@16` on macOS 14:
```cmake
-- CMAKE_AR: /usr/bin/ar
-- CMAKE_RANLIB: /opt/homebrew/opt/llvm@16/bin/llvm-ranlib
```
It’s essential that both AR and RANLIB originate from the same LLVM distribution, otherwise [linker errors](https://stackoverflow.com/questions/77444892/xcode-15-0-1-macos-sonoma-clang-archive-or-linking-issue) will arise.

While one could technically use `/usr/bin/ar` together with `/usr/bin/ranlib`, since we're using LLVM's `clang` and `clang++`, it's best to ensure consistency by using `llvm-ar` and `llvm-ranlib` from the same distribution.

To ensure consistency within the toolchain, we add an OS platform detection + loading the appropriate toolchain setting file before the `project(CLP ...)` call. We also add a compiler check after the toolchains have been loaded.
All the new CMake utilities are inside `./cmake/Toolchains`.

# Checklist

  breaking change.

# Validation performed

* [X] Existing workflows run successfully.

## Summary by CodeRabbit

- **Chores**
  - Refined the macOS build workflow to enhance dependency management and configuration.
  - Introduced the installation of a newer LLVM version to improve compiler support.
  - Improved toolchain setup and compatibility checks for a smoother build and development experience.
- **Documentation**
  - Updated compiler requirements to specify support for multiple C++20 features, enhancing clarity for users.


Bill-hbrhbr:
@coderabbitai review

kirkrodrigues (changes requested):
This PR needs a bit more to be complete. The reason our CI tests on macOS is so that we ensure users who use macOS can compile (and lint) CLP's core on macOS. Towards that goal, this PR doesn't really tell macOS 13 and 14 users how to compile CLP-core, so at the very least we need to update the [docs](https://github.com/y-scope/clp/blob/main/docs/src/dev-guide/components-core/index.md#native-environment). That said, I think a better approach might be:

* Add a set of compiler requirements in the main CMakeLists.txt like we did in [ystdlib-cpp](https://github.com/y-scope/ystdlib-cpp/blob/main/CMakeLists.txt#L7-L26).
* For macOS 13 & 14, add a CMake toolchain file that sets the required CMake variables for the compilers, `ar`, and `ranlib`.
* In the main CMakeLists.txt, if we detect the current OS is macOS 13 or 14, load this new toolchain file.

The benefits of this approach is that the user shouldn't need to set any paths or CMake variables manually. I haven't tested this approach, so maybe it will have some unforeseen issues, but I think it shouldn't take you too long to try.

Bill-hbrhbr:
@kirkrodrigues I followed your suggestions thru point 1 to 3. For updating the docs, since now I've moved installing `llvm@16` to `components/core/tools/scripts/lib_install/macos/install-all.sh`, I no longer think we need to explicitly update the docs.

components/core/CMakeLists.txt
```diff
@@ -1,6 +1,19 @@
 cmake_minimum_required(VERSION 3.16.3)
+
+set(CMAKE_TOOLCHAINS_DIR
```
kirkrodrigues, line 3: * Do we really need this variable?
* Don't forget we need to prefix it with the project name.

components/core/cmake/Toolchains/llvm-clang-toolchains.cmake
kirkrodrigues, line 1: The name should be `llvm-clang-toolchain.cmake` (singular), right?

components/core/cmake/Toolchains/llvm-clang-toolchains.cmake
```diff
@@ -0,0 +1,13 @@
+message(STATUS "Setting up LLVM v16 toolchains for MacOS 13 Ventura and MacOS 14 Sonoma...")
```
kirkrodrigues, line 1:
```suggestion
message(STATUS "Setting up LLVM v16 toolchain...")
```

components/core/cmake/Toolchains/llvm-clang-toolchains.cmake
```diff
@@ -0,0 +1,13 @@
...
+
+execute_process(
+    COMMAND
+        brew --prefix llvm@16
```
kirkrodrigues, line 5:
```suggestion
        "brew" "--prefix" "llvm@16"
```

Let's quote strings that aren't parameter or variable names to make the scripts more readable.

components/core/cmake/Toolchains/utils.cmake
```diff
@@ -0,0 +1,40 @@
+function(setup_toolchains)
+    if("Darwin" STREQUAL "${CMAKE_HOST_SYSTEM_NAME}")
+        execute_process(
+            COMMAND
+                sw_vers --productVersion
```
kirkrodrigues, line 5:
```suggestion
                "sw_vers" "--productVersion"
```

components/core/cmake/Toolchains/utils.cmake
```diff
@@ -0,0 +1,40 @@
...
+            OUTPUT_VARIABLE CMAKE_OSX_VERSION
```
kirkrodrigues, line 6: Let's not prefix the variable name with `CMAKE_` to avoid confusion. How about just `MACOS_VERSION`?

components/core/cmake/Toolchains/utils.cmake
```diff
@@ -0,0 +1,40 @@
...
+                "Toolchain file"
+            )
+        endif()
+    endif()
+endfunction()
+
+# Require compiler versions that support the C++20 features necessary for compiling CLP
+function(validate_compiler_versions)
+	if("AppleClang" STREQUAL "${CMAKE_CXX_COMPILER_ID}")
+		set(CLP_CMAKE_CXX_COMPILER_MIN_VERSION "16")
```
kirkrodrigues, line 22: Since functions have their own scope, how about let's simplify the name to `CXX_COMPILER_MIN_VERSION`?

components/core/tools/scripts/lib_install/macos/install-all.sh
```diff
@@ -16,6 +16,7 @@ brew install \
   go-task \
   java11 \
   libarchive \
+  llvm@16 \
```
kirkrodrigues, line 19: What happens if the user already has, for instance, llvm@19 installed?

kirkrodrigues (changes requested):

components/core/tools/scripts/lib_install/macos/install-all.sh
Bill-hbrhbr: As discussed offline, brew installs different versions of llvm clang toolchains side-by-side, so this should be a non-issue.

Bill-hbrhbr:
@coderabbitai void the rest of your existing comments. Too verbose for this PR.

components/core/cmake/Toolchains/llvm-clang-toolchain.cmake
```diff
@@ -0,0 +1,22 @@
+message(STATUS "Setting up LLVM v16 toolchain...")
+
+execute_process(
+    COMMAND
+        "brew" "--prefix" "llvm@16"
+    RESULT_VARIABLE BREW_RESULT
+    OUTPUT_VARIABLE LLVM_TOOLCHAIN_PREFIX
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+
```
kirkrodrigues, line 10:
```suggestion
```

To group error checking with the statement that may generate the error.

components/core/cmake/Toolchains/utils.cmake
```diff
@@ -0,0 +1,49 @@
+# This file contains utility functions for setting up toolchains and validating toolchain versions
+# to ensure compatibility with C++20 features required by the project.
```
kirkrodrigues, line 2:
```suggestion
# to ensure compatibility with the C++20 features required by the project.
```

components/core/cmake/Toolchains/utils.cmake
```diff
@@ -0,0 +1,49 @@
...
+
+# Sets up the appropriate toolchain file based on the host system.
+# For macOS versions below 15, uses the LLVM Clang toolchain.
+function(setup_toolchains)
```
kirkrodrigues, line 6:
```suggestion
function(setup_toolchains)
    # For macOS versions below 15, use the LLVM 16 Clang toolchain.
```

components/core/cmake/Toolchains/llvm-clang-toolchain.cmake
kirkrodrigues, line 1: Sorry, since we're hardcoding the version to 16, can we rename this file to llvm-16-clang-toolchain.cmake?

components/core/cmake/Toolchains/utils.cmake
```diff
@@ -0,0 +1,49 @@
...
+                "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Toolchains/llvm-clang-toolchain.cmake"
+                CACHE STRING
+                "Toolchain file"
+            )
+        endif()
+    endif()
+endfunction()
+
+# Checks if the compiler ID and version meet the minimum requirements to support C++20 features
+# required by the project.
```
kirkrodrigues, line 25:
```suggestion
# required by the project:
```

components/core/CMakeLists.txt
```diff
@@ -1,6 +1,14 @@
 cmake_minimum_required(VERSION 3.16.3)
+
+# Toolchain setup must come before the first project() call in the entire CMake buildsystem.
+# If CLP is not the top-level project, the following setup has no effect.
+include("cmake/Toolchains/utils.cmake")
```
kirkrodrigues, line 5:
```suggestion
include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/Toolchains/utils.cmake")
```

Unless that's wrong for some reason?

kirkrodrigues (changes requested):
A few more suggestions.

components/core/cmake/Toolchains/llvm-clang-toolchain.cmake
Bill-hbrhbr: renamed to `llvm-clang-16-toolchain.cmake`

kirkrodrigues (commented):
Can you change [this line](https://github.com/y-scope/clp/blob/36140bf46b8d4d93b4383c2b17ebe6dc5cfebe1e/docs/src/dev-guide/components-core/index.md?plain=1#L9) to:

```text
A recent compiler that supports C++20
```

The rationale being that we can avoid having to keep that doc up-to-date with the precise versions required, but if the user has too low a compiler version, the validation in this PR will inform them.

For the PR title, how about:

```
build(cmake): Require compiler versions that support the C++20 features necessary for compiling `ystdlib-cpp`; Add toolchain file for LLVM Clang 16 on macOS 13 & 14.
```

kirkrodrigues (approved):

PR #754, opened 2025-03-10
refactor(clp-s): Divide clp-s search AST code so that it can easily be turned into a library.

gibber9809:
# Description

This PR modifies the clp-s codebase so that the build can be split into a "clp_s::TimestampPattern" library, a "clp_s::search::ast" library, and the core clp binary.

To clean up this split we move around some code to eliminate circular dependencies and delete some duplicated/unnecessary code in the process.

These changes include:
* Moving tokenize_column_descriptor, unescape_kql_value, and related utilities from StringUtils:: to SearchUtils (which is part of the core ast build)
* Deleting the wildcard_match implementation in SearchUtils -- it is a duplicate of wildcard_match_unsafe from StringUtils that doesn't handle case sensitivity (remaining uses of wildcard_match have been replaced with the StringUtils implementation)
* Moving node_to_literal_type from SearchUtils.hpp to SchemaTree.hpp

# Checklist

  breaking change.

# Validation performed
* Verified that unit tests continue to pass
* Manually verified that refactors allow case insensitive wildcard search inside of unstructured arrays to work correctly

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced modules for search and timestamp pattern processing that improve query string tokenization and handling.
  - Added functions for converting four-byte hexadecimal sequences to UTF-8 and unescaping KQL values.
  - Implemented a new function to map node types to literal types.

- **Refactor**
  - Consolidated and restructured search utilities and string processing functions to streamline query operations.
  - Updated function calls to utilize a new namespace for improved organization.

- **Chores**
  - Updated build configurations and dependency management by removing deprecated components and adding new library linkages.
  - Expanded the set of source files compiled for the indexer executable.


Bill-hbrhbr:
At a first glance this PR looks a bit big, could at least separate library encapsulation and implementation detail changes into 2 separate PRs.

Also I'm wondering if you think removing `StringUtils` in `clp-s` and add the missing components to `clp::string_utils` is a good idea.

gibber9809:
> At a first glance this PR looks a bit big, could at least separate library encapsulation and implementation detail changes into 2 separate PRs.
> 
> Also I'm wondering if you think removing `StringUtils` in `clp-s` and add the missing components to `clp::string_utils` is a good idea.

OK -- will revert CMakeLists changes from this PR and leave just the code changes in this PR.

And yeah, removing clp_s::StringUtils and switching over to `clp::string_utils` for most stuff is the plan eventually.

components/core/src/clp_s/TimestampPattern.hpp
```diff
@@ -71,7 +70,10 @@ class TimestampPattern {
 
     // Methods
     /**
-     * Static initializer for class. This must be called before using the class.
+     * Initializes global data needed to use this class. This function must be called before using
```
wraymo, line 73:
```suggestion
     * Initializes global data required by this class. This function must be called before any other methods of the class are used.
```

components/core/src/clp_s/search/kql/kql.cpp
```diff
@@ -85,13 +86,13 @@ class ParseTreeVisitor : public KqlBaseVisitor {
         } else if (auto ret = NullLiteral::create_from_string(token)) {
             return ret;
         } else {
-            return StringLiteral::create(StringUtils::clean_up_wildcard_search_string(token));
+            return StringLiteral::create(clp::string_utils::clean_up_wildcard_search_string(token));
```
wraymo, line 89: Are we going to remove duplicated code in `StringUtils` and use `clp::string_utils` instead?

components/core/src/clp_s/search/SearchUtils.hpp
```diff
@@ -37,12 +32,32 @@ LiteralType node_to_literal_type(NodeType type);
 bool double_as_int(double in, FilterOperation op, int64_t& out);
 
 /**
- * Performs a wildcard match of a string against a pattern
- * @param s the string to match
- * @param p the pattern to match against
- * @return true if s matches p, false otherwise
+ * Converts a KQL string column descriptor delimited by '.' into a list of tokens. The
```
wraymo, line 35: Do you think it's better to put those functions in a class (maybe `SearchUtils`) like `StringUtils`?

wraymo (commented):
Thanks for the refactoring! The code is more modular now.

components/core/src/clp_s/search/kql/kql.cpp
gibber9809: In general yes, but I think that that should probably happen as its own PR. The only reason I make this change here is because I want `kql` to be independent of `clp_s/Utils`.

wraymo (approved):

PR #769, opened 2025-03-15
fix(clp-s): Fix authoritative timestamp parsing bug introduced in #731.

gibber9809:
# Description

This PR fixes a bug introduced during some refactoring in #731. In particular, that PR adds the condition `m_authoritative_timestamp.size() > (m_matched_timestamp_prefix_length + 1)` to an if statement wrapping the rest of the timestamp key resolution logic, which prevented being able to correctly match the root of a given object subtree for keys consisting of a single token.

The fix simply involves removing this condition from that outermost if branch and adjusting some other conditions accordingly.

Alternatively we could simplify the matching logic by inserting the namespace into the start of the vector of tokens and use that to remove the special case for matching the root of an object subtree.

# Checklist

  breaking change.

# Validation performed
* Manually validated that timestamp key can be recognized for variety of kv-ir and json streams

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced safety and reliability in internal archival processing by streamlining validation checks, reducing the risk of runtime errors.

### DIFF
--- a/components/core/src/clp_s/ArchiveWriter.cpp
+++ b/components/core/src/clp_s/ArchiveWriter.cpp
@@ -241,14 +241,15 @@ ArchiveWriter::append_message(int32_t schema_id, Schema const& schema, ParsedMes
 
 int32_t ArchiveWriter::add_node(int parent_node_id, NodeType type, std::string_view key) {
     auto const node_id{m_schema_tree.add_node(parent_node_id, type, key)};
-    if (NodeType::Object == type && m_matched_timestamp_prefix_node_id == parent_node_id
-        && m_authoritative_timestamp.size() > (m_matched_timestamp_prefix_length + 1))
-    {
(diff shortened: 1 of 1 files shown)


wraymo (commented):
Can you show an example what this PR fixes?

gibber9809:
> Can you show an example what this PR fixes?

All cases -- the bug prevents timestamp parsing from ever being successful. Consider a simple case of timestamp `"ts"` in namespace `""`. The vector of timestamp tokens will be of size 1, so the topmost condition of `m_authoritative_timestamp.size() > (m_matched_timestamp_prefix_length + 1)` will never be true because the matched prefix starts at 0 and `1 > 1` will never be true.

Adding the namespace to the start of the timestamp token vector can also solve the bug/simplify the code overall, but it requires more code change so I've gone with the change in this PR for now.

wraymo (approved):

PR #779, opened 2025-03-20
fix(clp-s): Reduce compiler warnings

wraymo:
# Description

This PR fixes a lot of compiler warnings when building `clp-s`. It is an essential step when integrating with other tools with the default `Werror` option. 

# Checklist

  breaking change.

# Validation performed
+ Built the clp-s library for velox the the default CMake flags, all previous warnings are resolved

### DIFF
--- a/components/core/src/clp_s/TimestampPattern.cpp
+++ b/components/core/src/clp_s/TimestampPattern.cpp
@@ -111,7 +111,7 @@ append_padded_value_notz(int value, char padding_character, size_t max_length, s
     if ("0" != value_str) {
         str.append(max_length - value_str.length(), padding_character);
         size_t last_zero = string::npos;
-        for (size_t last = value_str.size() - 1; last >= 0; --last) {
+        for (int last = value_str.size() - 1; last >= 0; --last) {
             if (value_str[last] == '0') {
                 last_zero = last;
             } else {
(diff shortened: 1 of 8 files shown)


components/core/src/clp_s/TimestampPattern.cpp
```diff
@@ -111,7 +111,7 @@ append_padded_value_notz(int value, char padding_character, size_t max_length, s
     if ("0" != value_str) {
         str.append(max_length - value_str.length(), padding_character);
         size_t last_zero = string::npos;
-        for (size_t last = value_str.size() - 1; last >= 0; --last) {
+        for (int last = value_str.size() - 1; last >= 0; --last) {
```
wraymo, line 114: This line is both a warning and a potential bug.

gibber9809 (changes requested):
Nice work! Just a few small comments.

I'm also wondering if we should set the [COMPILE_WARNING_AS_ERROR](https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_WARNING_AS_ERROR.html#prop_tgt:COMPILE_WARNING_AS_ERROR) target property for all parts of the clp-s build in order to catch regressions? Would help us avoid having to repeatedly fix this issue as we update the version of clp-s used by external projects.

gibber9809 (approved):
LGTM. For PR title how about
`refactor(clp-s): Fix compiler warnings that cause issues in Velox -Werror build`

Before merging could you also create a github issue for tracking what we talked about re warnings in the side channel?

jackluo923 (approved):

PR #784, opened 2025-03-24
feat(clp-s): Add hook to Projection class to help project columns in a specific order

gibber9809:
# Description

This PR adds the `get_ordered_matching_nodes` function to the `Projection` class to allow users of the class to see how the resolved nodes correspond to their original projection in order. This function has the same semantics as `get_matching_nodes_list` from the `OrderedProjection` class in the Velox integration. Once this PR is merged the Velox integration PR should be able to replace all usage of `OrderedProjection` with the regular `Projection` class.

Adding support for tracking order in this way seems to have added no measurable overhead, which is to be expected since the overhead should be O(N) in number of columns being projected, and this overhead is incurred one time when resolving columns.

# Checklist

  breaking change.

# Validation performed

* Manually validated that projection still functions as expected
* Manually validated that there is no measurable overhead for tracking order

## Summary by CodeRabbit

- **New Features**
	- Enhanced the matching mechanism to improve accuracy and consistency of search results.
	- Introduced a new feature that enables retrieval of ordered matching results for improved result clarity.

### DIFF
--- a/components/core/src/clp_s/search/Projection.cpp
+++ b/components/core/src/clp_s/search/Projection.cpp
@@ -54,8 +54,10 @@ void Projection::resolve_column(
 
     auto cur_node_id = tree->get_object_subtree_node_id_for_namespace(column->get_namespace());
     if (-1 == cur_node_id) {
+        m_ordered_matching_nodes.emplace_back(std::vector<int32_t>{});
         return;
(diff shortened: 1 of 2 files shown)


wraymo (approved):
This looks good to me

PR #790, opened 2025-03-28
refactor(clp-s): Split up clp-s code to create clp_s::search::ast and clp_s::TimestampPattern libraries.

gibber9809:
# Description

This PR splits up the clp-s build and introduces the clp_s::search::ast and clp_s:TimestampPattern libraries.

The diff is deceptively large, but contains (almost) no real code change. Mostly this PR consists of:
* Moving ast components into clp_s/search/ast and putting them in the clp_s::search::ast namespace
* Refactoring code to respect the new clp_s::search::ast namespace
* Adding some missing includes
* Changing the lint exceptions in lint-tasks.yml to account for moving code into clp_s/search/ast
* Modifying various CMakeLists files to create and use clp_s::search::ast and clp_s::TimestampPattern
* Move StringUtils::has_unescaped_wildcards into SearchUtils (this dependency was missed by the previous refactoring PR #754)

# Checklist

  breaking change.

# Validation performed

* Validated that unit tests pass succesfully

## Summary by CodeRabbit

- **Refactor**
  - Reorganized search functionality into a dedicated AST module.
  - Updated namespaces, includes, and function signatures for improved modularity and clarity.
- **Chores**
  - Removed obsolete wildcard detection utilities.
  - Streamlined build configurations and dependency management to simplify maintenance.


components/core/src/clp_s/Utils.cpp
```diff
@@ -274,18 +274,6 @@ bool StringUtils::is_wildcard(char c) {
     return false;
 }
 
-bool StringUtils::has_unescaped_wildcards(std::string const& str) {
```
wraymo, line 277: Is it because it's related to escaping so we move it to SearchUtils?

components/core/src/clp_s/search/Output.cpp
```diff
@@ -43,14 +57,6 @@ bool Output::filter() {
         return true;
     }
 
-    // Skip decompressing archive if it won't match based on the timestamp
-    // range index
-    EvaluateTimestampIndex timestamp_index(m_archive_reader->get_timestamp_dictionary());
```
wraymo, line 48: Are we checking the index elsewhere?

wraymo (commented):
Thanks for the refactoring. I'm generally fine with the changes. Just two questions:

1. Since we're moving timestamp patterns into a separate library, do we still want to create `TimestampPatternGlobalInstance`? What's the benefit of having a separate library here?

2. Could we document somewhere what `ast` refers to? I'm not sure if that's clear enough.

components/core/src/clp_s/Utils.cpp
gibber9809: It's because it is used by `StringLiteral`. Moving it into `SearchUtils` allows the ast library to not depend on `Utils`.

components/core/src/clp_s/search/Output.cpp
gibber9809: Yeah, we check it in the main flow before schema matching if you check `clp-s.cpp`.

I took it out because I thought it was a leftover from when we had sub-archives, but now I'm realizing that the reason we have this duplicated check here is because ambiguous columns can get resolved to the timestamp column after column resolution.

I will add it back.

gibber9809:
> Thanks for the refactoring. I'm generally fine with the changes. Just two questions:
> 
>     1. Since we're moving timestamp patterns into a separate library, do we still want to create `TimestampPatternGlobalInstance`? What's the benefit of having a separate library here?
> 
>     2. Could we document somewhere what `ast` refers to? I'm not sure if that's clear enough.

For 1 I was trying to make it more obvious that users of the ast library should call `TimestampPattern::init()` and trying to make the dependencies in the build a bit clearer -- i.e. both the ast library and the core clp-s json parsing code depend on `TimestampPattern`. I can get rid of the `clp_s::TimestampPattern` library if you don't think pulling it out is useful though?

For 2 I can add a comment in the `CMakeLists.txt` file to explain a bit about the library.

gibber9809:
> > Thanks for the refactoring. I'm generally fine with the changes. Just two questions:
> > ```
> > 1. Since we're moving timestamp patterns into a separate library, do we still want to create `TimestampPatternGlobalInstance`? What's the benefit of having a separate library here?
> > 
> > 2. Could we document somewhere what `ast` refers to? I'm not sure if that's clear enough.
> > ```
> 
> For 1 I was trying to make it more obvious that users of the ast library should call `TimestampPattern::init()` and trying to make the dependencies in the build a bit clearer -- i.e. both the ast library and the core clp-s json parsing code depend on `TimestampPattern`. I can get rid of the `clp_s::TimestampPattern` library if you don't think pulling it out is useful though?
> 
> For 2 I can add a comment in the `CMakeLists.txt` file to explain a bit about the library.

I pushed the changes that I described. I can get rid of the TimestampPattern library or not depending on what you decide.

wraymo (approved):

kirkrodrigues (approved):
Approving changes external to `components/core/src/clp_s`.

Edited PR title slightly to be grammatically correct and follow our convention of finishing PR titles with a period.

Issue #795, opened 2025-04-01
Resolve compilation errors for the newest CMake releases (version >= 4.0.0)


Bill-hbrhbr:
### Bug

PR to fix: #794 

### CLP version

0.3.0

### Environment

Bare MacOS-15 machine ran with `brew install cmake`.

### Reproduction steps

Follow the usual steps of compiling CLP.

junhaoliao:
In case someone has Kitware's cmake APT source added:

### To Pin Specific Version of `cmake` from Kitware

1. Check available versions
```bash
apt list -a cmake
apt list -a cmake-data

2. Create a pinning file of arbitrary name in `/etc/apt/preferences.d/`
```
sudo nano /etc/apt/preferences.d/kitware-cmake
```
Paste
```
Package: cmake
Pin: version 3.31.6-0kitware1ubuntu22.04.1
Pin-Priority: 1001

Package: cmake-data
Pin: version 3.31.6-0kitware1ubuntu22.04.1
Pin-Priority: 1001
```
3. Update package list and install the pinned version
```
sudo apt update
sudo apt install cmake
```

PR #796, opened 2025-04-01
build(core): Lock CMake version to < 4 across all platforms to avoid third-party library compilation issues.

Bill-hbrhbr:
# Description

Currently, multiple libraries that we depend on have the following specification: 
```
cmake_minimum_required(3.x)
```
where `3.x < 3.5`. This raises a CMake deprecation error if we use the newest CMake version 4.0.0 [(release note)](https://cmake.org/cmake/help/latest/release/4.0.html), which is now installed by default if we run `brew install cmake` on macos-15.

Since these third-party dependencies may take some time to upgrade their CMake configurations, we are temporarily locking the CMake version to be below `4.0.0` to avoid compatibility issues with the newer release.

This PR should be reverted once we resolve issue #795 with PR #794.

# Checklist

  breaking change.

# Validation performed

* [X] No source code changes. Compilation succeeds and unit tests pass as usual.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new script to verify that the installed CMake version meets compatibility requirements.
  - Updated the macOS installation process to install a specific version of CMake for better component support.
  - Added a new step to check the CMake version in installation scripts for Ubuntu and CentOS.

- **Chores**
  - Streamlined installation procedures across supported operating systems with refined command syntax and additional version checks.


components/core/tools/scripts/lib_install/macos/install-all.sh
```diff
@@ -7,9 +7,15 @@ set -e
 set -u
 
 brew update
+
+# Install CMake v3.31.6 since we're using yaml-cpp that doesn't yet support CMake v4 yet
```
junhaoliao, line 11:
```suggestion
# Install CMake v3.31.6 as ANTLR and yaml-cpp do not yet support CMake v4+.
```

components/core/tools/scripts/lib_install/macos/install-all.sh
```diff
@@ -7,9 +7,15 @@ set -e
...
+cmake_formula_path=/tmp/cmake.rb
+curl --fail --show-error --location --output "${cmake_formula_path}" \
+  https://raw.githubusercontent.com/Homebrew/homebrew-core/b4e46db74e74a8c1650b38b1da222284ce1ec5ce/Formula/c/cmake.rb
```
junhaoliao, line 14:
```suggestion
curl --fail --location --show-error \
  https://raw.githubusercontent.com/Homebrew/homebrew-core/b4e46db74e74a8c1650b38b1da222284ce1ec5ce/Formula/c/cmake.rb \
  --output "${cmake_formula_path}"
```

components/core/tools/scripts/lib_install/macos/install-all.sh
```diff
@@ -7,9 +7,15 @@ set -e
...
+brew install --formula "${cmake_formula_path}"
```
junhaoliao, line 15: Can the `--formula` argument be omitted? i.e.,

```suggestion
brew install "${cmake_formula_path}"
```

junhaoliao (changes requested):

components/core/tools/scripts/lib_install/macos/install-all.sh
kirkrodrigues: The version of curl on macOS seems to require that the options come before the URL.

components/core/tools/scripts/lib_install/macos/install-all.sh
kirkrodrigues: If we omit `--formula`, then `brew` will print a warning (and it's probably better to be explicit?).

junhaoliao: Right, i wasn't aware of the warning. Being explicit is also better. Please ignore the original suggestion.

components/core/tools/scripts/lib_install/macos/install-all.sh
kirkrodrigues: Sorry, scratch that. I was misinterpreting the error I saw.

junhaoliao: hmm... the suggested command seems to work on my mac:
```
~ % curl --version
curl 8.7.1 (x86_64-apple-darwin24.0) libcurl/8.7.1 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.12 nghttp2/1.63.0
Release-Date: 2024-03-27
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS GSS-API HSTS HTTP2 HTTPS-proxy IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL threadsafe UnixSockets
```

maybe macos-13 is using an older version?

components/core/tools/scripts/lib_install/check-cmake.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+# Exit on error
+set -e
```
kirkrodrigues, line 4:
```suggestion
# Exit on any error
set -e

# Error on undefined variable
set -u
```

components/core/tools/scripts/lib_install/check-cmake.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# Get the installed cmake version string
+cmake_version=$(cmake --version | head -n1 | awk '{print $3}')
```
kirkrodrigues, line 7:
```suggestion
cmake_version=$(cmake --version | head --lines 1 | awk '{print $3}')
```

To stay consistent with our convention of using long-form options in scripts. This option is also available on macOS: https://ss64.com/mac/head.html

components/core/tools/scripts/lib_install/check-cmake.sh
```diff
@@ -0,0 +1,14 @@
...
+cmake_major_version=$(echo "$cmake_version" | cut -d. -f1)
+
+# Check if version is 4.0 or higher
+if [[ "$cmake_major_version" -ge "4" ]]; then
```
kirkrodrigues, line 11:
```suggestion
# shellcheck disable=SC2071
if [[ "$cmake_version" < "4" ]]; then
```

Semantic versions have the nice property that ordering them in ASCII-order also orders them in version order. So we can use a string comparison and get an accurate result.

components/core/tools/scripts/lib_install/check-cmake.sh
```diff
@@ -0,0 +1,14 @@
...
+  echo "CMake version $cmake_version is not supported (>= 4.0)."
```
kirkrodrigues, line 12:
```suggestion
  echo "CMake version $cmake_version is currently unsupported (>= 4.0)."
```

components/core/tools/scripts/lib_install/check-cmake.sh
kirkrodrigues, line 1: How about `check-cmake-version.sh`?

kirkrodrigues (changes requested):

components/core/tools/scripts/lib_install/check-cmake.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+# Exit on error
+set -e
+
+# Get the installed cmake version string
+cmake_version=$(cmake --version | head -n1 | awk '{print $3}')
+cmake_major_version=$(echo "$cmake_version" | cut -d. -f1)
```
junhaoliao, line 8: Would it be better to write
```suggestion
cmake_major_version=$(cmake -E capabilities | jq ".version.major")
```

components/core/tools/scripts/lib_install/check-cmake-version.sh
```diff
@@ -0,0 +1,18 @@
...
+# Error on undefined variable
+set -u
+
+# Get the installed cmake version string
+cmake_version=$(cmake -E capabilities | jq ".version.string")
+cmake_major_version=$(cmake -E capabilities | jq ".version.major")
+
+# Check if version is 4.0 or higher
+# shellcheck disable=SC2071
+if ! [[ "$cmake_major_version" < "4" ]]; then
```
kirkrodrigues, line 15: We can compare directly against `cmake_version`, right?

```suggestion
if ! [[ "$cmake_version" < "4" ]]; then
```

kirkrodrigues (changes requested):

components/core/tools/scripts/lib_install/check-cmake-version.sh
Bill-hbrhbr: fixed in the newest commit.
Need to pass `--raw-output` flag to `jq` to remove the extra quotes around cmake version string

components/core/tools/scripts/lib_install/check-cmake-version.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Error on undefined variable
+set -u
+
+# Get the installed cmake version string
+cmake_version=$(cmake -E capabilities | jq --raw-output ".version.string")
+
+# Check if version is 4.0 or higher
+# shellcheck disable=SC2071
+if ! [[ $cmake_version < "4" ]]; then
```
junhaoliao, line 14: Just curious, is the quote around `4` to make the string comparison explicit?
Also, this will only work if the reported cmake version is less than 10, right?

junhaoliao (dismissed):
cc @kirkrodrigues 

for the PR title, how about
```
fix(cmake): Lock CMake version to `<4` across all platforms to avoid third-party library compilation issues; Add version check script (fixes #795).
```

components/core/tools/scripts/lib_install/check-cmake-version.sh
kirkrodrigues: @Bill-hbrhbr Let's switch back to your original solution.

kirkrodrigues (approved):

PR #797, opened 2025-04-01
feat(new-webui): Add support for page switching using react-router.

davemarco:
# Description

The menu in the new webui is currently static, and does nothing.

This PR adds [react-router](https://reactrouter.com/home) library which makes the menu links clickable.

In old webui just relied on regular links, but using react router is supposedly more performant. 

Note we are using data mode for react-router

# Checklist

  breaking change.

# Validation performed

Pages switch

## Summary by CodeRabbit

- **New Features**
  - Introduced two new pages: an Ingest Page and a Search Page, providing dedicated interfaces for content management and querying archives.
  - Upgraded the routing system to support dynamic page rendering and smooth, interactive navigation.

- **Style**
  - Standardised styling across navigation elements to ensure a consistent user experience.


components/log-viewer-webui/client/src/ui/MainLayout.tsx
```diff
@@ -31,7 +35,7 @@ const MainLayout = () => {
     const [collapsed, setCollapsed] = useState(false);
 
     return (
-        <Layout className={"main-layout"}>
+        <Layout className={styles["main-layout"]}>
```
junhaoliao, line 38: Let's use camelCasing for class names in CSS modules, following the recommendation at https://github.com/css-modules/css-modules/blob/master/docs/naming.md

Then we can use the dot-notation here

components/log-viewer-webui/client/src/ui/IngestView.tsx
```diff
@@ -0,0 +1,18 @@
...
+ */
+const IngestView = () => {
+    return (
+        <div>
+            <h1>Ingest View</h1>
+            <p>This is the Ingest view.</p>
+        </div>
+    );
+};
+
```
junhaoliao, line 17:
```suggestion


```

components/log-viewer-webui/client/src/routes.tsx
```diff
@@ -0,0 +1,19 @@
...
+    {
+        path: "/",
+        Component: MainLayout,
+        children: [
+            {path: "ingest", Component: IngestView},
+            {path: "search", Component: SearchView},
+        ],
+    },
+]);
+
```
junhaoliao, line 18:
```suggestion


```

components/log-viewer-webui/client/src/routes.tsx
junhaoliao, line 1: How do you feel about organzing the routes as this instead?
```
- src/
  - routes/
    - index.tsx
    - IngestView.tsx
    - ...
```

or
```
- src/
  - routes.tsx
  - page/
    - IngestView.tsx
    - ...
```

Naming the UI components directory as `src/import/ui` in the existing webui code was to follow Meteor conventions. Since `log-viewer-webui/client` is for UI only, let's avoid the name "UI" in namings.

components/log-viewer-webui/client/src/routes.tsx
```diff
@@ -0,0 +1,19 @@
+import {createBrowserRouter} from "react-router";
+
+import IngestView from "./ui/IngestView";
+import MainLayout from "./ui/MainLayout";
+import SearchView from "./ui/SearchView";
+
+
+const router = createBrowserRouter([
```
junhaoliao, line 8:
```suggestion
const ROUTER = createBrowserRouter([
```

components/log-viewer-webui/client/src/ui/SearchView.tsx
```diff
@@ -0,0 +1,18 @@
...
+ */
+const SearchView = () => {
+    return (
+        <div>
+            <h1>Search View</h1>
+            <p>This is the Search view.</p>
+        </div>
+    );
+};
+
```
junhaoliao, line 17:
```suggestion


```

junhaoliao (changes requested):

components/log-viewer-webui/client/src/ui/MainLayout.tsx
```diff
@@ -10,16 +14,16 @@ import {
...
 
-import "./MainLayout.css";
+import styles from "./MainLayout.module.css";
 
 
 const {Sider} = Layout;
 
 type MenuItem = Required<MenuProps>["items"][number];
 
 const items: MenuItem[] = [
```
junhaoliao, line 24: How about SIDEBAR_MENU_ITEMS?

components/log-viewer-webui/client/src/routes.tsx
junhaoliao: How about
```
- src/
  - components/
    - For any sharable components...
  - Layout/
    - index.tsx
    - routes.tsx
      - page/
        - IngestView.tsx
      - ...
```

components/log-viewer-webui/client/src/routes.tsx
junhaoliao: we've been using SNAKE_CASE for global constants but i do get variables are mostly `const` since EM6. I prefer keeping SNAKE_CASES, but let's quickly check with @kirkrodrigues before we commit.

components/log-viewer-webui/client/src/ui/MainLayout.tsx
junhaoliao: Let me quickly check the linter on my end.

junhaoliao (changes requested):

junhaoliao:
@davemarco 
Please also merge `main` into this branch. I added a fix on a build issue due to TS check errors this morning

components/log-viewer-webui/client/src/routes.tsx
junhaoliao: I discussed this offline with @kirkrodrigues and here're some consolidated thoughts (a copy of section https://www.notion.so/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?pvs=4#1cf04e4d9e6b80c7ac4fe69b846f5791)

## Naming convention for constants in JavaScript / TypeScript

JavaScript doesn't enforce deep immutability for `const` objects or arrays, so we apply a semantic naming convention to indicate intent:

- **SCREAMING_SNAKE_CASE**
    
    Use this for *truly immutable* values. This includes:
    
    - Primitive constant values
    e.g., `const SECONDS_IN_MINUTE = 60;`
    - Deeply frozen objects
    e.g., `const CONFIG = Object.freeze({...});`
    - Any value that is intended to remain constant throughout its lifetime.
- **camelCase**
    
    Use this for variables declared with `const` whose contents may still change (e.g., properties of objects or elements of arrays). This keeps the code readable and avoids excessive SCREAMING_SNAKE_CASE.
    

### Additional Notes

- **React `useState` variables** should *not* use SCREAMING_SNAKE_CASE, even though the bindings are immutable. Their values can change via re-renders, so they are not considered truly constant.
- **Constant functions** should follow normal **function naming rules**, based on their role:
    - Use **PascalCase** for **JSX functional components** (e.g., `const MyComponent = () => { ... }`), as required by React.
    - Use **camelCase** for all other functions (e.g., `const formatDate = () => { ... }`).
    
    Even though function bindings with `const` are immutable, the function itself is a callable entity, not a constant value, so we do not use SCREAMING_SNAKE_CASE.

junhaoliao: That said, let's make this object deeply frozen since the routes are not expected to be modified elsewhere. We will then rename the object and the filename to `ROUTES` as well.

components/log-viewer-webui/client/src/ui/MainLayout.tsx
junhaoliao: Let's make this deeply frozen with `Object.freeze()`.

junhaoliao (changes requested):

components/log-viewer-webui/client/src/ui/MainLayout.tsx
junhaoliao: > https://github.com/mrmckeb/typescript-plugin-css-modules

this plugin is only a Language Service Plugin, which means it only have effects in IDEs (than `tsc -b`).

---

How about this:
1. We always name the classes in camelCase in CSS modules.
2. If we want to use dot-notation access, we shall disable `noPropertyAccessFromIndexSignature` in `tsconfig.app.json`.
3. Alternatively, we can update `eslint.config.mjs` to add below block:
```js
{
     files: [
         "**/*.ts",
         "**/*.tsx",
     ],
     rules: {
         "dot-notation": "off",
         "@typescript-eslint/dot-notation": "error",
     },
 },

```
I will also be adding the above to `eslint-config-yscope` so the config lib users do not have to config those explicitly.

Let me know what you think.

components/log-viewer-webui/client/src/routes.tsx
junhaoliao: sure. i just thought of another reason that we don't want to make it deeply frozen: if we are to support any plugins like an optional page. this is fine to leave then

components/log-viewer-webui/client/src/routes.tsx
junhaoliao, line 1: 1. don't forget updating the page component names inside the files.
2. since this `routes.tsx` default exports `router`, let's name this file `router.tsx` to match the default export name then.

components/log-viewer-webui/client/eslint.config.mjs
```diff
@@ -28,6 +28,7 @@ const EslintConfig = [
     ...ReactConfigArray,
     {
         rules: {
+            "dot-notation": "off",
```
junhaoliao, line 31: Let's turn it off for TS only. like instead, we want to add this object to the end of the `EslintConfig` array:
```
{
     'files: [
         "**/*.ts",
         "**/*.tsx",
     ],
     rules: {
         "dot-notation": "off",
         "@typescript-eslint/dot-notation": "error",
     },
},
```

junhaoliao (changes requested):
good with a few nits

junhaoliao:
i edited the PR title directly to change the scope "webui" -> "new-webui"

junhaoliao (approved):

PR #809, opened 2025-04-09
feat(new-webui): MongoCDC Framework

AVMatthews:
# Description

New framework built on MongoCDC using socket.io meant to replace our current usage of meteor's publish-subscribe framework. 

TODO:
- Linting 
- Unit Tests
- Before merge return App.tsx back to original App.tsx (just changed for testing)

# Checklist

  breaking change.

# Validation performed

Manually added collection to mongo replica and tested to see that the contest of the collection could be requested through the client from the server.


components/log-viewer-webui/client/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "tsc -b && vite build",
     "lint:check": "eslint",
     "lint:fix": "eslint --fix",
-    "start": "vite",
+    "start": "vite --host 10.1.0.21",
```
AVMatthews, line 10: @junhaoliao and I added this --host while debugging. We obviously need to remove this, but I'm going to have to test it without it.

components/log-viewer-webui/client/src/App.tsx
junhaoliao, line 1: NOTE: changes in this file is to be removed.

components/log-viewer-webui/client/package.json
junhaoliao: NOTE: to be removed.

components/log-viewer-webui/client/src/mongoCDCLib/MongoReplicaCollectionCursor.ts
```diff
@@ -0,0 +1,67 @@
...
+            }, (response: { error?: any; data?: any }) => {
+                if (response.error) {
+                    return reject(response.error);
+                }
+
+                return resolve(response.data);
+            });
+        });
+    }
+}
```
junhaoliao, line 64: NOTE: not in use at the moment.

components/log-viewer-webui/client/src/mongoCDCLib/MongoReplicaCollectionReactiveCursor.ts
```diff
@@ -0,0 +1,66 @@
...
+interface ReactiveArrayCallback {
+    onData: (data: any) => void;
+    onError: (error: any) => void;
+}
+
+class MongoReplicaCollectionReactiveCursor extends MongoReplicaCollectionCursor {
+    /**
+     * @type {Listener}
+     */
+    private listener: Listener | null = null;
```
junhaoliao, line 18: For private properties / functions, can we use the `#propertyName` syntax?

components/log-viewer-webui/client/vite.config.ts
```diff
@@ -14,6 +14,11 @@ export default defineConfig({
                 target: "http://localhost:3000/",
```
junhaoliao, line 14: Since the comment in L12-13 applies to both routes:
1. Let move the comment to Line 11
2. Let say "target host and port" instead of "target" to improve clarity.

@coderabbitai Let's create an issue to track the effort of serving Fastify routes under "/api". And we should replace this "/query/" proxy entry with "/api/" to ensure the other RESTful APIs are proxied to the backend.

components/log-viewer-webui/server/package.json
```diff
@@ -30,6 +30,7 @@
     "fastify-plugin": "^5.0.1",
     "http-status-codes": "^2.3.0",
     "pino-pretty": "^13.0.0",
+    "socket.io": "^4.0.0",
```
junhaoliao, line 33: Let's use the latest
```suggestion
    "socket.io": "^4.8.1",
```

components/log-viewer-webui/client/package.json
```diff
@@ -21,13 +21,20 @@
...
+    "@eslint/js": "^9.19.0",
     "@types/react": "^19.0.10",
     "@types/react-dom": "^19.0.4",
     "@vitejs/plugin-react": "^4.3.4",
+    "eslint": "^9.19.0",
     "eslint-config-yscope": "latest",
+    "eslint-plugin-react": "^7.37.4",
+    "eslint-plugin-react-hooks": "^5.0.0",
+    "eslint-plugin-react-refresh": "^0.4.18",
+    "globals": "^15.14.0",
```
junhaoliao, line 37: We don't have to explicitly state those eslint dependencies because `eslint-config-yscope` lists those as "peer" dependencies, so `npm install` will install those even if not explicitly specified.
```suggestion
    "@types/react": "^19.0.10",
    "@types/react-dom": "^19.0.4",
    "@vitejs/plugin-react": "^4.3.4",
    "eslint-config-yscope": "latest"
```

components/log-viewer-webui/client/src/mongoCDCLib/useTracker.tsx
```diff
@@ -0,0 +1,40 @@
...
+            onData: handleData,
+            onError: (error: any) => {
+                console.error("Error fetching data:", error);
+            },
+        });
+
+        return () => {
+            unsubscribe();
+        };
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
junhaoliao, line 34:
```suggestion
        // `getCursor` is a function that's intentionally excluded from the dependencies array. Including it would
        // cause unnecessary re-subscriptions or even infinite loops if it's redefined on each render.
        // eslint-disable-next-line react-hooks/exhaustive-deps
```

components/log-viewer-webui/server/src/plugins/MongoReplicaServerPlugin.ts
```diff
@@ -0,0 +1,272 @@
...
+        try {
+            await mongoClient.connect();
+
+            return mongoClient.db(dbName);
+        } catch (e) {
+            throw new Error("MongoDB connection error", {cause: e});
+        }
+    }
+
+    /* eslint-disable max-lines-per-function */
```
junhaoliao, line 148: I understand this was ported from the prototype. Let's remove this and try to deduplicate the `socket` event registrations by iterating through a list of objects like
```
{eventName: "disconnect", listener: this.#socketDisconnectListener, args: null, callback: null},
...
```

components/log-viewer-webui/server/src/plugins/MongoReplicaServerPlugin.ts
```diff
@@ -0,0 +1,272 @@
...
+/**
+ * MongoDB replica set plugin for Fastify.
+ *
+ * @param app
+ * @param options
+ * @param options.mongoUri
+ * @param options.dbName
+ */
+// eslint-disable-next-line @stylistic/max-len
+const MongoReplicaServerPlugin = async (app: FastifyInstance, options: {mongoUri: string; dbName: string}): Promise<void> => {
```
junhaoliao, line 258: Return types of `Promise<void>` can be ommited. Since the promise resolves to `void` and the function is `async` (a syntactic sugar for wrapping some return value in a Promise).
```suggestion
const MongoReplicaServerPlugin = async (app: FastifyInstance, options: {mongoUri: string; dbName: string}) => {
```

components/log-viewer-webui/server/src/plugins/MongoReplicaServerPlugin.ts
```diff
@@ -0,0 +1,272 @@
...
+    app.decorate(
+        "MongoReplicaServer",
+        MongoReplicaServer.create({
+            fastify: app,
+            mongoUri: options.mongoUri,
+            dbName: options.dbName,
+        })
+    );
+};
+
```
junhaoliao, line 271:
```suggestion


```

components/log-viewer-webui/client/src/mongoCDCLib/MongoReplicaCollection.ts
```diff
@@ -0,0 +1,41 @@
...
+//     const collection = new MongoReplicaCollection("3");
+//     const cursor = collection.find({}, {});
+
+//     console.log(await cursor.toArray());
+// };
+
+//
+// await main();
+
+export default MongoReplicaCollection;
```
junhaoliao, line 41: we tend to reserve two empty lines before exports. Sorry about that the linter has not been smart enough to autofix this.
```suggestion


export default MongoReplicaCollection;
```

components/log-viewer-webui/client/src/mongoCDCLib/MongoReplicaCollectionReactiveCursor.ts
```diff
@@ -0,0 +1,66 @@
...
+            this.socket.emit("collection::find::unsubscribe", {
+                queryHash: queryHash,
+            });
+
+            this.socket.off("collection::find::update");
+        };
+    }
+}
+
+export default MongoReplicaCollectionReactiveCursor;
```
junhaoliao, line 66:
```suggestion


export default MongoReplicaCollectionReactiveCursor;
```

components/log-viewer-webui/client/src/mongoCDCLib/useTracker.tsx
```diff
@@ -0,0 +1,40 @@
...
+    }, dependencies);
+
+    return data;
+};
+
+export {useTracker};
```
junhaoliao, line 40: Let's do default export here:
```suggestion

export default useTracker;
```

We also need to update the import in the sample App.tsx:
```
import useTracker from ...;
```

components/log-viewer-webui/server/src/app.ts
```diff
@@ -56,6 +57,11 @@ const app = async ({
         await server.register(S3Manager, {region: settings.StreamFilesS3Region});
     }
 
+    await server.register(MongoReplicaServerPlugin, {
+        mongoUri: "mongodb://localhost:27017",
+        dbName: "clp-query-results",
```
junhaoliao, line 62: 1. For consistency with the Fastify-mongodb plugin, let's rename `dbName` as `database.`; similarly, let's break the `mongoUri` property into `host` and `name`.
2. We don't want to hardcoded the values here. They should come from `settings.xxx`. Note we need to add another entry in the settings.json file like `MongoDbQueryResultsCollectionName`.

components/log-viewer-webui/server/src/plugins/MongoReplicaServerPlugin.ts
```diff
@@ -0,0 +1,272 @@
+/* eslint-disable max-classes-per-file */
```
junhaoliao, line 1: Let's remove this and move the classes into separate files then import as needed

components/log-viewer-webui/server/src/plugins/MongoReplicaServerPlugin.ts
```diff
@@ -0,0 +1,272 @@
...
+            .command({replSetInitiate: {}});
+
+        fastify.log.info("Replica set initialized:", response);
+    } catch (e) {
+        if (e instanceof MongoServerError && "AlreadyInitialized" === e.codeName) {
+            return;
+        }
+        throw new Error("Failed to initialize replica set", {cause: e});
+    }
+};
```
junhaoliao, line 33: We can remove all references to this, since the clp-package should have done the init in some Python script when starting the package.
```suggestion
/**
 * Initialize the MongoDB replica set.
 *
 * @param {FastifyInstance} fastify
 * @return {Promise<void>}
 * @throws {Error} If the replica set initialization fails.
 */
const initializeReplicaSet = async (fastify: FastifyInstance): Promise<void> => {
    try {
        const directMongoClient = new MongoClient(
            "mongodb://localhost:27017",
            {replicaSet: "rs0", directConnection: true}
        );
        const response = await directMongoClient.db("admin").admin()
            .command({replSetInitiate: {}});

        fastify.log.info("Replica set initialized:", response);
    } catch (e) {
        if (e instanceof MongoServerError && "AlreadyInitialized" === e.codeName) {
            return;
        }
        throw new Error("Failed to initialize replica set", {cause: e});
    }
};
```

components/log-viewer-webui/server/src/plugins/MongoReplicaServerPlugin.ts
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * MongoDB replica set plugin for Fastify.
+ *
+ * @param app
+ * @param options
+ * @param options.mongoUri
+ * @param options.dbName
+ */
+// eslint-disable-next-line @stylistic/max-len
```
junhaoliao, line 257:
```suggestion
```

components/log-viewer-webui/server/src/plugins/MongoReplicaServerPlugin.ts
```diff
@@ -0,0 +1,272 @@
...
+            );
+        });
+    }
+}
+
+/**
+ * @typedef {object} MongoReplicaServerPluginOptions
+ * @property {string} mongoUri MongoDB URI
+ * @property {string} dbName Database name
+ */
```
junhaoliao, line 247: The property types are o longer needed in TypeScript. The descriptions also do not need to be ported since they just repeat the names.
```suggestion
```

components/log-viewer-webui/server/src/plugins/MongoReplicaServerPlugin.ts
```diff
@@ -0,0 +1,272 @@
...
+ * MongoDB replica set plugin for Fastify.
+ *
+ * @param app
+ * @param options
+ * @param options.mongoUri
+ * @param options.dbName
+ */
+// eslint-disable-next-line @stylistic/max-len
+const MongoReplicaServerPlugin = async (app: FastifyInstance, options: {mongoUri: string; dbName: string}): Promise<void> => {
+    console.log("registering")
```
junhaoliao, line 259: Let's remove this debug print
```suggestion
```

junhaoliao (changes requested):
nice work

i understand this PR is still WIP. Let's use the suggestions as a checklist before requesting for a formal review

junhaoliao:
@AVMatthews / @davemarco / @hoophalab Please voice any preferences if you feel the `mongoCdcLib` / `MongoReplicaCollection`, etc. can be renamed. I got some ideas but would like to hear your thoughts first.

junhaoliao:
> I looked briefly - does it make sense to keep this open as a WIP. Then have two smaller PRs for server and client. At almost ~1000, it may sense to split
> 
> 
> 
> Also for syntax, we normally only Uppercase first letter in code. so MongoCdc

Getting the stuff cleaned in this PR first then submitting separately is a good idea. 

By the way, I tend to avoid the naming anything with "CDC" at this point because we're not currently handling (and for performance reasons even in the future we will not be always handling) the delta changes in the change packets.

davemarco:
replaced by #880 and #892

PR #814, opened 2025-04-10
feat(new-webui): Mongo Framework Server

AVMatthews:
# Description

Server for new framework built on MongoCDC using socket.io meant to replace our current usage of meteor's publish-subscribe framework.

First PR to split #809 

TODO: A few listing errors left to solve

# Checklist

  breaking change.

# Validation performed

Manually added collection to mongo replica and tested to see that the contest of the collection could be requested through the client from the server.

## Summary by CodeRabbit

- **New Features**
  - Enabled real-time reactive MongoDB query subscriptions with live updates via WebSocket.
  - Added user interface support for subscribing and unsubscribing to MongoDB queries.
  - Integrated WebSocket communication for efficient, event-driven backend data updates.

- **Chores**
  - Added new dependencies to support WebSocket and real-time functionality.


davemarco:
I am still looking through but just for linting to start
- we should have comments on top of the new classes roughly explaining what they do
- Also comments for the internal functions

For some of the linting failures, It looks like they are related to missing types for socket io. They have some docs here that may help for adding missing types - https://socket.io/docs/v4/typescript/

junhaoliao (changes requested):
as discussed offline, let's remove all references to the name "replica". e.g., MongoReplicaServer -> FastifyMongoServer (i added "Fastify" in my proposal here because the name "MongoServer" alone can be ambiguous)

components/log-viewer-webui/server/src/plugins/MongoReplicaServer/index.ts
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /**
+     * Creates a listener for initializing a collection.
+     * Adds a reference to the collection when requested by the client.
+     *
+     * @param socket The socket instance that is requesting the collection.
+     * @return A function that handles the collection initialization event.
+     */
+    #getCollectionInitListener (socket: CustomSocket): ClientToServerEvents["collection::init"] {
```
davemarco, line 198: reminder - If functionality is changed to allow support for multiple collections, then this will need to be updated.

components/log-viewer-webui/server/src/plugins/MongoReplicaServer/index.ts
```diff
@@ -0,0 +1,296 @@
...
+            if ("undefined" === typeof collection) {
+                callback({error: "Collection not initialized"});
+
+                return;
+            }
+
+            const {queryHash, watcher} = collection.getWatcher(query, options, socket.id);
+            callback({queryHash});
+            // eslint-disable-next-line @typescript-eslint/no-misused-promises
+            watcher.changeStream.on("change", async () => {
```
davemarco, line 241: I looked online and saw an example where they get a`next` event from the change stream, and emit that, instead of doing another find query. Would that be possible here

Here is example i saw 
```
       // open a Change Stream on the "messages" collection
       changeStream = messages.watch();

       // set up a listener when change events are emitted
       changeStream.on("change", next => {
           // process any change event
           switch (next.operationType) {
               case 'insert':
                   console.log(next.fullDocument.message);
                   break;
               case 'update':
                   console.log(next.updateDescription.updatedFields.message);
           }
       });
```

components/log-viewer-webui/server/src/plugins/MongoReplicaServer/MongoReplicaServerCollection.ts
```diff
@@ -0,0 +1,141 @@
...
+ */
+class MongoReplicaServerCollection {
+    // Reference count for active subscriptions
+    private count: number;
+
+    // MongoDB collection instance
+    private collection: Collection;
+
+    // Map of active change streams keyed by query hash
+    private watchers: Map<string, Watcher>;
```
davemarco, line 42: Again i think we can move this out of constructor
```suggestion
    private watchers: Map<string, Watcher> = new Map();
```

components/log-viewer-webui/server/src/plugins/MongoReplicaServer/MongoReplicaServerCollection.ts
```diff
@@ -0,0 +1,141 @@
...
+    constructor (mongoDb: Db, collectionName: string) {
+        this.count = 0;
+        this.collection = mongoDb.collection(collectionName);
+        this.watchers = new Map();
+    }
+
+    /**
+     * Increment the reference count;
+     */
+    refAdd () {
```
davemarco, line 59: Maybe these reference counters should be replaced with array of socket ids

davemarco (changes requested):
I did my first pass of the code. 

Two larger comments.
1. It looks like each client can only subscribe to a single collection at once. Is that intended? It may make sense to add support to subscribe to multiple collections. Let me know
2. Right now it looks like the client needs to disconnect to unsubscribe from a collection. Maybe it makes senses to add a "collection::unsubscribe" command, so the client can remove the collection itself?. 

I will more at style in the next review. 

Edit per discussion - The client is set up to open multiple sockets. As a result, it can support multiple collections. We can look into moving to a single socket at a later date.

davemarco:
Also google a bit, and I think this is another option for the client
https://dev.to/bilelsalemdev/understanding-the-singleton-pattern-in-typescript-4kep#:~:text=Singleton%20Pattern%20with%20Socket.IO,ensure%20consistent%20and%20efficient%20communication.

components/log-viewer-webui/server/src/plugins/MongoReplicaServer/MongoReplicaServerCollection.ts
davemarco: ignore based on earlier discussion

junhaoliao:
> Here is an option for the client side using only one socket, but maintaining something close to the current interface. Not neccesary for now, but potential later optimization.
> 
> (1a) shared socket The main benefit is that the browser limits websockets to domain to 255 per this - https://news.ycombinator.com/item?id=30314281
> 
> https://dev.to/bilelsalemdev/understanding-the-singleton-pattern-in-typescript-4kep#:~:text=Singleton%20Pattern%20with%20Socket.IO,ensure%20consistent%20and%20efficient%20communication.
> 
> We could have a new function on the client
> 
> ```
> import { io, Socket } from "socket.io-client";
> 
> let sharedSocket: Socket | null = null;
> 
> export function getSharedSocket(): Socket {
>     if (!sharedSocket) {
>         // You can pass a URL here if needed (e.g., from environment vars)
>         sharedSocket = io();
>     }
>     return sharedSocket;
> }
> ```
> 
> then when creating new collection on the client, it uses the shared socket across all components
> 
> ```
> class MongoReplicaCollection {
>     private socket: Socket;
> 
>     constructor(collectionName: string) {
>         this.socket = getSharedSocket();
> 
>         this.socket.emit("collection::init", {
>             collectionName: collectionName,
>         });
>     }
> ```
> 
> Then for the cursor for the component could listen on something like this this.socket.on("collection::find::update${hash})",
> 
> where the hash is the collection, query, options.
> 
> 1a) Even if we dont do shared socket, it may still sense to still use a different event name for reactive arrays from the same client from the same collection. Something like this.socket.on("collection::find::update${hash})", where the hash is the query, option. It looks like if the client has multiple queries to the same collection, the notification will collide on the same event since they have the same name? Another option (1c) is maybe to force only one cursor per collection
> 
> (2) Rooms. Also another potential improvement is to consider rooms each query hash. Then we don't need to rerun the query for every client that subscribes.

(1a) we should do shared socket on the client side given it's been fairly clear how it can be achived
(1b) assigning different event names might cause challenge working with TS. instead, I agree we can include the query hash / identifier in the server->client update payload. it's likely required if we're to share the same socket on the client side; and even if not, it will be good for some sanity check
(2) i agree we should use rooms now, given how we will be able to share the same watcher across requests associated with an identical query. like in our latest discussion, we proposed that we want to wrap the watcher instance with a list of subscribers and only free the watcher if there is no subscriber left.

components/log-viewer-webui/server/src/plugins/MongoReplicaServer/index.ts
```diff
@@ -0,0 +1,296 @@
...
+import {
+    Server,
+    Socket,
+} from "socket.io";
+
+import MongoReplicaServerCollection from "./MongoReplicaServerCollection.js";
+
+
+// Define the events that the client can emit to the server
+type ClientToServerEvents = {
```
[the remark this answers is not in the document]
junhaoliao:
```
export type Response<T> = Error | Success<T>;
```
sure, i agree we need something like this

components/log-viewer-webui/server/src/plugins/MongoReplicaServer/index.ts
```diff
@@ -0,0 +1,296 @@
...
+            const collection = this.#collections.get(collectionName);
+
+            if ("undefined" === typeof collection) {
+                callback({error: "Collection not initialized"});
+
+                return;
+            }
+
+            const {queryHash, watcher} = collection.getWatcher(query, options, socket.id);
+            callback({queryHash});
```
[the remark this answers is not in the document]
junhaoliao: This "queryHash" is just an identifier for use in `client->server: unsubscribe` events later.

I believe a better approach is to maintain a object for `queryId` -> `query` mappings per collection, like:
```
{
  0: queryHash0,
  1: queryHash1,
}
```
and return the key (`queryId`) instead. when there's a new query, we will:
1. generate queryHash and try to match the values in the mappings object.
2. if found, return the key.
3. if not found, insert a new entry with key = `MAX(the object's keys) + 1`

components/log-viewer-webui/server/src/plugins/MongoReplicaServer/index.ts
AVMatthews: We decided against this for right now, but this might be useful to add in the future.

components/log-viewer-webui/server/src/plugins/MongoReplicaServer/index.ts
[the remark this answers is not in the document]
AVMatthews: I tried removing the arguments from the object and I would only get undefined as a value. I'm happy to discuss further.

components/log-viewer-webui/server/src/plugins/FastifyMongoServer/index.ts
```diff
@@ -0,0 +1,444 @@
...
+        data: Document[];
+    }) => void;
+    "error::query": (respArgs: {
+        queryId: number;
+        message: string;
+    }) => void;
+    "error::collection": (respArgs: {
+        collectionName: string;
+        message: string;
+    }) => void;
```
davemarco, line 70: Is it possible to just check for an error in the callback function on the client side? Like do we need these new messages?

davemarco (changes requested):
I added some more comments. I am still looking through, and will add more later.

davemarco:
I added a stylistic refactor here. and some other minor logic changes - https://github.com/y-scope/clp/pull/865

components/log-viewer-webui/server/src/plugins/MongoSocketIoServer/MongoWatcherCollection.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    /**
+     * Gets the watcher for a specific query. If the watcher does not exist, it creates a new one.
+     *
+     * @param queryParameters
+     * @param queryId
+     * @param socket
+     */
+    async getWatcher (
```
junhaoliao, line 54: This is an unusual design choice in the interface. The `getWatcher` method is mixing two very different responsibilities:

* Creating/retrieving a watcher (resource management)
* Subscribing a specific socket and sending initial data (communication)

This violates the single responsibility principle and creates an awkward interface that tightly couples distinct operations.

Can we implement those interfaces instead:
```
interface WatcherOptions<T> {
  onUpdate: (data: T[]) => void;
  onError: (error: Error) => void;
  debounceMillis: number;
}

class MongoWatcherCollection {
    constructor(mongoDb: Db, collectionName: string) {}

    async createOrGetWatcher<T>(queryId: QueryId, options: WatcherOptions<T>) {
        ...
        return watcher;
    }

    addSubscriber(queryId: QueryId, socketId: string) {}

    removeSubscriber (queryId: number, socketId: string): boolean {}

    // add this, or maybe we can simply extend MongoWatcherCollection from mongo.Collection
    find(...) {
        this.#collection.find(...);
    }
}
```

junhaoliao (changes requested):

junhaoliao:
It seems there're some conflicts with `main` as well. Please attempt merging from main and resolve any conflicts.

components/log-viewer-webui/server/src/plugins/MongoSocketIoServer/MongoWatcherCollection.ts
davemarco: I took inspiration from this in the PR. My implementation is not exactly the same. But uses the same idea

davemarco:
replaced by #880

PR #817, opened 2025-04-11
refactor(clp-s): Split some of `Output`'s responsibilities into a new `QueryRunner` class.

wraymo:
# Description

This PR refactors the clp-s `Output` class by extracting query execution logic into a dedicated `QueryRunner` class, enhancing modularity and reusability. Specifically:

* The `Output` class retains its `filter` functionality, responsible for query initialization and output message handling.
* A new, reusable `QueryRunner` class is introduced, encapsulating the query execution logic.
* To facilitate `QueryRunner` usage, certain private classes within the original `Output` have been made protected or public.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

- **New Features**
  - Introduced a new query execution engine for structured data searches, enabling more efficient and flexible query evaluation.
- **Refactor**
  - Simplified and streamlined the search output and filtering logic, delegating complex evaluation to the new query engine.
  - Updated interfaces and method signatures for improved clarity and maintainability.
- **Bug Fixes**
  - Improved memory management and object ownership for certain classes.
- **Chores**
  - Updated linting configuration to accommodate new source files.
- **Documentation**
  - Added a default virtual destructor to ensure proper cleanup in transformation classes.


kirkrodrigues:
It seems like there are a lot of clang-tidy violations in QueryRunner.cpp besides the misc-include-cleaner one we discussed offline. I'm assuming a lot of this PR is about moving code, so I'm fine with deferring these fixes to another PR if it makes @gibber9809's review easier. However, if the code in QueryRunner.cpp is mostly new, we should try to fix the clang-tidy violations in this PR rather than deferring it.

(To see the additional violations, you can add a `// NOLINTBEGIN(misc-include-cleaner)` at the beginning of the file and a `// NOLINTEND(misc-include-cleaner)` at the end of the file.)

components/core/src/clp_s/search/Output.hpp
```diff
@@ -18,372 +18,42 @@
...
 #include "OutputHandler.hpp"
+#include "QueryRunner.hpp"
 #include "SchemaMatch.hpp"
 
 using namespace simdjson;
 using namespace clp_s::search::clp_search;
 
 namespace clp_s::search {
-class Output : public FilterClass {
+class Output {
```
gibber9809, line 28: Should probably add a docstring for this class.

components/core/src/clp_s/search/Output.hpp
```diff
@@ -18,372 +18,42 @@
...
+           std::shared_ptr<ast::Expression> const& expr,
+           std::shared_ptr<ArchiveReader> const& archive_reader,
            std::unique_ptr<OutputHandler> output_handler,
            bool ignore_case)
-            : m_archive_reader(std::move(archive_reader)),
-              m_schema_tree(m_archive_reader->get_schema_tree()),
-              m_schemas(m_archive_reader->get_schema_map()),
-              m_match(match),
+            : m_query_runner(match, expr, archive_reader, ignore_case),
+              m_archive_reader(std::move(archive_reader)),
```
gibber9809, line 36:
```suggestion
              m_archive_reader(archive_reader),
```
I think it might be better to not `std::move` `archive_reader` since we also pass it to `m_query_runner`, so reordering could lead to a null pointer dereference.

wraymo:
I’ve addressed nearly all the warnings in `QueryRunner.hpp`. As for `QueryRunner.cpp`, there are still quite a few remaining. Some of them — like those related to cognitive complexity — aren’t straightforward to fix and could introduce breaking changes, which feels out of scope for this PR.

Since most of the methods in `QueryRunner` are from `Output`, we can tackle the remaining warnings in both classes together later. Beyond clang-tidy, some of those methods will likely need a bit of refactoring as well which will be handled in the future PR.

components/core/src/clp_s/search/SchemaMatch.hpp
```diff
@@ -19,6 +19,8 @@ class SchemaMatch : public ast::Transformation {
     // Constructor
     SchemaMatch(std::shared_ptr<SchemaTree> tree, std::shared_ptr<ReaderUtils::SchemaMap> schemas);
 
+    virtual ~SchemaMatch() = default;
```
gibber9809, line 22: I think the real problem is that `Transformation` doesn't have a virtual destructor right? I think we can get rid of this and just add a virtual destructor in `Transformation.hpp`.

components/core/src/clp_s/search/Output.cpp
```diff
@@ -62,62 +60,62 @@ bool Output::filter() {
...
         } else {
-            m_array_dict = m_archive_reader->read_array_dictionary(true);
+            m_archive_reader->read_array_dictionary(true);
         }
     }
 
-    populate_internal_columns();
-    populate_string_queries(top_level_expr);
+    m_query_runner.populate_internal_columns();
+    m_query_runner.populate_string_queries();
```
gibber9809, line 80: Since it probably isn't clear to the average reader what these two functions are doing do you think it makes sense to wrap them in a function like `global_init()  -> void` or something?
i.e. replace these two calls with `m_query_runner.global_init()` and the semantic is that it does one-time initialization work that is common to all searched schemas?

components/core/src/clp_s/search/Output.cpp
```diff
@@ -62,62 +60,62 @@ bool Output::filter() {
...
 
-        if (m_expression_value == EvaluatedValue::False) {
+        if (auto expression_value = m_query_runner.constant_propagate();
+            EvaluatedValue::False == expression_value)
+        {
             continue;
         }
 
-        add_wildcard_columns_to_searched_columns();
+        m_query_runner.add_wildcard_columns_to_searched_columns();
```
gibber9809, line 96: Similar to my other comment: do you think it makes sense to wrap `setup_schema` through this function call in a function like `schema_init -> EvaluatedValue`?

i.e. replace all of this with `if (EvaluatedValue::False == m_query_runner.schema_init(schema_id))...` with the semantic that it does all schema specific initialization (and tries to figure out if the schema may have no search results).

components/core/src/clp_s/search/QueryRunner.hpp
```diff
@@ -0,0 +1,441 @@
...
+#include "ast/FilterOperation.hpp"
+#include "ast/Literal.hpp"
+#include "clp_search/Query.hpp"
+#include "SchemaMatch.hpp"
+
+using namespace simdjson;
+using namespace clp_s::search::clp_search;
+
+namespace clp_s::search {
+class QueryRunner : public FilterClass {
```
gibber9809, line 39: Could you add a docstring for QueryRunner?

gibber9809: Also, if we do end up wrapping most of these functions like I suggest in my other comments we should probably also move them into `private` scope. Besides that this class looks good to me.

gibber9809 (changes requested):
Nice work! The code should be much easier to re-use once this is merged.

Mostly have some suggestions for how we can wrap the APIs presented by `QueryRunner` to present a simpler interface/hide some of the implementation details and make them easier to change.

components/core/src/clp_s/search/SchemaMatch.hpp
wraymo: I believe I originally added this line while addressing warnings in `QueryRunner.hpp`. However, since we've now included that file in `EXCLUDE_PATTERNS`, it looks like the change is no longer necessary. Would you prefer that I remove the line for now and instead add the virtual destructor later, when we tackle the clang-tidy warnings for `Transformation` and its derived classes?

gibber9809: I'd say may as well add the virtual destructor to `Transformation` because it's a small change.

components/core/src/clp_s/search/QueryRunner.hpp
```diff
@@ -0,0 +1,438 @@
...
+    virtual ~QueryRunner() = default;
+
+    QueryRunner(QueryRunner const&) = delete;
+    auto operator=(QueryRunner const&) -> QueryRunner& = delete;
+
+    QueryRunner(QueryRunner&&) = delete;
+    auto operator=(QueryRunner&&) -> QueryRunner& = delete;
+
+    /**
+     * Initializes the query processing context for the first time.
```
gibber9809, line 76:
```suggestion
     * Initializes query processing context that is common to all schemas.
```

components/core/src/clp_s/search/QueryRunner.hpp
```diff
@@ -0,0 +1,438 @@
...
+     */
+    void global_init();
+
+    /**
+     * Prepares the query processing context for a given schema.
```
gibber9809, line 81:
```suggestion
     * Initializes query processing context for a given schema.
```
Just to be a bit more consistent with how we word the docstring for `global_init`.

gibber9809 (changes requested):
Minor comments on docstrings but otherwise looks good to me.

gibber9809 (dismissed):
LGTM. For PR title how about
```
refactor(clp-s): Separate `Output` responsibilities into `Output` and `QueryRunner` classes.
```

kirkrodrigues (commented):
Is the PR title correct? It sounds weird to separate a class' (Output's) responsibilities into itself, right?

kirkrodrigues:
Sorry, can you resolve the conflicts?

kirkrodrigues (approved):

Issue #737, opened 2025-02-27
feat(clp-s): Add support for generic ingestion-unit-level range index inside of archives


gibber9809:
### Request

When ingesting files into clp-s there can be several properties related to the file as a whole that can potentially be relevant to both decompression and search. These properties include the name of the file, but can also include other things known at ingestion time such as the service that created the log file and the host on which the logs were generated. Being able to accurately track these properties would be useful for things like decompressing or filtering on names of specific log files and implementing more powerful top level filtering. Our most immediate need is to gain the ability to track the relationship between files and data, but we may as well solve that problem in a way that allows us to build out other powerful features.

Currently we have an archive-level tagging system which doesn't quite meet these goals. The main issue is that these tags only track properties at the archive granularity, but archives can contain data from several different log files and log files may even be split across archives. As well, these tags are currently "unstructured" -- they're simply a list of strings associated with an archive (as opposed to a structured set of key value pairs).

Generally we'd like to introduce a more powerful mechanism that allows us to associate named properties with the units of data that we ingest so that they can be leveraged for decompression and search.

### Possible implementation

We can accomplish this goal by combining the following three ideas:
1) Allow users to specify properties for each log file passed to ingestion that they would like tracked
2) Maintain a range-index within the metadata section of each archive which tracks all of the properties requested by the user
3) Maintain some or all of the properties within an archive's range index in the global metadata db for top-level filtering

1.0 Specifying config

A user might pass us configuration like the following :
```
{
  "file_name": "/my/log/file.json",
  "properties": {
    "hostname": "host_1",
    "service": "service_1"
  }
}
```
Which tells us the file and some properties they want to track about it. The configuration wouldn't necessarily look like this, and would likely have more options, but this should serve to sketch the idea.

2.0 Tracking properties within an archive

Within an archive we simply need to map logical ranges of records to these properties, with some special handling for tracking relationships to an original file.

Specifically we would introduce a new metadata packet type `RangeIndex` with the following schema:
(range[X, Y)<size_t, size_t>, (tag_name<str>, tag_type, tag_value<tag_type>)+)+

Here each row of the range index maps a logical range of records [X, Y) to a list of named and typed properties. This allows us to track different properties for different subsections of an archive.

In the long run we can use this to help narrow search within an archive: if a user gives us a query targeting these metadata properties on top of their regular search query we should be able to consult the range index in order to narrow search to a logical range of records.

2.1 Tracking the mapping to an original file

[...]

PR #827, opened 2025-04-14
feat(clp-s)!: Reserve the `$`, `!`, and `#` namespaces for future use.

gibber9809:
# Description

This PR reserves the `$`, `!`, and `#` namespaces for future use. Specifically we update the parsing and unescaping rules for columns to treat columns beginning with `$`, `!`, and `#` as belonging to those respective namespaces unless escaped with a backslash.

Note that each change to the column parsing and unescaping rules is a breaking archive format change because it affects the interpretation of the authoritative timestamp column stored in the archive. Even though this is a breaking change we don't increment the archive version because if this PR gets merged before the next release we get to batch it together with a similar column format change for which we did increment the archive version.

We have near-term plans to use the `$` namespace to allow users to reference columns in the archive range index (see #737). The `!` and `#` namespaces have no immediate use, but they are reserved in this PR to avoid having to introduce more breaking changes of this type for the foreseeable future.

# Checklist

  breaking change.

# Validation performed

* Added tests for column format changes

## Summary by CodeRabbit

- **New Features**
  - Enhanced query parsing to support additional namespace characters ($, !, #), expanding the range of recognized symbols in search queries.
  - Improved unescaping functionality to handle more special characters in queries.
  - Added new namespace constants to support expanded query syntax.

- **Documentation**
  - Updated JSON search syntax guidelines to clarify that keys starting with special symbols (\, @, $, !, #) must now be escaped.

- **Tests**
  - Improved test cases to verify the handling of multiple namespace characters in search queries.

### DIFF
--- a/docs/src/user-guide/reference-json-search-syntax.md
+++ b/docs/src/user-guide/reference-json-search-syntax.md
@@ -195,7 +195,7 @@ Keys containing the following literal characters must escape the characters usin
 Furthermore, keys that _start_ with the following literal characters must escape the characters
 using a `\` (backslash):
 
-a `\` (backslash):
+* `\` (backslash):
 * `@`
 * `$`
 * `!`
--- a/docs/src/user-guide/reference-json-search-syntax.md
+++ b/docs/src/user-guide/reference-json-search-syntax.md
@@ -195,7 +195,6 @@ Keys containing the following literal characters must escape the characters usin
 Furthermore, keys that _start_ with the following literal characters must escape the characters
 using a `\` (backslash):
 
-* `\` (backslash):
 * `@`
 * `$`
 * `!`
--- a/docs/src/user-guide/reference-json-search-syntax.md
+++ b/docs/src/user-guide/reference-json-search-syntax.md
@@ -192,7 +192,9 @@ Keys containing the following literal characters must escape the characters usin
 * `.`
 * `*`
 
-Furthermore, keys _starting with_ the following literal characters must escape the characters using
+Furthermore, keys that _start_ with the following literal characters must escape the characters
+using a `\` (backslash):
+
 a `\` (backslash):
 * `@`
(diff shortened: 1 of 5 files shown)


kirkrodrigues (commented):
Docs review.

docs/src/user-guide/reference-json-search-syntax.md
```diff
@@ -191,7 +191,15 @@ Keys containing the following literal characters must escape the characters usin
 * `"`
 * `.`
 * `*`
-* `@` (only when `@` is the first character of the key)
+
+Furthermore, keys that _start_ with the following literal characters must escape the characters
+using a `\` (backslash):
+
+a `\` (backslash):
```
wraymo, line 198:
```suggestion
* `\` (backslash):
```

wraymo (commented):
I'm generally okay with the changes in this PR. I just have one question: do we have any guideline that restricts the use of char constants? If not, we could store those namespace characters as `char` constants. That way, we can directly use `constants::cAutogenNamespace` in the switch statement and replace the `@` and `!` literals with `constants::cAutogenNamespace` and `constants::cReservedNamespace1`, respectively.

docs/src/user-guide/reference-json-search-syntax.md
```diff
@@ -191,7 +191,15 @@ Keys containing the following literal characters must escape the characters usin
 * `"`
 * `.`
 * `*`
-* `@` (only when `@` is the first character of the key)
+
+Furthermore, keys that _start_ with the following literal characters must escape the characters
+using a `\` (backslash):
+
+* `\` (backslash):
```
gibber9809, line 198:
```suggestion
```

kirkrodrigues (approved):
Docs review.

wraymo (approved):

PR #834, opened 2025-04-17
docs: Update copyright year dynamically.

kirkrodrigues:
# Description

Adds code to set the end of the copyright time range to the current year. The reason we didn't do this previously is because technically if you're building and publishing the site, there's no guarantee that you've modified it and so the copyright year should remain the same as it was previously. However, it's unlikely that someone would build and publish the site without having modified it first, so for convenience, we add the code.

# Checklist

  breaking change.

# Validation performed

* `task docs:serve`
* Validated that the end of the copyright time range showed up as 2025.

## Summary by CodeRabbit

- **Chores**
  - Updated documentation to automatically display the current year in the copyright notice.

### DIFF
--- a/docs/conf/conf.py
+++ b/docs/conf/conf.py
@@ -1,11 +1,11 @@
-from datetime import datetime
+from datetime import date
 
 # -- Project information -------------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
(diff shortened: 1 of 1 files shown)


junhaoliao (approved):

junhaoliao:
For the title, how about:
```
docs: Update copyright year dynamically.
```

PR #847, opened 2025-04-24
feat(clp-s): Add the write-side implementation for the archive range index.

gibber9809:
# Description

This PR adds the write side of the archive range index as described in #737.

In the current implementation each split of a file has an associated range index entry that looks something like the following:
`{"start_idx": N, "end_index": M, "properties": {...}}`
This records properties associated with that split of a file corresponding to the logical range of the archive [N, M).

The range index will take in user-defined metadata from kv-ir streams and also always set the following properties for each range:
`_filename`: The full path to the file corresponding to the range
`_file_split_number`: Which split of a file corresponds to the range (0-indexed and incremented each time an archive is split while ingesting a file)
`_archive_creator_id`: UUID associated with a specific _invocation_ of compression -- allows disambiguating files with the same path compressed across different compression invocations

Most of the logic for writing and maintaining the range index is contained in `RangeIndexWriter.{cpp,hpp}`, this class is wrapped in `ArchiveWriter.{cpp,hpp}`, and properties are written in the main parsing loop(s) of `JsonParser.cpp`.

We bump the minor version of the archive format in this PR (new version is `0.3.1`) because this does constitute a change to the archive format, but the change is non-breaking (`0.3.0` archives that do not have the archive range index metadata packet are valid `0.3.1` archives).

Proper tests will be added in the read-side PR.

# Checklist

  breaking change.

# Validation performed

* Manually validated that range index metadata section can be deserialized and has expected contents both for common case and during archive splitting.

## Summary by CodeRabbit

- **New Features**
  - Introduced range-based metadata tracking for archives, including file split numbering and unique archive creator IDs.
  - Added support for associating custom metadata fields with archive ranges and enhanced propagation of user-defined metadata.
  - Enhanced archive format with a new range index structure for detailed metadata serialization.
  - Added new methods to manage archive ranges, including adding fields and closing ranges.

- **Bug Fixes**
  - Improved handling of unknown metadata packets to maintain decompressor state and archive integrity.

- **Chores**
  - Updated archive patch version and build configuration to include new range index components.


components/core/src/clp_s/RangeIndexWriter.hpp
```diff
@@ -0,0 +1,106 @@
...
+    // Types
+    using handle_t = size_t;
+
+    /**
+     * Opens a new range starting at start_index.
+     * @param start_index
+     * @return A handle to the newly opened range on success or the relevant error code on failure.
+     */
+    [[nodiscard]] auto open_range(size_t start_index) -> OUTCOME_V2_NAMESPACE::std_result<handle_t>;
+
```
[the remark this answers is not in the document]
gibber9809: `open_range` already does this check.

components/core/src/clp_s/RangeIndexWriter.hpp
```diff
@@ -0,0 +1,117 @@
...
+    [[nodiscard]] auto empty() const -> bool { return m_ranges.empty(); }
+
+private:
+    // Types
+    struct Range {
+        explicit Range(size_t start) : start_index{start} {}
+
+        size_t start_index{0ULL};
+        std::optional<size_t> end_index{std::nullopt};
+        std::map<std::string, nlohmann::json> fields;
```
wraymo, line 110: Is that to support multiple data types and make serialization easier with the nlohmann library?

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -917,14 +975,83 @@ auto JsonParser::parse_from_ir() -> bool {
...
 
+        size_t file_split_number{0ULL};
         int32_t log_event_idx_node_id{};
-        auto add_log_event_idx_node = [&]() {
+        auto initialize_fields_for_archive = [&]() -> bool {
             if (m_record_log_order) {
                 log_event_idx_node_id
                         = add_metadata_field(constants::cLogEventIdxName, NodeType::Integer);
             }
+            if (auto const rc = m_archive_writer->add_field_to_current_range(
```
wraymo, line 985: Is it possible to deduplicate the code from line 985-1023 with code from line 517-555

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -581,6 +633,11 @@ bool JsonParser::parse() {
                 json_file_iterator.get_num_bytes_read() - bytes_consumed_up_to_prev_archive
         );
 
+        if (ErrorCodeSuccess != m_archive_writer->close_current_range()) {
+            m_archive_writer->close();
+            return false;
+        }
+
```
wraymo, line 640: Do you think it's better to close the range after checking the simdjson errors?

components/core/src/clp_s/RangeIndexWriter.cpp
```diff
@@ -0,0 +1,88 @@
...
+namespace clp_s {
+auto RangeIndexWriter::open_range(size_t start_index)
+        -> OUTCOME_V2_NAMESPACE::std_result<handle_t> {
+    for (auto const& range : m_ranges) {
+        if (start_index < range.start_index || false == range.end_index.has_value()
+            || start_index < range.end_index.value())
+        {
+            return std::errc::invalid_argument;
+        }
+    }
```
wraymo, line 27: Do we want to print an error message in this case (and probably better error code)?

components/core/src/clp_s/RangeIndexWriter.cpp
```diff
@@ -0,0 +1,88 @@
...
+    handle_t handle{m_ranges.size()};
+    m_ranges.emplace_back(start_index);
+    return handle;
+}
+
+auto RangeIndexWriter::close_range(handle_t handle, size_t end_index) -> ErrorCode {
```
wraymo, line 33: Not sure if `handle` always equals to `m_ranges.size() -1`. If that's the case, we may get rid of this variable.

wraymo (commented):
This is a great PR with a very clean design and implementation! I just have a few questions.

wraymo:
For the range index schema, does the following make sense?
```
[{
  "tag_name": ...,
  "tag_type": ...,
  "ranges": [{
      "start": ...,
      "end": ...,
      "value": ...,
      }, ...
   ]
}, ...
]
```

components/core/src/clp_s/ArchiveReaderAdaptor.cpp
```diff
@@ -97,6 +97,13 @@ ErrorCode ArchiveReaderAdaptor::try_read_archive_info(ZstdDecompressor& decompre
     return ErrorCodeSuccess;
 }
 
+auto
+ArchiveReaderAdaptor::try_read_unknown_metadata_packet(ZstdDecompressor& decompressor, size_t size)
```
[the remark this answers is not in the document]
gibber9809: Any metadata packet that this version of the code isn't familiar with. Generally I'm trying to make it so that if you use this version of the code to read a later archive format with more metadata packet types the code will just skip the metadata packet and keep reading as normal.

For this PR specifically it also makes it so that I can split the read side code into a separate PR but let the archive still be readable.

components/core/src/clp_s/ArchiveWriter.cpp
```diff
@@ -145,8 +153,12 @@ void ArchiveWriter::write_archive_metadata(
 
     ZstdCompressor compressor;
     compressor.open(archive_writer, m_compression_level);
-    uint8_t const num_packets{3U};
-    compressor.write_numeric_value(num_packets);
+    uint8_t num_optional_packets{0ULL};
+    if (false == m_range_index_writer.empty()) {
+        ++num_optional_packets;
```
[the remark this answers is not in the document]
gibber9809: By the spec I think the timestamp dictionary is optional, but we've always written it even when it is unused for the whole time we've had the metadata section.

Should I open an issue to refactor the timestamp dictionary writing code to actually write it only when it is used?

components/core/src/clp_s/RangeIndexWriter.hpp
gibber9809: Yep, that's the idea -- nlohmann makes it easy to serialize/derialize msgpack + makes it easy to support multiple types. Since the metadata section is relatively small I don't think that using nlohmann will cause any performance problems for us.

components/core/src/clp_s/JsonParser.cpp
gibber9809: Good point, will move it after.

gibber9809:
> For the range index schema, does the following make sense?
> 
> ```
> [{
>   "tag_name": ...,
>   "tag_type": ...,
>   "ranges": [{
>       "start": ...,
>       "end": ...,
>       "value": ...,
>       }, ...
>    ]
> }, ...
> ]
> ```

I think it would make it a little bit more complicated to search against compared to the current design, particularly for queries that touch multiple metadata fields.

components/core/src/clp_s/RangeIndexWriter.cpp
gibber9809: Yeah, I think it could potentially help catch some weird bugs, but probably not really necessary. I'll get rid of it.

components/core/src/clp_s/JsonParser.cpp
gibber9809: I think it's a bit hard with the way the code is currently written. In general I think we want to deduplicate a lot of the code between the kv-ir/json `parse_X` flows though. Will give it some thought.

wraymo (approved):

kirkrodrigues (approved):
Reviewed CMakeLists.txt

gibber9809:
Just making sure that the unit tests I have written so far in the read side PR pass when rebased on top of this branch, then I will merge.

gibber9809:
> Just making sure that the unit tests I have written so far in the read side PR pass when rebased on top of this branch, then I will merge.

Seems to be passing, will merge.

PR #875, opened 2025-05-01
fix(new-webui): Configure Vite `base` parameter to allow hosting in subdirectories (fixes #867).

### LINKED ISSUE #867: new-webui: Client app fails when hosted in subdirectory due to missing Vite `base` configuration
### Bug

(issue initially discovered by @haiqi96 )
The client application can only be served from the website root (/) because Vite's configuration is missing the `base` parameter that needs to be set to "./". This prevents the app from being hosted in any subdirectory or non-root path.

When attempting to serve the application from a path other than the root, the app fails to load assets correctly, resulting in 404 errors for scripts, stylesheets, and other resources; depending on the hosting environment, the response may even be routed to a different application which gives 200 and causes confusion in debugging. This occurs because all asset URLs are generated as absolute paths from the root instead of relative paths.

### CLP version

faddfc387710c8066c74ae42e52a6ce65ff6ed9b

### Environment

Chrome
Version 135.0.7049.115 

### Reproduction steps

1. Build the client application:
   ```
   cd components/log-viewer-webui/client
   npm install
   npm run build
   ```
2. Observe the assets were generated under `dist`.
3. Serve the application using a simple HTTP server with a path prefix:
   ```
   npx http-server
   ```
4. Open your browser and navigate to http://localhost:8080/dist/index.html
5. Open the browser debugger and observe the `index.html` loads but individual resources fail to load with 404 errors: Browser console shows errors for missing JavaScript and CSS files.
6. Inspect the `index.html` source and note that all asset URLs are absolute paths (e.g., /assets/main.js) instead of relative paths (./assets/main.js)

AVMatthews:
# Description

Fixes issue #867.

Adds base to the client Vite config.

# Checklist

* [ x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x ] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x ] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

Verified that the individual resources from the index.html in the list directory.

Started the webui, compressed and searched a file, and verified that the links froth search results directed correctly to the log viewer.

## Summary by CodeRabbit

- **Chores**
  - Updated asset paths to be relative to the current directory for improved compatibility when deploying built files.

### DIFF
--- a/components/log-viewer-webui/client/vite.config.ts
+++ b/components/log-viewer-webui/client/vite.config.ts
@@ -17,4 +17,5 @@ export default defineConfig({
         },
     },
     publicDir: "public",
+    base: "./",
 });
--- a/components/log-viewer-webui/client/vite.config.ts
+++ b/components/log-viewer-webui/client/vite.config.ts
@@ -4,7 +4,9 @@ import {defineConfig} from "vite";
 
 // https://vite.dev/config/
 export default defineConfig({
+    base: "./",
     plugins: [react()],
+    publicDir: "public",
     server: {
         port: 8080,
         proxy: {
@@ -16,6 +18,4 @@ export default defineConfig({
             },
         },
     },
-    publicDir: "public",
-    base: "./",
 });


components/log-viewer-webui/client/vite.config.ts
```diff
@@ -17,4 +17,5 @@ export default defineConfig({
         },
     },
     publicDir: "public",
+    base: "./",
```
junhaoliao, line 20: Let's alphabetize the properties. e.g.,
```
    base: "./",
    plugins: [react()],
    publicDir: "public",
    server: {
        port: 8080,
        proxy: {
            "/query/": {
                // Below target should match the server's configuration in
                // `components/log-viewer-webui/server/.env` (or `.env.local` if overridden)
                target: "http://localhost:3000/",
                changeOrigin: true,
            },
        },
    },
```

junhaoliao (approved):

PR #833, opened 2025-04-17
docs(kv-ir): Add background necessary to understand the kv-ir stream format.

kirkrodrigues:
# Description

This PR adds docs explaining any background information necessary to explain the kv-ir stream format. Future PRs will flesh out the format itself.

It takes over where #707 left off.

The theme for Mermaid flowcharts used in this PR is slightly different than the one used in existing flowcharts, so they've also been updated for consistency.

Note that there is seemingly no way to set the theme globally due to some kind of bug in sphinxcontrib-mermaid (mermaid_init_js doesn't work).

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

Validated the new pages render as expected in both light and dark modes.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

- **Documentation**
  - Added new documentation introducing the KV-IR stream format and its background, including detailed explanations of compression, schema encoding, and storage processes.
  - Updated the developer guide index to include the new KV-IR design documentation.
  - Changed the colour scheme of workflow diagrams for improved visual clarity.

PR #888, opened 2025-05-08
docs(kv-ir): Explain support for auto-generated KV pairs.

kirkrodrigues:
# Description

As the title says. Future PRs will flesh out the format itself.

This continues the work started in #707 and #833.

# Checklist

  breaking change.

# Validation performed

* Workflows succeed.
* New pages render as expected.

## Summary by CodeRabbit

- **Documentation**
  - Introduced a new guide explaining the distinction between auto-generated and user-generated key-value pairs in KV-IR streams, including query syntax and examples.
  - Updated background documentation for clearer references and direct linking to the new guide.
  - Added a navigation card and table of contents entry for the new section to improve discoverability.


docs/src/dev-guide/design-kv-ir-streams/auto-gen-kv-pairs.md
```diff
@@ -0,0 +1,92 @@
...
+One of the key differences between KV-IR streams and a format like [JSON lines][json-lines] is that
+KV-IR streams support storing KV pairs generated by the logging framework separately from those
+generated by the user. For instance, consider the structured [Zap] log printing statement (LPS) in
+[Figure 1](#figure-1) and its output, as JSON, in [Figure 2](#figure-2). Comparing the user-defined
+KV pairs in [Figure 1](#figure-1) and the output in [Figure 2](#figure-2), we can see that the
+logging library added the timestamp and level KV pairs automatically. (The key for the message,
+timestamp, and log level are Zap configuration options.) Storing these two categories of KV pairs in
+a format like JSON has the disadvantage that both categories share the same key namespace; so a user
+needs to pick keys that don't conflict with the logging library's keys, and the logging library
+needs to pick keys that are unlikely to conflict with a user's keys.
```
LinZhihao-723, line 12:
```suggestion
logging library added the timestamp and level KV pairs automatically.

Although the keys for the message, timestamp, and log level can be configured via Zap options,
storing these two categories of KV pairs in a format like JSON requires additional effort to prevent
keys conflicts, since both categories share the same key namespace:

* Users must avoid selecting keys that might conflict with those used by the logging framework—-this
  may even require knowledge of all reserved keys.
* The logging library must choose keys that are unlikely to collide with user-defined keys, and
  still implement a policy to resolve any conflicts that do occur.
```
* Add a little more details to explain why this is really a headache and inevitable.
* How about using `users` instead `a user`? Feel like it's easier to read.

docs/src/dev-guide/design-kv-ir-streams/auto-gen-kv-pairs.md
```diff
@@ -0,0 +1,92 @@
...
+
+Since KV-IR streams store auto-generated KV pairs and user-generated KV pairs in different
+namespaces, we differentiate which namespace to query using special query syntax. For instance, if a
+log event has both an auto-generated KV pair and user-generated KV pair with the key `timestamp`, we
+need some way to allow the user to query one of the two. We could use two different queries, one to
+query each namespace, but this wouldn't allow users to join filters from both namespaces. Instead,
+we use the key-prefix `@` to denote a filter on an auto-generated KV pair---e.g., `@timestamp: 0`
+denotes a filter for an auto-generated KV pair with the key `timestamp` and value `0`. Accordingly,
+filters on user-generated KV pairs whose keys start with the character `@` will need to escape the
+`@` symbol.
```
LinZhihao-723, line 67:
```suggestion
If a log event has both an auto-generated KV pair and user-generated KV pair with the same key,
e.g., `timestamp`, we need a mechanism to let the user query one or the other unambiguously.
Naively, we could use two different queries, one to query each namespace, but this wouldn't allow
users to join filters from both namespaces, and it would double the query evaluation workload. As a
solution, we introduce a special query syntax to differentiate the queried namespace: we use the
key-prefix `@` to denote a filter on an auto-generated KV pair---e.g., `@timestamp: 0` denotes a
filter for an auto-generated KV pair with the key `timestamp` and value `0`. Accordingly, filters on
user-generated KV pairs whose keys start with the character `@` will need to escape the
`@` symbol.
```
Reordered the flow a little bit as: <what we need (with an example)> -> <why the naive solution doesn't work> -> <what is the new syntax we provide>. What do u think?

docs/src/dev-guide/design-kv-ir-streams/auto-gen-kv-pairs.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Additional benefits
+
+One benefit of supporting auto-generated KV pairs is that our CLP logging libraries or plugins can
+use a consistent set of keys for auto-generated KV pairs, making it easier for users to filter for
+those keys. For instance, if a user is logging in both Java and Python, and they use CLP logging
+plugins to generate KV-IR streams, the streams generated should both use the same key for log
```
LinZhihao-723, line 74:
```suggestion
plugins to generate KV-IR streams, the generated streams should both use the same key for log
```
Should sound more natural?

LinZhihao-723 (changes requested):
Overall lgtm.
Left some comments for two paragraphs.

kirkrodrigues:
@coderabbitai review

LinZhihao-723 (approved):
For the PR title, how about:
```
docs(kv-ir): Add introduction for auto-generated KV pairs.
```

kirkrodrigues:
> 
> ```
> docs(kv-ir): Add introduction for auto-generated KV pairs.
> ```

Mm, that might be misinterpreted as we only added the intro for the section. How about:

```
docs(kv-ir): Explain support for auto-generated KV pairs.
```

LinZhihao-723:
> > For the PR title, how about:
> > ```
> > docs(kv-ir): Add introduction for auto-generated KV pairs.
> > ```
> 
> Mm, that might be misinterpreted as we only added the intro for the section. How about:
> 
> ```
> docs(kv-ir): Explain support for auto-generated KV pairs.
> ```

Sure. lgtm

PR #889, opened 2025-05-08
feat(clp-s): Add the read-side implementation for the archive range index.

gibber9809:
# Description

This PR follows up on #847 and implements the read side of the archive range index, and also adds end to end tests for this feature across both the JSON and kv-ir ingestion paths.

The code for reading the range index metdata packet lives in `ArchiveReaderAdaptor`, and access to the range index is forwarded to readers through `ArchiveReader`.

# Checklist

  breaking change.

# Validation performed

* Added end to end tests that verify that the archive range index acts as expected when ingesting JSON and kv-ir files

### DIFF
--- a/components/core/src/clp_s/ArchiveReader.hpp
+++ b/components/core/src/clp_s/ArchiveReader.hpp
@@ -117,6 +117,10 @@ public:
 
     std::shared_ptr<ReaderUtils::SchemaMap> get_schema_map() { return m_schema_map; }
 
+    auto get_range_index() const -> std::vector<ArchiveReaderAdaptor::RangeIndexEntry> const& {
+        return m_archive_reader_adaptor->get_range_index();
+    }
+
     /**
      * Writes decoded messages to a file.
      * @param writer
--- a/components/core/src/clp_s/ArchiveReaderAdaptor.hpp
+++ b/components/core/src/clp_s/ArchiveReaderAdaptor.hpp
@@ -1,10 +1,19 @@
 #ifndef CLP_S_ARCHIVEREADERADAPTOR_HPP
 #define CLP_S_ARCHIVEREADERADAPTOR_HPP
 
+#include <cstddef>
 #include <memory>
 #include <optional>
 #include <string>
 #include <string_view>
+#include <vector>
+
+// We use NOLINTNEXTLINE to satisfy clang-tidy here because while we don't use any symbols from
+// `nlohmann/json.hpp` directly this code does not compile without the definition of
+// `nlohmann::basic_json<>` found in the `nlohmann/json.hpp` header.
+// NOLINTNEXTLINE(misc-include-cleaner)
+#include <nlohmann/json.hpp>
+#include <nlohmann/json_fwd.hpp>
 
 #include "../clp/BoundedReader.hpp"
 #include "../clp/ReaderInterface.hpp"
@@ -26,6 +35,18 @@ public:
         // Constructors
         OperationFailed(ErrorCode error_code, char const* const filename, int line_number)
                 : TraceableException(error_code, filename, line_number) {}
+    };
+
+    struct RangeIndexEntry {
+        explicit RangeIndexEntry(size_t start_index, size_t end_index, nlohmann::json&& fields)
+                : start_index{start_index},
+                  end_index{end_index},
+                  // Note: brace initializer would make nlohmann wrap the fields object in an array.
+                  fields(fields) {}
+
+        size_t start_index;
+        size_t end_index;
+        nlohmann::json fields;
     };
 
     explicit ArchiveReaderAdaptor(Path const& archive_path, NetworkAuthOption const& network_auth);
@@ -62,6 +83,8 @@ public:
 
     ArchiveHeader const& get_header() const { return m_archive_header; }
 
+    std::vector<RangeIndexEntry> const& get_range_index() const { return m_range_index; }
+
 private:
     /**
      * Tries to read an ArchiveFileInfo packet from the archive metadata.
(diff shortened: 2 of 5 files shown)


components/core/src/clp_s/ArchiveReaderAdaptor.hpp
```diff
@@ -28,6 +38,18 @@ class ArchiveReaderAdaptor {
                 : TraceableException(error_code, filename, line_number) {}
     };
 
+    struct RangeIndexEntry {
```
wraymo, line 41: Not sure if we should put it inside `ArchiveReaderAdopter`

components/core/src/clp_s/ArchiveReader.hpp
```diff
@@ -117,6 +117,10 @@ class ArchiveReader {
 
     std::shared_ptr<ReaderUtils::SchemaMap> get_schema_map() { return m_schema_map; }
 
+    auto get_range_index() const -> std::vector<ArchiveReaderAdaptor::RangeIndexEntry> const& {
```
wraymo, line 120: I think this method will be used when we have search support?

components/core/tests/test-clp_s-range_index.cpp
```diff
@@ -0,0 +1,222 @@
...
+auto get_test_input_local_path() -> std::string;
+void serialize_record(
+        nlohmann::json const& auto_gen,
+        nlohmann::json const& user_gen,
+        clp::ffi::ir_stream::Serializer<clp::ir::eight_byte_encoded_variable_t>& serializer
+);
+void compress_ir();
+void compress(bool single_file_archive, bool from_ir);
+void check_archive_metadata(bool from_ir);
+
```
wraymo, line 42: Do you think we can have a test utility file which includes functions that can be shared by test-clp_s-end-to-end and test-clp_s-range_index (and maybe add path to compress/compress_ir signatures)

components/core/tests/test-clp_s-range_index.cpp
```diff
@@ -0,0 +1,222 @@
...
+#include "../src/clp_s/ArchiveReader.hpp"
+#include "../src/clp_s/InputConfig.hpp"
+#include "../src/clp_s/JsonParser.hpp"
+#include "TestOutputCleaner.hpp"
+
+constexpr std::string_view cTestRangeIndexArchiveDirectory{"test-range-index-archive"};
+constexpr std::string_view cTestRangeIndexIRDirectory{"test-range-index-ir"};
+constexpr std::string_view cTestRangeIndexInputFileDirectory{"test_log_files"};
+constexpr std::string_view cTestRangeIndexInputFile{"test_no_floats_sorted.jsonl"};
+constexpr std::string_view cTestRangeIndexIRInputFile{"test_no_floats_sorted.ir"};
```
wraymo, line 27: And some constants as well

components/core/tests/test-clp_s-range_index.cpp
```diff
@@ -0,0 +1,222 @@
...
+            user_gen_bytes.size()
+    )};
+    auto const auto_gen_obj{auto_gen_handle.get()};
+    auto const user_gen_obj{user_gen_handle.get()};
+    REQUIRE(msgpack::type::MAP == auto_gen_obj.type);
+    REQUIRE(msgpack::type::MAP == user_gen_obj.type);
+    REQUIRE(serializer.serialize_msgpack_map(auto_gen_obj.via.map, user_gen_obj.via.map));
+}
+
+void compress_ir() {
```
wraymo, line 80: Maybe we should rename this method to something like `generate_ir`? Or we could add `test_no_floats_sorted.ir` to `test_log_files`

wraymo (commented):
Nice work! I think it's overall very clear and well-structured. Just a few comments on renaming and refactoring.

components/core/src/clp_s/ArchiveReaderAdaptor.hpp
gibber9809: Would you prefer moving it outside of this class and still defining it in this file, or moving it to a separate header?

components/core/src/clp_s/ArchiveReader.hpp
gibber9809: Yes, that is the plan

components/core/tests/test-clp_s-range_index.cpp
gibber9809: I like `generate_ir`, let's go with that.

components/core/tests/test-clp_s-range_index.cpp
gibber9809: Sure, I think it is probably worth moving `compress` and `generate_ir` into a helper.

gibber9809: Actually on second thought I might only move `compress` into a helper for now since we have similar code for 3 different tests. I prefer leaving `generate_ir` as is because there is currently only one user.

components/core/tests/test-clp_s-range_index.cpp
gibber9809: I think I prefer defining them in each test because it makes it clear what files they depend on at a glance.

components/core/src/clp_s/ArchiveReaderAdaptor.hpp
gibber9809: Going to assume that moving it out of `ArchiveReaderAdaptor` but keeping it in the same file is fine for now.

wraymo: I prefer moving it outside `ArchiveReaderAdaptor`, and since it's very similar to `Range` in `RangeIndexWriter`. Is it possible to merge them into one struct?

gibber9809: Merging them might be a bit awkward because in `RangeIndexWriter` we need to know whether `end_index` has been specified so we wrap that field in `optional` but on the read-side we know that it always exists so wrapping it in `optional` would be cumbersome.

wraymo (approved):

components/core/tests/ClpSTestUtils.cpp
kirkrodrigues, line 1: Since these ClpSTestUtils.cpp/hpp don't contain classes/types, they should be named using snake_case, right? See our internal guidelines: [C++ -> File naming and organization](https://www.notion.so/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?pvs=4#1a504e4d9e6b80d58247e1fba204a3f7) (I know we should open-source them so they're easier to find... We're trying, we're trying.).

kirkrodrigues (commented):
Something's also wrong with the clang-tidy tasks where they're not linting tests properly. Will probably need your help to fix the linting errors in the PR where we fix the task.

kirkrodrigues (approved):
Stamping tests.

Issue #840, opened 2025-04-22
Kv-pair IR stream: Search implementation plan.


LinZhihao-723:
This is the implementation plan for #747. The prototype implementation is in [this](https://github.com/gibber9809/clp/tree/kvir-search-draft) branch.

## Overview

As an extension of #747, this issue discusses the implementation plan and tracks the implementation progress. To support searching kv-pair IR stream, we will break the required features into a PR series with the following steps:
- **Relevant KQL AST Utilities** to make clp-s KQL query compatible with kv-pair IR stream.
- **Projection Handler Interface** to define the projection handling interface.
- **Query Handler** to handle the core search logic during deserialization.
- **Deserializer Integration** to integrate the features above into the current IR deserializer.
- **clp-s Integration** to integrate the overall search feature into `clp-s`' cli.

### Relevant KQL AST Utilities

The current KQL AST is designed to operate on clp-s archives, which have a different schema tree implementation comparing to kv-pair IR stream.
To adapt these differences, we need some utility code to:
- Convert IR's schema tree type to `clp_s::search::ast::LiteralTypeBitmask`
- Convert IR's node-type-value-pair to `clp_s::search::ast::LiteralType`
- Evaluate a KQL filter expression against the deserialized IR value.

### Projection Handler Interface

As discussed in #747, users need to define their own projection resolution handler to maintain the full-key-to-node-ID mapping across the stream.
This mapping should be applied by user-level code, which is outside the deserializer.

The *Projection Handler* is a concept that defines the interface for users to implement their own logic to handle projection resolution.

### Query Handler

Query handler is an object that:
- Hold all the necessary data structures to proceed streaming IR search/projection.
- Be responsible for evaluating the query on the deserialized node-id-value-pairs.

As designed to be a part of the deserializer, the following APIs are needed:
- ***column_resolution_update***: Handle column resolution and update the relevant key-to-node-ID mapping required by the query or the projection (possibly calling the projection handler). Will be called whenever schema tree node insertion IR unit is deserialized.
- ***evaluate_node_id_value_pairs***: Execute the query on the given node-ID-value-pairs. Will be called whenever the log event IR unit is deserialized.

This object implements the core search logic so most of the engineering efforts will be spent here.

### Deserializer Integration

The components discussed above will eventually be integrated into the current IR deserializer. As discussed above, if a query is given, the deserializer will behave as following to handle the query:
- For schema tree node insertion IR unit, the deserializer will call the relevant query handler's API to update the column resolution.
[...]

PR #890, opened 2025-05-08
feat(kv-ir): Integrate `QueryHandler` into the IR stream `Deserializer`.

LinZhihao-723:
# Description

## References

* This PR is one of the PR series #840.
* This PR depends on #882.

## Overview

This PR integrates `QueryHandler` into `Deserializer`. `QueryHandler` is added as a templated parameter to `Deserializer`, where it can be an instance of `clp::ffi::ir_stream::search::QueryHandler` as introduced in the previous PRs in this series, or `clp::ffi::ir_stream::search::EmptyQueryHandler` (an empty class) to perform no query handling. This design ensures the type/functionality of a deserializer is determined statically: it's either designed to read the entire stream, or is configured to read the stream with KQL query filtering or column projections.

The factory function of the deserializer is templated to have two different versions:
- The first one takes no query handler (which preserves the existing API). The constructed deserializer behaves in the same way as before this PR.
- The second one explicitly takes a query handler, which returns a deserializer with query handler support.

The C++ typing system will treat the deserializers above as different types. However, if users use the `auto` keyword for creation, the type can be automatically deduced according to the chosen factory function.

# Checklist

  breaking change.

# Validation performed

* [x] Add test cases to ensure the integration is functionally correct.

## Summary by CodeRabbit

- **New Features**
  - Added support for evaluating search queries against key-value pair log events, enabling advanced query evaluation and column projection during deserialization.
  - Introduced new error codes and improved error handling for query evaluation.

- **Bug Fixes**
  - Enhanced handling of empty or null query expressions to avoid errors in query processing.

- **Refactor**
  - Streamlined and restructured query handler interfaces and internal evaluation logic for improved maintainability and extensibility.

- **Tests**
  - Expanded and improved test coverage for query evaluation against log events, including new utility functions and comprehensive test cases.

- **Chores**
  - Added and updated utility headers and build configuration to support new query handler features.


LinZhihao-723:
@coderabbitai stop

components/core/src/clp/ffi/ir_stream/search/QueryHandlerReq.hpp
LinZhihao-723, line 1: @coderabbitai stop

LinZhihao-723, line 1: @coderabbitai review

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
LinZhihao-723, line 1: @coderabbitai review

LinZhihao-723:
@coderabbitai pause

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
LinZhihao-723, line 1: @coderabbitai review

components/core/src/clp/ffi/ir_stream/search/QueryHandlerReq.hpp
```diff
@@ -0,0 +1,43 @@
...
+ * deserialize an IR stream without performing any query evaluation.
+ */
+struct EmptyQueryHandler {};
+
+/**
+ * A type trait to determine if a given type is a valid `clp::ffi::ir_stream::search::QueryHandler`.
+ * @tparam T The type to check.
+ */
+template <typename T>
+struct IsQueryHandler : std::false_type {};
```
kirkrodrigues, line 21: How about `IsNonEmptyQueryHandler`? The name `IsQueryHandler` implies that `EmptyQueryHandler` should also satisfy this trait.

components/core/src/clp/ffi/ir_stream/search/QueryHandlerReq.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Specialization of `IsQueryHandler` for `clp::ffi::ir_stream::search::QueryHandler`.
+ * This specialization evaluates to `true` for valid `QueryHandler` types.
```
kirkrodrigues, line 25:
```suggestion
```

I think the fact that we're specializing the trait already implies it would evaluate to true. The definition is also simple to read.

components/core/src/clp/ffi/ir_stream/search/QueryHandlerReq.hpp
```diff
@@ -0,0 +1,43 @@
...
+ * @tparam NewProjectedSchemaTreeNodeCallbackType
+ */
+template <NewProjectedSchemaTreeNodeCallbackReq NewProjectedSchemaTreeNodeCallbackType>
+struct IsQueryHandler<QueryHandler<NewProjectedSchemaTreeNodeCallbackType>> : std::true_type {};
+
+/**
+ * Requirements for a query handler that can be used with`clp::ffi::ir_stream::Deserializer`.
+ * A valid query handler must either be `EmptyQueryHandler` or satisfy the `IsQueryHandler` trait,
+ * and must also be move-constructible.
+ * @tparam QueryHandlerType The type to validate as a query handler.
```
kirkrodrigues, line 35:
```suggestion
 * Requirements for a query handler that can be used with `clp::ffi::ir_stream::Deserializer`. A
 * valid query handler must be:
 *
 * - an `EmptyQueryHandler` or satisfy the `IsQueryHandler` trait.
 * - move constructible.
 *
 * @tparam QueryHandlerType The type to check.
```

components/core/src/clp/ffi/ir_stream/search/QueryHandlerReq.hpp
kirkrodrigues, line 21: Am I right that the primary use of this type trait is really to hide the `NewProjectedSchemaTreeNodeCallbackType` template parameter from the `Deserializer`? If so, could we add some text about that? Otherwise, it looks a bit odd to have a type trait where there will only ever be one specialization.

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -20,37 +20,62 @@
...
+        IrUnitHandlerInterface IrUnitHandler,
+        search::QueryHandlerReq QueryHandlerType = search::EmptyQueryHandler>
 requires(std::move_constructible<IrUnitHandler>)
 class Deserializer {
 public:
     // Factory function
     /**
-     * Creates a deserializer by reading the stream's preamble from the given reader.
+     * Creates a deserializer with an empty query handler.
+     * This factory function is used when the deserializer doesn't requires any query evaluations.
```
kirkrodrigues, line 47:
```suggestion
     * Creates a deserializer with an empty query handler (for use when the deserializer won't be
     * used to perform queries or column projections).
```

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -20,37 +20,62 @@
...
+            -> OUTCOME_V2_NAMESPACE::std_result<Deserializer>
+    requires std::is_same_v<QueryHandlerType, search::EmptyQueryHandler>
+    {
+        return create_generic(reader, std::move(ir_unit_handler), {});
+    }
+
+    /**
+     * Creates a deserializer with a query handler.
+     * This factory function is used when the deserializer needs to perform query evaluations or
+     * column projections.
```
kirkrodrigues, line 64:
```suggestion
     * Creates a deserializer with a query handler (for use when the deserializer will be used to
     * perform queries or column projections).
```

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -20,37 +20,62 @@
...
 #include "protocol_constants.hpp"
+#include "search/QueryHandlerReq.hpp"
 #include "utils.hpp"
 
 namespace clp::ffi::ir_stream {
 /**
  * A deserializer for reading IR units from a CLP kv-pair IR stream. An IR unit handler should be
- * provided to perform user-defined operations on each deserialized IR unit.
+ * provided to perform user-defined operations on each deserialized IR unit. Additionally, a query
+ * handler can be provided to handle KQL queries and column projections.
```
kirkrodrigues, line 30:
```suggestion
 * handler can be provided to handle queries and column projections.
```

Just in case we rename the query language in the future, it's probably easier to not mention it here.

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -65,34 +90,44 @@ class Deserializer {
 
     // Methods
     /**
-     * Deserializes the stream from the given reader up to and including the next log event IR unit.
+     * Deserializes the stream from the given reader up to and including the next log event IR unit,
+     * and invokes the user-defined IR unit handler according to the deserialized IR unit type.
+     *
+     * NOTE: If the deserialized IR unit is `IrUnitType::LogEvent` and the query handler is not
+     * `search::EmptyQueryHandler`, `handle_log_event` will only be invoked if the query evaluation
```
kirkrodrigues, line 97:
```suggestion
     * `search::EmptyQueryHandler`, `handle_log_event` will only be invoked if the query handler
     
```

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -65,34 +90,44 @@ class Deserializer {
...
-     * @return IRUnitType::LogEvent if a log event IR unit is deserialized, or an error code
+     * @return IrUnitType::LogEvent if a log event IR unit is deserialized, or an error code
      * indicating the failure:
      * - Forwards `deserialize_ir_unit_kv_pair_log_event`'s return values if it failed to
      *   deserialize and construct the log event.
      * - Forwards `handle_log_event`'s return values from the user-defined IR unit handler on
      *   unit handling failure.
-     * @return IRUnitType::SchemaTreeNodeInsertion if a schema tree node insertion IR unit is
+     * - Forwards `search::QueryHandler::evaluate_kv_pair_log_event`'s return values on failure, if
+     *   `m_query_handler` is not `search::EmptyQueryHandler`.
```
kirkrodrigues, line 113:
```suggestion
     *   `QueryHandlerType` is not `search::EmptyQueryHandler`.
```

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -65,34 +90,44 @@ class Deserializer {
...
+     * @return IrUnitType::SchemaTreeNodeInsertion if a schema tree node insertion IR unit is
      * deserialized, or an error code indicating the failure:
      * - Forwards `deserialize_ir_unit_schema_tree_node_insertion`'s return values if it failed to
      *   deserialize and construct the schema tree node locator.
      * - Forwards `handle_schema_tree_node_insertion`'s return values from the user-defined IR unit
      *   handler on unit handling failure.
+     * - Forwards `search::QueryHandler::update_partially_resolved_columns`'s return values on
+     *   failure, if `m_query_handler` is not `search::EmptyQueryHandler`.
```
kirkrodrigues, line 121:
```suggestion
     *   failure, if `QueryHandlerType` is not `search::EmptyQueryHandler`.
```

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -118,10 +153,34 @@ class Deserializer {
...
+     * - std::errc::protocol_error if the IR stream is corrupted
+     * - std::errc::protocol_not_supported if either:
+     *   - the IR stream contains an unsupported metadata format;
+     *   - the IR stream's version is unsupported;
+     *   - or the IR stream's user-defined metadata is not a JSON object.
+     */
+    [[nodiscard]] static auto create_generic(
+            ReaderInterface& reader,
+            IrUnitHandler ir_unit_handler,
+            QueryHandlerType query_handler
```
kirkrodrigues, line 172: Missing from docstring.

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -176,13 +239,43 @@ auto Deserializer<IrUnitHandler>::create(ReaderInterface& reader, IrUnitHandler
...
 
+/**
+ * Wrapper of `Deserializer`'s factory function to enable auto type deduction.
+ * @param reader
+ * @param ir_unit_handler
+ * @return Forwards `Deserializer::create`'s return values.
+ */
 template <IrUnitHandlerInterface IrUnitHandler>
+requires std::is_move_constructible_v<IrUnitHandler>
+[[nodiscard]] auto make_deserializer(
```
kirkrodrigues, line 257: * Why do we need this if we have the factory methods?
* Why are these defined after `Deserializer`'s constructor?

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -176,13 +239,43 @@ auto Deserializer<IrUnitHandler>::create(ReaderInterface& reader, IrUnitHandler
...
-    return Deserializer{std::move(ir_unit_handler), std::move(metadata_json)};
+    return Deserializer{
+            std::move(ir_unit_handler),
+            std::move(metadata_json),
+            std::move(query_handler)
+    };
 }
 
+/**
+ * Wrapper of `Deserializer`'s factory function to enable auto type deduction.
```
kirkrodrigues, line 250:
```suggestion
 * Wrapper for `Deserializer`'s factory function to enable automatic type deduction.
```

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -176,13 +239,43 @@ auto Deserializer<IrUnitHandler>::create(ReaderInterface& reader, IrUnitHandler
...
+ * @param reader
+ * @param ir_unit_handler
+ * @return Forwards `Deserializer::create`'s return values.
+ */
 template <IrUnitHandlerInterface IrUnitHandler>
+requires std::is_move_constructible_v<IrUnitHandler>
```
kirkrodrigues, line 256: Can we add this to `IrUnitHandlerInterface`?

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -176,13 +239,43 @@ auto Deserializer<IrUnitHandler>::create(ReaderInterface& reader, IrUnitHandler
...
+[[nodiscard]] auto make_deserializer(
+        ReaderInterface& reader, IrUnitHandler ir_unit_handler
+) -> OUTCOME_V2_NAMESPACE::std_result<Deserializer<IrUnitHandler>>;
+
+/**
+ * Wrapper of `Deserializer`'s factory function to enable auto type deduction.
```
kirkrodrigues, line 262:
```suggestion
 * Wrapper for `Deserializer`'s factory function to enable automatic type deduction.
```

kirkrodrigues (changes requested):
Reviewed the parts that are different from #882.

components/core/src/clp/ffi/ir_stream/search/QueryHandlerReq.hpp
LinZhihao-723: Not quite. Deserializer takes the query handler type as a template parameter. QueryHandler itself is a class template, and this type trait is used to check whether the type provided to Deserializer is an instantiation of the QueryHandler<T> template. C++20 does not provide a built-in way to directly check whether a type is an instantiation of a particular class template, so we use a custom type trait for that purpose.

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
LinZhihao-723: > Why do we need this if we have the factory methods?

The factory function doesn't allow auto type deduction. Let's say we create a query handler with a lambda projection callback:
```C++
    auto new_projected_schema_tree_node_callback
            = []([[maybe_unused]] bool is_auto_gen,
                 [[maybe_unused]] SchemaTree::Node::id_t node_id,
                 [[maybe_unused]] std::string_view key) -> outcome_v2::std_result<void> {
        return outcome_v2::success();
    };
    auto query_handler{
            QueryHandler<decltype(new_projected_schema_tree_node_callback)>::create(
                    new_projected_schema_tree_node_callback,
                    std::shared_ptr<clp_s::search::ast::Expression>(),
                    {},
                    false
            ).value() // Ignoring the error
    };
```
Then, if we use the factory:
```C++
    auto deserializer_without_type_deduction{
            Deserializer<IrUnitHandler, QueryHandler<decltype(new_projected_schema_tree_node_callback)>>::
                    create(reader, IrUnitHandler{}, std::move(query_handler)).value()
    };
```
the user needs to provide both `IrUnitHandler` and `QueryHandler<decltype(new_projected_schema_tree_node_callback)>` explicitly since the factory is a method of this template class.

With `make_deserializer` helper, we can then rely on the type deduction:
```C++
    auto deserializer_with_type_deduction{
            make_deserializer(buf_reader, IrUnitHandler{}, std::move(query_handler))).value()
    };
```
which makes it somehow easier to use (by avoiding `decltype`; it's tricky to explicitly write out `QueryHandler<T>`'s instantiated type).

LinZhihao-723: > Why are these defined after `Deserializer`'s constructor?

Sorry, I don't fully get this. They're non-member variables, and the definition ordering matches the declaration.

That said, maybe we can create a separate file called `make_deserializer.hpp` and move these two methods there instead?

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
LinZhihao-723: I'd prefer to fix this in another PR. This concept was added before we have the new concept naming rule, so we should also rename it to `IrUnitHandlerInterfaceReq`.

components/core/src/clp/ffi/ir_stream/search/test/test_deserializer_integration.cpp
LinZhihao-723, line 1: Also added test cases for deserializer integration with queries.
Can u help review this? @kirkrodrigues

components/core/src/clp/ffi/ir_stream/search/test/test_deserializer_integration.cpp
```diff
@@ -0,0 +1,316 @@
...
+    };
+    REQUIRE_FALSE(query_handler_result.has_error());
+
+    BufferReader buf_reader{
+            size_checked_pointer_cast<char const>(ir_stream_bytes.data()),
+            ir_stream_bytes.size()
+    };
+    auto deserializer_result{
+            make_deserializer(buf_reader, IrUnitHandler{}, std::move(query_handler_result.value()))
+    };
```
LinZhihao-723, line 291: fyi, this is what I mean by https://github.com/y-scope/clp/pull/890#discussion_r2083644722.
The type is auto-deduced: no need to explicitly state the type of unit handler and query handler.
@kirkrodrigues

components/core/src/clp/ffi/ir_stream/search/QueryHandlerReq.hpp
kirkrodrigues: Gotcha.

components/core/src/clp/ffi/ir_stream/search/QueryHandlerReq.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace clp::ffi::ir_stream::search {
+/**
+ * Defines an empty query handler that can be used with `clp::ffi::ir_stream::Deserializer` to
+ * deserialize an IR stream without performing any query evaluation.
+ */
+struct EmptyQueryHandler {};
+
+/**
+ * A type trait to determine if a given type is a valid `clp::ffi::ir_stream::search::QueryHandler`.
```
kirkrodrigues, line 17:
```suggestion
 * A type trait to determine if a given type is an instantiation of
 * `clp::ffi::ir_stream::search::QueryHandler`.
```

components/core/src/clp/ffi/ir_stream/search/QueryHandlerReq.hpp
```diff
@@ -0,0 +1,46 @@
...
+ * @tparam T The type to check.
+ */
+template <typename T>
+struct IsNonEmptyQueryHandler : std::false_type {};
+
+/**
+ * Specialization of `IsQueryHandler` for `clp::ffi::ir_stream::search::QueryHandler`.
```
kirkrodrigues, line 24:
```suggestion
 * Specialization of `IsNonEmptyQueryHandler` for `clp::ffi::ir_stream::search::QueryHandler`.
```

Can't rely on CLion to get everything right.

components/core/src/clp/ffi/ir_stream/search/QueryHandlerReq.hpp
```diff
@@ -0,0 +1,46 @@
...
+ */
+template <NewProjectedSchemaTreeNodeCallbackReq NewProjectedSchemaTreeNodeCallbackType>
+struct IsNonEmptyQueryHandler<QueryHandler<NewProjectedSchemaTreeNodeCallbackType>>
+        : std::true_type {};
+
+/**
+ * Requirements for a query handler that can be used with `clp::ffi::ir_stream::Deserializer`. A
+ * valid query handler must be:
+ *
+ * - an `EmptyQueryHandler` or satisfy the `IsQueryHandler` trait.
```
kirkrodrigues, line 35:
```suggestion
 * - an `EmptyQueryHandler` or satisfy the `IsNonEmptyQueryHandler` trait.
```

Always double-check your docstrings.

components/core/src/clp/ffi/ir_stream/search/test/utils.hpp
```diff
@@ -86,6 +95,93 @@ class ColumnQueryPossibleMatches {
...
+ * @return The encoded result.
+ */
+template <typename encoded_variable_t>
+requires std::is_same_v<encoded_variable_t, ir::eight_byte_encoded_variable_t>
+         || std::is_same_v<encoded_variable_t, ir::four_byte_encoded_variable_t>
+[[nodiscard]] auto get_encoded_text_ast(std::string_view text)
+        -> clp::ir::EncodedTextAst<encoded_variable_t>;
+
+/**
+ * Unpacks and serializes the given msgpack bytes using kv serializer.
```
kirkrodrigues, line 112:
```suggestion
 * Unpacks and serializes the given msgpack bytes using the given serializer.
```

components/core/src/clp/ffi/ir_stream/search/test/utils.hpp
```diff
@@ -86,6 +95,93 @@ class ColumnQueryPossibleMatches {
...
+        return false;
+    }
+
+    // NOLINTNEXTLINE(misc-include-cleaner)
+    auto const user_gen_msgpack_byte_handle{msgpack::unpack(
+            clp::size_checked_pointer_cast<char const>(user_gen_msgpack_bytes.data()),
+            user_gen_msgpack_bytes.size()
+    )};
+
+    auto const user_gen_msgpack_obj{user_gen_msgpack_byte_handle.get()};
```
kirkrodrigues, line 170:
```suggestion
    auto const user_gen_msgpack_obj{user_gen_msgpack_byte_handle.get()};

```

components/core/src/clp/ffi/ir_stream/search/test/test_deserializer_integration.cpp
```diff
@@ -0,0 +1,316 @@
...
+// The following clang-tidy suppression should be removed once the circular dependency is resolved.
+// NOLINTNEXTLINE(misc-header-include-cycle)
+#include "../../decoding_methods.hpp"
+
+namespace clp::ffi::ir_stream::search::test {
+namespace {
+using JsonPair = std::pair<nlohmann::json, nlohmann::json>;
+
+/**
+ * Class that implements `clp::ffi::ir_stream::IrUnitHandlerInterface` for testing purpose.
```
kirkrodrigues, line 37:
```suggestion
 * Implementation of `clp::ffi::ir_stream::IrUnitHandlerInterface` for testing purposes.
```

components/core/src/clp/ffi/ir_stream/search/test/test_deserializer_integration.cpp
```diff
@@ -0,0 +1,316 @@
...
+private:
+    std::vector<KeyValuePairLogEvent> m_deserialized_log_events;
+    bool m_is_complete{false};
+};
+
+/**
+ * Serializes a vector of JSON object pairs into a key-value pair IR stream.
+ * @tparam encoded_variable_t
+ * @param json_pairs A vector of JSON pairs. The objects in the pair are serialized as
+ * auto-generated kv-pairs and user-generated kv-pairs, respectively.
```
kirkrodrigues, line 82:
```suggestion
 * Serializes a vector of JSON object pairs into a key-value pair IR stream, where each pair is
 * treated as containing an object with auto-generated KV pairs and an object with user-generated KV
 * pairs.
 * @tparam encoded_variable_t
 * @param json_pairs
```

components/core/src/clp/ffi/ir_stream/search/test/test_deserializer_integration.cpp
```diff
@@ -0,0 +1,316 @@
...
+ */
+template <typename encoded_variable_t>
+[[nodiscard]] auto serialize_json_pairs_into_kv_pair_ir_stream(
+        std::vector<JsonPair> const& json_pairs
+) -> std::vector<int8_t>;
+
+/**
+ * Serializes a vector of JSON object pairs into a string.
+ * @param json_pairs
+ * @return A string representing the serialized JSON pairs.
```
kirkrodrigues, line 93:
```suggestion
 * Serializes a vector of JSON object pairs into a string, where each pair represents the
 * auto-generated and user-generated KV pairs in a log event.
 * @param json_pairs
 * @return The string.
```

components/core/src/clp/ffi/ir_stream/search/test/test_deserializer_integration.cpp
```diff
@@ -0,0 +1,316 @@
...
+        "[ffi][ir_stream][search][QueryHandler]",
+        ir::four_byte_encoded_variable_t,
+        ir::eight_byte_encoded_variable_t
+) {
+    constexpr std::string_view cIntKey{"int_key"};
+    constexpr std::string_view cStrKey{"str_key"};
+    constexpr std::string_view cObjKey{"obj_key"};
+    constexpr std::string_view cUnresolvableKey{"unresolvable_key"};
+    constexpr int cIntBase{INT16_MAX};
+    constexpr std::string_view cTestSubStr{"test"};
```
kirkrodrigues, line 141: * Why substring instead of string?
* Substring is usually one word, right? So then it should be `cTestSubstr`.

components/core/src/clp/ffi/ir_stream/search/test/test_deserializer_integration.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+    nlohmann::json const base_obj_0
+            = {{cIntKey, cIntBase - 1}, {cStrKey, fmt::format("VarStr:{}", cTestSubStr)}};
+    nlohmann::json const base_obj_1 = {
+            {cIntKey, cIntBase + 1},
+            {cStrKey,
+             fmt::format("This is a ClpStr: {}; And it contains a var={}", cTestSubStr, cIntBase)}
```
kirkrodrigues, line 148:
```suggestion
             fmt::format("This is a {} ClpStr; And it contains a var={}", cTestSubStr, cIntBase)}
```

Otherwise it looks like it's saying "test" is the ClpSttr.

components/core/src/clp/ffi/ir_stream/search/test/test_deserializer_integration.cpp
```diff
@@ -0,0 +1,316 @@
...
+    JsonPair const json_pair_0{base_obj_0, base_obj_1};
+    JsonPair const json_pair_1{base_obj_1, base_obj_0};
+    JsonPair const json_pair_2{nested_obj_0, nested_obj_1};
+    JsonPair const json_pair_3{nested_obj_1, nested_obj_0};
+
+    std::vector<JsonPair> const json_pairs_to_serialize{
+            {base_obj_0, base_obj_1},
+            {base_obj_1, base_obj_0},
+            {nested_obj_0, nested_obj_1},
+            {nested_obj_1, nested_obj_0}
```
kirkrodrigues, line 162:
```suggestion
            json_pair_0,
            json_pair_1,
            json_pair_2,
            json_pair_3
```

components/core/src/clp/ffi/ir_stream/search/test/test_deserializer_integration.cpp
```diff
@@ -0,0 +1,316 @@
...
+                            cIntKey,
+                            cIntBase
+                    ),
+                    json_pairs_to_serialize
+            ),
+            std::make_pair(fmt::format("*.{}.*: *", cUnresolvableKey), std::vector<JsonPair>{})
+    );
+    CAPTURE(kql_query_str);
+
+    auto query_stream{std::istringstream{kql_query_str}};
```
kirkrodrigues, line 271:
```suggestion
    std::istringstream query_stream{kql_query_str};
```

kirkrodrigues (changes requested):

components/core/src/clp/ffi/ir_stream/search/test/test_deserializer_integration.cpp
```diff
@@ -0,0 +1,316 @@
...
+            {nested_obj_1, nested_obj_0}
+    };
+
+    auto const ir_stream_bytes{
+            serialize_json_pairs_into_kv_pair_ir_stream<TestType>(json_pairs_to_serialize)
+    };
+    CAPTURE(ir_stream_bytes.size());
+
+    auto const wildcard_str_query{fmt::format("*{}*", cTestSubStr)};
+    auto const [kql_query_str, expected_json_pairs] = GENERATE_COPY(
```
kirkrodrigues, line 171: Did you follow some process to come up with these test cases or were they picked at random?

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
kirkrodrigues: Continuing your example, this seems to work for me:

```cpp
    auto deserializer_without_type_deduction{
            Deserializer<IrUnitHandler, decltype(query_handler)>::create(reader, IrUnitHandler{}, std::move(query_handler)).value()
    };
```

Anyway, that does require putting query_handler into a variable first which is a little tedious, so I can see the utility of having a method to help the compiler deduce the type after it's been unwrapped from whatever outcome is doing.

> Sorry, I don't fully get this. They're non-member variables, and the definition ordering matches the declaration.

The file is currently ordered like this:

1. `class Deserializer {...}`
2. `Deserializer::create_generic` implementation
3. `make_deserializer` prototype
4. `make_deserializer` prototype
5. `Deserializer::deserialize_next_ir_unit` implementation
6. `make_deserializer` implementation
7. `make_deserializer` implementation

Why is 2 between 1 & 3 instead of between 4 & 5?

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
kirkrodrigues: Can you file an issue?

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
LinZhihao-723: - Yeah, the goal is to avoid explicitly listing template arguments inside `<>` to use the factory.
- Oh true that's my mistake.

components/core/src/clp/ffi/ir_stream/search/test/test_deserializer_integration.cpp
LinZhihao-723: - It's a substring since:
  - The actual string is not exactly `"test"` but a string that contains `"test"`
  - We wrap it with two wildcards to generate the string query for testing.
- True, true. I thought it's `sub-string` or `sub string`. mb

components/core/src/clp/ffi/ir_stream/search/test/test_deserializer_integration.cpp
LinZhihao-723: Not randomly picked. Based on the JSON pairs input, I created the following:
1. String search on non-wildcard columns only matches the flattened ones.
2. String search on wildcard columns matches all.
3. Integer comparison on non-wildcard columns only matches the flattened ones.
4. Integer comparison on wildcard columns also matches inner keys.
5. A combination of `AND` on auto-gen and user-gen columns (results should be equivalent to 3).
6. A combination of `OR` on auto-gen and user-gen columns (matches all)
7. A combination of `AND` with `NOT` on auto-gen and user-gen columns (results should be equivalent to 5)
8. Top-level `NOT` with `AND` on auto-gen and user-gen columns that matches all.
9. Unresolvable column that matches nothing.

Search features covered:
- Column resolutions on wildcards
- Column resolutions on auto-gen/user-gen namespaces
- Basic `AND`/`OR` logic
- Inverter logic (`NOT`) on filter
- Inverter logic (`NOT`) on the top
- Unresolvable columns

In the next push I will also include a simple case to ensure unresolvable columns are not invertible.

I understand that not all supported types (e.g., boolean and float) and filter operations are covered, but they're tested in query hander's unit tests.

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
LinZhihao-723: #901

taskfiles/lint.yaml
```diff
@@ -750,6 +753,9 @@ tasks:
           # yamllint enable rule:line-length
           FLAGS:
             - "--config-file '{{.ROOT_DIR}}/.clang-tidy'"
+            # The following line filter is added to ignore msgpack issue #1098:
+            # https://github.com/msgpack/msgpack-c/issues/1098
+            - "-line-filter \"[{\\\"name\\\":\\\"cpp11_zone.hpp\\\",\\\"lines\\\":[[197,197]]}]\""
```
kirkrodrigues, line 758:
```suggestion
            - >-
              --line-filter "[{\"name\":\"cpp11_zone.hpp\",\"lines\":[[197,197]]}]"
```

* So it's more readable.
* Probably a good idea to extract it into a variable rather than duplicating it, right?

taskfiles/lint.yaml
```diff
@@ -750,6 +753,9 @@ tasks:
           # yamllint enable rule:line-length
           FLAGS:
             - "--config-file '{{.ROOT_DIR}}/.clang-tidy'"
+            # The following line filter is added to ignore msgpack issue #1098:
+            # https://github.com/msgpack/msgpack-c/issues/1098
```
kirkrodrigues, line 757:
```suggestion
            # Ignore https://github.com/msgpack/msgpack-c/issues/1098
```

kirkrodrigues (approved):
For the PR title, how about:

```
feat(kv-ir): Integrate `QueryHandler` into the IR stream `Deserializer`.
```

PR #899, opened 2025-05-13
refactor(new-webui): Enhance server implementation with Fastify toolchain.

davemarco:
# Description

PR refactors existing server code to match fastify maintainers [demo](https://github.com/fastify/demo) application. If someone wants to learn fastify i also recommend this [video](https://www.youtube.com/watch?v=x3SG71Ut2tA&ab_channel=Platformatic)

Note this PR dosent add any features, and does not migrate the old webui to new format. It just sets up a new folder for future new webui code. Eventually the old webui code should be ported in a future PR. 

Some changes include the following:

Better modularity - previously each plugin maintained a separate connection to the database, now its easier
to import an application-wide shared database pool into any plugin.

Better package start commands - I had some trouble when testing the server PR with the current package commands. I decided to just copy their recommended commands.

Adds fastify helper repos to reduce code( environment var parser, sensible http request/reply templates, test boilerplate) - I removed our own environment parsing code and the templates should make writting route code easier.

Security - adds fastify rate limiter plugin.

Graceful termination - adds plugin to ensure graceful shutdown on exit. 

# Checklist

  breaking change.

# Validation performed

I tested commands and test runs. I have not yet retested the package, but will do so before merging.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new Fastify v2 server architecture with modular plugins for environment configuration, MySQL, MongoDB, rate limiting, and error handling.
  - Added example Fastify plugins and routes for demonstration purposes.
  - Enhanced server startup with improved logging and graceful shutdown support.

- **Bug Fixes**
  - Improved error handling for server errors and 404 responses, including rate limiting to mitigate abuse.

- **Refactor**
  - Streamlined database plugin usage to rely on pre-registered plugins.
  - Simplified and modernized server initialization and environment management.

- **Documentation**
  - Updated development guide to reflect new server commands and usage instructions.

- **Chores**
  - Updated configuration and environment files for consistency and best practices.
  - Revised test setup with new helpers for easier server testing.


components/log-viewer-webui/server/src/fastify-v2/plugins/external/env.ts
davemarco, line 1: @junhaoliao - it looks like the path variables are not being used in old webui code? Like they just come from settings instead? if so, should i remove path variables (i.e. log viewer directory etc... from here and env file)?

components/log-viewer-webui/server/src/fastify-v2/app.ts
junhaoliao, line 1: could you explain why we want to split a "v2" app from "v1"?

components/log-viewer-webui/server/src/fastify-v2/app.ts
```diff
@@ -0,0 +1,114 @@
...
+    // test app. Some decorators may not be exposed in production.
+    delete opts.skipOverride;
+
+    // Loads all external plugins. Registered first as application plugins might depend on them.
+    await fastify.register(fastifyAutoload, {
+        dir: path.join(import.meta.dirname, "plugins/external"),
+        options: {...opts},
+    });
+
+    /* eslint-disable no-warning-comments */
```
junhaoliao, line 39:
```suggestion
    // eslint-disable-next-line no-warning-comments
```

components/log-viewer-webui/server/src/utils/tap.ts
junhaoliao, line 1: would it be better to move it to `test/utils/tap.ts`?

junhaoliao, line 1: this does not need to be address in the current PR:
what do you think about the other testing frameworks such as jest, mocha, or vitest? both of them seem more popular than tap.

components/log-viewer-webui/server/.env
```diff
@@ -6,3 +6,5 @@ HOST=localhost
 PORT=3000
 CLP_DB_USER=clp-user
 CLP_DB_PASS=
+
+FASTIFY_CLOSE_GRACE_DELAY=1000
```
junhaoliao, line 10: can we configure this in the config json file since this is not credential related?

junhaoliao (changes requested):
thanks for cleaning the server up

i posted all my comments for a first iteration review

junhaoliao (changes requested):
for the PR title, would this be better?
```
refactor(new-webui): Enhance server implementation with Fastify plugins and toolchain.
```

components/log-viewer-webui/server/src/fastify-v2/plugins/external/mongoDb.ts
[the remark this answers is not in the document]
davemarco: I modified to match the external name. re this
> Also, I do think adding fastify as a prefix 

They are all from fastify at the moment, so i dont think we need that yet

components/log-viewer-webui/server/src/main.ts
```diff
@@ -2,34 +1,72 @@
...
+
+    closeWithGrace(
+        {delay: Number(process.env.FASTIFY_CLOSE_GRACE_DELAY || DEFAULT_FASTIFY_CLOSE_GRACE_DELAY)},
+        async ({err}) => {
+            if (err) {
+                app.log.error(err);
+            }
+
+            await app.close();
+        }
```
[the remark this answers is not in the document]
davemarco: based on the repo, it dosent do this. https://github.com/mcollina/close-with-grace
perhaps the function itself(close with grace) does the then, catch

components/log-viewer-webui/server/src/utils/tap.ts
davemarco: It might try and run tests on the tap file? maybe we add sort of setting to only test .test?

davemarco: I think the whole testing infra needs to redone. Like we probably need seperate unit tests and integration tests. Like the current test is really a toy, and dosent do anything. I think when we get there, we can change framework if we want

davemarco:
> for the PR title, would this be better?
> 
> ```
> refactor(new-webui): Enhance server implementation with Fastify plugins and toolchain.
> ```

refactor(new-webui): Enhance server implementation with Fastify toolchain.?

components/log-viewer-webui/server/src/utils/tap.ts
junhaoliao: > I think the whole testing infra needs to redone.

I agree

> I think when we get there, we can change framework if we want

sounds good

junhaoliao: @coderabbitai create an issue to track testing infra improvement in the new-webui

junhaoliao: ah i see your point. i think any code is considered test code only if they are in a test hook.

did you try moving it into `test/` though? if moving it doesn't break any tests, i prefer keeping it as `test/tap.ts` or `test/utils/tap.ts` since this is strictly test utils.

components/log-viewer-webui/server/src/main.ts
```diff
@@ -1,35 +1,72 @@
...
         process.exit(1);
     }
 };
 
-main().catch((e: unknown) => {
-    console.error(e);
-    process.exit(1);
-});
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
+init();
```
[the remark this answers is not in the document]
junhaoliao: i agree with the rabbit. Node.js 22 supports calling `await` in the global scope

components/log-viewer-webui/server/src/fastify-v2/plugins/external/env.ts
davemarco: I think there are already coming for settings.json, so i just deleted the env variables

components/log-viewer-webui/server/src/utils/tap.ts
davemarco: okay i moved test, and added a patten to prevent it trying to test utils.

components/log-viewer-webui/server/src/utils/tap.ts
```diff
@@ -0,0 +1,44 @@
+// Reference: https://github.com/fastify/demo/blob/main/test/helper.ts
+
+import path from "node:path";
+
+import {FastifyInstance} from "fastify";
+import {build as buildApplication} from "fastify-cli/helper.js";
+import {Test} from "tap";
+
+
+const AppPath = path.join(import.meta.dirname, "../app.ts");
```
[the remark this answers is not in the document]
davemarco: I think i just took from reference by accident

davemarco:
okay @junhaoliao - I just rebuilt the package, and tested and looks like nothing is broken. I am okay with it.

components/log-viewer-webui/server/src/main.ts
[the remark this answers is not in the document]
junhaoliao: > [In case one of them is emitted, it will call the given function. **If it is emitted again, it will terminate the process abruptly.**](https://github.com/mcollina/close-with-grace?tab=readme-ov-file#closewithgraceopts-fn-err-signal-manual-)

I think we should at least attempt logging any exception before the process is terminated abruptly.

components/log-viewer-webui/server/src/main.ts
```diff
@@ -1,35 +1,72 @@
...
+    // Only if the program is running in an interactive terminal.
+    if (process.stdout.isTTY) {
+        return {
+            level: "info",
+            transport: {
+                target: "pino-pretty",
+                options: {
+                    translateTime: "HH:MM:ss Z",
+                    ignore: "pid,hostname",
+                },
```
junhaoliao, line 29: any reason we don't want to use the default settings?

components/log-viewer-webui/server/src/fastify-v2/plugins/external/env.ts
```diff
@@ -0,0 +1,100 @@
...
+        // Security
+        RATE_LIMIT_MAX: {
+            type: "number",
+            default: 100,
+        },
+    },
+};
+
+export const autoConfig = {
+
```
junhaoliao, line 59:
```suggestion
```

components/log-viewer-webui/server/src/fastify-v2/app.ts
junhaoliao: Is this pushed?

components/log-viewer-webui/server/src/main.ts
```diff
@@ -1,35 +1,72 @@
...
-        },
-        sqlDbPass: envVars.CLP_DB_PASS,
-        sqlDbUser: envVars.CLP_DB_USER,
-    });
+const init = async (): Promise<void> => {
+    // fp must be used to override default error handler.
+    app.register(fp(serviceApp));
+
+    closeWithGrace(
+        {delay: Number(process.env.FASTIFY_CLOSE_GRACE_DELAY || DEFAULT_FASTIFY_CLOSE_GRACE_DELAY)},
```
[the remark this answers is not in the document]
davemarco: I dont fully understand ur question but i dont believe we need to since using fastify logger.

components/log-viewer-webui/server/src/main.ts
davemarco: I think these options will only run in dev environment, not in production. I think the idea is too hide some information to make the logs less verbose, and make the timestamp more readable. If you want can remove, but I dont think impacts production logs.

components/log-viewer-webui/server/.env
```diff
@@ -1,6 +1,6 @@
-CLIENT_DIR=../client/dist
-STREAMS_DATA_DIR=../../../build/clp-package/var/data/streams
-LOG_VIEWER_DIR=../yscope-log-viewer/dist
+# Must always set to production
+# @see {@link https://www.youtube.com/watch?v=HMM7GJC5E2o}
```
junhaoliao, line 2: how about:
```suggestion
# Using NODE_ENV with non-"production" values is an antipattern due to inconsistent behavior in dependencies across environments.
# Reference: https://nodejs.org/en/learn/getting-started/nodejs-the-difference-between-development-and-production
```
1. this explains the rationale directly
2. i do know that YouTuber is a very famous Fastify maintainer in the community but i believe quoting a Node.js official blog is better than the YouTube video, because:
   1. people don't have to watch a 20-min video
   2. it would appear less opinionated

components/log-viewer-webui/server/src/main.ts
junhaoliao: For local debugging, I believe using local time, rather than UTC time, is more intuitive. At our current scale, i agree we can hide pid in addition to hostname.
```suggestion
                options: {
                    ignore: "pid,hostname",
                },
```

components/log-viewer-webui/server/src/main.ts
```diff
@@ -1,35 +1,71 @@
...
+ * Do not use NODE_ENV to determine what logger (or any env related feature) to use
+ *
+ * @return Logger options for Fastify.
+ * @see {@link https://www.youtube.com/watch?v=HMM7GJC5E2o}
+ */
+const getLoggerOptions = () => {
+    // Only if the program is running in an interactive terminal.
+    if (process.stdout.isTTY) {
+        return {
+            level: "info",
```
junhaoliao, line 23: Is this better? e.g., users wouldn't have to modify any VCS tracked file to change the log level
```suggestion
            level: process.env.LOG_LEVEL ?? "info",
```

components/log-viewer-webui/server/src/main.ts
```diff
@@ -1,35 +1,71 @@
...
+
+const DEFAULT_FASTIFY_CLOSE_GRACE_DELAY = 500;
+const DEFAULT_PORT = 3000;
+
+/**
+ * Do not use NODE_ENV to determine what logger (or any env related feature) to use
+ *
+ * @return Logger options for Fastify.
+ * @see {@link https://www.youtube.com/watch?v=HMM7GJC5E2o}
+ */
```
junhaoliao, line 18: we may not have to explain this again (to those who didn't know we were using an antipattern in determining the logger config previously):
```suggestion
/**
 * Generates logger configuration options based on the environment.
 *
 * @return Logger options for Fastify.
 */
```

components/log-viewer-webui/server/src/test/tap.ts
```diff
@@ -0,0 +1,47 @@
...
+import path from "node:path";
+
+import {FastifyInstance} from "fastify";
+import {build as buildApplication} from "fastify-cli/helper.js";
+import {Test} from "tap";
+
+
+// eslint-disable-next-line no-warning-comments
+// TODO: Setup testing for new webui fastify app in `fastify-v2/app.ts`. Testing will need
+// to be split into unit tests and integregation tests (require clp package running).
```
junhaoliao, line 12:
```suggestion
// to be split into unit tests and integration tests (require clp package running).
```

components/log-viewer-webui/server/src/test/tap.ts
```diff
@@ -0,0 +1,47 @@
...
+const appPath = path.join(import.meta.dirname, "../app.ts");
+
+/**
+ * Provides test configuration options.
+ *
+ * @return
+ */
+export const config = () => {
```
junhaoliao, line 20:
```suggestion
const config = () => {
```

components/log-viewer-webui/server/src/test/tap.ts
```diff
@@ -0,0 +1,47 @@
...
+    };
+};
+
+/**
+ * Automatically build and tear down fastify app.
+ *
+ * @param t The test instance.
+ * @return Fastify instance.
+ */
+export const build = async (t: Test): Promise<FastifyInstance> => {
```
junhaoliao, line 34:
```suggestion
const build = async (t: Test): Promise<FastifyInstance> => {
```
...
```
export {build};
```

components/log-viewer-webui/server/.env
junhaoliao, line 2: given this, do we still need the `if ("production" === process.env.NODE_ENV)` check in `src/routes/static.ts`?

components/log-viewer-webui/server/src/app.ts
```diff
@@ -15,30 +12,28 @@ import queryRoutes from "./routes/query.js";
...
-    sqlDbPass,
-}: AppProps): Promise<FastifyInstance> => {
-    const server = fastify(fastifyOptions);
-
+const FastifyV1App: FastifyPluginAsync<AppPluginOptions> = async (
+    fastify: FastifyInstance,
+    opts: AppPluginOptions
+) => {
+    const {sqlDbUser, sqlDbPass} = opts;
     if ("test" !== process.env.NODE_ENV) {
```
junhaoliao, line 35: if any value of `process.env.NODE_ENV` than `production` is an antipattern, is there a clean way to skip registration of certain plugins?

components/log-viewer-webui/client/package.json
```diff
@@ -7,8 +7,8 @@
     "build": "tsc -b && vite build",
     "lint:check": "eslint",
     "lint:fix": "eslint --fix",
-    "start": "vite",
-    "antd": "VITE_USE_ANTD_APP=true npm run start"
+    "dev": "vite",
```
junhaoliao, line 10:
```suggestion
    "dev": "vite",
```

components/log-viewer-webui/client/package.json
```diff
@@ -7,8 +7,8 @@
     "build": "tsc -b && vite build",
     "lint:check": "eslint",
     "lint:fix": "eslint --fix",
-    "start": "vite",
-    "antd": "VITE_USE_ANTD_APP=true npm run start"
+    "dev": "vite", 
+    "antd": "VITE_USE_ANTD_APP=true npm run dev"
```
junhaoliao, line 11: i prefer leaving this unchanged in this PR and submit another one to also update the scripts in `component/log-viewer-webui/package.json`

junhaoliao (changes requested):
i submitted my final batch of comments

components/log-viewer-webui/server/.env
davemarco: I think we can decide later how to remove it when we refactor old code. I dont think it's really an issue though just to remove and run two clients. Another options is to pass a fastify option to disable it. But i think can be decided later when old code refactored

components/log-viewer-webui/server/src/app.ts
davemarco: I think we definitely should not, but the current test requires this, so i left this as a future item when old code refactored. In general, i dont think we should be disabling any plugins. I think we should have integration test with the package running all the plugins, or just launch certain plugins individually and test them? Also unit tests with no plugins running. I think this can be decided more when we get to #903, but for simplicity for now i left it. A potential option in short term could also be to pass a fastify option instead of using NODE_ENV.

davemarco: I updated issue to remove these NODE_ENV references

components/log-viewer-webui/server/src/main.ts
davemarco: I also changed the default log level from silent to info for production. This is how we had it before.

components/log-viewer-webui/client/package.json
davemarco: @coderabbitai please make issue to update client scripts to use dev instead of start

components/log-viewer-webui/server/.env
davemarco: I added commennt to testing issue.

davemarco (commented):
responded to ur comments

junhaoliao (approved):

components/log-viewer-webui/server/settings.json
```diff
@@ -8,9 +8,9 @@
     "MongoDbName": "clp-query-results",
     "MongoDbStreamFilesCollectionName": "stream-files",
 
-    "ClientDir": "../client/dist",
-    "LogViewerDir": "../yscope-log-viewer/dist",
-    "StreamFilesDir": "../../../build/clp-package/var/data/streams",
+    "ClientDir": "../../client/dist",
+    "LogViewerDir": "../../yscope-log-viewer/dist",
+    "StreamFilesDir": "../../../../build/clp-package/var/data/streams",
```
junhaoliao, line 13: It seems we still need to add one more level up (i.e., an additional `../`) per path to get the paths resolved correctly.

Also, the `.` current directory is assumed to be where `settings.json` located.

I believe we should change `rootDirname` -> `rootDirname = path.resolve(dirname, "../../../..")` in `src/routes/static.ts`, given where the `static.js` is in `dist` now.

components/log-viewer-webui/server/src/fastify-v2/app.ts
```diff
@@ -0,0 +1,114 @@
...
+    FastifyInstance,
+    FastifyPluginOptions,
+} from "fastify";
+
+import FastifyV1App from "../app.js";
+
+
+const INTERNAL_SERVER_ERROR_CODE = 500;
+const RATE_LIMIT_MAX_REQUESTS = 3;
+const RATE_LIMIT_TIME_WINDOW_MS = 500;
```
junhaoliao, line 16: those values should be configurable via the environment variables

components/log-viewer-webui/server/src/fastify-v2/plugins/external/env.ts
```diff
@@ -0,0 +1,75 @@
...
+declare module "fastify" {
+    export interface FastifyInstance {
+        config: {
+            PORT: number;
+            HOST: string;
+            CLP_DB_USER: string;
+            CLP_DB_PASS: string;
+
+            RATE_LIMIT_MAX: number;
```
junhaoliao, line 13: we should also add `RATE_LIMIT_MAX` to `.env`

components/log-viewer-webui/server/src/fastify-v2/plugins/external/env.ts
```diff
@@ -0,0 +1,75 @@
...
+            default: "clp-user",
+        },
+        CLP_DB_PASS: {
+            type: "string",
+        },
+
+        // Security
+        RATE_LIMIT_MAX: {
+            type: "number",
+            default: 100,
```
junhaoliao, line 53: serving 100 requests in a minute is too few, considering we also serve the frontend with the Fastify server. each page load of the ingest page is 11 requests, not to mention the page auto fetches the ingest stats every 10 seconds, which yields an additional (60 seconds / 10 seconds * 3 requests) = 18 requests per minute per viewer.

we should extend this to something higher probably like 1000

Issue #907, opened 2025-05-15
bug(clp-s): Handle pruning when dynamically expanding wildcards during evaluation.


gibber9809:
### Bug

Evaluation code generally expects that each column in the query matches *at least one* column in a schema. For pure wildcard columns though this isn't necessarily true -- it can be the case that upon dynamically expanding a wildcard we find that no column in the schema matches the wildcard.

During schema matching we currently only check that at least one column with a type matching the wildcard column exists somewhere in the archive (irrespective of namespace and subtree type) and don't bother checking for matching types on a per-schema basis. I.e. we don't consider pure wildcard columns in `intersect_and_sub_expr`.

The current behaviour during evaluation is to return `false` from the evaluation of a wildcard filter that matches no columns, but this leads to incorrect evaluation when the wildcard filter is inverted.

For example the query: `NOT *: 0`
against the dataset
```
{"a": "b"}
{"c": 0}
```

will yield the result `{"a": "b"}`. This result is unexpected since our query semantics involve evaluating against columns with matching type but the returned record has no such column.

The simplest solution is to update our evaluation code in `QueryRunner` to use `True`, `False`, and `Pruned` like we do in kv-ir evaluation so that dynamic wildcard expansion can result in dynamic pruning of the AST.

### CLP version

47ff53ae931fafcd85fd42fb4b42f646a28d1fc2

### Environment

ubuntu 22.04 container

### Reproduction steps

See issue.

PR #908, opened 2025-05-15
feat(clp-s): Add support for searching against MPT subtrees with specific types.

gibber9809:
# Description

This PR adds support to the clp-s AST for specifying search against specific subtrees of the MPT. The immediate use-case is to allow search on the `log_event_idx` column in the metadata subtree.

Within `ColumnDescriptor` we now have a `subtree_type` string property wrapped in an `optional`. The associated search semantics are:
1. If `subtree_type` is not set then resolve the column against every subtree with matching namespace except the `NodeType::Metadata` subtree
2. If `subtree_type` is set then resolve only against the subtree with matching namespace and type

The current supported mappings for `subtree_type` are `"object" -> `NodeType::Object` and `"metadata"` -> `NodeType::Metadata`.

We modify `SchemaMatch` to implement these semantics during column resolution and `QueryRunner` to make dynamically expanded wildcards respect `subtree_type`. Note that this dynamic expansion runs into some existing limitations with how we expand wildcards: see #907.

Finally we add some search tests which check that we search the metadata subtree iff `subtree_type` is `metadata`.

# Checklist

  breaking change.

# Validation performed

* Added test for search against metadata subtree

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying and resolving columns against multiple subtree types, including metadata, in search queries.
  - Introduced new constants for subtree types to improve clarity in configuration and output.

- **Improvements**
  - Enhanced the detail and structure of column descriptor output for better debugging and readability.
  - Unified and streamlined internal management of subtree identifiers for improved consistency.
  - Improved safety and const-correctness in node access during schema operations.

- **Bug Fixes**
  - Prevented wildcard filters from being incorrectly applied to metadata columns unless explicitly intended.

- **Tests**
  - Added new tests for metadata-aware queries and direct AST-based search validation.
  - Extended existing tests to verify correct handling of subtree type properties in parsed queries.


components/core/src/clp_s/search/ast/ColumnDescriptor.hpp
```diff
@@ -280,6 +282,12 @@ class ColumnDescriptor : public Literal {
         m_namespace = descriptor_namespace;
     }
 
+    auto get_subtree_type() const -> std::optional<std::string> const& { return m_subtree_type; }
```
wraymo, line 285: Can we get rid of `get_node_type_for_subtree_type` and have a method `get_subtree_node_type` instead?

components/core/tests/test-clp_s-search.cpp
```diff
@@ -56,6 +67,33 @@ auto get_test_input_local_path() -> std::string {
...
 }
 
+auto create_first_record_match_metadata_query() -> std::shared_ptr<clp_s::search::ast::Expression> {
+    auto zero_literal = clp_s::search::ast::Integral::create_from_int(0);
+    auto one_literal = clp_s::search::ast::Integral::create_from_int(1);
+    auto non_matching_idx = clp_s::search::ast::ColumnDescriptor::create_from_escaped_tokens(
+            {std::string{clp_s::constants::cLogEventIdxName}},
+            std::string{clp_s::constants::cDefaultNamespace}
+    );
+    auto object_subtree_non_matching_idx = non_matching_idx->copy();
```
wraymo, line 77: Maybe change idx to column? I personally prefer something like `column`, `column_with_object_subtree_type`, `column_with_metadata_subtree_type`

components/core/tests/test-clp_s-search.cpp
```diff
@@ -56,6 +67,33 @@ auto get_test_input_local_path() -> std::string {
     return (tests_dir / get_test_input_path_relative_to_tests_dir()).string();
 }
 
+auto create_first_record_match_metadata_query() -> std::shared_ptr<clp_s::search::ast::Expression> {
```
wraymo, line 70: I didn't get the point of this test. Seems that we create an OR expression with three operands. And one operand matches a record. But if so, what's the use of other two operands? Or can we create another two test cases with the other two operands

components/core/tests/test-clp_s-search.cpp
```diff
@@ -56,6 +67,33 @@ auto get_test_input_local_path() -> std::string {
...
+    auto zero_literal = clp_s::search::ast::Integral::create_from_int(0);
+    auto one_literal = clp_s::search::ast::Integral::create_from_int(1);
```
wraymo, line 72: Why do we need to create two literals?

wraymo (commented):
A good PR overall! Just have a few questions about `set_subtree_type` and the unit tests.

components/core/tests/test-clp_s-search.cpp
gibber9809: It tests that column resolution is treating subtree type as expected in all cases. Each column has name "log_event_idx" and different subtree type. The column that matches against 0 has subtree type "metadata" and both columns matching against 1 have "object" and no subtree type respectively. If column resolution respects subtree type correctly then only the filter against 0 should get resolved and the only matching record should be record 0, but if either of the two other columns are erroneously resolved then the results will incorrectly also contain record 1.

components/core/tests/test-clp_s-search.cpp
gibber9809: See other comment.

components/core/src/clp_s/search/ast/ColumnDescriptor.hpp
```diff
@@ -280,6 +282,12 @@ class ColumnDescriptor : public Literal {
...
+
+    auto set_subtree_type(std::optional<std::string> subtree_type) {
```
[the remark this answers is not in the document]
gibber9809: We discussed it in the thread last week, but you can also just take a look at [this part](https://github.com/gibber9809/clp/blob/b3faef9ba1c876dcaf717d7ad6f8ba792bab364e/components/core/src/clp_s/search/EvaluateMetadataFilters.cpp#L107C68-L107C88) of the follow-up PR.

It allows us to transform queries against the metadata range index into queries against the `log_event_idx` column in that transformation pass.

components/core/tests/test-clp_s-search.cpp
wraymo: Oh, I see your point. It's a compact way to test three filters in one expression

Issue #905, opened 2025-05-15
Add support for projection in `clp::ffi::KeyValuePairLogEvent`.


LinZhihao-723:
### Request

The current `clp::ffi::KeyValuePairLogEvent` doesn't support projection. Projection to schema-tree node ID resolution has been added as a part of `QueryHandler`, but we don't have the native support to use this projection resolution in log event object to "select" projected node-ID-value pairs.

### Possible implementation

Maybe maintain a bitmap to track all projected leaf nodes. Need more thoughts before we come up a formal implementation plan.

PR #909, opened 2025-05-16
feat(clp-s): Add support for searching KV-pair IR streams through the clp-s cli.

LinZhihao-723:
# Description

## References

* This PR depends on #890.
* This PR is one of the PR series #840.

## Overview

This PR adds cli support for searching KV-IR streams through the `clp-s` executable. It is built on top of the current `clp-s` search utilities, assuming that the input path that contains `.clp.zst` will be considered an IR file.

As the first PR to support KV-IR search, the current implementation has the following limitations:
* The current IR search only supports for `stdout` output. The current implementation can be extended to support other output methods when needed.
* The current IR search doesn't support projection. This will be resolved by implementing #905.
* The current IR search doesn't support any `count` operations.
* The current IR search doesn't support timestamp filtering, as we don't have a way to specify an authoritative timestamp in a KV-IR stream yet.

Some behaviours to notice:
* The search result is formatted a JSON string that has the following schema: `{"auto_generated_kv_pairs":{xxx},"user_generated_kv_pairs":{xxx}}`.
* Input-wise, the current implementation supports both reading from a file in local FS and reading through the network reader with proper authentication.
* When failing to read the input as a KV-IR stream, even if the input path contains `.clp.zst`, we will fall back to archive search just in case the file path of the archive unintentionally contains the IR extension.
* We don't fail on truncated IR streams since the searched file might be an ongoing IR stream. This provides compatibility for real-time search.

# Checklist

  breaking change.

# Validation performed

* [x] Ensure searching an IR file from the local file system works as expected.
* [x] Ensure searching an IR file served by a local-host server works as expected.
* [x] Ensure searching an archive with `.clp.zst` in its path works as expected (will actually perform archive search).
* [x] Esnrue searching truncated IR streams only returns warnings, but not errors.

## Summary by CodeRabbit

- **New Features**
  - Added support for searching key-value IR streams, including improved error handling and output to standard output.
  - Enhanced input detection to automatically perform IR stream searches when applicable.

- **Bug Fixes**
  - Improved handling of various error scenarios during IR stream processing, with clear warnings and fallbacks.

- **Chores**
  - Updated build configuration to include new search-related files and dependencies.


components/core/src/clp_s/kv_ir_search.cpp
```diff
@@ -0,0 +1,328 @@
...
+    [[nodiscard]] static auto handle_schema_tree_node_insertion(
+            [[maybe_unused]] bool is_auto_generated,
+            [[maybe_unused]] clp::ffi::SchemaTree::NodeLocator schema_tree_node_locator,
+            [[maybe_unused]] std::shared_ptr<clp::ffi::SchemaTree const> const& schema_tree
+    ) -> IRErrorCode {
+        return IRErrorCode::IRErrorCode_Success;
+    }
+
+    [[nodiscard]] auto handle_end_of_stream() -> IRErrorCode {
+        m_is_complete = true;
```
gibber9809, line 96:
```suggestion
```
I think we can get rid of this variable since it isn't used/we don't have a getter for it.

components/core/src/clp_s/kv_ir_search.cpp
```diff
@@ -0,0 +1,328 @@
...
+        return IRErrorCode::IRErrorCode_Success;
+    }
+
+private:
+    // Constructor
+    IrUnitHandler() = default;
+
+    // Variable
+    bool m_is_complete{false};
```
gibber9809, line 105:
```suggestion
```

components/core/src/clp_s/clp-s.cpp
```diff
@@ -358,9 +363,51 @@ int main(int argc, char const* argv[]) {
...
+                        error.message()
+                );
+                if (KvIrSearchError{KvIrSearchErrorEnum::ProjectionSupportNotImplemented} == error
+                    || KvIrSearchError{KvIrSearchErrorEnum::UnsupportedOutputHandlerType} == error
+                    || KvIrSearchError{KvIrSearchErrorEnum::CountSupportNotImplemented} == error)
+                {
+                    // These errors are treated as non-fatal since it's because we currently don't
+                    // have supports for these features.
+                    continue;
+                }
```
gibber9809, line 399: I think this leads to us always not searching an archive if it happens to have this extension and the search uses any of these more advanced features.

I know that that isn't really an expected case, but could we go into the fallback of attempting archive search when encountering these errors to handle this edge case?

gibber9809 (changes requested):
Just a few minor comments, but change mostly looks good.

components/core/src/clp_s/clp-s.cpp
LinZhihao-723: Changed. Added a log on this case.

gibber9809 (approved):
LGTM. Maybe for the title
`feat(clp-s): Add support for searching KV-pair IR streams through the clp-s cli.`

Looking through our past commit messages seems like we usually call it "KV-pair IR format" or "KV-pair IR stream" instead of the shorter "KV-IR".

LinZhihao-723:
> LGTM. Maybe for the title `feat(clp-s): Add support for searching KV-pair IR streams through the clp-s cli.`
> 
> Looking through our past commit messages seems like we usually call it "KV-pair IR format" or "KV-pair IR stream" instead of the shorter "KV-IR".

Sure. Let's keep using "KV-pair IR streams" for this PR to be explicit. In the future PRs, we might use `KV-IR` in short since it will be the official short form (not IRv2, lol).

PR #923, opened 2025-05-26
feat(datasets): Add storage-type column to datasets table, and document datasets table.

kirkrodrigues:
# Description

As the title says. Adding the storage type column allows us to separate the storage type from the directory---the alternative being to use a URI containing both the type and directory (e.g., `s3://abc`), which would be harder for external tools to use.

# Checklist

  breaking change.

# Validation performed

* Validated that compression & search work correctly with the clp-text and clp-json packages.
* Validated that when using the clp-json package, the datasets table is created, with the added column.
* Validated that the new docs render correctly.

## Summary by CodeRabbit

- **New Features**
	- Added a new column for archive storage type to the datasets table, enhancing dataset metadata.
- **Documentation**
	- Introduced a new section detailing the metadata database and datasets table schema.
	- Updated the developer guide index to include new design documentation topics.


docs/src/dev-guide/index.md
```diff
@@ -80,6 +80,8 @@ tooling-gh-workflows
 :hidden:
 
 design-project-structure
+
```
haiqi96, line 83: why an extra empty line?

haiqi96: I see, we can keep the line.

PR #925, opened 2025-05-26
fear: Remove ANTLR generator dependencies.

anlowee:
# Description

This PR is to remove the ANTLR's generator dependencies which needs Java. Instead, we directly put genearted code into the codebase.

# Checklist

  breaking change.

# Validation performed


components/core/src/clp_s/search/kql/CMakeLists.txt
```diff
@@ -1,19 +1,32 @@
-ANTLR_TARGET(
-        KqlParser
-        Kql.g4
-        LEXER PARSER VISITOR
-        PACKAGE kql
+#ANTLR_TARGET(
+#        KqlParser
+#        Kql.g4
+#        LEXER PARSER VISITOR
+#        PACKAGE kql
```
gibber9809, line 5: I think we may want to change this to be namespaced just to avoid namespace collisions when we pull clp-s into other projects.

This PACKAGE argument is supposed to control the namespace of the generated code. Not sure if it supports nested namespaces, but maybe we can try changing this to `clp_s::search::kql` and updating the generated code?

components/core/src/clp_s/search/sql/CMakeLists.txt
```diff
@@ -1,19 +1,32 @@
-ANTLR_TARGET(
-        SqlParser
-        Sql.g4
-        LEXER PARSER VISITOR
-        PACKAGE sql
+#ANTLR_TARGET(
+#        SqlParser
+#        Sql.g4
+#        LEXER PARSER VISITOR
+#        PACKAGE sql
```
gibber9809, line 5: Same as my comment for kql -- maybe we can try changing this to `clp_s::search::sql` and regenerating the code?

components/core/src/clp_s/search/kql/CMakeLists.txt
gibber9809: I would also suggest changing the library name from kql to something namespaced, but I'm doing that in #924, so no need to do that part here.

gibber9809 (changes requested):
Nice work!

Just have a suggestion about the namespace for the generated code.

components/core/src/clp_s/search/kql/antlr_generated/antlr4-runtime.h
```diff
@@ -0,0 +1,168 @@
...
+#include "Token.h"
+#include "TokenFactory.h"
+#include "TokenSource.h"
+#include "TokenStream.h"
+#include "TokenStreamRewriter.h"
+#include "UnbufferedCharStream.h"
+#include "UnbufferedTokenStream.h"
+#include "Version.h"
+#include "Vocabulary.h"
+#include "Vocabulary.h"
```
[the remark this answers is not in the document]
anlowee: antlr4-runtime.h is from the official antlr repo, so do not modify it

components/core/src/clp_s/search/sql/CMakeLists.txt
```diff
@@ -1,19 +1,32 @@
...
+set(ANTLR_GENERATED_SQL_CODE
+		${ANTLR_GENERATED_SQL_CODE_DIR}/SqlBaseVisitor.cpp
+		${ANTLR_GENERATED_SQL_CODE_DIR}/SqlBaseVisitor.h
+		${ANTLR_GENERATED_SQL_CODE_DIR}/SqlLexer.cpp
+		${ANTLR_GENERATED_SQL_CODE_DIR}/SqlLexer.h
+		${ANTLR_GENERATED_SQL_CODE_DIR}/SqlParser.cpp
+		${ANTLR_GENERATED_SQL_CODE_DIR}/SqlParser.h
+		${ANTLR_GENERATED_SQL_CODE_DIR}/SqlVisitor.cpp
+		${ANTLR_GENERATED_SQL_CODE_DIR}/SqlVisitor.h
 )
```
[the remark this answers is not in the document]
anlowee: I prefer add them one by one, as we do in other CMakes.

components/core/src/clp_s/search/kql/antlr_generated/KqlLexer.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+#include "KqlLexer.h"
+
+
+using namespace antlr4;
+
+using namespace kql;
+
+
+using namespace antlr4;
```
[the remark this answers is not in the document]
anlowee: do not modify any for the auto generated code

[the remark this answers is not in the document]
anlowee: I think this is a codegen wart.

The first one using namespace antlr4; comes from a general template header section.

Another using namespace antlr4; is added after using namespace your.custom.package; (e.g., clp_s::search::sql), to ensure ANTLR symbols are in scope again — even though they never left.

components/core/src/clp_s/search/kql/CMakeLists.txt
anlowee: Fixed

components/core/src/clp_s/search/sql/antlr_generated/SqlLexer.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+#include "SqlLexer.h"
+
+
+using namespace antlr4;
+
+using namespace clp_s::search::sql;
+
+
+using namespace antlr4;
```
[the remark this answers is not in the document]
anlowee: as I said, this is auto generated by ANTLR as a codegen wart. Don't touch it.

taskfile.yaml
```diff
@@ -86,6 +86,58 @@
...
+        git status --porcelain \
+        {{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/kql/antlr_generated \
+        {{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/sql/antlr_generated
+      - |
+        git status --porcelain \
+        {{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/kql/antlr_generated \
+        {{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/sql/antlr_generated \
+        | grep . > /dev/null \
+        && exit 1 \
+        || exit 0
```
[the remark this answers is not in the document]
anlowee: but git diff cannot detect the untracked files, what if we want to generate new files?

.github/workflows/clp-s-antlr-generation.yaml
```diff
@@ -0,0 +1,42 @@
...
+    name: "antlr-generation-check"
+    strategy:
+      matrix:
+        os: ["macos-latest", "ubuntu-latest"]
+    runs-on: "${{matrix.os}}"
+    steps:
+      - uses: "actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683"
+        with:
+          submodules: "recursive"
+
```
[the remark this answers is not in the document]
anlowee: I copied this from other workflow files, so don't blame me.

components/core/src/clp_s/search/kql/antlr_generated/Kql.interp
wraymo, line 1: Do we need files like `*.interp`, `*.tokens`?

anlowee: I don't know, but it seems CMake also generates them

taskfile.yaml
```diff
@@ -86,6 +92,53 @@ tasks:
...
+          -Xexact-output-dir \
+          clp_s/search/kql/Kql.g4
+      - |
+        java -jar {{.G_ANTLR_JAR_PATH}} \
+          -Dlanguage=Cpp \
+          -no-listener \
+          -visitor \
+          -package clp_s::search::sql \
+          -o clp_s/search/sql/antlr_generated \
+          -Xexact-output-dir \
```
wraymo, line 126: where does it come from

components/core/src/clp_s/search/kql/antlr_generated/Kql.interp
anlowee: removed

wraymo (approved):

components/core/src/clp_s/search/kql/CMakeLists.txt
```diff
@@ -1,24 +1,33 @@
-ANTLR_TARGET(
-        KqlParser
-        Kql.g4
-        LEXER PARSER VISITOR
-        PACKAGE kql
+set(ANTLR_GENERATED_KQL_CODE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/antlr_generated)
```
kirkrodrigues: How about ANTLR_GENERATED_SOURCE_DIR (to be consistent with how CMake names source directories)?

gibber9809 (dismissed):

taskfiles/deps/main.yaml
```diff
@@ -84,39 +84,62 @@ tasks:
           TARBALL_SHA256: "987ce98f02eefbaf930d6e38ab16aa05737234d7afbab2d5c4ea7adbe50c28ed"
           TARBALL_URL: "https://github.com/abseil/abseil-cpp/archive/refs/tags/20230802.1.tar.gz"
 
-  antlr:
```
kirkrodrigues, line 87: * As discussed offline, let's add this `antlr` task back and use it in `clp-s-generate-antlr`, instead of doing the same work inside `clp-s-generate-antlr`.
* Let's rename the task to `antlr-jar` so differentiate it from `antlr-runtime`.

.github/workflows/clp-s-antlr-generation.yaml
```diff
@@ -0,0 +1,42 @@
+name: "clp-s-antlr-generation"
+
+on:
+  pull_request:
+  push:
+  schedule:
+    # Run daily at 00:15 UTC (the 15 is to avoid periods of high load)
+    - cron: "15 0 * * *"
```
kirkrodrigues: We don't need to run this workflow periodically, right? The dependencies related to this workflow are fixed, so unless something unexpected happens, nothing should change if we run the workflow at a different time.

.github/workflows/clp-s-antlr-generation.yaml
```diff
@@ -0,0 +1,42 @@
...
+        os: ["macos-latest", "ubuntu-latest"]
+    runs-on: "${{matrix.os}}"
+    steps:
+      - uses: "actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683"
+        with:
+          submodules: "recursive"
+
+      - uses: "actions/setup-python@42375524e23c412d93fb67b49958b491fce71c38"
+        with:
+          python-version: "3.11"
```
kirkrodrigues: Why do we need Python in this workflow?

taskfile.yaml
```diff
@@ -86,6 +92,59 @@ tasks:
           FLAVOUR: "text"
           STORAGE_ENGINE: "clp"
 
+  clp-s-generate-antlr:
```
kirkrodrigues: How about `clp-s-generate-parsers`?

kirkrodrigues: Let's add this task as a dependency of `core-generate` so that when the package is built, it uses the latest generated code.

taskfile.yaml
```diff
@@ -86,6 +92,59 @@ tasks:
...
+    preconditions:
+      - sh: "which curl >/dev/null"
+        msg: "curl is required to download the ANTLR JAR"
+      - sh: "which java >/dev/null"
+        msg: "Java is required for ANTLR code generation"
```
kirkrodrigues: Since we're going to move back to using the `antlr` task, let's remove these preconditions. I'm also not sure how helpful they are since the shell would already report that the commands are missing, right? (And we document what dependencies are required to build CLP anyway.)

taskfile.yaml
```diff
@@ -86,6 +92,59 @@ tasks:
...
+    sources:
+      - "{{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/kql/Kql.g4"
+      - "{{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/sql/Sql.g4"
+    dir: "{{.G_CORE_COMPONENT_DIR}}/src/"
+    generates:
```
kirkrodrigues: Because of the way task tracks generated files, this may be unreliable. Instead, can you switch to using checksum tasks like we do for `webui-node-modules`? Let me know if you're confused.

taskfile.yaml
```diff
@@ -86,6 +92,59 @@ tasks:
...
+      - "rm -rf {{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/kql/antlr_generated/*"
+      - "rm -rf {{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/sql/antlr_generated/*"
+      - "curl -o {{.G_ANTLR_JAR_PATH}} {{.G_ANTLR_DOWNLOAD_URL}}"
+      - |
+        java -jar {{.G_ANTLR_JAR_PATH}} \
+          -Dlanguage=Cpp \
+          -no-listener \
+          -visitor \
+          -package clp_s::search::kql \
+          -o clp_s/search/kql/antlr_generated \
```
kirkrodrigues: I talked to Devin and we agreed that it would be better to put the generated code in its
own namespace---`clp_s::search::kql::generated`---and to call the directory `generated`, so that it matches the namespace name.

taskfile.yaml
```diff
@@ -86,6 +92,59 @@ tasks:
...
+    deps:
+      - "clp-s-generate-antlr"
+    cmds:
+      - |
+        git status --porcelain \
+        {{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/kql/antlr_generated \
+        {{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/sql/antlr_generated \
+        | grep . > /dev/null \
+        && exit 1 \
+        || exit 0
```
kirkrodrigues: I don't really think we need a task to do this since it's really only going to be used in the workflow. So could we just move this into the GH workflow?

kirkrodrigues: I think this indentation is more clear:

```suggestion
      - |-
        git status --porcelain \
          {{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/kql/antlr_generated \
          {{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/sql/antlr_generated \
          | grep . > /dev/null \
            && exit 1 \
            || exit 0
```

components/core/src/clp_s/search/kql/CMakeLists.txt
```diff
@@ -1,25 +1,35 @@
...
+
+set(ANTLR_GENERATED_KQL_CODE
```
kirkrodrigues: How about `ANTLR_GENERATED_SOURCES` (to be consistent with the variable naming in other CMake files)?

components/core/src/clp_s/search/kql/CMakeLists.txt
```diff
@@ -1,25 +1,35 @@
...
 add_library(
         clp_s_search_kql
-        ../antlr_common/ErrorListener.hpp
-        ${ANTLR_KqlParser_CXX_OUTPUTS}
-        kql.cpp
-        kql.hpp
+        ${CMAKE_CURRENT_SOURCE_DIR}/../antlr_common/ErrorListener.hpp
+        ${ANTLR_GENERATED_KQL_CODE}
+        ${CMAKE_CURRENT_SOURCE_DIR}/kql.cpp
+        ${CMAKE_CURRENT_SOURCE_DIR}/kql.hpp
```
kirkrodrigues: Do we absolutely need to use `CMAKE_CURRENT_SOURCE_DIR` here?

kirkrodrigues: Lets alphabetize these.

components/core/src/clp_s/search/sql/CMakeLists.txt
kirkrodrigues, line 1: Same comments as those in the kql CMakeLists.txt

taskfile.yaml
```diff
@@ -35,6 +35,12 @@ vars:
   # Versions
   G_PACKAGE_VERSION: "0.2.0-dev"
 
+  # ANTLR codegen related
+  G_ANTLR_VERSION: "4.13.1"
```
kirkrodrigues: Note that I switched to 4.13.2 for the `antlr-runtime` task.

.github/workflows/clp-s-antlr-generation.yaml
```diff
@@ -0,0 +1,42 @@
+name: "clp-s-antlr-generation"
```
kirkrodrigues: How about `clp-s-generated-code-committed-check`?

.github/workflows/clp-s-antlr-generation.yaml
```diff
@@ -0,0 +1,42 @@
...
+  workflow_dispatch:
+
+concurrency:
+  group: "${{github.workflow}}-${{github.ref}}"
+  # Cancel in-progress jobs for efficiency
+  cancel-in-progress: true
+
+jobs:
+  antlr-generation-check:
+    name: "antlr-generation-check"
```
kirkrodrigues: How about `antlr-generated-code-committed-check`?

taskfile.yaml
```diff
@@ -86,6 +92,59 @@ tasks:
...
+          -no-listener \
+          -visitor \
+          -package clp_s::search::sql \
+          -o clp_s/search/sql/antlr_generated \
+          -Xexact-output-dir \
+          clp_s/search/sql/Sql.g4
+      - "rm -f {{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/kql/antlr_generated/*.interp"
+      - "rm -f {{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/kql/antlr_generated/*.tokens"
+      - "rm -f {{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/sql/antlr_generated/*.interp"
+      - "rm -f {{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/sql/antlr_generated/*.tokens"
```
kirkrodrigues:
```suggestion
      - for: ["kql", "sql"]
        cmd: |-
          java -jar {{.G_ANTLR_JAR_PATH}} \
            -Dlanguage=Cpp \
            -no-listener \
            -visitor \
            -package clp_s::search::{{.ITEM}} \
            -o clp_s/search/{{.ITEM}}/antlr_generated \
            -Xexact-output-dir \
            clp_s/search/{{.ITEM}}/*.g4 \
            && find "{{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/{{.ITEM}}/antlr_generated" \
              -type f -and -not \( -name "*.cpp" -o -name "*.h" \) \
              -exec rm {} \;
```

kirkrodrigues (changes requested):

components/core/src/clp_s/search/kql/CMakeLists.txt
anlowee: It's sugested by coderabbit if I remember correctly.

kirkrodrigues: And what was its reasoning?

components/core/src/clp_s/search/sql/sql.cpp
```diff
@@ -15,14 +15,13 @@
...
 using clp_s::search::antlr_common::ErrorListener;
-using sql::SqlBaseVisitor;
-using sql::SqlLexer;
-using sql::SqlParser;
 
 using clp_s::search::ast::EmptyExpr;
 using clp_s::search::ast::Expression;
 
 namespace clp_s::search::sql {
+using namespace generated;
```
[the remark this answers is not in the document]
anlowee: WDYT about this one and the previous one? @kirkrodrigues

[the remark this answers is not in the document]
kirkrodrigues: I agree with the rabbit.

components/core/src/clp_s/search/kql/CMakeLists.txt
anlowee: I guess no. Because previously when I tried building Velox it told me these generated files cannot be found, so I just added them though it turned out this didn't fix that issue. Yea I can remove them, shall I?

kirkrodrigues: Yeah, let's remove it to stay consistent with other CMakeLists.

.github/workflows/clp-s-generated-code-committed-check.yaml
```diff
@@ -0,0 +1,42 @@
...
+  workflow_dispatch:
+
+concurrency:
+  group: "${{github.workflow}}-${{github.ref}}"
+  # Cancel in-progress jobs for efficiency
+  cancel-in-progress: true
+
+jobs:
+  antlr-generated-code-committed-check:
+    name: "antlr-generated-code-committed-check"
```
kirkrodrigues:
```suggestion
  antlr-code-committed:
    name: "antlr-code-committed"
```

.github/workflows/clp-s-generated-code-committed-check.yaml
```diff
@@ -0,0 +1,42 @@
+name: "clp-s-generated-code-committed-check"
```
kirkrodrigues: On second thought these names become too long when displayed, lol. How about

```suggestion
name: "clp-s-generated-code-checks"
```

taskfile.yaml
```diff
@@ -91,6 +91,44 @@ tasks:
           FLAVOUR: "text"
           STORAGE_ENGINE: "clp"
 
+  clp-s-generate-parsers:
```
kirkrodrigues: Can we move this task to be just above the `core` task?

taskfile.yaml
```diff
@@ -91,6 +91,44 @@ tasks:
...
+    vars:
+      CHECKSUM_FILE: "{{.G_BUILD_DIR}}/{{.TASK}}.md5"
+      OUTPUT_DIR:
```
kirkrodrigues: Since this contains more than one output directory, how about `OUTPUT_DIRS`?

taskfile.yaml
```diff
@@ -91,6 +91,44 @@ tasks:
...
+        - "{{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/kql/generated"
+        - "{{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/sql/generated"
+    deps:
+      - "deps:antlr-jar"
+      - task: "utils:checksum:validate"
+        vars:
+          CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
+          INCLUDE_PATTERNS: ["{{.OUTPUT_DIR}}"]
```
kirkrodrigues: This won't pass the list as a list but will instead stringify the list. You'll notice that with the current code, when you run `clp-s-generate-parsers` repeatedly, it always seems to re-run. This means that it's not working properly.

To pass it as a list, change to:

```suggestion
          INCLUDE_PATTERNS:
            ref: ".OUTPUT_DIR"
```

taskfile.yaml
```diff
@@ -91,6 +91,44 @@ tasks:
...
+    generates: ["{{.CHECKSUM_FILE}}"]
+    sources:
+      - "{{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/kql/Kql.g4"
+      - "{{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/sql/Sql.g4"
+    dir: "{{.G_CORE_COMPONENT_DIR}}/src/"
+    cmds:
+      - for: ["kql", "sql"]
+        cmd: |-
+          rm -rf "{{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/{{.ITEM}}/generated/*" \
```
kirkrodrigues: Bash won't interpret `*` inside quotes, so we need to put them outside the quotes:

```suggestion
          rm -rf "{{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/{{.ITEM}}/generated/"* \
```

taskfile.yaml
```diff
@@ -91,6 +91,44 @@ tasks:
...
+            -no-listener \
+            -visitor \
+            -package clp_s::search::{{.ITEM}}::generated \
+            -o clp_s/search/{{.ITEM}}/generated \
+            -Xexact-output-dir \
+            clp_s/search/{{.ITEM}}/*.g4 \
+            && find "{{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/{{.ITEM}}/generated" \
+              -type f -and -not \( -name "*.cpp" -o -name "*.h" \) \
+              -exec rm {} \;
+      # This command must be last
```
kirkrodrigues: Need to put empty line before comment blocks.

```suggestion

      # This command must be last
```

taskfile.yaml
```diff
@@ -91,6 +91,44 @@ tasks:
...
+      - task: "utils:checksum:compute"
+        vars:
+          CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
+          INCLUDE_PATTERNS: ["{{.OUTPUT_DIR}}"]
```
kirkrodrigues:
```suggestion
          INCLUDE_PATTERNS:
            ref: ".OUTPUT_DIR"
```

taskfiles/deps/main.yaml
```diff
@@ -20,6 +20,10 @@ vars:
   # Library names
   G_FMT_LIB_NAME: "fmt"
 
+  # Antlr related
```
kirkrodrigues:
```suggestion
  # Antlr
```

kirkrodrigues (approved):
For the PR title, how about:

```
refactor(clp-s): Add ANTLR-generated parsers to codebase to eliminate need for code dependencies at build time.
```

PR #929, opened 2025-05-26
feat(clp-s): Add support for querying kv-pairs in the archive range index.

gibber9809:
# Description

This PR adds support for querying fields in the archive range index. Any column in the `$` namespace is treated as a query against the metadata range index. Filters from the `$` namespace are evaluated against the range index and either converted into filters on a range of `log_event_idx` or constant propagated away.

This logic has been implemented in the `EvaluateMetadataFilters` transformation pass. The pass has an iterative implementation that walks the AST to find and rewrite filters on columns in the `$` namespace and has another iterative implementation that resolves columns against metadata fields in each section of the range index. Expression evaluation is performed by `clp::ffi::ir_stream::search::evaluate_filter_against_literal_type_value_pair` -- since this evaluation function doesn't yet support array search filters against the archive range index also do not support array search.

Note that this PR enables search on fields in the archive range index but not projection.

# Checklist

  breaking change.

# Validation performed

* Added search test cases that test various aspects of column resolution and expression evaluation on the archive range index

## Summary by CodeRabbit

- **New Features**
  - Added support for filtering search results using metadata fields from the archive range index. Users can now query metadata fields (e.g., filename, file split number, archive creator ID) by prefixing keys with a `$` symbol.
- **Bug Fixes**
  - Enhanced search logic to skip unnecessary processing when no results match the metadata filters.
- **Documentation**
  - Updated user guide with instructions and examples for querying kv-pairs in the archive range index.
- **Tests**
  - Expanded test coverage for metadata range index filtering, including new queries and improved diagnostics.


components/core/src/clp_s/search/EvaluateMetadataFilters.cpp
```diff
@@ -0,0 +1,322 @@
...
+                case nlohmann::json::value_t::boolean:
+                    if (col->matches_type(ast::LiteralType::BooleanT)) {
+                        std::optional<clp::ffi::Value> bool_value{
+                                clp::ffi::Value{cur_field.template get<bool>()}
+                        };
+                        auto ret{evaluate_filter_against_literal_type_value_pair(
+                                filter_expr,
+                                ast::LiteralType::BooleanT,
+                                bool_value,
+                                m_case_sensitive_match
```
wraymo, line 182: is it possible to deduplicate the code a bit using something like
```
    auto eval = [&](ast::LiteralType type, std::optional<clp::ffi::Value> value) -> bool {
        auto result = evaluate_filter_against_literal_type_value_pair(
            filter_expr, type, value, m_case_sensitive_match);
        return !result.has_error() && (filter_expr->is_inverted() != result.value());
    };
```

components/core/src/clp_s/search/EvaluateMetadataFilters.cpp
```diff
@@ -0,0 +1,292 @@
...
+ * @param text
+ * @return The four byte encoded text AST corresponding to the input string.
+ */
+auto get_encoded_text_ast(std::string_view text)
+        -> clp::ir::EncodedTextAst<clp::ir::four_byte_encoded_variable_t> {
+    std::string logtype;
+    std::vector<clp::ir::four_byte_encoded_variable_t> encoded_vars;
+    std::vector<int32_t> dict_var_bounds;
+    clp::ffi::encode_message(text, logtype, encoded_vars, dict_var_bounds);
+    std::vector<std::string> dict_vars;
```
wraymo: Can we reserve the size?

components/core/src/clp_s/search/EvaluateMetadataFilters.cpp
```diff
@@ -0,0 +1,292 @@
...
+    std::vector<std::pair<ast::DescriptorList::iterator, nlohmann::json const&>> work_list;
+    work_list.emplace_back(col->descriptor_begin(), fields);
+    // Allow prefix wildcard to match zero tokens.
+    if (col->descriptor_begin() != col->descriptor_end() && col->descriptor_begin()->wildcard()) {
+        work_list.emplace_back(++col->descriptor_begin(), fields);
+    }
+
+    auto evaluate_expr
+            = [&](ast::LiteralType type, std::optional<clp::ffi::Value> const& value) -> bool {
+        auto ret{evaluate_filter_against_literal_type_value_pair(
```
wraymo: Just curious, why are we using clp::ffi methods to evaluate filter? Is it because we don't expose similar methods in clp_s?

components/core/src/clp_s/search/EvaluateMetadataFilters.cpp
```diff
@@ -0,0 +1,292 @@
...
+            }
+        } else if (cur_field.is_object() && cur_it->wildcard()) {
+            auto cur_it_tmp = cur_it++;
+            // Allow wildcard to continue matching tokens or continue on to next descriptor token.
+            for (auto const& field : cur_field) {
+                work_list.emplace_back(cur_it_tmp, field);
+                work_list.emplace_back(cur_it, field);
+            }
+        } else if (cur_field.is_object()) {
+            if (cur_field.contains(cur_it->get_token())) {
```
wraymo: Can we merge these two into one condition

components/core/src/clp_s/search/EvaluateMetadataFilters.cpp
```diff
@@ -0,0 +1,292 @@
...
+        auto ret{evaluate_filter_against_literal_type_value_pair(
+                filter_expr,
+                type,
+                value,
+                m_case_sensitive_match
+        )};
+        return false == ret.has_error() && filter_expr->is_inverted() != ret.value();
+    };
+
+    while (false == work_list.empty()) {
```
wraymo: I'm a bit confused about the purpose of this while loop. Is it because we want to handle nested column in the future? and are we planning to store nested structures in the RangeIndex fields?

components/core/tests/test-clp_s-search.cpp
```diff
@@ -216,6 +236,7 @@ TEST_CASE("clp-s-search", "[clp-s][search]") {
     ));
 
     for (auto const& [query, expected_results] : queries_and_results) {
+        CAPTURE(query);
```
wraymo, line 239: Do we need this?

docs/src/user-guide/reference-json-search-syntax.md
```diff
@@ -80,6 +80,20 @@ follows:
...
 ```
 
+### Querying kv-pairs in the archive range index
+
+The archive range index maps metadata about each file ingested into an archive to a logical range
+of the archive. Filters that reference fields in this metadata transparently get converted to
+filters that match corresponding logical ranges of an archive.
+
+To search for records corresponding to a kv-pair within the archive range index you can prefix the
+key with `$`. For example, the following query searches for records from a file that was passed to
```
wraymo: should we mention the range index fields we currently support?

wraymo (commented):
Nice work! For ` EvaluateMetadataFilters`, it seems that it's specifically used for the range index rather than searching against the metadata tree. Should we rename it to ` EvaluateRangeIndexFilters`?

components/core/tests/test-clp_s-search.cpp
gibber9809: It helps with debugging -- now when one of the queries fail this test will show you the specific query as well as the test assertion that failed.

components/core/src/clp_s/search/EvaluateMetadataFilters.cpp
gibber9809: That is part of it, but also
1. The current `evaluate_filter` functions in clp-s rely on a bunch of internal state, so even if we did expose them they would be hard to use
2. There are really good tests for the clp::ffi evaluate methods, so it should be pretty reliable

components/core/src/clp_s/search/EvaluateMetadataFilters.cpp
gibber9809: Yes, it is for that and for handling wildcards tokens in the column descriptor.

docs/src/user-guide/reference-json-search-syntax.md
gibber9809: I'll mention them in the SFA format design doc that I'm writing, but I think the current example with `$_filename` should be enough for most users.

The `$_archive_creator_id` and `$_file_split_number` fields will probably only be needed by ourselves and advanced users.

wraymo: Got it, that makes sense

docs/src/user-guide/reference-json-search-syntax.md
```diff
@@ -80,6 +80,20 @@ follows:
...
+compression as `test.jsonl`:
+
+```
+$_filename: "test.jsonl"
+```
```
kirkrodrigues:
````suggestion
### Querying file-level metadata kv-pairs

clp-json stores some metadata about each file that's compressed into an archive (e.g., the file's
name). To filter for log events that correspond to some kv-pair in this metadata, you can prefix
the key with `$`:

```
$key: value
```

For example, to query for log events that were compressed from a file whose name was `test.jsonl`:

```
$_filename: "test.jsonl"
```

To query for a key named `$key` within the default namespace, the `$` character can be escaped as
follows:

```
\$key: value
```
````

I think for a normal user, this is as much information as they would need. What do you think?

Comment (gibber9809) [reply docs/src/user-guide/reference-json-search-syntax.md]: That's fair. I guess I can leave info about how the metadata search actually works for developer docs.

**Review by wraymo — APPROVED** [review]

File: docs/src/user-guide/reference-json-search-syntax.md [hunk 9]
```diff
@@ -80,6 +80,29 @@ follows:
 \@key: value
 ```
 
+### Querying file-level metadata kv-pairs
+
+clp-json stores some metadata about each file that's compressed into an archive (e.g., the file's
+name). To filter for log events that correspond to some kv-pair in this metadata, you can prefix
+the key with `$`:
```
kirkrodrigues:
```suggestion
clp-json stores some metadata about each file that is compressed into an archive (e.g., the file's
name). To filter for log events that correspond to some kv-pair in this metadata, you can prefix the
key with `$`:
```

docs/src/user-guide/reference-json-search-syntax.md
```diff
@@ -80,6 +80,29 @@ follows:
...
+```
+
+For example, to query for log events that were compressed from a file whose name was `test.jsonl`:
+
+```
+$_filename: "test.jsonl"
+```
+
+To query for a key named `$key` within the default namespace, the `$` character can be escaped as
+follows:
```
kirkrodrigues:
```suggestion
To query for a key named `$key` in an event, the `$` character can be escaped as follows:
```

kirkrodrigues (changes requested):
Sorry, some minor rephrasing.

kirkrodrigues (approved):
Approving docs.

PR #955, opened 2025-06-02
build(core): Add flags that external projects can use to disable any part of the build.

gibber9809:
# Description

This PR adds flags that external projects can use to disable almost every part of the clp core build. This allows external projects to depend on some parts of clp without needing to install dependencies used by components the external project does not need.

We also bump the required cmake version to 3.21 to ensure all the cmake features we use are supported (in particular the `SOURCE_SUBDIR` argument for `FetchContent_Declare/ExternalProject_Add` in 3.18 and `PROJECT_IS_TOP_LEVEL` in 3.21).

These flags are all prefixed with `CLP_BUILD_` and the corresponding targets are wrapped in `if(CLP_BUILD_...)`. All of these `CLP_BUILD_` flags are cache options defaulting to true, so by default are targets are built.

These flags are used to derive a set of `CLP_NEED_X` flags that indicate which library dependencies are required to satisfy all of the targets being built. Each `find_package` call is wrapped by the `CLP_NEED_X` flag for the corresponding library.

We write one `validate_x_dependencies` and `set_x_dependencies` function for each `CLP_BUILD_` flag. The validate function for each target ensures that  the `CLP_BUILD_X` flags for all of its dependencies are true, and emits a `FATAL_ERROR` message otherwise. The set dependencies function for a target sets all of the `CLP_NEED_` flags directly linked against by the target.

Together these functions ensure that
1. The user has specified a legal combination of `CLP_BUILD_X` options
2. `find_package` is called for only the set of libraries actually needed by the build
3. `add_library` and `add_executable` are called only for the targets where `CLP_BUILD_X` is true

For convenience we wrap all of the validate/set functions in a function called `validate_and_setup_all_clp_dependency_flags`. To allow flags set inside of these functions to be exposed as variables to the caller users need to follow a call to that function with `convert_clp_dependency_properties_to_variables`.

# Checklist

  breaking change.

# Validation performed

* Validated that a project with the following cmake script could add clp as an external project and build only the necessary components to link against `clp_s::search`
```
cmake_minimum_required(VERSION 3.21)

project(TEST_PROJ LANGUAGES CXX C)

# Modules copied from clp for simplicity
set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")

# Dependencies taken from clp for simplicity
include("${CMAKE_SOURCE_DIR}/../clp/build/deps/core/cmake-settings/all.cmake"
        OPTIONAL
        RESULT_VARIABLE CLP_DEPS_SETTINGS_FILE_PATH
)

include(FetchContent)

FetchContent_Declare(
        clp
        SOURCE_DIR "${CMAKE_SOURCE_DIR}/../clp/"
        SOURCE_SUBDIR "components/core/"
)

set(CLP_USE_STATIC_LIBS ON CACHE BOOL "")
set(CLP_BUILD_EXECUTABLES OFF CACHE BOOL "")
set(CLP_BUILD_TESTING OFF CACHE BOOL "")
set(CLP_BUILD_CLP_REGEX_UTILS OFF CACHE BOOL "")
set(CLP_BUILD_CLP_S_ARCHIVEWRITER OFF CACHE BOOL "")
set(CLP_BUILD_CLP_S_JSONCONSTRUCTOR OFF CACHE BOOL "")
set(CLP_BUILD_CLP_S_REDUCER_DEPENDENCIES OFF CACHE BOOL "")
set(CLP_BUILD_CLP_S_SEARCH_SQL OFF CACHE BOOL "")

FetchContent_MakeAvailable(clp)

add_executable(test test.cpp)
target_compile_features(test PRIVATE cxx_std_20)
target_link_libraries(test clp_s::search)
```

## Summary by CodeRabbit

- **New Features**
  - Introduced build-time options to selectively enable or disable building specific binaries, libraries, and tests using configuration flags.
  - Users can now exclude certain executables or components from the build process, providing greater flexibility and control over what is built.

- **Chores**
  - Updated minimum required CMake version.
  - Improved build configuration logic for external dependencies and test targets.


components/core/src/clp_s/search/sql/CMakeLists.txt
```diff
@@ -5,20 +5,22 @@ ANTLR_TARGET(
...
+        target_compile_features(clp_s_search_sql PRIVATE cxx_std_20)
+        target_include_directories(clp_s_search_sql PRIVATE ${ANTLR_SqlParser_OUTPUT_DIR})
+        target_link_libraries(
+                clp_s_search_sql
+                PRIVATE
+                antlr4_static
+                clp_s::search::ast
+                spdlog::spdlog
+        )
+endif()
```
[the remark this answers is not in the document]
gibber9809: True, but antlr code generation at build time is being removed by another PR, so I written things this way to reduce conflicts.

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,211 @@
...
+    CLP_NO_NEED_YSTDLIB
+    "Do not include ystdlib in the build."
+    OFF
+)
+
+option(
+    CLP_NO_NEED_ZSTD
+    "Do not include zstd in the build."
+    OFF
+)
```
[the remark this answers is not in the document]
gibber9809: Each block is alphabetized (though some blocks haven't been explicitly commented as separate blocks in the BUILD section e.g. binary options, clp options, clp-s options within build).

components/core/src/clp/clo/CMakeLists.txt
```diff
@@ -151,32 +151,34 @@ set(
...
-        ystdlib::error_handling
-        ZStd::ZStd
-)
-# Put the built executable at the root of the build directory
-set_target_properties(
-        clo
-        PROPERTIES
-        RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}"
-)
+if(NOT CLP_NO_BUILD_BINARIES)
```
wraymo: Do we want to control each individual binary?

gibber9809:
> I'm not sure if we're giving the user too much flexibility by allowing them to enable or disable every dependency individually. Is there a clean approach where we can handle dependency resolution ourselves — for example, by automatically including only the dependencies required by the targets the user selects?

I don't think cmake has functionality to do this automatically for the `ExternalProject_Add` and related flows unfortunately, at least not that I could find. We could of course add some higher level flags to request certain components and then use those flags to set all of these lower level flags, but that wouldn't be trivial to implement/maintain.

components/core/src/clp/clo/CMakeLists.txt
gibber9809: Yeah it could be worthwhile. I think it doesn't matter for the immediate use-case (exposing enough options so that external projects can pull in individual libraries without needing to pull in all of our targets and dependencies), but it could be useful e.g. to create more targeted installs of clp.

kirkrodrigues:
> For convenience we wrapp all of the validate/set functions in a function called validate_and_setup_all_clp_dependency_flags. To allow flags set inside of these functions to be exposed as variables to the caller users need to follow a call to that function with convert_clp_dependency_properties_to_variables.

I'm probably missing something---why do we need `convert_clp_dependency_properties_to_variables` rather than setting the variables directly (instead of setting properties first)?

gibber9809:
> > For convenience we wrapp all of the validate/set functions in a function called validate_and_setup_all_clp_dependency_flags. To allow flags set inside of these functions to be exposed as variables to the caller users need to follow a call to that function with convert_clp_dependency_properties_to_variables.
> 
> I'm probably missing something---why do we need `convert_clp_dependency_properties_to_variables` rather than setting the variables directly (instead of setting properties first)?

It has to do with cmake's scoping rules. If you set a variable inside of a function it gets function-scoped. You can put variables in the parent's scope instead by using the `PARENT_SCOPE` argument with `set`. Since I'm setting these flags inside of _nested_ functions though the parent scope would just be scoped to the parent function (though I could set the same variables again in `validate_and_setup_all_clp_dependency_flags` with `PARENT_SCOPE` to get around this).

Properties let you have a little bit more control over scope than variables -- when I set all of the properties at `DIRECTORY` scope they become available in the scope of the directory of the cmake file they're set in (or the directory that that cmake file is `include`d in by another cmake file). Our main cmake file should actually be able to use the properties directory using `get_property`, but imo its easier to read the main cmake script if they're forwarded as variables.

If we prefer though I can just forward the variables through multiple levels of nesting via `set(... PARENT_SCOPE)` to avoid properties altogether -- have to forward them one way or another at any rate.

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,489 @@
...
+        set_clp_s_search_sql_dependencies()
+    endif()
+
+    if (CLP_BUILD_CLP_S_TIMESTAMPPATTERN)
+        validate_clp_s_timestamppattern_dependencies()
+        set_clp_s_timestamppattern_dependencies()
+    endif()
+endfunction()
+
+function (convert_clp_dependency_properties_to_variables)
```
kirkrodrigues, line 445: Can we replace the body of this method with a list + loop? Something like:

```cmake
    list(APPEND CLP_NEED_FLAGS
            CLP_NEED_ABSL
            CLP_NEED_ANTLR
            CLP_NEED_BOOST
            CLP_NEED_CATCH2
            CLP_NEED_CURL
            CLP_NEED_DATE
            CLP_NEED_FMT
            CLP_NEED_LOG_SURGEON
            CLP_NEED_LIBARCHIVE
            CLP_NEED_LZMA
            CLP_NEED_MARIADB
            CLP_NEED_MONGOCXX
            CLP_NEED_MSGPACKCXX
            CLP_NEED_NLOHMANN_JSON
            CLP_NEED_OPENSSL
            CLP_NEED_SIMDJSON
            CLP_NEED_SPDLOG
            CLP_NEED_SQLITE
            CLP_NEED_YAMLCPP
            CLP_NEED_YSTDLIB
            CLP_NEED_ZSTD
    )

    foreach(FLAG IN LISTS CLP_NEED_FLAGS)
        get_property(VALUE DIRECTORY PROPERTY "${FLAG}")
        set("${FLAG}" "${VALUE}" PARENT_SCOPE)
    endforeach ()
```

kirkrodrigues (commented):
Haven't finished reviewing everything, but in the interest of time, could you take a look at my suggestion and see if that can be applied throughout the PR?

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,458 @@
+# Build Target Options
+option(
+    CLP_BUILD_BINARIES
```
kirkrodrigues: How about `CLP_BUILD_EXECUTABLES`? I feel like `BINARIES` encompasses libraries which makes `CLP_BUILD_BINARIES` a bit confusing.

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,458 @@
...
+    "Build all binaries."
```
kirkrodrigues:
```suggestion
    "Build all executables."
```

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,458 @@
...
+    ON
+)
+
+option(
+    CLP_BUILD_TESTS
```
kirkrodrigues: How about `CLP_BUILD_TESTING` to stick with the naming convention of CMake's [BUILD_TESTING](https://cmake.org/cmake/help/latest/variable/BUILD_TESTING.html)? (This is also what other projects do.)

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,458 @@
...
+    ON
+)
+
+option(
+    CLP_BUILD_CLP_S_TIMESTAMPPATTERN
+    "Build clp_s::timestamp_pattern."
+    ON
+)
+
+function(validate_clp_dependencies_for_target TARGET)
```
kirkrodrigues:
```suggestion
# Validates that the `CLP_BUILD_` options required by `TARGET_CLP_BUILD_OPTION` are `ON`.
#
# @param {string} TARGET_CLP_BUILD_OPTION
# @param {string[]} ARGN The required `CLP_BUILD_` options
function(validate_clp_dependencies_for_target TARGET)
```

kirkrodrigues: How about `TARGET_CLP_BUILD_OPTION`?

kirkrodrigues:
```suggestion
function(validate_clp_dependencies_for_target TARGET)
    if (NOT DEFINED TARGET_CLP_BUILD_OPTION OR TARGET_CLP_BUILD_OPTION STREQUAL "")
        message(FATAL_ERROR "TARGET_CLP_BUILD_OPTION can't be unset or empty.")
    endif()

```

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,458 @@
...
+    foreach(DEPENDENCY IN LISTS ARGN)
+        if (NOT "${${DEPENDENCY}}")
+            message(FATAL_ERROR "${TARGET} requires ${DEPENDENCY}=ON")
+        endif()
+    endforeach()
+endfunction()
+
+function(set_clp_need_flags)
```
kirkrodrigues, line 111:
```suggestion
# Sets the given `CLP_NEED_` flags as directory properties
#
# @param {string[]} ARGV The `CLP_NEED_` flags to set.
function(set_clp_need_flags)
```

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,458 @@
...
+    foreach(NEEDS_FLAG IN LISTS ARGV)
+        set_property(DIRECTORY PROPERTY "${NEEDS_FLAG}" TRUE)
```
kirkrodrigues:
```suggestion
        set_property(DIRECTORY PROPERTY "${NEEDS_FLAG}" ON)
```

For consistency. Unless `TRUE` is required for some reason?

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,458 @@
...
+        CLP_BUILD_CLP_REGEX_UTILS
+        CLP_BUILD_CLP_STRING_UTILS
+        CLP_BUILD_CLP_S_ARCHIVEREADER
+        CLP_BUILD_CLP_S_ARCHIVEWRITER
+        CLP_BUILD_CLP_S_IO
+        CLP_BUILD_CLP_S_JSONCONSTRUCTOR
+        CLP_BUILD_CLP_S_SEARCH
+        CLP_BUILD_CLP_S_SEARCH_AST
+        CLP_BUILD_CLP_S_SEARCH_KQL
+    )
```
kirkrodrigues: * How come `CLP_BUILD_CLP_S_CLP_DEPENDENCIES` and `CLP_CLP_S_REDUCER_DEPENDENCIES` are not in this list?
* Why is `CLP_BUILD_CLP_REGEX_UTILS` in this list?

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,458 @@
...
+        CLP_NEED_MONGOCXX
+        CLP_NEED_MSGPACKCXX
+        CLP_NEED_NLOHMANN_JSON
+        CLP_NEED_OPENSSL
+        CLP_NEED_SIMDJSON
+        CLP_NEED_SPDLOG
+        CLP_NEED_SQLITE
+        CLP_NEED_YAMLCPP
+        CLP_NEED_YSTDLIB
+        CLP_NEED_ZSTD
```
kirkrodrigues: * Alphabetize
* How come absl & curl aren't in this list?

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,458 @@
...
+    )
+endfunction()
+
+function(validate_clp_tests_dependencies)
+    validate_clp_dependencies_for_target("CLP_BUILD_TESTS"
```
kirkrodrigues: It's a little odd to quote one argument but not the others. Any reason why? Personally, I would get rid of the quotes.

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,458 @@
...
+        CLP_BUILD_CLP_REGEX_UTILS
+        CLP_BUILD_CLP_STRING_UTILS
+        CLP_BUILD_CLP_S_SEARCH_AST
+        CLP_BUILD_CLP_S_SEARCH_KQL
+        CLP_BUILD_CLP_S_SEARCH_SQL
```
kirkrodrigues, line 161: How come `CLP_BUILD_CLP_S_TIMESTAMPPATTERN` isn't in the list?

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,458 @@
...
+endfunction()
+
+function(set_clp_s_timestamppattern_dependencies)
+    set_clp_need_flags(
+        CLP_NEED_DATE
+        CLP_NEED_SPDLOG
+    )
+endfunction()
+
+function(validate_and_setup_all_clp_dependency_flags)
```
kirkrodrigues, line 371:
```suggestion
# Validates that for each target whose `CLP_BUILD_` option is `ON`, the `CLP_BUILD_` options for
# the target's dependencies are also `ON`; Sets the required `CLP_NEED_` flags for any target that
# will be built.
function(validate_and_setup_all_clp_dependency_flags)
```

components/core/cmake/Options/options.cmake
gibber9809: I think I missed the indexer CMakeLists.txt when making this list -- will add absl and curl.

components/core/cmake/Options/options.cmake
gibber9809: Both right -- will update. I think I just missed the clp-s and indexer CMakeLists when populating this list.

gibber9809: Actually I think it might have been that there was a version of this script where I got rid of both of those flags, but it looks like I didn't put them in this list when I added them back.

components/core/cmake/Options/options.cmake
gibber9809: To me it makes it easier to distinguish the target from the dependencies, but I'll get rid of it.

components/core/cmake/Options/options.cmake
gibber9809: Will add it.

gibber9809:
> Thinking about it, I think there's a cleaner approach to achieve the goals of this PR:
> 
>     * For each CMake target, define the list of `CLP_BUILD_` options it requires
> 
>     * If a target's `CLP_BUILD_` option is set, iterate over each each target's dependent options to check if each one is set
> 
>     * Move the required `find_package` calls for each target into the target's CMake list so that the `NEED_` flag tracking will be done automatically.
>       
>       * To ensure we still use a consistent of external dependency versions and options, we can define these in the root CMakeLists.txt.
>       * Calling `find_package` multiple times should be fine as long the library's CMake configure script is written correctly.
> 
> 
> That said, since this PR is blocking other PRs, let's defer the above to another PR (ideally, immediately after this one is merged).
> 
> Inline I left some comments about missing internal/external dependencies. If the omissions weren't intentional, can you recheck all of the dependencies?

Does moving the `find_package` calls work for something like `boost` though where we depend on many different components of a package? I'm not sure, but I think it might just use the list of components from the first `find_package` call for boost, which could cause problems if we use different lists of components in different `find_package` calls.

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,475 @@
...
+
+function(set_clp_tests_dependencies)
+    set_clp_need_flags(
+        CLP_NEED_ABSL
+        CLP_NEED_BOOST
+        CLP_NEED_CATCH2
+        CLP_NEED_DATE
+        CLP_NEED_FMT
+        CLP_NEED_LOG_SURGEON
+        CLP_NEED_LIBARCHIVE
```
kirkrodrigues:
```suggestion
        CLP_NEED_LIBARCHIVE
        CLP_NEED_LOG_SURGEON
```

components/core/cmake/Options/options.cmake
```diff
@@ -0,0 +1,475 @@
...
+    validate_clp_dependencies_for_target(CLP_BUILD_CLP_S_IO
+        CLP_BUILD_CLP_S_CLP_DEPENDENCIES
+    )
+endfunction()
+
+function(set_clp_s_io_dependencies)
+    set_clp_need_flags(
+        CLP_NEED_BOOST
+        CLP_NEED_FMT
+        CLP_NEED_MSGPACKCXX
```
kirkrodrigues: I might be missing something, but is this required? I don't see it in target_link_libs.

gibber9809: You're right -- missed that one as well.

wraymo (approved):

PR #956, opened 2025-06-02
feat(new-webui): Sort search results by timestamp.

davemarco:
# Description

PR adds ability to sort by timestamp.

I took a look at old webui, and It implemented server side sorting. I thought the idea of this is that it would requery, and you could get either the earliest or latest 1000 results. 

After looking at it further, the server side sorting is only in the results cache, and not actually in clp, or clp-s. Like it dosent requery. As a results, the server side sorting dosent actually change the available results, and the same functionality can be achieved with client-side sorting. 

To simplify until we have fancier queries, that can be resubmitted, I simply implemented client side sorting to achieve the same functionality, which is simpler.  It's not that it would be hard, but this is simpler. 

# Checklist

  breaking change.

# Validation performed

Results can be sorted

## Summary by CodeRabbit

- **New Features**
  - Improved sorting behaviour for the "timestamp" column in the search results table, including a default ascending order and updated sort direction cycling.
- **Style**
  - Adjusted user interaction by removing the option to clear sorting on the "timestamp" column.

---
  - https://app.asana.com/0/0/1210417831657504

### DIFF
--- a/components/log-viewer-webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/typings.tsx
+++ b/components/log-viewer-webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/typings.tsx
@@ -3,6 +3,7 @@ import dayjs from "dayjs";
 
 import {DATETIME_FORMAT_TEMPLATE} from "../../../../typings/datetime";
 import Message from "./Message";
+import {getStreamId} from "./utils";
 
 
 /**
@@ -10,12 +11,13 @@ import Message from "./Message";
  */
 interface SearchResult {
     _id: string;
-    timestamp: number;
-    message: string;
+    archive_id: string;
     filePath: string;
-    orig_file_path: string;
-    orig_file_id: string;
     log_event_ix: number;
+    message: string;
+    orig_file_id: string;
+    orig_file_path: string;
+    timestamp: number;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


components/log-viewer-webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/typings.tsx
```diff
@@ -24,9 +24,17 @@ interface SearchResult {
...
+
+        // Specifying a third sort direction removes ability for user to cancel sorting.
+        sortDirections: [
+            "ascend",
+            "descend",
+            "ascend",
+        ],
         title: "Timestamp",
         width: 15,
     },
```
junhaoliao, line 42: (optional: utterly nit-picking) since there're many fields already, i'd probably group the fields:
```suggestion
    {
        dataIndex: "timestamp",
        key: "timestamp",

        title: "Timestamp",
        render: (timestamp: number) => dayjs(timestamp).format(DATETIME_FORMAT_TEMPLATE),
        width: 15,
        
        // Specifying a third sort direction removes ability for user to cancel sorting.
        defaultSortOrder: "ascend",
        sortDirections: [
            "ascend",
            "descend",
            "ascend",
        ],
        sorter: (a, b) => a.timestamp - b.timestamp,
    },
```

junhaoliao (approved):
for the PR title, how about:
```
feat(new-webui): Sort search results by timestamp.
```

components/log-viewer-webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/typings.tsx
davemarco: I would be happy to do this, but the linter will complain about sort keys

PR #957, opened 2025-06-02
feat(new-webui): Enable query submission via Enter key (resolves #943).

### LINKED ISSUE #943: [new-webui] Make the Enter key start a search query.
### Request

As the title says. This is necessary for feature parity with the current webui.

### Possible implementation

Wrap the search input in a form that has an onSubmit handler.

davemarco:
# Description

When user hits enter, the query will now submit. Resolves #943 

Implementation is just to wrap the search controls in a form, and prevent default behaviour from triggering page load.

I also explicitly removed behaviour for cancel button, I think this is better. I think on old webui enter would also cancel query.

# Checklist

  breaking change.

# Validation performed

Tested enter submits query.

## Summary by CodeRabbit

- **Refactor**
  - Updated search controls to use a form element with proper submit and cancel button types for improved accessibility and consistent behaviour.


components/log-viewer-webui/client/src/pages/SearchPage/SearchControls/index.tsx
```diff
@@ -11,11 +11,17 @@ import TimeRangeInput from "./TimeRangeInput";
...
     return (
-        <div className={styles["searchControlsContainer"]}>
-            <QueryInput/>
-            <TimeRangeInput/>
-            <SearchButton/>
-        </div>
+        <form
+            onSubmit={(e) => {
+                e.preventDefault();
+            }}
```
junhaoliao: let's extract this and use `useCallback` to wrap the definition

davemarco: I just moved it out of component. The effect should be the same just without useCallback

components/log-viewer-webui/client/src/pages/SearchPage/SearchControls/index.tsx
```diff
@@ -4,18 +4,29 @@ import SearchButton from "./SearchButton/SearchButton";
...
 
 
+/**
+ * Prevents the default behavior to avoid page reload when submitting query.
+ *
+ * @param e
+ */
+const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+};
```
junhaoliao: Let's use `ev` for "event" as a convention, as we usually use `e` for "error".
```suggestion
 * @param ev
 */
const handleSubmit = (ev: React.FormEvent<HTMLFormElement>) => {
    ev.preventDefault();
};
```

junhaoliao (approved):

PR #963, opened 2025-06-04
feat(core): Add environment variable to limit parallelism for core's build tasks.

kirkrodrigues:
# Description

When building the dependencies for `core` as well as core itself, a machine can run out of memory if the build parallelism isn't restricted. We've been seeing something like this happen in the GH workflows for #925. Thus, this PR adds an environment variable `CLP_CORE_MAX_PARALLELISM_PER_BUILD_TASK` that, as the name suggests, when set, it limit the parallelism of each build task.

Of course, if we run many concurrent tasks then a machine could still run out of memory; however, in our testing, it seems like it's just specific builds like `core`'s and `antlr-runtime`'s that can exceed the capacity of a single machine, probably because they have many source files they can compile concurrently. So until the number of concurrent tasks we can run matches the number of concurrent files they can compile, task-level concurrency is probably a non-issue.

For now, this environment variable is undocumented. Once we add proper tasks to build & test `core`, we can document the environment variable on the page that describes those tasks.

# Checklist

  breaking change.

# Validation performed

* Compared the runtime of the following commands on a 12-core machine and observed the former was proportionally faster than the latter:
  ```bash
  task clean && rm -r .task && task core
  ```
  and
  ```bash
  task clean && rm -r .task && CLP_CORE_MAX_PARALLELISM_PER_BUILD_TASK=6 task core
  ```

## Summary by CodeRabbit

- **Chores**
	- Enhanced build workflows to automatically adjust parallelism according to available processors, improving efficiency of dependency processing and build tasks on macOS, CentOS, and Ubuntu systems.

### DIFF
--- a/.github/workflows/clp-core-build-macos.yaml
+++ b/.github/workflows/clp-core-build-macos.yaml
@@ -74,7 +74,7 @@ jobs:
       - run: "./tools/scripts/deps-download/init.sh"
         shell: "bash"
 
-      - run: "task deps:core"
+      - run: "CLP_CORE_MAX_PARALLELISM_PER_BUILD_TASK=$(getconf _NPROCESSORS_ONLN) task deps:core"
         shell: "bash"
 
       - name: "Build CLP-core and run unit tests"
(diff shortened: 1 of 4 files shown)


anlowee (approved):

junhaoliao:
it will be great if the env var `CLP_CORE_MAX_PARALLELISM_PER_BUILD_TASK` can be documented. @coderabbitai create an doc request issue to track this

PR #969, opened 2025-06-05
refactor(core): Replace `outcome_v2::std_result` with `ystdlib::error_handling::Result`; Remove `outcome` dependency.

sitaowang1998:
# Description

This pr removes `outcome` library from the project. This pr replaces `outcome_v2::std::result` in the `core` with `ystdlib::error_handling::Result`. This pr also bumps `ystdlib-cpp` version so it now depends on `Boost::outcome` instead of `outcome`.

# Checklist

  breaking change.

# Validation performed
* [x] GitHub workflows pass.
* [x] core builds and unit tests pass.

## Summary by CodeRabbit

- **Refactor**
  - Updated the underlying error handling mechanism throughout the application to use a new result type from ystdlib. This change affects log event deserialization, query handling, search utilities, serialization, regex translation, and related components. All error and result reporting now utilize the new approach, with no impact on user-facing functionality or workflows.

- **Chores**
  - Removed the obsolete "outcome" dependency and updated the "ystdlib" dependency to a new version.


components/core/src/clp/regex_utils/CMakeLists.txt
```diff
@@ -12,11 +12,6 @@ add_library(
         ${REGEX_UTILS_HEADER_LIST}
 )
 add_library(clp::regex_utils ALIAS regex_utils)
-target_include_directories(regex_utils
-        PUBLIC
-        ../
```
LinZhihao-723, line 17: This is still needed right?

LinZhihao-723 (changes requested):
Only one more comment and we should be ready to merge. For the PR title, how about:
```
refactor(core): Replace `outcome_v2::std_result` with `ystdlib::error_handling::Result`; Remove `outcome` from the build dependency.
```

components/core/src/clp/regex_utils/CMakeLists.txt
sitaowang1998: Fixed.

PR #977, opened 2025-06-07
fix(core): Try to exhaust Zstd's internal buffers when they might contain unconsumed data. (fixes #976).

### LINKED ISSUE #976: `clp::streaming_compression::zstd::Decompressor` prematurely returns EOF when buffered data remains
### Bug

(originally reported by @jackluo923 )

The `clp::streaming_compression::zstd::Decompressor` incorrectly returns EOF errors when the input buffer is fully consumed, even when the zstd decompression context still contains buffered data that can be flushed.

According to the [zstd manual](http://facebook.github.io/zstd/zstd_manual.html#Chapter9), when `output.pos == output.size` (output buffer is full), there may still be data remaining in the decompression context. The manual states that developers should "call `ZSTD_decompressStream()` again to flush whatever remains in the buffer."

In the next call of bytes reading from the decompressor, our current implementation only checks if more bytes can be consumed from the input buffer and returns EOF if not, completely ignoring that calling `ZSTD_decompressStream()` could yield additional data from the internal buffers.

### CLP version

c69d4cf798651d4e71270e4aa0c4a28ce8b4f296

### Environment

the issue was original found from clp-ffi-js which is compiled by Emscripten 3.0.67 (derived from LLVM)

this env was used to reproduce with below steps:
```
g++ -v
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/11/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none:amdgcn-amdhsa
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 11.4.0-1ubuntu1~22.04' --with-bugurl=file:///usr/share/doc/gcc-11/README.Bugs --enable-languages=c,ada,c++,go,brig,d,fortran,objc,obj-c++,m2 --prefix=/usr --with-gcc-major-version-only --program-suffix=-11 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-bootstrap --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-plugin --enable-default-pie --with-system-zlib --enable-libphobos-checking=release --with-target-system-zlib=auto --enable-objc-gc=auto --enable-multiarch --disable-werror --enable-cet --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none=/build/gcc-11-XeT9lY/gcc-11-11.4.0/debian/tmp-nvptx/usr,amdgcn-amdhsa=/build/gcc-11-XeT9lY/gcc-11-11.4.0/debian/tmp-gcn/usr --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu --with-build-config=bootstrap-lto-lean --enable-link-serialization=2
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 11.4.0 (Ubuntu 11.4.0-1ubuntu1~22.04) 
```

### Reproduction steps

1. Download and extract the test file `compressed-logs.clp.zst` from [compressed-logs.clp.zst.zip](https://github.com/user-attachments/files/20636715/compressed-logs.clp.zst.zip)

2. Add this test case to `components/core/tests/test-StreamingCompression.cpp`:

    ```cpp
    SECTION("ZStd compression without flush") {
        decompressor = std::make_unique<clp::streaming_compression::zstd::Decompressor>();
        clp::ReadOnlyMemoryMappedFile const memory_mapped_compressed_file{
            "/home/junhao/Downloads/compressed-logs.clp.zst"
        };
        auto const compressed_file_view{memory_mapped_compressed_file.get_view()};
        decompressor->open(compressed_file_view.data(), compressed_file_view.size());
    
        std::ranges::fill(decompressed_buffer.begin(), decompressed_buffer.end(), 0);
        
        // First read should succeed
        auto read_result = decompressor->try_read_exact_length(decompressed_buffer.data(), 4);
        REQUIRE(read_result == ErrorCode_Success);
        
        // Second read should also succeed but currently fails with EOF
        read_result = decompressor->try_read_exact_length(decompressed_buffer.data(), 1);
        REQUIRE(read_result == ErrorCode_Success);  // This assertion fails
    
        decompressor->close();
    }
    ```

3. Run the test - the second `try_read_exact_length` call will fail with EOF

4. However, using `zstdcat` confirms the file contains approximately 260 uncompressed bytes, which is significantly more than the 5 bytes (4 + 1) requested in the test:

    ```bash
    $ zstdcat compressed-logs.clp.zst | wc -c
    compressed-logs.clp.zst : Read error (39) : premature end 
       260
    ```

    The second `try_read_exact_length` call should succeed and read the requested number of bytes, not return an EOF error.

junhaoliao:
# Description

1. According to the zstd manual, check for frame continuation by checking the return value of `ZSTD_decompressStream`: http://facebook.github.io/zstd/zstd_manual.html#Chapter9
2. Refactor buffer refill logic into a separate method.

# Checklist

  breaking change.

# Validation performed

1. Added the test case in #976 . Ran all cases and observed success.
2. Replace the implementation in clp-ffi-js and was able to decode one log event with the code in https://github.com/y-scope/clp-ffi-js/issues/88 

## Summary by CodeRabbit

- **Refactor**
  - Improved the reliability and maintainability of the decompression process by centralizing and simplifying buffer refill logic.
  - Enhanced internal tracking of decompression progress for more accurate status reporting.

### DIFF
--- a/components/core/src/clp_s/indexer/IndexManager.cpp
+++ b/components/core/src/clp_s/indexer/IndexManager.cpp
@@ -1,10 +1,12 @@
 #include "IndexManager.hpp"
 
 #include <filesystem>
+#include <memory>
 #include <stack>
 #include <string>
 
 #include "../archive_constants.hpp"
+#include "../TimestampDictionaryReader.hpp"
 
 namespace clp_s::indexer {
 IndexManager::IndexManager(
@@ -43,7 +45,10 @@ void IndexManager::update_metadata(std::string const& dataset_name, Path const& 
     ArchiveReader archive_reader;
     archive_reader.open(archive_path, NetworkAuthOption{});
 
-    traverse_schema_tree_and_update_metadata(archive_reader.get_schema_tree());
+    traverse_schema_tree_and_update_metadata(
+            archive_reader.get_schema_tree(),
+            archive_reader.get_timestamp_dictionary()
+    );
 }
 
 std::string IndexManager::escape_key_name(std::string_view const key_name) {
@@ -94,9 +99,14 @@ std::string IndexManager::escape_key_name(std::string_view const key_name) {
 }
 
 void IndexManager::traverse_schema_tree_and_update_metadata(
-        std::shared_ptr<SchemaTree> const& schema_tree
+        std::shared_ptr<SchemaTree> const& schema_tree,
+        std::shared_ptr<TimestampDictionaryReader> const& timestamp_dict
 ) {
     if (nullptr == schema_tree) {
+        return;
+    }
+
+    if (nullptr == timestamp_dict) {
         return;
     }
 
@@ -129,7 +139,12 @@ void IndexManager::traverse_schema_tree_and_update_metadata(
         path_buffer += escape_key_name(node.get_key_name());
         if (children_ids.empty() && NodeType::Object != node_type && NodeType::Unknown != node_type)
         {
-            m_field_update_callback(path_buffer, node_type);
+            // Always index authoritative timestamp as `NodeType::DateString`
+            auto const indexed_node_type
+                    = timestamp_dict->get_authoritative_timestamp_column_ids().contains(node_id)
+                              ? NodeType::DateString
+                              : node_type;
+            m_field_update_callback(path_buffer, indexed_node_type);
         }
 
         for (auto child_id : children_ids) {
--- a/components/core/src/clp_s/indexer/IndexManager.hpp
+++ b/components/core/src/clp_s/indexer/IndexManager.hpp
@@ -2,10 +2,12 @@
 #define CLP_S_INDEXER_INDEXMANAGER_HPP
 
 #include <functional>
+#include <memory>
 #include <optional>
 
 #include "../../clp/GlobalMetadataDBConfig.hpp"
 #include "../ArchiveReader.hpp"
+#include "../TimestampDictionaryReader.hpp"
 #include "MySQLIndexStorage.hpp"
 
 namespace clp_s::indexer {
@@ -64,8 +66,12 @@ private:
     /**
      * Traverses the schema tree and updates the metadata
      * @param schema_tree
+     * @param timestamp_dict
      */
-    void traverse_schema_tree_and_update_metadata(std::shared_ptr<SchemaTree> const& schema_tree);
+    void traverse_schema_tree_and_update_metadata(
+            std::shared_ptr<SchemaTree> const& schema_tree,
+            std::shared_ptr<TimestampDictionaryReader> const& timestamp_dict
+    );
 
     OutputType m_output_type{OutputType::Database};
     std::shared_ptr<MySQLIndexStorage> m_mysql_index_storage;
--- a/components/core/src/clp_s/indexer/indexer.cpp
+++ b/components/core/src/clp_s/indexer/indexer.cpp
@@ -31,7 +31,10 @@ int main(int argc, char const* argv[]) {
     }
 
     try {
-        clp_s::indexer::IndexManager index_manager(command_line_arguments.get_db_config());
+        clp_s::indexer::IndexManager index_manager(
+                command_line_arguments.get_db_config(),
+                command_line_arguments.should_create_table()
+        );
         index_manager.update_metadata(
                 command_line_arguments.get_dataset_name(),
                 command_line_arguments.get_archive_path()


components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
```diff
@@ -34,67 +34,33 @@ auto Decompressor::try_read(char* buf, size_t num_bytes_to_read, size_t& num_byt
...
-                    m_compressed_stream_block.pos = 0;
-                    m_compressed_stream_block.size = m_read_buffer_length;
-                    break;
-                }
-                default:
-                    throw OperationFailed(ErrorCode_Unsupported, __FILENAME__, __LINE__);
+        if (m_compressed_stream_block.pos == m_compressed_stream_block.size
+            && false == m_frame_has_more_data)
+        {
+            auto const error_code{refill_compressed_buffer()};
```
junhaoliao: i did the refactoring as i was navigating through the long function. let me know if it's better to revert the function extraction.

components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
```diff
@@ -34,67 +34,33 @@ auto Decompressor::try_read(char* buf, size_t num_bytes_to_read, size_t& num_byt
...
             SPDLOG_ERROR(
-                    "streaming_compression::zstd::Decompressor: ZSTD_decompressStream() error: "
-                    "{}",
-                    ZSTD_getErrorName(error)
+                    "streaming_compression::zstd::Decompressor: ZSTD_decompressStream() error: {}",
+                    ZSTD_getErrorName(ret)
             );
             return ErrorCode_Failure;
         }
+        m_frame_has_more_data = (ret > 0);
```
junhaoliao: adding a test case into components/core/tests/test-StreamingCompression.cpp is possible but may require some significant amount of refactoring of the current cases. let me know if i should proceed with the changes

components/core/src/clp/streaming_compression/zstd/Decompressor.hpp
```diff
@@ -120,6 +120,14 @@ class Decompressor : public ::clp::streaming_compression::Decompressor {
     };
 
     // Methods
+    /**
+     * Refills the compressed buffer with data from the underlying reader.
```
kirkrodrigues:
```suggestion
     * Refills m_compressed_stream_block with data from the underlying input medium.
```

components/core/src/clp/streaming_compression/zstd/Decompressor.hpp
```diff
@@ -120,6 +120,14 @@ class Decompressor : public ::clp::streaming_compression::Decompressor {
...
+     *
+     * @return ErrorCode_Success on success
+     * @return ErrorCode_EndOfFile if no more data is available
```
kirkrodrigues, line 127:
```suggestion
     * @return ErrorCode_EndOfFile if no more data is available
     * @return Forwards `ReaderInterface::try_read`'s return values.
```

components/core/src/clp/streaming_compression/zstd/Decompressor.hpp
```diff
@@ -120,6 +120,14 @@ class Decompressor : public ::clp::streaming_compression::Decompressor {
...
+     */
+    [[nodiscard]] auto refill_compressed_buffer() -> ErrorCode;
```
kirkrodrigues: How about `refill_compressed_stream_block` to match the variable name?

components/core/src/clp/streaming_compression/zstd/Decompressor.hpp
```diff
@@ -142,6 +150,7 @@ class Decompressor : public ::clp::streaming_compression::Decompressor {
     ZSTD_inBuffer m_compressed_stream_block{};
 
     size_t m_decompressed_stream_pos{0ULL};
+    bool m_frame_has_more_data{false};
```
kirkrodrigues: How about `m_zstd_frame_has_more_data`?

components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
```diff
@@ -34,67 +34,33 @@ auto Decompressor::try_read(char* buf, size_t num_bytes_to_read, size_t& num_byt
...
-                }
-                default:
-                    throw OperationFailed(ErrorCode_Unsupported, __FILENAME__, __LINE__);
+        if (m_compressed_stream_block.pos == m_compressed_stream_block.size
+            && false == m_frame_has_more_data)
+        {
+            auto const error_code{refill_compressed_buffer()};
+            if (ErrorCode_Success != error_code) {
+                num_bytes_read = decompressed_stream_block.pos;
+                return decompressed_stream_block.pos > 0 ? ErrorCode_Success : error_code;
```
kirkrodrigues: If `error_code` isn't `ErrorCode_EndOfFile`, then I don't think it's appropriate to return `ErrorCode_Success` to the caller.

kirkrodrigues: I think another bug in the existing code is that we don't update `m_decompressed_stream_pos` before we return here. Can you add it?

components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
```diff
@@ -228,6 +194,37 @@ auto Decompressor::get_decompressed_stream_region(
...
+            m_compressed_stream_block.pos = 0;
+            m_compressed_stream_block.size = m_read_buffer_length;
+            return ErrorCode_Success;
+        }
+        default:
+            throw OperationFailed(ErrorCode_Unsupported, __FILENAME__, __LINE__);
+    }
+}
+
 auto Decompressor::reset_stream() -> void {
```
kirkrodrigues, line 232: We need to set `m_frame_has_more_data` to false in this method.

components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
kirkrodrigues: What kind of refactoring?

kirkrodrigues (changes requested):

components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
junhaoliao: that I extracted a helper method. i think it's good for clarity though it's not strictly required for the fix

kirkrodrigues: You extracted a helper method for the test cases?

junhaoliao: sorry I was super blind... thought I was looking at another comment

a case that best illustrates the issue is that it writes a small chunk (less than ZSTD_BLOCKSIZE) of compressed data, creates a decompressor with a truncated view of the compressed data, and only expects correctness of certain leading uncompressed bytes with no error, until the rest in the frame is fully consumed 

in the current test code, the "compress" and "decompress_and_compare" helpers use for-loops to write and expect full correctness of more than one chunk of data. There's also no negative tests. I was thinking we should extract the compressor / file reader / decompressor creation logic so that we don't necessarily repeat those across different cases

anyways, let me clean up the test source file on my end, and then we can discuss whether / how much we want to check things in

components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
junhaoliao: > that I extracted a helper method. i think it's good for clarity though it's not strictly required for the fix

@kirkrodrigues I thought your comment https://github.com/y-scope/clp/pull/977/files#r2133962400 was posted on this thread. sorry for the confusion

components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
junhaoliao: my attempt to add the test case and refactor:
```cpp
#include <algorithm>
#include <array>
#include <cstring>
#include <memory>
#include <numeric>
#include <string>
#include <string_view>
#include <utility>

#include <boost/filesystem/operations.hpp>
#include <catch2/catch.hpp>
#include <ystdlib/containers/Array.hpp>
#include <zstd.h>

#include "../src/clp/ErrorCode.hpp"
#include "../src/clp/FileWriter.hpp"
#include "../src/clp/ReadOnlyMemoryMappedFile.hpp"
#include "../src/clp/streaming_compression/Compressor.hpp"
#include "../src/clp/streaming_compression/Decompressor.hpp"
#include "../src/clp/streaming_compression/lzma/Compressor.hpp"
#include "../src/clp/streaming_compression/lzma/Decompressor.hpp"
#include "../src/clp/streaming_compression/passthrough/Compressor.hpp"
#include "../src/clp/streaming_compression/passthrough/Decompressor.hpp"
#include "../src/clp/streaming_compression/zstd/Compressor.hpp"
#include "../src/clp/streaming_compression/zstd/Decompressor.hpp"

using clp::ErrorCode_Success;
using clp::FileWriter;
using clp::streaming_compression::Compressor;
using clp::streaming_compression::Decompressor;
using std::string;
using std::string_view;
using ystdlib::containers::Array;

namespace {
constexpr string_view cCompressedFilePath{"test_streaming_compressed_file.bin"};
constexpr size_t cBufferSize{128L * 1024 * 1024};  // 128MB
constexpr auto cCompressionChunkSizes = std::to_array<size_t>(
        {0,
         cBufferSize / 100,
         cBufferSize / 50,
         cBufferSize / 25,
         cBufferSize / 10,
         cBufferSize / 5,
         cBufferSize / 2,
         cBufferSize}
);

auto fill_buffer(Array<char>& buffer, size_t alphabet_length) -> void;

auto compress(
        std::unique_ptr<Compressor> compressor,
        char const* src,
        std::span<size_t const> const& chunk_sizes
) -> void;

auto decompress_and_compare(
        std::unique_ptr<Decompressor> decompressor,
        Array<char> const& uncompressed_buffer,
        Array<char>& decompressed_buffer
) -> void;

auto fill_buffer(Array<char>& buffer, size_t const alphabet_length) -> void {
    for (size_t i{0}; i < buffer.size(); ++i) {
        buffer.at(i) = static_cast<char>(('a' + (i % alphabet_length)));
    }
}

auto compress(
        std::unique_ptr<Compressor> compressor,
        char const* src,
        std::span<size_t const> const& chunk_sizes
) -> void {
    FileWriter file_writer;
    file_writer.open(string(cCompressedFilePath), FileWriter::OpenMode::CREATE_FOR_WRITING);
    compressor->open(file_writer);
    for (auto const chunk_size : chunk_sizes) {
        compressor->write(src, chunk_size);
    }
    compressor->close();
    file_writer.close();
}

auto decompress_and_compare(
        std::unique_ptr<Decompressor> decompressor,
        Array<char> const& uncompressed_buffer,
        Array<char>& decompressed_buffer
) -> void {
    clp::ReadOnlyMemoryMappedFile const memory_mapped_compressed_file{string(cCompressedFilePath)};
    auto const compressed_file_view{memory_mapped_compressed_file.get_view()};
    decompressor->open(compressed_file_view.data(), compressed_file_view.size());

    size_t num_uncompressed_bytes{0};
    for (auto const chunk_size : cCompressionChunkSizes) {
        // Clear the buffer to ensure that we are not comparing values from a previous test
        std::ranges::fill(decompressed_buffer.begin(), decompressed_buffer.end(), 0);
        REQUIRE(
                (ErrorCode_Success
                 == decompressor->get_decompressed_stream_region(
                         num_uncompressed_bytes,
                         decompressed_buffer.data(),
                         chunk_size
                 ))
        );
        REQUIRE(std::equal(
                uncompressed_buffer.begin(),
                uncompressed_buffer.begin() + chunk_size,
                decompressed_buffer.begin()
        ));
        num_uncompressed_bytes += chunk_size;
    }

    REQUIRE(
            (std::accumulate(
                     cCompressionChunkSizes.cbegin(),
                     cCompressionChunkSizes.cend(),
                     size_t{0}
             )
             == num_uncompressed_bytes)
    );
}
}  // namespace

TEST_CASE("StreamingCompression", "[StreamingCompression]") {
    constexpr size_t cAlphabetLength{26};

    std::unique_ptr<Compressor> compressor;
    std::unique_ptr<Decompressor> decompressor;

    Array<char> decompressed_buffer(cBufferSize);
    Array<char> uncompressed_buffer(cBufferSize);
    fill_buffer(uncompressed_buffer, cBufferSize);

    SECTION("ZStd single phase compression") {
        compressor = std::make_unique<clp::streaming_compression::zstd::Compressor>();
        compress(std::move(compressor), uncompressed_buffer.data(), cCompressionChunkSizes);
        decompressor = std::make_unique<clp::streaming_compression::zstd::Decompressor>();
        decompress_and_compare(std::move(decompressor), uncompressed_buffer, decompressed_buffer);
    }

    SECTION("Passthrough compression") {
        compressor = std::make_unique<clp::streaming_compression::passthrough::Compressor>();
        compress(std::move(compressor), uncompressed_buffer.data(), cCompressionChunkSizes);
        decompressor = std::make_unique<clp::streaming_compression::passthrough::Decompressor>();
        decompress_and_compare(std::move(decompressor), uncompressed_buffer, decompressed_buffer);
    }

    SECTION("LZMA compression") {
        compressor = std::make_unique<clp::streaming_compression::lzma::Compressor>();
        compress(std::move(compressor), uncompressed_buffer.data(), cCompressionChunkSizes);
        decompressor = std::make_unique<clp::streaming_compression::lzma::Decompressor>();
    }

    boost::filesystem::remove(string(cCompressedFilePath));
}

TEST_CASE("SmallZstdFile", "[SmallZstdFile]") {
    constexpr size_t cAlphabetLength{26};

    std::unique_ptr<Compressor> compressor;
    std::unique_ptr<Decompressor> decompressor;
    Array<char> decompressed_buffer(cAlphabetLength);
    Array<char> uncompressed_buffer(cAlphabetLength);
    fill_buffer(uncompressed_buffer, cAlphabetLength);

    SECTION("ZStd single phase compression") {
        compressor = std::make_unique<clp::streaming_compression::zstd::Compressor>();
        auto const single_chunk{std::to_array<size_t>({cAlphabetLength})};
        compress(std::move(compressor), uncompressed_buffer.data(), single_chunk);

        decompressor = std::make_unique<clp::streaming_compression::zstd::Decompressor>();
        clp::ReadOnlyMemoryMappedFile const memory_mapped_compressed_file{
                string(cCompressedFilePath)
        };
        auto const compressed_file_view{memory_mapped_compressed_file.get_view()};
        decompressor->open(
                compressed_file_view.data(),
                // Truncate the EOF byte to let the decompressor consume the whole first frame on
                // the first read.
                compressed_file_view.size() - 1
        );

        // Suppose after the truncate, the decompressor can still read half of the first frame.
        constexpr auto cValidReadLength{cAlphabetLength / 2};
        for (size_t i{0}; i < cValidReadLength; ++i) {
            REQUIRE(
                    (ErrorCode_Success
                     == decompressor->try_read_exact_length(decompressed_buffer.begin() + i, 1))
            );
        }
        REQUIRE(std::equal(
                uncompressed_buffer.begin(),
                uncompressed_buffer.begin() + cValidReadLength,
                decompressed_buffer.begin()
        ));
    }
}
```

components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
```diff
@@ -34,67 +34,38 @@ auto Decompressor::try_read(char* buf, size_t num_bytes_to_read, size_t& num_byt
...
-                    break;
+        if (m_compressed_stream_block.pos == m_compressed_stream_block.size
+            && false == m_zstd_frame_has_more_data)
+        {
+            auto const error_code{refill_compressed_stream_block()};
+            if (ErrorCode_Success != error_code) {
+                m_decompressed_stream_pos += decompressed_stream_block.pos;
+
+                num_bytes_read = decompressed_stream_block.pos;
+                if (ErrorCode_EndOfFile == error_code && decompressed_stream_block.pos > 0) {
```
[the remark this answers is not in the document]
kirkrodrigues: @junhaoliao I think the rabbit might be correct about this? If `Zstd_decompressStream` exhausts `m_compressed_stream_block` but `Zstd_decompressStream` still returns `> 0`, then we should try to read more from the underlying data, right?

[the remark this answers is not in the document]
junhaoliao: https://facebook.github.io/zstd/zstd_manual.html#Chapter9
> `@return` ...
any other value > 0, which means there is still some decoding or flushing to do to complete current frame

assuming `ZSTD_decompressStream()` implementation is correct, it should return > 0 only if a future call of it is allowed (and expected). in that case, we don't have to load new data into the input buffer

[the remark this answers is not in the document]
kirkrodrigues: I understand your interpretation (I think that's what I thought initially too), but unfortunately I think they mean it very literally:

* if the return is `>0`, then the frame hasn't been fully decoded. If at this point the input buffer has been exhausted, then the only way to decode more of the frame would be to provide more input---which is why the return value is a hint for how much input to provide.
* if `output.pos == output.size`, then there might be more data that Zstd_decompressStream can provide, *without* providing more input data.

I think this sentence also agrees with that interpretation:

> The function tries to flush all data decoded immediately, respecting output buffer size.

Fwiw, I wrote a simple unit test to check the interpretation and it does seem like we'll get into an infinite loop if we hit this case.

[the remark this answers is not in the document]
junhaoliao: > if output.pos == output.size, then there might be more data that Zstd_decompressStream can provide, without providing more input data.

i updated 
```
m_zstd_frame_has_more_data = decompressed_stream_block.pos == decompressed_stream_block.size;
```
which now strictly follows the recommendation of the zstd manual

components/core/src/clp/streaming_compression/zstd/Decompressor.hpp
```diff
@@ -142,6 +151,7 @@ class Decompressor : public ::clp::streaming_compression::Decompressor {
     ZSTD_inBuffer m_compressed_stream_block{};
 
     size_t m_decompressed_stream_pos{0ULL};
+    bool m_zstd_frame_has_more_data{false};
```
junhaoliao: more accurately, this should be `m_zstd_frame_might_has_more_data`

components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
```diff
@@ -34,67 +34,39 @@ auto Decompressor::try_read(char* buf, size_t num_bytes_to_read, size_t& num_byt
 
     num_bytes_read = 0;
 
-    ZSTD_outBuffer decompressed_stream_block = {buf, num_bytes_to_read, 0};
+    ZSTD_outBuffer decompressed_stream_block{buf, num_bytes_to_read, 0};
     while (decompressed_stream_block.pos < num_bytes_to_read) {
         // Check if there's data that can be decompressed
```
kirkrodrigues:
```suggestion
```

Comment feels unnecessary and slightly misleading now.

components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
```diff
@@ -34,67 +34,39 @@ auto Decompressor::try_read(char* buf, size_t num_bytes_to_read, size_t& num_byt
...
-                    m_compressed_stream_block.size = m_read_buffer_length;
-                    break;
+        if (m_compressed_stream_block.pos == m_compressed_stream_block.size
+            && false == m_zstd_frame_has_more_data)
+        {
+            auto const error_code{refill_compressed_stream_block()};
+            if (ErrorCode_Success != error_code) {
+                m_decompressed_stream_pos += decompressed_stream_block.pos;
+
+                num_bytes_read = decompressed_stream_block.pos;
```
kirkrodrigues:
```suggestion
                m_decompressed_stream_pos += decompressed_stream_block.pos;
                num_bytes_read = decompressed_stream_block.pos;

```

Just feels like those two operations should be grouped together.

components/core/src/clp/streaming_compression/zstd/Decompressor.hpp
kirkrodrigues: I agree. Let's use `m_zstd_frame_might_have_more_data` (grammar).

kirkrodrigues (changes requested):
Let's defer the unit tests to another PR where we can make them more exhaustive and robust.

For the PR title, how about:

```
fix(core): Try to exhaust Zstd's internal buffers when they might contain unconsumed data. (fixes #976).
```

components/core/src/clp/streaming_compression/zstd/Decompressor.cpp
junhaoliao: i thought we intentionally grouped the num_bytes_read assignment and the return statement together since `num_bytes_read` is an output of the method. shall we also group the ones at the end of this method body?

junhaoliao: for now i also made the changes at the bottom of the method and removed the "Update decompression stream position" comment which i feel is unnecessary. we can revert if you disagree

kirkrodrigues (approved):

junhaoliao:
> Let's defer the unit tests to another PR where we can make them more exhaustive and robust.

@coderabbitai create an issue to track the unit tests improvement

PR #998, opened 2025-06-10
fix(package): Prevent remapping of host filesystem paths when mounting output directory in CLP execution containers. (fixes #960)

### LINKED ISSUE #960: Don't remap host filesystem paths when mounting them in CLP's execution containers
### Request

Input and output directories that the CLP package operates on are mounted into the container as follows:

* The directory set via the `logs_input.directory` config gets mounted to `/mnt/logs/<logs_input.directory>`
* The directory set via the `archive_output.storage.directory` config gets mounted to `/mnt/archive-output`
* The directory set via the `stream_output.storage.directory` config gets mounted to `/mnt/stream-output`
* The directory set via the `logs_directory` config gets mounted to `/opt/clp/var/log`
* The directory set via the `data_directory` config gets mounted to `/opt/clp/var/data`

Although this provides some level of obfuscation of host paths, it presents some challenges for users:

* Paths printed to the console are relative to the container mount rather than the host filesystem (see #602).
* External users (e.g., Presto) of the datasets table (see #868) will see a container-relative path (e.g., /mnt/archive-output/dataset1) and will then need to translate that to a path on the filesystem.
  * This problem is exacerbated if the external user is also running in its own container.

So the proposal is to simply mount the host filesystem paths specified by these configs directly into the container, without remapping them to a different path. Obviously, we need to ensure that the paths specified in the config file *can* be mapped into the container without causing issues, but that should be doable since we control what container the package runs in.

### Possible implementation

* Validate that the paths specified in the config file don't overlap with paths that exist in the container (e.g., ensure they don't overlap with `/opt`).
  * The one exception to this rule would be `logs_input.directory` since it's very likely that a user might configure `/var/log` or `/tmp` as their input directory and we may not be able to mount this into the container. For this case, I think we can handle continue to mount it into the container as we do currently and document the behaviour---since we retain the full input path, but just prefix it, the user shouldn't get too confused. In the future, we could potentially ensure that when we print out the path, we remove the `/mnt/logs-input` prefix.
* Mount the configured paths directly into the container without remapping them.

haiqi96:
# Description

# Checklist

  breaking change.

# Validation performed

### DIFF
--- a/components/clp-package-utils/clp_package_utils/general.py
+++ b/components/clp-package-utils/clp_package_utils/general.py
@@ -227,7 +227,6 @@ def generate_container_config(
             DockerMountType.BIND, input_logs_dir, container_clp_config.logs_input.directory, True
         )
 
-    container_clp_config.data_directory = CONTAINER_CLP_HOME / "var" / "data"
     if not is_path_already_mounted(
         clp_home, CONTAINER_CLP_HOME, clp_config.data_directory, container_clp_config.data_directory
     ):
@@ -235,7 +234,6 @@ def generate_container_config(
             DockerMountType.BIND, clp_config.data_directory, container_clp_config.data_directory
         )
 
-    container_clp_config.logs_directory = CONTAINER_CLP_HOME / "var" / "log"
     if not is_path_already_mounted(
         clp_home, CONTAINER_CLP_HOME, clp_config.logs_directory, container_clp_config.logs_directory
     ):
@@ -243,7 +241,6 @@ def generate_container_config(
             DockerMountType.BIND, clp_config.logs_directory, container_clp_config.logs_directory
         )
 
-    container_clp_config.archive_output.set_directory(pathlib.Path("/") / "mnt" / "archive-output")
     if not is_path_already_mounted(
         clp_home,
         CONTAINER_CLP_HOME,
@@ -256,7 +253,6 @@ def generate_container_config(
             container_clp_config.archive_output.get_directory(),
         )
 
-    container_clp_config.stream_output.set_directory(pathlib.Path("/") / "mnt" / "stream-output")
     if not is_path_already_mounted(
         clp_home,
         CONTAINER_CLP_HOME,
@@ -368,6 +364,9 @@ def load_config_file(
 
     clp_config.make_config_paths_absolute(clp_home)
     clp_config.load_execution_container_name()
+
+    validate_path_for_container_mount(clp_config.data_directory)
+    validate_path_for_container_mount(clp_config.logs_directory)
 
     # Make data and logs directories node-specific
     hostname = socket.gethostname()
@@ -509,6 +508,9 @@ def validate_worker_config(clp_config: CLPConfig):
     clp_config.validate_archive_output_config()
     clp_config.validate_stream_output_config()
 
+    validate_path_for_container_mount(clp_config.archive_output.get_directory())
+    validate_path_for_container_mount(clp_config.stream_output.get_directory())
+
 
 def validate_webui_config(
     clp_config: CLPConfig, logs_dir: pathlib.Path, settings_json_path: pathlib.Path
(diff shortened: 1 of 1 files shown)


components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -509,6 +508,9 @@ def validate_worker_config(clp_config: CLPConfig):
     clp_config.validate_archive_output_config()
     clp_config.validate_stream_output_config()
 
+    validate_path_for_container_mount(clp_config.archive_output.get_directory())
+    validate_path_for_container_mount(clp_config.stream_output.get_directory())
+
```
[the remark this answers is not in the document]
haiqi96: when storage type is S3, the get_directory() will return the staging directory so it's not an issue

haiqi96:
Let me know if you think there's any docs to update.

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -509,6 +508,9 @@ def validate_worker_config(clp_config: CLPConfig):
     clp_config.validate_archive_output_config()
     clp_config.validate_stream_output_config()
 
+    validate_path_for_container_mount(clp_config.archive_output.get_directory())
+    validate_path_for_container_mount(clp_config.stream_output.get_directory())
```
kirkrodrigues, line 513: I think these are technically redundant since this function is only called from start_clp after the config is loaded, right?

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -537,3 +539,33 @@ def validate_log_viewer_webui_config(clp_config: CLPConfig, settings_json_path:
...
+        CONTAINER_CLP_HOME,
+        CONTAINER_INPUT_LOGS_ROOT_DIR,
+    ]
+
+    if not path.is_absolute():
+        raise ValueError(f"Path: `{path}` must be absolute:")
+
+    for prefix in RESTRICTED_PREFIXES:
+        if path.is_relative_to(prefix):
+            raise ValueError(f"Invalid path: `{path}` is under the restricted prefix '{prefix}'.")
```
```suggestion
            raise ValueError(
                f"Invalid path: `{path}` cannot be under '{prefix}' which may overlap with a path"
                f" in the container."
            )
```

components/clp-package-utils/clp_package_utils/general.py
haiqi96: Why? these functions are not called anywhere else though.
Do you mean put them inside `validate_archive_output_config()` and `validate_stream_output_config`?

haiqi96:
I don't think we need to check `aws_config_directory` since
1. It is a read only mount, so we don't write to it
2. It has to be mounted  to a specific path in the container to make it work for the default user.

haiqi96:
@kirkrodrigues how about
`# Note: This path cannot (or must not?) overlap with any path used in clp's execution container.`

kirkrodrigues:
> I don't think we need to check `aws_config_directory` since
> 
>     1. It is a read only mount, so we don't write to it
> 
>     2. It has to be mounted  to a specific path in the container to make it work for the default user.

Wouldn't we need to guard against someone setting the data/logs/archive/stream output path to the same path?

components/clp-package-utils/clp_package_utils/general.py
kirkrodrigues: Sorry, didn't notice they're different directories.

kirkrodrigues:
> @kirkrodrigues how about `# Note: This path cannot (or must not?) overlap with any path used in clp's execution container.`

```yaml
# NOTE: This directory must not overlap with any path used in CLP's execution container. An error
# will be raised if so.
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -537,3 +540,37 @@ def validate_log_viewer_webui_config(clp_config: CLPConfig, settings_json_path:
...
+        pathlib.Path("/root"),
+        pathlib.Path("/run"),
+        pathlib.Path("/sbin"),
+        pathlib.Path("/srv"),
+        pathlib.Path("/sys"),
+        pathlib.Path("/usr"),
+        pathlib.Path("/var"),
+        CONTAINER_CLP_HOME,
+        CONTAINER_INPUT_LOGS_ROOT_DIR,
+        CONTAINER_AWS_CONFIG_DIRECTORY,
```
kirkrodrigues: Sorry, should've mentioned this in the previous review, but can we alphabetize these?

PR #995, opened 2025-06-10
feat(new-webui): Support loading injected client-side settings.json at app startup.

hoophalab:
# Description

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

- **Chores**
	- Updated configuration loading to use a TypeScript module for application settings instead of direct JSON imports.
	- Improved error handling when loading settings from the server.
	- Removed obsolete comments related to configuration management.

PR #1003, opened 2025-06-12
fix(new-webui): Support top-level await by updating browser compatibility target in Vite config.

davemarco:
# Description

#995 introduced a top-level await to support settings loading.

However, the vite config was not configured for this leading to an error during build.

This PR changed the vite config to support top-level await on modern browsers.

# Checklist

  breaking change.

# Validation performed

App now builds. 

### DIFF
--- a/components/log-viewer-webui/client/vite.config.ts
+++ b/components/log-viewer-webui/client/vite.config.ts
@@ -7,6 +7,9 @@ import {defineConfig} from "vite";
 // https://vite.dev/config/
 export default defineConfig({
     base: "./",
+    build: {
+        target: "esnext",
+    },
     plugins: [
         react(),
     ],


components/log-viewer-webui/client/vite.config.ts
```diff
@@ -7,6 +7,9 @@ import {defineConfig} from "vite";
 // https://vite.dev/config/
 export default defineConfig({
     base: "./",
+    build: {
+        target: "esnext",
```
junhaoliao, line 11: https://esbuild.github.io/api/#target
> `esnext` which means that by default, esbuild will assume all of the latest JavaScript and CSS features are supported.

seems that's exactly what we want since we use / will use many latest features

davemarco:
npm looks to be down, so ci failing

PR #1005, opened 2025-06-13
feat(clp-s): Expose archive range index metadata in printed archive statistics; Forward archive-level statistics to c++ writers.

gibber9809:
# Description

This PR exposes the contents of the archive range index (including filename, etc.) in the printed archive statistics to enable the package and other users to index this kind of archive-level metadata. Additionally, we refactor how archive-level statistics are represented internally and forward this new representation to the write boundaries of the c++ code such that users of our write-side library can work with archive-level statistics without having to hack around with the printed statistics or read newly-written archives.

Besides including the archive-level meta in the archive-level statistics we also add an "is_split" flag which indicates whether the last file ingested into an archive has content continuing into another archive.

# Checklist

  breaking change.

# Validation performed
* Added test for forwarded archive stats
* Manually confirmed that printed archive stats contain expected content

## Summary by CodeRabbit

- **New Features**
  - Archive statistics, including metadata and range index information, are now returned and available after archive creation.
  - Enhanced reporting of archive statistics in a structured JSON format.

- **Bug Fixes**
  - Improved handling and validation of archive metadata and range index entries in tests.

- **Refactor**
  - Unified archive statistics and metadata handling for consistency across the application.
  - Modularized and extended test logic for archive metadata validation.

- **Chores**
  - Updated method signatures and internal logic to support new statistics and metadata features.
  - Made explicit discarding of unused return values to improve code clarity.
  - Adjusted subprocess output decoding to UTF-8 for improved compatibility.

### DIFF
--- a/components/core/src/clp_s/ArchiveWriter.hpp
+++ b/components/core/src/clp_s/ArchiveWriter.hpp
@@ -55,7 +55,6 @@ public:
 
     // Methods
     /**
-     * Dumps the contents of `ArchiveStats` as a JSON object in a string.
      * @return The contents of `ArchiveStats` as a JSON object in a string.
      */
     [[nodiscard]] auto as_string() const -> std::string {
--- a/components/core/src/clp_s/RangeIndexWriter.hpp
+++ b/components/core/src/clp_s/RangeIndexWriter.hpp
@@ -74,9 +74,6 @@ public:
 
     /**
      * Writes ranges to a `ZstdCompressor` then clears internal state.
-     *
-     * TODO: refactor this method to return the metadata as a `ystdlib::error_handling::Result`.
-     *
      * @param writer
      * @param metadata The metadata written to the `ZstdCompressor` returned by reference.
      * @return ErrorCodeSuccess on success or the relevant error code on failure.


components/core/src/clp_s/ArchiveWriter.hpp
```diff
@@ -31,6 +33,49 @@ struct ArchiveWriterOption {
...
 };
 
+struct ArchiveStats {
+    explicit ArchiveStats(
+            std::string id,
+            epochtime_t begin_timestamp,
+            epochtime_t end_timestamp,
+            size_t uncompressed_size,
+            size_t compressed_size,
+            nlohmann::json metadata,
```
wraymo: Seems it's only for range index? Should we call it `metadata`?

components/core/src/clp_s/ArchiveWriter.hpp
```diff
@@ -80,9 +125,11 @@ class ArchiveWriter {
     void open(ArchiveWriterOption const& option);
 
     /**
-     * Closes the archive writer
+     * Closes the archive writer.
+     * @param is_split Whether the last file ingested into the archive is split.
```
wraymo, line 150: I guess every archive except the last one is a split? Why do we need this information?

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -1140,19 +1140,20 @@ auto JsonParser::parse_from_ir() -> bool {
...
         }
     }
     return true;
 }
 
-void JsonParser::store() {
-    m_archive_writer->close();
+auto JsonParser::store() -> std::vector<ArchiveStats> {
+    m_archive_stats.emplace_back(m_archive_writer->close());
+    return std::move(m_archive_stats);
```
wraymo, line 1152: Do we actually use this return value except for the test case?

wraymo (commented):
Thanks for improving archive statistics! Just have a few questions

components/core/src/clp_s/JsonParser.cpp
gibber9809: Not right now, but I figure it will be useful when we write c++ spider tasks to do ingestion.

components/core/src/clp_s/ArchiveWriter.hpp
gibber9809: I think it could be helpful for error handling in some scenarios -- e.g. if some file gets split across two archives and compression fails while the second archive is being compressed we can use this flag to tell that the last file in the first archive hasn't been fully ingested and handle the error appropriately.

components/core/src/clp_s/ArchiveWriter.hpp
gibber9809: Right, `range_index` is probably a better name.

components/core/src/clp_s/ArchiveWriter.hpp
wraymo: How would looking at the printed `ArchiveStats` help in this case? As I understand it, regardless of whether an error occurs or not, the last archive is never a split.

gibber9809: Say a user is ingesting files A and B and we produce two archives, where B is split between the first and the second archive. Consider the scenario where the second archive fails compression such that no archive stats get printed for the second archive.

In this scenario, without the is_split flag, the user can see that the first archive contains the files "A" and "B", but has no reliable way of determining that "B" is not fully ingested. However, if they do have the "is_split" flag then they know that they _should_ expect at least one more archive containing the rest of "B".

Right now, since we still close/print archive stats for failed archives it's not as useful, but if we improve the error handling in `JsonConstructor` I think it could be pretty useful. The c++ code-path doesn't have this issue since we're only forwarding archive stats for the succesful archives.

wraymo: Do we have a direct way to figure out which file is split across two archives if one archive contains multiple files

wraymo: Oh the range index is part of the statistics

wraymo (dismissed):

components/core/src/clp_s/ArchiveWriter.cpp
```diff
@@ -107,8 +107,17 @@ void ArchiveWriter::close() {
...
+            m_timestamp_dict.get_begin_timestamp(),
+            m_timestamp_dict.get_end_timestamp(),
+            m_uncompressed_size,
+            m_compressed_size,
+            archive_range_index,
+            is_split
+    };
     if (m_print_archive_stats) {
-        print_archive_stats();
+        std::cout << archive_stats.as_string() << std::endl;
```
kirkrodrigues:
```suggestion
        std::cout << archive_stats.as_string() << '\n';
```

To resolve clang-tidy violation.

components/core/src/clp_s/ArchiveWriter.hpp
```diff
@@ -31,6 +33,49 @@ struct ArchiveWriterOption {
...
+        return json_msg.dump(-1, ' ', false, nlohmann::json::error_handler_t::ignore);
+    }
+
+    std::string const id;
+    epochtime_t const begin_timestamp{};
+    epochtime_t const end_timestamp{};
+    size_t const uncompressed_size{};
+    size_t const compressed_size{};
+    nlohmann::json const range_index;
+    bool const is_split{};
```
kirkrodrigues: I guess the reason you're const qualifying these variables is so that they're not independently modified by a caller? clang-tidy understandably has concerns about const-qualified members, so I wonder if we should just convert `ArchiveStats` to a full-fledged class so that these members are private? Is any caller actually using these members directly?

components/core/src/clp_s/RangeIndexWriter.hpp
```diff
@@ -9,6 +9,7 @@
 #include <vector>
 
 #include <nlohmann/json.hpp>
+#include <ystdlib/error_handling/ErrorCode.hpp>
```
kirkrodrigues: Is this necessary in this PR?

components/core/src/clp_s/RangeIndexWriter.hpp
```diff
@@ -74,10 +75,14 @@ class RangeIndexWriter {
 
     /**
      * Writes ranges to a `ZstdCompressor` then clears internal state.
+     *
+     * TODO: refactor this method to return the metadata as a `ystdlib::error_handling::Result`.
```
kirkrodrigues: TODOs generally get forgotten so we prefer to open issues in the repo. That also functions as a good entry point for new contributors. So can we convert this to an issue?

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -346,7 +346,7 @@ def run_clp(
         if not line:
             last_line_decoded = True
         else:
-            stats = json.loads(line.decode("ascii"))
+            stats = json.loads(line.decode("utf-8"))
```
kirkrodrigues, line 349: Out of interest, what motivated this change?

kirkrodrigues (changes requested):
A few comments after skimming the PR.

components/core/src/clp_s/ArchiveWriter.hpp
gibber9809: Yeah, this was a rabbit suggestion. I agree it makes sense to just turn it into a full class at this point though.

components/core/src/clp_s/RangeIndexWriter.hpp
gibber9809: No, will remove.

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
gibber9809: Rabbit suggested just using utf-8 output instead of using the escape to ascii approach because it makes utf-8 content human readable, which I agree with. Since I'm dumping the range index into this metadata now we can expect it to contain some user-inserted kv-pairs, and that will probably be easier for users to spot-check if unicode content renders as unicode characters.

components/core/src/clp_s/RangeIndexWriter.hpp
gibber9809: Put it up as #1025.

components/core/src/clp_s/ArchiveWriter.hpp
```diff
@@ -31,6 +33,71 @@ struct ArchiveWriterOption {
...
+              m_begin_timestamp{begin_timestamp},
+              m_end_timestamp{end_timestamp},
+              m_uncompressed_size{uncompressed_size},
+              m_compressed_size{compressed_size},
+              m_range_index(std::move(range_index)),  // Avoid {} to prevent wrapping in JSON array.
+              m_is_split{is_split} {}
+
+    // Methods
+    /**
+     * Dumps the contents of `ArchiveStats` as a JSON object in a string.
```
kirkrodrigues: Nit: Seems redundant.
```suggestion
```

components/core/src/clp_s/RangeIndexWriter.hpp
kirkrodrigues: Do we still need the TODO comment?

PR #1026, opened 2025-06-23
fix(clp-s)!: Do not record the archive range index when log-ordering is disabled.

gibber9809:
# Description

Without log-ordering the archive range index has limited usefulness in that the metadata in the range index can not reliably be associated with specific records in the archive. Currently, we still record the archive range index even when log-ordering is disabled. This could sometimes be useful (e.g. if an archive consists of a single range), but probably isn't worth supporting unless users specifically request it.

This PR is marked "fix" because the archive range index was intended to be an optional section partially because of this situation.

# Checklist

  breaking change.

# Validation performed

* Manually validated that an archive produced with the `--disable-log-order` flag does not contain a `RangeIndex` metadata packet.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of log order recording: metadata fields and archive range closing now occur only when log order recording is enabled, reducing unnecessary operations when disabled.

- **Documentation**
  - Enhanced the help text for the "disable-log-order" option to clarify it also disables recording the archive range index.


wraymo (approved):
LGTM

wraymo (approved):

PR #1033, opened 2025-06-25
build: Update to the latest version of log-surgeon.

SharafMohamed:
# Description
#### Reason for update:
- Allows for the use of capture-groups.
- Several bug-fixes inside log-surgeon.

#### Code changes:
- Use new lexer `scan` API (now returns an optional token, instead of taking a token by reference).
- Update to match log-surgeon renames:
  - Change all instances of enum class `SymbolID` to `SymbolId`, and update corresponding enum names (drop the `Id`/`ID` suffix).
  - Drop `Regex` prefix for `ByteNfaState`.
  - Change lexer method `add_delimiters` to `set_delimiters`.
- Update timestamps in `schema.txt` and update `equals` variable to use capture groups.
- Remove all instances of `reverse_lexer` as its no longer supported by `log-surgeon`. As a result, fall back to grepping all files if search variable has a leading wildcard.
- Rename all instances of `forward_lexer` to just `lexer`.
- Update `is_possible_input` to use an `std::array` in place of a c-style array.

### Unit-test changes
- `>` is now in the action table of `int`, so is a suggested character to proceed ` ` in the failure test.
- `?` is now used in capture groups, so has to be escaped in the schema.

### Taskfile changes
- Add a task to install `microsoft.gsl`.
- Update `log-surgeon` task to: 
  - depend on `fmt` and `microsoft.gsl`.
  - disable testing.
  - use the newest version.

# Validation performed
- Updated unit-tests pass.
- CI passes.

## Summary by CodeRabbit

* **New Features**
  * Expanded support for a wide variety of timestamp formats in schema configuration.
  * Added Microsoft GSL as a supported dependency.

* **Bug Fixes**
  * Improved handling of delimiters and special characters in schema files.
  * Corrected error messages and token end symbol usage in schema parser tests.
  * Added runtime error handling for timestamp encoding failures.

* **Refactor**
  * Updated naming conventions for token type enums to improve clarity.
  * Streamlined lexer loading by removing unnecessary boolean parameters and simplifying function signatures.
  * Consolidated lexer usage from separate forward and reverse lexers to a single lexer instance across core components.
  * Improved internal data structures for lexer utilities.

* **Chores**
  * Updated dependency management configuration and library versions.


components/core/CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ string(LENGTH "${CMAKE_SOURCE_DIR}/" SOURCE_PATH_SIZE)
 add_definitions("-DSOURCE_PATH_SIZE=${SOURCE_PATH_SIZE}")
 
 # Profiling options
-add_definitions(-DPROF_ENABLED=0)
+add_definitions(-DPROF_ENABLED=1)
```
davidlion: I'm guessing this is accidentally left on ya?

components/core/src/clp/Grep.cpp
```diff
@@ -866,21 +866,25 @@ bool Grep::get_bounds_of_next_potential_var(
...
                 // auto const& set = search_token.m_type_ids_set;
-                // if (set.find(static_cast<int>(log_surgeon::SymbolID::TokenUncaughtStringID))
+                // if (set.find(static_cast<int>(log_surgeon::SymbolId::TokenUncaughtStringID))
                 //            == set.end()
-                //     && set.find(static_cast<int>(log_surgeon::SymbolID::TokenEndID))
+                //     && set.find(static_cast<int>(log_surgeon::SymbolId::TokenEndID))
                 //            == set.end())
                 // {
                 //     is_var = true;
                 // }
```
davidlion: Let's remove this comment and document it somewhere for when we do the performance benchmarking.
@kirkrodrigues do you think it makes sense to have a separate meta issue in clp tracking log-surgeon performance todos?

davidlion (changes requested):

components/core/src/clp/Grep.cpp
kirkrodrigues: Yeah, makes sense.

davidlion: @SharafMohamed can you do the same thing you did for log-surgeon. (The difference being these performance issues are related to the code using log-surgeon, not just a call to log-surgeon itself.)

SharafMohamed: This TODO should be obsolete with the new search PR. I'll remove it, but I don't think we need to track it.

components/core/config/schemas.txt
```diff
@@ -1,19 +1,97 @@
...
 int:\-{0,1}[0-9]+
 float:\-{0,1}[0-9]+\.[0-9]+
 
+// ===
 // Dictionary variables
+// ===
+
 hex:[a-fA-F]+
+
+// Note: the way `equals` and `hasNumber` are written only work if `=` is a delim
```
davidlion: @SharafMohamed Can you clarify this? I can sort of imagine something since `equals` has `=` in it, but I don't understand why `hasNumber` won't work without `=` as a delim.

components/core/config/schemas.txt
```diff
@@ -1,19 +1,97 @@
-// Delimiters
-delimiters: \t\r\n!"#$%&'\(\)\*,:;<>?@\[\]\^_`\{\|\}~
+// ===
+// Timestamps
+// ===
+
+// E.g. 2015-01-31T15:50:45.392
+timestamp:\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}
```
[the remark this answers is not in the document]
SharafMohamed: Fixed.

components/core/config/schemas.txt
```diff
@@ -1,19 +1,97 @@
...
 
+// ===
 // Dictionary variables
+// ===
+
 hex:[a-fA-F]+
+
+// Note: the way `equals` and `hasNumber` are written only work if `=` is a delim
+equals:.*=(?<val>.*[a-zA-Z0-9].*)
 hasNumber:.*\d.*
```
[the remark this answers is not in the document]
SharafMohamed: `.` is only treated as non delimiter characters in log-surgeon.

components/core/config/schemas.txt
```diff
@@ -1,19 +1,97 @@
...
+
+// E.g. 2015/01/31 15:50:45
+timestamp:\d{4}/\d{2}/\d{2} \d{2}:\d{2}:\d{2}
+
+// E.g. 15/01/31 15:50:45
+timestamp:\d{2}/\d{2}/\d{2} \d{2}:\d{2}:\d{2}
+
+// E.g. 150131  9:50:45
+timestamp:\d{2}\d{2}\d{2} [ 0-9]{2}:\d{2}:\d{2}
+
```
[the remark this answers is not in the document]
SharafMohamed: Leading spaces are allowed, but log-surgeon doesn't support that style of regex you recommended.

components/core/config/schemas.txt
```diff
@@ -1,19 +1,98 @@
-// Delimiters
-delimiters: \t\r\n!"#$%&'\(\)\*,:;<>?@\[\]\^_`\{\|\}~
+// ===
+// Timestamps
```
kirkrodrigues: Does this cover all the known timestamp patterns in `TimestampPattern.cpp`?

components/core/config/schemas.txt
```diff
@@ -1,19 +1,98 @@
...
+// ===
+
+// E.g. 2015-01-31T15:50:45.392
+timestamp:\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}
```
kirkrodrigues: Everywhere we use `T`, technically we could also support a space. Same for periods and commas. So:

```suggestion
// E.g. 2015-01-31T15:50:45.392
// E.g. 2015-01-31T15:50:45,392
// E.g. 2015-01-31 15:50:45.392
// E.g. 2015-01-31 15:50:45,392
timestamp:\d{4}\-\d{2}\-\d{2}[T ]\d{2}:\d{2}:\d{2}[,\.]\d{3}
```

Unless that causes any issues?

components/core/config/schemas.txt
```diff
@@ -1,19 +1,98 @@
...
+timestamp:\d{2}\-\d{2} \d{2}:\d{2}:\d{2}\.\d{3}
+
+// E.g. 2016-05-08 11:34:04.083464
+timestamp:\d{4}\-\d{2}\-\d{2} \d{2}:\d{2}:\d{2}\.\d{6}
+
+// ===
+// Delimiters
+// ===
+
+delimiters: \t\r\n!"#$%&'\(\)\*,:;\<\>\?@\[\]\^_`\{\|\}~=
```
kirkrodrigues:
```suggestion
delimiters: \t\r\n!"#$%&'\(\)\*,:;\<=\>\?@\[\]\^_`\{\|\}~
```

To maintain ASCII ordering. If we don't want to maintain ASCII ordering, that's fine, but we can probably come up with a better ordering than the current (e.g., all brackets, parentheses, and braces together).

components/core/config/schemas.txt
```diff
@@ -1,19 +1,98 @@
...
 int:\-{0,1}[0-9]+
 float:\-{0,1}[0-9]+\.[0-9]+
 
+// ===
 // Dictionary variables
+// ===
+
 hex:[a-fA-F]+
+
+// Note: the way `equals` and `hasNumber` are written only work if `=` is a delimiter, due to
```
kirkrodrigues:
```suggestion
// NOTE: `equals` and `hasNumber` only work if `=` is a delimiter, due to
```

components/core/src/clp/clo/clo.cpp
```diff
@@ -476,17 +476,12 @@ static bool search_archive(
...
     if (std::filesystem::exists(schema_file_path)) {
         use_heuristic = false;
-        // Create forward lexer
-        forward_lexer.reset(new log_surgeon::lexers::ByteLexer());
-        load_lexer_from_file(schema_file_path.string(), false, *forward_lexer);
-
-        // Create reverse lexer
-        reverse_lexer.reset(new log_surgeon::lexers::ByteLexer());
-        load_lexer_from_file(schema_file_path.string(), true, *reverse_lexer);
+        lexer.reset(new log_surgeon::lexers::ByteLexer());
```
kirkrodrigues: Can we use `make_unique`?

components/core/src/clp/clg/clg.cpp
```diff
@@ -592,46 +587,24 @@ int main(int argc, char const* argv[]) {
...
 
         // Perform search
-        if (!search(search_strings,
-                    command_line_args,
-                    archive_reader,
-                    *forward_lexer_ptr,
-                    *reverse_lexer_ptr,
-                    use_heuristic))
-        {
+        if (!search(search_strings, command_line_args, archive_reader, *lexer_ptr, use_heuristic)) {
```
kirkrodrigues, line 594: If I'm not mistaken, `*lexer_ptr` could dereference a garbage pointer if no schema file exists. This bug wasn't introduced by this PR, so we could fix it in a new PR... but we should definitely fix it.

kirkrodrigues: The same applies in `clo.cpp`.

davidlion: @coderabbitai Please create an issue to investigate and refactor the construction of lexers and how they are passed through cl{g,o,p}'s API with `use_heuristic`.

components/core/src/clp/clo/clo.cpp
```diff
@@ -476,17 +476,11 @@ static bool search_archive(
...
         use_heuristic = false;
-        // Create forward lexer
-        forward_lexer.reset(new log_surgeon::lexers::ByteLexer());
-        load_lexer_from_file(schema_file_path.string(), false, *forward_lexer);
-
-        // Create reverse lexer
-        reverse_lexer.reset(new log_surgeon::lexers::ByteLexer());
-        load_lexer_from_file(schema_file_path.string(), true, *reverse_lexer);
+        load_lexer_from_file(schema_file_path.string(), *lexer);
     }
```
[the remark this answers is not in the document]
SharafMohamed: Changed.

components/core/src/clp/clo/clo.cpp
SharafMohamed: We swapped it to use a `ByteLexer` object directly instead of a `unique_ptr`.

components/core/config/schemas.txt
davidlion: It is best effort, but all patterns in `TimestampPattern.cpp` should be in the schema now. The timestamps requiring log boundary knowledge are categorized separately and will be addressed in the future.

kirkrodrigues (approved):

davidlion (approved):

PR #1023, opened 2025-06-20
fix(package): Refactor dataset handling in the package.

haiqi96:
# Description

This PR refactored the following:
1. added shared method for getting the names of the tables in the MySQL metadata database.
2. updated s3_put methods to support specifying a relative path as upload destination.

The PR also fixes a bug where old webui fails to load before the first CLP-S compression job
- The current design does not create the tables with default dataset until a compression job with dataset = default is created.
- The old webui querying log statistics from the tables with the default dataset, but those tables are not created when package starts. 

As a result, the old webui will shown an empty page before the first compression job is launched. This is fixed by creating the set of tables with dataset = default when the package starts, at [here](https://github.com/y-scope/clp/pull/1023/files#diff-8b30e73ca3e57a2797b26e898193bb8c1ed598e64fa7d84f04695d8113ad460dR60)

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

Manually tested:
1. CLP & CLP-S when using local filesystem as archive storage.
2. CLP & CLP-S when using S3 as archive storage.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **Refactor**
  * Centralized and standardized database table name construction using dynamic helper functions instead of hardcoded suffixes.
  * Updated dataset handling across multiple modules for improved consistency and modularity, including removal of storage engine parameters and related conditional logic.
  * Simplified and clarified S3 object key construction and dataset path handling without mutating configuration objects.
  * Modified function and class signatures to accept optional dataset parameters where applicable.
  * Encapsulated dataset addition and caching logic for improved dataset management.
  * Introduced conditional dataset assignment in search and query job configurations based on storage engine settings.
  * Added explicit storage engine configuration entry and related constants for unified dataset handling in the web UI.

* **Style**
  * Improved type annotations for dataset attributes, allowing them to be optional.

* **Chores**
  * Removed unused table suffix constants from configuration files to reduce redundancy.

PR #1035, opened 2025-06-25
feat(package): Add garbage collectors to delete archives and search results based on a configurable retention policy.

haiqi96:
# Description

Note: this PR is based on https://github.com/y-scope/clp/pull/1023

This PR includes the following change.
1. Added a new `retention_cleaner` component to the package that launches one or more tasks via asyncio to handle retention for different components.
2. Updated clp_config.py to support specifying retention period and retention frequencies for archives and search results.
3. Updated S3 api for deleting objects from a bucket and updated the require permissions in the user Docs.
4. Implemented a targetBuffer class to support fault tolerance.

This PR also handles a potential race condition between search scheduler and archive retention. It's possible that an archive to be removed by the retention cleaner is still being searched.
To avoid this, we require the following:
1. the search scheduler will not search on any archives that is considered out-of-date
2. the archive retention cleaner shall not start deleting archives until all search jobs that starts before the `expiry_ts` finishes, which ensures that no jobs will be searching on the archive.

Note:
This PR doesn't support stream retention yet. The stream retention will be properly supported in a future PR.

# Checklist

  breaking change.

# Validation performed

- Tested corner case and basic functionality of archive retention for both clp-json and clp-text on fs. The test script can be found at https://github.com/haiqi96/retention_tester
- Manually tested S3 basic functionality
- Manually tested search results garbage collection

## Summary by CodeRabbit

* **New Features**
  * Added a configurable garbage collector component with CLI start/stop integration and per-container config handling.

* **Improvements**
  * New retention-period and sweep-interval configuration; scheduler now filters by retention and a DB index improves job lookup performance.
  * S3 utilities support batched object deletion; IAM examples updated to include s3:DeleteObject.
  * Standardized directory validation and safer, coordinated metadata-plus-storage deletion with fault-tolerant buffering.


components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -1111,6 +1115,89 @@
...
+    # fmt: off
+    retention_cleaner_cmd = [
+        "python3", "-u",
+        "-m", "job_orchestration.retention.retention_cleaner",
+        "--config", str(container_clp_config.logs_directory / container_config_filename),
+    ]
+    cmd = container_start_cmd + retention_cleaner_cmd
+    subprocess.run(cmd, stdout=subprocess.DEVNULL, check=True)
+
+    logger.info(f"Started {component_name}.")
```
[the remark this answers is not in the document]
haiqi96: why prefer this to f string?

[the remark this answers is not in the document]
haiqi96: Interesting, thanks. For this time, I guess I will keep the line in consistent with the rest of files. Will consider this in the future PRs.

components/job-orchestration/job_orchestration/retention/search_results_handler.py
```diff
@@ -0,0 +1,73 @@
...
+async def search_results_retention(
+    clp_config: CLPConfig, log_directory: pathlib.Path, logging_level: str
+) -> None:
+    configure_logger(logger, logging_level, log_directory, SEARCH_RESULTS_RETENTION_HANDLER_NAME)
+
+    job_frequency_secs = clp_config.retention_cleaner.job_frequency.search_results * MIN_TO_SECONDS
+    while True:
+        _handle_search_results_retention(
+            clp_config.results_cache, clp_config.webui.results_metadata_collection_name
+        )
```
haiqi96: note to myself, good to have a catch re-throw here to output the error log.

haiqi96: Just for more context, currently if an async task fails, the execption message will only gets logged in mean retention_cleaner's log, but not logged in the log specific to the task.

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -308,3 +310,44 @@ def s3_put(s3_config: S3Config, src_file: Path, dest_path: str) -> None:
...
+
+def s3_delete_objects(s3_config: S3Config, object_keys: Set[str]) -> None:
+    """
+    Deletes a set of objects from the S3 bucket using AWS's DeleteObjects operation.
+    The objects to be deleted are identified by keys relative to the `s3_config.key_prefix`.
+    Note: `boto3.delete_objects` is limited to 1000 objects per call. This method divides the input
+    object keys into chunks of 1000 in size and calls `boto3.delete_objects` multiple times until
+    all objects are deleted.
+
+    :param s3_config: S3 configuration specifying the upload destination and credentials.
```
haiqi96:
```suggestion
    :param s3_config: S3 configuration specifying the credentials and the bucket to perform deletion.
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -508,8 +495,11 @@ def validate_results_cache_config(
     )
 
 
-def validate_worker_config(clp_config: CLPConfig):
+def validate_logs_input_config(clp_config: CLPConfig):
     clp_config.validate_logs_input_config()
+
+
+def validate_output_config(clp_config: CLPConfig):
```
LinZhihao-723: As discussed offline, how about `validate_output_storage_config`?

components/clp-py-utils/clp_py_utils/clp_metadata_db_utils.py
```diff
@@ -219,6 +219,47 @@ def create_metadata_db_tables(db_cursor, table_prefix: str, dataset: str | None
     _create_files_table(db_cursor, table_prefix, dataset)
 
 
+def delete_archives_from_metadata_db(
+    db_cursor, archive_ids: List[str], table_prefix: str, dataset: str | None
+):
+    """
+    Deletes the given list of archives from the metadata database. Also deletes associated entries
+    from `files` and `archive_tags` table that reference the target archives.
```
LinZhihao-723:
```suggestion
def delete_archives_from_metadata_db(
    db_cursor, archive_ids: List[str], table_prefix: str, dataset: str | None
) -> None:
    """
    Deletes archives from the metadata database specified by a list of IDs. It also deletes
    the associated entries from `files` and `archive_tags` tables that reference these archives.

```
* Added `-> None` to be consistent with other functions in this file.
* Polished the comment a little bit. Also, we need to leave an empty line between paragraphs to make the docstring appropriately rendered.

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -1051,6 +1055,89 @@ def start_reducer(
...
+    instance_id: str,
+    clp_config: CLPConfig,
+    container_clp_config: CLPConfig,
+    mounts: CLPDockerMounts,
+):
+    component_name = RETENTION_CLEANER_COMPONENT_NAME
+
+    retention_cleaner_required = is_retention_cleaner_required(clp_config)
+    if not retention_cleaner_required:
+        logger.info(f"No retention period is configured, skipping {component_name} creation...")
```
LinZhihao-723:
```suggestion
    if not is_retention_cleaner_required(clp_config):
        logger.info(f"Retention period is not configured, skipping {component_name} creation...")
```
How about this?

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -1141,7 +1230,14 @@ def main(argv):
             COMPRESSION_WORKER_COMPONENT_NAME,
             QUERY_WORKER_COMPONENT_NAME,
         ):
-            validate_worker_config(clp_config)
+            validate_logs_input_config(clp_config)
```
LinZhihao-723, line 1246: Just curious: I know that eventually we need the query worker to be able to read the input logs, but why the query worker needs to read the input logs now?

components/clp-py-utils/clp_py_utils/clp_config.py
LinZhihao-723, line 1: Have some suggestions in terms of naming. I think this file is strongly bound to our YAML config. I will leave comments in the config file then.

components/package-template/src/etc/clp-config.yml
```diff
@@ -99,6 +105,15 @@
...
 #  target_uncompressed_size: 134217728  # 128 MB
 #
+## Configuration for retention jobs
+#retention_cleaner:
+#  logging_level: "INFO"
+#
+#  # Frequency (in minutes) at which each retention job runs
+#  job_frequency:
+#    archives: 60
+#    search_results: 30
```
LinZhihao-723: Since this is directly user-oriented, let's polish the naming. The following suggestion has been confirmed by @kirkrodrigues 

```suggestion
## Garbage collector config
#garbage_collector:
#  logging_level: "INFO"
#
#  # Interval (in minutes) at which garbage collector jobs run
#  sweep_interval:
#    archive: 60
#    search_result: 30
```

* `retention_cleaner` -> `garbage_collector`
* `job_frequency` -> `sweep_interval`
* Drop the plarul forms for `archives` and `search_results`

components/package-template/src/etc/clp-config.yml
```diff
@@ -71,6 +74,9 @@
 #    # error will be raised if so.
 #    directory: "var/data/archives"
 #
+#  # Retention period for archives, in minutes. Set to null to disable retention.
```
LinZhihao-723:
```suggestion
#  # Retention period for archives, in minutes. Set to null to disable automatic deletion.
```
"disable retention" isn't a correct term: it sounds like we just don't retain the archive at all, lol

components/package-template/src/etc/clp-config.yml
```diff
@@ -52,6 +52,9 @@
 #  db_name: "clp-query-results"
 #  stream_collection_name: "stream-files"
 #
+#  # Retention period for search results, in minutes. Set to null to disable retention.
```
LinZhihao-723:
```suggestion
#  # Retention period for search results, in minutes. Set to null to disable automatic deletion.
```
Same.

LinZhihao-723 (changes requested):
Comments 2/x.
Let's fix these naming suggestions first, since they may propagate across the entire PR.

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
haiqi96: 80% You are right, it shouldn't need the input logs. but let me double check

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -308,3 +310,44 @@ def s3_put(s3_config: S3Config, src_file: Path, dest_path: str) -> None:
         s3_client.put_object(
             Bucket=s3_config.bucket, Body=file_data, Key=s3_config.key_prefix + dest_path
         )
+
+
+def s3_delete_objects(s3_config: S3Config, object_keys: Set[str]) -> None:
```
LinZhihao-723, line 315: How about `s3_delete_objects` -> `s3_delete_by_keys`?

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -308,3 +310,44 @@ def s3_put(s3_config: S3Config, src_file: Path, dest_path: str) -> None:
...
+    """
+    Deletes a set of objects from the S3 bucket using AWS's DeleteObjects operation.
+    The objects to be deleted are identified by keys relative to the `s3_config.key_prefix`.
+    Note: `boto3.delete_objects` is limited to 1000 objects per call. This method divides the input
+    object keys into chunks of 1000 in size and calls `boto3.delete_objects` multiple times until
+    all objects are deleted.
+
+    :param s3_config: S3 configuration specifying the credentials and the bucket to perform deletion.
```
LinZhihao-723:
```suggestion
    :param s3_config: The S3 config specifying the credentials and the bucket to perform deletion.
```
The current line already exceeds 100-char limit.

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -308,3 +310,44 @@ def s3_put(s3_config: S3Config, src_file: Path, dest_path: str) -> None:
...
         )
+
+
+def s3_delete_objects(s3_config: S3Config, object_keys: Set[str]) -> None:
+    """
+    Deletes a set of objects from the S3 bucket using AWS's DeleteObjects operation.
+    The objects to be deleted are identified by keys relative to the `s3_config.key_prefix`.
+    Note: `boto3.delete_objects` is limited to 1000 objects per call. This method divides the input
+    object keys into chunks of 1000 in size and calls `boto3.delete_objects` multiple times until
+    all objects are deleted.
```
LinZhihao-723:
```suggestion
    Deletes objects from an S3 bucket. The objects are identified by keys relative to
    `s3_config.key_prefix`.

    Note: The AWS S3 `DeleteObjects` API, used by `boto3.client.delete_objects`, supports a maximum
    of 1,000 objects per request. This method splits the provided keys into batches of up to 1,000
    and issues multiple delete requests until all objects are removed.
```

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -308,3 +310,44 @@ def s3_put(s3_config: S3Config, src_file: Path, dest_path: str) -> None:
...
+    s3_client = _create_s3_client(s3_config, boto3_config)
+
+    MAX_OBJECTS_PER_REQUEST = 1000
+    for chunked_keys in _iterate_by_chunk(
+        s3_config.key_prefix, object_keys, MAX_OBJECTS_PER_REQUEST
+    ):
+        s3_client.delete_objects(
+            Bucket=s3_config.bucket,
+            Delete=_gen_deletion_config(chunked_keys),
+        )
```
LinZhihao-723: Let's try to avoid using inline functions. So how about rewriting it in the following way:
```suggestion
    boto3_config = Config(retries=dict(total_max_attempts=3, mode="adaptive"))
    s3_client = _create_s3_client(s3_config, boto3_config)
    max_num_objects_to_delete_per_batch = 1000
    objects_to_delete: List[str] = []
    for relative_obj_key in object_keys:
        objects_to_delete.append(s3_config.key_prefix + relative_obj_key)
        if len(objects_to_delete) < max_num_objects_to_delete_per_batch:
            continue
        s3_client.delete_objects(
            Bucket=s3_config.bucket,
            Delete=_gen_deletion_config(objects_to_delete),
        )
        objects_to_delete = []

    if len(objects_to_delete) != 0:
        s3_client.delete_objects(
            Bucket=s3_config.bucket,
            Delete=_gen_deletion_config(objects_to_delete),
        )
```
A few things to notice:
- This assumes `_gen_deletion_config` is moved outside
- This removes `_iterate_by_chunk`:
  - I don't think this iterator is needed: with an iterator it's actually making the code even more complicated
  - I replaced `chunk` by `batch`, since `chunk` in S3 storage could mean storage units: an object may have multiple chunks. "Deleting by chunk" is misleading in this case.
- `max_num_objects_to_delete_per_batch` to avoid having a local constant
- Let's not use `if some_list` to check whether a list is not empty.

components/job-orchestration/job_orchestration/garbage_collector/garbage_collector.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Load configuration
+    config_path = Path(parsed_args.config)
+    try:
+        clp_config = CLPConfig.parse_obj(read_yaml_config_file(config_path))
+    except ValidationError as err:
+        logger.error(err)
+        return 1
+    except Exception as ex:
+        logger.error(ex)
```
LinZhihao-723:
```suggestion
    except Exception:
        logger.exception("Failed to parse CLP configuration file.")
```
We can log like this for exceptions, right?

components/job-orchestration/job_orchestration/garbage_collector/garbage_collector.py
```diff
@@ -0,0 +1,91 @@
...
+        return 1
+
+    retention_tasks: Dict[str, Tuple[Optional[int], Callable]] = {
```
LinZhihao-723: How about: `retention_tasks` -> `gc_task_configs`

components/job-orchestration/job_orchestration/garbage_collector/garbage_collector.py
```diff
@@ -0,0 +1,91 @@
...
+        ARCHIVE_GARBAGE_COLLECTOR_NAME: (
+            clp_config.archive_output.retention_period,
+            archive_garbage_collector,
+        ),
+        SEARCH_RESULT_GARBAGE_COLLECTOR_NAME: (
+            clp_config.results_cache.retention_period,
+            search_result_garbage_collector,
+        ),
+    }
+    tasks_handler: List[asyncio.Task[None]] = []
```
LinZhihao-723: How about: `tasks_handler` -> `gc_tasks`

components/job-orchestration/job_orchestration/garbage_collector/garbage_collector.py
```diff
@@ -0,0 +1,91 @@
...
+    # Create retention tasks
+    for task_name, (retention_period, task_method) in retention_tasks.items():
+        if retention_period is not None:
+            logger.info(f"Creating {task_name} task with retention = {retention_period} minutes")
+            garbage_collection_task = asyncio.create_task(
+                task_method(clp_config, logs_directory, logging_level), name=task_name
+            )
+            tasks_handler.append(garbage_collection_task)
+        else:
+            logger.info(f"No retention period configured, skip creating {task_name} task.")
```
LinZhihao-723:
```suggestion
        if retention_period is None:
            logger.info(f"No retention period configured, skip creating {task_name} task.")
            continue
        logger.info(f"Creating {task_name} task with retention period = {retention_period} minutes")
        garbage_collection_task = asyncio.create_task(
            task_method(clp_config, logs_directory, logging_level), name=task_name
        )
        tasks_handler.append(garbage_collection_task)
```
- Reduce the indentation level
- "with retention" -> "with retention period"

components/job-orchestration/job_orchestration/garbage_collector/garbage_collector.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Poll and report any task that finishes unexpectedly
+    while tasks_handler:
```
LinZhihao-723:
```suggestion
    while len(tasks_handler) != 0:
```

components/job-orchestration/job_orchestration/garbage_collector/garbage_collector.py
```diff
@@ -0,0 +1,91 @@
...
+        done, _ = await asyncio.wait(tasks_handler, return_when=asyncio.FIRST_COMPLETED)
+        for task in done:
+            tasks_handler.remove(task)
+            task_name = task.get_name()
+            try:
+                _ = task.result()
+                logger.info(f"Task {task_name} unexpectedly terminated without an error.")
+            except Exception as e:
+                logger.exception(f"Task {task_name} failed with exception: {e}")
```
LinZhihao-723:
```suggestion
                logger.exception(f"Task {task_name} failed.")
```
The exception should be automatically logged by `exception()` right?

components/job-orchestration/job_orchestration/garbage_collector/garbage_collector.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Poll and report any task that finishes unexpectedly
+    while tasks_handler:
+        done, _ = await asyncio.wait(tasks_handler, return_when=asyncio.FIRST_COMPLETED)
+        for task in done:
+            tasks_handler.remove(task)
+            task_name = task.get_name()
+            try:
+                _ = task.result()
+                logger.info(f"Task {task_name} unexpectedly terminated without an error.")
```
LinZhihao-723: I assume these gc tasks should be running forever. If it ends, then there should be a clear error. Shall we log it as a warning or an error?

components/job-orchestration/job_orchestration/garbage_collector/garbage_collector.py
```diff
@@ -0,0 +1,91 @@
...
+            except Exception as e:
+                logger.exception(f"Task {task_name} failed with exception: {e}")
+
+    logger.error("All garbage collection tasks unexpectedly terminated.")
```
LinZhihao-723:
```suggestion
    logger.error("All garbage collection tasks terminated unexpectedly.")
```
Both "unexpectedly terminated" and "terminated unexpectedly" are grammatically correct, but the later seems more common.

components/job-orchestration/job_orchestration/garbage_collector/garbage_collector.py
```diff
@@ -0,0 +1,91 @@
...
+    return -1
```
LinZhihao-723: coderabbit may raise a good point. Shall we return 1 instead? Like, do we have a convention on the error return code?

components/job-orchestration/job_orchestration/garbage_collector/garbage_collector.py
```diff
@@ -0,0 +1,91 @@
...
+            archive_garbage_collector,
+        ),
+        SEARCH_RESULT_GARBAGE_COLLECTOR_NAME: (
+            clp_config.results_cache.retention_period,
+            search_result_garbage_collector,
+        ),
+    }
+    tasks_handler: List[asyncio.Task[None]] = []
+
+    # Create retention tasks
```
LinZhihao-723:
```suggestion
    # Create GC tasks
```

components/job-orchestration/job_orchestration/garbage_collector/garbage_collector.py
```diff
@@ -0,0 +1,91 @@
...
+        if retention_period is not None:
+            logger.info(f"Creating {task_name} task with retention = {retention_period} minutes")
+            garbage_collection_task = asyncio.create_task(
+                task_method(clp_config, logs_directory, logging_level), name=task_name
+            )
+            tasks_handler.append(garbage_collection_task)
+        else:
+            logger.info(f"No retention period configured, skip creating {task_name} task.")
+
+    # Poll and report any task that finishes unexpectedly
```
LinZhihao-723:
```suggestion
    # Poll and report any task that finished unexpectedly
```

LinZhihao-723 (changes requested):
Comments 4/x
Reviewed the main entry of the garbage collector.

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,145 @@
...
+    :param target: Relative path of the file or directory to remove.
+    """
+    path_to_remove = fs_storage_config.directory / target
+    if not path_to_remove.exists():
+        return
+
+    if path_to_remove.is_dir():
+        shutil.rmtree(path_to_remove)
+    else:
+        os.remove(path_to_remove)
```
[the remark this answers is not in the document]
haiqi96: imo, if those operations fails, it's very likely due to a permission issue.  in that case I think it's better to just let the entire job exit with exception since it will fail again anyway

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,145 @@
...
+        Writes any new targets added since initialization to the recovery file.
+        """
+        if len(self._targets_to_persist) == 0:
+            return
+
+        with open(self._recovery_file_path, "a") as f:
+            for target in self._targets_to_persist:
+                f.write(f"{target}\n")
+
+        self._targets_to_persist.clear()
```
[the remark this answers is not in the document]
haiqi96: same reasoning as above.

components/job-orchestration/job_orchestration/garbage_collector/garbage_collector.py
haiqi96: Sure, also renamed some other variables

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,145 @@
...
+        for target in targets:
+            remove_fs_target(storage_config, target)
+    else:
+        raise ValueError(f"Unsupported Storage type: {storage_type}")
+
+
+class TargetsBuffer:
+    """
+    A class representing an in-memory buffer for targets with fault-tolerance support.
+    This class support recovering from a previous failure by reading previously persisted targets
```
LinZhihao-723:
```suggestion
    A class representing an in-memory buffer for targets with fault-tolerance support.

    This class supports recovering from a previous failure by reading previously persisted targets
```

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,145 @@
...
+        if len(self._targets_to_persist) == 0:
+            return
+
+        with open(self._recovery_file_path, "a") as f:
+            for target in self._targets_to_persist:
+                f.write(f"{target}\n")
+
+        self._targets_to_persist.clear()
+
+    def flush(self):
```
LinZhihao-723: Do u think `clear` would be a better name?

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+        Clear the in-memory buffer of targets and remove the recovery file.
+        This is intended to be called after the caller finished processing all targets (i.e. when
+        recovery is no longer needed for the targets.)
```
LinZhihao-723:
```suggestion
        Clears the in-memory buffer of targets and removes the recovery file.

        This is intended to be called after the caller finished processing all targets (i.e., when
        recovery is no longer needed for the targets).
```

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,145 @@
...
+    The `expiry_epoch` is calculated as `expiry_epoch = cur_time - retention_secs`.
+
+    :param: retention_minutes: Retention period in minutes.
+    :return: The UTC epoch of the expiry time.
+    """
+    return int(time.time() - retention_minutes * MIN_TO_SECONDS)
+
+
+def get_oid_with_expiry_time(expiry_epoch_secs: int) -> ObjectId:
+    return ObjectId.from_datetime(datetime.utcfromtimestamp(expiry_epoch_secs))
```
LinZhihao-723:
```suggestion
    return ObjectId.from_datetime(datetime.fromtimestamp(expiry_epoch_secs, tz=UTC))
```
`utcfromtimestamp` is deprecated. We can use `fromtimestamp` and import `UTC` from `datetime`.

LinZhihao-723 (changes requested):
Comments 5/x.
Reviewed utils.

components/job-orchestration/job_orchestration/garbage_collector/archive_garbage_collector.py
```diff
@@ -0,0 +1,217 @@
...
+    guaranteed not to be searched by any running query jobs.
+
+    If no query jobs are running, the expiry time is set to `current_time - retention_period`.
+    If a query job is running and was created at `creation_time`, the query scheduler guarantees
+    that it will not search any archive whose end_ts < (creation_time - retention_period).
+    In this case, the expiry time can be safely adjusted to `creation_time - retention_period`.
+
+    Note: This function does not consider query jobs that started before
+    `current_time - retention_period`, as such long-running jobs are likely hanging.
+    Including them would prevent the expiry time from advancing.
```
LinZhihao-723:
```suggestion
    `current_time - retention_period`, as such long-running jobs are likely hanging. Including them
    would prevent the expiry time from advancing.
```

components/job-orchestration/job_orchestration/garbage_collector/archive_garbage_collector.py
```diff
@@ -0,0 +1,217 @@
...
+        logger.debug(f"Discovered running query job created at {job_creation_time}.")
+        logger.debug(f"Using adjusted archive_expiry_epoch=`{archive_expiry_epoch}`.")
+    else:
+        archive_expiry_epoch = int(current_epoch_secs) - retention_period_secs
+        logger.debug(f"Using archive_expiry_epoch=`{archive_expiry_epoch}`.")
+
+    return archive_expiry_epoch
+
+
+def _handle_archive_garbage_collection(
```
LinZhihao-723: How about:
`_handle_archive_garbage_collection` -> `_collect_and_sweep_expired_archives`

- "collect" indicates we get a expiry epoch as a cutoff for removal
- "sweep" means we will remove the archives

components/job-orchestration/job_orchestration/garbage_collector/archive_garbage_collector.py
```diff
@@ -0,0 +1,217 @@
...
+
+    delete_targets(archive_output_config, targets_to_delete)
+
+    # Prepare the log message
+    dataset_msg: str
+    deleted_targets: List[str]
+    if dataset is not None:
+        dataset_log_msg = f" from dataset `{dataset}`"
+        # Note: If dataset is not None, targets are expected to be in the format `<dataset>/<archive_id>`
+        deleted_targets = [target.split("/")[1] for target in targets_to_delete]
```
LinZhihao-723:
```suggestion
        # Note: If dataset is not None, targets are expected to be in the format of
        # `<dataset>/<archive_id>`
```
This comment exceeds 100-char limit. Not sure why the linter isn't alerting.

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
```diff
@@ -0,0 +1,86 @@
...
+                collection_timestamp = _get_latest_doc_timestamp(job_results_collection)
+
+                if collection_timestamp < expiry_epoch:
+                    logger.debug(f"Removing search results of job: {job_id}.")
+                    _remove_result_metadata(
+                        results_cache_db, results_metadata_collection_name, job_id
+                    )
+                    job_results_collection.drop()
+    except Exception:
+        logger.exception("Failed to delete search results from the results cache.")
```
LinZhihao-723: Did u miss raising the exception after the logging statement?

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
```diff
@@ -0,0 +1,86 @@
...
+
+                job_results_collection = results_cache_db.get_collection(job_id)
+                collection_timestamp = _get_latest_doc_timestamp(job_results_collection)
+
+                if collection_timestamp < expiry_epoch:
+                    logger.debug(f"Removing search results of job: {job_id}.")
+                    _remove_result_metadata(
+                        results_cache_db, results_metadata_collection_name, job_id
+                    )
+                    job_results_collection.drop()
```
LinZhihao-723:
```suggestion
                if collection_timestamp >= expiry_epoch:
                    continue

                logger.debug(f"Removing search results of job: {job_id}.")
                _remove_result_metadata(results_cache_db, results_metadata_collection_name, job_id)
                job_results_collection.drop()
```
I'd highly prefer we avoid writing long blocks inside if in a loop, especially in Python. The context manager and try block already introduce extra levels of indentation, lol.

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def _remove_result_metadata(
+    database: pymongo.database.Database, results_metadata_collection_name: str, job_id: str
+) -> None:
+    results_metadata_collection = database.get_collection(results_metadata_collection_name)
+    results_metadata_collection.delete_one({MONGODB_ID_KEY: job_id})
+
+
+def _handle_search_result_garbage_collection(
```
LinZhihao-723: Same, how about `_collect_and_sweep_expired_search_results`?

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
```diff
@@ -0,0 +1,86 @@
...
+logger = get_logger(SEARCH_RESULT_GARBAGE_COLLECTOR_NAME)
+
+
+def _get_latest_doc_timestamp(collection: pymongo.collection.Collection) -> int:
+    latest_doc = collection.find_one(sort=[(MONGODB_ID_KEY, pymongo.DESCENDING)])
+    if latest_doc:
+        object_id = latest_doc[MONGODB_ID_KEY]
+        if isinstance(object_id, ObjectId):
+            return int(object_id.generation_time.timestamp())
+        raise ValueError(f"{object_id} is not a ObjectID")
```
LinZhihao-723:
```suggestion
        raise ValueError(f"{object_id} is not an ObjectID")
```

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
```diff
@@ -0,0 +1,86 @@
...
+
+    return 0
```
LinZhihao-723:
```suggestion
    if latest_doc is None:
        return 0

    object_id = latest_doc[MONGODB_ID_KEY]
    if isinstance(object_id, ObjectId):
        return int(object_id.generation_time.timestamp())
    raise ValueError(f"{object_id} is not a ObjectID")
```

LinZhihao-723 (changes requested):
Comments 6/x.
Reviewed the actual GC implementations.

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
haiqi96: hmm, I was feeling that maybe the collector and retry later instead of failing, but maybe you are right, it's better to let it just fail loudly.

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,145 @@
...
+    if StorageType.S3 == storage_type:
+        s3_delete_objects(storage_config.s3_config, targets)
+    elif StorageType.FS == storage_type:
+        for target in targets:
+            delete_fs_target(storage_config, target)
+    else:
+        raise ValueError(f"Unsupported Storage type: {storage_type}")
+
+
+class TargetsBuffer:
```
LinZhihao-723: How about `DeletionCandidatesBuffer`?

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,145 @@
...
+            return
+
+        if not self._recovery_file_path.is_file():
+            raise ValueError(f"{self._recovery_file_path} does not resolve to a file")
+
+        with open(self._recovery_file_path, "r") as f:
+            for line in f:
+                self._targets.add(line.strip())
+
+    def add_target(self, target: str) -> None:
```
LinZhihao-723: How about `add_candidate`?

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,145 @@
...
+        if target not in self._targets:
+            self._targets.add(target)
+            self._targets_to_persist.append(target)
+
+    def get_targets(self) -> Set[str]:
```
LinZhihao-723: How about `get_candidates`?

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,145 @@
...
+        return self._targets
+
+    def persists_new_targets(self) -> None:
```
LinZhihao-723: How about `persist_new_candidates`?
Notice that it should be `persist`, not `persists`

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,145 @@
...
+    if not path_to_delete.exists():
+        return
+
+    if path_to_delete.is_dir():
+        shutil.rmtree(path_to_delete)
+    else:
+        os.remove(path_to_delete)
+
+
+def delete_targets(output_config: ArchiveOutput, targets: Set[str]) -> None:
```
LinZhihao-723: How about:
- `delete_targets` -> `execute_deletion`
- `targets` -> `deletion_candidates`

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,145 @@
...
+    :return: The UTC epoch of the expiry time.
+    """
+    return int(time.time() - retention_minutes * MIN_TO_SECONDS)
+
+
+def get_oid_with_expiry_time(expiry_epoch_secs: int) -> ObjectId:
+    return ObjectId.from_datetime(datetime.utcfromtimestamp(expiry_epoch_secs))
+
+
+def delete_fs_target(fs_storage_config: FsStorage, target: str) -> None:
```
LinZhihao-723: How about:
- `delete_fs_targets` -> `execute_fs_deletion`
- `targets` -> `deletion_candidates`

components/job-orchestration/job_orchestration/garbage_collector/archive_garbage_collector.py
```diff
@@ -0,0 +1,217 @@
...
+            targets_buffer.persists_new_targets()
+            db_conn.commit()
+    except Exception:
+        logger.exception("Failed to delete archives from the database. Aborting deletion.")
+        if db_conn in locals() and db_conn.is_connected():
+            db_conn.rollback()
+
+    targets_to_delete = targets_buffer.get_targets()
+    num_targets_to_delete = len(targets_to_delete)
+    if num_targets_to_delete == 0:
```
LinZhihao-723: Minor:
```suggestion
    if 0 == num_targets_to_delete:
```

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -670,7 +677,15 @@ def handle_pending_query_jobs(
                     continue
 
                 search_config = SearchJobConfig.parse_obj(job_config)
-                archives_for_search = get_archives_for_search(db_conn, table_prefix, search_config)
+                cutoff_epoch_msecs: Optional[int] = None
```
LinZhihao-723: `cutoff_epoch_msecs` is a bit vague here since it doesn't indicate in which direction it cuts off: either > or <.
How about `archive_end_ts_lower_bound`? In this way, we know:
- `archive_end_ts` indicates the unit will be the same as our storage
- `lower_bound` indicates the direction of the cutoff

components/clp-py-utils/clp_py_utils/s3_utils.py
haiqi96: I feel delete objects is better since it also aligns with the api name used in this method.

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,147 @@
...
+    The `expiry_epoch` is calculated as `expiry_epoch = cur_time - retention_secs`.
+
+    :param: retention_minutes: Retention period in minutes.
+    :return: The UTC epoch of the expiry time.
+    """
+    return int(time.time() - retention_minutes * MIN_TO_SECONDS)
+
+
+def get_oid_with_expiry_time(expiry_epoch_secs: int) -> ObjectId:
+    return ObjectId.from_datetime(datetime.fromtimestamp(expiry_epoch_secs, tz=timezone.utc))
```
haiqi96, line 55: @LinZhihao-723 UTC is only available after python 3.11, so use timezone.utc for now

LinZhihao-723: oh ok. Didn't notice that since my local Python version is indeed 3.11. sgtm.

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
haiqi96: Changed the code that exception will only be caught by top-level loop runner and will cause the colletor to fail

components/clp-package-utils/clp_package_utils/scripts/native/archive_manager.py
```diff
@@ -325,13 +324,13 @@ def _delete_archives(
...
-        sql_adapter: SQL_Adapter = SQL_Adapter(database_config)
-        clp_db_connection_params: dict[str, any] = (
-            database_config.get_clp_connection_params_and_type(True)
-        )
-        table_prefix = clp_db_connection_params["table_prefix"]
+    sql_adapter: SQL_Adapter = SQL_Adapter(database_config)
+    clp_db_connection_params: dict[str, any] = database_config.get_clp_connection_params_and_type(
+        True
+    )
+    table_prefix = clp_db_connection_params["table_prefix"]
```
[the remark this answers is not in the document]
haiqi96: this should be fine.  we don't expect those lines to raise any error in normal execution.

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -508,8 +495,11 @@ def validate_results_cache_config(
     )
 
 
-def validate_worker_config(clp_config: CLPConfig):
+def validate_logs_input_config(clp_config: CLPConfig):
```
LinZhihao-723:
```suggestion
def validate_logs_input_config(clp_config: CLPConfig) -> None:
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -508,8 +495,11 @@ def validate_results_cache_config(
...
     clp_config.validate_logs_input_config()
+
+
+def validate_output_storage_config(clp_config: CLPConfig):
```
LinZhihao-723:
```suggestion
def validate_output_storage_config(clp_config: CLPConfig) -> None:
```

components/job-orchestration/job_orchestration/garbage_collector/archive_garbage_collector.py
```diff
@@ -0,0 +1,216 @@
...
+
+    archive_output_config = clp_config.archive_output
+    storage_engine = clp_config.package.storage_engine
+    validate_storage_type(archive_output_config, storage_engine)
+
+    sweep_interval_secs = clp_config.garbage_collector.sweep_interval.archive * MIN_TO_SECONDS
+    recovery_file = clp_config.logs_directory / f"{ARCHIVE_GARBAGE_COLLECTOR_NAME}.tmp"
+
+    logger.info(f"{ARCHIVE_GARBAGE_COLLECTOR_NAME} started.")
+    # Start retention loop
```
LinZhihao-723:
```suggestion
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -590,3 +580,13 @@ def validate_dataset_name(clp_table_prefix: str, dataset_name: str) -> None:
             f"Invalid dataset name: `{dataset_name}`. Names can only be a maximum of"
             f" {dataset_name_max_len} characters long."
         )
+
+
+def is_retention_configured(clp_config: CLPConfig) -> bool:
```
LinZhihao-723: `is_retention_configured` -> `is_retention_period_configured`

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,147 @@
...
+                f.write(f"{target}\n")
+
+        self._candidates_to_persist.clear()
+
+    def clear(self):
+        """
+        Clear the in-memory buffer of candidates and remove the recovery file.
+
+        This is intended to be called after the caller finished processing all candidates (i.e. when
+        recovery is no longer needed for the candidates.)
```
LinZhihao-723:
```suggestion
        Clears the in-memory buffer of candidates and removes the recovery file.

        This is intended to be called after the caller finished processing all candidates (i.e.,
        when recovery is no longer needed for the candidates.)
```
U probably missed this comment: https://github.com/y-scope/clp/pull/1035#discussion_r2252161744

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,147 @@
...
+        raise ValueError(f"Unsupported Storage type: {storage_type}")
+
+
+class DeletionCandidatesBuffer:
+    """
+    A class representing an in-memory buffer for candidates with fault-tolerance support.
+
+    This class support recovering from a previous failure by reading previously persisted candidates
+    from a recovery file. The user is expected to explicitly call `persists_new_candidates` to
+    persist any new candidates on to the disk for fault-tolerance purposes.
```
LinZhihao-723:
```suggestion
    Represents an in-memory buffer for deletion candidates with fault-tolerance support.

    This class supports recovering from a previous failure by reading previously persisted
    candidates from a recovery file. The user is expected to explicitly call
    `persist_new_candidates` to persist any new candidates on to the disk for fault-tolerance
    purposes.
```

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,147 @@
...
+        if target not in self._candidates:
+            self._candidates.add(target)
+            self._candidates_to_persist.append(target)
+
+    def get_candidates(self) -> Set[str]:
+        return self._candidates
+
+    def persist_new_candidates(self) -> None:
+        """
+        Writes any new candidates added since initialization to the recovery file.
```
LinZhihao-723: This isn't true, right? It should be since the last call of `persist_new_candidates`, not "since initialization". Can u update this description?

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
```diff
@@ -0,0 +1,85 @@
...
+            job_results_collection = results_cache_db.get_collection(job_id)
+            collection_timestamp = _get_latest_doc_timestamp(job_results_collection)
+            if collection_timestamp >= expiry_epoch:
+                continue
+
+            _delete_result_metadata(results_cache_db, results_metadata_collection_name, job_id)
+            job_results_collection.drop()
+            deleted_job_ids.append(int(job_id))
+
+    logger.debug(f"deleted search results of job(s): {deleted_job_ids}.")
```
LinZhihao-723:
```suggestion
    logger.debug(f"Deleted search results of job(s): {deleted_job_ids}.")
```

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -0,0 +1,147 @@
...
+            return
+
+        if not self._recovery_file_path.is_file():
+            raise ValueError(f"{self._recovery_file_path} does not resolve to a file")
+
+        with open(self._recovery_file_path, "r") as f:
+            for line in f:
+                self._candidates.add(line.strip())
+
+    def add_candidate(self, target: str) -> None:
```
LinZhihao-723: `target` -> `candidate`?

LinZhihao-723 (changes requested):
A few missing comments.

components/job-orchestration/job_orchestration/garbage_collector/utils.py
haiqi96: True, should have pulled your suggestion before I made any local change.

components/job-orchestration/job_orchestration/garbage_collector/utils.py
```diff
@@ -114,33 +115,35 @@ def __init__(self, recovery_file_path: pathlib.Path):
...
 
     def get_candidates(self) -> Set[str]:
         return self._candidates
 
     def persist_new_candidates(self) -> None:
         """
-        Writes any new candidates added since initialization to the recovery file.
+        Writes any new candidates from `_candidates_to_persist` to the recovery file and clears
+        the buffer.
+        The method returns immediately without writing if `_candidates_to_persist` is empty.
```
LinZhihao-723:
```suggestion
        Persists any new deletion candidates buffered in `_candidates_to_persist` by appending them
        to the recovery file, then clears `_candidates_to_persist`.
        This method returns immediately if `_candidates_to_persist` is empty.
```

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
```diff
@@ -0,0 +1,88 @@
...
+
+            _delete_result_metadata(results_cache_db, results_metadata_collection_name, job_id)
+            job_results_collection.drop()
+            deleted_job_ids.append(int(job_id))
+
+    if len(deleted_job_ids) != 0:
+        logger.debug(f"Deleted search results of job(s): {deleted_job_ids}.")
+    else:
+        logger.debug(f"No search results matched the expiry criteria.")
+
```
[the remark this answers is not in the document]
LinZhihao-723: We forward the errors to the top-level catch, as DB exceptions are considered non-recoverable in our use-case. We will exit on such errors.

LinZhihao-723 (approved):
For the PR title, how about:
```
feat(package): Add garbage collectors to delete archives and search results based on a configurable retention policy.
```

components/clp-package-utils/clp_package_utils/scripts/native/archive_manager.py
```diff
@@ -358,21 +356,7 @@ def _delete_archives(
...
-
-            db_cursor.execute(
-                f"""
-                DELETE FROM `{get_archive_tags_table_name(table_prefix, dataset)}`
-                WHERE archive_id in ({ids_list_string})
-                """
-            )
+            delete_archives_from_metadata_db(db_cursor, archive_ids, table_prefix, dataset)
             for archive_id in archive_ids:
                 logger.info(f"Deleted archive {archive_id} from the database.")
```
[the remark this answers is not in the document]
haiqi96: This is unrelated to this PR. will skip.

PR #1004, opened 2025-06-13
feat(clp-package): Expose dataset selection in package CLI scripts and web UI. (Deprecated)

Bill-hbrhbr:
# Description

# This PR is deprecated due to merge conflicts that would bloat the PR size after resolution.

# clp-package:

## clp-package-utils

Add dataset arguments `--dataset` to various package scripts‘ command-line interface. If no dataset argument is provided, the default dataset name is `default`.

## webui

Add dataset selector to new webui. Dataset selector has to query backend to get datasets, and then display them. I also now pass the dataset to the log viewer connector. 

The `cachedDataset` is the dataset when query is submitted. We have to do this since the user could change the selector, then click a link, which would result in an error. Instead we cache the dataset when they submitted the query. 
Screenshot 2025-06-23 at 12 23 39 PM

# Checklist

* [X] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [X] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [X] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* [X] Verified correctness in the following use cases:
  * [X] CLI
  * [x] UI

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

**New Features**
  * Added support for specifying a dataset name via a new --dataset option in compression, decompression, archive management, and search scripts.
  * Introduced a dataset selector UI to the search page, allowing users to choose and view available datasets.

**Bug Fixes**
  * Improved validation for dataset existence in the metadata database.

**Refactor**
  * Centralized dataset validation logic and updated internal usage to leverage new validation functions.

**Style**
  * Added new styling for dataset label and selector components in the web UI.

**Chores**
  * Removed unused SQL configuration constants and query functions from the ingest page code.

PR #1036, opened 2025-06-26
feat(clp-package): Add utility function to validate dataset's existence in CLI scripts.

Bill-hbrhbr:
# Description

Add utility functions to allow CLP package scripts to check whether the dataset name exists in the metadata, both upon script startup and during execution.
* Scripts can now fail fast with a clear error if users reference a non-existent dataset, or if users specify a dataset when using non `CLP_S` storage engine.
* During execution, supports dataset checking with a local cache, which gets refreshed with each query to the remote metadata database.

The facilitates PR #1004 if we want output error messages to users when they search for a dataset that currently does not exist in the metadata database.

# Checklist

  breaking change.

# Validation performed

* [X] Tested with CLP package.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced dataset validation to prevent processing of jobs referencing non-existent datasets.

* **Refactor**
  * Streamlined dataset validation by replacing helper functions with inline checks, improving efficiency and clarity.


components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -574,3 +577,23 @@ def validate_path_for_container_mount(path: pathlib.Path) -> None:
...
+
+
+def validate_dataset(clp_config: CLPConfig, dataset: typing.Optional[str]) -> None:
+    """
+    Checks if the provided dataset currently exists in the metadata database.
+    :param clp_config:
+    :param dataset:
+    """
+    if not dataset:
+        raise ValueError(f"`{dataset}` is not a valid dataset name.")
```
haiqi96: It's weird that we let the dataset to be optional but immediately raise if it's None.

I kind of see that you are preparing for PR1023, but I feel that perhaps we can write code in the way such that caller with a dataset=None will never call this method.

That might turn into some code like the following, but it seems mypy is ok with it.

```
def validate_dataset(dataset: str):
   # do something

def func_caller():
    # do something
    input_arg: Optional[str]
    if input_arg is not None:
        validate_dataset(input_arg)
```

components/clp-py-utils/clp_py_utils/clp_metadata_db_utils.py
```diff
@@ -157,6 +157,30 @@ def fetch_existing_datasets(
     return {row["name"] for row in rows}
 
 
+def validate_and_cache_dataset(
```
haiqi96: same comment, we might be able to avoid calling validate_and_cache_dataset when dataset is None.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -656,7 +644,10 @@ def handle_pending_query_jobs(
             table_prefix = clp_metadata_db_conn_params["table_prefix"]
             if StorageEngine.CLP_S == clp_storage_engine:
                 dataset = QueryJobConfig.parse_obj(job_config).dataset
-                if not _validate_dataset(db_cursor, table_prefix, dataset, existing_datasets):
+                dataset_exists, existing_datasets = validate_and_cache_dataset(
+                    db_cursor, table_prefix, dataset, existing_datasets
+                )
+                if not dataset_exists:
```
haiqi96: what if we simplify the logic to be:
```
if dataset not in existing_datasets and not validate_and_dataset(dataset):
   report error and continue

# iirc adding an existing item in the set should have no impact?
existing_datasets.add(dataset)
```

or
```
if existing_datasets is None:
    existing_datasets = fetch_existing_datasets(db_cursor, table_prefix)
if dataset not in existing_datasets
   if not validate_and_dataset(dataset):
       report error and continue
   existing_datasets.add(dataset)
```
since this logic is only used once, I feel we might not need a seperate function that returns a tuple.

Bill-hbrhbr: I don't know if in the future we support different retention periods for datasets, resulting in some old dataset name expiring. For now, of course if datasets never expire, the local cache can only grow

haiqi96: I don't think I get your concern, what would be an example of problem is dataset expires?
but I see the issue in my suggestion that I didn't consider the case where existing_datasets could be null.

That said, I still feel there is a way to simplify the code. Edited my second suggestion in the previous reply.

- Another question: if there a case where the existing_datasets will be set to None? It seems to be inited as set().

Bill-hbrhbr: you are right that there's no need to consider `None` for `existing_datasets`. It's just that sometimes this function would be called without supplying this local cache, but I guess we can make the default value just an empty set.
Regarding the _dataset expiring_, I guess the more appropriate way to phrase my concern is that in the future we may allow user the action of deleting datasets. In this case, the local cache is not strictly growing, so we cannot simply use `existing_datasets.add(dataset)`

Bill-hbrhbr: On a second thought, it's not a good practice to set the `existing_datasets` default to an empty set, see python [mutable default trap](https://stackoverflow.com/questions/41686829/why-does-pycharm-warn-about-mutable-default-arguments-how-can-i-work-around-the).

Also for your second suggestion:
```
if existing_datasets is None:
    existing_datasets = fetch_existing_datasets(db_cursor, table_prefix)
if dataset not in existing_datasets
   if not validate_and_dataset(dataset):
       report error and continue
   existing_datasets.add(dataset)
```
if the function caller supplies no local cache but wants one after the function executes, the function needs to explicitly return the cache. `existing_datasets = fetch_existing_datasets(db_cursor, table_prefix)` replaces the function arg, which is not an in-place update.

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -574,3 +577,21 @@ def validate_path_for_container_mount(path: pathlib.Path) -> None:
...
+    db_config: Database = clp_config.database
+    sql_adapter: SQL_Adapter = SQL_Adapter(db_config)
+    clp_db_connection_params: dict[str, any] = db_config.get_clp_connection_params_and_type(True)
+    table_prefix: str = clp_db_connection_params["table_prefix"]
+    with closing(sql_adapter.create_connection(True)) as db_conn, closing(
+        db_conn.cursor(dictionary=True)
+    ) as db_cursor:
+        dataset_exists, _ = validate_and_cache_dataset(db_cursor, table_prefix, dataset)
+        if not dataset_exists:
+            raise ValueError(f"Dataset `{dataset}` does not exist.")
```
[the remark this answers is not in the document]
Bill-hbrhbr: this script is meant to be called once upon function startup.

components/clp-py-utils/clp_py_utils/clp_metadata_db_utils.py
```diff
@@ -185,6 +185,32 @@ def fetch_existing_datasets(
...
+    """
+    if existing_datasets is not None and dataset in existing_datasets:
+        return True
+
+    # NOTE: This assumes we never delete a dataset.
+    new_datasets = fetch_existing_datasets(db_cursor, table_prefix)
+    if existing_datasets is not None:
+        existing_datasets.update(new_datasets)
+    return dataset in existing_datasets
+
```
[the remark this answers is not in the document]
Bill-hbrhbr: does the new commit fix the problem?

components/clp-py-utils/clp_py_utils/clp_metadata_db_utils.py
```diff
@@ -185,6 +185,32 @@ def fetch_existing_datasets(
...
+    :return: Whether the dataset exists.
+    """
+    if existing_datasets is not None and dataset in existing_datasets:
+        return True
+
+    # NOTE: This assumes we never delete a dataset.
+    new_datasets = fetch_existing_datasets(db_cursor, table_prefix)
+    if existing_datasets is not None:
+        existing_datasets.update(new_datasets)
+    return dataset in new_datasets
```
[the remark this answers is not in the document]
Bill-hbrhbr: we want to do an in-place update

components/clp-py-utils/clp_py_utils/clp_metadata_db_utils.py
```diff
@@ -185,6 +185,32 @@ def fetch_existing_datasets(
...
+
```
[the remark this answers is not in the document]
Bill-hbrhbr: dataset validation is performed once in the wrapper scripts, but repeated DB queries are necessary for constantly running schedulers on both the compression and the query side.
Refresh is only needed when things are not found in the local cache. Dataset collection updates should not be too frequent

Bill-hbrhbr:
@coderabbitai help me check all the unused or missing imports

Bill-hbrhbr:
@coderabbitai
from __future__ import annotations
is used for python pipe notations before py 3.10

components/clp-py-utils/clp_py_utils/clp_metadata_db_utils.py
```diff
@@ -185,6 +185,32 @@ def fetch_existing_datasets(
     return {row["name"] for row in rows}
 
 
+def validate_and_cache_dataset(
+    db_cursor,
+    table_prefix: str,
+    dataset: str,
+    existing_datasets: Set[str] | None = None,
```
haiqi96: what if we remove `existing_datasets` from the parameter cache let query_scheduler handle the caching instead?

in my opinion, adding this `existing_datasets` parameter complicates the function, and also make it hard to write the docstrings (at least I can't come up with something good).

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -556,3 +562,31 @@ def validate_path_for_container_mount(path: pathlib.Path) -> None:
...
+    :param clp_config:
+    :param input_dataset: Dataset from the CLI.
+    :return: The validated dataset to use.
+    :raise: ValueError
+    """
+    storage_engine: StorageEngine = clp_config.package.storage_engine
+    if StorageEngine.CLP_S != storage_engine:
+        if input_dataset is not None:
+            raise ValueError("Dataset selection is only enabled for CLP_S storage engine.")
+        return None
```
haiqi96: Sorry, I think you answered my question once but it got burried by a lot of other reviews.

Is there a case where you expect we call `validate_dataset` when storage_engine is not CLP_S?

Looking at the code, if input_dataset is None and the storage_engine is CLP (clp-text), the dataset will be defaulted to CLP_DEFAULT_DATASET_NAME. but as far as I know, clp-text should use an empty 'input_dataset' string, so the logic looks incorrect to me.

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -556,3 +562,31 @@ def validate_path_for_container_mount(path: pathlib.Path) -> None:
                 f"Invalid path: `{path}` cannot be under '{prefix}' which may overlap with a path"
                 f" in the container."
             )
+
+
+def validate_dataset(clp_config: CLPConfig, input_dataset: Optional[str]) -> Optional[str]:
```
haiqi96: Since the function name is "validate"_dataset, I don't think it should return a string. some suggestions have been made in PR1050

Bill-hbrhbr: ok. so here's the plan
1. If we are using `CLP_S`, we turn `None` into `default`
2. Move dataset validation into native scripts.

Bill-hbrhbr:
@coderabbitai
check for unused or missing imports

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -556,3 +562,21 @@ def validate_path_for_container_mount(path: pathlib.Path) -> None:
                 f"Invalid path: `{path}` cannot be under '{prefix}' which may overlap with a path"
                 f" in the container."
             )
+
+
+def validate_dataset(clp_config: CLPConfig, dataset: str) -> None:
+    """
+    Checks if the provided dataset currently exists in the metadata database.
```
haiqi96:
```suggestion
    Checks if `dataset` currently exists in the metadata database.
```
Or if Kirk has any better suggestion

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -556,3 +562,21 @@ def validate_path_for_container_mount(path: pathlib.Path) -> None:
                 f"Invalid path: `{path}` cannot be under '{prefix}' which may overlap with a path"
                 f" in the container."
             )
+
+
+def validate_dataset(clp_config: CLPConfig, dataset: str) -> None:
```
haiqi96: technically we only need to provide a DatabaseConfig?
```suggestion
def validate_dataset(db_config: Database, dataset: str) -> None:
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -7,14 +7,18 @@
...
 import typing
 import uuid
+from contextlib import closing
 from enum import auto
 from typing import List, Optional, Tuple
 
 import yaml
 from clp_py_utils.clp_config import (
     CLP_DEFAULT_CREDENTIALS_FILE_PATH,
+    CLP_DEFAULT_DATASET_NAME,
```
Bill-hbrhbr:
```suggestion
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -23,12 +27,14 @@
     WEBUI_COMPONENT_NAME,
     WorkerConfig,
 )
+from clp_py_utils.clp_metadata_db_utils import validate_and_cache_dataset
```
Bill-hbrhbr:
```suggestion
```

Bill-hbrhbr:
@coderabbitai
check for unused or missing imports

Bill-hbrhbr:
@coderabbitai
review imports

haiqi96 (approved):
Looks good to me.

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -556,3 +560,20 @@ def validate_path_for_container_mount(path: pathlib.Path) -> None:
                 f"Invalid path: `{path}` cannot be under '{prefix}' which may overlap with a path"
                 f" in the container."
             )
+
+
+def validate_dataset(db_config: Database, dataset: str) -> None:
+    """
+    Checks if `dataset` currently exists in the metadata database.
```
```suggestion
    Validates that `dataset` exists in the metadata database.

```

Let's try to include an empty line between the docstring's description and parameters.

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -556,3 +560,20 @@ def validate_path_for_container_mount(path: pathlib.Path) -> None:
...
+    :param db_config:
+    :param dataset:
+    :raise: ValueError
```
```suggestion
    :raise: ValueError if the dataset doesn't exist.
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -556,3 +560,20 @@ def validate_path_for_container_mount(path: pathlib.Path) -> None:
...
+    """
+    sql_adapter: SQL_Adapter = SQL_Adapter(db_config)
+    clp_db_connection_params: dict[str, any] = db_config.get_clp_connection_params_and_type(True)
+    table_prefix: str = clp_db_connection_params["table_prefix"]
```
```suggestion
    sql_adapter = SQL_Adapter(db_config)
    clp_db_connection_params = db_config.get_clp_connection_params_and_type(True)
    table_prefix = clp_db_connection_params["table_prefix"]
```

Instead of explicit type annotations here, ideally, we should

* update `get_clp_connection_params_and_type` to have an explicit return type.
* add mypy linting.

We can address the above as we clean up the code for additional linting. For now, I would prefer that we get rid of the explicit type annotations since it might be something hard to find and remove later (without just reading through all the code).

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -556,3 +560,20 @@ def validate_path_for_container_mount(path: pathlib.Path) -> None:
...
+    with closing(sql_adapter.create_connection(True)) as db_conn, closing(
+        db_conn.cursor(dictionary=True)
+    ) as db_cursor:
+        if dataset not in fetch_existing_datasets(db_cursor, table_prefix):
+            raise ValueError(f"Dataset `{dataset}` does not exist.")
```
```suggestion
            raise ValueError(f"Dataset `{dataset}` doesn't exist.")
```

kirkrodrigues (approved):
Modified the PR title directly.

PR #868, opened 2025-04-30
feat(clp-json): Use dataset-specific tables and archive directories for compression, decompression, and search.

Bill-hbrhbr:
# Description

This PR updates the compression, decompressoin, and search workflows to support dataset-specific operations when using the `CLP_S` storage engine:
* All metadata tables should now be prefixed with `clp_{dataset_name}_`, with the exception of the following:
  * `clp_datasets`: The central table where all datasets info are stored
  * `compression_jobs`
  * `compression_tasks`
  * `query_jobs`
  * `query_tasks`
* Archive paths for `fs` and `s3`, and stream output directories are suffixed with the dataset name to ensure physical isolation.

This PR succeeds: 
* #819
* #839 
* #831 
* #846
* #864

and covers step no. 3 of the dataset feature implementation [plan](https://docs.google.com/document/d/1BXDG1LSFOS7I52BRc0s1ahk_9MebCiH2HlLefLJ9unM/edit?tab=t.0#heading=h.kjcfz55f6si7).

# Checklist

* [X] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [X] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [X] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

For `clp-s`, the initial db table set:
image

CMD with interface changes taking effect:
```
bingranhu@baker22:/home/bingranhu/clp/build/clp-package$ ./sbin/compress.sh fs ~/dataset/data/
2025-04-30T05:57:07.264 INFO [compress] Compression job 1 submitted.
2025-04-30T05:57:08.799 INFO [compress] Compression finished.
2025-04-30T05:57:08.799 INFO [compress] Compressed 1.77KB into 889.00B (2.04x). Speed: 2.39KB/s.
bingranhu@baker22:/home/bingranhu/clp/build/clp-package$ ./sbin/search.sh "level: INFO"
: {"timestamp":"2025-04-10T23:10:45Z","level":"INFO","service":"auth-service","message":"User login successful","user_id":"abc123","ip":"192.168.1.10"}
: {"timestamp":"2025-04-10T23:10:45Z","level":"INFO","service":"auth-service","message":"User login successful","user_id":"abc123","ip":"192.168.1.10"}
: {"timestamp":"2025-04-10T23:10:45Z","level":"INFO","service":"auth-service","message":"User login successful","user_id":"abc123","ip":"192.168.1.10"}
[...]

PR #1042, opened 2025-06-26
fix(new-webui): Add support to query metadata from multiple datasets (fixes #1024).

### LINKED ISSUE #1024: Archive metadata SQL queries no longer function since tables are renamed
### Bug

PR #868 added prefixes to archive specific tables; however, webui used these tables to get metadata like compressed size and uncompressed size

Not exactly sure what the best solution for this is

### CLP version

bcb7f54 

### Environment

ubuntu

### Reproduction steps

Start webui see stats are missing and errors from server

davemarco:
This PR is blocked by #1004. We cannot test fix for UI, until multiple datasets can be ingested 

# Description

PR #868 modified the names of the metadata tables leading to issue #1024. The table names were changed, breaking the ui. 

The pr is fixes the issue by first querying all datasets, then issuing a large runtime generated query that include a subquery for each dataset, and finally combines the results using UNION ALL. The large query should be more performant than sending multiple queries for each dataset.

This is even more complicated since clp does not have datasets, so in that case, the dataset query is not sent, and the old query is used. 

The solution is a bit crude, but should be fine for a limited number of datasets. I will create an issue about another solution we discussed where we add a dataset column instead of renaming the tables.

# Checklist

  breaking change.

# Validation performed


components/webui/client/src/pages/IngestPage/SpaceSavings/index.tsx
```diff
@@ -1,72 +1,53 @@
...
-                    throw new Error("Space savings response is undefined");
-                }
-                setCompressedSize(spaceSavings.total_compressed_size);
-                setUncompressedSize(spaceSavings.total_uncompressed_size);
-            })
-            .catch((e: unknown) => {
-                console.error("Failed to fetch space savings stats", e);
-            });
-    }, []);
+    const {data: datasetNames = [], isSuccess: isSuccessDatasetNames} = useQuery({
```
davemarco, line 28: for reviewer, this enabled looks a bit weird, but this is there recommendation in the docs for dependant queries. https://tanstack.com/query/latest/docs/framework/react/guides/dependent-queries

components/webui/client/src/pages/IngestPage/Details/DetailsCard.tsx
```diff
@@ -14,12 +15,14 @@ interface DetailsCardProps {
  * @param props
  * @param props.title
  * @param props.stat
+ * @param props.isLoading
  * @return
  */
-const DetailsCard = ({title, stat}: DetailsCardProps) => {
+const DetailsCard = ({title, stat, isLoading}: DetailsCardProps) => {
```
hoophalab, line 21: This inheritance-style dependencies is too crazy.
We have
`DetailsCard -> StatCard -> DashboardCard`
You are passing `isLoading` all the way down to `DashboardCard` which is a typical problem of inheritance

Suggestion:
1. Combine all three into one: `DetailsCard` as all cards requires title and stat/content
2. `DetailsCard` accepts `title` as a property
3. `DetailsCard` accepts content text (i.e. `No timestamp data` and `0.00%`) as react node in children, so that space saving can have a different fontsize.
4. two options:
    1. All other cards uses `DetailsCard` directly and `isLoading` becomes a page state in zustand store
    2. Have a `...props` parameter in all concrete cards such as `File` `Message`... which are directed to `DetailsCard`

components/webui/client/src/api/sql.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Query the SQL server with the queryString.
+ *
+ * @param queryString
+ * @return
+ */
+const querySql = async <T>(queryString: string) => {
+    return axios.post<T>("/query/sql", {queryString});
+};
+
+export {querySql};
```
hoophalab: The original idea of having separate `sql.ts` is: the developer of ingestion page don't need to touch any other code -- every change can be done in `src/pages/IngestPage`.

However, this idea doesn't work. Since you are already create an `src/api` folder, then let's do it properly:

1. create a API module that one-to-one maps all restful endpoint to a typescript function.
2. create a backend manager class **on top of** the thin api that includes all shared fetchers such as `fetchDatasetNames`.

components/webui/client/src/pages/IngestPage/SpaceSavings/sql.ts
```diff
@@ -24,15 +42,90 @@ SELECT
...
-    SpaceSavingsItem,
-    SpaceSavingsResp,
+export type {SpaceSavingsItem};
+export {
+    buildMultiDatasetSpaceSavingsSql,
+    executeSpaceSavingsQuery,
+    fetchClpSpaceSavings,
+    fetchClpsSpaceSavings,
+    getSpaceSavingsSql,
+    SPACE_SAVINGS_DEFAULT,
```
hoophalab: Take a look at `sql.ts` — for some features, it just builds a SQL query string, and for others, it actually executes the query. This kind of inconsistency makes the interface confusing.

One idea: move all of this logic into the "backend manager class" mentioned in another comment. That way, it’s all in one place with a clear and consistent responsibility.

Also, consider doing the same for the `sql.ts` file in `IngestPage`. Move its logic into the backend manager as well, and only expose `fetch* functions` instead of `build*Sql` ones.

components/webui/client/src/api/sql.ts
davemarco: for 1. i think the src/api is already the api module no?.
for 2. I dont think we need a class since there is no saved state. But i think we can move shared fetchers like fetchDatasetNames to src/api/ maybe in a datasets.ts file

components/webui/client/src/pages/IngestPage/SpaceSavings/sql.ts
davemarco: I think these are different from the dataset query since they are not shared. Junhao original idea for the widget is that they can send a self-contained SQL query, so in some sense it makes sense these fetches at here and only the query/sql is in api module.

However, i do think i exported too many functions. I think we only need to export fetchClpSpaceSavings and fetchClpsSpaceSavings, the others can be removed

components/webui/client/src/pages/IngestPage/Details/DetailsCard.tsx
davemarco: I did also think these cards can be refactored when working on the code. But probably a different PR.

components/webui/client/src/api/sql.ts
hoophalab: > for 1. i think the src/api is already the api module no?

yes, but everything is mixed now. Can we create a module one-to-one maps each endpoint that fastify exposes?

> for 2. I dont think we need a class since there is no saved state. But i think we can move shared fetchers like fetchDatasetNames to src/api/ maybe in a datasets.ts file

yes, it doesn't need to be a class. but we should move shared fetchers to `src/api/ `

components/webui/client/src/pages/IngestPage/SpaceSavings/sql.ts
hoophalab: Ok, I think there are two layers of abstractions.

(1) [submitExtractStreamJob](https://github.com/davemarco/clp/blob/e463470a8693c8a17ce5b90a9156a947688fe7aa/components/webui/client/src/api/query.ts#L22), [querySql](https://github.com/davemarco/clp/blob/e463470a8693c8a17ce5b90a9156a947688fe7aa/components/webui/client/src/api/sql.ts#L11) ..., which only calls `axios.get`/`axios.post`

(2) `fetchClpSpaceSavings`, `fetchClpsSpaceSavings`, `fetchDatasetNames` build the query and returns the result as a promise.

We could either

1. move both functions in (1) (2) to `client/src/api`, which obviously deviates from Junhao's original idea

2. move functions in (1) to `client/src/api` and keep functions in (2) in `client/src/pages/*`

However, we shouldn't put `fetchDatasetNames` in `client/src/api` while leave `fetchClpSpaceSavings`, `fetchClpsSpaceSavings` in  `client/src/pages/*`.

If we choose method 2., then there will should be two `fetchDatasetNames` separate functions -- one in card, another one in search page

components/webui/client/src/pages/IngestPage/Details/DetailsCard.tsx
hoophalab: Can we create a quick refactor PR to address this before merging? It shouldn’t take long — I’m happy to submit one and you review that. However, we shouldn't let the current `isLoading` solution make it into the main branch.

davemarco: okay this is reasonable, but you will have to merge into this PR

components/webui/client/src/api/sql.ts
davemarco: for 1. do you mean like a new folder for each. I think that is okay

components/webui/client/src/pages/IngestPage/SpaceSavings/sql.ts
davemarco: okay I prefer one as lesser of two evils

components/webui/client/src/api/sql.ts
hoophalab: > okay I prefer one as lesser of two evils

If we go with that approach, then I feel it's best to:

1. We keep `src/api/search.ts`, `src/api/query.ts`, and `src/api/sql.ts` as they are.

2. We move the other fetchers into `src/fetchers.ts`.

hoophalab:
@davemarco we need to resolve conflicts as #1050 is merged

components/webui/client/src/api/sql.ts
davemarco: what about src/api/fetchers.ts?

davemarco: I think the fetchers are making requests so okay to be in the api folder, and keep one file out of the top level code

davemarco:
fyi @hoophalab - I merged main into this, and it looks like it still works

components/webui/client/src/pages/IngestPage/Details/index.tsx
```diff
@@ -1,104 +1,74 @@
...
     return (
         <div>
             <TimeRange
-                beginDate={beginDate}
-                endDate={endDate}/>
+                beginDate={dayjs.utc(details.begin_timestamp)}
+                endDate={dayjs.utc(details.end_timestamp)}
+                isLoading={isPending}/>
         </div>
     );
```
hoophalab, line 73: Quick question: I might be missing some context here. Why don’t we show Messages and Files for CLP-s? @davemarco

kirkrodrigues: The clp-s storage engine doesn't currently keep track of the number of log events and files in the archives.

components/webui/client/src/api/sql.ts
hoophalab: Yeah, sure. Let's put them in `src/api/fetchers.ts`.

Also, I’m thinking about adding a generic `fetchSqlFirstRow` function. A sketch

```
const fetchSqlFirstRow = async <T>(queryString: string): Promise<T> => {
    const resp = await querySql<T[]>(queryString);
    const [firstRow] = resp.data;
    if ("undefined" === typeof firstRow) {
        throw new Error(`SQL ${queryString} does not contain data.`);
    }
    return firstRow;
};
```

This way, we can keep all the `build*Sql functions` inside `src/pages/IngestPage` which aligns with Junhao’s original idea.

davemarco: i dont think this works since some like the dataset fetch multiple rows.

components/webui/client/src/pages/IngestPage/SpaceSavings/sql.ts
davemarco: After looking at this further I think its weird to have the build* in the pages code, and the fetch* in the api folder. I actually think the current approach is not that bad, and importing the fetch datasets from the datasets folder is not great but is not that terrible.

davemarco:
@hoophalab -
1.  I moved the the axios function into their own modules.
2. Removed some unnecessary export form the sql files. 

3. re: Moving fetch functions. Overall on further thought I think it is weird to have the build sql in one part of the code and fetch in another. Overall I think it is better as it is now

4. I am waiting on your card refactor, then I will merge main

davemarco:
@hoophalab - merged i main and there was a conflict. I need you to reapprove

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -871,15 +871,20 @@ def start_webui(
     # Read, update, and write back client's and server's settings.json
     clp_db_connection_params = clp_config.database.get_clp_connection_params_and_type(True)
     table_prefix = clp_db_connection_params["table_prefix"]
-    dataset: Optional[str] = None
     if StorageEngine.CLP_S == clp_config.package.storage_engine:
-        dataset = CLP_DEFAULT_DATASET_NAME
+        archives_table_name = ""
+        files_table_name = ""
+    else:
```
hoophalab, line 877: We may want to pass null values in the future, but this would require changes in `settings.json`'s typing and proper null value handling, instead of relying on the implicit conversion ` ${null} == "null"`.

@coderabbitai can you create an issue on this?

davemarco:
asking @junhaoliao for approval since @hoophalab dosent have write access

PR #1043, opened 2025-06-26
fix(package): Explicilty specify required fields for update in `update_archive_metadata` method. (fixes #1039)

### LINKED ISSUE #1039: bug(clp-json): Compression fails in `update_archive_metadata` because update function does not parse relevant fields out of `--print-archive-stats` output.
### Bug

In `update_archive_metadata` (`compression_task.py:94`) we assume that all fields in the archive statistics from `--print-archive-stats` directly corresponds to a field in the archives table, and attempt to dump every key and value into the archive stats table (after setting some default values).

After merging #1005 we now emit `"range_index"` as an array as well as the `"is_split"` flag. Since the current code makes no effort to validate which fields it tries to insert into the database the metadata update fails.

The `update_archive_metadata` function should be updated to explicitly insert only the fields of interest. This should be every field in the "archives" table (see `init-db.py`):
1. `begin_timestamp` (default `0`)
2. `end_timestamp` (default `0`)
3. `creator_id` -- (insert `""` to match current behaviour)
4. `creation_ix` -- (insert `0` to match current behaviour)
5. `size` -- this corresponds to compressed size
6. `uncompressed_size`
7. `id` -- this corresponds to archive id

### CLP version

8ccccce7998fa1690dbbc7ad2752ac76de30a012

### Environment

Ubuntu.

### Reproduction steps

Attempt any compression while using the `clp-json` package.

haiqi96:
# Description

As described in #1039 , the current `update_archive_metadata` relies on clp-s to output the exact fields for metadata update. However, this assumption is not always true and has been broken by a recent change.

This pr updates `update_archive_metadata` so it expliclity specifies the required fields from the clp-s's archive_stats.
This pr also updates the logic for handling the `default_fields`. We removed begin_ts and end_ts from the default list since they will be output by clp-s (when timestamp is not found, those fields will be 0). We also add a strict check to ensure the fields to add does not exist in the clp-s's archive stats.

# Checklist

  breaking change.

# Validation performed

Ran package with storage_engine = clp-s.
Confirmed that a single input file can be compressed without error.

### DIFF
--- a/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
+++ b/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
@@ -93,19 +93,29 @@ def update_job_metadata_and_tags(db_cursor, job_id, table_prefix, tag_ids, archi
 
 def update_archive_metadata(db_cursor, table_prefix, archive_stats):
     archive_stats_defaults = {
-        "begin_timestamp": 0,
-        "end_timestamp": 0,
         "creator_id": "",
         "creation_ix": 0,
     }
     for k, v in archive_stats_defaults.items():
-        archive_stats.setdefault(k, v)
-    keys = ", ".join(archive_stats.keys())
-    value_placeholders = ", ".join(["%s"] * len(archive_stats))
+        if k in archive_stats:
+            raise ValueError(f"Unexpected key '{k}' in archive stats")
+        archive_stats[k] = v
+
+    fields_to_update = [
+        "begin_timestamp",
+        "end_timestamp",
+        "id",
+        "size",
+        "uncompressed_size",
+    ]
+    fields_to_update.extend(archive_stats_defaults.keys())
+
+    keys = ", ".join(fields_to_update)
+    value_placeholders = ", ".join(["%s"] * len(fields_to_update))
     query = (
         f"INSERT INTO {table_prefix}{ARCHIVES_TABLE_SUFFIX} ({keys}) VALUES ({value_placeholders})"
     )
-    db_cursor.execute(query, list(archive_stats.values()))
+    db_cursor.execute(query, [archive_stats[key] for key in fields_to_update])
 
 
 def _generate_fs_logs_list(
--- a/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
+++ b/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
@@ -92,30 +92,30 @@ def update_job_metadata_and_tags(db_cursor, job_id, table_prefix, tag_ids, archi
 
 
 def update_archive_metadata(db_cursor, table_prefix, archive_stats):
-    archive_stats_defaults = {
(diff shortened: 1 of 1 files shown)


gibber9809 (dismissed):
LGTM. I think it's fine to not check that those keys exist like rabbit is suggesting since part of the intended behaviour of --print-archive-stats is for those fields to exist. We currently don't document what fields are guaranteed to exist in the output of --print-archive-stats, but that's a separate issue. Granted, I'd also be fine with explicitly checking that the fields exist if we want to be cautious.

I think the PR title is fine by me as well.

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -92,20 +92,30 @@ def update_job_metadata_and_tags(db_cursor, job_id, table_prefix, tag_ids, archi
 
 
 def update_archive_metadata(db_cursor, table_prefix, archive_stats):
```
kirkrodrigues, line 94: I think it'd be clearer for future devs if we implement it as follows:

* Create default_creation_stats with the default values
* Validate the default keys don't exist
* Iterate over `stats_to_update` and add each kv-pair to `default_creation_stats`
* The rest of the logic, except in the query, use `default_creation_stats`

Obviously, the names wouldn't make sense anymore, so how about renaming:

* `default_creation_stats` -> `stats_to_update`
* `stats_to_update` -> `required_stat_names`

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -92,20 +92,30 @@ def update_job_metadata_and_tags(db_cursor, job_id, table_prefix, tag_ids, archi
...
-    archive_stats_defaults = {
-        "begin_timestamp": 0,
-        "end_timestamp": 0,
-        "creator_id": "",
+    default_creation_stats = {
         "creation_ix": 0,
+        "creator_id": "",
```
kirkrodrigues, line 98:
```suggestion
        # Use defaults for values clp-s doesn't output
        "creation_ix": 0,
        "creator_id": "",
```

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -92,20 +92,30 @@ def update_job_metadata_and_tags(db_cursor, job_id, table_prefix, tag_ids, archi
...
     }
-    for k, v in archive_stats_defaults.items():
-        archive_stats.setdefault(k, v)
-    keys = ", ".join(archive_stats.keys())
-    value_placeholders = ", ".join(["%s"] * len(archive_stats))
+    for key, default_value in default_creation_stats.items():
```
kirkrodrigues:
```suggestion

    # Validate clp-s doesn't output the set kv-pairs
    for key, default_value in default_creation_stats.items():
```

PR #1053, opened 2025-06-30
fix(new-webui): Drop timezone from `Date`s returned by antd's `DatePicker.RangePicker` to match timezone (UTC) shown in search results (fixes #1085).

### LINKED ISSUE #1085: [new-webui] Time range selected via selector is offset from time range sent in query
### Bug

Time ranges selected via the selector are offset from the timezone shown, resulting in the wrong time range filter in the query.

### CLP version

f1979be993c1b0d166670576ee2e24e5d6cfcf87

### Environment

Firefox 140.0.2 (64-bit)

### Reproduction steps

* Start the package
* Compress a dataset
* Perform a search in the UI
* Select a custom time range
* Send the query
* Open the console
* Observe that the time range sent with the query is offset from the time range selected

davemarco:
# Description

@kirkrodrigues mentioned a bug that the range picker is not accurately selecting dates. 

PR fixes this issue by dropping any timezone offset supplied by antd range picker.

PR also drops `showNow` options since it selects the time in the local timezone. This does not work for our design since the whole in UI is in utc.

# Checklist

  breaking change.

# Validation performed

tested the submitted timestamp in utc is the same when (1) using the selector, and (2) when clearing the timestamp first then using the selector.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the date range picker in the custom time range selection to prevent clearing the selection and hide the "now" button, ensuring consistent timezone handling.


kirkrodrigues:
> however, the "allowClear" and "now" options reset the timezone to the local timezone.

How'd you figure that out?

components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/index.tsx
```diff
@@ -68,8 +68,12 @@ const TimeRangeInput = () => {
                 onChange={handleSelectChange}/>
             {timeRangeOption === TIME_RANGE_OPTION.CUSTOM && (
                 <DatePicker.RangePicker
+
+                    // `ShowNow` and `allowClear` options are disabled since they set the timezone
+                    //  of the range picker to the local timezone.
```
kirkrodrigues:
```suggestion
                    // NOTE: We disable `showNow` and `allowClear` since they set the timezone of
                    // the range picker to the local timezone.
```

davemarco:
> > however, the "allowClear" and "now" options reset the timezone to the local timezone.
> 
> How'd you figure that out?

I tested. If you look at logs from the UI in the console. You will see that the submitted timestamp is 4 hours off (after you clear it). The now option was also using my local time, and not the utc time.

kirkrodrigues:
Entering a timestamp by typing also seems to trigger the issue. Any ideas how we could fix/mitigate that as well?

davemarco:
> Entering a timestamp by typing also seems to trigger the issue. Any ideas how we could fix/mitigate that as well?

This is kind of annoying. Maybe we do a gross [manipulation](https://day.js.org/docs/en/manipulate/utc)  (i think the timeline does something like this) when updating the state ? I think you can use `utc(true)` and it will drop the existing offset. The manipulation should have no effect if already in utc, and if not, it will force it to utc?

It dosent seem antd provides good support for timezones.

Originally i wanted to avoid this, and I thought setting the starting timestamp in utc was elegant fix, but it turns out it dosent work. 

we will still have to drop "now" feature since it will be local time, but allow clear should be fine

kirkrodrigues:
> Maybe we just keep the selector in the local timezone, and due a gross [manipulation](https://day.js.org/docs/en/manipulate/utc) when submitting the query (i think the timeline does something like this)? I think you can use `utc(true)` and it will drop the existing offset. It dosent seem antd provides good support for timezones

Sounds reasonable for now.

davemarco:
@kirkrodrigues - i made the fix. It appears to work, overall it is not that messy.

components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/index.tsx
```diff
@@ -68,8 +72,9 @@ const TimeRangeInput = () => {
                 onChange={handleSelectChange}/>
             {timeRangeOption === TIME_RANGE_OPTION.CUSTOM && (
                 <DatePicker.RangePicker
+
```
kirkrodrigues:
```suggestion
```

kirkrodrigues (commented):
Could you update the PR description about why we removed `showNow` as well as the updated validation steps?

kirkrodrigues (approved):
For the PR title, how about:

```
fix(new-webui): Drop timezone from `Date`s returned by antd's `DatePicker.RangePicker` to match timezone (UTC) shown in search results.
```

PR #1049, opened 2025-06-28
fix(clp-s): Link msgpack-cxx to clp_s_clp_dependencies target.

wraymo:
# Description

The target `clp_s_clp_dependencies` did not link against `msgpack`, which is required by `clp/ffi/ir_stream/Serializer.cpp`. This patch adds msgpack to the target's link libraries to resolve missing header issues.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

+ Install `msgpack-cxx` to a custom (non-system) path
+ Without this change, it reports the error
```
In file included from /home/clp/components/core/src/clp/ffi/ir_stream/Serializer.hpp:10,
                 from /home/clp/components/core/src/clp/ffi/ir_stream/Serializer.cpp:1:
/usr/local/include/msgpack.hpp:10:10: fatal error: msgpack/object.hpp: No such file or directory
   10 | #include "msgpack/object.hpp"
```
+ With this change, `clp_s_clp_dependencies` is built successfully

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

No user-facing changes in this release.

PR #1055, opened 2025-07-01
fix(clp-s): Expose msgpack-cxx publicly via clp_s_archive_reader and clp_s_archive_writer; Add CLP_NEED_MSGPACKCXX flag when enabling clp_s_clp_dependencies target.

wraymo:
# Description

The target `clp_s_archive_reader` and `clp_s_archive_writer` did not publicly expose `msgpack-cxx`, which led to missing header errors when other targets included its headers. This PR fixes the issue by publicly linking `msgpack-cxx` to `clp_s_archive_reader`. Additionally, it adds the `CLP_NEED_MSGPACKCXX` flag when enabling the `clp_s_clp_dependencies` target—an omission from PR #1049.

# Checklist

  breaking change.

# Validation performed

+ Install `msgpack-cxx` to a custom (non-system) path
+ Without this change, it reports the error
```
src/components/core/src/clp_s/search/clp_search/Grep.hpp:11,
                 from /home/yscope/actions-runner/_work/velox/velox/velox/_build/release/_deps/clp-src/components/core/src/clp_s/search/clp_search/Grep.cpp:3:
/home/yscope/actions-runner/_work/velox/velox/velox/_build/release/_deps/clp-src/components/core/src/clp_s/search/clp_search/../../SingleFileArchiveDefs.hpp:8:10: fatal error: msgpack.hpp: No such file or directory
    8 | #include "msgpack.hpp"
```
+ With this change, `clp_s_clp_dependencies` is built successfully

## Summary by CodeRabbit

* **Chores**
  * Updated internal dependency configuration to include an additional required library.
  * Adjusted the order of linked libraries for improved consistency.

### DIFF
--- a/components/core/cmake/Options/options.cmake
+++ b/components/core/cmake/Options/options.cmake
@@ -248,6 +248,7 @@ function(set_clp_s_clp_dependencies_dependencies)
         CLP_NEED_BOOST
         CLP_NEED_CURL
         CLP_NEED_FMT
+        CLP_NEED_MSGPACKCXX
         CLP_NEED_NLOHMANN_JSON
         CLP_NEED_OPENSSL
         CLP_NEED_SPDLOG
--- a/components/core/src/clp_s/CMakeLists.txt
+++ b/components/core/src/clp_s/CMakeLists.txt
@@ -303,6 +303,7 @@ if(CLP_BUILD_CLP_S_ARCHIVEREADER)
                 PUBLIC
                 absl::flat_hash_map
                 clp_s::io
+                msgpack-cxx
                 nlohmann_json::nlohmann_json
                 PRIVATE
                 Boost::url
@@ -310,7 +311,6 @@ if(CLP_BUILD_CLP_S_ARCHIVEREADER)
                 clp_s::timestamp_pattern
                 ${CURL_LIBRARIES}
                 fmt::fmt
-                msgpack-cxx
                 spdlog::spdlog
         )
 endif()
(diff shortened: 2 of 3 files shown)


gibber9809 (changes requested):
Sorry, two things:

The first is that for this `Grep` include path (and in general `msgpack-cxx`) seems to get exposed through things built as part of `clp_s::archive_reader` or `clp_s::archive_writer` -- in this case it seems to be a chain of `Grep` -> `EncodedVariableInterpreter` -> `DictionaryReader` -> `ArchiveReaderAdaptor` -> `SingleFileArchiveDefs` -> `msgpack`. Since msgpack is getting exposed via the public interfaces of `archive_{reader,writer}` we should just publically link against `msgpack-cxx` for these two targets instead of adding it to the link flags for `clp_s::search`.

The second is that for your previous PR I forgot to mention that we should add the `CLP_NEED_MSGPACKCXX` to line 251 of `components/core/cmake/Options/options.cmake` to indicate that `clp_s::clp_dependencies` links against `msgpack-cxx`.

wraymo:
@gibber9809 But without adding it to `clp_s::search`, this target cannot be compiled individually?

gibber9809:
> @gibber9809 But without adding it to `clp_s::search`, this target cannot be compiled individually?

Since `clp_s::search` links against `clp_s::archive_reader` I don't think there's an issue -- there isn't a way to compile `clp_s::search` without compiling and and linking against `clp_s::archive_reader` anyway.

gibber9809 (commented):
Could we also make msgpack-cxx public for clp_s::archive_writer for the same reason?

gibber9809 (approved):
LGTM. Title is fine as well.

kirkrodrigues (approved):

PR #1064, opened 2025-07-02
fix(new-webui): Improve visibility of case-sensitivity toggle (resolves #1058). 

### LINKED ISSUE #1058: [new-webui] Improve visibility of case-sensitivity toggle
### Request

I think the current case sensitivity toggle has two issues:

* By default, it's not obvious that it's a button that can be clicked since it has no outline. I know other software uses this kind of border-less theming, but I don't think it works well for new users.
* I can't tell whether it's on or off since the only difference is that it has an outline when it's one on of the states.

### Possible implementation

* Give the button an outline at all times.
* Use colour to indicate whether it's turned on or off.

davemarco:
# Description

Changes toggle styling per issue #1058 

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **Style**
  * Improved the visual feedback of the case sensitivity toggle button, making its appearance change more clearly when enabled or disabled. The button now displays "Aa" directly and updates its colour and style based on the current state.


junhaoliao (approved):
let's try to fill the validation section in the future to speed up reviews

inactive:
image

inactive during a search:
image

active:
image

active during a search:
image

---

i suggest that when isCaseSensitive is inactive but the button is disable (during an in progress query), we can set the button variant to "text" to reduce confusions:
```
    let buttonColor: ButtonColorType = "default";
    let buttonVariant: ButtonVariantType = "outlined";
    if (isCaseSensitive) {
        buttonColor = "primary";
        buttonVariant = "solid";
    } else if (disabled) {
        buttonVariant = "text";
    }
```

otherwise, i think the PR is good to go

since @kirkrodrigues created the issue - do you have any more comments?

davemarco:
okay i made ur change, but left the colour

PR #1027, opened 2025-06-23
refactor(webui): Integrate existing Mongo SocketIO plugin into new Fastify architecture; Pass logger into plugin instead of entire Fastify Instance (resolves #885).  

davemarco:
# Description

PR #880 added new mongo plugin, but this was complete prior to #899 (fastify architecture refactor). This PR begins the process of migrating old plugins/routes from old architecture to new architecture.  There will be several more PRs. When all the PRs are finished, the `fastify-v2` directory can be flattened(its files/dir brought into src)

Specifically, this PR moves plugin to fastify-v2 directory, modifies the plugin to use shared mongo client, and passes the logger to plugin instead of entire fastify instance (which happens to resolve #885)

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

Tested a query, and still ran in webui

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **Refactor**
  * Improved logging by routing all log messages through the application's logger instead of the console.
  * Simplified the setup and integration of the MongoDB socket server, making plugin registration asynchronous and streamlining dependency injection.
  * Updated internal type import paths for consistency.

* **Chores**
  * Removed unused MongoDB initialization code.
  * Removed the MongoDB socket server plugin registration from the main application setup.

PR #1065, opened 2025-07-03
refactor(webui): Migrate extract stream route and plugin to new Fastify architecture.

davemarco:
This PR is dependent on #1027, #1028 being merged

# Description

PR moves extract stream route and plugin into new fastify architecture. This is a chain of PRs which started with https://github.com/y-scope/clp/pull/1027

There was alot of duplication between the plugin that submitted search jobs, and the old extract stream plugin (they both submit jobs to the same mysql database), so i completely refactored so they share more code. As a result, the PR is not trivial. 

# Checklist

  breaking change.

# Validation performed

tested streams with clp and clp-s

## Summary by CodeRabbit

* **New Features**
  * Added backend support and API route for stream file extraction with metadata retrieval.
  * Introduced schema validation for stream file extraction requests.

* **Refactor**
  * Consolidated job management into a unified query job manager for search, aggregation, and stream extraction.
  * Updated API routes and server plugins to use new job management classes and removed legacy database plugins.
  * Simplified server setup by removing old database-related routes and plugins.

* **Bug Fixes**
  * Enhanced validation and error handling for stream extraction jobs and metadata queries.

* **Chores**
  * Removed deprecated backend files, types, and utilities related to prior database and job management systems.


davemarco:
@hoophalab , i merged main into this and some smaller changes

components/webui/server/src/fastify-v2/routes/api/stream-files/index.ts
```diff
@@ -0,0 +1,87 @@
...
+
+            if (fastify.hasDecorator("s3Manager") && "undefined" !== typeof fastify.s3Manager) {
+                streamMetadata.path = await fastify.s3Manager.getPreSignedUrl(
+                    `s3://${settings.StreamFilesS3PathPrefix}${streamMetadata.path}`
+                );
+            } else {
+                streamMetadata.path = `/streams/${streamMetadata.path}`;
+            }
+
+            return streamMetadata;
```
hoophalab, line 80: Can we add an explicit type conversion here?
The return type of `getExtractedStreamFileMetadata` is  `StreamFileMongoDocument`, but the return type is `typeof StreamFileMetadataSchema`. It took me some time to confirm `StreamFileMongoDocument` is a superset of `typeof StreamFileMetadataSchema`. The old code doesn't have this problem because its response is not type safe at all xD
```suggestion

            return streamMetadata as Static<typeof StreamFileMetadataSchema>;
```
where `import {Static} from "@sinclair/typebox";`
* Do not commit suggestion directly as the syntax could be incorrect

components/webui/client/src/api/stream-files/index.ts
```diff
@@ -38,7 +38,7 @@ const submitExtractStreamJob = async ({
     onUploadProgress,
 }: SubmitExtractStreamJobProps): Promise<AxiosResponse<ExtractStreamResp>> => {
     return await axios.post(
-        "/query/extract-stream",
+        "/api/stream-files/extract",
```
hoophalab, line 41: qq: should this be considered as a breaking change `refactor(webui)!:` or it is not because this is an internal api?

kirkrodrigues: Our policy is to mark breaking changes for changes that would be visible to users (since that requires that we bump the relevant version number). So if this is purely internal to this repo, I would say it's not breaking.

components/webui/server/src/fastify-v2/routes/api/stream-files/index.ts
davemarco: I refactored the code to use the same base type for `StreamFileMetadataSchema` and `StreamFileMongoDocument` to make things clearer and prevent duplication.

We generally try and avoid type conversion, and I dont think neccesary here. Typebox/Fastify/Typescript will verify that the response matches the declared typebox type, like if you didnt return `StreamFileMetadataSchema` here, there would be a built error and linter will complain. Like there is already a type assertion in effect, so I dont believe we should cast it.

hoophalab: If that's the method we are going to take, then I would recommend rename `StreamFileMongoDocument` as  `StreamFileMetadata` (or `StreamFileMetadataSchemaType` or something similar).

davemarco: okay done

davemarco:
@hoophalab ready for rereview

components/webui/server/src/app.ts
junhaoliao, line 1: qq: what's the plan for the static server and s3manager plugins in this file? i.e., do we plan to migrate them as well in the near future, or to keep it in this "v1" plugin?

junhaoliao (approved):

components/webui/server/src/app.ts
davemarco: i already put up the s3 manager pr. The static i will do later. we are going to delete the v1 plugin

PR #1066, opened 2025-07-04
fix(clp-s): Add DictionaryWriter sources to clp_s_search and link against clp_s::io.

wraymo:
# Description

`clp_s_search` targets depends on `VariableEncoder`, which in turn depends on `VariableWriter`, However, `VariableWriter` was not included in the list of source files for the build, resulting in a "symbol not found" linker error. Consider the follow simple source file that links against `clp_s_search`,

```
#include "../VariableEncoder.hpp"
#include <vector>
#include <string>

int main() {
    clp_s::VariableEncoder encoder;
    std::string msg = "test log message";
    clp_s::LogTypeDictionaryEntry logtype_dict_entry;
    clp_s::VariableDictionaryWriter var_dict;
    std::vector<int64_t> encoded_vars;

    encoder.encode_and_add_to_dictionary(msg, logtype_dict_entry, var_dict, encoded_vars);
    return 0;
}
```
Compiling this will result in a linker error indicating a missing symbol for `VariableDictionaryWriter::add_entry`. This PR fixes this issue.

# Checklist

  breaking change.

# Validation performed

Applied the changes and there's no issue compiling the above code.

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration to include new dictionary-related components and dependencies. No changes to user-facing features.


gibber9809 (approved):
LGTM! I don't really like that we're forced to pull in writer dependencies on the search path because of one function in `VariableEncoder` that we don't use, but I think its fine to avoid refactoring this for now since we intend to delete this version of the CLP encoding code in the near future anyway.

For the title how about:
`fix(clp-s): Add DictionaryWriter sources to clp_s_search and link against clp_s::io.`

Just because the actual source files are `DictionaryWriter.{cpp,hpp}` and `DictionaryEntry.{cpp,hpp}`.

PR #428, opened 2024-06-07
core-clp: Rewrite wildcard matching method and add systematic unit tests (fixes #427).

kirkrodrigues:
# Description

This PR:

* rewrites `clp::string_utils::wildcard_match_unsafe_case_sensitive` both to fix #427, to slightly simplify the logic, and to add more comments to explain the algorithm;
* completely rewrites the unit tests for wildcard matching so that they (hopefully) systematically test all cases;
* replaces the wildcard performance test to be more realistic by matching against several lines (rather than a single line) from an example log file.

# Validation performed

Validated unit tests passed.

Issue #427, opened 2024-06-06
`clp::string_utils::wildcard_match_unsafe_case_sensitive` fails to match with certain queries


kirkrodrigues:
### Bug

For instance, `clp::string_utils::wildcard_match_unsafe_case_sensitive` fails to match 'a?c' with the wildcard string 'a\??'.

This bug affects both `clp` and `clp-s` but is hidden by #243 and #404.

Thanks to @wraymo for reporting it.

### CLP version

2c37cc6

### Environment

Ubuntu 22.04

### Reproduction steps

Write a unit test which tests `clp::string_utils::wildcard_match_unsafe_case_sensitive("a?c", "a\\??")`.

PR #1070, opened 2025-07-05
core-clp: Add function to preserve escaped wildcards during search (fixes #243).

### LINKED ISSUE #243: Escaped and wildcard "?" not properly handled during search
### Bug

Currently CLG replaces ?  with * in the search query because we don't handle ? wildcard and fallback to decompression and match.
However, this doesn't consider the case when a ? is escaped.

**_Note: the query discussed below are queries seen by CLP. if you enter the query on the commandline via bash, you may need to add extra escapes for bash to properly interpret the query._**

For example, considering the query `INFO \? TEXT`, which shall match a plain text ? in the log. current CLG will replace the ? with * and end up searching for the wrong results.

While I tried a quick fix (not submitted) for the issue above, there were other bugs that I noticed.

For example, considering the query `INFO \\? TEXT`. The query shall match a plain text \ in the log, then ? will server as wildcard and match any character. However the CLG is not returning the correct results.

### CLP version

5d6ff54d375af35f8aadd7d551672c255f627ab9

### Environment

22.04 Ubuntu

### Reproduction steps

Attaching log and queries that can be used for testing and reproducing the issue.

Currently running the query in the CLP won't generate the correctly matched results.

**_Note: the queries attached are what should be seen by CLP. if you enter the query on the commandline via bash, you may need to add extra escapes for bash to properly interpret the query._**

[query.txt](https://github.com/y-scope/clp/files/14031548/query.txt)
[log.txt](https://github.com/y-scope/clp/files/14031549/log.txt)

quinntaylormitchell:
# Description

Issue #243 describes a bug wherein queries containing the character sequence `\?` were not being matched with the appropriate search results.

`?` wildcard characters are replaced with `*` during search, but this was being done indiscriminately, i.e. `?` was being replaced with `*` regardless of whether or not it was escaped. This PR adds a helper function to `components/core/src/clp/Grep.cpp` called `replace_unescaped_wildcards()` which replaces the call to `std::replace()` that was being performed in `Grep::process_raw_query()`

`replace_unescaped_wildcards()` calls `std::replace_if()` for the string passed into it. The predicate function passed into `std::replace_if()` is a lambda function which returns `true` if the current (non-backslash) character is preceded by an even number of backslashes, i.e. that that character is not escaped. A boolean called `escaped` alternates between `false` and `true` upon encountering a `\` character.

`std::replace_if()` has the same complexity as `std::replace()`, and is single-run-through. The predicate (lambda) function does not require any caching other than the boolean `escaped`.

_NOTE: This PR does not fix the bug that affects queries that contain the character sequence `\\?`; this bug is described in issue #427 and fixed in PR #428._

# Checklist

  breaking change.

# Validation performed

Ran queries from [query.txt](https://github.com/y-scope/clp/files/14031548/query.txt) on the logs in [log.txt](https://github.com/y-scope/clp/files/14031549/log.txt), as described in the "Reproduction steps" section of issue #243. All queries using `\?` now return the correct results.

### DIFF
--- a/components/core/src/clp/Grep.cpp
+++ b/components/core/src/clp/Grep.cpp
@@ -496,6 +496,28 @@ SubQueryMatchabilityResult generate_logtypes_and_vars_for_subquery(
 }
 }  // namespace
 
+void replace_unescaped_wildcards(
+        std::string& search_string
+) {
+    auto unescaped = [escaped = false](char c) mutable
+    {
+        bool should_replace = ('?' == c && !escaped);
+        if ('\\' == c) {
+            escaped = !escaped;
+        } else {
+            escaped = false;
+        }
+        return should_replace;
+    };
+    
+    std::replace_if(
+        search_string.begin(),
+        search_string.end(),
+        unescaped,
+        '*'
+    );
+}
+
 std::optional<Query> Grep::process_raw_query(
         Archive const& archive,
         string const& search_string,
@@ -522,12 +544,9 @@ std::optional<Query> Grep::process_raw_query(
         // Replace '?' wildcards with '*' wildcards since we currently have no support for
         // generating sub-queries with '?' wildcards. The final wildcard match on the decompressed
         // message uses the original wildcards, so correctness will be maintained.
-        std::replace(
-                search_string_for_sub_queries.begin(),
-                search_string_for_sub_queries.end(),
-                '?',
-                '*'
-        );
+        // Use replace_unescaped_wildcards() to avoid changing escaped wildcard characters.        
+        replace_unescaped_wildcards(search_string_for_sub_queries);
+
         // Clean-up in case any instances of "?*" or "*?" were changed into "**"
         search_string_for_sub_queries
                 = clean_up_wildcard_search_string(search_string_for_sub_queries);


quinntaylormitchell:
@coderabbitai review

components/core/src/clp/Grep.cpp
```diff
@@ -522,12 +536,9 @@ std::optional<Query> Grep::process_raw_query(
...
         // generating sub-queries with '?' wildcards. The final wildcard match on the decompressed
         // message uses the original wildcards, so correctness will be maintained.
-        std::replace(
-                search_string_for_sub_queries.begin(),
-                search_string_for_sub_queries.end(),
-                '?',
-                '*'
-        );
+        // Use replace_unescaped_wildcards() to avoid changing escaped wildcard characters.
+        replace_unescaped_wildcards(search_string_for_sub_queries);
```
haiqi96: some comments:
1. instead of having `// Use replace_unescaped_wildcards() to avoid changing escaped wildcard characters.`, let's add docstring for replace_unescaped_wildcards instead.
2. Let's rename replace_unescaped_wildcards to something else since now it doesn't explicitly mention which wildcard it replaces. I see 3 options.
- rename the function name to `replace_unescaped_question_wildcards` so it's clear what the method is doing.
- Update the  function argument to accept a wildcard character, so we can let it either replace either * or ?
- Update the function argument to accept a list of wildcard characters, so we can replace both * and ?.

While the option1 is the easiest to implement, 2 and 3 may be able to support different use-case in the future. @kirkrodrigues  do you see any value in 2nd and 3rd option?

components/core/src/clp/Grep.cpp
```diff
@@ -496,6 +496,20 @@ SubQueryMatchabilityResult generate_logtypes_and_vars_for_subquery(
 }
 }  // namespace
 
+void replace_unescaped_wildcards(std::string& search_string) {
+    auto unescaped = [escaped = false](char c) mutable {
```
haiqi96: It's interesting to see this capture initialization and mutable approach, this seems to be quite modern feature and could be confusing for people who have never seen it before (actually, neither I or Kirk see such usage).
Let's keep the feature simple by:
1. let escaped to be a local variable under replace_unescaped_wildcards instead.
2. Capture and update escaped by reference.

components/core/src/clp/Grep.cpp
kirkrodrigues: Potentially. Perhaps we could make the function general so that it takes the escape character and character to escape as arguments; then we can move it into string_utils and unit test it easily.

haiqi96: do you mean `replacing character` and `character to replace`?

components/core/src/clp/Grep.cpp
```diff
@@ -496,6 +496,20 @@ SubQueryMatchabilityResult generate_logtypes_and_vars_for_subquery(
...
+        bool should_replace = ('?' == c && false == escaped);
```
haiqi96: We have both ? and '\\' defined in components/core/src/clp/regex_utils/constants.hpp. Can we use the constant from that file directly?

@kirkrodrigues asking Kirk since regex_utils/constants.hpp doesn't seem to be a general header that any random file is supposed to include.

haiqi96: nit
```suggestion
        auto const should_replace = ('?' == c && false == escaped);
```

components/core/src/clp/Grep.cpp
haiqi96: let's find a better name for this method.

components/core/src/clp/Grep.cpp
kirkrodrigues: I mean a signature like `replace_unescaped_wildcards(char escape_char, char char_to_escape, string& s)`. So in this case, we'd call it as `replace_unescaped_wildcards('\\', '?', search_string_for_sub_queries)`.

haiqi96: Moving this into string_utils and adding unit-tests sounds like the way to go.

haiqi96: why does it need to take `escape_char` as an argument? it would only makes sense to me if we rename the method to
`replace_unprefixed_wildcards(char prefix_char, char char_to_replace, string& s)`

kirkrodrigues: Sorry, I meant to rename the method as well to something like `replace_unescaped_char`. The reason to take in `escape_char` is that not every scenario uses `\\` as the escape character. I feel like more developers would understand what an `escape_char` is compared to `prefix_char`? Cause `prefix_char` implies that we only replace the character if it's preceded by `prefix_char`---but that doesn't take into account cases like `\\?` (where the `escape_char` is escaped).

haiqi96: It would be hard to imagine such use case, but as long as the docstring is clear I think it should be fine.
We should also let user specifiy src char and target char, right? not all use case replaces with ? with *.

kirkrodrigues: Right, right. Forgot to specify that, lol.

components/core/src/clp/Grep.cpp
kirkrodrigues: Unfortunately, I think that would create a circular dependency between the two libraries. Perhaps it's better to move the wildcard meta characters from regex_utils/constants into string_utils/constants.

quinntaylormitchell: @kirkrodrigues makes sense --- sounds a bit out-of-scope though, should we leave that for another PR?

components/core/src/clp/Grep.cpp
quinntaylormitchell: thanks @haiqi96 and @kirkrodrigues! I made the following changes:
- [x]  Move replace_unescaped_wildcards() function to string_utils.cpp and added docstring; Rename replace_unescaped_wildcards() to replace_unescaped_char()
- [x]  Modify replace_unescaped_char() to take escape_char, src_char, and target_char as arguments; update docstring and call to replace_unescaped_char() accordingly; lint
- [x]  Add unit tests for replace_unescaped_char() to test-string_utils.cpp
- [x]  Modify lambda function in replace_unescaped_char() by locally declaring ‘escaped’ variable and then passing by reference
- [x]  Change bool to auto const

components/core/src/clp/Grep.cpp
quinntaylormitchell: change is made!

components/core/src/clp/Grep.cpp
kirkrodrigues: I think it's within scope. Essentially we have a new use for the constants, so we need to move them into a place where they're re-usable.

components/core/src/clp/string_utils/string_utils.cpp
```diff
@@ -129,6 +129,23 @@ bool is_wildcard(char c) {
...
+        if (escape_char == c) {
+            escaped = !escaped;
+        } else {
+            escaped = false;
+        }
+        return should_replace;
+    };
+
+    std::replace_if(src_string.begin(), src_string.end(), unescaped, target_char);
+}
```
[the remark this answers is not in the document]
quinntaylormitchell: @kirkrodrigues what do you think about this? from my research about this, it seems that it is true that there's no guarantee that the predicate will be called in a specific order, but also I read that `src_string.begin()` and `src_string.end()` are `ForwardIterators` so they can only be incremented, which would makes it (almost) impossible for the computer to organize the calculations out-of-order

components/core/tests/test-string_utils.cpp
```diff
@@ -22,6 +23,43 @@ TEST_CASE("to_lower", "[to_lower]") {
...
+    check("trailing?", "trailing*");
+    check("multiple??q", "multiple**q");
+
+    // replacements with backslashes present
+    check("a\\\\?b", "a\\\\*b");
+    check("\\\\?abc", "\\\\*abc");
+    check("abc\\\\?", "abc\\\\*");
+
+    // no replacements with backslashes present
+    check("a\\?b", "a\\?b");
```
haiqi96: Wonder can we try using string literals https://en.cppreference.com/w/cpp/language/string_literal.html? it might be able to remove some extra \ from the strings.
People can get really confused when decoding so many escape, lol

components/core/src/clp/string_utils/string_utils.cpp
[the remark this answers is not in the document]
haiqi96: Interesting, I will need to take a look.

components/core/src/clp/Grep.cpp
```diff
@@ -519,15 +520,11 @@ std::optional<Query> Grep::process_raw_query(
...
-        std::replace(
-                search_string_for_sub_queries.begin(),
-                search_string_for_sub_queries.end(),
-                '?',
-                '*'
-        );
+        // Replace unescaped '?' wildcards with '*' wildcards since we currently have no support
+        // for generating sub-queries with '?' wildcards. The final wildcard match on the
+        // decompressed message uses the original wildcards, so correctness will be maintained.
+        replace_unescaped_char('\\', '?', '*', search_string_for_sub_queries);
```
haiqi96: as suggested in one of our comment, can we use constant instead of hardcoded magic char?

They are currently defined in regex_utils but you might need to move them to string_utils, as discces in https://github.com/y-scope/clp/pull/1070#discussion_r2195134176

components/core/src/clp/string_utils/string_utils.cpp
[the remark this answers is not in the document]
quinntaylormitchell: I feel like it might be a false concern because of the `ForwardIterators` thing that I mentioned, but let me know if you find anything @haiqi96

components/core/src/clp/Grep.cpp
quinntaylormitchell: @kirkrodrigues if I move the wildcard meta characters from `regex_utils/constants` into `string_utils/constants`, they are still needed in `components/core/src/clp/regex_utils/regex_translation_utils.cpp`. should I just include `string_utils/constants` in that file?

kirkrodrigues: Yeah.

components/core/tests/test-string_utils.cpp
quinntaylormitchell: the string literals helped I think!

components/core/src/clp/Grep.cpp
quinntaylormitchell: done!

components/core/src/clp/string_utils/string_utils.cpp
```diff
@@ -129,6 +129,23 @@ bool is_wildcard(char c) {
     return false;
 }
 
+void
+replace_unescaped_char(char escape_char, char src_char, char target_char, std::string& src_string) {
+    bool escaped = false;
+
+    auto unescaped = [escape_char, src_char, &escaped](char c) {
```
haiqi96: I feel that we don't need to expliclity specify what has to be captured.
Compare to a wildcard capture [&], I guess your current version prevents us from modifying the escape_char? 
Wonder if we make argument to be `char const escape_char, char const src_char` and use a [&[ capture.  In that case, can we modify the value (say, do something like escape_char = 'a') in the lambda?

components/core/src/clp/string_utils/string_utils.hpp
```diff
@@ -61,6 +61,16 @@ std::string replace_characters(
  */
 void to_lower(std::string& str);
 
+/**
+ * Replace unescaped instances of src_char with target_char in src_string
```
haiqi96:
```suggestion
 * Replace unescaped instances of `src_char` with `target_char` in `src_string`
```

components/core/src/clp/string_utils/string_utils.hpp
```diff
@@ -61,6 +61,16 @@ std::string replace_characters(
...
+ * @param escape_char The character which indicates escape
+ * @param src_char The character which should be replaced if unescaped
+ * @param target_char The character which should replace src_char
+ * @param search_string The string which needs to have src_char replaced
```
haiqi96: I think we can remove the comments for src_char, target_char and search_string.
Also don't forget to rename search_string to src_string in the comment.

components/core/src/clp/string_utils/string_utils.hpp
```diff
@@ -61,6 +61,16 @@ std::string replace_characters(
...
+ */
+void
+replace_unescaped_char(char escape_char, char src_char, char target_char, std::string& src_string);
```
haiqi96: Another nit. let's use trailinh return.
```suggestion
auto
replace_unescaped_char(char escape_char, char src_char, char target_char, std::string& src_string) -> void;
```

components/core/src/clp/string_utils/string_utils.cpp
```diff
@@ -129,6 +129,23 @@ bool is_wildcard(char c) {
     return false;
 }
 
+void
```
haiqi96: same comment, let's use trailing return.

components/core/tests/test-string_utils.cpp
```diff
@@ -22,6 +23,43 @@ TEST_CASE("to_lower", "[to_lower]") {
...
 }
 
+TEST_CASE("replace_unescaped_char", "[replace_unescaped_char]") {
+    auto check = [](std::string in, std::string const& expected) {
+        replace_unescaped_char('\\', '?', '*', in);
+        REQUIRE(in == expected);
+    };
+
+    // initial
+    check(R"(no change)", R"(no change)");
```
haiqi96: I feel we can put this under // additional edge cases

let's also replace `backslashes` -> `escapes` (since by default, escape is backslash)

Maybe also add a simple test case when using a custom escape character? you can add it as a new section. (and put existing tests under a section as well)

components/core/src/clp/string_utils/string_utils.cpp
quinntaylormitchell: sure, I'll change it to a wildcard capture.

components/core/src/clp/string_utils/string_utils.hpp
quinntaylormitchell: @haiqi96 sure I can do that; out of curiosity though, what's the advantage of using a trailing return? none of the other functions in string_utils use it

components/core/src/clp/string_utils/string_utils.cpp
quinntaylormitchell: same question as above

components/core/src/clp/string_utils/string_utils.cpp
```diff
@@ -129,6 +129,27 @@ bool is_wildcard(char c) {
     return false;
 }
 
+auto replace_unescaped_char(
+        char const escape_char,
+        char const src_char,
+        char const target_char,
+        std::string& src_string
+) -> void {
+    bool escaped = false;
```
haiqi96: nit. let's use braclet init.
```suggestion
    bool escaped{false};
```

quinntaylormitchell: oh ok, why is that?

components/core/tests/test-string_utils.cpp
```diff
@@ -22,6 +23,51 @@ TEST_CASE("to_lower", "[to_lower]") {
...
+        replace_unescaped_char(escape_char, src_char, target_char, in);
+        REQUIRE(in == expected);
+    };
+
+    SECTION("Default characters '\\', '?', and '*'") {
+        // replacements with no escape chars present
+        check('\\', '?', '*', R"(a?b)", R"(a*b)");
+        check('\\', '?', '*', R"(?leading)", R"(*leading)");
+        check('\\', '?', '*', R"(trailing?)", R"(trailing*)");
+        check('\\', '?', '*', R"(multiple??q)", R"(multiple**q)");
```
haiqi96: Can we also use defined constant from string_utils for the first 3 arguments?

quinntaylormitchell: for sure :)

components/core/src/clp/string_utils/string_utils.hpp
haiqi96: I believe the trailing return type has more value when it comes to some complicated templated functions, and it works better with auto. For this function though, using trailing doesn't have any functional impact. It's mainly to align with our new coding standard. The string_utils might be written at the time before we use this standard so its not update-to-date.

components/core/src/clp/string_utils/string_utils.cpp
haiqi96: The first one allows implict type conversion. For example, if you mistakenly write bool escape = -1; I believe it will work but it might not do what you intended it to do.

The second one require explicit type matching so you can't do something like bool escaped{-1}; (you can try and it should fail the compile).

There's also some implications on which constructor it calls. but for this case it is not that complicated. Try to use {} since it's our coding standard.

quinntaylormitchell: sounds good!

components/core/src/clp/string_utils/string_utils.hpp
quinntaylormitchell: gotcha, thanks!

components/core/src/clp/string_utils/string_utils.cpp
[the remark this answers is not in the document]
haiqi96: Yeah I agree with Quinn. I guess the concern of rabbit is that if the iterator is custom one instead of from simply from begin to end, the order of visiting the chars in the string maybe out of control.
But given that the iterator is string.begin(), it's hard to see how it can go out of order.
@coderabbitai any other insights or concern?

components/core/src/clp/string_utils/string_utils.cpp
```diff
@@ -129,6 +129,27 @@ bool is_wildcard(char c) {
...
 
+auto replace_unescaped_char(
+        char const escape_char,
+        char const src_char,
+        char const target_char,
+        std::string& src_string
+) -> void {
+    bool escaped{false};
+
+    auto unescaped = [&](char c) {
```
haiqi96: let's also rename the lambda method since the current name is not so intuitive. also the trailing return
```suggestion
    auto should_replace_char = [&](char c) -> bool {
```

components/core/src/clp/string_utils/string_utils.cpp
```diff
@@ -129,6 +129,27 @@ bool is_wildcard(char c) {
...
+        auto const should_replace = (src_char == c && false == escaped);
+        if (escape_char == c) {
+            escaped = !escaped;
+        } else {
+            escaped = false;
+        }
+        return should_replace;
+    };
+
+    std::replace_if(src_string.begin(), src_string.end(), unescaped, target_char);
```
haiqi96:
```suggestion
    std::replace_if(src_string.begin(), src_string.end(), should_replace_char, target_char);
```

components/core/src/clp/string_utils/string_utils.hpp
```diff
@@ -61,6 +61,16 @@ std::string replace_characters(
...
+/**
+ * Replace unescaped instances of `src_char` with `target_char` in `src_string`
+ * @param escape_char
+ * @param src_char
+ * @param target_char
+ * @param src_string The string which needs to have src_char replaced
+ */
+auto
+replace_unescaped_char(char escape_char, char src_char, char target_char, std::string& src_string)
+        -> void;
```
haiqi96: Sorry, just thought about naming again, both I and Kirk feel that `target` can be interpreted as both `src` and `dst`. 
let's do the following instead
src_char -> from_char
target_char -> to_char

Also. I think we should add comments for @param escape_char so it's clear to reader that this method accetps any char as an "escape char" instead of only looking at '\\\\'

components/core/src/clp/string_utils/string_utils.hpp
```diff
@@ -61,6 +61,15 @@ std::string replace_characters(
  */
 void to_lower(std::string& str);
 
+/**
+ * Replace unescaped instances of `from_char` with `to_char` in `src_string`
+ * @param escape_char The character which indicates escape; can be any character
```
haiqi96:
```suggestion
 * @param escape_char The user-specified character used for escape.
```

components/core/src/clp/string_utils/string_utils.hpp
```diff
@@ -61,6 +61,15 @@ std::string replace_characters(
...
+ * @param from_char
+ * @param to_char
+ * @param src_string The string which needs to have `from_char` replaced
```
haiqi96: I feel this should be very self explanatory.

```suggestion
 * @param src_string
```

components/core/src/clp/string_utils/constants.hpp
```diff
@@ -0,0 +1,13 @@
+#ifndef CLP_STRING_UTILS_CONSTANTS_HPP
+#define CLP_STRING_UTILS_CONSTANTS_HPP
+
+namespace clp::string_utils {
+// Wildcard meta characters
+constexpr char cZeroOrMoreCharsWildcard{'*'};
+constexpr char cSingleCharWildcard{'?'};
+
+// Escape meta characters
```
haiqi96: technically we don't need this comment
```suggestion
```

components/core/src/clp/string_utils/string_utils.hpp
```diff
@@ -61,6 +61,15 @@ std::string replace_characters(
  */
 void to_lower(std::string& str);
 
+/**
+ * Replace unescaped instances of `from_char` with `to_char` in `src_string`
+ * @param escape_char The user-specified character used for escape.
```
haiqi96:
```suggestion
 * @param escape_char The character used for escaping.
```

haiqi96:
For title, how about
`fix(core-clp): Preserve escaped wildcards in clp search queries (fixes #243)`

@kirkrodrigues any suggestions?

kirkrodrigues:
> For title, how about `fix(core-clp): Preserve escaped wildcards in clp search queries (fixes #243)`
> 
> @kirkrodrigues any suggestions?

How about this?

`fix(core-clp): Preserve escaped wildcards in queries (fixes #243).`

haiqi96 (dismissed):
The change looks good to me.

components/core/src/clp/string_utils/string_utils.hpp
```diff
@@ -61,6 +61,16 @@ std::string replace_characters(
...
 
+/**
+ * Replace unescaped instances of `from_char` with `to_char` in `src_string`.
+ *
+ * @param escape_char The character used for escaping
+ * @param from_char
+ * @param to_char
+ * @param src_string
+ */
+auto replace_unescaped_char(char escape_char, char from_char, char to_char, std::string& src_string)
```
kirkrodrigues: * For more logical grouping, can we move this function up to be just after `replace_characters`?
* Make sure to reflect the same ordering in `string_utils.cpp`.

components/core/src/clp/string_utils/string_utils.hpp
```diff
@@ -61,6 +61,16 @@ std::string replace_characters(
  */
 void to_lower(std::string& str);
 
+/**
+ * Replace unescaped instances of `from_char` with `to_char` in `src_string`.
+ *
+ * @param escape_char The character used for escaping
+ * @param from_char
+ * @param to_char
+ * @param src_string
```
kirkrodrigues:
```suggestion
 * @param src_string String in which to replace the characters
```

components/core/src/clp/string_utils/string_utils.hpp
kirkrodrigues: `src_string` sounds a little odd to me. How about just `str`?

components/core/tests/test-string_utils.cpp
```diff
@@ -22,6 +27,79 @@ TEST_CASE("to_lower", "[to_lower]") {
...
+                    char from_char,
+                    char to_char,
+                    std::string in,
+                    std::string const& expected) {
+        replace_unescaped_char(escape_char, from_char, to_char, in);
+        REQUIRE(in == expected);
+    };
+
+    SECTION("Default characters '\\', '?', and '*'") {
+        // replacements with no escape chars present
```
kirkrodrigues: Nit: We usually start comments with capital letters.

components/core/tests/test-string_utils.cpp
```diff
@@ -22,6 +27,79 @@ TEST_CASE("to_lower", "[to_lower]") {
...
+        // additional edge cases
+        check(cEscapeChar,
+              cSingleCharWildcard,
+              cZeroOrMoreCharsWildcard,
+              R"(no change)",
+              R"(no change)");
+        check(cEscapeChar, cSingleCharWildcard, cZeroOrMoreCharsWildcard, R"()", R"()");
+        check(cEscapeChar, cSingleCharWildcard, cZeroOrMoreCharsWildcard, R"(\)", R"(\)");
+        check(cEscapeChar, cSingleCharWildcard, cZeroOrMoreCharsWildcard, R"(\\)", R"(\\)");
+        check(cEscapeChar, cSingleCharWildcard, cZeroOrMoreCharsWildcard, R"(?\)", R"(*\)");
```
kirkrodrigues: Instead of repeating the `check` call repeatedly with similar arguments, you could create a vector of the values that change, and loop over that, right? In fact, Catch2 has some support for [this](https://github.com/catchorg/Catch2/blob/devel/docs/generators.md).

You could write something like:

```c++
        auto [str, expected] = GENERATE(
                as<std::pair<string, string>>{},

                // replacements with no escape chars present
                std::pair{R"(a?b)", R"(a*b)"},
                std::pair{R"(a\\?b)", R"(a\\*b)"},
                std::pair{R"(multiple??q)", R"(multiple**q)"}

                // ...
        );

        check(cEscapeChar, cSingleCharWildcard, cZeroOrMoreCharsWildcard, str, expected);
```

components/core/tests/test-string_utils.cpp
```diff
@@ -22,6 +27,79 @@ TEST_CASE("to_lower", "[to_lower]") {
...
+    auto check = [](char escape_char,
+                    char from_char,
+                    char to_char,
+                    std::string in,
+                    std::string const& expected) {
+        replace_unescaped_char(escape_char, from_char, to_char, in);
+        REQUIRE(in == expected);
+    };
+
+    SECTION("Default characters '\\', '?', and '*'") {
```
kirkrodrigues:
```suggestion
    SECTION("Conventional escape and wildcard characters") {
```

components/core/tests/test-string_utils.cpp
```diff
@@ -22,6 +27,79 @@ TEST_CASE("to_lower", "[to_lower]") {
...
+              cZeroOrMoreCharsWildcard,
+              R"(no change)",
+              R"(no change)");
+        check(cEscapeChar, cSingleCharWildcard, cZeroOrMoreCharsWildcard, R"()", R"()");
+        check(cEscapeChar, cSingleCharWildcard, cZeroOrMoreCharsWildcard, R"(\)", R"(\)");
+        check(cEscapeChar, cSingleCharWildcard, cZeroOrMoreCharsWildcard, R"(\\)", R"(\\)");
+        check(cEscapeChar, cSingleCharWildcard, cZeroOrMoreCharsWildcard, R"(?\)", R"(*\)");
+    }
+
+    SECTION("Custom escape character with custom src/target ('w' → 'e')") {
```
kirkrodrigues:
```suggestion
    SECTION("Unconventional escape and wildcard characters") {
```

components/core/tests/test-string_utils.cpp
quinntaylormitchell: I'll try it out!

components/core/src/clp/string_utils/constants.hpp
```diff
@@ -0,0 +1,12 @@
+#ifndef CLP_STRING_UTILS_CONSTANTS_HPP
+#define CLP_STRING_UTILS_CONSTANTS_HPP
+
+namespace clp::string_utils {
+// Wildcard meta characters
+constexpr char cZeroOrMoreCharsWildcard{'*'};
+constexpr char cSingleCharWildcard{'?'};
+
+constexpr char cEscapeChar{'\\'};
```
kirkrodrigues: How about `cWildcardEscapeChar`?

components/core/src/clp/string_utils/string_utils.cpp
```diff
@@ -113,6 +113,27 @@ string replace_characters(
...
+    bool escaped{false};
+
+    auto should_replace_char = [&](char c) -> bool {
+        auto const should_replace = (from_char == c && false == escaped);
+        if (escape_char == c) {
+            escaped = !escaped;
+        } else {
+            escaped = false;
+        }
+        return should_replace;
```
kirkrodrigues: Although this is functionally correct, it feels a little unintuitive to me. One part of that is we're first checking if `c` is `from_char`, but then immediately check if `c` is `escape_char`. `escape_char` and `from_char` can't be the same character (as we should note in the docstring), so writing the code in this way feels unintuitive. What do you think about writing the code like this:

```suggestion
        if (escaped) {
            escaped = false;
        } else if (escape_char == c) {
            escaped = true;
        } else if (from_char == c) {
            return true;
        }
        return false;
```

Essentially, since we're iterating over the string character by character, each if-block handles a single case:

- The character is escaped, in which case we can ignore it and do no replacement.
- The character is the escape character, in which case we should set the escaped flag and do no replacement.
- The character is the character to replace, in which case we should replace it.

components/core/src/clp/string_utils/string_utils.hpp
```diff
@@ -55,6 +55,17 @@ std::string replace_characters(
         bool escape
 );
 
+/**
+ * Replace unescaped instances of `from_char` with `to_char` in `str`.
+ *
```
kirkrodrigues, line 60:
```suggestion
 *
 * NOTE: `from_char` and `escape_char` must not be the same character. If they are, the function's
 * behaviour is undefined.
 *
```

components/core/src/clp/GrepCore.hpp
```diff
@@ -162,15 +163,16 @@ std::optional<Query> GrepCore::process_raw_query(
...
-        // generating sub-queries with '?' wildcards. The final wildcard match on the decompressed
-        // message uses the original wildcards, so correctness will be maintained.
-        std::replace(
-                search_string_for_sub_queries.begin(),
-                search_string_for_sub_queries.end(),
-                '?',
-                '*'
+        // Replace unescaped '?' wildcards with '*' wildcards since we currently have no support
+        // for generating sub-queries with '?' wildcards. The final wildcard match on the
+        // decompressed message uses the original wildcards, so correctness will be maintained.
```
kirkrodrigues:
```suggestion
        // Replace unescaped '?' wildcards with '*' wildcards since we currently have no support for
        // generating sub-queries with '?' wildcards. The final wildcard match on the decompressed
        // message uses the original wildcards, so correctness will be maintained.
```

components/core/tests/test-string_utils.cpp
```diff
@@ -22,6 +27,57 @@ TEST_CASE("to_lower", "[to_lower]") {
...
 }
 
+TEST_CASE("replace_unescaped_char", "[replace_unescaped_char]") {
+    auto check = [](char escape_char,
+                    char from_char,
+                    char to_char,
+                    std::string in,
+                    std::string const& expected) {
+        replace_unescaped_char(escape_char, from_char, to_char, in);
+        REQUIRE(in == expected);
```
kirkrodrigues:
```suggestion
        REQUIRE((in == expected));
```

components/core/src/clp/regex_utils/regex_translation_utils.cpp
```diff
@@ -177,7 +181,7 @@ auto normal_state_transition(
         case '.':
             state.set_next_state(TranslatorState::RegexPatternState::Dot);
             break;
-        case cEscapeChar:
+        case cWildcardEscapeChar:
```
kirkrodrigues: Sorry, I should've checked what this would affect before giving you the suggestion. I think the name `cEscapeChar` makes sense in the context of `regex_utils,` whereas in `string_utils,`, `cEscapeChar` is vague. I think what we really need to do is create a separate constant for the wildcard escape character (even though their value will be the same).

components/core/src/clp/string_utils/string_utils.hpp
```diff
@@ -58,13 +58,20 @@ std::string replace_characters(
...
  * @param from_char
  * @param to_char
  * @param str String in which to replace the characters
  */
-auto replace_unescaped_char(char escape_char, char from_char, char to_char, std::string& str)
-        -> void;
+auto replace_unescaped_char(
+        char const escape_char,
+        char const from_char,
+        char const to_char,
```
kirkrodrigues: Adding `const` for primitives in a function prototype has no effect. Did you mean to add them in the function's definition?

quinntaylormitchell: ah no, they are already `const` in the definition, I was just following Rabbit that said it would be cleaner to have `const` in the prototype too. I can change them back

kirkrodrigues:
clp-core-build-macos failures are unrelated and due to #1218.

PR #1095, opened 2025-07-10
feat(clp-package)!: Add query engine option to support starting only compression and UI components when using the Presto query engine.

davemarco:
# Description

### Background:
The UI team is working on making the [package compatible](https://docs.google.com/document/d/1pDvZdQdvYCFnXzLmngLadsVC62EYOy8J2SUW5MKoR3k/edit?tab=t.0#heading=h.pfeqv8zax5vw) with the Presto connector.

This PR disables unnecessary components when using Presto and passes a corresponding flag to the Web UI.

### Changes:

Introduced a new query_engine option in the package config, which can be set to either "native" or "presto".

"native" refers to the current behavior of the package.

"presto" disables internal query-related components, as querying will now be handled by Presto.

Note:
The stop-clp.sh script has not been modified, as it already gracefully handles the case where components are not running (i.e., it does not fail if it attempts to stop components that were never started).

# Checklist

  breaking change.

No docs for this until fully implemented

# Validation performed

Started and stopped the package with both presto and native options

## Summary by CodeRabbit

- New Features
  - Added configurable query engine options (CLP, CLP‑S, Presto) across the app, including config files and the web UI.
  - Startup now launches only components compatible with the selected query engine and provides a clear message when a requested target can’t run with the current configuration.

- Chores
  - Updated packaging to set the query engine automatically, aligning it with the selected storage engine.
  - Refreshed configuration template to include an explicit query_engine setting.


davemarco:
@haiqi96 @gibber9809 , since this isnt really a UI pr, i asked kirk for who should review, and he mentioned u guys

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -1106,6 +1108,16 @@ def main(argv):
...
         clp_config = load_config_file(config_file_path, default_config_file_path, clp_home)
 
+        # Exit early if target part of native query engine but Presto is configured
+        if QueryEngine.PRESTO == clp_config.package.query_engine:
+            if target in (
+                QUERY_SCHEDULER_COMPONENT_NAME,
+                QUERY_WORKER_COMPONENT_NAME,
+                REDUCER_COMPONENT_NAME,
+            ):
+                logger.error(f"{target} not available when using Presto as query engine")
```
haiqi96: Nit:
```suggestion
                logger.error(f"{target} not available when using {clp_config.package.query_engine} as query engine")
```

Maybe can also create a variable query_engine for reuse.

davemarco:
also tagging @junhaoliao since this may be used in cloud

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
davemarco: what about just `QueryEngine.PRESTO`

haiqi96: Sure, works for me.

davemarco:
I'm okay with devin's suggestion. It will also add an extra line above each component for for the new condition.
`ANY_COMPONENT_NAME in runnable_components:` 
@haiqi96 what do u think

haiqi96:
> I'm okay with devin's suggestion. It will also add an extra line above each component for for the new condition. `ANY_COMPONENT_NAME in runnable_components:` @haiqi96 what do u think

Sure, this sounds like a more general approach and can be extended in the future

davemarco:
@gibber9809 @haiqi96 - I changed per ur comment

davemarco:
I also added some validation to prevent using clp text with presto

components/package-template/src/etc/clp-config.yml
```diff
@@ -12,6 +12,7 @@
 #
 #package:
 #  storage_engine: "clp"
+#  query_engine: "native"
```
kirkrodrigues: I think this can be confusing since we're sort of mixing terms.

Currently:

| storage_engine | query_engine | Meaning                                |
|----------------|--------------|----------------------------------------|
| clp            | native       | Compress with clp, query with clp      |
| clp-s          | native       | Compress with clp-s, query with clp-s  |
| clp            | presto       | X Unsupported                          |
| clp-s          | presto       | Compress with clp-s, query with Presto |

So, "native" means "clp" or "clp-s" depending on the storage engine. In some sense, that prevents user error, but we then need to explain what "native" means.

On the other hand, if we do this:

| storage_engine | query_engine | Meaning                                |
|----------------|--------------|----------------------------------------|
| clp            | clp          | Compress with clp, query with clp      |
| clp            | clp-s        | X Unsupported                          |
| clp-s          | clp-s        | Compress with clp-s, query with clp-s  |
| clp-s          | clp          | X Unsupported                          |
| clp            | presto       | X Unsupported                          |
| clp-s          | presto       | Compress with clp-s, query with Presto |

Then, I think it's clearer what query_engine's possible values mean. We do need to do some extra validation in the script, to prevent clp w/ clp-s or clp-s w/ clp, but that shouldn't be hard.

All that said, I think you mentioned we would need to add more options for the query_engine field?

davemarco: I used "native" so users wouldnt need to change two fields when switching packages.  "default" was another option. Also i guess the query engine for clp is clg which maybe is slightly confusing. 

Yes I want to add a presto section, outside of package. something like this
```
presto:
  host: "localhost"
  port: 8080
```

gibber9809 (dismissed):
Sorry for taking a while to re-review. Talked with @kirkrodrigues and have one more idea to clean up the component starting. Generally the idea is that we can clean up this mix of targets, components, and runnable components using some set intersections.

So
1. Add a utility function (probably in `clp_config.py`) `get_components_for_target` that tells us the set of components a user is requesting to start for a target:
`components_to_start = get_components_for_target(target)`

2. Do a set intersection between the components to start and the runnable components: `components_to_start = components_to_start.intersect(set(runnable_components))`

3. If the intersection is empty then log an error -- I think this would replace the current error log around `start_clp.py:1113`

4. Rewrite the two level checks to start each components as:
```
if COMPONENT_X in components_to_start:
  # start_x()
...
```

davemarco:
@gibber9809
How should we handle actions that aren’t tied to explicit components?
For example, the `controller `target runs the `create_db_tables` action, but not `start_db`.

This doesn’t seem compatible with the format you suggested:

```
if COMPONENT_X in components_to_start:
    # start_x()
 ```

**Comment by davemarco** [comment]
@kirkrodrigues rerequesting review

File: components/clp-package-utils/clp_package_utils/scripts/start_clp.py [hunk 3]
```diff
@@ -22,9 +22,11 @@
     COMPRESSION_WORKER_COMPONENT_NAME,
     CONTROLLER_TARGET_NAME,
     DB_COMPONENT_NAME,
+    get_components_for_target,
     QUERY_JOBS_TABLE_NAME,
     QUERY_SCHEDULER_COMPONENT_NAME,
     QUERY_WORKER_COMPONENT_NAME,
+    QueryEngine,
```
kirkrodrigues:
```suggestion
```

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -1106,6 +1109,15 @@ def main(argv):
         config_file_path = pathlib.Path(parsed_args.config)
         clp_config = load_config_file(config_file_path, default_config_file_path, clp_home)
 
+        runnable_components = clp_config.get_runnable_components()
+        components_to_start = get_components_for_target(target)
+        components_to_start = components_to_start.intersection(runnable_components)
+
+        # Exit early if no components to start
+        if not components_to_start:
```
kirkrodrigues:
```suggestion
        if len(components_to_start) == 0:
```

Although this is less Pythonic, it's more explicit and is less error prone in cases where developers forget the type of the variable.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -1,6 +1,6 @@
 import pathlib
 from enum import auto
-from typing import Literal, Optional, Union
+from typing import List, Literal, Optional, Set, Union
```
```suggestion
from typing import Literal, Optional, Set, Union
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -43,12 +43,48 @@
...
+    RESULTS_CACHE_COMPONENT_NAME,
+    COMPRESSION_SCHEDULER_COMPONENT_NAME,
+    COMPRESSION_WORKER_COMPONENT_NAME,
+    QUERY_SCHEDULER_COMPONENT_NAME,
+    QUERY_WORKER_COMPONENT_NAME,
+    REDUCER_COMPONENT_NAME,
+    WEBUI_COMPONENT_NAME,
+}
+
+PRESTO_RUNNABLE_COMPONENTS = {
```
kirkrodrigues: This name makes it sound like these are the Presto components we're going to start rather than the components that we'll run when using the Presto query engine. I tried renaming it, but I don't think it can make it much less confusing. That said, with a bit of refactoring, I think things can be generally more clear. I can't make suggestions on all relevant lines so I'll just describe the changes:

1. Define these general sets of components above `# Target names`

    ```python
    GENERAL_SCHEDULING_COMPONENTS = {
        QUEUE_COMPONENT_NAME,
        REDIS_COMPONENT_NAME,
    }

    COMPRESSION_COMPONENTS = GENERAL_SCHEDULING_COMPONENTS | {
        DB_COMPONENT_NAME,
        COMPRESSION_SCHEDULER_COMPONENT_NAME,
        COMPRESSION_WORKER_COMPONENT_NAME,
    }

    QUERY_COMPONENTS = GENERAL_SCHEDULING_COMPONENTS | {
        DB_COMPONENT_NAME,
        QUERY_SCHEDULER_COMPONENT_NAME,
        QUERY_WORKER_COMPONENT_NAME,
        REDUCER_COMPONENT_NAME,
    }

    UI_COMPONENTS = {
        RESULTS_CACHE_COMPONENT_NAME,
        WEBUI_COMPONENT_NAME,
    }

    ALL_COMPONENTS = COMPRESSION_COMPONENTS | QUERY_COMPONENTS | UI_COMPONENTS
    ```

2. Define a target to components map, after the target names:

    ```python
    TARGET_TO_COMPONENTS = {
        ALL_TARGET_NAME: ALL_COMPONENTS,
        CONTROLLER_TARGET_NAME: GENERAL_SCHEDULING_COMPONENTS | {
            COMPRESSION_SCHEDULER_COMPONENT_NAME,
            QUERY_SCHEDULER_COMPONENT_NAME,
        }
    }
    ```

3. In get_runnable_components, apply the suggestions I made.

4. In get_components_for_target, apply the suggestions I made.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -757,6 +826,12 @@ def load_redis_credentials_from_file(self):
                 f"Credentials file '{self.credentials_file_path}' does not contain key '{ex}'."
             )
 
+    def get_runnable_components(self) -> Set[str]:
+        if QueryEngine.PRESTO == self.package.query_engine:
+            return PRESTO_RUNNABLE_COMPONENTS
```
```suggestion
            return COMPRESSION_COMPONENTS | UI_COMPONENTS
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -757,6 +826,12 @@ def load_redis_credentials_from_file(self):
...
+        else:
+            return ALL_RUNNABLE_COMPONENTS
```
```suggestion
            return ALL_COMPONENTS
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -791,3 +866,14 @@ def dump_to_primitive_dict(self):
...
 
         return d
+
+
+def get_components_for_target(target: str) -> Set[str]:
+    if target == ALL_TARGET_NAME:
+        return ALL_RUNNABLE_COMPONENTS
+    elif target == CONTROLLER_TARGET_NAME:
+        return CONTROLLER_COMPONENTS
+    elif target in ALL_RUNNABLE_COMPONENTS:
```
```suggestion
    if target in TARGET_TO_COMPONENTS:
        return TARGET_TO_COMPONENTS[target]
    elif target in ALL_COMPONENTS:
```

davemarco:
will test building package again before merging

kirkrodrigues (approved):
feat(clp-package)!: Add query engine option to support starting only compression components when using the Presto query engine.

davemarco:
builds now. Also tested package startup for clp, and presto, and did not encounter issues

davemarco:
re: title it also starts the UI components right

kirkrodrigues:
> re: title it also starts the UI components right

True. How about:

```
feat(clp-package)!: Add query engine option to support starting only compression and UI components when using the Presto query engine.
```

I tested the tasks to build the package tars and they worked (with the hotfix for the sed issue).

PR #1099, opened 2025-07-11
refactor(webui): Migrate example tests to new Fastify architecture.

davemarco:
# Description

PR moves example tests into new fastify architecture. This is a chain of PRs which started with https://github.com/y-scope/clp/pull/1027

I removed building the app from the test, since we are only testing the route.

I left some unused test helpers since they are new(after refactor), and may be useful for future integration tests. 

# Checklist

  breaking change.

# Validation performed

Tests pass

## Summary by CodeRabbit

* **Refactor**
  * Updated internal route handling and plugin registration for improved consistency.
* **Tests**
  * Adjusted test setup for example routes without changing test behaviour or outcomes.

### DIFF
--- a/components/webui/server/src/test/example.test.ts
+++ b/components/webui/server/src/test/example.test.ts
@@ -1,11 +1,13 @@
+import fastify from "fastify";
 import {StatusCodes} from "http-status-codes";
 import tap, {Test} from "tap";
 
-import {build} from "./tap.js";
+import routes from "../fastify-v2/routes/api/example/index.js";
 
 
 tap.test("Tests the example routes", async (t: Test) => {
-    const server = await build(t);
+    const server = fastify();
+    server.register(routes);
 
     let resp = await server.inject({
         method: "GET",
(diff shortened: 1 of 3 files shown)


components/webui/server/src/test/example.test.ts
```diff
@@ -1,11 +1,13 @@
...
+import routes from "../fastify-v2/routes/api/example/index.js";
 
 
 tap.test("Tests the example routes", async (t: Test) => {
-    const server = await build(t);
+    const server = fastify();
+    server.register(routes);
 
     let resp = await server.inject({
         method: "GET",
```
hoophalab, line 13: github cannot comment on unchanged lines, but the url below this line should be "/api/example/get/Alice" right?

davemarco: annoyingly no because the route is imported directly, i believe fastify only takes the directory structure if you use the auto import method we are using in app.ts. 

So this is correct

davemarco:
fyi u can test with npm run test. u should see all pass

junhaoliao (approved):
lgtm. deferring to @hoophalab 's review

PR #1104, opened 2025-07-14
build(core): Enable link-time optimization for improved runtime performance.

LinZhihao-723:
# Description

From our microbenchmark result, enabling LTO (link-time optimization) can achieve up to 10% performance boost for clp-text and clp-json (with dependent libraries linked statically), tested on real user input. As its value has already been proven, this PR formally enables LTO through CMake for all core targets.

### clp-text

The following benchmark was run on an Intel Core i9-14900K processor with 64GB of RAM, in Ubuntu 22.04 WSL, using GCC 11 for compilation. The affinity was set to core 0-3 (performance cores). The time is reported by the wall clock. Tested on an 18GB dataset (raw text) from our unstructured solution users.

Compression:
Without LTO: 108.8s
With LTO: 95.6s
Improvement: 13.8%

Search (with query `*original publishedPatch*`):
Without LTO: 28.2s
With LTO: 27.5s
Improvement: 2.5%

Decompression:
Without LTO: 51.5s
With LTO: 50.7s
Improvement: 1.6%

### clp-json

The following benchmark was run on an Intel Core i7-9700K processor with 16GB of RAM, in Ubuntu 22.04, using GCC 11 for compilation. The time is reported by the wall clock. Tested on a 640MB dataset (kv-ir, 8.8GB before compression) from our structured solution users.

Compression:
Without LTO: 153.3s
With LTO: 150.9s
Improvement: 1.6%

Search (with query `*time*`):
Without LTO: 22.61s
With LTO: 21.72s
Improvement: 4.1%

Decompression:
Without LTO: 23.82s
With LTO: 23.37s
Improvement: 1.9%

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **Chores**
  * Improved build configuration to automatically enable link-time optimisation (LTO) during Release builds when supported, with clear status messages.


components/core/CMakeLists.txt
```diff
@@ -21,6 +21,15 @@ if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
 endif()
 
+include(CheckIPOSupported)
+check_ipo_supported(RESULT IPO_SUPPORTED OUTPUT IPO_CHECK_MSG)
+if (IPO_SUPPORTED)
+    message(STATUS "Link-time optimization enabled.")
+    set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)
+else()
+    message(WARNING "Link-time optimization is not enabled: ${IPO_CHECK_MSG}")
+endif()
```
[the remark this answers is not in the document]
LinZhihao-723: Done. Enabling only for release build.

kirkrodrigues:
Can we add some details about the microbenchmark performed?

LinZhihao-723:
> Can we add some details about the microbenchmark performed?

Added in the PR description.

LinZhihao-723:
> Can we resturture the LTO optimization as 'release' build opt-in feature? i.e. LTO is enabled or no-op rather than LTO is enabled or disabled. Rigjt now, functionally it's opt-in based solution, but in the log message, we are still suggesting LTO is a feature we disable.

I'm not sure I fully understand what u suggest. The reason to "disable" it for non-release build is to avoid paying longer build time when it's unnecessary. At this point I don't see a need to allow users to turn it on/off, since cmake hides the underlying compiler details and the compiler may not even support LTO when users enable it.

kirkrodrigues:
> > Can we resturture the LTO optimization as 'release' build opt-in feature? i.e. LTO is enabled or no-op rather than LTO is enabled or disabled. Rigjt now, functionally it's opt-in based solution, but in the log message, we are still suggesting LTO is a feature we disable.
> 
> I'm not sure I fully understand what u suggest. The reason to "disable" it for non-release build is to avoid paying longer build time when it's unnecessary. At this point I don't see a need to allow users to turn it on/off, since cmake hides the underlying compiler details and the compiler may not even support LTO when users enable it.

I *think* (I might be wrong) Jack means that it's odd to print a warning or status message when LTO is not enabled, since the only reason it wouldn't enabled is because the processor doesn't support it or it's not turned on for debug builds. I kind of agree in that regard. It's probably clearer to print a message only when it's turned on.

LinZhihao-723:
> > > Can we resturture the LTO optimization as 'release' build opt-in feature? i.e. LTO is enabled or no-op rather than LTO is enabled or disabled. Rigjt now, functionally it's opt-in based solution, but in the log message, we are still suggesting LTO is a feature we disable.
> > 
> > 
> > I'm not sure I fully understand what u suggest. The reason to "disable" it for non-release build is to avoid paying longer build time when it's unnecessary. At this point I don't see a need to allow users to turn it on/off, since cmake hides the underlying compiler details and the compiler may not even support LTO when users enable it.
> 
> I _think_ (I might be wrong) Jack means that it's odd to print a warning or status message when LTO is not enabled, since the only reason it wouldn't enabled is because the processor doesn't support it or it's not turned on for debug builds. I kind of agree in that regard. It's probably clearer to print a message only when it's turned on.

Sure. I guess I should drop the two "not enabled" messages then.

components/core/CMakeLists.txt
```diff
@@ -21,6 +21,15 @@ if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     set(CMAKE_BUILD_TYPE "${default_build_type}" CACHE STRING "Choose the type of build." FORCE)
 endif()
 
+if(CMAKE_BUILD_TYPE MATCHES "Release")
+    include(CheckIPOSupported)
+    check_ipo_supported(RESULT IPO_SUPPORTED OUTPUT IPO_CHECK_MSG)
```
kirkrodrigues:
```suggestion
    check_ipo_supported(RESULT IPO_SUPPORTED OUTPUT)
```

Since we don't use `IPO_CHECK_MSG` any more.

components/core/CMakeLists.txt
```diff
@@ -21,6 +21,15 @@ if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     set(CMAKE_BUILD_TYPE "${default_build_type}" CACHE STRING "Choose the type of build." FORCE)
 endif()
 
+if(CMAKE_BUILD_TYPE MATCHES "Release")
```
kirkrodrigues:
```suggestion
if (CMAKE_BUILD_TYPE MATCHES "Release")
```

For consistency.

PR #1107, opened 2025-07-16
refactor(build): Remove `CLP_NEEDS_X` flags for marking required dependencies; bump ystdlib to y-scope/ystdlib-cpp@c03806a.

gibber9809:
# Description

This PR removes the `CLP_NEEDS_X` flags that were used to manually track library dependencies of each target and switches to simply calling `find_package` (or equivalent) for each dependency of each target guarded by its `CLP_BUILD_X` flag. This makes it so that required dependencies can be resolved next to their targets instead of requiring book-keeping in a separate file.

To simplify the process of resolving each dependency multiple times and avoid duplicating logic to set up flags specific to certain dependencies we add utility functions for resolving each dependency. These functions are named `clp_find_X` where `X` is the name of the dependency.

We also add two functions `get_clp_checked_find(X)` and `set_clp_checked_find(X)` that the `clp_find_X` functions can use to mark and check whether they have previously run. This is currently only used by `clp_find_ystdlib` to make the setup logic run only once, since the way we set up `ystdlib` will error out if run multiple times, and by `clp_find_mariadb` to only emit a warning message one time.

Dev docs describing how to use `clp_find_X` vs `find_package(X)` as well as the `CLP_BUILD_Y` flags will be added in a follow-up PR.

# Checklist

  breaking change.

# Validation performed

* Validated that our velox clp connector from the velox 0.293 build in our fork builds if pointing to the version of the clp repo from this branch
* Validated that the build succeeds when any combination of `CLP_BUILD_TESTING` and `CLP_BUILD_EXECUTABLES` is `OFF`

## Summary by CodeRabbit

- Refactor
  - Centralized dependency discovery and linking for more consistent library detection and simpler validation flow.

- Chores
  - Standardized module include resolution and expanded per-target dependency checks to surface configuration issues earlier.
  - Simplified dependency flag handling and reduced propagated build-property plumbing.

- New Features
  - Default build type set to Release when unspecified.
  - Optional interprocedural optimization (IPO) enabled for Release builds when supported.

---
  - https://app.asana.com/0/0/1212854499949998


components/core/cmake/Options/options.cmake
```diff
@@ -309,171 +191,89 @@ function(validate_clp_s_search_dependencies)
...
-    )
-endfunction()
-
 # Validates that for each target whose `CLP_BUILD_` option is `ON`, the `CLP_BUILD_` options for
-# the target's dependencies are also `ON`; Sets the required `CLP_NEED_` flags for any target that
-# will be built.
-function(validate_and_setup_all_clp_dependency_flags)
+# the target's dependencies are also `ON`.
+function(validate_all_clp_dependency_flags)
     if (CLP_BUILD_EXECUTABLES)
```
Bill-hbrhbr: Update `validate_clp_dependencies_for_target` so that we can remove this `if (CLP_BUILD_EXECUTABLES)`. Ditto for all the following if blocks.

components/core/cmake/find_utils.cmake
```diff
@@ -0,0 +1,137 @@
...
+    FindDynamicLibraryDependencies(sqlite "${sqlite_DYNAMIC_LIBS}")
+endmacro()
+
+# Find and setup ystdlib.
+function(clp_find_ystdlib)
+    # We can not call `add_subdirectory` for the same directory multiple times.
+    get_clp_checked_find(ystdlib)
+    if (CLP_CHECKED_FIND)
+        return()
+    endif()
```
Bill-hbrhbr: guard the other macros with `get_clp_checked_find` and `set_clp_checked_find` as well to remove duplicate library `find_package()` messages.
I guess you need `if(NOT CLP_CHECKED_FIND)` since macros don't have `return()`

As a side note, I've wondered if using the `<lib_FOUND>` flag can achieve the same header guarding purpose, e.g. guard `clp_find_boost` with `if(NOT Boost_FOUND)`. However, in my own local experiments, I still see duplicate `find_package()` messages. Perhaps there are race conditions during the cmake generation phase.

components/core/cmake/find_utils.cmake
```diff
@@ -0,0 +1,137 @@
...
+    set(sqlite_DYNAMIC_LIBS "dl;m;pthread")
+    include("${PROJECT_SOURCE_DIR}/cmake/Modules/FindLibraryDependencies.cmake")
+    FindDynamicLibraryDependencies(sqlite "${sqlite_DYNAMIC_LIBS}")
+endmacro()
+
+# Find and setup ystdlib.
+function(clp_find_ystdlib)
+    # We can not call `add_subdirectory` for the same directory multiple times.
+    get_clp_checked_find(ystdlib)
+    if (CLP_CHECKED_FIND)
```
Bill-hbrhbr:
```suggestion
    if(CLP_CHECKED_FIND)
```
Remove space between if and condition

components/core/src/clp/clg/CMakeLists.txt
```diff
@@ -130,6 +130,20 @@ set(
 )
 
 if(CLP_BUILD_EXECUTABLES)
+        find_package(absl REQUIRED)
```
Bill-hbrhbr: We should change all `find_package` to `clp_find_X` for consistency and ease of maintainence.

components/core/src/clp/clg/CMakeLists.txt
```diff
@@ -130,6 +130,19 @@ set(
...
+        clp_find_boost()
+        find_package(date REQUIRED)
+        find_package(fmt REQUIRED)
+        find_package(log_surgeon REQUIRED)
+        clp_find_mariadb_client()
+        find_package(nlohmann_json REQUIRED)
+        find_package(spdlog REQUIRED)
+        clp_find_sqlite()
+        find_package(yaml-cpp REQUIRED)
+        clp_find_ystdlib()
```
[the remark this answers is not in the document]
Bill-hbrhbr: This seems to be a valid suggestion.

components/core/src/clp_s/CMakeLists.txt
```diff
@@ -103,6 +103,17 @@ set(
...
+        clp_find_curl()
+        find_package(fmt REQUIRED)
+        find_package(log_surgeon REQUIRED)
+        clp_find_msgpack()
+        find_package(nlohmann_json REQUIRED)
+        find_package(OpenSSL REQUIRED)
+        find_package(spdlog REQUIRED)
+        clp_find_ystdlib()
+        clp_find_zstd()
+
```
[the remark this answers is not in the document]
Bill-hbrhbr: @coderabbitai Do not raise suggestions to specify CONFIG or MODULE in `find_package`

components/core/src/clp/clp/CMakeLists.txt
```diff
@@ -168,6 +168,20 @@
...
+        find_package(date REQUIRED)
+        find_package(fmt REQUIRED)
+        find_package(log_surgeon REQUIRED)
+        find_package(spdlog REQUIRED)
+        clp_find_sqlite()
+        clp_find_libarchive()
+        clp_find_mariadb_client()
+        find_package(nlohmann_json REQUIRED)
+        find_package(yaml-cpp REQUIRED)
+        clp_find_ystdlib()
```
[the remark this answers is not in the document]
Bill-hbrhbr: ditto

Bill-hbrhbr (changes requested):

components/core/src/clp/clg/CMakeLists.txt
gibber9809: Ah, I was originally doing things that way, but @kirkrodrigues preferred not having macros wrapping the straightforward `find_package` calls, so I switched to this approach.

kirkrodrigues: If you both feel that's clearer, we could. I just felt like a 3-line macro for each of these was adding a lot more code.

components/core/cmake/find_utils.cmake
gibber9809: I don't think we can guard the entirety of the macros (at least for most of them) since generally `find_package` just makes things available in the current scope, so if we do a build where e.g. we disable unit testing then have header guards set up s.t. we only call `find_package` one time for each dependency they'll end up being made available in one subdirectory each and nowhere else.

I can try making warnings emit only one time each, though tbh it seems like the only thing emitting warnings is the version range on boost (which happens inside the `find_package` call. Maybe I can just fix our `find_package` call and version checking for boost? After that we should be in a state where developers only see each warning once (since we do have a header guard for each time we emit warnings, and we will no longer have warnings inside any `find_package` calls).

components/core/src/clp/clg/CMakeLists.txt
Bill-hbrhbr: I see the reasoning. We could leave `find_package` calls without version requirements _macro-less_.

That said, if we eventually need to enforce a version requirement for one of the libraries, having already replaced all `find_package(X)` calls with `clp_find_X()` means it would just be a one-line change, rather than a grep and replace across all CMakeLists.

gibber9809: Yeah I think I agree with Bill on this one, it isn't _that_ much more code and it will make maintenance a bit easier, so we might as well use macros for everything.

components/core/cmake/find_utils.cmake
gibber9809: I ended up doing this + changing the `clp_find_boost` macro s.t. we silence the output from inside of the boost `find_package` call after the first invocation. Note that you'll still see logs for boost being found twice, because we also find boost in `tools/yscope-dev-utils/taskfiles/boost/CMakeLists.txt`.

components/core/cmake/Options/options.cmake
```diff
@@ -309,171 +196,48 @@ function(validate_clp_s_search_dependencies)
...
-    endif()
-
-    if (CLP_BUILD_TESTING)
-        validate_clp_tests_dependencies()
-        set_clp_tests_dependencies()
-    endif()
-
-    if (CLP_BUILD_CLP_REGEX_UTILS)
-        validate_clp_regex_utils_dependencies()
-        set_clp_regex_utils_dependencies()
```
Bill-hbrhbr: think you forgot to remove this function `set_clp_regex_utils_dependencies`

components/core/cmake/find_utils.cmake
```diff
@@ -0,0 +1,213 @@
...
+# Sets a flag indicating that we have run the find function for LIBRARY.
+# @param {string} LIBRARY
+function(set_clp_checked_find LIBRARY)
+    set_property(DIRECTORY "${PROJECT_SOURCE_DIR}" PROPERTY "clp_checked_find_${LIBRARY}" TRUE)
+endfunction()
+
+# Finds and sets up the the absl library.
+# @return Forwards any variables from the `find_package` call.
+macro(clp_find_absl)
+    find_package(absl REQUIRED)
```
Bill-hbrhbr, line 23: I feel like adding back the `message("{<lib>_VERSION}")` or  `message("{<lib>_VERSION_STRING}")` is helpful. Especially as a person who's been working on deps a lot, it assures me when I see the version used is the version installed.
That being said, did you investigate the overly verbose messages when `find_package()` is called multiple times for libraries that emit CMake warnings?
Lastly, if we add the version logging, we would need to header guard all macros.

(Let me know if you and Kirk have discussed this point and reached a different conclusion about logging version strings)

Bill-hbrhbr (changes requested):

components/core/cmake/find_utils.cmake
gibber9809: Yeah like I mention in my other comment I:
1. Fixed the warning caused by our incorrect usage of `find_package` for boost
2. Made the `find_package` call `QUIET` for boost after the first invocation since boost itself logs the found version within `FindBoost`

For logging the version of each library I see where you're coming from, but I feel like if we have a specific version requirement we should just specify it in the `find_package` call, and if it isn't satisfied the user will get a message saying that we were unable to find the correct version anyway.

If you really feel like its useful as a maintainer of the cmake scripts to log the found versions anyway though I'd prefer splitting that change off into a smaller follow-up PR, since this one is already pretty big.

Bill-hbrhbr (dismissed):
passing off to Kirk

components/core/cmake/Options/options.cmake
```diff
@@ -312,171 +192,48 @@ function(validate_clp_s_search_dependencies)
...
-    endif()
-
+# the target's dependencies are also `ON`.
+function(validate_all_clp_dependency_flags)
+    validate_clp_binaries_dependencies()
+    validate_clp_tests_dependencies()
+    validate_clp_regex_utils_dependencies()
     # clp::string_utils has no dependencies
```
Bill-hbrhbr, line 227:
```suggestion
```

taskfiles/deps/main.yaml
```diff
@@ -600,8 +600,8 @@ tasks:
             - "-DCMAKE_POSITION_INDEPENDENT_CODE=ON"
             - "-Dystdlib_BUILD_TESTING=OFF"
           LIB_NAME: "ystdlib"
-          TARBALL_SHA256: "65990dc2bcc4a355c2181bfe31a7800f492309d1bcd340f52a34e85047e61bc8"
-          TARBALL_URL: "https://github.com/y-scope/ystdlib-cpp/archive/9ed78cd.tar.gz"
+          TARBALL_SHA256: "379c04c86715f39cffa09bb3ebc22b4c45af9a63ea3efa2081d210289fcd2af6"
+          TARBALL_URL: "https://github.com/y-scope/ystdlib-cpp/archive/c03806a.tar.gz"
```
Bill-hbrhbr, line 604: Need to update this commit version in
https://github.com/y-scope/clp/blob/main/docs/src/dev-docs/components-core/index.md

Bill-hbrhbr (approved):
nit picks

components/core/cmake/Options/options.cmake
```diff
@@ -312,171 +192,48 @@ function(validate_clp_s_search_dependencies)
...
-    endif()
-
-    if (CLP_BUILD_CLP_REGEX_UTILS)
-        validate_clp_regex_utils_dependencies()
-        set_clp_regex_utils_dependencies()
-    endif()
-
+# the target's dependencies are also `ON`.
+function(validate_all_clp_dependency_flags)
+    validate_clp_binaries_dependencies()
```
Bill-hbrhbr, line 223: technically these validate functions are just wrapper around other singular function calls. but yeah IMO keeping the wrapper doesn't harm

Bill-hbrhbr:
For a draft title, how about:
```
refactor(build): Replace `CLP_NEEDS_X` flags with `clp_find_X` calls for dependency marking; bump `ystdlib` to y-scope/ystdlib-cpp@c03806a to remove `find_package` include guard.
```

PR #1108, opened 2025-07-16
feat(webui): Integrate Monaco editor for future Presto SQL query input.

davemarco:
# Description

PR adds the monaco editor to the webui for SQL input, and adds support for highlighting and autocomplete for sql syntax like `SELECT`, `WHERE` etc. 

I also added a "run" button, which seemed more appropriate than "search" since we support more queries with sql. 

They only render when presto is set as query engine in settings.

# Checklist

  breaking change.

# Validation performed

Editor loads. To validate yourself, you must change   from ClpQueryEngine from native to presto in settings.json in public folder.

## Summary by CodeRabbit

* **New Features**
  * Introduced a SQL editor with syntax highlighting and autocomplete for SQL queries.
  * Added a dedicated SQL query input and "Run" button for Presto query engine workflows.
  * UI now dynamically switches between native and Presto query controls based on configuration.

* **Enhancements**
  * Added configuration options to select the query engine type in settings.
  * Improved editor responsiveness and user experience with dynamic sizing and tooltips.

* **Chores**
  * Updated dependencies to support Monaco Editor integration.
  * Adjusted build configuration for optimized Monaco Editor loading.

### DIFF
--- a/components/core/CMakeLists.txt
+++ b/components/core/CMakeLists.txt
@@ -21,6 +21,15 @@ if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     set(CMAKE_BUILD_TYPE "${default_build_type}" CACHE STRING "Choose the type of build." FORCE)
 endif()
 
+if (CMAKE_BUILD_TYPE MATCHES "Release")
+    include(CheckIPOSupported)
+    check_ipo_supported(RESULT IPO_SUPPORTED OUTPUT)
+    if(IPO_SUPPORTED)
+        message(STATUS "Link-time optimization enabled.")
+        set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)
+    endif()
+endif()
+
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 # Set general compressor


components/webui/client/src/components/SqlEditor/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+type SqlEditorProps = Omit<EditorProps, "language">;
+
+/**
+ * Monaco editor with highlighting and autocomplete for SQL syntax.
+ *
+ * @param props
+ * @return
+ */
+const SqlEditor = (props: SqlEditorProps) => {
```
hoophalab, line 28: This isn't your issue, but when we resize the browser window's width below a threshold. The Monaco editor doesn't seem to update its width correctly, which can cause the "Run" button to get clipped.

image

components/webui/client/src/pages/SearchPage/SearchControls/Presto/RunButton/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Renders a button to run the SQL query.
+ *
+ * @return
+ */
+const RunButton = () => {
+    const queryString = useSearchStore((state) => state.queryString);
+
+    const isQueryStringEmpty = queryString === SEARCH_STATE_DEFAULT.queryString;
```
hoophalab: `isQueryStringEmpty` is confusing.
```suggestion
    const isQueryStringEmpty = queryString === "";
```

I find it counter-intuitive to check for equality against a default value, especially if we set default to something frequently used later (e.g., `SEARCH_STATE_DEFAULT.queryString === "SELECT * FROM default"`).

components/webui/client/src/pages/SearchPage/SearchControls/Presto/RunButton/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+    let tooltipTitle = "";
+    if (isQueryStringEmpty) {
+        tooltipTitle = "Enter SQL query to run";
+    }
```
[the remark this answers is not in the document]
hoophalab: I agree with coderabbitai

```suggestion
    const isQueryStringEmpty = queryString === SEARCH_STATE_DEFAULT.queryString;
    const tooltipTitle = isQueryStringEmpty ?  "Enter SQL query to run" : "";
```

components/webui/client/src/pages/SearchPage/SearchControls/Presto/SqlQueryInput/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+
+/**
+ * Renders SQL query input.
+ *
+ * @return
+ */
+const SqlQueryInput = () => {
+    const handleChange = useCallback((value: string | undefined) => {
+        const {updateQueryString} = useSearchStore.getState();
```
hoophalab, line 15:
```suggestion
        const updateQueryString = useSearchStore((state) => state.updateQueryString);
```
simplify the state dependency

components/webui/client/src/pages/SearchPage/SearchControls/Presto/RunButton/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+    return (
+        <Tooltip title={tooltipTitle}>
+            <Button
+                className={styles["gradientButton"] || ""}
```
hoophalab: Let's move the gradientButton's CSS file to a shared folder, maybe in `src/components`. We shouldn't have one component's styles depends on another. Otherwise, if someone updates the submit button later, they'll unintentionally change the run button's appearance too.

components/webui/client/src/pages/SearchPage/SearchControls/Presto/RunButton/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+    const isQueryStringEmpty = queryString === SEARCH_STATE_DEFAULT.queryString;
+
+    let tooltipTitle = "";
+    if (isQueryStringEmpty) {
+        tooltipTitle = "Enter SQL query to run";
+    }
+
+    return (
+        <Tooltip title={tooltipTitle}>
```
hoophalab, line 24: For a consistent look, we should disable this button when the querystring is empty, just like we did with the submit button.

components/webui/client/src/pages/SearchPage/SearchControls/index.tsx
```diff
@@ -27,10 +31,21 @@ const SearchControls = () => {
...
+                            {CLP_STORAGE_ENGINES.CLP_S === SETTINGS_STORAGE_ENGINE && <Dataset/>}
+                            <QueryInput/>
+                            <TimeRangeInput/>
+                            <SearchButton/>
+                        </>
+                    ) :
+                    (
+                        <>
+                            <SqlQueryInput/>
+                            <RunButton/>
```
hoophalab, line 46: Besides the Monaco width issue, what do you think about moving the run button below the SQL editor and aligning it to the right?

image

components/webui/client/src/pages/SearchPage/SearchControls/Presto/RunButton/index.tsx
hoophalab: Or, ideally, we create a reusable GradientButton component and let both submit button and run button use GradientButton.

components/webui/client/src/components/SqlEditor/index.tsx
davemarco: I found a fix for this with css. so now fixed

components/webui/client/src/pages/SearchPage/SearchControls/Presto/RunButton/index.tsx
davemarco: sure i changed this here and the other submit button

components/webui/client/src/pages/SearchPage/SearchControls/Presto/RunButton/index.tsx
davemarco: Instead of adding overhead of a custom component, i just made this button solid green.

components/webui/client/src/pages/SearchPage/SearchControls/Presto/SqlQueryInput/index.tsx
davemarco: junhao decided for these callbacks we should use the method i used. If you use getState, the output is not reactive, then you dont need the state as a dependency in the hook. like you dont need [updateQueryString] like u would without get state

components/webui/client/src/pages/SearchPage/SearchControls/index.tsx
davemarco: I thought about it at first, but it will complicate searchQueryStatus. Like right now searchQueryStatus is a separate component under SearchControls. I think it looks weird to have searchQueryStatus under the button. I though about putting searchQueryStatus next to the button, but it will just be mess to render searchQueryStatus in different locations for "presto" and "native". I thought just putting the button on the right makes "native" and "presto" have similiar layout simplifying the codebase

davemarco (commented):
see comments

components/webui/client/src/pages/SearchPage/SearchControls/Presto/SqlQueryInput/index.tsx
hoophalab: If that's the case, we should probably plan on unifying this coding style. For example

https://github.com/y-scope/clp/blob/8fc0dd710bf61cd76720542aa3c5b7e19f7112df/components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/index.tsx#L31


https://github.com/y-scope/clp/blob/8fc0dd710bf61cd76720542aa3c5b7e19f7112df/components/webui/client/src/pages/SearchPage/SearchControls/Dataset/index.tsx#L23

There might be other instances like these too.

Can you confirm these lines need to be refactored? If yes, let's ask coderabbit to create an issue on this
@davemarco

components/webui/client/package.json
```diff
@@ -16,6 +16,7 @@
     "@ant-design/v5-patch-for-react-19": "^1.0.3",
     "@emotion/react": "^11.14.0",
     "@emotion/styled": "^11.14.0",
+    "@monaco-editor/react": "^4.7.0",
```
junhaoliao, line 19: What's the bundle size after adding Monaco?

One reason we avoided using this `@monaco-editor/react` was that the bundle size increases significantly as all Monaco assets were packed into the bundle. Is it still the case today?

davemarco: is seems reasonable here? https://bundlephobia.com/package/@monaco-editor/react@4.7.0

components/webui/client/src/pages/SearchPage/SearchControls/Presto/SqlQueryInput/index.tsx
hoophalab: @coderabbitai can you create an issue based on the discussion?

components/webui/client/package.json
junhaoliao: i ran `npm run build` and this is the output:
```
> webui-client@0.1.0 build
> tsc -b && vite build

vite v6.2.6 building for production...
✓ 5687 modules transformed.
dist/index.html                     0.77 kB │ gzip:   0.41 kB
dist/assets/index-UZaMzv9P.css      2.73 kB │ gzip:   1.06 kB
dist/assets/index-CaK4Bu86.js   2,613.68 kB │ gzip: 838.16 kB

(!) Some chunks are larger than 500 kB after minification. Consider:
- Using dynamic import() to code-split the application
- Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/configuration-options/#output-manualchunks
- Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
✓ built in 6.88s
```

previously, it was
```
vite v6.2.6 building for production...
✓ 5670 modules transformed.
dist/index.html                     0.77 kB │ gzip:   0.41 kB
dist/assets/index-CG-hk4vl.css      2.69 kB │ gzip:   1.04 kB
dist/assets/index-BdLJKTKx.js   2,593.66 kB │ gzip: 830.59 kB
```

so we can see the package size increase is not significant possibly because only the `@monaco-editor/react` are included. however, i wonder where the `monaco-editor` assets are added

junhaoliao: by default, the loader-config loads Monaco assets from a CDN (on my end it uses https://cdn.jsdelivr.net/npm/monaco-editor): https://github.com/suren-atoyan/monaco-react/blob/master/README.md#loader-config 

This would mean that if a user accesses the webui without access to the Internet or behind a country / corporate managed firewall, they would not be able to use the search page.

can we follow https://github.com/suren-atoyan/monaco-react/blob/master/README.md#use-monaco-editor-as-an-npm-package to config the monaco assets imports? (there's a Vite specific section)

junhaoliao: on my end, if i just use their Vite example loader code, the bundle size bumped from 2.5MB to 15MB.

i think the issue is with
```
import * as monaco from 'monaco-editor';
```

if we use https://github.com/y-scope/yscope-log-viewer/blob/main/src/components/Editor/MonacoInstance/bootstrap.ts as a reference and import only what we need, the bundle size increase should be less than 972.2KB

image

components/webui/client/src/components/SqlEditor/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+        };
+    }, [monaco]);
+
+    return (
+        <Editor
+            language={"sql"}
+
+            // Use white background while loading (default is grey) so transition to editor with
+            // white background is less jarring.
+            loading={<div style={{backgroundColor: "white", height: "100%", width: "100%"}}/>}
```
junhaoliao, line 123: (just an idea - not strictly needed in this PR)
Would https://ant.design/components/skeleton#skeleton-demo-active be a better alternative?

junhaoliao (commented):
i tested the UI and it looks very cool

i'll approve once https://github.com/y-scope/clp/pull/1108#discussion_r2216851567 is addressed

components/webui/client/src/components/SqlEditor/index.tsx
davemarco: i tried it already! it looks worse since the grey lines in the skeleton appear and disapear. the plain white box just looks better

components/webui/client/package.json
davemarco: actually never mind. we do need the whole bundle for the autocomplete to work. If you only want the editor, then we can't have autocomplete i think

davemarco: >  import only what we need, the bundle size increase should be less than 972.2KB 

this wont work since it will break the autocomplete, unless you find another thing to import that can fix. I was able to fix syntax highlighting but not autocomplete

davemarco: adding this may fix, but not sure why import "monaco-editor/esm/vs/editor/editor.all.js"

components/webui/client/src/components/SqlEditor/monaco-config.ts
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable import/default */
+import {loader} from "@monaco-editor/react";
+import * as monaco from "monaco-editor/esm/vs/editor/editor.api";
+import EditorWorker from "monaco-editor/esm/vs/editor/editor.worker?worker";
+
+import "monaco-editor/esm/vs/basic-languages/sql/sql.contribution.js";
+import "monaco-editor/esm/vs/editor/editor.all.js";
+
```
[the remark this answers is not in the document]
davemarco: the autocomplete dosent work without it

components/webui/client/src/components/SqlEditor/monaco-config.ts
junhaoliao, line 1: Can we rename this as `monaco-loader.ts` or `monaco-boostrap.ts`? The name "config" can be a bit confusing since we're loading assets / configuring the Monaco React loader rather than doing Monaco configurations here.

junhaoliao (changes requested):

components/webui/client/src/components/SqlEditor/monaco-config.ts
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable import/default */
+import {loader} from "@monaco-editor/react";
+import * as monaco from "monaco-editor/esm/vs/editor/editor.api";
+import EditorWorker from "monaco-editor/esm/vs/editor/editor.worker?worker";
+
+import "monaco-editor/esm/vs/basic-languages/sql/sql.contribution.js";
+import "monaco-editor/esm/vs/editor/editor.all.js";
```
junhaoliao: we can import only what we need, which saves another half MB:

```suggestion
/* eslint-disable import/default, @stylistic/max-len */

import {loader} from "@monaco-editor/react";
import * as monaco from "monaco-editor/esm/vs/editor/editor.api";
import EditorWorker from "monaco-editor/esm/vs/editor/editor.worker?worker";

import "monaco-editor/esm/vs/basic-languages/sql/sql.contribution.js";
import "monaco-editor/esm/vs/editor/contrib/clipboard/browser/clipboard.js";
import "monaco-editor/esm/vs/editor/contrib/contextmenu/browser/contextmenu.js";
import "monaco-editor/esm/vs/editor/contrib/find/browser/findController.js";
import "monaco-editor/esm/vs/editor/contrib/wordHighlighter/browser/wordHighlighter.js";
import "monaco-editor/esm/vs/editor/contrib/suggest/browser/suggestController.js";
import "monaco-editor/esm/vs/editor/contrib/placeholderText/browser/placeholderText.contribution.js";


/* eslint-enable @stylistic/max-len */
```

junhaoliao: i've found ways to lazy load the monaco assets but the code is not very organized at this point. i'll investigate further in the future

@coderabbitai create an issue to support lazy loading of the monaco assets

components/webui/client/src/components/SqlEditor/monaco-config.ts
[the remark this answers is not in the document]
junhaoliao: see above comment

components/webui/client/package.json
junhaoliao: I now understand this 972.2KB is counting sizes of assets only in the "min" bundle and does not include any ESM APIs. Since we use the ESM APIs, we do need to import differently (more).

Let's see the comment at "monaco-config.ts" to reduce the scope of import. In the future, we should support lazy loading those.

junhaoliao (changes requested):
added some comments about loading monaco assets

components/webui/client/src/pages/SearchPage/SearchControls/Presto/SqlQueryInput/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+        updateQueryString(value || "");
+    }, []);
+
+    return (
+        <div className={styles["input"] || ""}>
+            <SqlEditor
+                height={"150px"}
```
junhaoliao: i find it kinda awkward to show a giant box if the user hasn't input any query / the query is just a single line. Let's see if we can get the editor to auto resize base on the content
```suggestion
```

components/webui/client/src/components/SqlEditor/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Monaco editor with highlighting and autocomplete for SQL syntax.
+ *
+ * @param props
+ * @return
+ */
+const SqlEditor = (props: SqlEditorProps) => {
+    const monaco = useMonaco();
+
+
```
junhaoliao, line 30: redundant empty line

components/webui/client/src/components/SqlEditor/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Monaco editor with highlighting and autocomplete for SQL syntax.
+ *
+ * @param props
+ * @return
+ */
+const SqlEditor = (props: SqlEditorProps) => {
+    const monaco = useMonaco();
+
```
junhaoliao:
```suggestion
    const [isContentMultiline, setIsContentMultiline] = useState<boolean>(false);

    const handleMonacoMount = useCallback((editor: editor.IStandaloneCodeEditor) => {
        editor.onDidContentSizeChange((ev) => {
            if (false === ev.contentHeightChanged) {
                return;
            }
            if (null === monaco) {
                throw new Error("Unexpected null Monaco instance");
            }

            const domNode = editor.getDomNode();
            if (null === domNode) {
                throw new Error("Unexpected null editor DOM node");
            }

            const model = editor.getModel();
            if (null === model) {
                throw new Error("Unexpected null editor model");
            }

            const lineHeight = editor.getOption(monaco.editor.EditorOption.lineHeight);
            const contentHeight = editor.getContentHeight();
            const approxWrappedLines = Math.round(contentHeight / lineHeight);

            setIsContentMultiline(1 < approxWrappedLines);

            if (5 >= approxWrappedLines) {
                domNode.style.height = `${contentHeight}px`;
            } else {
                domNode.style.height = `${lineHeight * 5}px`;
            }
        });
    }, [monaco]);
```

components/webui/client/src/components/SqlEditor/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+    }, [monaco]);
+
+    return (
+        <Editor
+            language={"sql"}
+
+            // Use white background while loading (default is grey) so transition to editor with
+            // white background is less jarring.
+            loading={<div style={{backgroundColor: "white", height: "100%", width: "100%"}}/>}
+            options={{
```
junhaoliao, line 124:
```suggestion
            options={{
                automaticLayout: true,
                folding: isContentMultiline,
```

components/webui/client/src/components/SqlEditor/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+                fontSize: 14,
+                lineNumbers: "on",
```
junhaoliao:
```suggestion
                lineNumbers: isContentMultiline ?
                    "on" :
                    "off",
```

components/webui/client/src/components/SqlEditor/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+                lineNumbersMinChars: 2,
+                minimap: {enabled: false},
+                overviewRulerBorder: false,
+                placeholder: "Enter your SQL query",
+                scrollBeyondLastLine: false,
```
junhaoliao, line 137:
```suggestion
                renderLineHighlightOnlyWhenFocus: true,
                scrollBeyondLastLine: false,
```

components/webui/client/src/components/SqlEditor/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+                wordWrap: "on",
+            }}
+            {...props}/>
```
junhaoliao, line 141:
```suggestion
            onMount={handleMonacoMount}
            {...props}/>
```

components/webui/client/src/components/SqlEditor/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+    return (
+        <Editor
+            language={"sql"}
+
+            // Use white background while loading (default is grey) so transition to editor with
+            // white background is less jarring.
+            loading={<div style={{backgroundColor: "white", height: "100%", width: "100%"}}/>}
+            options={{
+                fontSize: 14,
```
junhaoliao: maybe increase the font size as well
```suggestion
                // Avoid 30 to prevent scroll being visible even with just a single line.
                fontSize: 29.9,
```

components/webui/client/src/components/SqlEditor/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+
+    useEffect(() => {
+        if (null === monaco) {
+            return () => {
+            };
+        }
+
+        // Adds autocomplete suggestions for SQL keywords on editor load
+        const provider = monaco.languages.registerCompletionItemProvider("sql", {
```
junhaoliao: for lazy users who don't even wanna type a single character
```suggestion
        const provider = monaco.languages.registerCompletionItemProvider("sql", {
            triggerCharacters: [
                " ",
                "\n",
            ],
```

components/webui/client/src/components/SqlEditor/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+            provideCompletionItems: (model, position) => {
+                const word = model.getWordUntilPosition(position);
+                const range = {
+                    startLineNumber: position.lineNumber,
+                    endLineNumber: position.lineNumber,
+                    startColumn: word.startColumn,
+                    endColumn: word.endColumn,
+                };
+                const suggestions = sqlLanguage.keywords.map((keyword: string) => ({
+                    detail: "SQL Keyword",
```
junhaoliao: The name "Keyword" can be confusing to end users.

how about just
```suggestion
                    detail: "SQL",
```

in the future we can add UDF as well, so maybe this is better:
```suggestion
                    detail: "Presto SQL (CLP)",
```

junhaoliao (changes requested):

components/webui/client/src/components/SqlEditor/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+                    insertText: `${keyword} `,
+                    kind: monaco.languages.CompletionItemKind.Keyword,
+                    label: keyword,
+                    range: range,
+                }));
+
```
junhaoliao, line 54:
```suggestion

                const typedWord = model.getValueInRange(range);
                if (0 < typedWord.length) {
                    suggestions.push({
                        detail: "Current",
                        insertText: `${typedWord}\n`,
                        kind: monaco.languages.CompletionItemKind.Text,
                        label: typedWord,
                        range: range,
                    });
                }

```

components/webui/client/src/components/SqlEditor/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+                return {suggestions: suggestions};
```
junhaoliao: to make sure `provideCompletionItems()` always get called even if the listed suggestions still can match users' existing inputs.
```suggestion
                return {
                    suggestions: suggestions,
                    incomplete: true,
                };
```

junhaoliao:
@hoophalab all yours again -
please see if the above suggestions make sense. i don't have any more comments for this PR

components/webui/client/src/components/SqlEditor/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+        <Editor
+            language={"sql"}
+
+            // Use white background while loading (default is grey) so transition to editor with
+            // white background is less jarring.
+            loading={<div style={{backgroundColor: "white", height: "100%", width: "100%"}}/>}
+            options={{
+                automaticLayout: true,
+                folding: isContentMultiline,
+                fontSize: 18,
```
junhaoliao: @hoophalab we discussed offline that setting the `fontSize` to 30 could make the UI look awkward. Still, i think it makes sense to let the input box's height to match the Run button's height. how about:
```suggestion
                fontSize: 20,
                lineHeight: 30,
```

components/webui/client/src/components/SqlEditor/index.tsx
junhaoliao: Following https://github.com/y-scope/clp/pull/1108/files#r2229464629 , i think it makes sense to reapply this sugguestion

components/webui/client/src/pages/SearchPage/SearchControls/Presto/SqlQueryInput/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+const SqlQueryInput = () => {
+    const handleChange = useCallback((value: string | undefined) => {
+        const {updateQueryString} = useSearchStore.getState();
+        updateQueryString(value || "");
+    }, []);
+
+    return (
+        <div className={styles["input"] || ""}>
+            <SqlEditor
+                onChange={handleChange}/>
```
junhaoliao:
```suggestion
            <SqlEditor onChange={handleChange}/>
```

junhaoliao (approved):
for the PR title, how about:
```
feat(webui): Integrate Monaco editor for future Presto SQL query input.
```

PR #1114, opened 2025-07-20
fix(core-clp): Construct file paths in test cases relative to test file location rather than working directory (fixes #124).

### LINKED ISSUE #124: Schema unit tests fail if run outside build directory.
### Bug

If the unit tests are run outside the build directory, e.g.:
`./build/unitTest` instead of `cd build && ./unitTest`
you will get 11 failures related to missing schema files.

This is caused by relative paths being used in the schema unit tests.

An error example:
```
/home/lion/yscope/clp-ffi-go/cpp/clp/components/core/tests/test-Grep.cpp:21: FAILED:
due to unexpected exception with message:
ReaderInterface operation failed

[2023-05-20 23:25:59.016] [info] [test-ParserWithUserSchema.cpp:72] File not found: /home/lion/yscope/clp-ffi-go/cpp/clp/components/tests/test_schema_files/missing_schema.txt
``` 

### CLP version

460d377

### Environment

ghcr.io/y-scope/clp/clp-core-dependencies-x86-ubuntu-focal:main

### Reproduction steps

1. build the latest CLP
2. run `unitTest` outside of your build directory (i.e. run `unitTest` from any directory other than the file's parent)

quinntaylormitchell:
# Description

Issue #124 describes how some unit tests fail if you run `unitTest` outside the build directory. The offending test cases are in `test-GrepCore.cpp` and `test-ParserWithUserSchema.cpp`.

This PR adds methods that construct file paths using the `__FILE__` macro from `std::filesystem::path`.

_NOTE: in the future we should deduplicate the code by having some directory called `test-utils` or something where we can store methods like `get_test_dir_local_path()`._

# Checklist

  breaking change.

# Validation performed

`unitTest` has been run from inside and outside the build directory; all test cases pass.

## Summary by CodeRabbit

- **Tests**
  - Replaced hard-coded relative test paths with centralized, dynamically resolved test directories to improve environment-independent reliability.
  - Switched test asset handling to standard filesystem utilities and standardized loading of schema and query files across suites.
  - No changes to product behaviour or user-facing functionality.


quinntaylormitchell:
@coderabbitai review

quinntaylormitchell:
I mention in the Description that we should deduplicate the code that gets local paths and store it in some directory called `test-utils` or something. Would that be appropriate for this PR?

quinntaylormitchell:
@coderabbitai review

components/core/tests/test-ParserWithUserSchema.cpp
```diff
@@ -28,6 +29,16 @@ using log_surgeon::SchemaParser;
...
 
+namespace {
+[[nodiscard]] auto get_test_dir_local_path() -> std::string;
+
+auto get_test_dir_local_path() -> std::string {
+    std::filesystem::path const current_file_path{__FILE__};
+    auto const tests_dir{current_file_path.parent_path()};
+    return tests_dir.string();
+}
+}  // namespace
```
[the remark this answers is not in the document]
quinntaylormitchell: waiting on advice from colleagues on whether or not it is appropriate to do that in this PR, or whether that should be saved for another PR

components/core/tests/test-ParserWithUserSchema.cpp
```diff
@@ -183,10 +201,11 @@
 
 TEST_CASE("Test reverse lexer", "[Search]") {
     ByteLexer reverse_lexer;
-    std::string schema_file_name = "../tests/test_schema_files/search_schema.txt";
+    std::string schema_file_name
+            = get_test_dir_local_path() + "/test_schema_files/search_schema.txt";
     std::string schema_file_path = boost::filesystem::weakly_canonical(schema_file_name).string();
     load_lexer_from_file(schema_file_path, false, reverse_lexer);
-    FileReader file_reader{"../tests/test_search_queries/easy.txt"};
+    FileReader file_reader{get_test_dir_local_path() + "/test_search_queries/easy.txt"};
```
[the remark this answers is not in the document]
quinntaylormitchell: You are correct, but not sure if this is within scope of this PR.

components/core/tests/test-ParserWithUserSchema.cpp
```diff
@@ -80,8 +91,8 @@
 }
 
 TEST_CASE("Test error for missing schema file", "[LALR1Parser][SchemaParser]") {
-    std::string file_path = "../tests/test_schema_files/missing_schema.txt";
-    std::string file_name = boost::filesystem::weakly_canonical(file_path).string();
+    std::string file_path = get_test_dir_local_path() + "/test_schema_files/missing_schema.txt";
+    // std::string file_name = boost::filesystem::weakly_canonical(file_path).string();
     REQUIRE_THROWS_WITH(
```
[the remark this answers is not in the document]
quinntaylormitchell: why do we prefer this?

quinntaylormitchell:
@coderabbitai review

components/core/tests/test-ParserWithUserSchema.cpp
```diff
@@ -159,10 +173,11 @@ TEST_CASE("Test creating log parser without delimiters", "[LALR1Parser][LogParse
 
 TEST_CASE("Test forward lexer", "[Search]") {
     ByteLexer forward_lexer;
-    std::string schema_file_name = "../tests/test_schema_files/search_schema.txt";
+    std::string schema_file_name
+            = (get_test_dir() / "test_schema_files/search_schema.txt").string();
     std::string schema_file_path = boost::filesystem::weakly_canonical(schema_file_name).string();
     load_lexer_from_file(schema_file_path, false, forward_lexer);
-    FileReader file_reader{"../tests/test_search_queries/easy.txt"};
+    FileReader file_reader{(get_test_dir() / "test_search_queries/easy.txt").string()};
```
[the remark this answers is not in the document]
quinntaylormitchell: outside of the scope of this PR

components/core/tests/test-ParserWithUserSchema.cpp
```diff
@@ -79,8 +89,7 @@ void decompress(std::string archive_dir, std::string output_dir) {
 }
 
 TEST_CASE("Test error for missing schema file", "[LALR1Parser][SchemaParser]") {
-    std::string file_path = "../tests/test_schema_files/missing_schema.txt";
-    std::string file_name = boost::filesystem::weakly_canonical(file_path).string();
+    std::string file_path = (get_test_dir() / "test_schema_files/missing_schema.txt").string();
```
haiqi96: nit:
1. We can store the value of `get_test_dir()` to a variable before all TEST_CASEs so that we can reuse the variable instead of recalling the method multiple times. Since `test_schema_files` is shared by all test_files, we can put it in the shared variable as well, maybe name it as `test_schema_files_dir`.
2: we can use auto const for file_path, since `.string()` has an explicit return type.
3. we can delay calling string() till when we actually need to. 

Combine it all together, 

```
auto const test_schema_file_path{test_schema_files_dir / "empty_schema.txt"};
REQUIRE_THROWS_WITH(
            generate_schema_ast(test_schema_file_path.string()),
            "Schema:1:1: error: empty file\n"
)
```



Ditto for all other occurence. Let's always declare a test_schema_path variable, instead of directly putting a long line (such as `(get_test_dir() / "test_schema_files/search_schema.txt").string()`) into the function parameter.

components/core/tests/test-GrepCore.cpp
```diff
@@ -17,9 +18,18 @@ using log_surgeon::SchemaParser;
...
+auto get_test_dir() -> std::filesystem::path {
+    std::filesystem::path const current_file_path{__FILE__};
+    return current_file_path.parent_path();
+}
+}  // namespace
+
 TEST_CASE("get_bounds_of_next_potential_var", "[get_bounds_of_next_potential_var]") {
     ByteLexer lexer;
-    load_lexer_from_file("../tests/test_schema_files/search_schema.txt", lexer);
+    load_lexer_from_file((get_test_dir() / "test_schema_files/search_schema.txt").string(), lexer);
```
haiqi96: please make sure you test and run through linter.
```suggestion
    auto const test_schema_file_path{get_test_dir() / "test_schema_files" / "search_schema.txt"}
    load_lexer_from_file(test_schema_file_path.string(), lexer);
```

components/core/tests/test-ParserWithUserSchema.cpp
quinntaylormitchell: agreed, will do.

components/core/tests/test-ParserWithUserSchema.cpp
```diff
@@ -79,56 +93,55 @@ void decompress(std::string archive_dir, std::string output_dir) {
...
+            generate_schema_ast(file_path.string()),
             "Schema:2:11: error: expected ':' before ' ' token\n"
             "          delimiters : \\r\\n\n"
             "                    ^\n"
     );
 }
 
 TEST_CASE("Test creating schema parser", "[LALR1Parser][SchemaParser]") {
-    generate_schema_ast("../tests/test_schema_files/easy_schema.txt");
+    generate_schema_ast((test_schema_dir / "easy_schema.txt").string());
```
haiqi96: Let's break the line into two lines for better readability. Ditto for other lines below
```suggestion
    auto const schema_file_path{test_schema_dir / "easy_schema.txt"};
    generate_schema_ast(schema_file_path.string());
```

components/core/tests/test-ParserWithUserSchema.cpp
```diff
@@ -158,10 +171,10 @@ TEST_CASE("Test creating log parser without delimiters", "[LALR1Parser][LogParse
 
 TEST_CASE("Test lexer", "[Search]") {
     ByteLexer lexer;
-    std::string schema_file_name = "../tests/test_schema_files/search_schema.txt";
-    std::string schema_file_path = boost::filesystem::weakly_canonical(schema_file_name).string();
+    auto const schema_file_name = (test_schema_dir / "search_schema.txt").string();
+    auto const schema_file_path = boost::filesystem::weakly_canonical(schema_file_name).string();
     load_lexer_from_file(schema_file_path, lexer);
-    FileReader file_reader{"../tests/test_search_queries/easy.txt"};
+    FileReader file_reader{(test_query_dir / "easy.txt").string()};
```
haiqi96: ditto for 1 line -> 2 lines

components/core/tests/test-ParserWithUserSchema.cpp
```diff
@@ -158,10 +171,10 @@ TEST_CASE("Test creating log parser without delimiters", "[LALR1Parser][LogParse
 
 TEST_CASE("Test lexer", "[Search]") {
     ByteLexer lexer;
-    std::string schema_file_name = "../tests/test_schema_files/search_schema.txt";
-    std::string schema_file_path = boost::filesystem::weakly_canonical(schema_file_name).string();
+    auto const schema_file_name = (test_schema_dir / "search_schema.txt").string();
+    auto const schema_file_path = boost::filesystem::weakly_canonical(schema_file_name).string();
```
haiqi96: I know you didn't write this code, but it's weird that 
1. schema_file_name is only used once.
2. filesystem::weakly_canonical should allow a `path` type argument, so I am not sure why we first convert schema_file_name to string.
3. it's not clear to me why we use `weakly_canonical` here, whereas we don't use `weakly_canonical` in other tests.

Another nit: I believe we have been trying to remove the dependency on boost::filesystem. Since you include <filesystem>, can you try replacing boost::filesystem with std::filesystem?

components/core/tests/test-ParserWithUserSchema.cpp
```diff
@@ -79,56 +93,55 @@ void decompress(std::string archive_dir, std::string output_dir) {
 }
 
 TEST_CASE("Test error for missing schema file", "[LALR1Parser][SchemaParser]") {
-    std::string file_path = "../tests/test_schema_files/missing_schema.txt";
-    std::string file_name = boost::filesystem::weakly_canonical(file_path).string();
+    auto const file_path = test_schema_dir / "missing_schema.txt";
     REQUIRE_THROWS_WITH(
-            generate_schema_ast(file_path),
-            "Failed to read '" + file_path + "', error_code="
+            generate_schema_ast(file_path.string()),
```
haiqi96: sorry for my oversight. since we need to use file_path.string() twice for this test_case, let's store it as a separate variable
`auto const file_path_string = file_path.string()`

components/core/tests/test-ParserWithUserSchema.cpp
quinntaylormitchell: >weird

not sure!

>replacing boost::filesystem with std::filesystem

done.

haiqi96: you haven't addressed the comment on weakly_canonical, right?
I am hoping that we can rewrite the code like 
```
auto const schema_file_path = test_schema_dir / "search_schema.txt;
```
and use the path as string() where appropriate.

Looking at the doc of https://en.cppreference.com/w/cpp/filesystem/canonical.html. it seems we want to make sure there's no dot in the path. 
This leads to one question, is it possible for the return value of `get_test_dir()` (or to be more specific, `std::filesystem::path const current_file_path{__FILE__}`) to contain any dot or dot-dot element?

Quickly looking at __FILE__, it seems there's no guarantee it would not contain dot or dot-dot.

So I think what we can make everything simple, convert the return value of get_test_dir() into absolute path by using std::filesystem::canonical(). Imo, it is also safer to return the absolute path.

components/core/tests/test-GrepCore.cpp
```diff
@@ -17,9 +18,22 @@ using log_surgeon::SchemaParser;
...
+namespace {
+[[nodiscard]] auto get_test_dir() -> std::filesystem::path;
+
+auto get_test_dir() -> std::filesystem::path {
+    std::filesystem::path const current_file_path{__FILE__};
+    return current_file_path.parent_path();
+}
+
+auto const test_dir = get_test_dir();
+auto const test_schema_dir = test_dir / "test_schema_files";
```
kirkrodrigues: clang-tidy complains about these lines:

> Clang-Tidy: Initialization of 'test_dir' with static storage duration may throw an exception that cannot be caught.

Can we move them into the actual test case?

components/core/tests/test-ParserWithUserSchema.cpp
```diff
@@ -28,6 +28,19 @@ using log_surgeon::SchemaAST;
 using log_surgeon::SchemaVarAST;
 using log_surgeon::Token;
 
+namespace {
+[[nodiscard]] auto get_test_dir() -> std::filesystem::path;
+
+auto get_test_dir() -> std::filesystem::path {
+    std::filesystem::path const current_file_path{__FILE__};
+    return std::filesystem::canonical(current_file_path.parent_path());
```
kirkrodrigues, line 38: If we're using `canonical` here, maybe do the same in the other file?

components/core/tests/test-ParserWithUserSchema.cpp
```diff
@@ -28,6 +28,19 @@ using log_surgeon::SchemaAST;
...
+}
+
+auto const test_dir = get_test_dir();
+auto const test_schema_dir = test_dir / "test_schema_files";
+auto const test_query_dir = test_dir / "test_search_queries";
```
kirkrodrigues: Ditto about clang-tidy.

Since we reuse some of these values in multiple cases, can we change them into functions instead?

components/core/tests/test-GrepCore.cpp
```diff
@@ -17,9 +18,22 @@ using log_surgeon::SchemaParser;
 using log_surgeon::SchemaVarAST;
 using std::string;
 
+namespace {
+[[nodiscard]] auto get_test_dir() -> std::filesystem::path;
+
+auto get_test_dir() -> std::filesystem::path {
```
kirkrodrigues: How about `get_tests_dir`? `get_test_dir` sounds like it might be a function to get a testing directory.

components/core/tests/test-GrepCore.cpp
```diff
@@ -17,9 +18,22 @@ using log_surgeon::SchemaParser;
...
+    std::filesystem::path const current_file_path{__FILE__};
+    return current_file_path.parent_path();
+}
+
+auto const test_dir = get_test_dir();
```
kirkrodrigues: Ditto about naming.

components/core/tests/test-GrepCore.cpp
kirkrodrigues: How about `test_schema_files_dir`?

components/core/tests/test-ParserWithUserSchema.cpp
kirkrodrigues: How about `test_queries_dir`?

components/core/tests/test-ParserWithUserSchema.cpp
```diff
@@ -79,56 +92,59 @@ void decompress(std::string archive_dir, std::string output_dir) {
...
+            generate_schema_ast(file_path.string()),
             "Schema:2:11: error: expected ':' before ' ' token\n"
             "          delimiters : \\r\\n\n"
             "                    ^\n"
     );
 }
 
 TEST_CASE("Test creating schema parser", "[LALR1Parser][SchemaParser]") {
-    generate_schema_ast("../tests/test_schema_files/easy_schema.txt");
+    auto const schema_file_path{test_schema_dir / "easy_schema.txt"};
```
kirkrodrigues: I won't complain if someone wants to use `auto ... =` compared to `auto ...{...}`, but we should be consistent. Which one would you prefer to use?

components/core/tests/test-GrepCore.cpp
quinntaylormitchell: agreed

components/core/tests/test-GrepCore.cpp
quinntaylormitchell: agreed

components/core/tests/test-ParserWithUserSchema.cpp
quinntaylormitchell: agreed, forgot about that thank you

components/core/tests/test-GrepCore.cpp
quinntaylormitchell: definitely

quinntaylormitchell: agreed

components/core/tests/test-ParserWithUserSchema.cpp
quinntaylormitchell: agreed

components/core/tests/test-ParserWithUserSchema.cpp
quinntaylormitchell: I prefer `auto ... =` because it's what I'm used to; I think I used `auto ...{...}` because Haiqi suggested it, but then I forgot to change the others. I will use `auto ... =` for now.

components/core/tests/test-ParserWithUserSchema.cpp
quinntaylormitchell: definitely

components/core/tests/test-GrepCore.cpp
```diff
@@ -17,9 +18,21 @@ using log_surgeon::SchemaParser;
...
+
+auto get_tests_dir() -> std::filesystem::path {
+    std::filesystem::path const current_file_path{__FILE__};
+    return std::filesystem::canonical(current_file_path.parent_path());
+}
+}  // namespace
+
 TEST_CASE("get_bounds_of_next_potential_var", "[get_bounds_of_next_potential_var]") {
+    auto const tests_dir = get_tests_dir();
+    auto const test_schema_files_dir = tests_dir / "test_schema_files";
```
kirkrodrigues:
```suggestion
    auto const test_schema_files_dir = get_tests_dir() / "test_schema_files";
```

components/core/tests/test-GrepCore.cpp
```diff
@@ -17,9 +18,21 @@ using log_surgeon::SchemaParser;
...
     ByteLexer lexer;
```
kirkrodrigues: Nit: Can we move this one line lower so that the order of the variable declarations matches the function's arguments?

components/core/tests/test-ParserWithUserSchema.cpp
```diff
@@ -158,10 +181,10 @@ TEST_CASE("Test creating log parser without delimiters", "[LALR1Parser][LogParse
...
 TEST_CASE("Test lexer", "[Search]") {
     ByteLexer lexer;
-    std::string schema_file_name = "../tests/test_schema_files/search_schema.txt";
-    std::string schema_file_path = boost::filesystem::weakly_canonical(schema_file_name).string();
-    load_lexer_from_file(schema_file_path, lexer);
-    FileReader file_reader{"../tests/test_search_queries/easy.txt"};
+    auto const schema_file_path = get_test_schema_files_dir() / "search_schema.txt";
+    load_lexer_from_file(schema_file_path.string(), lexer);
+    auto const query_file_path = get_test_queries_dir() / "easy.txt";
+    FileReader file_reader = query_file_path.string();
```
kirkrodrigues: This isn't an `auto` case, right?

components/core/tests/test-GrepCore.cpp
quinntaylormitchell: agreed

components/core/tests/test-GrepCore.cpp
quinntaylormitchell: definitely

components/core/tests/test-ParserWithUserSchema.cpp
quinntaylormitchell: sorry, what's this referring to?

kirkrodrigues: It should be the below, right?

```suggestion
    FileReader file_reader{query_file_path.string()};
```

PR #1125, opened 2025-07-25
fix(package): Fail job and report failure to user for compression jobs that encounter at least one invalid input path (fixes #308).

### LINKED ISSUE #308: compression_scheduler doesn't report failure with partially invalid input paths
### Bug

When handling compression job with multiple input paths, the compression scheduler silently log and skip any input that doesn't resolve to any real on disk path. 

The compression is only marked as failure if none of the input paths resolve to valid path, otherwise CLP will skip compressing any invalid input path and report success.

We should update the reporting system so that we can report this kind of case as partially fail.

### CLP version

469019e329bea4a0e8fc351b3d26aa976ab7f891

### Environment

Ubuntu 22.04

### Reproduction steps

- Build and start a CLP package
- run` ./sbin/compress.sh <valid_path> <invalid_path>`
- An error log can be observed in compression schdeuler log, but the compression returns success

quinntaylormitchell:
# Description

This PR fixes issue #308 by changing the reporting system that provides the user with information about the status of their compression jobs.

Any compression job that encounters at least one invalid path will be failed. The user will be notified that at least one invalid path has been encountered, and they will be directed to a log file that communicates all these invalid paths.

# Checklist

  breaking change.

# Validation performed

Compressed three arrangements of log files:

1. one real, one fake
2. one real, two fake
3. two fake

The correct log file was produced in all cases, and the user was notified correctly in all cases.

## Summary by CodeRabbit

* **New Features**
  * Per-job user-facing log created for invalid filesystem paths; job status includes a link to the log when available.
  * Scheduling now collects and surfaces path validation details to users.

* **Bug Fixes**
  * Stronger validation of top-level and nested input paths to prevent partial scheduling when any path is invalid.
  * Clearer failure messages for S3 issues and unsupported inputs.
  * Jobs that produce no tasks are no longer auto-marked as failed.


quinntaylormitchell:
@coderabbitai review

haiqi96:
Quick question: what would appear on webui if a job partially fail?

Edit: I guess this is irrelevant now since a job will either pass or fail.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -207,7 +207,20 @@ def search_and_schedule_new_tasks(
...
+                update_compression_job_metadata(
+                    db_cursor,
+                    job_id,
+                    {
+                        "status": CompressionJobStatus.FAILED,
+                        "status_msg": f"Failed to process one or more file input paths.",
+                    },
+                )
+                db_conn.commit()
+                continue
```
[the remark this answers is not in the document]
quinntaylormitchell: my concern is that only the first invalid path will be reported if we report `{str(err)}`, which may be confusing to the user. @haiqi96 what do you think?

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -109,7 +109,7 @@ def _process_fs_input_paths(
             file, empty_directory = validate_path_and_get_info(CONTAINER_INPUT_LOGS_ROOT_DIR, path)
         except ValueError as ex:
             logger.error(str(ex))
-            continue
+            raise ValueError(str(ex))
```
haiqi96: We are duplicating the try-catch in both this method and it's caller, which is unnecessary. In addition, if we are going to log an exception, we can use logger.exception which will print some fancy stack trace for you.

@kirkrodrigues 
Not an expert of error handling so want to get your opinion. I feel it's not a good practice to catch an error and re-raise it. We can go either
1. Remove the try-except block in `_process_fs_input_paths` so it will throw the exception, and let caller handle the exception with try-catch. What Quinn currently have in the caller code should work.
2. Update `_process_fs_input_paths` run try-catch, but return a True/False to indicate success or failure

I feel we should go with 2? but 1 isn't that bad either imo.

haiqi96: Nvm, now I realized we also need to report the path that causes the error. which gives a good reason to re-raise with a different error message. 

Can you raise an error message saying something like "blabla file is invalid blabla" and store that as our status message? (it should be within the 512 char limit if we only report the first file that causes the error)

@kirkrodrigues wonder if you see any better way to propagate the error message? Alternatively, the method can return a tuple of [Bool, Optional[str]] (success_flag, failure_flag), but I don't feel it is any better.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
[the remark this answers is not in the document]
haiqi96: I think it is an acceptable bahavior, since many tools also only report the first error they observe. Given that our length of our status_message is limited, it would be safer to only report the first error.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
quinntaylormitchell: I've changed it so that the first invalid file path is now reported; @kirkrodrigues let us know if you see a better way to propagate the error message.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -207,7 +207,20 @@ def search_and_schedule_new_tasks(
...
+            try:
+                _process_fs_input_paths(input_config, paths_to_compress_buffer)
+            except Exception as err:
+                logger.exception("Failed to process one or more file input paths.")
+                update_compression_job_metadata(
+                    db_cursor,
+                    job_id,
+                    {
+                        "status": CompressionJobStatus.FAILED,
+                        "status_msg": f"Failed to process file input path: {str(err)}",
```
haiqi96: nit: how about
```suggestion
                        "status_msg": f"Failed to process input: {str(err)}",
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -207,7 +207,20 @@ def search_and_schedule_new_tasks(
 
         input_type = input_config.type
         if input_type == InputType.FS.value:
-            _process_fs_input_paths(input_config, paths_to_compress_buffer)
+            try:
+                _process_fs_input_paths(input_config, paths_to_compress_buffer)
+            except Exception as err:
+                logger.exception("Failed to process one or more file input paths.")
```
haiqi96: nit: maybe we should do:
1. store `Failed to process input: {str(err)}` as an error_msg variable
2. use it both in `logger.exception` and for `update_compression_job_metadata`?

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
haiqi96: now I think about it again, we don't really need to catch in _process_fs_input_paths since we are just rethrowing the error? 
we can simply let caller at line [211](https://github.com/y-scope/clp/blob/b31b2bb40982d82a978e1fba0f3f5ac35468d20c/components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py#L211) to catch the exception (and I think you don't even need to change the caller try except)

I will ask Kirk to confirm

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -207,7 +207,20 @@ def search_and_schedule_new_tasks(
...
+                update_compression_job_metadata(
+                    db_cursor,
+                    job_id,
+                    {
+                        "status": CompressionJobStatus.FAILED,
+                        "status_msg": f"Failed to process input: {str(err)}",
```
haiqi96: Likely you don't need str() for err, since f string will do the conversion for you. I could be wrong though. can you try removing str() and see how things go?

quinntaylormitchell: it works!

quinntaylormitchell:
@haiqi96 and @kirkrodrigues, here are my thoughts on what I should change re. catching the error:

I could omit `logger.error(str(ex))` from `_process_fs_input_paths`, and then it would be logged only once (at the caller around line 211)

OR

I could omit `logger.exception(error_msg)` (at the caller around line 211) and then it would still be logged only once (in `process_fs_input_paths`).

Not sure about the whole "double-raising" thing; I don't know how else to pass the error message from `validate_path_and_get_info` all the way to the caller around line 211.

kirkrodrigues:
Sorry, just getting to this now. I feel failing on the first invalid path is a bit harsh since a user might be compressing a directory with files that change or get deleted (e.g., files in `/var/log` during log rotation).

Ideally, we should report a partial error with the job and what paths failed, so that the user could retry just those paths if they like (and maybe one day we can even support retrying the files at the click of a button). However, before then, I think a reasonable compromise is to log the failing paths and inform the user to check the logs for which paths failed. What do y'all think?

haiqi96:
> Sorry, just getting to this now. I feel failing on the first invalid path is a bit harsh since a user might be compressing a directory with files that change or get deleted (e.g., files in `/var/log` during log rotation).
> 
> Ideally, we should report a partial error with the job and what paths failed, so that the user could retry just those paths if they like (and maybe one day we can even support retrying the files at the click of a button). However, before then, I think a reasonable compromise is to log the failing paths and inform the user to check the logs for which paths failed. What do y'all think?



> Sorry, just getting to this now. I feel failing on the first invalid path is a bit harsh since a user might be compressing a directory with files that change or get deleted (e.g., files in `/var/log` during log rotation).
> 
> Ideally, we should report a partial error with the job and what paths failed, so that the user could retry just those paths if they like (and maybe one day we can even support retrying the files at the click of a button). However, before then, I think a reasonable compromise is to log the failing paths and inform the user to check the logs for which paths failed. What do y'all think?

I am not against it. in that case, the `_process_fs_input_paths` method needs to somehow let caller knows what path has failed. I have the following proposal:

1. _process_fs_input_paths() shall return a List[str] of failing path. an empty list means no failure. Alternatively, it can return a Tuple of (success, failing_path): [bool, List[str]] 
2. The caller checks if _process_fs_input_paths succeed. if there's any failing path:
    - It will locally generate a log file, where each line of the log file is a failing path.
    - it shall mark the job as failed without proceeding, with a short error message stating that (1) the job failed because of invalid input path(s), (2), point the user to the generated log file on the disk

@kirkrodrigues any comment?

kirkrodrigues:
> > Sorry, just getting to this now. I feel failing on the first invalid path is a bit harsh since a user might be compressing a directory with files that change or get deleted (e.g., files in `/var/log` during log rotation).
> > Ideally, we should report a partial error with the job and what paths failed, so that the user could retry just those paths if they like (and maybe one day we can even support retrying the files at the click of a button). However, before then, I think a reasonable compromise is to log the failing paths and inform the user to check the logs for which paths failed. What do y'all think?
> 
> > Sorry, just getting to this now. I feel failing on the first invalid path is a bit harsh since a user might be compressing a directory with files that change or get deleted (e.g., files in `/var/log` during log rotation).
> > Ideally, we should report a partial error with the job and what paths failed, so that the user could retry just those paths if they like (and maybe one day we can even support retrying the files at the click of a button). However, before then, I think a reasonable compromise is to log the failing paths and inform the user to check the logs for which paths failed. What do y'all think?
> 
> I am not against it. in that case, the `_process_fs_input_paths` method needs to somehow let caller knows what path has failed. I have the following proposal:
> 
>     1. _process_fs_input_paths() shall return a List[str] of failing path. an empty list means no failure. Alternatively, it can return a Tuple of (success, failing_path): [bool, List[str]]
> 
>     2. The caller checks if _process_fs_input_paths succeed. if there's any failing path:
>        
>        * It will locally generate a log file, where each line of the log file is a failing path.
>        * it shall mark the job as failed without proceeding, with a short error message stating that (1) the job failed because of invalid input path(s), (2), point the user to the generated log file on the disk
> 
> 
> @kirkrodrigues any comment?

Sure

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -103,7 +103,7 @@ def _process_fs_input_paths(
     """
     Iterates through all files in fs_input_conf and adds their metadata to
     `paths_to_compress_buffer`.
-    NOTE: This method skips files that don't exist.
+    NOTE: This method fails on files that don't exist.
     :param fs_input_conf:
     :param paths_to_compress_buffer:
     """
```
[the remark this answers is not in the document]
quinntaylormitchell: no longer relevant

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -217,7 +217,21 @@
...
+                update_compression_job_metadata(
+                    db_cursor,
+                    job_id,
+                    {
+                        "status": CompressionJobStatus.FAILED,
+                        "status_msg": error_msg,
+                    },
+                )
+                db_conn.commit()
+                continue
```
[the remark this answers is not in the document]
quinntaylormitchell: outdated

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -99,22 +99,25 @@ def update_compression_job_metadata(db_cursor, job_id, kv):
...
+) -> bool:
     """
     Iterates through all files in fs_input_conf and adds their metadata to
-    `paths_to_compress_buffer`.
-    NOTE: This method skips files that don't exist.
+    `paths_to_compress_buffer`. Returns True if all paths were processed successfully, else False.
+    NOTE: On encountering an invalid path, this method skips that path, sets paths_ok to False, and continues.
     :param fs_input_conf:
     :param paths_to_compress_buffer:
     """
```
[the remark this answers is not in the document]
quinntaylormitchell: The `validate_path_and_get_info()` function already logs invalid paths in the compression_scheduler logs. We do not need to log them twice.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -99,22 +99,25 @@ def update_compression_job_metadata(db_cursor, job_id, kv):
...
+
     for path_idx, path in enumerate(fs_input_conf.paths_to_compress, start=1):
         path = Path(path)
 
         try:
             file, empty_directory = validate_path_and_get_info(CONTAINER_INPUT_LOGS_ROOT_DIR, path)
         except ValueError as ex:
             logger.error(str(ex))
+            paths_ok = False
             continue
```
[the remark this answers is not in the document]
quinntaylormitchell: The `validate_path_and_get_info()` function already logs invalid paths in the compression_scheduler logs. We do not need to log them twice.

quinntaylormitchell:
@haiqi96 and @kirkrodrigues, I’ve pushed a new implementation. The core elements of the new strategy are a combination of your proposals, as follows:

1. Don’t fail compression jobs upon encountering an invalid path; complete the compression job if there is at least one valid path
2. Log the failing paths somewhere
3. Inform the user to check the logs for which paths failed

That being said, I felt that part of the solution that Haiqi proposed, i.e.

>_process_fs_input_paths() shall return a List[str] of failing path

could instead be fulfilled by the logging infrastructure already used in [this function](https://github.com/y-scope/clp/blob/main/components/clp-py-utils/clp_py_utils/compression.py#L94) (the failed path is already being recorded to the `compression_scheduler` log, which can be viewed by the user). With that in mind, it makes more sense to me to return `bool` from `_process_fs_input_paths`; if we are simply reporting to the user that at least one path failed, and telling them to check the logs for the failed paths, then the caller of `_process_fs_input_paths` doesn’t need to know the path(s) that failed, as the failing path(s) have already been logged in a way that the user can access.

Let me know what you think.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -217,7 +223,20 @@ def search_and_schedule_new_tasks(
...
+                    "Check the compression_scheduler log for more details."
+                )
+                update_compression_job_metadata(
+                    db_cursor,
+                    job_id,
+                    {
+                        "status_msg": error_msg,
+                    },
+                )
+                db_conn.commit()
```
haiqi96, line 288: Based on what I agree with Kirk, we should mark job with partial failing paths as failed.

Refer to https://github.com/y-scope/clp/pull/1125#issuecomment-3229402925

haiqi96:
> > > Sorry, just getting to this now. I feel failing on the first invalid path is a bit harsh since a user might be compressing a directory with files that change or get deleted (e.g., files in `/var/log` during log rotation).
> > > Ideally, we should report a partial error with the job and what paths failed, so that the user could retry just those paths if they like (and maybe one day we can even support retrying the files at the click of a button). However, before then, I think a reasonable compromise is to log the failing paths and inform the user to check the logs for which paths failed. What do y'all think?
> > 
> > 
> > > Sorry, just getting to this now. I feel failing on the first invalid path is a bit harsh since a user might be compressing a directory with files that change or get deleted (e.g., files in `/var/log` during log rotation).
> > > Ideally, we should report a partial error with the job and what paths failed, so that the user could retry just those paths if they like (and maybe one day we can even support retrying the files at the click of a button). However, before then, I think a reasonable compromise is to log the failing paths and inform the user to check the logs for which paths failed. What do y'all think?
> > 
> > 
> > I am not against it. in that case, the `_process_fs_input_paths` method needs to somehow let caller knows what path has failed. I have the following proposal:
> > ```
> > 1. _process_fs_input_paths() shall return a List[str] of failing path. an empty list means no failure. Alternatively, it can return a Tuple of (success, failing_path): [bool, List[str]]
> > 
> > 2. The caller checks if _process_fs_input_paths succeed. if there's any failing path:
> >    
> >    * It will locally generate a log file, where each line of the log file is a failing path.
> >    * it shall mark the job as failed without proceeding, with a short error message stating that (1) the job failed because of invalid input path(s), (2), point the user to the generated log file on the disk
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @kirkrodrigues any comment?
> 
> Sure



> @haiqi96 and @kirkrodrigues, I’ve pushed a new implementation. The core elements of the new strategy are a combination of your proposals, as follows:
> 
> 1. Don’t fail compression jobs upon encountering an invalid path; complete the compression job if there is at least one valid path
> 2. Log the failing paths somewhere
> 3. Inform the user to check the logs for which paths failed
> 
> That being said, I felt that part of the solution that Haiqi proposed, i.e.
> 
> > _process_fs_input_paths() shall return a List[str] of failing path
> 
> could instead be fulfilled by the logging infrastructure already used in [this function](https://github.com/y-scope/clp/blob/main/components/clp-py-utils/clp_py_utils/compression.py#L94) (the failed path is already being recorded to the `compression_scheduler` log, which can be viewed by the user). With that in mind, it makes more sense to me to return `bool` from `_process_fs_input_paths`; if we are simply reporting to the user that at least one path failed, and telling them to check the logs for the failed paths, then the caller of `_process_fs_input_paths` doesn’t need to know the path(s) that failed, as the failing path(s) have already been logged in a way that the user can access.
> 
> Let me know what you think.

I don't agree with the approach of logging the paths in compression_scheduer.
1. It's easier to point user to a specific path, rather than saying "check compression_scheduer log" since most users (and even some people in our team who don't use package frequently) don't know where is compression_scheduer's log.
2. Imagine if the compression scheduler has been running for a while and logged many different things, it would be hard for user to find the exact portion of logs that he/she should read.

Because of the points above, it would makes sense to generate a separate log file for clear/efficient error reporting.

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -86,6 +86,8 @@ def handle_job_update(db, db_cursor, job_id, no_progress_reporting):
             if not no_progress_reporting:
                 logger.info("Compression finished.")
                 print_compression_job_status(job_row)
+                if job_row["status_msg"]:
+                    logger.warning(f"{job_row['status_msg']}")
```
haiqi96: wouldn't need this if we don't allow parital success

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -257,7 +276,7 @@ def search_and_schedule_new_tasks(
                 job_id,
                 {
                     "status": CompressionJobStatus.FAILED,
-                    "status_msg": "invalid input path",
+                    "status_msg": "All provided paths were invalid.",
```
haiqi96: My gut feeling is that we don't need the if clause anymore, because any invalid path should be caught by the previous if-else and end up skipping the rest of the loop with `continue`. Just to be safe, can you take a look if there's any other reason that would have caused len(tasks) to be 0?

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
quinntaylormitchell: ah right. to be honest, I agree with Kirk's original message where he says

>I feel failing on the first invalid path is a bit harsh

I didn't feel that way in the beginning, but now I feel like changing the behaviour of the `compress.sh` command (changing it so that the job fails immediately) is more drastic than just printing the warning to the user.

@kirkrodrigues what do you think?

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
quinntaylormitchell: >any invalid path should be caught by the previous if-else

you're right about that, but there needs to be some differentiation between jobs that have only invalid paths and jobs that have a mixture of valid and invalid paths.

your suggestion here is quite correct if we are doing outright failure upon encountering an invalid path, but I'm still not convinced about that. I will revisit this comment when we clear that up

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
quinntaylormitchell: right; I will revisit this comment when we clear the "partial success" discussion up

quinntaylormitchell:
>most users (and even some people in our team who don't use package frequently) don't know where is compression_scheduer's log

@haiqi96 that's true. But you also mention:

>It's easier to point user to a specific path

... I could rig it so that it prints out the specific path of the `compression_scheduler` log for the user?

As for this point:

>it would be hard for user to find the exact portion of logs that he/she should read

that's fair.

@haiqi96, I guess my objection to your proposal that "`_process_fs_input_paths()` shall return a `List[str]`" lies in the fact that the paths are already being logged in the `compression_scheduler` logs; my approach just directs the user there.

Also, see [my comment](https://github.com/y-scope/clp/pull/1125#discussion_r2319842893) re. the "immediate failure" vs "continue compression" discussion.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -250,18 +321,18 @@ def search_and_schedule_new_tasks(
...
+        #     update_compression_job_metadata(
+        #         db_cursor,
+        #         job_id,
+        #         {
+        #             "status": CompressionJobStatus.FAILED,
+        #             "status_msg": "All provided paths were invalid.",
+        #         },
+        #     )
+        #     db_conn.commit()
+        #     continue
```
quinntaylormitchell: @kirkrodrigues do you think that this commented section can be deleted? I have confirmed (best that I can) that this `if` statement will only be entered if all paths are invalid (there are no other conditions that would cause `len(tasks)` to be 0)

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -99,28 +99,37 @@ def update_compression_job_metadata(db_cursor, job_id, kv):
...
-    `paths_to_compress_buffer`.
-    NOTE: This method skips files that don't exist.
+    Iterates through all files in fs_input_conf and validates their paths. Adds the metadata for
+    each valid path to `paths_to_compress_buffer` as long as an invalid path has not yet been
+    encountered. The error messages from all invalid paths are added to the list
+    `invalid_path_messages`. `paths_ok` set to False upon encountering an invalid path.
     :param fs_input_conf:
     :param paths_to_compress_buffer:
+    :return List of error messages about invalid paths.
     """
```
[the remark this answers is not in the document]
quinntaylormitchell: we don't need to add any paths to the `paths_to_compress_buffer` after encountering a single invalid path because the entire job will be failed. we do need to keep check though, because we need to see which other paths are invalid.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -217,7 +264,31 @@
...
+                update_compression_job_metadata(
+                    db_cursor,
+                    job_id,
+                    {
+                        "status": CompressionJobStatus.FAILED,
+                        "status_msg": error_msg,
+                    },
+                )
+                db_conn.commit()
+                continue
```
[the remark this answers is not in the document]
quinntaylormitchell: we've decided to fail the job on any invalid path.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -250,18 +321,18 @@
...
+        #     update_compression_job_metadata(
+        #         db_cursor,
+        #         job_id,
+        #         {
+        #             "status": CompressionJobStatus.FAILED,
+        #             "status_msg": "All provided paths were invalid.",
+        #         },
+        #     )
+        #     db_conn.commit()
+        #     continue
```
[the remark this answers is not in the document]
quinntaylormitchell: I'll wait for @kirkrodrigues to rule on whether we need this section

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -99,28 +99,37 @@ def update_compression_job_metadata(db_cursor, job_id, kv):
...
-) -> None:
+) -> List[str]:
     """
-    Iterates through all files in fs_input_conf and adds their metadata to
-    `paths_to_compress_buffer`.
-    NOTE: This method skips files that don't exist.
+    Iterates through all files in fs_input_conf and validates their paths. Adds the metadata for
+    each valid path to `paths_to_compress_buffer` as long as an invalid path has not yet been
+    encountered. The error messages from all invalid paths are added to the list
+    `invalid_path_messages`. `paths_ok` set to False upon encountering an invalid path.
```
```suggestion
    Iterates through all paths in `fs_input_conf`, validates them, and adds metadata for each valid
    path to `paths_to_compress_buffer` as long as an invalid path hasn't yet been encountered.
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -99,28 +99,37 @@ def update_compression_job_metadata(db_cursor, job_id, kv):
...
     :param fs_input_conf:
     :param paths_to_compress_buffer:
+    :return List of error messages about invalid paths.
```
```suggestion
    :return: List of error messages about invalid paths.
```

Careful not to miss the second colon.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -159,6 +174,38 @@ def _process_s3_input(
         paths_to_compress_buffer.add_file(object_metadata)
 
 
+def write_failed_path_log(
```
kirkrodrigues: This should start with an `_` to indicate it's private.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -159,6 +174,38 @@ def _process_s3_input(
...
+    invalid_path_messages: List[str], logs_directory: Path, job_id: Any
+) -> Optional[Path]:
+    """
+    Writes the error messages in `invalid_path_messages` to a log file called
+    failed_paths_<job_id>.txt that the user can access. File is located at <logs_directory>/user;
+    this directory is created if it does not already exist.
```
```suggestion
    Writes the error messages in `invalid_path_messages` to a log file,
    `{logs_directory}/user/failed_paths_{job_id}.txt`. The directory will be created if it doesn't
    already exist.
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -159,6 +174,38 @@ def _process_s3_input(
...
+    :param invalid_path_messages:
+    :param logs_directory:
+    :param job_id
+    :return Path to the written log file; None if error is encountered.
```
```suggestion
    :param job_id:
    :return: Path to the written log file or `None` if error is encountered.
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -159,6 +174,38 @@ def _process_s3_input(
...
+    try:
+        user_logs_dir.mkdir(parents=True, exist_ok=True)
+    except Exception:
+        return None
+
+    log_path = user_logs_dir / f"failed_paths_{job_id}.txt"
+    try:
+        with log_path.open("w", encoding="utf-8") as f:
+            timestamp = datetime.datetime.now().isoformat(timespec="seconds")
+            f.write(f"Failed input paths log\nGenerated: {timestamp}\n\n")
```
```suggestion
            f.write(f"Failed input paths log.\nGenerated at {timestamp}.\n\n")
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -159,6 +174,38 @@ def _process_s3_input(
...
+            for msg in invalid_path_messages:
+                f.write(f"{str(msg).rstrip()}\n")
```
```suggestion
                f.write(f"{msg.rstrip()}\n")
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -217,7 +264,31 @@ def search_and_schedule_new_tasks(
 
         input_type = input_config.type
         if input_type == InputType.FS.value:
-            _process_fs_input_paths(input_config, paths_to_compress_buffer)
+            invalid_path_messages = _process_fs_input_paths(input_config, paths_to_compress_buffer)
+            if len(invalid_path_messages) > 0:
+                base_msg = "At least one of your input paths could not be processed. "
```
```suggestion
                base_msg = "At least one of your input paths could not be processed."
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -217,7 +264,31 @@ def search_and_schedule_new_tasks(
...
+
+                user_log_path = write_failed_path_log(
+                    invalid_path_messages, clp_config.logs_directory, job_id
+                )
+                if user_log_path is None:
+                    error_msg = base_msg + (
+                        "Check the compression scheduler logs in "
+                        f"{clp_config.logs_directory} for more details."
```
```suggestion
                        f" Check the compression scheduler logs in {clp_config.logs_directory} for"
                        " more details."
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -217,7 +264,31 @@ def search_and_schedule_new_tasks(
...
+                    )
+                else:
+                    error_msg = base_msg + (f"Check {user_log_path} for more details.")
```
```suggestion
                    error_msg = base_msg + f" Check {user_log_path} for more details."
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -250,18 +321,18 @@ def search_and_schedule_new_tasks(
...
+        #     update_compression_job_metadata(
+        #         db_cursor,
+        #         job_id,
+        #         {
+        #             "status": CompressionJobStatus.FAILED,
+        #             "status_msg": "All provided paths were invalid.",
+        #         },
+        #     )
+        #     db_conn.commit()
+        #     continue
```
```suggestion
```

PR #1126, opened 2025-07-25
build(deps-dev): Bump `yscope-dev-utils` to the newest version.

Bill-hbrhbr:
# Description

As title.

# Checklist

  breaking change.

# Validation performed

* [X] Existing task workflows in GitHub runners succeed as usual.

### DIFF
--- a/tools/scripts/deps-download/init.sh
+++ b/tools/scripts/deps-download/init.sh
@@ -10,7 +10,7 @@ script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 project_root_dir="$script_dir/../../../"
 download_dep_script="$script_dir/download-dep.py"
 
-readonly YSCOPE_DEV_UTILS_COMMIT_SHA="76aecbd6d0db6e2078005431946b5c2e716c4291"
+readonly YSCOPE_DEV_UTILS_COMMIT_SHA="b965f19f3e11c548f32bf7ab8e28ff38906be018"
 python3 "${download_dep_script}" \
     "https://github.com/y-scope/yscope-dev-utils/archive/${YSCOPE_DEV_UTILS_COMMIT_SHA}.zip" \
     "yscope-dev-utils-${YSCOPE_DEV_UTILS_COMMIT_SHA}" \


.gitignore
```diff
@@ -2,3 +2,4 @@
 .task/
 build/
 **/dist/
+.clang-format
```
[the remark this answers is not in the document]
Bill-hbrhbr: the .clang-format is commited inside a submodule and cp-ed over to CLP, so we ignore it

docs/src/dev-guide/contributing-linting.md
```diff
@@ -17,6 +17,7 @@ To run the linting tools, besides commonly installed tools like `tar`, you'll ne
 * python3-venv (for the version of Python installed)
 * [Task] >= 3.38.0 and < 3.43.0
   * We constrain the version due to unresolved [issues][clp-issue-872].
+* [uv]
```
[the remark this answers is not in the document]
Bill-hbrhbr: Currently there's no minimum constraint

Bill-hbrhbr:
@coderabbitai
can you make an issue in `yscope-dev-utils` saying that exclude_patterns in `check-cpp-static-full` is confusing as part of it belongs to the source file root path and part of it belongs to the test root path?
There's currently no good solution. It's just confusing if we happen to have similar relative paths in these two dirs.

taskfiles/lint.yaml
```diff
@@ -187,568 +191,567 @@ tasks:
...
+            - "test-ParserWithUserSchema.cpp"
+            - "test-query_methods.cpp"
+            - "test-Segment.cpp"
+            - "test-SQLiteDB.cpp"
+            - "test-Stopwatch.cpp"
+            - "test-string_utils.cpp"
+            - "test-TimestampPattern.cpp"
+            - "test-utf8_utils.cpp"
+            - "test-Utils.cpp"
           # yamllint enable rule:line-length
```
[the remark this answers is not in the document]
Bill-hbrhbr: currently there's only one reference of this list, so these measures are not very necessary. Also in the future we aim to remove these exclusions almost entirely as we resolve clang-tidy issues one by one.

components/core/tools/docker-images/clp-env-base-ubuntu-jammy/Dockerfile
```diff
@@ -7,6 +7,9 @@ ADD ./tools/scripts/lib_install ./tools/scripts/lib_install
 
 RUN ./tools/scripts/lib_install/ubuntu-jammy/install-all.sh
 
+# Install uv
```
davidlion: nit: unnecessary/redundant comment
```suggestion
```

components/core/tools/scripts/lib_install/macos/install-all.sh
```diff
@@ -33,6 +33,11 @@ curl \
 /Formula/g/go-task.rb
 brew install --formula "$task_formula_path"
 
+# Install `uv`
```
davidlion: nit: unnecessary/redundant comment
```suggestion
```

.gitignore
davidlion: Is there a reason to only add the format config and not tidy? Also, since we're touching this file we might as well fix the ordering:
```
.clang-format
.clang-tidy
.task/
build/
dist/
```

(`**/dist/` should be equivalent to `dist/`)

components/core/tools/docker-images/clp-env-base-centos-stream-9/Dockerfile
```diff
@@ -7,6 +7,9 @@ ADD ./tools/scripts/lib_install ./tools/scripts/lib_install
 
 RUN ./tools/scripts/lib_install/centos-stream-9/install-all.sh
 
+# Install uv
```
davidlion: nit: unnecessary/redundant comment
```suggestion
```

docs/src/dev-guide/contributing-linting.md
```diff
@@ -15,8 +15,10 @@ To run the linting tools, besides commonly installed tools like `tar`, you'll ne
 * `md5sum`
 * Python 3.9 or newer
 * python3-venv (for the version of Python installed)
-* [Task] >= 3.38.0 and < 3.43.0
-  * We constrain the version due to unresolved [issues][clp-issue-872].
+* [Task] >= 3.40.0 and < 3.43.0
+  * Minimum version 3.40.0 is required for [yscope-dev-utils].
+  * We constrain the version to < 3.43.0 due to unresolved [issues][clp-issue-872].
+* [uv]
```
davidlion: Lets add the minimum version that dev utils says for `uv` here.
```suggestion
* [uv] >= 0.7.10
```

components/core/tools/docker-images/clp-env-base-ubuntu-jammy/Dockerfile
```diff
@@ -7,6 +7,9 @@ ADD ./tools/scripts/lib_install ./tools/scripts/lib_install
...
+RUN pip3 install --no-cache-dir uv==0.8.3
```
davidlion: Can we drop the version here since we don't fix the version in CI or mac (brew)?
```suggestion
RUN pip3 install --no-cache-dir uv
```

components/core/tools/docker-images/clp-env-base-centos-stream-9/Dockerfile
```diff
@@ -7,6 +7,9 @@ ADD ./tools/scripts/lib_install ./tools/scripts/lib_install
...
+RUN pip3 install --no-cache-dir uv==0.8.3
```
davidlion: Can we drop the version here since we don't fix the version in CI or mac (brew)?
```suggestion
RUN pip3 install --no-cache-dir uv
```

.github/workflows/clp-core-build.yaml
```diff
@@ -360,6 +360,7 @@ jobs:
           #   task lint:check-cpp-${{(github.event_name == 'schedule') && 'full' || 'diff'}}
           run_command: >-
             CLP_CORE_MAX_PARALLELISM_PER_BUILD_TASK=$(getconf _NPROCESSORS_ONLN)
+            HOME=/tmp
```
davidlion, line 362: What is the benefit of this change? There is no explanation in the description.

Bill-hbrhbr:
@coderabbitai
Looking at my last several commits, due to issue 
https://github.com/actions/runner-images/issues/6775
we have to do `chown $(id -u):$(id -g) -R .` to avoid dubious ownership, and we need to pass `--user` of github runner during action `run-on-image`.
Right now we mount the deps image to `/mnt/repo`. Once we define a proper github runner home, we can mount our image their inside `$HOME/clp`, and our `uv` generated files will also have a place to store their files. The potential directories that could be used by `uv` are listed here: https://specifications.freedesktop.org/basedir-spec/latest/#basics
We set `/tmp` for a temporary $HOME.

Now make an issue saying that we need a proper $HOME for github runner defined in the future, with all the files inside $HOME having the uid and gid of the github runner to avoid dubious ownership.

Bill-hbrhbr:
@coderabbitai

Update the issue 1145.
We need a proper $HOME so that programs like `uv` will have a place to store their files. Currently, without a proper github runner home, the $HOME variable defaults to root dir `/` and github runner doesn't have sudo rights to write inside this dir.
we already have `chown` and `--user` in place to avoid the dubious ownership problem, so your title is wrong. I'm saying we cannot remove `chown` and `--user` because they exist to workaround https://github.com/actions/runner-images/issues/6775. The currently problem is that we don't have a proper $HOME variable.

taskfiles/lint.yaml
```diff
@@ -118,12 +118,18 @@ tasks:
...
-          ROOT_PATHS: *cpp_source_files
+          INCLUDE_FILENAME_PATTERNS: &cpp_include_filename_patterns
+            - "*.cpp"
+            - "*.h"
+            - "*.hpp"
+            - "*.inc"
+          ROOT_PATHS: &cpp_root_paths
+            - "{{.G_CORE_COMPONENT_DIR}}/src"
+            - "{{.G_CORE_COMPONENT_DIR}}/tests"
           VENV_DIR: "{{.G_LINT_VENV_DIR}}"
```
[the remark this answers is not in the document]
Bill-hbrhbr: give me a codeblock on how this may be done

.github/workflows/clp-core-build.yaml
```diff
@@ -358,8 +358,10 @@ jobs:
           # TODO: When enough files are passing clang-tidy, switch to a full pass on schedule only.
           # run_command: >-
           #   task lint:check-cpp-${{(github.event_name == 'schedule') && 'full' || 'diff'}}
+          # TODO: https://github.com/y-scope/clp/issues/1145
```
davidlion:
```suggestion
          # TODO: We set HOME due to permission issues inside the container used in run-on-image.
          # See: https://github.com/y-scope/clp/issues/1145
```

We can drop the other `TODO` as the use of the cache makes it obsolete.

davidlion (changes requested):

PR #1127, opened 2025-07-25
feat(webui): Add support for submitting Presto SQL queries.

hoophalab:
# Description

1. Adds route for Presto search queries.
2. Submits Presto search queries on 'Run' button click.
3. Currently, the search results are printed to the server console. They need to be pushed to mongdo db in future PRs.

@davemarco, after experimenting with presto-client-node, I feel we'll still need a `PrestoSearchManager` for handling cancel requests.

Also, we are probably going to store presto query's states in server's memory, which doesn't sound good.

# Checklist

  breaking change.

# Validation performed

1. Executed `SELECT * FROM tpch.sf100.customer limit 10` in the Presto search webui and verified that the data appeared correctly in the server's console.

2. Executed `SELECT * FROM tpch.sf100.customer` and verified that the data streamed as intended to the server's console.

## Summary by CodeRabbit

* **New Features**
  * Added support for submitting Presto SQL queries from the user interface.
  * Introduced configuration options for connecting to a Presto server.
  * Users can now execute queries using a dedicated "Run" button in the Presto search controls.

* **Chores**
  * Updated server dependencies to include Presto client libraries and cleaned up duplicate entries.
  * Added new API endpoints and schemas to facilitate Presto query submission and job tracking.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -13,7 +13,7 @@ this project:
     the size of repo as we add and update images.
 * [Node.js] >= 16 to be able to [view the output](#viewing-the-output)
 * Python 3.10 or later
-* [Task] >= 3.38.0 and < 3.43.0
+* [Task] >= 3.40.0 and < 3.43.0
   * We constrain the version due to unresolved [issues][clp-issue-872].
 
 ## Build Commands
--- a/docs/src/dev-guide/building-package.md
+++ b/docs/src/dev-guide/building-package.md
@@ -13,7 +13,7 @@ prebuilt version instead, check out the [releases](https://github.com/y-scope/cl
 * Python 3.9 or newer
 * python3-dev
 * python3-venv (for the version of Python installed)
-* [Task] >= 3.38.0 and < 3.43.0
+* [Task] >= 3.40.0 and < 3.43.0
   * We constrain the version due to unresolved [issues][clp-issue-872].
 
 ## Setup
--- a/docs/src/dev-guide/components-core/index.md
+++ b/docs/src/dev-guide/components-core/index.md
@@ -9,7 +9,7 @@ CLP core is the low-level component that performs compression, decompression, an
 * A recent compiler that fully supports C++20 features such as
   * std::span
   * std::source_location
-* [Task] >= 3.38.0 and < 3.43.0
+* [Task] >= 3.40.0 and < 3.43.0
   * We constrain the version due to unresolved [issues][clp-issue-872].
 
 To build, we require some source dependencies, packages from package managers, and libraries built


components/webui/client/src/api/presto-search.ts
```diff
@@ -0,0 +1,43 @@
...
+} from "@sinclair/typebox";
+import {Value} from "@sinclair/typebox/value";
+import axios from "axios";
+
+
+// eslint-disable-next-line no-warning-comments
+// TODO: Replace with shared type from the `@common` directory once refactoring is completed.
+// Currently, server schema types require typebox dependency so they cannot be moved to the
+// `@common` directory with current implementation.
+const StringSchema = Type.String({
```
junhaoliao: How about `NonEmptyString`

components/webui/server/settings.json
```diff
@@ -16,5 +16,9 @@
     "StreamTargetUncompressedSize": 134217728,
     "StreamFilesS3Region": null,
     "StreamFilesS3PathPrefix": null,
-    "StreamFilesS3Profile": null
+    "StreamFilesS3Profile": null,
+
+    "EnablePresto": true,
```
junhaoliao: Since currently we can only toggle the query backend between Presto and "native", would it be better to conform to the same settings interface on the client side. i.e.,
```
"ClpQueryEngine": "native", // or "presto"
```

components/webui/server/settings.json
```diff
@@ -16,5 +16,9 @@
...
+    "PrestoHost": "localhost",
+    "PrestoPort": 8080
```
junhaoliao: It might be better to launch Presto on a different port because we commonly use 8080 as the docs debug server's port.

components/webui/server/src/routes/api/presto-search/index.ts
```diff
@@ -0,0 +1,89 @@
...
+            schema: {
+                body: PrestoSearchJobCreationSchema,
+                response: {
+                    [StatusCodes.CREATED]: PrestoJobSchema,
+                    [StatusCodes.INTERNAL_SERVER_ERROR]: ErrorSchema,
+                },
+                tags: ["Search"],
+            },
+        },
+
```
junhaoliao, line 39: if not added by the linter
```suggestion
```

components/webui/server/package.json
```diff
@@ -38,10 +38,12 @@
     "fastify-plugin": "^5.0.1",
     "http-status-codes": "^2.3.0",
     "pino-pretty": "^13.0.0",
+    "presto-client": "^1.1.0",
```
junhaoliao, line 41: It seems the official https://www.npmjs.com/package/@prestodb/presto-js-client is slightly more popular than the `presto-client` package. Shall we use `@prestodb/presto-js-client` instead, in the hope of getting better long-term support?

components/webui/server/src/routes/api/presto-search/index.ts
```diff
@@ -0,0 +1,89 @@
...
+        async (request, reply) => {
+            const {queryString} = request.body;
+
+            let searchJobId: Nullable<string> = null;
+
+            searchJobId = await new Promise((resolve) => {
+                client.execute({
+                    data: (error, data, columns) => {
+                        if (null === searchJobId) {
+                            request.log.error("Data arrived from Presto before the search job id.");
```
junhaoliao: in what scenario is this expected to happen?

components/webui/server/src/routes/api/presto-search/index.ts
```diff
@@ -0,0 +1,89 @@
...
+                        }
+                        request.log.info({error, searchJobId, columns, data}, "Presto data");
```
junhaoliao: can we improve this message a bit?

junhaoliao (dismissed):
(i haven't validated the changes) on a high level, the code lgtm

we shall see if "@prestodb/presto-js-client" can be a better choice than "presto-client" before making further changes

components/webui/client/src/api/presto-search.ts
junhaoliao, line 1: sorry for missing this in the previous reviews - for naming source files, shall we consistently use `camelCase.<extension>`? e.g. `prestoSeach.ts` here.

davemarco: I feel like we do this for classes, and use the `-` for regular names. At least this is closer to the clp c++ style. anyways let me know your thoughts but i am okay with presto-search.ts

components/webui/client/src/api/presto-search.ts
davemarco: I removed this for now and removed typebox to match other api calls. We can revisit this when i migrate common types in workspace PR.

components/webui/server/package.json
davemarco: I was researching this before hoophalab started the pr, and I realized the `@prestodb/presto-js-client` dosent support streaming. The presto API routes are versioned `v1` so unlikely to change in future. If we do need to update, we can eventually migrate the existing `presto-client` into our repo since its small package <1000 lines.

components/webui/server/src/routes/api/presto-search/index.ts
davemarco: its not. i removed but i think was there maybe for typings purposes

components/webui/server/src/routes/api/presto-search/index.ts
davemarco: these are going to get removed, i added a todo for it

junhaoliao (commented):
didn't test but the code lgtm. @hoophalab please do a final quick look

davemarco:
> @davemarco, after experimenting with presto-client-node, I feel we'll still need a PrestoSearchManager for handling cancel requests.

There is a kill api no? Like the client just sends back the id, and we kill the query

hoophalab:
> > @davemarco, after experimenting with presto-client-node, I feel we'll still need a PrestoSearchManager for handling cancel requests.
> 
> There is a kill api no? Like the client just sends back the id, and we kill the query

I was thinking when the user sends an ID to cancel a query, we need to tell `presto-client-node` to stop it. Based on my previous research, this can be only done by returning true in the cancel callback.

This means we'll need a map to link search job IDs to their cancellation status. But storing state in server's memory might not be ideal for failure recovery or out-of-memory issues.

However, we should be fine with merely storing ephemeral search job ids. So, let's proceed with that approach.

davemarco:
@hoophalab - so your saying we cant just use this https://github.com/tagomoris/presto-client-node?tab=readme-ov-file#killquery_id-callback?

hoophalab:
> @hoophalab - so your saying we cant just use this https://github.com/tagomoris/presto-client-node?tab=readme-ov-file#killquery_id-callback?

I didn't notice this function before. This function looks correct and we can use it.

kirkrodrigues (approved):
Deferring to @hoophalab and @junhaoliao's reviews.

components/webui/server/settings.json
```diff
@@ -16,5 +16,9 @@
     "StreamTargetUncompressedSize": 134217728,
     "StreamFilesS3Region": null,
     "StreamFilesS3PathPrefix": null,
-    "StreamFilesS3Profile": null
+    "StreamFilesS3Profile": null,
+
+    "ClpQueryEngine": "native",
+    "PrestoHost": "localhost",
+    "PrestoPort": 8889
```
junhaoliao, line 23: we should also configure these in `server_settings_json_updates` in `components/clp-package-utils/clp_package_utils/scripts/start_clp.py`

PR #1128, opened 2025-07-25
fix(core): Drop fractional seconds from formatted timestamps used for S3 signed-URL generation (fixes #1131).

### LINKED ISSUE #1131: AWS S3 authentication fails due to fractional seconds in timestamps
## Description

AWS S3 presigned URL generation was failing due to incorrect timestamp formatting that included fractional seconds. AWS requires timestamps to follow the ISO8601 Long Format `yyyyMMdd'T'HHmmss'Z'` without fractional seconds, but the current implementation was producing timestamps like `2023-03-22T03-45-46.234232`.

## Root Cause

The `get_formatted_timestamp_string` function in `components/core/src/clp/aws/AwsAuthenticationSigner.cpp` was formatting timestamps directly from `std::chrono::system_clock::time_point` without truncating to second precision, which could include sub-second precision depending on the fmt library version or C++ version used.

## Impact

- S3 authentication fails when using clp-s tool with timestamp keys
- Prevents successful compression operations with S3 storage

## Solution

The issue was resolved by explicitly casting the timepoint to second precision using `std::chrono::time_point_cast<std::chrono::seconds>` before formatting.

## References

- Fixed in PR: https://github.com/y-scope/clp/pull/1128
- Comment: https://github.com/y-scope/clp/pull/1128#discussion_r2233169218

## Validation

The fix was validated by:
1. Exporting AWS credentials
2. Building the clp-s tool  
3. Running compression command with S3 authentication and timestamp key
4. Confirming the tool failed without the fix but succeeded after applying the change

haiqi96:
# Description

Aws requires that the timestamp in presigned Url strictly follow the ISO8601 Long Format "yyyyMMdd'T'HHmmss'Z'"
The current implementation of timestamp [formatter](https://fmt.dev/11.1/syntax/#chrono-format-specifications) 'S' will add fractional number and generate timestamp such as `2023-03-22T03-45-46.234232 which violates the requirement of AWS.

This PR resolves the issue by adding an explicit timepoint cast to convert the timestamp into second precision.

Note: I am not sure why this was not an issue in the previous commit. Perhaps it was some undefined behavior and recently we updated either fmt verison or c++ version that revealed this issue. 
I have verified that commit 037cf108f62aee7baba88345e5fc4d38788c5c35 doesn't have this issue, so perhaps we can do a binary search to find out the first commit revealing the issue if we want to understand the root cause

# Checklist

  breaking change.

# Validation performed
1. Exported access_key_id and secret_access_key in my terminal (please contact me if you want to reproduce with the same set of key)
2. built clp-s
3. Ran the compression cmd `./clp-s c output --print-archive-stats --auth s3 --timestamp-key 't.$date' https://yscope-log-compression-dataset-us-west-1.s3.us-west-1.amazonaws.com/mongodb-8gb/mongod.log.2023-03-22T03-45-46`
4. Verified that
- clp-s failed without my change
- clp-s successfully compressed the log from s3 with my change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timestamp formatting to ensure times are displayed with second-level precision, avoiding display of sub-second values.

### DIFF
--- a/components/core/src/clp/aws/AwsAuthenticationSigner.cpp
+++ b/components/core/src/clp/aws/AwsAuthenticationSigner.cpp
@@ -91,7 +91,8 @@ auto is_unreserved_characters(char c) -> bool {
 
 auto get_formatted_timestamp_string(std::chrono::system_clock::time_point const& timestamp)
         -> string {
-    return fmt::format("{:%Y%m%dT%H%M%SZ}", timestamp);
+    auto const timestamp_secs = std::chrono::time_point_cast<std::chrono::seconds>(timestamp);
+    return fmt::format("{:%Y%m%dT%H%M%SZ}", timestamp_secs);
 }
 
 auto get_formatted_date_string(std::chrono::system_clock::time_point const& timestamp) -> string {


haiqi96:
> LGTM. I'm also curious why this hasn't been an issue before.
> 
> For title since this is under `src/clp/aws` maybe it should be `fix(clp)` or `fix(core)`? Also the PR title should end with a period.



> LGTM. I'm also curious why this hasn't been an issue before.
> 
> For title since this is under `src/clp/aws` maybe it should be `fix(clp)` or `fix(core)`? Also the PR title should end with a period.

Added another note in the description about a commit that is known to work (with this issue hidden)

kirkrodrigues:
Caused by #1054 which upgraded fmtlib.

components/core/src/clp/aws/AwsAuthenticationSigner.cpp
```diff
@@ -91,7 +91,8 @@ auto is_unreserved_characters(char c) -> bool {
 
 auto get_formatted_timestamp_string(std::chrono::system_clock::time_point const& timestamp)
         -> string {
-    return fmt::format("{:%Y%m%dT%H%M%SZ}", timestamp);
+    auto const timestamp_secs = std::chrono::time_point_cast<std::chrono::seconds>(timestamp);
+    return fmt::format("{:%Y%m%dT%H%M%SZ}", timestamp_secs);
```
kirkrodrigues, line 95: @coderabbitai Can you open a bug report corresponding to this fix?

kirkrodrigues (approved):
Deferring to @gibber9809's review.

For the PR title, how about:

```
fix(core): Drop fractional seconds from formatted timestamps used for S3 signed-URL generation (fixes #1131).
```

PR #1129, opened 2025-07-25
fix(package): Fix potential error code in the package. 

haiqi96:
# Description

There are few places of code in CLP that could cause issues when running on cloud. This PR fixes all of them

This PR adds the following 3 fixes:
1. Only add stream_output_dir to the necessary_mount if stream storage type is filesystem. This prevent the webui from mounting the stream_storage_dir when storing stream on s3.
A potential error could be caused by the existing code: 1. If webui runs on another a different node from extraction worker, so the staged-stream dir wouldn't be created on webui's node. 2. On certain machine, for example ec2's arm instance. docker doesn't allow binding to a non-existing directory.

2. Add __str__ and to_str() method for scheduler enum. On ec2's arm machine, without those methods, python convert enums to int instead of string and has caused failures when updating the database.

3. Properly implement dump_to_primitive_dict for S3IngestionConfig.

# Checklist

  breaking change.

# Validation performed

### DIFF
--- a/components/clp-package-utils/clp_package_utils/scripts/start_clp.py
+++ b/components/clp-package-utils/clp_package_utils/scripts/start_clp.py
@@ -957,7 +957,6 @@ def start_webui(
     ]
     necessary_mounts = [
         mounts.clp_home,
-        mounts.stream_output_dir,
     ]
     if StorageType.S3 == stream_storage.type:
         auth = stream_storage.s3_config.aws_authentication
@@ -973,6 +972,9 @@ def start_webui(
                 necessary_mounts.append(mounts.aws_config_dir)
             if aws_env_vars:
                 necessary_env_vars.extend(aws_env_vars)
+    else:
+        necessary_mounts.append(mounts.stream_output_dir)
+
     append_docker_options(container_cmd, necessary_mounts, necessary_env_vars)
     container_cmd.append(clp_config.execution_container)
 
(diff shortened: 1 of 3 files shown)


junhaoliao (changes requested):
the current changes in start_clp.py lgtm

as @haiqi96 brought up offline, a warning message about the stream file folder being non-existent is printed when the stream files should be served on s3. let's wrap https://github.com/y-scope/clp/blob/ac6f359737309351a22a6fd7a8e9a0daee9ca8b2/components/webui/server/src/routes/static.ts#L20-L28 with a check on the `settings`.


as for what to check from the settings, i believe checking for null of `StreamFilesDir` might be good in terms of semantics, which will require us to supply it conditionally in start_clp.py . what do you think?

components/webui/server/src/routes/static.ts
```diff
@@ -18,14 +18,21 @@ const routes: FastifyPluginAsync = async (fastify) => {
     const rootDirname = path.resolve(dirname, "../../../..");
 
     let streamFilesDir = settings.StreamFilesDir;
-    if (false === path.isAbsolute(streamFilesDir)) {
-        streamFilesDir = path.resolve(rootDirname, streamFilesDir);
+
+    // Only register /streams if `streamFilesDir` is set (i.e. fs support is configured in package)
```
junhaoliao:
```suggestion
    // Register /streams only if `streamFilesDir` is set (i.e., FS support is enabled in the
    // package)
```

junhaoliao (dismissed):
lgtm. adding @davemarco & @hoophalab for awareness

junhaoliao (approved):

Issue #1090, opened 2025-07-09
[build] `ystdlib` dependency fails to build on older macOS


sitaowang1998:
### Bug

#1038 introduces installing `ystdlib` as a CMake package. However, on macOS, all dependencies are built with `AppleClang` instead of `clang++` from `llvm@16` installed through `homebrew`. On older macOS, i.e. macOS-13 and macOS-14, the old versions of `AppleClang` do not support C++ 20, which is required by `ystdlib`. Thus, `ystdlib` cannot be built on these platforms. #1038 temporarily disables GitHub workflows for these two platforms.

# Solutions explored
We have tried to set up environment variables `LLVM_PREFIX`, `CC`, `CXX`, `AR`, `RANLIB` through `GITHUB_ENV` to instruct `CMake` to use `clang++` to build all dependencies. However, the environment variables confuses `clang-tidy` to use `llvm@16` system libraries, while `clp` and all dependencies are built using macOS sdk system libraries for compatibility with `homebrew` installed libraries.

### CLP version

6815890

### Environment

macOS-13 and macOS-14

### Reproduction steps

GitHub workflows fail

PR #1038, opened 2025-06-26
build!: Install `ystdlib` as a package to avoid duplicate `ystdlib` files.

sitaowang1998:
# Description

Both `clp` and `spider` add `ystdlib` as subdirectory. When integrating `spider` into `clp`, compilation fails because duplicated `ystdlib` files.

This PR updates `ystdlib` version to use the latest version that supports install `ystdlib` as a package. This PR changes the dependency install tasks to install `ystdlib` as a package.

Because the updated `ystdlib` requires `CMake` minimum version 3.23, this PR also updates minimum `CMake` version required and updates GitHub workflows to install `CMake` 3.23.5 for ubuntu jammy.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed
* [x] GitHub workflows pass.
* [x] `ystdlib` installed as a package on ubuntu noble.
* [x] Core compiles with `ystdlib` package.  

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

PR #1130, opened 2025-07-25
ci: Drop support for macOS 13 & 14 to prepare to upgrade ystdlib-cpp; Lock GH runner images to latest available rather than latest stable.

davidlion:
# Description

This PR drops CI support for `macos-13` and `macos-14` to enable upgrading ystdlib-cpp (#1038). #1090 outlines a problem #1038 ran into when trying to support the older macos versions. Although this PR doesn't fix the fundamental problem, it does work around it in the short term.

We also explicitly lock the Github runner images' version to the latest available rather than using `-latest`. This means we will need to manually/explicitly keep these versions updated as Github releases new images, but it avoids random breaks in CI and accidentally dropping versions when `-latest` changes .

Finally, we do some minor refactoring to change any lists from:

```yaml
list: ["entry"]
```

to:

```yaml
list:
  - "entry"
```

# Checklist

  breaking change.

# Validation performed

CI passing.

## Summary by CodeRabbit

* **Chores**
  * Updated GitHub Actions workflows to use explicit operating system versions, replacing generic labels with specific versions (e.g., "ubuntu-24.04" and "macos-15").
  * Removed older macOS versions from the macOS build workflow, now targeting only "macos-15".
  * Adjusted conditional steps in workflows to match the updated OS version identifiers.

### DIFF
--- a/.github/workflows/clp-core-build-macos.yaml
+++ b/.github/workflows/clp-core-build-macos.yaml
@@ -45,8 +45,6 @@ jobs:
     strategy:
       matrix:
         os:
-          - "macos-13"
-          - "macos-14"
           - "macos-15"
         use_shared_libs:
           - true
--- a/.github/workflows/clp-core-build.yaml
+++ b/.github/workflows/clp-core-build.yaml
@@ -42,8 +42,7 @@ concurrency:
 jobs:
   filter-relevant-changes:
     name: "filter-relevant-changes"
-    runs-on:
-      - "ubuntu-24.04"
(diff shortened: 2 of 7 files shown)


kirkrodrigues (approved):
For the PR title, how about:

```
ci: Drop support for macOS 13 & 14 to prepare to upgrade ystdlib-cpp; Lock GH runner images to latest available rather than latest stable.
```

We should also mention latest stable -> latest available in the PR description.

PR #1132, opened 2025-07-27
feat(presto-clp): Add Docker compose setup for Presto cluster that can connect to clp-json.

anlowee:
# Description

This PR creates a docker-compose project that can quickly set up a docker stack with Presto coordinator and Presstimmo worker for CLP connector E2E demo.

# Checklist

  breaking change.

# Validation performed

Did E2E testing with processed MongoDB 256MB dataset.

## Summary by CodeRabbit

**New Features**
  * Added deployment scripts, configuration templates, and environment files for setting up Presto with CLP integration using Docker Compose.
  * Introduced automated configuration generation for both coordinator and worker nodes.
  * Added environment variable support for flexible deployment.
  * Included JVM and logging configuration templates for the Presto coordinator.
  * Added scripts to generate user environment variable files and automate setup with Python virtual environments.

**Documentation**
  * Added a new user guide on using Presto with CLP, including setup instructions and usage examples.
  * Updated the documentation index and overview to include the new Presto guide.

**Chores**
  * Expanded linting to cover deployment files.
  * Added dependency requirements for deployment scripts.
  * Added .gitignore for deployment scripts directory.


docs/src/user-guide/guides-using-presto.md
```diff
@@ -0,0 +1,118 @@
...
+
+```sql
+SELECT foo.bar FROM default LIMIT 1;
+```
+
+## Limitations
+
+The Presto CLP integration the following limitations at present:
+
+* Nested-fields that contain special characters (e.g., `t.$date`, where `$` is considered a special
```
kirkrodrigues: @anlowee Do you know exactly what characters are allowed and disallowed?

tools/deployment/presto-clp/coordinator/config-template/metadata-filter.json
```diff
@@ -0,0 +1,2 @@
+{
+}
```
kirkrodrigues, line 2: @anlowee I think we should explain to the user how to configure this for the timestamp field in their logs, right? And also that it may need to be different for each dataset they compress.

docs/src/user-guide/guides-using-presto.md
anlowee: any chars except `_` when using Presstimo

tools/deployment/presto-clp/coordinator/config-template/metadata-filter.json
anlowee: Can we direct them to the related presto-doc section?

docs/src/user-guide/guides-overview.md
```diff
@@ -12,6 +12,13 @@ Using object storage
 Using CLP to ingest logs from object storage and store archives on object storage.
 :::
 
+:::{grid-item-card}
+:link: guides-using-presto
+Using Presto with CLP
```
anlowee, line 17: Shall we replace all presto to prestissimo?

anlowee: Or maybe mention it somewhere, because in our image name there is "prestissimo"

tools/deployment/presto-clp/scripts/set-up-config.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+set -eu
+set -o pipefail
+
+script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+
+cUsage="Usage: ${BASH_SOURCE[0]} <clp-package-dir>"
+if [ "$#" -lt 1 ]; then
```
anlowee, line 9: shall we use `-eq`?

tools/deployment/presto-clp/scripts/generate-user-env-vars-file.py
```diff
@@ -0,0 +1,172 @@
...
+
+    credentials_file_path = clp_package_dir / "etc" / "credentials.yml"
+    with open(credentials_file_path, "r") as credentials_file:
+        credentials = yaml.safe_load(credentials_file)
+
+    database_user = _get_config_value(credentials, "database.user")
+    database_password = _get_config_value(credentials, "database.password")
+    if not database_user or not database_password:
+        logger.error(
+            "database.user and database.password must be specified in '%s'.", credentials_file_path
```
anlowee, line 128: Shall we tell user that to launch clp-package first? Because that will generate the credentials.

docs/src/user-guide/guides-using-presto.md
```diff
@@ -0,0 +1,126 @@
...
+```sql
+SELECT * FROM default LIMIT 1;
+```
+
+All kv-pairs in each log event can be queried directly using dot-notation. For example, if your logs
+contain the field `foo.bar`, you can query it using:
+
+```sql
+SELECT foo.bar FROM default LIMIT 1;
+```
```
anlowee, line 134: Shall we also mention that this only works when the user didn't setup any "required" metadata filters

tools/deployment/presto-clp/coordinator-common.env
```diff
@@ -0,0 +1,5 @@
+PRESTO_COORDINATOR_HTTPPORT=8080
+PRESTO_COORDINATOR_SERVICENAME=presto-coordinator
+
+# node.properties
+PRESTO_COORDINATOR_NODEPROPERTIES_ENVIRONMENT=production
```
anlowee, line 5: I used "common" was because `PRESTO_COORDINATOR_HTTPPORT` and `PRESTO_COORDINATOR_SERVICENAME` were not used only in certain config file, but across coordinator's configs and worker's configs.

Why we put `node.properties` section here?

docs/src/user-guide/guides-overview.md
kirkrodrigues: https://prestodb.io/docs/0.285/develop/presto-native.html

> Prestissimo is a C++ drop-in replacement for Presto workers.

Presto is the product, Prestissimo is basically a sub-component, right?

tools/deployment/presto-clp/scripts/set-up-config.sh
kirkrodrigues: Do you mean `-ne`? `-lt` is better for long term maintenance if we add more args, right?

docs/src/user-guide/guides-using-presto.md
kirkrodrigues: Good idea. That should probably go in the section about configuring the metadata filter.

tools/deployment/presto-clp/coordinator-common.env
kirkrodrigues: These are properties that are needed in both the coordinator and worker, right?

tools/deployment/presto-clp/coordinator/config-template/metadata-filter.json
kirkrodrigues: That's not published and is more general than they need, right? We should write a simplified section for them here.

tools/deployment/presto-clp/scripts/set-up-config.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+source "${script_dir}/.venv/bin/activate"
+
+echo "Installing required Python packages..."
+pip3 install -r "${script_dir}/requirements.txt"
+
+echo "Generating environment variables file for user-configured properties..."
+python3 "${script_dir}/generate-user-env-vars-file.py" \
+    --clp-package-dir "${clp_package_dir}" \
+    --output-file "${script_dir}/../.env"
```
junhaoliao, line 28: the `.env` file should be added to https://github.com/y-scope/clp/blob/main/tools/deployment/presto-clp/.gitignore 

tools/deployment/presto-clp/docker-compose.yaml
```diff
@@ -0,0 +1,48 @@
...
+        condition: "service_healthy"
+    entrypoint: ["/bin/bash", "-c", "/scripts/generate-configs.sh && /opt/entrypoint.sh"]
+    env_file:
+      - ".env"
+      - "coordinator-common.env"
+      - "worker.env"
+    volumes:
+      - "./worker/config-template:/configs:ro"
+      - "./worker/scripts:/scripts:ro"
+      - "${CLP_ARCHIVES_DIR}:${CLP_ARCHIVES_DIR}"
```
junhaoliao, line 37: The Presto Docker Compose project and the CLP Package Docker Compose project (introduced in #1178) use different conventions for file mounting, which causes Presto queries to return zero rows.

* **Presto project:** It mounts the host's `/path/to/var/data/archives` directory (or `/path/to/var/data/staged-archives` when using S3 output) into the container **at the same absolute path** as on the host.
* **CLP Package project:** It always mounts the host's `/path/to/var/data/archives` directory to `/var/data/archives` in the container, and `/path/to/var/data/staged-archives` to `/var/data/staged-archives`.

Before #1178, when the components were orchestrated via Python subprocess calls to the `docker` CLI, they followed the same mapping convention as the current Presto project.

With the `clp-s` storage engine, when archives are compressed through the Package project, the metadata DB's `clp-datasets` table stores the `archive_storage_directory` field using container paths (e.g., `/var/data/archives`) instead of the corresponding host paths (e.g., `/path/to/var/data/archives`). This mismatch prevents the Presto coordinator from locating the archives, as they are not mounted under `/var/data/archives` in the Presto container.

@coderabbitai create an issues to track the fix for this inconsistency

tools/deployment/presto-clp/worker/config-template/config.properties
junhaoliao, line 1: we should set `system-memory-gb` to be about 90% of available machine memory of the deployment according to https://prestodb.io/docs/current/presto_cpp/properties.html#system-memory-gb

@coderabbitai file an issue

tools/deployment/presto-clp/docker-compose.yaml
```diff
@@ -0,0 +1,48 @@
...
+    healthcheck:
+      test:
+        - "CMD"
+        - "curl"
+        - "-f"
+        - "${PRESTO_COORDINATOR_CONFIGPROPERTIES_DISCOVERY_URI}/v1/info"
+      interval: "10s"
+      retries: 30
+
+  presto-worker:
```
junhaoliao, line 24: we should define a `healthcheck` for the worker as well; otherwise, users who run `docker compose up --detach --wait` would observe the services are started successfully, but the worker service may silently fail (since we `--detach` the output from the console), which may cause confusions to users about the status of the cluster

@coderabbitai file an issue

PR #1133, opened 2025-07-28
fix(package): Add `StatusIntEnum` class to match the behavior of `IntEnum` in Python 3.11.

haiqi96:
# Description

This PR adds `StatusIntEnum` which derived on `IntEnum` so that `__str__`'s behavior is aligned with python 3.11, as documented in [here](https://docs.python.org/3/library/enum.html#:~:text=with%201.-,Changed%20in%20version%203.11%3A%20__str__()%20is%20now%20int.__str__()%20to%20better%20support%20the%20replacement%20of%20existing%20constants%20use%2Dcase.%20__format__()%20was%20already%20int.__format__()%20for%20that%20same%20reason.,-class%20enum)

In our code, we currently rely on the database connector to implicilty convert the scheduler enums when writting them to the database. 
Note that when using mysql, the library (likely `mysql-connector-python` ) converts the enum using __str__(). With a lower python version, the enum will be converted to string and cause an error.

As a sollution, this PR introduce `StatusIntEnum` which overloads the __str__ method, and let all status related enum to derive on `StatusIntEnum`.

To reproduce, change the database type to "mysql" and compress any input logs. the compression scheduler will fail with error log: 

```
  File "/opt/clp/lib/python3/site-packages/mysql/connector/connection_cext.py", line 697, in cmd_query
    self._cmysql.query(
_mysql_connector.MySQLInterfaceError: Unknown column 'CompressionJobStatus.RUNNING' in 'field list'
```

Note the same issue is not observed when using mariadb, which could be internally handled by mariadb's python library.

# Checklist

  breaking change.

# Validation performed

Configured database.type = mysql
Compressed mongodb log from local file storage, ensured the compression scheduler doesn't crash

## Summary by CodeRabbit

* **New Features**
  * Enhanced status display options for compression job and task statuses, allowing for improved string representations in the user interface.


components/job-orchestration/job_orchestration/scheduler/constants.py
```diff
@@ -17,18 +17,36 @@ class CompressionJobStatus(IntEnum):
     SUCCEEDED = auto()
     FAILED = auto()
 
+    def __str__(self) -> str:
+        return str(self.value)
```
junhaoliao, line 21: for ease of debugging, would it be more intuitive to show the enum names intead?

haiqi96:
> > On ec2's arm machine, without those methods, python convert enums to int instead of string and has caused failures when updating the database.
> 
> Mind giving a code example of how the enum is converted?

one example can be found here
https://github.com/y-scope/clp/blob/main/components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py#L249.

Inside update_compression_job_metadata(), the compression_job_status enum will be converted, I guess the __str__ method will be implicitly called.

components/job-orchestration/job_orchestration/scheduler/constants.py
haiqi96: Might have given a wrong description in my PR (now fixed). we want it to be converted to int to match with what database expects.

components/job-orchestration/job_orchestration/scheduler/constants.py
junhaoliao, line 1: instead of overriding individual classes' `__str__` method, can we add a mixin like
```py
class IntEnumStrMixin:
    """
    Delegates __str__ to int.__str__, matching the behavior of IntEnum
    in Python 3.11+.
    TODO: Remove this when our minimum supported Python version is 3.11+.
    """

    def __str__(self):
        return int.__str__(self.__getattribute__("value"))

class CompressionJobStatus(IntEnumStrMixin, IntEnum):
    PENDING = 0
    RUNNING = auto()
    SUCCEEDED = auto()
    FAILED = auto()

```
alternatively, we can also extend the `IntEnum` class into a custom class and let the existing enum classes extend from it.

haiqi96: Updated my code. let me know if you agree with the current approach. If you don't see any issues, I will then update the PR description.

junhaoliao (approved):

junhaoliao (approved):
for the title, how about
```
fix(package): Add `__str()__` and `to_str()` methods to job orchestration enums for MySQL connector compatibility.
```

haiqi96:
> for the title, how about
> 
> ```
> fix(package): Add `__str()__` and `to_str()` methods to job orchestration enums for MySQL connector compatibility.
> ```

I will think about the title since now I am introducing a new class. Will let you know when I finish updating the PR.

junhaoliao (approved):
For the title, let's capitalize the "P" in "Python"

PR #1136, opened 2025-07-29
fix(webui): Cast `BIGINT` values as `UNSIGNED` to fix MySQL-specific type errors in dashboard stats queries (fixes #1137).

### LINKED ISSUE #1137: Fix invalid SQL queries causing MySQL compatibility issues in WebUI
## Problem Description

The WebUI fails to load timestamp ranges properly when using MySQL as the underlying storage engine due to invalid SQL syntax in CAST operations.

## Root Cause

The keyword "INTEGER" is not a valid target for CAST operations in MySQL, causing SQL syntax errors.

## Impact

- Timestamp range functionality broken in WebUI when using MySQL storage engine
- Error logs show: "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'INTEGER'"

## Solution

Replace "INTEGER" with "UNSIGNED" in CAST operations to ensure MySQL compatibility while maintaining MariaDB compatibility.

## Steps to Reproduce

1. Configure MySQL as the underlying storage engine
2. Compress logs with valid timestamps  
3. Check WebUI - timestamp range fails to load properly
4. Check container logs for SQL syntax errors

## Related

- Fixed in PR: https://github.com/y-scope/clp/pull/1136
- Requested by: @haiqi96

## Validation

- [x] Issue resolved with UNSIGNED keyword
- [x] MariaDB compatibility maintained
- [x] WebUI timestamp functionality restored

haiqi96:
# Description

This PR updates the sql query used by the webui. The keyword "INTEGER" seems to not be a valid target for CAST in mysql, and will cause a failure if the underlying database type is mysql. This PR replaces the keyword with "UNSIGNED" to resolve the issue.

# Steps to reproduce
1. Modify the config file to use mysql as underlying database type
2. compress any logs with valid timestamp
3. Check the webui, you can observe that timestamp range is not properly loaded.
4. run docker logs <webui_component_name>, you can see error logs saying `"...You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'INTEGER\n ...`

# Checklist

  breaking change.

# Validation performed
1. Confirmed that the issue described above is gone.
2. Confirmed that when database type is MariaDB, there's no new issue.

## Summary by CodeRabbit

* **Refactor**
  * Updated the handling of message count values to use unsigned integers in dataset details displays.


haiqi96:
@coderabbitai Can you creata an issue based on this PR?

components/webui/client/src/pages/IngestPage/Details/sql.ts
junhaoliao, line 1: @coderabbitai Ideally, the server should catch and handle any query errors instead of leaving it uncaught which causes 500 errors to return the client. create an issue to track this

junhaoliao (approved):
i was able to reproduce the issue with `database.type` changed to "mysql" in `clp-config.yml`. with the fix, it is confirmed that the error is no longer observed

for the PR title, how about:
```
fix(webui): Cast as UNSIGNED to fix MySQL-specific type errors in dashboard stats queries.
```

cc @davemarco @hoophalab for awareness

kirkrodrigues:
> for the PR title, how about:
> 
> ```
> fix(webui): Cast as UNSIGNED to fix MySQL-specific type errors in dashboard stats queries.
> ```


* Cast what as unsigned?
* We should mention the issue number, right? Or is the original issue not fixed?

haiqi96:
> > for the PR title, how about:
> > ```
> > fix(webui): Cast as UNSIGNED to fix MySQL-specific type errors in dashboard stats queries.
> > ```
> 
> * Cast what as unsigned?
> * We should mention the issue number, right? Or is the original issue not fixed?

How about `Cast COALESCE result as UNSIGNED`

kirkrodrigues:
> How about `Cast COALESCE result as UNSIGNED`

It's actually the type of the arguments to `COALESCE`, right?

https://dev.mysql.com/doc/refman/8.4/en/comparison-operators.html#function_coalesce

> The return type of [COALESCE()](https://dev.mysql.com/doc/refman/8.4/en/comparison-operators.html#function_coalesce) is the aggregated type of the argument types.

So maybe it should be:

```
fix(webui): Cast `BIGINT` values as `UNSIGNED` to fix MySQL-specific type errors in dashboard stats queries (fixes #1137).
```

(Don't forget to add the `#` before the issue number.)

haiqi96:
> > How about `Cast COALESCE result as UNSIGNED`
> 
> It's actually the type of the arguments to `COALESCE`, right?
> 
> https://dev.mysql.com/doc/refman/8.4/en/comparison-operators.html#function_coalesce
> 
> > The return type of [COALESCE()](https://dev.mysql.com/doc/refman/8.4/en/comparison-operators.html#function_coalesce) is the aggregated type of the argument types.
> 
> So maybe it should be:
> 
> ```
> fix(webui): Cast `BIGINT` values as `UNSIGNED` to fix MySQL-specific type errors in dashboard stats queries (fixes #1137).
> ```
> 
> (Don't forget to add the `#` before the issue number.)

reading the code again, I still believe what's casted is the result of COALESCE? for example, for the second change, what's casted is `COALESCE(SUM(num_messages), 0)`.
But in either case, the suggested title looks good.

kirkrodrigues:
> reading the code again, I still believe what's casted is the result of COALESCE? for example, for the second change, what's casted is COALESCE(SUM(num_messages), 0).

Yeah, it is the result of COALESCE, but COALESCE's return type is the aggregated type of its arguments, right? And based on the database schema, the type of num_messages is BIGINT, no?

haiqi96:
> > reading the code again, I still believe what's casted is the result of COALESCE? for example, for the second change, what's casted is COALESCE(SUM(num_messages), 0).
> 
> Yeah, it is the result of COALESCE, but COALESCE's return type is the aggregated type of its arguments, right? And based on the database schema, the type of num_messages is BIGINT, no?

right, right. I think we are on the same page.

PR #1138, opened 2025-07-29
refactor(clp): Prepare parsing & search code for deduplication with clp-s:

gibber9809:
# Description

This PR is the first of several that is attempting to bring the core CLP parsing and search code to a state where it can be shared by both clp and clp-s. The full end to end integration can be seen [here](https://github.com/gibber9809/clp/tree/clp-core-refactor-prototype).

This initial PR does two things:
1. It changes `EncodedVariableInterpreter` to use templates instead of dictionary and dictionary entry implementations specific to CLP

This will end up allowing both clp and clp-s to share the same core logic while using their own respective dictionary implementations. These templates will be enhanced in a follow-up PR to use concepts in order to ensure dictionaries passed to these methods follow a certain interface.

2. We change several methods in `EncodedVariableInterpreter` and the various dictionary classes to accept `std::string_view` instead of `std::string const&`.

This allows us to avoid some string copies (more once combined with a follow-up PR that modifies `Grep`). It also makes the `EncodedVariableInterpreter` interface agree more with the ffi parsing interfaces.

As part of changing the dictionary writers to accept `std::string_view` we also switch their internal hashmap representation to `absl::flat_hash_map<std::string,...>`. The performance of flat hash map is a bit better than `std::unordered_map`, but the main advantage here is that absl has defined their hash functions such that `std::string_view` can be hashed in the same way as `std::string` to do lookups in a hashmap keyed by `std::string` (which allows us to avoid a string copy on the fast path), whereas with `std::unordered_map` we would have to provide our own hash function to accomplish the same thing which is error prone and forces us to use a more verbose template for `std::unordered_map`.

# Checklist

  breaking change.

# Validation performed

* Ensured that all unit tests pass
* Verified that after the entire sequence of planned PRs performance looks good

## Summary by CodeRabbit

* **New Features**
  * Improved efficiency and flexibility of variable encoding, decoding, and dictionary operations by introducing templated and generic implementations.
  * Enhanced error diagnostics during variable decoding and log message reconstruction.

* **Refactor**
  * Updated interfaces to use `std::string_view` for more efficient string handling.
  * Centralized variable placeholder methods into a single utility, simplifying and modernizing the API.
  * Replaced internal map types with more efficient alternatives for better performance.
  * Simplified variable encoding implementation by removing redundant methods and consolidating logic.
  * Updated query processing to use centralized variable handling methods for consistency.

* **Chores**
  * Updated build configuration to include necessary dependencies for improved performance and compatibility.
  * Added required header includes in test files for better code organization.

### DIFF
--- a/components/core/src/clp/DictionaryReader.hpp
+++ b/components/core/src/clp/DictionaryReader.hpp
@@ -1,7 +1,10 @@
 #ifndef CLP_DICTIONARYREADER_HPP
 #define CLP_DICTIONARYREADER_HPP
 
+#include <iterator>
 #include <string>
+#include <string_view>
+#include <utility>
 #include <vector>
 
 #include <boost/algorithm/string.hpp>
@@ -33,6 +36,9 @@ public:
         // Methods
         char const* what() const noexcept override { return "DictionaryReader operation failed"; }
     };
+
+    using dictionary_id_t = DictionaryIdType;
+    using entry_t = EntryType;
 
     // Constructors
     DictionaryReader() : m_is_open(false), m_num_segments_read_from_index(0) {
@@ -85,7 +91,7 @@ public:
      * @return a vector of matching entries, or an empty vector if no entry matches.
      */
     std::vector<EntryType const*>
-    get_entry_matching_value(std::string const& search_string, bool ignore_case) const;
+    get_entry_matching_value(std::string_view search_string, bool ignore_case) const;
     /**
      * Gets the entries that match a given wildcard string
      * @param wildcard_string
@@ -93,7 +99,7 @@ public:
      * @param entries Set in which to store found entries
      */
     void get_entries_matching_wildcard_string(
-            std::string const& wildcard_string,
+            std::string_view wildcard_string,
             bool ignore_case,
             std::unordered_set<EntryType const*>& entries
     ) const;
@@ -235,7 +241,7 @@ DictionaryReader<DictionaryIdType, EntryType>::get_value(DictionaryIdType id) co
 template <typename DictionaryIdType, typename EntryType>
 std::vector<EntryType const*>
 DictionaryReader<DictionaryIdType, EntryType>::get_entry_matching_value(
-        std::string const& search_string,
+        std::string_view search_string,
         bool ignore_case
 ) const {
     if (false == ignore_case) {
@@ -252,7 +258,11 @@ DictionaryReader<DictionaryIdType, EntryType>::get_entry_matching_value(
     }
 
     std::vector<EntryType const*> entries;
-    auto const search_string_uppercase = boost::algorithm::to_upper_copy(search_string);
+    std::string search_string_uppercase;
+    std::ignore = boost::algorithm::to_upper_copy(
+            std::back_inserter(search_string_uppercase),
+            search_string
+    );
     for (auto const& entry : m_entries) {
         if (boost::algorithm::to_upper_copy(entry.get_value()) == search_string_uppercase) {
             entries.push_back(&entry);
@@ -263,7 +273,7 @@ DictionaryReader<DictionaryIdType, EntryType>::get_entry_matching_value(
 
 template <typename DictionaryIdType, typename EntryType>
 void DictionaryReader<DictionaryIdType, EntryType>::get_entries_matching_wildcard_string(
-        std::string const& wildcard_string,
+        std::string_view wildcard_string,
         bool ignore_case,
         std::unordered_set<EntryType const*>& entries
 ) const {
--- a/components/core/src/clp/EncodedVariableInterpreter.hpp
+++ b/components/core/src/clp/EncodedVariableInterpreter.hpp
@@ -2,14 +2,18 @@
 #define CLP_ENCODEDVARIABLEINTERPRETER_HPP
 
 #include <string>
+#include <string_view>
+#include <unordered_set>
 #include <vector>
 
+#include "DictionaryConcepts.hpp"
+#include "ffi/ir_stream/decoding_methods.hpp"
 #include "ir/LogEvent.hpp"
 #include "ir/types.hpp"
 #include "Query.hpp"
+#include "spdlog_with_specializations.hpp"
 #include "TraceableException.hpp"
-#include "VariableDictionaryReader.hpp"
-#include "VariableDictionaryWriter.hpp"
+#include "type_utils.hpp"
 
 namespace clp {
 /**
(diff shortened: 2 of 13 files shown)


components/core/src/clp/LogTypeDictionaryEntry.hpp
```diff
@@ -1,6 +1,8 @@
 #ifndef CLP_LOGTYPEDICTIONARYENTRY_HPP
 #define CLP_LOGTYPEDICTIONARYENTRY_HPP
 
+#include <string>
```
haiqi96: It seems the string is not used in this header.

components/core/src/clp/LogTypeDictionaryEntry.hpp
```diff
@@ -137,10 +105,10 @@ class LogTypeDictionaryEntry : public DictionaryEntry<logtype_dictionary_id_t> {
      * @return true if another variable was found, false otherwise
      */
     bool parse_next_var(
-            std::string const& msg,
+            std::string_view msg,
             size_t& var_begin_pos,
             size_t& var_end_pos,
-            std::string& var
+            std::string_view& var
```
haiqi96, line 110: at first glance, It's not very intuitive to return a string_view by reference. 
I double checked the code and I think indeed will work, as the function updates the `var` so it reference to the correct porition of the string. so I guess I will approve this.

@kirkrodrigues just want to double check if you also agree.

components/core/src/clp/EncodedVariableInterpreter.hpp
```diff
@@ -47,14 +50,47 @@ class EncodedVariableInterpreter {
     // Methods
     static encoded_variable_t encode_var_dict_id(variable_dictionary_id_t id);
     static variable_dictionary_id_t decode_var_dict_id(encoded_variable_t encoded_var);
+
+    /**
+     * Adds a dictionary variable placeholder to the given logtype
+     * @param logtype
+     */
+    static void add_dict_var(std::string& logtype) {
+        logtype += enum_to_underlying_type(ir::VariablePlaceholder::Dictionary);
```
haiqi96: can we use append instead of +=?

components/core/src/clp/LogTypeDictionaryEntry.hpp
gibber9809: I do agree that it would be better to end up changing the interface (actually in a few places besides this as well), but my reasoning for keeping it mostly the same besides switching to string_view in this PR was to minimize how much call-sites for all of this code need to change to simplify the whole series of PRs.

components/core/src/clp/EncodedVariableInterpreter.hpp
gibber9809: Sure -- this was inherited from the old code, but may as well clean it up.

components/core/src/clp/EncodedVariableInterpreter.hpp
```diff
@@ -141,63 +194,359 @@ class EncodedVariableInterpreter {
      * dictionary
      * @return false otherwise
      */
+    template <
+            typename VariableDictionaryReaderType,
+            typename VariableDictionaryEntryType = VariableDictionaryReaderType::entry_t>
```
haiqi96: any plan to use a concept to enforce that VariableDictionaryReaderType have an entry_t type?

gibber9809: Yeah that's part of the concept (see [here](https://github.com/gibber9809/clp/blob/2bd2abdbd279cd7fca7fd9d78960e6cc1ecbe74a/components/core/src/clp/DictionaryConcepts.hpp#L252)).

haiqi96 (approved):
In general the change looks good to me. Locally compiled it and made sure unittests passed.

Given that there are more changes to come, I would prefer to approve this one instead of keep reviewing it (and get diminished return)

components/core/src/clp/LogTypeDictionaryEntry.hpp
kirkrodrigues: Yeah, seems fine.

components/core/src/clp/EncodedVariableInterpreter.hpp
```diff
@@ -77,16 +113,19 @@ class EncodedVariableInterpreter {
...
      * them to the variable dictionary if necessary)
+     * @tparam VariableDictionaryWriterType
+     * @tparam LogTypeDictionaryEntryType
      * @param message
      * @param logtype_dict_entry
      * @param var_dict
      * @param encoded_vars
      * @param var_ids
      */
+    template <typename VariableDictionaryWriterType, typename LogTypeDictionaryEntryType>
```
kirkrodrigues: Nit: Can we swap these two template params for consistency? (Also in the docstring and definition.)

components/core/src/clp/EncodedVariableInterpreter.hpp
kirkrodrigues: My IDE suggests:

```suggestion
            typename VariableDictionaryEntryType = typename VariableDictionaryReaderType::entry_t>
```

Since `entry_t` is a template type that's dependent on what `VariableDictionaryReaderType` is. I suppose the compiler could infer that since the lhs is a typename, the rhs must also be a typename, but I'm not sure. I'll leave it up to you based on your knowledge. Same for line 220.

components/core/src/clp/DictionaryReader.hpp
```diff
@@ -85,15 +91,15 @@ class DictionaryReader {
      * @return a vector of matching entries, or an empty vector if no entry matches.
      */
     std::vector<EntryType const*>
-    get_entry_matching_value(std::string const& search_string, bool ignore_case) const;
+    get_entry_matching_value(std::string_view search_string, bool ignore_case) const;
```
kirkrodrigues, line 94: There's a test case in `test-EncodedVariableInterpreter.cpp` that currently casts a string_view to a string to call this method. We can now get rid of those casts.

kirkrodrigues (commented):
For the PR title & body, how about:

```
refactor(clp): Prepare for deduplication with clp-s:

- Use templates in `EncodedVariableInterpreter` instead of CLP dictionary implementations.
- Use `std::string_view` where possible in `EncodedVariableInterpreter` and dictionary classes.
```

components/core/src/clp/EncodedVariableInterpreter.hpp
gibber9809: I think you're right -- I'll update it

PR #1143, opened 2025-07-30
refactor(clp): Prepare `Grep` code for deduplication with clp-s:

gibber9809:
# Description

This PR is the second in a series of PRs that is attempting to bring the core CLP parsing and search code to a state where it can be shared by both clp and clp-s. The full end to end integration can be seen [here](https://github.com/gibber9809/clp/tree/clp-core-refactor-prototype).

This PR focuses on splitting up `Grep` into the parts that will be shared (`GrepCore`) and the parts specific to CLP (left as `Grep`), and making the parts that will be shared independent of specific CLP implementations. In particular this PR:

1. Moves `process_raw_query` and associated utility functions to `GrepCore`

2. Modifies functions in `GrepCore` to use templates in order to remove dependencies on clp specific dictionary implementations -- this will be augmented with concepts in a follow-up PR in order to guarantee that the dictionaries follow a certain interface.

2. Moves the `QueryToken` class into its own files `QueryToken.{cpp,hpp}`

3. Refactors `Query` and `Subquery` classes to work in terms of dictionary ids (and segment ids) instead of pointers to dictionary entries in order to remove dependence on clp-specific dictionary implementations.

4. Refactors `process_raw_query` flow in a few ways
- Leaves adding implicit wildcards up to the caller. This makes the implementation better able to serve the needs of clp and clp-s, and makes it so that the interface shouldn't need to be changed if we interally switch to ffi-style subquery generation
- Moves segment index calculation outside of `process_raw_query` -- now callers that are interested in finding matches based on the segment index must call `calculate_ids_of_matching_segments` on the resulting `Query` object from `process_raw_query`. This simplifies the `process_raw_query` flow and allows callers that don't care about the segment index (clp-s) to not have to worry about the segment index calculation flow. 

# Checklist

  breaking change.

# Validation performed

* Validated that all unit tests pass
* Verified that after the entire sequence of planned PRs performance looks good 


components/core/src/clp/GrepCore.cpp
```diff
@@ -0,0 +1,273 @@
...
+                } else if (lexer.is_delimiter(c)) {
+                    // Found delimiter that's not also a wildcard
+                    break;
+                }
+            }
+        }
+
+        if (end_pos > begin_pos) {
+            bool has_prefix_wildcard = ('*' == value[begin_pos]) || ('?' == value[begin_pos]);
+            bool has_suffix_wildcard = ('*' == value[end_pos - 1]) || ('?' == value[begin_pos]);
```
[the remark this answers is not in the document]
gibber9809: This does indeed look like a bug, though I confirmed that it is copied exactly from the previous version of the code.

Maybe @haiqi96 can confirm that he also thinks this is a bug, then I'll fix it.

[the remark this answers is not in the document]
haiqi96: Yeah I agree it looks like a bug.

components/core/src/clp/QueryToken.cpp
```diff
@@ -0,0 +1,170 @@
+#include "QueryToken.hpp"
+
+#include <string>
+
+#include "Defs.h"
+#include "EncodedVariableInterpreter.hpp"
+
+namespace clp {
+QueryToken::QueryToken(
+        std::string const& query_string,
```
haiqi96: nit: you can use using std::string in this file since std::string is used multiple times.

components/core/src/clp/Query.hpp
```diff
@@ -36,32 +35,36 @@ class QueryVar {
...
-    bool m_is_precise_var;
-    bool m_is_dict_var;
+    bool m_is_precise_var{};
+    bool m_is_dict_var{};
 
-    encoded_variable_t m_precise_var;
+    encoded_variable_t m_precise_var{};
     // Only used if the precise variable is a dictionary variable
-    VariableDictionaryEntry const* m_var_dict_entry;
+    variable_dictionary_id_t m_var_dict_entry{};
```
haiqi96: since we don't store pointer to entries anymore, how about renaming them tom something like m_var_dict_id

haiqi96:
I feel that moving segment index calculation outside the procress_raw_query is not the best design, especially know that developer needs to explicitly call calculate_ids_of_matching_segments, but I see it is because we want to refactor only the common logic into GrepCore. 

In addition, I believe that our long-term goal is to drop segment support in CLP as well, so l would say the change proposed in this PR is fine.

components/core/tests/test-Grep.cpp
haiqi96, line 1: shall we rename this to be test-GrepCore.cpp?

haiqi96 (approved):
Did my best effort review and the change looks good as most of them copying and paste with adding template parameters.

components/core/src/clp/StringReader.cpp
```diff
@@ -7,8 +7,6 @@
 #include <cassert>
 #include <cerrno>
 
-#include <boost/filesystem.hpp>
```
kirkrodrigues: This is unrelated to the PR, right?

components/core/src/clp/clo/CMakeLists.txt
```diff
@@ -57,6 +59,8 @@ set(
         ../Platform.hpp
         ../Profiler.cpp
         ../Profiler.hpp
+        ../QueryToken.cpp
+        ../QueryToken.hpp
```
kirkrodrigues: Should be after Query.{c,h}pp

components/core/src/clp/clg/clg.cpp
```diff
@@ -205,8 +212,11 @@ static bool search(
         std::set<segment_id_t> ids_of_segments_to_search;
         bool is_superseding_query = false;
         for (auto const& search_string : search_strings) {
-            auto query_processing_result = Grep::process_raw_query(
-                    archive,
+            auto const& log_dict{archive.get_logtype_dictionary()};
```
kirkrodrigues: In some places in the PR, we're using `log` instead of `logtype`. I think that might lead to confusion at some point. Can we update them to `logtype`?

components/core/src/clp/Query.cpp
```diff
@@ -65,20 +68,24 @@ bool QueryVar::matches(encoded_variable_t var) const {
...
     if (m_is_precise_var) {
         auto& ids_of_segments_containing_query_var
-                = m_var_dict_entry->get_ids_of_segments_containing_entry();
+                = get_segments_containing_var_dict_id(m_var_dict_id);
         inplace_set_intersection(ids_of_segments_containing_query_var, segment_ids);
     } else {
         set<segment_id_t> ids_of_segments_containing_query_var;
-        for (auto entry : m_possible_var_dict_entries) {
-            auto& ids_of_segments_containing_var = entry->get_ids_of_segments_containing_entry();
+        for (auto entry : m_possible_var_dict_ids) {
```
kirkrodrigues: var_dict_id?

components/core/src/clp/Query.cpp
```diff
@@ -202,4 +185,18 @@ void Query::make_sub_queries_relevant_to_segment(segment_id_t segment_id) {
     }
     m_prev_segment_id = segment_id;
 }
+
+void Query::calculate_ids_of_matching_segments(
+        std::function<std::set<segment_id_t> const&(logtype_dictionary_id_t)>
+                get_segments_containing_log_dict_id,
+        std::function<std::set<segment_id_t> const&(variable_dictionary_id_t)>
+                get_segments_containing_var_dict_id
```
kirkrodrigues: For all these function pointers, clang-tidy warns that we should use a const reference. Can we do so?

components/core/src/clp/Query.hpp
```diff
@@ -80,45 +83,47 @@ class SubQuery {
...
     void clear();
 
     bool wildcard_match_required() const { return m_wildcard_match_required; }
 
     size_t get_num_possible_logtypes() const { return m_possible_logtype_ids.size(); }
 
-    std::unordered_set<LogTypeDictionaryEntry const*> const& get_possible_logtype_entries() const {
-        return m_possible_logtype_entries;
+    std::unordered_set<logtype_dictionary_id_t> const& get_possible_logtypes() const {
+        return m_possible_logtype_ids;
```
kirkrodrigues: Shall we rename this to `m_possible_logtypes` to match (and since we don't have `m_possible_logtype_entries` anymore)?

components/core/src/clp/Query.hpp
```diff
@@ -139,13 +144,14 @@ class SubQuery {
      * Whether the given variables contain the subquery's variables in order (but not necessarily
      * contiguously)
      * @param vars
+     * @tparam EncodedVariableContainerType A random access list of `clp::encoded_variable_t`.
```
kirkrodrigues:
```suggestion
     * @tparam EncodedVariableContainerType A random access list of `clp::encoded_variable_t`.
     * @param vars
```

components/core/src/clp/QueryToken.hpp
```diff
@@ -0,0 +1,66 @@
...
+    size_t m_begin_pos{};
+    size_t m_end_pos{};
+    std::string m_value;
+
+    // Type if variable has unambiguous type
+    Type m_type{};
+    // Types if variable type is ambiguous
+    std::vector<Type> m_possible_types;
+    // Index of the current possible type selected for generating a subquery
+    size_t m_current_possible_type_ix{};
```
kirkrodrigues: I guess the addition of braces is to resolve clang-tidy violations? If so, is the reason to fix these but not others just since these are small enough to not majorly affect the diff?

I would still explicitly set a default for enum types like `Type`.

components/core/src/clp/GrepCore.hpp
```diff
@@ -0,0 +1,424 @@
...
+
+namespace clp {
+class GrepCore {
+public:
+    // Methods
+    /**
+     * Processes a raw user query into a Query.
+     *
+     * Note: callers are responsible for ensuring that the search string does not contain repeated
+     * wildcards "**" e.g. by using `clp::string_utils::clean_up_wildcard_search_string`.
```
kirkrodrigues:
```suggestion
     * NOTE: Callers are responsible for ensuring that the search string is "clean", where clean is
     * defined by the criteria in `clean_up_wildcard_search_string`.
```

It's usually clearer and more consistent to use `NOTE` rather than `Note` (similar to `TODO`).

components/core/src/clp/StringReader.cpp
gibber9809: Sort of -- it doesn't have to be part of this PR but it does make it so that clp-s doesn't need to pull in `boost::filesystem` into the library that `GrepCore` will get pulled into in the follow-up PR to integrate with clp-s.

kirkrodrigues: Gotcha. Let's throw it in the follow-up PR then since it sounds necessary for that one but not this one.

components/core/src/clp/QueryToken.hpp
gibber9809: That and it made it easier for me to reason about the code knowing that things would be default initialized how I would expect. I'll explicitly set a default for `Type` as well.

gibber9809:
> What's the body of the PR going to be? (Usually I put that as the first lines of the PR body (before the description).

I updated the PR description with what I'd put as the PR body @kirkrodrigues.

kirkrodrigues:
Edited the PR body directly (you can check the edit history for details).

PR #1150, opened 2025-08-01
feat(webui): Dump query results from Presto into MongoDB.

davemarco:
# Description

Draft: TBD

# Checklist

  breaking change.

# Validation performed


components/webui/server/src/routes/api/presto-search/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+    row: unknown[],
+    columns: {name: string}[]
+): Record<string, unknown> => {
+    const obj: Record<string, unknown> = {};
+    columns.forEach((col, idx) => {
+        obj[col.name] = row[idx];
+    });
+
+    return obj;
+};
```
hoophalab, line 25: I feel this function provides little abstraction. It's best to put the content inside
```
        const resultDocs = data.map((row) => prestoRowToObject(row, columns));
```

What's your opinion?

components/webui/server/src/routes/api/presto-search/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    if (0 < data.length && searchJobId) {
+        const collection = mongoDb.collection(searchJobId);
+        const resultDocs = data.map((row) => prestoRowToObject(row, columns));
+        collection.insertMany(resultDocs).catch((err: unknown) => {
+            log.error(err, "Failed to insert Presto results into MongoDB");
+        });
+    }
+};
```
hoophalab: This function is just taking all the variables from the data callback as its parameters.

It would be much better to just inline it.

If we really want to keep it, it needs a few changes:
1. don't pass in `isResolved`. That error check should happen directly inside the `data` callback.
2. don't pass in `log`. This function can return `insertMany`'s promise directly.

components/webui/server/src/routes/api/presto-search/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+    mongoDb,
+    searchJobId,
+}: InsertPrestoRowsToMongoProps): void => {
+    if (false === isResolved) {
+        log.error("Presto data received before searchJobId was resolved; skipping insert.");
+
+        return;
+    }
+
+    if (0 < data.length && searchJobId) {
```
hoophalab: There's no need to check searchJobId here. In the previous code, it was a Nullable type. However, since we've already established that searchJobId will always be resolved before the data arrives, checking for undefined is redundant.

components/webui/server/src/routes/api/presto-search/utils.ts
davemarco: ur right dosent provide much functionality, but i thought the code was difficult to read/understand, so i felt wrapping it in a function and adding jsdoc would make more readable

components/webui/server/src/routes/api/presto-search/utils.ts
davemarco: okay i kept function but made changes. I want to keep the utility since i feel like this function will get more complicated over time, and want to keep the main route cleaner

components/webui/server/src/routes/api/presto-search/utils.ts
davemarco: okay its gone, and i moved this check outside the utility

components/webui/server/src/routes/api/presto-search/typings.ts
```diff
@@ -0,0 +1,14 @@
...
+
+
+interface InsertPrestoRowsToMongoProps {
+    data: unknown[][];
+    columns: {name: string}[];
+    searchJobId: string;
+    isResolved: boolean;
+    mongoDb: Db;
+    log: FastifyBaseLogger;
+}
```
hoophalab: this interface isn't updated, and we probably don't need this interface since there are less than 5 arguments.

davemarco:
@junhaoliao can you approve this

junhaoliao (approved):
Deferring to @hoophalab ‘s review

PR #1152, opened 2025-08-03
refactor(package): Standardize how configuration and credentials are passed to containers (resolves #1149).

### LINKED ISSUE #1149: Standardize credentials and configuration passing for Python components
### Request

Currently, the `start_clp.py` script uses inconsistent patterns for passing configuration and 
credentials to containers, leading to complex merging logic and security concerns. The main issues 
are:

1. Ad-hoc configuration merging: Many components receive a temporary YAML file that is a merged 
   result of parts of `clp-config.yml` and `credentials.yml`. This requires logic within the 
   `start_clp.py` script to perform the merge and generates numerous temporary files.
2. Non-uniform configuration subsets & credential handling: The approach differs across components.
   - `clp-db-creator` receives only a specific section (e.g., a `db-config.yml` with all fields at
     the top-level from just the `database` section of `clp-config.yml`) rather than a standard 
     full configuration object.
   - Credential passing is inconsistent: some components as mentioned above get merged credentials,
     while `clp-webui` already use environment variables.

We shall standardize configuration and credential delivery to improve consistency and security by:
- Separating credentials from general configuration.
   - Instead, passing all credentials via environment variables to avoid leaking them in temporary files.
- Using a single generated configuration file for all non-credential configuration data. (instead of generating config files per component)

### Possible implementation

1. Modify `CLPConfig` class to include a `generated_config_file_path` field to store the path to the generated config file.

2. Update `start_clp.py` script to:
   - Generate a single `clp-config-generated.yml` file containing all non-credential configuration
   - Pass database credentials via environment variables (`CLP_DB_USER`, `CLP_DB_PASS`) instead of embedding them in temporary files

3. Standardize configuration passing for Python components:
   1. Modify the `clp-db-table-creator` components to:
      - Read their configuration from the generated `clp-config-generated.yml` file instead of separate db-config files
      - Load database credentials from environment variables
   2. Modify the following Python components in `start_clp` to:
      - Load their non-credential configuration from `clp-config-generated.yml`
      - Read credentials from environment variables:
     - `clp-db-table-creator` (composed of `components/clp-py-utils/clp_py_utils/create-db-tables.py`, `initialize-clp-metadata-db.py`, and `initialize-orchestration-db.py`)
     - `clp-compression-scheduler` (`components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py`)
     - `clp-query-scheduler` (`components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py`)
     - `clp-compression-worker` (`components/job-orchestration/job_orchestration/executor/compress/compression_task.py`)
     - `clp-query-worker` (`components/job-orchestration/job_orchestration/executor/query/fs_search_task.py` and `extract_stream_task.py`)
     - `clp-reducer` (`components/job-orchestration/job_orchestration/reducer/reducer.py`)

4. Update all other affected Python components and their wrappers to read credentials from environment variables rather than from configuration files, which includes:
   - `compress.sh`
   - `decompress.sh`
   - `search.sh`
   - `admin-tools/archive-manager.sh`

junhaoliao:
# Description

This PR standardizes how configuration and credentials are passed to containers, addressing security concerns and code complexity in the `start_clp.py` script.

The key changes are:
Standardize configuration and credential delivery to improve consistency and security by:
- Separating credentials from general configuration.
   - Instead, passing all credentials via environment variables to avoid leaking them in temporary files.
- Using a single generated configuration file for all non-credential configuration data. (instead of generating config files per component)

See #1149 for more details.

# Checklist

  breaking change.

# Validation performed

## Code / File inspection
1. Inspected that all components properly load credentials from environment variables.
2. Confirmed that the single generated configuration file contains all necessary non-credential configuration data

## Manual tests
1. Tested container startup with the new configuration approach:
   - `cd clp-package/sbin/`
   - `./start-clp.sh`
2. Tested wrapper scripts:
   - `./compress.sh ~/samples/hive-24hr/`
   - `./decompress.sh x`
   - `./search.sh "1"`
   - `./admin-tools/archive-manager.sh find`
3. Performed a search with query string "1" in the webui and observed results returned successfully.
4. Enabled `garbage_collector` and observed archives / search results are purged in below configurations:

    1. `results_cache.retention_period = 1` + `archive_output.retention_period = 5800000` + performed a search: observed search results were cleaned in around one minute after query completed.
    2. `results_cache.retention_period = 1` + `archive_output.retention_period = 1` + performed a search: observed no search result; archives were cleaned in around one minute after compression.
    ```yaml
    results_cache:
      # Retention period for search results, in minutes. Set to null to disable automatic deletion.
      retention_period: 1
    
    ## Where archives should be output to
    archive_output:
      # Retention period for archives, in minutes. Set to null to disable automatic deletion.
      retention_period: 1 # 5800000 for validation with the result cache 
    
    ## Garbage collector config
    garbage_collector:
      logging_level: "INFO"
    
      # Interval (in minutes) at which garbage collector jobs run
      sweep_interval:
        archive: 1
        search_result: 1
    ```

## Summary by CodeRabbit

* **New Features**
  * Inject extra environment variables into container startup; add shared generated container config file and mount support; new helpers to construct common, credential and connection env lists.

* **Bug Fixes**
  * Load DB/queue/Redis credentials from environment when present; omit sensitive fields from serialized exports; improved error logging with tracebacks.

* **Documentation**
  * Added -c short alias and clarified --config help text.

* **Refactor**
  * Centralized container config dumping and env/credential handling; standardized generated config filename usage.


components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -36,6 +37,8 @@
 )
 from strenum import KebabCaseStrEnum
 
+logger = logging.getLogger(__file__)
```
haiqi96: Looks like we end up not using this?

btw, I believe we prefer to not log in the library-like method unless we really can't work around it.

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -278,6 +282,19 @@ def generate_container_config(
...
+        CONTAINER_CLP_HOME,
+        clp_config.generated_config_file_path,
+        container_clp_config.generated_config_file_path,
+    ):
+        docker_mounts.generated_config_dir = DockerMount(
+            DockerMountType.BIND,
+            clp_config.generated_config_file_path,
+            container_clp_config.generated_config_file_path,
+        )
+
```
haiqi96: I would suggest we reorder this with `aws_config_directory`, since `aws_config_directory` is the optional mount. (that said, input logs is also optional but we put it to the first... so I guess my suggestion is just a personal preference)
If you agree to do so, please also reorder the member declaration.

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -313,11 +330,15 @@ def dump_container_config(
 
 
 def generate_container_start_cmd(
-    container_name: str, container_mounts: List[Optional[DockerMount]], container_image: str
+    container_name: str,
+    extra_env_vars: dict[str, str],
```
haiqi96: let's use Dict instead. you would also need to import Dict from typings. 
Also wonder we put it as the second arguments in between the container configs? maybe put it to the end of the argument list?

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -590,3 +618,59 @@ def validate_dataset_name(clp_table_prefix: str, dataset_name: str) -> None:
...
+
+
+def generate_common_environment_variables(
+    container_clp_config,
+    include_clp_home=True,
+    include_db_credentials=False,
+    include_queue_credentials=False,
+    include_redis_credentials=False,
+    include_celery_connection_params=False,
+    extra_vars=None,
```
haiqi96: I personally feel that `extra_vars` is contradicting with the `common` in the function name.

Imo, the env_vars can be divided into 3 categories:
1. The true "common" one: `PYTHONPATH` and `CLP_HOME`
2. Credentials.
3. caller specific vars, which is the extra_var here.

I would consider creating separate method for 1 and 2. for now, it can be an overkill, but If in the future there are more common envvars then we can see more value.

What I can think of:
```
def generate_common_environment_variables(container_clp_config) -> List[str]
// Note: for generate_credentials_to_env_vars, I don't think the clp config has to be `container` specific.
def generate_credentials_to_env_vars(clp_config, flags..) -> List[str]

env_vars = generate_common_environment_variables(container_clp_config)
// If credential is needed.
env_vars.extend(generate_credentials_to_env_vars(container_clp_config, flags...))
...

```

For extra_vars, we can remove the `extra_vars` from this method, can the caller can append any specific env_vars to the returned vector of this method.

components/clp-package-utils/clp_package_utils/scripts/archive_manager.py
```diff
@@ -206,14 +206,17 @@ def main(argv: typing.List[str]) -> int:
     generated_config_path_on_container, generated_config_path_on_host = dump_container_config(
         container_clp_config, clp_config, container_name
     )
-
+    extra_env_vars = {
+        "CLP_DB_USER": clp_config.database.username,
+        "CLP_DB_PASS": clp_config.database.password,
+    }
```
haiqi96: If we have separate method for credentials and common_env_var, maybe we can utilize the credential_method?

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -590,3 +618,59 @@ def validate_dataset_name(clp_table_prefix: str, dataset_name: str) -> None:
             f"Invalid dataset name: `{dataset_name}`. Names can only be a maximum of"
             f" {dataset_name_max_len} characters long."
         )
+
+
+def generate_common_environment_variables(
+    container_clp_config,
```
haiqi96: Don't forget about the annotation.
container_clp_config: CLPConfig

I would also consider renaming the method since the env_vars are specific to containers.

components/clp-package-utils/clp_package_utils/general.py
haiqi96: since it has "extra' in the name, shall this be an optional as well?

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
```diff
@@ -112,11 +113,18 @@ def handle_extract_stream_cmd(
...
     )
     if clp_config is None:
         return -1
+    # Override credentials with environment variables
+    try:
+        clp_config.database.username = os.environ["CLP_DB_USER"]
+        clp_config.database.password = os.environ["CLP_DB_PASS"]
+    except KeyError as e:
+        logger.error(f"Missing environment variable: {e}")
+        return -1
```
haiqi96: I feel we can put this logic into native/utils.py since I see it duplicated for a few times.
Note this would require us to log within a method in native/utils.py, there is an [example](https://github.com/y-scope/clp/blob/main/components/job-orchestration/job_orchestration/executor/utils.py#L9) of how we do similar things but I would like to confirm with Kirk.

@kirkrodrigues Can we create a helper method in utils which takes in a logger as argument?

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -748,23 +730,17 @@ def generic_start_worker(
...
+        include_clp_home=True,
+        include_celery_connection_params=True,
+        extra_vars=[
+            f"CLP_CONFIG_PATH={container_clp_config.generated_config_file_path}",
+            f"CLP_LOGS_DIR={container_logs_dir}",
+            f"CLP_LOGGING_LEVEL={worker_config.logging_level}",
+            f"CLP_WORKER_LOG_PATH={container_worker_log_path}",
+        ],
+    )
     necessary_mounts = [
```
haiqi96: unrelated to this change, but I feel maybe each clp_config object should have their own get_necssary_mounts().

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -31,6 +31,7 @@
 )
 from clp_py_utils.sql_adapter import SQL_Adapter
 from job_orchestration.executor.compress.celery import app
+from job_orchestration.executor.utils import load_worker_config
```
haiqi96: this is not used. 
I guess you intended to replace code at [line503](https://github.com/y-scope/clp/pull/1152/files#diff-50c45493d448043133c73d8bcb83e6aea018ac8fc568c7e7c7dc3993d22e36d6R503-R506) with load_worker_config.

worker config was created because we didn't want to include more-than-necessary information for worker. but since now you are moving sensitive information to EnvVar, it's likely we can also get rid of WorkerConfig and use the common generated config.

If you don't have time, we should at least create an issue for this.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1134,7 +1134,8 @@ async def main(argv: List[str]) -> int:
     config_path = pathlib.Path(parsed_args.config)
     try:
         clp_config = CLPConfig.parse_obj(read_yaml_config_file(config_path))
-    except ValidationError as err:
+        clp_config.load_database_credentials_from_env()
+    except (ValidationError, ValueError) as err:
```
haiqi96, line 1158: line 1142 is outside of this PR, but let's be consistent with compression_scheduler and make this to be logger.exception() as well?

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -404,11 +404,12 @@ def main(argv):
...
     try:
         clp_config = CLPConfig.parse_obj(read_yaml_config_file(config_path))
-    except ValidationError as err:
+        clp_config.load_database_credentials_from_env()
+    except (ValidationError, ValueError) as err:
         logger.error(err)
         return -1
     except Exception as ex:
-        logger.error(ex)
+        logger.exception(ex)
```
haiqi96: As discussed offline, we can replace this with

`logger.exception(<custom error message>)`

ditto for other similar cases.

components/clp-py-utils/clp_py_utils/initialize-orchestration-db.py
```diff
@@ -36,14 +39,23 @@ def main(argv):
...
-        if database_config is None:
-            raise ValueError(f"Database configuration file '{parsed_args.config}' is empty.")
-        sql_adapter = SQL_Adapter(database_config)
+        clp_config = CLPConfig.parse_obj(read_yaml_config_file(config_path))
+        clp_config.load_database_credentials_from_env()
+    except (ValidationError, ValueError) as err:
+        logger.error(err)
+        return -1
+    except Exception as ex:
+        logger.exception(ex)
```
haiqi96: ditto

components/clp-py-utils/clp_py_utils/initialize-clp-metadata-db.py
```diff
@@ -39,16 +41,23 @@ def main(argv):
...
+    # Load configuration
+    config_path = pathlib.Path(parsed_args.config)
+    try:
+        clp_config = CLPConfig.parse_obj(read_yaml_config_file(config_path))
+        clp_config.load_database_credentials_from_env()
+    except (ValidationError, ValueError) as err:
+        logger.error(err)
+        return -1
+    except Exception as ex:
+        logger.exception(ex)
```
haiqi96: ditto

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -757,6 +762,26 @@ def load_redis_credentials_from_file(self):
...
+            self.database.password = os.environ["CLP_DB_PASS"]
+        except KeyError as ex:
+            raise ValueError(f"Missing environment variable: {ex}")
+
+    def load_queue_credentials_from_env(self):
+        try:
+            self.queue.username = os.environ["CLP_QUEUE_USER"]
+            self.queue.password = os.environ["CLP_QUEUE_PASS"]
+        except KeyError as ex:
+            raise ValueError(f"Missing environment variable: {ex}")
```
haiqi96: maybe we can create a resuable method for this
```
def get_env_var(name) -> str
        try:
            return = os.environ[name]
        except KeyError as ex:
            raise ValueError(f"Missing environment variable: {ex}")
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -608,6 +609,7 @@ class CLPConfig(BaseModel):
     data_directory: pathlib.Path = pathlib.Path("var") / "data"
     logs_directory: pathlib.Path = pathlib.Path("var") / "log"
     aws_config_directory: Optional[pathlib.Path] = None
+    generated_config_file_path: pathlib.Path = logs_directory / ".clp-config.yml"
```
haiqi96: if we have this as a member variable, it also means user can overwrite this file, which seems unnecessary?
I also feel it's a little awkward that you need to insert a "host" into the path at some point.

What if we don't store this variable, but instead create a getter_method for this generated_config_file_path (say, `get_generated_config_file_path()`)

```
def get_generated_config_file_path(self) -> pathlib.Path:
    return self.logs_directory/".clp-config.yml"
```

haiqi96 (commented):
Best effort review for now

components/clp-package-utils/clp_package_utils/general.py
junhaoliao: > Also wonder we put it as the second arguments in between the container configs? maybe put it to the end of the argument list?

i meant to order the args based on their usages but that may not make sense. i agree it's better to put the "extra"_env_vars at the end of the arg list. 

would it be even better if we make it optional? e.g. `[]` as default

junhaoliao: please ignore my last comment. i just saw https://github.com/y-scope/clp/pull/1152/files#r2258207729

all good then

components/clp-package-utils/clp_package_utils/scripts/archive_manager.py
junhaoliao: the extracted method still generates env vars in `KEY=VALUE` format so it doesn't fit perfectly here

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
junhaoliao: pending @kirkrodrigues 's comment

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -605,40 +603,34 @@ def generic_start_scheduler(
...
-            f"{container_clp_config.redis.host}:{container_clp_config.redis.port}/"
-            f"{container_clp_config.redis.query_backend_database}"
+    env_vars = [
+        *generate_common_environment_variables(),
+        *generate_credential_environment_variables(
+            container_clp_config, include_db_credentials=True
         ),
+        *generate_celery_connection_environment_variables(container_clp_config),
         f"CLP_LOGS_DIR={container_logs_dir}",
         f"CLP_LOGGING_LEVEL={clp_config.query_scheduler.logging_level}",
```
junhaoliao, line 615: seems we have a misconfiguration in the original code where a compression scheduler's log level is set with the configured query scheduler's log level.

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -356,8 +353,9 @@ def generate_container_start_cmd(
         "--name", container_name,
         "--log-driver", "local"
     ]
-    for key, value in extra_env_vars.items():
-        container_start_cmd.extend(["-e", f"{key}={value}"])
+    if extra_env_vars:
```
haiqi96: I believe we prefer to use explicit check, so let's use 
```suggestion
    if extra_env_vars is not None:
```

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -925,37 +926,37 @@ def start_webui(
...
         else:
             aws_mount, aws_env_vars = generate_container_auth_options(
                 clp_config, WEBUI_COMPONENT_NAME
             )
             if aws_mount:
                 necessary_mounts.append(mounts.aws_config_dir)
-            if aws_env_vars:
-                necessary_env_vars.extend(aws_env_vars)
+                if aws_env_vars:
+                    env_vars.extend(aws_env_vars)
```
haiqi96: aws_env_vars and aws_mount are independent (one for profile and one for env_var credentials), so they shouldn't be nested.
```suggestion
            if aws_env_vars:
                env_vars.extend(aws_env_vars)
```

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
haiqi96: @junhaoliao after this PR gets merged, can you ask rabbit to create a reminder issue so we can check if we still need worker_config

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -270,6 +272,18 @@ def generate_container_config(
...
+        CONTAINER_CLP_HOME,
+        clp_config.generated_config_file_path,
+        container_clp_config.generated_config_file_path,
+    ):
+        docker_mounts.generated_config_dir = DockerMount(
+            DockerMountType.BIND,
+            clp_config.generated_config_file_path,
+            container_clp_config.generated_config_file_path,
+        )
+
```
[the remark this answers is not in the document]
haiqi96: We are using --mount which seems to work for file.

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -590,3 +617,71 @@ def validate_dataset_name(clp_table_prefix: str, dataset_name: str) -> None:
             f"Invalid dataset name: `{dataset_name}`. Names can only be a maximum of"
             f" {dataset_name_max_len} characters long."
         )
+
+
+def generate_common_environment_variables(
+    include_clp_home=True,
+) -> List[str]:
```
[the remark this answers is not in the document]
haiqi96: Can you check the code again?  i don't think we are calling generate_common_environment_variables with container_clp_config

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -94,6 +94,7 @@ def __init__(self, clp_home: pathlib.Path, docker_clp_home: pathlib.Path):
         self.archives_output_dir: typing.Optional[DockerMount] = None
         self.stream_output_dir: typing.Optional[DockerMount] = None
         self.aws_config_dir: typing.Optional[DockerMount] = None
+        self.generated_config_dir: typing.Optional[DockerMount] = None
```
haiqi96: I guess first, this should be `generated_config_file` instead of dir.

And I guess it might be better to just mount the parent of generated_config_file? (so mount a dir).
In our current code, it won't do anything since generated_config_file.parent would be the same as logs_dir, but it can prevent some silly future error.

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -278,6 +291,7 @@ def generate_container_config(
             clp_config.aws_config_directory,
             container_clp_config.aws_config_directory,
         )
+
```
haiqi96: I am fine with a new line, but in case if it's not intended.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -770,6 +793,14 @@ def dump_to_primitive_dict(self):
             d["aws_config_directory"] = str(self.aws_config_directory)
         else:
             d["aws_config_directory"] = None
+
+        # Remove sensitive information
+        d["database"].pop("username", None)
+        d["database"].pop("password", None)
+        d["queue"].pop("username", None)
+        d["queue"].pop("password", None)
+        d["redis"].pop("username", None)
```
haiqi96: Sorry for another comment, how about we create a `dump_to_primitive_dict() `for database, queue and redis, and we remove the password/username in these methods?
I feel this would be less error prone than having a separate block.

haiqi96: Another thing I thought about is to override the .dict() method of those three classes, but that would make it harder to developer to debug (since without reading the code, someone may not expect .dict() omits certain fields), so dump_to_primitive_dict() is safer.

components/clp-package-utils/clp_package_utils/general.py
junhaoliao: i agree mounting the parent directory is better because Docker's binding mount would create a directory when the specified path doesn't exist, causing issues in debugging.

However, i just tested that `is_path_already_mounted()` doesn't behave as expected and we can end up requesting to mount the logs_dir twice, which gives below error:
```
docker: Error response from daemon: Duplicate mount point: /home/junhao/workspace/clp/build/clp-package/var/log/GIGABYTE
...
subprocess.CalledProcessError: Command '['docker', 'run', '-i', '--network', 'host', '--rm', '--name', 'clp-database-table-creator-8367', '--log-driver', 'local', '-u', '1000:1000', '--mount', 'type=bind,src=/home/junhao/workspa
ce/clp/build/clp-package,dst=/opt/clp', '--mount', 'type=bind,src=/home/junhao/workspace/clp/build/clp-package/var/data/GIGABYTE,dst=/home/junhao/workspace/clp/build/clp-package/var/data/GIGABYTE', '--mount', 'type=bind,src=/hom
e/junhao/workspace/clp/build/clp-package/var/log/GIGABYTE,dst=/home/junhao/workspace/clp/build/clp-package/var/log/GIGABYTE', '--mount', 'type=bind,src=/home/junhao/workspace/clp/build/clp-package/var/log/GIGABYTE,dst=/home/junh
ao/workspace/clp/build/clp-package/var/log/GIGABYTE', '-e', 'PYTHONPATH=/opt/clp/lib/python3/site-packages', '-e', 'CLP_HOME=/opt/clp', '-e', 'CLP_DB_USER=clp-user', '-e', 'CLP_DB_PASS=WuWM7ka3J70', 'ghcr.io/y-scope/clp/clp-exec
ution-x86-ubuntu-jammy:main', 'python3', '/opt/clp/lib/python3/site-packages/clp_py_utils/create-db-tables.py', '--config', '/home/junhao/workspace/clp/build/clp-package/var/log/GIGABYTE/.clp-config.yml', '--storage-engine', 'clp']' returned non-zero exit status 125
```

maybe we should just correct the name of this field to be `generated_config_file`?

components/clp-package-utils/clp_package_utils/general.py
junhaoliao: let me revert since it's not strictly related to the PR

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -605,40 +603,34 @@ def generic_start_scheduler(
...
-            f"{container_clp_config.redis.query_backend_database}"
+    env_vars = [
+        *generate_common_environment_variables(),
+        *generate_credential_environment_variables(
+            container_clp_config, include_db_credentials=True
         ),
+        *generate_celery_connection_environment_variables(container_clp_config),
         f"CLP_LOGS_DIR={container_logs_dir}",
         f"CLP_LOGGING_LEVEL={clp_config.query_scheduler.logging_level}",
     ]
```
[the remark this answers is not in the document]
junhaoliao: will fix in a future PR

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -748,19 +734,10 @@ def generic_start_worker(
...
-        ),
-        (
-            f"RESULT_BACKEND=redis://default:{container_clp_config.redis.password}@"
-            f"{container_clp_config.redis.host}:{container_clp_config.redis.port}/{redis_database}"
-        ),
-        f"CLP_HOME={CONTAINER_CLP_HOME}",
-        f"CLP_CONFIG_PATH={container_clp_config.logs_directory / container_config_filename}",
+    env_vars = [
+        *generate_common_environment_variables(include_clp_home=True),
+        *generate_celery_connection_environment_variables(container_clp_config),
```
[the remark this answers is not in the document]
junhaoliao: adding the explicit argument would make the code more readable, no?

components/clp-package-utils/clp_package_utils/general.py
junhaoliao: for now, i just renamed it to `generated_config_file`

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
junhaoliao: @coderabbit let's create an issue to track the request to remove the `WorkerConfig` class because it will no longer be needed after the changes in this current PR

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -158,6 +159,15 @@ def get_clp_connection_params_and_type(self, disable_localhost_socket_connection
...
 
+    def dump_to_primitive_dict(self):
+        d = self.dict()
+
+        # Remove sensitive information
+        d.pop("username", None)
+        d.pop("password", None)
+
+        return d
+
```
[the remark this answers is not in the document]
haiqi96: perhaps we can create a new class for any dataclass with password and username. but I would say let's not putting everything in this PR (unless @junhaoliao have time).

haiqi96 (dismissed):
Looks good to me.

for title, credential  -> credentials ?

And don't forget to make sure linter also passes.

components/clp-package-utils/clp_package_utils/scripts/archive_manager.py
```diff
@@ -206,14 +206,17 @@ def main(argv: typing.List[str]) -> int:
     generated_config_path_on_container, generated_config_path_on_host = dump_container_config(
         container_clp_config, clp_config, container_name
     )
-
     necessary_mounts: typing.List[CLPDockerMounts] = [
```
kirkrodrigues: While we're in the area:

```suggestion
    necessary_mounts: typing.List[DockerMount] = [
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -590,3 +610,71 @@ def validate_dataset_name(clp_table_prefix: str, dataset_name: str) -> None:
...
+
+def generate_common_environment_variables(
+    include_clp_home=True,
+) -> List[str]:
+    """
+    Generate a list of common environment variables for Docker containers.
+
+    :param include_clp_home:
```
kirkrodrigues: How about `include_clp_home_env_var`? Otherwise, this name is a little vague (e.g, maybe it means all env vars should be relative to clp_home).

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -590,3 +610,71 @@ def validate_dataset_name(clp_table_prefix: str, dataset_name: str) -> None:
...
+
+
+def generate_credential_environment_variables(
+    container_clp_config: CLPConfig,
+    include_db_credentials=False,
+    include_queue_credentials=False,
+    include_redis_credentials=False,
+) -> List[str]:
+    """
+    Generate a list of credential environment variables for Docker containers.
```
```suggestion
    Generates a list of credential environment variables for Docker containers.
```

components/clp-py-utils/clp_py_utils/initialize-clp-metadata-db.py
```diff
@@ -1,14 +1,16 @@
 #!/usr/bin/env python3
 import argparse
 import logging
+import os
```
kirkrodrigues:
```suggestion
```

components/clp-py-utils/clp_py_utils/initialize-orchestration-db.py
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 import argparse
 import logging
+import os
```
kirkrodrigues:
```suggestion
```

junhaoliao:
converting to draft before conflicts are resolved

junhaoliao:
i merged main and re-validated
@kirkrodrigues it's back ready for review now

junhaoliao:
sorry i forgot to run lint. putting it back to draft for now

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
junhaoliao: actually i think we can just replace this with
```
clp_config.load_database_credentials_from_env()
```

junhaoliao: then check if the credentials are loaded here. if not, log a message those aren't set

junhaoliao:
@kirkrodrigues it's now ready for review again after some stylistic changes

components/clp-package-utils/clp_package_utils/scripts/native/archive_manager.py
```diff
@@ -1,5 +1,6 @@
 import argparse
 import logging
+import os
```
kirkrodrigues:
```suggestion
```

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -1,6 +1,7 @@
 import argparse
 import datetime
 import logging
+import os
```
kirkrodrigues:
```suggestion
```

components/clp-package-utils/clp_package_utils/scripts/archive_manager.py
```diff
@@ -204,14 +204,17 @@ def main(argv: typing.List[str]) -> int:
...
         mounts.clp_home,
         mounts.logs_dir,
         mounts.archives_output_dir,
     ]
+    extra_env_vars = {
+        "CLP_DB_USER": clp_config.database.username,
+        "CLP_DB_PASS": clp_config.database.password,
```
kirkrodrigues: Can we use the constants from clp_config.py?

components/clp-package-utils/clp_package_utils/scripts/compress.py
```diff
@@ -222,8 +222,12 @@ def main(argv):
 
     _generate_logs_list(clp_config.logs_input.type, container_logs_list_path, parsed_args)
 
+    extra_env_vars = {
+        "CLP_DB_USER": clp_config.database.username,
+        "CLP_DB_PASS": clp_config.database.password,
```
kirkrodrigues: Ditto

components/clp-package-utils/clp_package_utils/scripts/decompress.py
```diff
@@ -122,8 +122,13 @@ def handle_extract_file_cmd(
                 container_paths_to_extract_file_path,
             )
         )
+
+    extra_env_vars = {
+        "CLP_DB_USER": clp_config.database.username,
+        "CLP_DB_PASS": clp_config.database.password,
```
kirkrodrigues: Ditto

components/clp-package-utils/clp_package_utils/scripts/decompress.py
```diff
@@ -193,9 +198,13 @@ def handle_extract_stream_cmd(
     generated_config_path_on_container, generated_config_path_on_host = dump_container_config(
         container_clp_config, clp_config, container_name
     )
+    extra_env_vars = {
+        "CLP_DB_USER": clp_config.database.username,
+        "CLP_DB_PASS": clp_config.database.password,
```
kirkrodrigues: Ditto

components/clp-package-utils/clp_package_utils/scripts/decompress.py
```diff
@@ -193,9 +198,13 @@ def handle_extract_stream_cmd(
...
+    }
     necessary_mounts = [mounts.clp_home, mounts.logs_dir]
```
kirkrodrigues: Can we move this above extra_env_vars?

components/clp-package-utils/clp_package_utils/scripts/search.py
```diff
@@ -117,10 +117,13 @@ def main(argv):
     generated_config_path_on_container, generated_config_path_on_host = dump_container_config(
         container_clp_config, clp_config, container_name
     )
-
     necessary_mounts = [mounts.clp_home, mounts.logs_dir]
+    extra_env_vars = {
+        "CLP_DB_USER": clp_config.database.username,
+        "CLP_DB_PASS": clp_config.database.password,
```
kirkrodrigues: Ditto

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -1034,14 +1008,20 @@ def start_reducer(
...
     # fmt: on
     env_vars = [
-        f"PYTHONPATH={clp_site_packages_dir}",
+        *generate_common_environment_variables(include_clp_home=True),
+        *generate_credential_environment_variables(
+            container_clp_config,
+            include_db_credentials=True,
+            include_queue_credentials=True,
+            include_redis_credentials=True,
+        ),
```
kirkrodrigues:
```suggestion
```

Don't think the reducer start script needs any credentials, does it?

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
kirkrodrigues, line 1: We need to update garbage_collector to use the new stuff.

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -1274,6 +1254,9 @@ def main(argv):
     clp_config.data_directory.mkdir(parents=True, exist_ok=True)
     clp_config.logs_directory.mkdir(parents=True, exist_ok=True)
 
+    # Dump the container config
+    dump_container_config(container_clp_config, clp_config, "clp")
```
kirkrodrigues: We're relying on dump_container_config to continue using `.{container_name}-config.yml` and `get_generated_config_file_path` to use `.clp-config.yml`. Any way to avoid having them go out of sync?

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -327,13 +340,17 @@ def dump_container_config(
...
 
 def generate_container_start_cmd(
-    container_name: str, container_mounts: List[Optional[DockerMount]], container_image: str
+    container_name: str,
+    container_mounts: List[Optional[DockerMount]],
+    container_image: str,
+    extra_env_vars: Optional[Dict[str, str]] = None,
 ) -> List[str]:
     """
     Generates the command to start a container with the given mounts and name.
```
```suggestion
    Generates the command to start a container with the given mounts, environment variables, and
    name.

```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -349,6 +366,9 @@ def generate_container_start_cmd(
         "--name", container_name,
         "--log-driver", "local"
     ]
+    if extra_env_vars is not None:
```
kirkrodrigues: Might be worth doing something like:

- Create an env_vars set with PYTHONPATH
- Check if extra_env_vars is non-empty
- If extra_env_vars contains PYTHONPATH, error out.
- Extend env_vars with extra_env_vars
- Update the container start cmd

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -590,3 +610,71 @@ def is_retention_period_configured(clp_config: CLPConfig) -> bool:
         return True
 
     return False
+
+
+def generate_common_environment_variables(
```
kirkrodrigues: How about `get_common_env_vars_list`? Similarly for the other methods.

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -590,3 +610,71 @@ def is_retention_period_configured(clp_config: CLPConfig) -> bool:
...
+    include_clp_home=True,
+) -> List[str]:
+    """
+    Generate a list of common environment variables for Docker containers.
+
+    :param include_clp_home:
+    :return: List of environment variable strings in the format "KEY=VALUE".
```
```suggestion
    :param include_clp_home:
    :return: A list of common environment variables for Docker containers in the format "KEY=VALUE".
```

Similarly for the other methods.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -878,6 +917,9 @@ def get_runnable_components(self) -> Set[str]:
     def dump_to_primitive_dict(self):
         d = self.dict()
         d["logs_input"] = self.logs_input.dump_to_primitive_dict()
+        d["database"] = self.database.dump_to_primitive_dict()
+        d["queue"] = self.queue.dump_to_primitive_dict()
+        d["redis"] = self.redis.dump_to_primitive_dict()
```
kirkrodrigues: Can we move this block one line up so that the paths remain grouped?

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -404,11 +404,12 @@ def main(argv):
...
         clp_config = CLPConfig.parse_obj(read_yaml_config_file(config_path))
-    except ValidationError as err:
+        clp_config.database.load_credentials_from_env()
+    except (ValidationError, ValueError) as err:
         logger.error(err)
         return -1
-    except Exception as ex:
-        logger.error(ex)
+    except Exception:
+        logger.exception("Failed to initialize compression scheduler.")
```
kirkrodrigues: Can we use the component name instead?

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1153,11 +1153,12 @@ async def main(argv: List[str]) -> int:
...
         clp_config = CLPConfig.parse_obj(read_yaml_config_file(config_path))
-    except ValidationError as err:
+        clp_config.database.load_credentials_from_env()
+    except (ValidationError, ValueError) as err:
         logger.error(err)
         return -1
-    except Exception as ex:
-        logger.error(ex)
+    except Exception:
+        logger.exception("Failed to initialize query scheduler.")
```
kirkrodrigues: Ditto

kirkrodrigues (changes requested):

components/clp-package-utils/clp_package_utils/scripts/archive_manager.py
haiqi96: I actually have the fix in https://github.com/y-scope/clp/pull/1186

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
junhaoliao: right, sorry for the confusion
now removed

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
junhaoliao: how about adding a `def dump_shared_container_config(container_clp_config: CLPConfig, clp_config: CLPConfig)` (or `dump_global_container_config`), extracting a helper from the existing dumping logic in `dump_container_config`, then letting both `dump_shared_container_config` and `dump_container_config` call the helper?

kirkrodrigues: Hm, how would that avoid the paths going out of sync?

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
junhaoliao: updated. also updated the verification steps in the PR description

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
junhaoliao: sorry, i forgot to explain the out of sync prevention part.

(tl;dr - it might be easier to just look at the modification: f852ca48e9bb90c49dce40067a927c81c315933a)

the above explanation is to say we should avoid assigning "clp" as a container name, so we extract a helper to just write some config object to some file.
`dump_container_config` should still take a container name as an arg, but `dump_shared_config` should reference a constant `CLP_GENERATED_CONFIG_FILENAME` instead of accepting it as an argument. `get_generated_config_file_path` should ref that constant for the filename as well.

alternatively, we can also create a config_filename generator function and use that in `get_generated_config_file_path`

if you agree with the approach, maybe my next step is to rename "generated config" as "shared config" for accuracy.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
junhaoliao: also updated the "Starting compression scheduler" message

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
junhaoliao: also updated the "Query scheduler started." message

components/clp-package-utils/clp_package_utils/general.py
junhaoliao: > If extra_env_vars contains PYTHONPATH, error out.

Will it be more future proof if we allow override to happen? if we explain that behaviour in the docstring of param `extra_env_vars`

junhaoliao: changes are made at c0949cc9acc5624f7270ecf0aad41df7acb4513a without the `PYTHONPATH` check

junhaoliao:
@kirkrodrigues 
i haven't merged the main branch into the PR because the latest commit on main branch broken the package which causes inconvenience in validating this PR. i'll do that after the issue is resolved on the main branch. for now, please help taking a look into the proposed changes per opening discussion if you may

haiqi96:
> @kirkrodrigues i haven't merged the main branch into the PR because the latest commit on main branch broken the package which causes inconvenience in validating this PR. i'll do that after the issue is resolved on the main branch. for now, please help taking a look into the proposed changes per opening discussion if you may

My previous commit has caused an issue for clp compression. I came up with a fix and Junhao has tested it locally and made sure it work. Will submit a PR with the fix tomorrow.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -233,6 +242,25 @@ def get_clp_connection_params_and_type(self, disable_localhost_socket_connection
...
+            self.password = get_config_value(config, f"{DB_COMPONENT_NAME}.password")
+        except KeyError as ex:
+            raise ValueError(
+                f"Credentials file '{credentials_file_path}' does not contain key '{ex}'."
+            )
+
+    def load_credentials_from_env(self):
+        self.username = _get_env_var(CLP_DB_USER_ENV_VAR_NAME)
+        self.password = _get_env_var(CLP_DB_PASS_ENV_VAR_NAME)
+
```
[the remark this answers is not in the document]
junhaoliao: _get_env_var() already throws, no?

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -233,6 +242,28 @@ def get_clp_connection_params_and_type(self, disable_localhost_socket_connection
...
+        if config is None:
+            raise ValueError(f"Credentials file '{credentials_file_path}' is empty.")
+        try:
+            self.username = get_config_value(config, f"{DB_COMPONENT_NAME}.user")
+            self.password = get_config_value(config, f"{DB_COMPONENT_NAME}.password")
+        except KeyError as ex:
+            raise ValueError(
+                f"Credentials file '{credentials_file_path}' does not contain key '{ex}'."
+            )
+
```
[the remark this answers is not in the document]
junhaoliao: This seems beneficial in the long run.
@haiqi96 what do you think? shall we do this in another PR?

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -882,6 +930,9 @@ def get_runnable_components(self) -> Set[str]:
     def dump_to_primitive_dict(self):
         d = self.dict()
         d["logs_input"] = self.logs_input.dump_to_primitive_dict()
+        d["database"] = self.database.dump_to_primitive_dict()
+        d["queue"] = self.queue.dump_to_primitive_dict()
+        d["redis"] = self.redis.dump_to_primitive_dict()
```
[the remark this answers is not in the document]
junhaoliao: i feel below is more maintainable?
```
    nested_keys = (
        "logs_input",
        "database",
        "queue",
        "redis",
        "archive_output",
        "stream_output",
    )

    d = self.dict(exclude=set(nested_keys))
    for key in nested_keys:
        d[key] = getattr(self, key).dump_to_primitive_dict()

    # ... path string conversions unchanged
    return d
```

@kirkrodrigues shall we make the change in this PR or another one?

components/clp-py-utils/clp_py_utils/clp_config.py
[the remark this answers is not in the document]
haiqi96: True, maybe changing user to username in the clp-config file? I don't think we need it for this release, and we might want to provide users a patch file so they can covert their old config that uses `user` to `username`

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -611,40 +610,34 @@ def generic_start_scheduler(
...
-            f"{container_clp_config.redis.query_backend_database}"
+    env_vars = [
+        *generate_common_environment_variables(),
+        *generate_credential_environment_variables(
+            container_clp_config, include_db_credentials=True
         ),
+        *generate_celery_connection_environment_variables(container_clp_config),
         f"CLP_LOGS_DIR={container_logs_dir}",
         f"CLP_LOGGING_LEVEL={clp_config.query_scheduler.logging_level}",
     ]
```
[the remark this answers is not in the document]
junhaoliao: the mismatching component specific logging level is a valid concern. i thought i created an issue for this some time ago but couldn't find it. @coderabbitai please help create an issue

components/clp-py-utils/clp_py_utils/clp_config.py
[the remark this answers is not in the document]
kirkrodrigues: Sure, we can make it in this PR.

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -1105,16 +1068,16 @@ def start_garbage_collector(
...
 
-    necessary_env_vars = [
-        f"PYTHONPATH={clp_site_packages_dir}",
-        f"CLP_HOME={CONTAINER_CLP_HOME}",
-        f"CLP_LOGS_DIR={container_logs_dir}",
-        f"CLP_LOGGING_LEVEL={clp_config.garbage_collector.logging_level}",
-    ]
     necessary_mounts = [
         mounts.clp_home,
         mounts.logs_dir,
```
kirkrodrigues, line 1071: Don't we need to add `mounts.generated_config_file` here?

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -308,33 +322,48 @@ def generate_worker_config(clp_config: CLPConfig) -> WorkerConfig:
     return worker_config
 
 
-def dump_container_config(
-    container_clp_config: CLPConfig, clp_config: CLPConfig, container_name: str
-) -> Tuple[pathlib.Path, pathlib.Path]:
+def dump_config(container_clp_config: CLPConfig, clp_config: CLPConfig, config_filename: str):
     """
     Writes the given config to the logs directory so that it's accessible in the container.
```
```suggestion
    Writes the given container config to the logs directory, so that it's accessible in the
    container.

```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -308,33 +322,48 @@ def generate_worker_config(clp_config: CLPConfig) -> WorkerConfig:
...
-    config_file_path_on_container = container_clp_config.logs_directory / container_config_filename
+    config_file_path_on_host = clp_config.logs_directory / config_filename
+    config_file_path_on_container = container_clp_config.logs_directory / config_filename
     with open(config_file_path_on_host, "w") as f:
         yaml.safe_dump(container_clp_config.dump_to_primitive_dict(), f)
 
     return config_file_path_on_container, config_file_path_on_host
 
 
+def dump_container_config(
```
kirkrodrigues: Can we just get rid of this function and rename `dump_config` to `dump_container_config`?

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -308,33 +322,48 @@ def generate_worker_config(clp_config: CLPConfig) -> WorkerConfig:
...
+    container_clp_config: CLPConfig, clp_config: CLPConfig, container_name: str
+) -> Tuple[pathlib.Path, pathlib.Path]:
+    return dump_config(container_clp_config, clp_config, f".{container_name}-config.yml")
+
+
+def dump_shared_config(
```
kirkrodrigues: How about `dump_shared_container_config`?

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -308,33 +322,48 @@ def generate_worker_config(clp_config: CLPConfig) -> WorkerConfig:
...
+    container_clp_config: CLPConfig, clp_config: CLPConfig
+) -> Tuple[pathlib.Path, pathlib.Path]:
```
kirkrodrigues, line 351:
```suggestion
) -> Tuple[pathlib.Path, pathlib.Path]:
    """
    Dumps the given container config to `CLP_GENERATED_CONFIG_FILENAME` in the logs directory, so
    that it's accessible in the container.

    :param container_clp_config:
    :param clp_config:
    """
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -308,33 +322,48 @@ def generate_worker_config(clp_config: CLPConfig) -> WorkerConfig:
...
 ) -> List[str]:
     """
-    Generates the command to start a container with the given mounts and name.
+    Generates the command to start a container with the given mounts, environment variables, and
+    name.
+
     :param container_name:
     :param container_mounts:
     :param container_image:
+    :param extra_env_vars: Environment variables to set on top of predefined ones.
```
```suggestion
    :param extra_env_vars: Environment variables to set on top of the predefined ones.
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -599,3 +634,71 @@ def is_retention_period_configured(clp_config: CLPConfig) -> bool:
...
+
+
+def get_common_env_vars_list(
+    include_clp_home_env_var=True,
+) -> List[str]:
+    """
+    Generate a list of common environment variables for Docker containers.
+
+    :param include_clp_home_env_var:
+    :return: A list of common environment variables for Docker containers in the format "KEY=VALUE".
```
```suggestion
    :param include_clp_home_env_var:
    :return: A list of common environment variables for Docker containers, in the format
    "KEY=VALUE".
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -599,3 +634,71 @@ def is_retention_period_configured(clp_config: CLPConfig) -> bool:
...
+    clp_site_packages_dir = CONTAINER_CLP_HOME / "lib" / "python3" / "site-packages"
+    env_vars = [f"PYTHONPATH={clp_site_packages_dir}"]
+
+    if include_clp_home_env_var:
+        env_vars.append(f"CLP_HOME={CONTAINER_CLP_HOME}")
+
+    return env_vars
+
+
+def generate_credential_env_vars_list(
```
kirkrodrigues: How about `get_credential_env_vars_list`?

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -599,3 +634,71 @@ def is_retention_period_configured(clp_config: CLPConfig) -> bool:
...
+    include_redis_credentials=False,
+) -> List[str]:
+    """
+    Generates a list of credential environment variables for Docker containers.
+
+    :param container_clp_config:
+    :param include_db_credentials:
+    :param include_queue_credentials:
+    :param include_redis_credentials:
+    :return: A list of common environment variables for Docker containers in the format "KEY=VALUE".
```
```suggestion
    :param container_clp_config:
    :param include_db_credentials:
    :param include_queue_credentials:
    :param include_redis_credentials:
    :return: A list of credential environment variables for Docker containers, in the format
    "KEY=VALUE".
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -599,3 +634,71 @@ def is_retention_period_configured(clp_config: CLPConfig) -> bool:
...
+        env_vars.append(f"CLP_QUEUE_USER={container_clp_config.queue.username}")
+        env_vars.append(f"CLP_QUEUE_PASS={container_clp_config.queue.password}")
+
+    if include_redis_credentials:
+        env_vars.append(f"CLP_REDIS_PASS={container_clp_config.redis.password}")
+
+    return env_vars
+
+
+def generate_celery_connection_env_vars_list(container_clp_config: CLPConfig) -> List[str]:
```
kirkrodrigues: How about `get_celery_connection_env_vars_list`?

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -599,3 +634,71 @@ def is_retention_period_configured(clp_config: CLPConfig) -> bool:
...
+    """
+    Generate a list of Celery connection environment variables for Docker containers.
+
+    :param container_clp_config:
+    :return: A list of common environment variables for Docker containers in the format "KEY=VALUE".
```
```suggestion
    :param container_clp_config:
    :return: A list of Celery connection environment variables for Docker containers, in the format
    "KEY=VALUE".
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -236,6 +246,28 @@ def get_clp_connection_params_and_type(self, disable_localhost_socket_connection
...
+            self.username = get_config_value(config, f"{DB_COMPONENT_NAME}.user")
+            self.password = get_config_value(config, f"{DB_COMPONENT_NAME}.password")
+        except KeyError as ex:
+            raise ValueError(
+                f"Credentials file '{credentials_file_path}' does not contain key '{ex}'."
+            )
+
+    def load_credentials_from_env(self):
+        """
+        :raise ValueError if any expected environment variable is not set.
```
```suggestion
        :raises ValueError: if any expected environment variable is not set.
```

Seems like that's the correct format according to: https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -841,40 +922,8 @@ def load_execution_container_name(self):
...
-        if config is None:
-            raise ValueError(f"Credentials file '{self.credentials_file_path}' is empty.")
-        try:
-            self.redis.password = get_config_value(config, f"{REDIS_COMPONENT_NAME}.password")
-        except KeyError as ex:
-            raise ValueError(
-                f"Credentials file '{self.credentials_file_path}' does not contain key '{ex}'."
-            )
+    def get_generated_config_file_path(self) -> pathlib.Path:
+        return self.logs_directory / ".clp-config.yml"
```
```suggestion
        return self.logs_directory / CLP_GENERATED_CONFIG_FILENAME
```

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
junhaoliao: right we should. it was still functional without that mount because the logs directory was mounted, though we should really add it

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: sorry for missing this

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: i use `raise` instead of `raises` because of the preference we have with JSDoc

they might be the same according to this doc:
https://www.sphinx-doc.org/en/master/usage/domains/python.html#:~:text=raises%2C%20raise%2C%20except%2C%20exception%3A%20That%20(and%20when)%20a%20specific%20exception%20is%20raised.

and `raise` is also recognized by the PyCharm

i'm fine using any. let's document any convention once we decide

kirkrodrigues: Sure, we can use `raise`. (Btw, the other difference is putting another colon after the type.)

components/clp-py-utils/clp_py_utils/clp_config.py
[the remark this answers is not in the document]
junhaoliao: @coderabbitai create an issue to track this. please also include Haiqi's config file compatibility concern in the issue

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -308,33 +322,57 @@ def generate_worker_config(clp_config: CLPConfig) -> WorkerConfig:
...
     """
-    container_config_filename = f".{container_name}-config.yml"
-    config_file_path_on_host = clp_config.logs_directory / container_config_filename
-    config_file_path_on_container = container_clp_config.logs_directory / container_config_filename
+    config_file_path_on_host = clp_config.logs_directory / config_filename
+    config_file_path_on_container = container_clp_config.logs_directory / config_filename
     with open(config_file_path_on_host, "w") as f:
         yaml.safe_dump(container_clp_config.dump_to_primitive_dict(), f)
 
     return config_file_path_on_container, config_file_path_on_host
```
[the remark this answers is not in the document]
junhaoliao: if the parent path doesn't exist, it's probably fine to let it fail to show any implementation error during development

junhaoliao:
@kirkrodrigues i just revalidated. it's ready for review again

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -80,9 +81,18 @@
 CLP_DEFAULT_CREDENTIALS_FILE_PATH = pathlib.Path("etc") / "credentials.yml"
 CLP_DEFAULT_DATA_DIRECTORY_PATH = pathlib.Path("var") / "data"
 CLP_DEFAULT_DATASET_NAME = "default"
+CLP_SHARED_CONFIG_FILENAME = ".clp-config.yml"
 CLP_METADATA_TABLE_PREFIX = "clp_"
```
```suggestion
CLP_METADATA_TABLE_PREFIX = "clp_"
CLP_SHARED_CONFIG_FILENAME = ".clp-config.yml"
```

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -1281,6 +1243,9 @@ def main(argv):
     clp_config.data_directory.mkdir(parents=True, exist_ok=True)
     clp_config.logs_directory.mkdir(parents=True, exist_ok=True)
 
+    # Dump the container config
```
kirkrodrigues:
```suggestion
```

kirkrodrigues (dismissed):
A few minor nits.

For the PR title, how about:

```
refactor(package): Standardize how configuration and credentials are passed to containers (resolves #1149).
```

kirkrodrigues (approved):

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -350,6 +388,12 @@ def generate_container_start_cmd(
         "--name", container_name,
         "--log-driver", "local"
     ]
+    env_vars = {
+        "PYTHONPATH": clp_site_packages_dir,
```
junhaoliao, line 392: the above `"-e", f"PYTHONPATH={clp_site_packages_dir}",` line was not removed, as a mistake possibly during branch syncs with `main`, which causes the environment variable to be defined twice. The duplication isn't harmful in the correctness though. 

@coderabbitai create an issue to track the removal of the deplicate

PR #1160, opened 2025-08-04
feat(webui): Focus/refocus query input box for ease of use.

### LINKED ISSUE #1068: [new-webui] Focus/refocus query input box for ease of use
### Request

* When a user clicks on the search page, we should set the focus to the query input box so that the user doesn't need to do so manually before they can search.
* If a user was focused on the query input box and then they press enter to search, after the search completes, focus should return to the query input box.

Technically, we could *always* set the focus to the query input box after a query completes, but this might affect usability. Ideally, focus should return to whatever input the user used before starting the query (moreover, we should support starting a new query while the previous query is running without having to manually cancel the query, but that's a different issue). So if the user was focused on the time range selector before starting a query, focus should return to it after the query completes.

### Possible implementation

Not entirely sure.

davemarco:
# Description

PR focuses input box when user enters search page, and when search completes.

Resolves #1068. Not exactly sure what the best way to "return to whatever input the user used before starting the query" mentioned in the issue, so for now it just refocuses on the input when the query completes.

# Checklist

  breaking change.

# Validation performed

Tested it focuses on input on load, and on query success. 

## Summary by CodeRabbit

* **New Features**
  * The query input box now automatically receives focus when the search interface is in its default or completed state, improving usability.


davemarco:
@junhaoliao can you approve this

junhaoliao (approved):
Deferring to @hoophalab ‘s review

PR #1163, opened 2025-08-05
feat(clp-s)!: Use core clp parsing and search code in clp-s; Bump archive version to 0.4.0.

gibber9809:
# Description

This PR is the third in a series of PRs (see #1143 for the second) that is attempting to bring the core CLP parsing and search code to a state where it can be shared by both clp and clp-s. The full end to end integration can be seen [here](https://github.com/gibber9809/clp/tree/clp-core-refactor-prototype).

This PR switches out the core encoding and search code for log-text in clp-s to use the clp implementations. Since the previous two PRs did most of the heavy lifting to make this possible the bulk of this change is comprised of:

1. Making clp-s' log-type dictionary entry (in `clp_s/DictionaryEntry.{cpp,hpp}`) act like clp's log-type dictionary entry (in `clp/LogTypeDictionaryEntry.{cpp,hpp}`). This makes clp-s' log-type dictionary entry follow the interface defined by the concept that will be added in a follow-up PR.
2. Deleting clp-s' version of encoding/decoding/search for log-text
3. Updating cmake scripts to include the clp implementations of encoding/decoding/search for log-text

There is also a small change in the clp parsing utilities to eliminate a performance issue spotted during profiling.

Of these the most complex part is probably the first point, so it might be useful to look at the concept from the final prototype implementation as well as compare the clp-s logtype-dictionary entry implementation with the clp version.

Overall this change leads to an extremely marginal improvement in compression ratio, and compression and decompression performance that ranges from about even to somewhat significantly better across the open source datasets.

I have not yet benchmarked search performance as thoroughly, but ad-hoc tests didn't reveal any issues.

Compression ratio results:

| dataset | new ratio | old ratio |
| ----- | ----- | ----- |
| cockroachdb | 27.974 | 28.313 |
| mongodb | 172.359 | 170.949 |
| elasticsearch | 152.243 | 152.630 |
| spark-event-logs | 57.555 | 57.448 |
| postgresql | 39.532 | 39.124 |

Compression speed results:
| dataset | new time (s) | old time (s) |
| ----- | ----- | ----- |
| cockroachdb | 147.066 | 169.978 |
| mongodb | 747.591 | 771.486 |
| elasticsearch | 70.752 | 71.229 |
| spark-event-logs | 81.316 | 83.806 |
| postgresql | 6.522 | 6.774 |

Decompression speed results show a similar trend. Overall performance seems better (though before the change to the clp parsing utilities performance was worse in most cases).

# Checklist

  breaking change.

# Validation performed

* Validated that compression ratio is the same or better than before
* Validated that compression and decompression speed is the same or better than before
* Validated that all tests pass

## Summary by CodeRabbit

* **New Features**
  * Improved type safety and clarity for dictionary IDs and variable handling.
  * Enhanced support for efficient string handling using string views.

* **Refactor**
  * Unified and modernized placeholder and variable management in log and variable dictionaries.
  * Switched to more explicit type aliases for identifiers and improved serialization logic.

* **Bug Fixes**
  * Increased performance and reliability in log type parsing and variable encoding/decoding.

* **Chores**
  * Removed deprecated and redundant internal encoding/decoding and search query components.
  * Updated archive format version to ensure compatibility with new features.

* **Style**
  * Improved const-correctness and code readability in utility classes.

* **Documentation**
  * Updated method signatures and type usage in public interfaces for better clarity.


components/core/src/clp_s/CMakeLists.txt
```diff
@@ -48,17 +56,28 @@ set(
         ../clp/FileDescriptor.hpp
         ../clp/FileReader.cpp
         ../clp/FileReader.hpp
+        ../clp/GrepCore.cpp
+        ../clp/GrepCore.hpp
         ../clp/hash_utils.cpp
         ../clp/hash_utils.hpp
+        ../clp/ir/LogEvent.hpp
+        ../clp/ir/types.hpp
```
haiqi96: let's alphabetically reorder these two

components/core/src/clp_s/indexer/CMakeLists.txt
```diff
@@ -13,12 +13,36 @@ set(
...
         ../../clp/hash_utils.cpp
         ../../clp/hash_utils.hpp
+        ../../clp/ir/LogEvent.hpp
+        ../../clp/ir/types.hpp
+        ../../clp/ir/constants.hpp
+        ../../clp/ir/EncodedTextAst.cpp
+        ../../clp/ir/EncodedTextAst.hpp
+        ../../clp/ir/parsing.cpp
+        ../../clp/ir/parsing.hpp
+        ../../clp/ir/parsing.inc
```
haiqi96: ditto: hope I get the order right?
```suggestion
        ../../clp/ir/constants.hpp
        ../../clp/ir/EncodedTextAst.cpp
        ../../clp/ir/EncodedTextAst.hpp
        ../../clp/ir/LogEvent.hpp
        ../../clp/ir/parsing.cpp
        ../../clp/ir/parsing.hpp
        ../../clp/ir/parsing.inc
        ../../clp/ir/types.hpp
```

components/core/src/clp_s/search/QueryRunner.hpp
```diff
@@ -152,6 +152,8 @@ class QueryRunner : public FilterClass {
             std::vector<std::pair<ExpressionType, ast::OpList::iterator>>>
             m_expression_state;
 
+    log_surgeon::lexers::ByteLexer m_dummy_lexer;
```
haiqi96: maybe `m_placeholder_lexer`?

components/core/src/clp_s/DictionaryReader.hpp
```diff
@@ -176,7 +183,11 @@ DictionaryReader<DictionaryIdType, EntryType>::get_entry_matching_value(
     }
 
     std::vector<EntryType const*> entries;
-    auto const search_string_uppercase = boost::algorithm::to_upper_copy(search_string);
+    std::string search_string_uppercase;
+    std::ignore = boost::algorithm::to_upper_copy(
+            std::back_inserter(search_string_uppercase),
+            search_string
+    );
```
haiqi96, line 190: I have approved a previous PR with this change, but just curious why the updated version is preferred?

gibber9809: It's not really preferred, but the other signature doesn't let you have a different and output type which is a problem for us since we want to pass in a string_view and get a strong as a result.

gibber9809: It's not really preferred, but the other signature doesn't let you have a different and output type which is a problem for us since we want to pass in a string_view and get a string as a result.

components/core/src/clp_s/ColumnWriter.cpp
```diff
@@ -74,15 +83,14 @@ void ClpStringColumnWriter::store(ZstdCompressor& compressor) {
...
+    clp::variable_dictionary_id_t id{};
+    m_var_dict->add_entry(std::get<std::string>(value), id);
     m_variables.push_back(id);
-    return sizeof(int64_t);
+    return sizeof(clp::variable_dictionary_id_t);
 }
 
 void VariableStringColumnWriter::store(ZstdCompressor& compressor) {
-    size_t size = m_variables.size() * sizeof(int64_t);
+    size_t size = m_variables.size() * sizeof(clp::variable_dictionary_id_t);
```
haiqi96: we can use auto?

it's not related to your change, but since we already have some auto used in this file, replacing size_t with auto doesn't harm

components/core/src/clp_s/ColumnWriter.cpp
```diff
@@ -49,15 +57,16 @@ void BooleanColumnWriter::store(ZstdCompressor& compressor) {
 }
 
 size_t ClpStringColumnWriter::add_value(ParsedMessage::variable_t& value) {
-    std::string string_var = std::get<std::string>(value);
-    uint64_t id;
-    uint64_t offset = m_encoded_vars.size();
-    VariableEncoder::encode_and_add_to_dictionary(
-            string_var,
+    uint64_t offset{m_encoded_vars.size()};
+    m_temp_var_dict_ids.clear();
```
haiqi96: I feel that making this to be local is better than creating a class member specifically for a temporary usage purpose.

components/core/src/clp_s/ColumnWriter.hpp
```diff
@@ -174,6 +174,7 @@ class ClpStringColumnWriter : public BaseColumnWriter {
 
     std::vector<int64_t> m_logtypes;
```
haiqi96: seems to be a type mismatch?

at [here](https://github.com/y-scope/clp/pull/1163/files#diff-27d0ee35cfdbf2ec8591f4f29fbd02618e17e1e87ed605ee88564aa03e58056cR71-R72), the return type of encode_log_dict_id is uint64_t.

You might also want to double check type for m_encoded_vars.

Not sure if we need to make a specific type for `encode_log_dict_id` like we do in clp.

components/core/src/clp_s/ColumnWriter.hpp
```diff
@@ -193,7 +194,7 @@ class VariableStringColumnWriter : public BaseColumnWriter {
 
 private:
     std::shared_ptr<VariableDictionaryWriter> m_var_dict;
-    std::vector<int64_t> m_variables;
+    std::vector<clp::variable_dictionary_id_t> m_variables;
```
haiqi96: shall we rename this to `m_variable_ids` (or `m_variables_ids`?)?

components/core/src/clp_s/ColumnWriter.hpp
```diff
@@ -174,6 +174,7 @@ class ClpStringColumnWriter : public BaseColumnWriter {
 
     std::vector<int64_t> m_logtypes;
     std::vector<int64_t> m_encoded_vars;
```
haiqi96: somewhere in encode_and_add_to_dictionary we are implicilty converting encoded_var_t to int64_t. when we push an encoded_var to this vector This doesn't cause any issue because `encoded_var_t ` is currently the same as int64_t, but to prevent future issues, I think we should either do a explicit cast, or use  `clp::encoded_var_t ` as the vector datatype?

components/core/src/clp_s/DictionaryEntry.cpp
haiqi96, line 1: For the DictionaryEntry.cpp/hpp, since you are doing a major refactor, I will skip the old version and focus on understanding how's the newer version different from what we use in CLP. (let me know if there;s anything you want me to look at in the older version).

haiqi96, line 1: From high-level, I am trying to understand what's the difference between the clp and clp-s's dictionary entry, and what's preventing us to reuse the clp's one (or maybe create a common one that can be shared by the two).

I found out that between CLP and CLP-S, most code of dictionary entry are alike, and Here's the major difference that I noticed;
- Only CLP's dictionary entry has segment support.
- CLP-S' dictionary entry has lazy read support.
- CLP-S uses string_view whereas CLP uses string const& for some function interface.
- They seem to use different errorcode.

And a few other differences that we can potentially work around?
- CLP-S dictionary entry only supports Zstd compressor.
- CLP-S has a decode method, which I think can be supported for CLP as well.

I feel that there can be some refactor/class inheritance we can do to reduce code duplications, but not necessarily in this PR. wonder have you thought about it?

gibber9809: I think they also have different encoding when writing to file.

It is something I thought about, but I figured it is simpler to not share it for now. We'll probably end up adding lazy initialization support + optional support for encoding the entry the clp-s way to the clp implementation to further deduplicate the core in the future.

components/core/src/clp_s/ColumnWriter.cpp
gibber9809: I prefer class member over local in this instance since it helps avoid a large number of memory allocations in some frequently run code.

components/core/src/clp_s/DictionaryEntry.cpp
haiqi96: True, and I didn't notice that variable dictionary entries handle the id in slight different way.

If you don't plan to work on the deduplication in short term, I think it might be a good idea to create an issue and document the differences between the CLP and CLP-S, so later on when someone has to work on this, he/she has something to refer to. how do you think?

components/core/src/clp_s/ColumnWriter.cpp
haiqi96: Fair

components/core/src/clp_s/DictionaryEntry.cpp
gibber9809: That makes sense. I can write an issue.

gibber9809: Up as #1165.

components/core/src/clp_s/ColumnWriter.hpp
gibber9809: Good catch -- I'm adding a `using encoded_log_dict_id_t = uint64_t` that I'll use all over and  changed the encoded variables to `clp::encoded_variable_t` respectively.``

components/core/src/clp_s/ColumnWriter.hpp
gibber9809: Changing this to `clp::encoded_var_t`.

components/core/src/clp_s/ColumnWriter.hpp
gibber9809: I think I'll go with `m_var_dict_ids` since I see that use `xyz_var_dict_ids` used in `clp/Query.hpp` at least.

components/core/src/clp_s/ColumnWriter.cpp
davidlion: Fwiw, I agree with Haiqi. Unless we have metrics to back up the importance of this it feels very much like a premature optimization.

components/core/src/clp_s/search/QueryRunner.hpp
```diff
@@ -152,7 +152,7 @@ class QueryRunner : public FilterClass {
             std::vector<std::pair<ExpressionType, ast::OpList::iterator>>>
             m_expression_state;
 
-    log_surgeon::lexers::ByteLexer m_dummy_lexer;
+    log_surgeon::lexers::ByteLexer m_placeholder_lexer;
```
haiqi96: I am fine with this. but David's comment could also apply to this member variable.

components/core/src/clp_s/ColumnWriter.hpp
```diff
@@ -141,7 +144,7 @@ class ClpStringColumnWriter : public BaseColumnWriter {
      * @param encoded_id
      * @return the encoded log dict id
      */
-    static clp::logtype_dictionary_id_t get_encoded_log_dict_id(uint64_t encoded_id) {
+    static clp::logtype_dictionary_id_t get_encoded_log_dict_id(encoded_log_dict_id_t encoded_id) {
         return static_cast<clp::logtype_dictionary_id_t>(encoded_id & cLogDictIdMask);
```
haiqi96, line 148: it might sounds silly, but shall we do another cast here from uint64_t to clp::logtype_dictionary_id_t?

components/core/src/clp_s/ColumnWriter.hpp
```diff
@@ -160,7 +163,8 @@ class ClpStringColumnWriter : public BaseColumnWriter {
      * @param offset
      * @return The encoded log dict id
      */
-    static uint64_t encode_log_dict_id(clp::logtype_dictionary_id_t id, uint64_t offset) {
+    static encoded_log_dict_id_t
+    encode_log_dict_id(clp::logtype_dictionary_id_t id, uint64_t offset) {
         return static_cast<uint64_t>(id) | (offset << cOffsetBitPosition);
```
haiqi96: shall we do another cast from uint64_t to encoded_log_dict_id_t?

components/core/src/clp_s/ColumnWriter.cpp
gibber9809: That's fair -- I'm kicking off some compression benchmarks now that I'll check on later today. I'll make the change depending on the results.

gibber9809: Yeah it seems to help a bit with some datasets but probably not enough to justify it. I'll get rid of the class member.

components/core/src/clp_s/ColumnWriter.hpp
gibber9809: I think I'll just change that `uint64_t` cast to `encoded_log_dict_id_t` cast

components/core/src/clp_s/ColumnWriter.hpp
```diff
@@ -117,6 +117,9 @@ class BooleanColumnWriter : public BaseColumnWriter {
 
 class ClpStringColumnWriter : public BaseColumnWriter {
 public:
+    // Types
+    using encoded_log_dict_id_t = uint64_t;
+
```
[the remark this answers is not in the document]
gibber9809: This header is currently already included in all places this type needs to be used.

haiqi96 (approved):
Looks good to me. Note that for DictionaryEntry.cpp/hpp, since most parts of code are copied from clp's DictionaryEntry classes, I only reviewed the difference introduced by this PR.

kirkrodrigues (approved):
Deferring to @haiqi96's review.

PR #1166, opened 2025-08-06
feat(docker): Add container image containing the CLP package, for Docker Compose integration (resolves #1164).

### LINKED ISSUE #1164: Create a new Docker image with package contents for Docker Compose integration
### Request

We currently build a `clp-execution-base-ubuntu-jammy` Docker image that only includes the dependencies required to run CLP components. The actual CLP package contents (binaries, Python packages, configuration files, etc.) are mapped into the container at runtime by the `start_clp.py` script. This approach is inconvenient as we plan to migrate to Docker Compose for starting package components. 

The current workflow requires the `start_clp.py` script to:
1. Create Docker mount options for the package directories
2. Map the package contents into containers at runtime
3. Handle various configuration files and environment variables

This approach is not compatible with Docker Compose which expects all necessary files to be present in the container image.

We need to create a new Docker image that includes all the CLP package contents directly in the image, allowing us to use Docker Compose without requiring the `start_clp.py` script to map package contents at runtime.

### Possible implementation

1. **Create a new Dockerfile**:
   - Create a new Dockerfile at `tools/docker-images/`
   - Use a new container name `clp-execution-package-ubuntu-jammy` to avoid compatibility issues with `clp-execution-base-ubuntu-jammy`
   - Duplicate the base image setup from `clp-execution-base-ubuntu-jammy` instead of using `clp-execution-base-ubuntu-jammy` as a base
   - Copy the built CLP package contents directly into the image during build time
   - Set appropriate ownership and permissions for the copied files

2. **Add a new Taskfile task**:
   - Add a new task in `taskfile.yaml` to build this new Docker image
   - The task should depend on the `package` task to ensure the package is built first
   - Name the task something like `docker-image-package-ubuntu-jammy`

3. **Integration approach**:
   - The new image should contain all necessary files in the expected locations:
     - Binaries in `/opt/clp/bin`
     - Python packages in `/opt/clp/lib/python3/site-packages`
     - Configuration files in `/opt/clp/etc`
     - Web UI files in `/opt/clp/var/www/webui`
     - Other necessary directories and files
   - Environment variables should be pre-configured or easily configurable via Docker Compose
   - The image should be self-contained and able to run CLP components without external volume mounts for package contents

4. **Build process modifications**:
   - The new Dockerfile should reference the built package from `build/clp-package`
   - The build task should properly tag the image for use with Docker Compose
   - Consider adding build arguments for customization if needed

5. **CI/CD workflow updates**:
   - Create a new GitHub Actions workflow similar to `clp-execution-image-build.yaml` 
   - The workflow should automatically build and publish the new image when changes are made to its Dockerfile or related files
   - Use path filtering to trigger builds only when relevant files change
   - Follow the same pattern as the existing workflow for consistency

6. **Migration path**:
   - The new image should maintain backward compatibility with existing configurations where possible
   - Documentation should be updated to reflect the new approach
   - The `start_clp.py` script will be modified in a later phase to support both approaches during transition

When implementing this feature, we should carefully review how the current package build process works in `taskfile.yaml` and ensure the Docker image build integrates seamlessly with it. The new image should be built after the package is built, and should include all the contents that are currently mapped at runtime.

junhaoliao:
# Description

This PR introduces a new container image for the CLP package and refactors the image building infrastructure. Key changes include:

- Unified the GitHub Action for building container images to support both "execution" and "package" types
- Added a new task `package-image` inside GH workflow `clp-core-build` for building the "package" image
- Created a new Dockerfile and build script for the package image in `tools/docker-images/clp-package/`
- Updated dependencies in both build and runtiem container images
- Added documentation for building the package image and using the new Docker images
- Renamed the GitHub Action from clp-execution-image-build to clp-build-runtime-image
- Added a new Taskfile for building Docker images locally

The new package image will enable Docker Compose integration as described in issue #1164. Once the package's container orchestration is replaced by Docker Compose, the clp-execution-ubuntu-jammy image can be removed.

# Checklist

  breaking change.
  **Explanation: I did not mark this as breaking because the execution image has not been removed.**
  **Explanation: We shall add docs after the Docker Compose scripts are checked in because the new container image cannot be used by general users standalone anyways.**

# Validation performed

1. Pushed to `junliaoliao/clp:main` and observed newly added job `package-image` completed successfully: ~~https://github.com/junhaoliao/clp/actions/runs/17224822650/job/48867575560~~ (latest: https://github.com/junhaoliao/clp/actions/runs/17445954258/job/49540713145 )
2. `docker run -it ghcr.io/junhaoliao/clp/clp-package:main /bin/bash` locally. Verified:
    * [x] `du -sh /opt/`:
        ```
        450M    /opt/ # which matches `build/clp-package` size right after the package is built in a cleanly cloned CLP repo
        ```
    * [x] `root@4d45478222a6:/# env | grep -E "PATH|CLP_HOME"`
        ```
        CLP_HOME=/opt/clp
        PYTHONPATH=/opt/clp/lib/python3/site-packages
        PATH=/opt/clp/bin:/opt/clp/sbin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
        ```

## Summary by CodeRabbit

- New Features
  - Introduced a minimal “clp-package” Docker image and a task to build it (docker-images:package).

- Bug Fixes
  - Addressed permission issues by setting a safe npm cache path during container runs.

- Chores
  - Renamed and streamlined workflows to “clp-artifact-build” with a new package-image job.
  - Unified image build action with dynamic image selection and leaner dependencies.

- Documentation
  - Added guidance for building the package Docker image.
  - Documented the new clp-package container.
  - Updated workflow references in tooling and contributor docs.


.github/actions/run-on-image/action.yaml
```diff
@@ -45,6 +45,7 @@ runs:
     - run: >-
         docker run
         --user $(id -u):$(id -g)
+        --env npm_config_cache=/tmp/.npm
```
junhaoliao, line 49: otherwise `/.npm` is used, where the default user does not have permission to write by default.

components/core/tools/scripts/lib_install/ubuntu-jammy/install-prebuilt-packages.sh
junhaoliao, line 1: i understand the added dependencies are not strictly "core" related but i believe some others here (e.g. unzip) are in a similar situation. we might want to split the package specific dependencies in anther refactoring PR, though i personally don't see too much benefit yet. any concerns / other ideas?

.github/workflows/clp-package-image-build.yaml
```diff
@@ -0,0 +1,103 @@
+name: "clp-package-image-build"
+
+on:
+  push:
+  workflow_dispatch:
+
+env:
+  DEPS_IMAGE_NAME_PREFIX: "clp-core-dependencies-x86-"
```
junhaoliao: (Similar to the comment at https://github.com/y-scope/clp/pull/1166/files#r2256441706)

I see how it can be awkward to build the package with a "core" dependencies container, but i haven't seen any quicker way to build the package. if there's no significant benefit in differentiating between "core" and "package" dependencies, shall we just unify them (simply rename the existing build env containers).

tools/docker-images/clp-package-ubuntu-jammy/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# Flatten the image
+FROM scratch
+COPY --from=base / /
+
+ENV PATH="/opt/clp/sbin:${PATH}"
+ENV PATH="/opt/clp/bin:${PATH}"
+ENV PYTHONPATH="/opt/clp/lib/python3/site-packages"
+ENV CLP_HOME="/opt/clp"
```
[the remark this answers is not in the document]
junhaoliao: wouldn't splitting into multiple lines be more VCS friendly?

tools/docker-images/clp-package-ubuntu-jammy/build.sh
kirkrodrigues, line 1: How about calling the directory just `clp-package`? Or do you anticipate that we'll need different container OSes for some reason?

tools/docker-images/clp-package-ubuntu-jammy/setup-scripts/install-prebuilt-packages.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -eu
+set -o pipefail
+
+apt-get update
+DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
+  ca-certificates \
+  libcurl4 \
+  libmariadb-dev \
+  python3
```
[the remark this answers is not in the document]
junhaoliao: with the replacement, the image size reduced from 822MB to 775MB

taskfile.yaml
```diff
@@ -162,6 +162,17 @@ tasks:
           CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
           INCLUDE_PATTERNS: ["{{.OUTPUT_DIR}}"]
 
+  docker-image-package:
```
kirkrodrigues: To avoid clutter, I would recommend moving this into a new taskfile, `taskfiles/docker-images.yaml`, and call it `package` (so it could be invoked with `docker-images:package`.

.github/actions/clp-image-build/action.yaml
```diff
@@ -1,8 +1,14 @@
-name: "clp-execution-image-build"
-description: "Builds a container image that contains the dependencies necessary
-to run the CLP package."
+name: "clp-image-build"
```
kirkrodrigues: How about `clp-runtime-image-build`? Otherwise, it's confusing why we have this action and also `clp-core-build-containers`.

.github/actions/clp-image-build/action.yaml
```diff
@@ -1,8 +1,14 @@
...
+description: "Builds a CLP container image."
```
kirkrodrigues:
```suggestion
description: "Builds a container image to be used for running CLP."
```

.github/actions/clp-image-build/action.yaml
```diff
@@ -1,8 +1,14 @@
...
 
 inputs:
+  image_type:
+    description: "Type of image to build"
+    required: true
+    type: "choice"
+    options:
+      - "execution"
+      - "package"
```
kirkrodrigues: These don't seem to be valid parameters in the context of [action metadata files](https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#inputs)

.github/actions/clp-image-build/action.yaml
```diff
@@ -50,15 +56,29 @@ runs:
...
+
+    - name: "Determine Dockerfile Path"
+      id: "dockerfile"
+      shell: "bash"
+      run: |
+        base_path="./tools/docker-images"
+        platform="${{inputs.platform_id}}-${{inputs.platform_version_id}}"
+
+        if [[ "${{inputs.image_type}}" == "execution" ]]; then
+          path="$base_path/clp-execution-base-$platform/Dockerfile"
```
kirkrodrigues: How about dockerfile_path to avoid confusion with "$PATH"?

.github/actions/run-on-image/action.yaml
kirkrodrigues: Can we add a comment about this?

.github/workflows/clp-package-image-build.yaml
kirkrodrigues: Yeah, I think we can rename the existing containers and make them sufficient for building the entire CLP package (since the requirements for building the package are not a lot more than what's required for building core). That said, let's defer this to another PR to avoid delaying this one.

.github/workflows/clp-package-image-build.yaml
```diff
@@ -0,0 +1,104 @@
...
+        env:
+          OS_NAME: "ubuntu-jammy"
+        with:
+          image_name: "${{env.DEPS_IMAGE_NAME_PREFIX}}${{env.OS_NAME}}"
+          docker_context: "components/core"
+          docker_file: "components/core/tools/docker-images/clp-env-base-${{env.OS_NAME}}\
+            /Dockerfile"
+          push_deps_image: >-
+            ${{github.event_name != 'pull_request' && github.ref == 'refs/heads/main'}}
+          token: "${{secrets.GITHUB_TOKEN}}"
```
kirkrodrigues: I don't think this is a good idea since it might cause some issues with two different workflows trying to publish the same container, not to mention the resources it'll use.

I think a quick solution might be to move this package image build into clp-core-build workflow.

.github/workflows/clp-package-image-build.yaml
```diff
@@ -0,0 +1,104 @@
...
+
+  ubuntu-jammy-package-image:
+    name: "ubuntu-jammy-package-image"
+    needs: "ubuntu-jammy-deps-image"
+    runs-on: "ubuntu-24.04"
+    env:
+      OS_NAME: "ubuntu-jammy"
```
kirkrodrigues: Can we move this into the scope of the step that requires it?

tools/docker-images/clp-package-ubuntu-jammy/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM ubuntu:jammy AS base
+
+COPY ./build/clp-package /opt/clp
+
+WORKDIR /root
+
+COPY ./tools/docker-images/clp-package-ubuntu-jammy/setup-scripts ./setup-scripts
+RUN ./setup-scripts/install-prebuilt-packages.sh && \
+    rm -rf ./setup-scripts/
```
kirkrodrigues:
```suggestion
RUN ./setup-scripts/install-prebuilt-packages.sh \
    && rm -rf ./setup-scripts/
```

tools/docker-images/clp-package-ubuntu-jammy/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# Flatten the image
+FROM scratch
+COPY --from=base / /
+
+ENV PATH="/opt/clp/sbin:${PATH}"
+ENV PATH="/opt/clp/bin:${PATH}"
+ENV PYTHONPATH="/opt/clp/lib/python3/site-packages"
+ENV CLP_HOME="/opt/clp"
```
kirkrodrigues:
```suggestion
ENV CLP_HOME="/opt/clp"
ENV PATH="${CLP_HOME}/sbin:${PATH}"
ENV PATH="${CLP_HOME}/bin:${PATH}"
ENV PYTHONPATH="${CLP_HOME}/lib/python3/site-packages"
```

tools/docker-images/clp-package-ubuntu-jammy/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+USER 1000:1000
```
kirkrodrigues: Why do we need this?

tools/docker-images/clp-package-ubuntu-jammy/build.sh
```diff
@@ -0,0 +1,35 @@
...
+elif [ "$arch" = "aarch64" ]; then
+    arch_name="arm64"
+else
+    echo "Error: Unsupported architecture - $arch"
+    exit 1
+fi
+
+build_cmd=(
+    docker build
+    --tag "clp-package-${arch_name}-ubuntu-jammy:dev"
```
kirkrodrigues: Similarly, how about naming the image `clp-package:dev`?

tools/docker-images/clp-package-ubuntu-jammy/setup-scripts/install-prebuilt-packages.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -eu
+set -o pipefail
+
+apt-get update
+DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
+  ca-certificates \
+  libcurl4 \
+  libmariadb3 \
+  python3
```
kirkrodrigues: Can we make the same changes in the execution image? I know we'll get rid of it soon, but just in case we don't, I don't want to forget.

.github/actions/clp-image-build/action.yaml
junhaoliao: how about `clp-build-runtime-image`? to be consistent with the word order in `clp-core-build-containers`

kirkrodrigues: Sure

.github/actions/clp-image-build/action.yaml
junhaoliao: i see. i was looking at https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputsinput_idtype and mistakenly thought they shared the same type

they're removed now

tools/docker-images/clp-package-ubuntu-jammy/build.sh
junhaoliao: sure. let's keep `clp-package` until we find need to maintain multiple bases at the same time, which seems unlikely so long our dev / runtime deps containers are working as expected

tools/docker-images/clp-package-ubuntu-jammy/build.sh
junhaoliao: sure. removing the arch name from the repo name / tag name seems quite beneficial as well

tools/docker-images/clp-package-ubuntu-jammy/Dockerfile
junhaoliao: In general it should be a good exercise to set the user to non-root to harden the security, though it might not with the way we use the image (see #1178 )

i can see how this can be very helpful if we plan to provide a single-container docker image in the future (instead of just docker-compose), though if at the moment this causes confusion, we should remove
```suggestion
```

.github/workflows/clp-package-image-build.yaml
junhaoliao: > to move this package image build into clp-core-build workflow

if we move it into clp-core-build, i believe we need to update the workflows trigger conditions. instead of an allow list, we might need to write it as
```
on:
  pull_request:
    paths: &monitored_paths
      - "**"
      - "!.github/ISSUE_TEMPLATE/**"
      - "!.github/*"
      - "!components/core/tools/scripts/lib_install/macos/**"
      - "!docs/**"
  push:
    paths: *monitored_paths
```

then the other jobs in the workflow will rely on `filter-relevant-changes` to see if they should run, as they currently do.

is my understanding correct? are there any other paths / patterns that need to be added to the deny list (`!`)?

tools/docker-images/clp-package-ubuntu-jammy/setup-scripts/install-prebuilt-packages.sh
junhaoliao: made the change in `tools/docker-images/clp-execution-base-ubuntu-jammy/setup-scripts/install-prebuilt-packages.sh` and validated with `execution_container: "clp-execution-x86-ubuntu-jammy:dev"` set in `clp-config.yaml`:
1. compress s3 file to s3 archive
2. (webUI) search s3 archive with aggregation
3. stream decompression onto s3

.github/workflows/clp-package-image-build.yaml
kirkrodrigues: That's correct. That pattern should be fine.

.github/actions/clp-build-runtime-image/action.yaml
```diff
@@ -17,11 +19,11 @@ inputs:
     required: false
   platform_id:
     description: "Platform ID of the container (e.g. ubuntu)"
-    required: true
+    required: false
```
junhaoliao, line 22: no longer needed for building the new image `clp-package`

.github/actions/clp-build-runtime-image/action.yaml
```diff
@@ -17,11 +19,11 @@ inputs:
...
   platform_version_id:
     description: "Platform VERSION_ID / VERSION_CODENAME of the container
       (e.g. jammy, focal, etc.)"
-    required: true
+    required: false
```
junhaoliao, line 26: no longer needed for building the new image `clp-package`

.github/workflows/clp-core-build.yaml
```diff
@@ -374,3 +361,41 @@ jobs:
...
+
+  package-image:
+    name: "package-image"
+    if: "!cancelled() && !failure() && (
+        needs.filter-relevant-changes.outputs.ubuntu_jammy_image_changed == 'false' ||
+        needs.ubuntu-jammy-deps-image.result == 'success'
+    )"
+    needs:
+      - "filter-relevant-changes"
+      - "ubuntu-jammy-deps-image"
```
junhaoliao, line 519: not quite certain if this works if `ubuntu-jammy-deps-image` fails its `if`

tools/docker-images/clp-package/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN apt-get clean \
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# Flatten the image
+FROM scratch
+COPY --from=base / /
+
+ENV CLP_HOME="/opt/clp"
+ENV PATH="${CLP_HOME}/sbin:${PATH}"
+ENV PATH="${CLP_HOME}/bin:${PATH}"
```
junhaoliao:
```suggestion
ENV PATH="${CLP_HOME}/bin:${PATH}"
ENV PATH="${CLP_HOME}/sbin:${PATH}"
```

.github/workflows/clp-package-image-build.yaml
junhaoliao: @coderabbitai create an issue to track the effort of creating a universal build env image

.github/actions/clp-build-runtime-image/action.yaml
```diff
@@ -45,20 +47,37 @@ runs:
...
+          image_name="clp-${{inputs.image_type}}-$platform"
+        else
+          dockerfile_path="$base_path/clp-package/Dockerfile"
+          image_name="clp-${{inputs.image_type}}"
+        fi
+
+        echo "DOCKERFILE=$dockerfile_path" >> "$GITHUB_OUTPUT"
+        echo "IMAGE_NAME=$image_name" >> "$GITHUB_OUTPUT"
+
     - name: "Update Metadata"
```
kirkrodrigues:
```suggestion
    - name: "Extract GitHub Metadata"
```

.github/actions/clp-build-runtime-image/action.yaml
```diff
@@ -45,20 +47,37 @@ runs:
...
-      id: "meta"
+      id: "update-meta"
```
kirkrodrigues: How about `extract-gh-meta`?

.github/actions/run-on-image/action.yaml
```diff
@@ -42,9 +42,11 @@ runs:
     - run: "./tools/scripts/deps-download/init.sh"
       shell: "bash"
 
+    # `--env npm_config_cache`: Override to prevent permission issues with `/.npm`.
```
kirkrodrigues:
```suggestion
    # `--env npm_config_cache` overrides the default `/.npm` to avoid permission issues.
```

.github/workflows/clp-core-build.yaml
```diff
@@ -374,3 +360,41 @@ jobs:
             .task/checksum/utils-cpp-lint-clang-tidy-*
             build/lint-clang-tidy
           key: "${{steps.cache-restore-lint-check-cpp-static-full.outputs.cache-primary-key}}"
+
+  package-image:
+    name: "package-image"
+    if: "!cancelled() && !failure() && (
+        needs.filter-relevant-changes.outputs.ubuntu_jammy_image_changed == 'false' ||
+        needs.ubuntu-jammy-deps-image.result == 'success'
+    )"
```
kirkrodrigues:
```suggestion
    if: >-
      !cancelled() && !failure() && (
        needs.filter-relevant-changes.outputs.ubuntu_jammy_image_changed == 'false' ||
        needs.ubuntu-jammy-deps-image.result == 'success'
      )
```

.github/workflows/clp-core-build.yaml
kirkrodrigues: Yeah, I'm not sure. We should probably test. I feel like that's why we used `success()` before, but maybe I'm wrong.

tools/docker-images/clp-package/build.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set -eu
+set -o pipefail
+
+script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+repo_root=${script_dir}/../../../
+
+build_cmd=(
+    docker build
+    --tag "clp-package:dev"
```
kirkrodrigues, line 11: Sorry, in hindsight, I think we should include the architecture in the image name (to clearly indicate to users that it's not an ARM image). What do you think?

.github/workflows/clp-core-build.yaml
junhaoliao: we could test more thoroughly, though what we have written should already work. see https://github.com/junhaoliao/clp/actions/runs/17405920984/workflow for a modification on an unrelated file, where `package-image` still runs with `Filter ubuntu_jammy_image = false`. I believe that's because we have this condition set up in `package-image`'s `if`: 
```
needs.filter-relevant-changes.outputs.ubuntu_jammy_image_changed == 'false' ||
needs.ubuntu-jammy-deps-image.result == 'success'
```

tools/docker-images/clp-package/build.sh
junhaoliao: 1. while i see misusing an image built on a different arch could harm performance, 
2. using a unified name should make our container orchestration infra platform agnostic.

i feel the benefits that (2.) brings outweighs the harm (1.) could potentially cause. (1.) should not be an issue once we start looking into building mutli platform images

docs/src/dev-docs/tooling-gh-workflows.md
```diff
@@ -90,6 +90,10 @@ This workflow builds CLP-core on macOS and runs its unit tests.
 This workflow builds a container image that contains the dependencies necessary to run the CLP
 package.
 
+## clp-package-image-build
+
+This workflow builds a container image that contains the CLP package.
```
kirkrodrigues: We need to update this file to reflect the latest changes, right?

.github/workflows/clp-core-build.yaml
kirkrodrigues, line 1: Since this is no longer just to build core, how about renaming it to something like `clp-artifact-build.yaml`?

kirkrodrigues (commented):
Last comments (hopefully).

.github/workflows/clp-core-build.yaml
junhaoliao: i intentionally left the `clp-core-macos` workflow unchanged since it still only builds the core binaries. let me know if that should be updated as well

docs/src/dev-docs/contributing-linting.md
```diff
@@ -32,7 +32,7 @@ To also apply any automatic fixes:
 task lint:fix
 ```
 
-[clp-core-build]: https://github.com/y-scope/clp/blob/main/.github/workflows/clp-core-build.yaml
+[clp-artifact-build]: https://github.com/y-scope/clp/blob/main/.github/workflows/clp-artifact-build.yaml
```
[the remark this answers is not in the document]
junhaoliao: the file should also has been renamed in the PR? could you double check?

kirkrodrigues (approved):
For the PR title, how about:

```
feat(docker): Add container image containing the CLP package, for Docker Compose integration (resolves #1164).
```

tools/docker-images/clp-package/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    && rm -rf ./setup-scripts/
+
+# Remove cached files
+RUN apt-get clean \
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+ENV CLP_HOME="/opt/clp"
+ENV PATH="${CLP_HOME}/bin:${PATH}"
+ENV PATH="${CLP_HOME}/sbin:${PATH}"
+ENV PYTHONPATH="${CLP_HOME}/lib/python3/site-packages"
```
junhaoliao, line 16: Below syntax reduces building layers while keeping VCS friendliness:
```
ENV KEY1=VAL1
    KEY2=VAL2
```

@coderabbitai create an issue for implementing the improvement

tools/docker-images/clp-package/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+USER 1000:1000
+
+COPY ./build/clp-package /opt/clp
+
+# Flatten the image
+FROM scratch
+COPY --from=base / /
```
junhaoliao, line 24: While currently the `USER` and `ENV` commands are unused, their effects would be lost after flattening the image. i.e., Those settings should be done after the image is flattened.

tools/docker-images/clp-package/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM ubuntu:jammy AS base
+
+WORKDIR /root
+
+COPY ./tools/docker-images/clp-package/setup-scripts ./setup-scripts
```
junhaoliao, line 5: @coderabbitai it appears that we should always use `COPY --link` to speed up builds. if you agree, file an issue

PR #1168, opened 2025-08-06
feat(webui): Create result metadata document in MongoDB to track search signal for Presto queries.

davemarco:
# Description

Create MongoDB document in results metadata collection to keep track of presto search state. 

Using the same collection for regular clp and presto since the document format is pretty much identical, and can query a single collection rather than two for aggregate statistics. I added a new field just in case we need to distinguish between the two in the future. Also using presto search signal type instead of clp search signal types.

# Checklist

  breaking change.

# Validation performed

Document created and state changes based on status.

## Summary by CodeRabbit

* **New Features**
  * Enhanced tracking and storage of search query engine type and metadata for search results, including support for Presto queries.
* **Bug Fixes**
  * Improved handling and typing of job identifiers and search signals, reducing potential issues with search job updates.
* **Chores**
  * Updated configuration to set the default query engine to "clp".
  * Streamlined code by consolidating query engine definitions and updating type usage for consistency.
  * Simplified search results metadata updates by removing deprecated utility functions.
  * Adjusted search controls rendering logic based on query engine settings.

### DIFF
--- a/docs/src/user-guide/guides-using-presto.md
+++ b/docs/src/user-guide/guides-using-presto.md
@@ -31,8 +31,32 @@ Using Presto with CLP requires:
 
 ### Setting up CLP
 
-Follow the [quick-start](./quick-start/index.md) guide to set up CLP and compress your logs. A
-sample dataset that works well with Presto is [postgresql].
+1. Follow the [quick-start](./quick-start/index.md) guide to download and extract the CLP package,
+   but don't start the package just yet.
+2. Before starting the package, update the package's config as follows:
+
+    * Open `etc/clp-config.yml` located within the package.
+    * Uncomment the `database` section.
+    * Change `database.host` value to a non-localhost hostname/IP.
+    * After the change, the `database` section should look something like this:
+
+      ```yaml
+      database:
+        type: "mariadb"  # "mariadb" or "mysql"
+        host: "<new-IP-address>"
+        port: 3306
+        name: "clp-db"
+      ```
+
+    :::{note}
+    This change is necessary since the Presto containers run on a Docker network, whereas CLP's
+    database runs on the host network. So `localhost` refers to two different entities in those
+    networks. This limitation will be addressed in the future when we unify Presto and CLP's
+    deployment infrastructure.
+    :::
+
+3. Continue following the [quick-start](./quick-start/index.md#using-clp) guide to start CLP and
+   compress your logs. A sample dataset that works well with Presto is [postgresql].
 
 ### Setting up Presto
 


components/webui/server/src/routes/api/presto-search/index.ts
```diff
@@ -100,9 +111,29 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
 
+                            // Insert metadata and resolve queryId on first call
                             if (false === isResolved) {
+                                searchResultsMetadataCollection.insertOne({
+                                    _id: queryId,
+                                    lastSignal: stats.state,
+                                    errorMsg: null,
+                                    queryEngine: CLP_QUERY_ENGINES.PRESTO,
+                                }).catch((err: unknown) => {
+                                    request.log.error(err, "Failed to insert Presto metadata");
```
hoophalab, line 121: probably need to simply await the `insertOne` function call here, so that uncaught error rejects the promise

components/webui/server/src/routes/api/presto-search/index.ts
```diff
@@ -50,6 +60,7 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
             let searchJobId: string;
 
             try {
+                // eslint-disable-next-line max-lines-per-function
                 searchJobId = await new Promise<string>((resolve, reject) => {
                     let isResolved = false;
                     Presto.client.execute({
```
hoophalab, line 65: I'm running into this bug:
1. run a long query: e.g. `SELECT * FROM tpch.sf10000.orders`
2. the presto query timed out (why can it time out?)
Here are the logs
```
[App] [16:43:47.390] INFO (3631010): Received 8092 rows from Presto query
[App]     reqId: "req-1"
[App] [16:43:48.251] INFO (3631010): Received 8083 rows from Presto query
[App]     reqId: "req-1"
[App] [16:43:49.129] INFO (3631010): Received 8092 rows from Presto query
[App]     reqId: "req-1"
[App] [16:43:49.675] INFO (3631010): Presto search failed
[App]     reqId: "req-1"
[App]     message: "execution error:query timed out"
```
3. `presto-client-node` library only emits a error callback and doesn't emit a `state` callback for this (why?), so mongo db isn't updated.

components/webui/server/src/routes/api/search/utils.ts
```diff
@@ -38,11 +38,14 @@ const updateSearchResultsMeta = async ({
...
+        _id: jobId,
     };
 
+    if (lastSignal) {
+        filter.lastSignal = lastSignal;
+    }
+
     const modifier = {
         $set: fields,
     };
```
hoophalab: I would highly recommend just inline this function in both `routes/api/search` in `routes/api/presto-search`

Compared with how much hassle needed to make `updateSearchResultsMeta` work for both endpoints,
we only need one line in both endpoints
```
await searchResultsMetadataCollection.updateOne({_id: queryId}, {lastSignal: stats.state});
```

We are already calling `searchResultsMetadataCollection.insertOne` directly in `presto-serach` anyway

components/webui/server/src/routes/api/presto-search/index.ts
davemarco: i added a flag to remove the timeout. Looking at their code, it looks like they just delete the request when the timeout is reached, which is why the state isnt updated. We could set the state in the error callback, but i think better just to remove timeout

davemarco: nice catch

components/webui/server/src/routes/api/presto-search/index.ts
davemarco: i thinks its fine not to await. then we need to await on every call. There is already error log, and if this breaks, there will be a few other errors in the front end that this is missing.

components/webui/server/src/routes/api/search/utils.ts
davemarco: okay i changed

davemarco:
yes  - working on error state in other PR

@junhaoliao can you approve

kirkrodrigues (approved):

components/webui/client/public/settings.json
```diff
@@ -1,6 +1,6 @@
 {
     "ClpStorageEngine": "clp",
-    "ClpQueryEngine": "native",
+    "ClpQueryEngine": "clp",
```
junhaoliao, line 3: we should also configure this in `server_settings_json_updates` in `components/clp-package-utils/clp_package_utils/scripts/start_clp.py`

Issue #1045, opened 2025-06-27
feat(package): Handle tasks in `RUNNING` state on compression scheduler restart.


gibber9809:
### Request

When the compression scheduler restarts some jobs in the database may already be in `RUNNING` state because:
1. The compression scheduler crashed/terminated and lost track of a job that could either be completed or still running
2. The compression scheduler marked a job as `RUNNING` and subsequently restarted, but no workers attempted to run the task (e.g. because they themselves were shut down)

The second scenario is very likely to happen after #1037 is implemented if the compression scheduler and compression workers are on the same machine, because we first shut down the workers before attempting to shut down the compression scheduler. This makes it likely that for some period of time the compression scheduler continues to issue jobs after the compression workers stop accepting new jobs.

### Possible implementation

This problem can be solved in two parts:

1. On restart, dispatch any jobs in `RUNNING` state to the compression workers
* This allows us to retry jobs that either finished after the compression scheduler crashed/terminated, or never started because workers were not accepting jobs
* This wouldn't allow us to address the scenario where a job is dispatched before the compression scheduler crashes and terminates after the compression scheduler restarts, but handling this case well would likely complicate state management

2. Attempt to gracefully terminate the compression scheduler and compression workers at the same time to minimize the period where the compression scheduler continues dispatching jobs while the compression workers no longer accept jobs.

kirkrodrigues:
For proposal 1, I would prefer we don't try to retry the jobs since, if they ran partially, we will have an inconsistent amount of data compressed; and then retrying the job may duplicate some of that data. Before we have more robust fault tolerance, the simpler (and I think intuitive for users) solution would seem to be what @haiqi96 suggested offline---mark the job as failed.

junhaoliao:
note that the container orchestration mechanism will soon be replaced by Docker Compose, where containers are run as serviced and are automatically restarted individually depending on their own health statuses. e.g., if a compression worker crashes, it will automatically restart without restarting the the scheduler. let's factor that in when finalizing the solution for this issue

haiqi96:
> note that the container orchestration mechanism will soon be replaced by Docker Compose, where containers are run as serviced and are automatically restarted individually depending on their own health statuses. e.g., if a compression worker crashes, it will automatically restart without restarting the the scheduler. let's factor that in when finalizing the solution for this issue

[...]

PR #1169, opened 2025-08-06
feat(package): Add support for gracefully shutting down the compression scheduler and compression workers (resolves #1037)

### LINKED ISSUE #1037: feat(package): Add support for gracefully shutting down the compression scheduler and compression workers.
### Request

Currently our stop package scripts will stop every container running on a given machine in dependency order using `docker stop`, which [by default](https://docs.docker.com/reference/cli/docker/container/stop/) will send the main process inside of the container `SIGTERM` followed by `SIGKILL` after a 10 second grace period.

Unfortunately, the compression scheduler currently has no signal handler to initiate a graceful shutdown on `SIGTERM`, and a 10 second grace period is insufficient  for a graceful shutdown in our system. This means that the current setup can leave jobs in a bad state or potentially duplicate or lose data if a shutdown occurs while compression jobs are running.

### Possible implementation

To ensure that running compression jobs complete before shutdown the following two conditions should be sufficient, even in a distributed setup:
1. Upon receiving `SIGTERM` a compression worker finishes any currently running job and shuts down without fetching a new job
2. Upon receiving `SIGTERM` the compression scheduler stops dispatching new jobs and waits for currently running jobs to finish before terminating.

Since we can not guarantee liveness we must also support forcefully terminating after some grace period (implemented by sending a `SIGKILL` or otherwise).

Celery workers already [register signal handlers that do most of what we want](https://docs.celeryq.dev/en/stable/userguide/workers.html#warm-shutdown) by default (except they seem to use `SIGTERM`/`SIGINT`/`SIGQUIT`).

We would also need to write a signal handler for the compression scheduler that handles graceful shutdown as described.

After that it would just be a matter of modifying our stop script to send the compression worker and scheduler a "soft" kill signal followed by a "hard" kill signal after a reasonable timeout (e.g. 5 minutes). The "hard" timeout could be made  configurable via command line argument on the package stop script.

AVMatthews:
# Description

- Enables the warm shutdown feature on celery workers
- Add signal handlers to compression scheduler and compression worker. The compression scheduler signal handler will stop the scheduler from scheduling new jobs and will instead just poll existing jobs.
- Extends docker shutdown timeouts for current running compression tasks to complete gracefully. 
-> 5 min for scheduler and 1 min for workers
- Reorder shutdowns so that the compression scheduler is shut down before the compression workers.
-> The docker stop commands will wait the full timeout time if the processes on the docker take advantage of it and no further docker stop commands will be given. So with the new ordering the scheduler is stopped first with 5 mins of waiting to hear from existing jobs. The workers are stopped after with 60 additional seconds to wrap up it's current job. If you have an especially long job, this can still lead it issues. Issue #1045 speaks to how we may be able to address this in the future. 

# Checklist

  breaking change.

# Validation performed

- Testing starting and stopping the package both with jobs and without. 
- Saw that the scheduler stopped assigning new jobs, but continues to poll for already assigned jobs.
- Double checked the the successful jobs were still readable and searchable using the webui.


components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -49,6 +50,18 @@
 
 scheduled_jobs = {}
 
+recieved_signal = False
+
+
+def signal_handler(sig, frame):
+    global recieved_signal
+    recieved_signal = True
```
gibber9809:
```suggestion
def sigterm_handler(sig, frame):
    global received_sigterm
    received_sigterm = True
```

Since we're only handling SIGTERM I think it makes sense to give more specific names.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -49,6 +50,18 @@
 
 scheduled_jobs = {}
 
+recieved_signal = False
```
gibber9809:
```suggestion
received_sigterm = False
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -49,6 +50,18 @@
...
+
+def signal_handler(sig, frame):
+    global recieved_signal
+    recieved_signal = True
+    logger.info("Received SIGTERM in compression scheduler")
+
+
+# Register the signal handler
+signal.signal(signal.SIGTERM, signal_handler)
+
```
gibber9809: I think we should register the handler in main right after we set up the logger since we use the logger in our signal handler.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -430,13 +443,14 @@ def main(argv):
...
         while True:
             try:
-                search_and_schedule_new_tasks(
-                    db_conn,
-                    db_cursor,
-                    clp_metadata_db_connection_config,
-                    clp_config.archive_output,
-                    existing_datasets,
-                )
+                if False == recieved_signal:
```
gibber9809:
```suggestion
                if not received_sigterm:
```
Seems like we use `if not` more than `if False ==` more often than not for python (though this file uses `False ==` a bunch).

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -49,6 +50,18 @@
 
 scheduled_jobs = {}
 
+recieved_signal = False
+
+
+def signal_handler(sig, frame):
+    global recieved_signal
+    recieved_signal = True
+    logger.info("Received SIGTERM in compression scheduler")
```
gibber9809: Seems like we have a warn log in the compression worker. We should probably consistently use info or warn for both.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -430,13 +443,14 @@ def main(argv):
...
+                if False == recieved_signal:
+                    search_and_schedule_new_tasks(
+                        db_conn,
+                        db_cursor,
+                        clp_metadata_db_connection_config,
+                        clp_config.archive_output,
+                        existing_datasets,
+                    )
                 poll_running_jobs(db_conn, db_cursor)
                 time.sleep(clp_config.compression_scheduler.jobs_poll_delay)
```
gibber9809, line 527: I think if we received sigterm and there are no more running jobs we should exit here -- no reason to wait the full 5 minutes once running jobs are done.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -430,13 +443,14 @@ def main(argv):
...
             except KeyboardInterrupt:
```
gibber9809, line 528: While we're making this change we should probably either change the log in the next line to "Forcefully shutting down." or actually make it graceful.

gibber9809 (changes requested):
Nice work! Main thing is that we also need to have the compression scheduler exit after it has received SIGTERM and all running jobs have finished. Besides that I just had a few small style comments.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -408,6 +404,10 @@ def poll_running_jobs(db_conn, db_cursor):
     for job_id in jobs_to_delete:
         del scheduled_jobs[job_id]
 
+    if received_sigterm and 0 == len(scheduled_jobs):
+        logger.info("Recieved sigterm, No more running jobs. Exiting.")
```
gibber9809, line 55:
```suggestion
        logger.info("Received sigterm, no more running jobs. Exiting.")
```
Minor typo.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -444,16 +460,17 @@ def main(argv):
...
-                    db_cursor,
-                    clp_metadata_db_connection_config,
-                )
+                if not received_sigterm:
+                    search_and_schedule_new_tasks(
+                        clp_config,
+                        db_conn,
+                        db_cursor,
+                        clp_metadata_db_connection_config,
+                    )
```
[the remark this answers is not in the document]
gibber9809: As far as I can tell this race is benign. If we schedule a few more jobs before shutdown it shouldn't be a problem.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -55,6 +56,14 @@
 
 scheduled_jobs = {}
 
+received_sigterm = False
+
+
+def sigterm_handler(sig, frame):
```
kirkrodrigues: Instead of `sig` (which could be misinterpreted as "signature"), how about `signal_number`?

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -55,6 +56,14 @@
...
+    global received_sigterm
+    received_sigterm = True
+    logger.info("Received SIGTERM in compression scheduler")
```
kirkrodrigues: Won't the logs be printed with information about the caller of the logging statement (and they're written to component-specific files anyway)? If so, I think this is sufficient.

```suggestion
    logger.info("Received SIGTERM.")
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -395,6 +404,10 @@ def poll_running_jobs(db_conn, db_cursor):
     for job_id in jobs_to_delete:
         del scheduled_jobs[job_id]
 
+    if received_sigterm and 0 == len(scheduled_jobs):
+        logger.info("Received sigterm, no more running jobs. Exiting.")
```
```suggestion
        logger.info("Received SIGTERM and there're no more running jobs. Exiting.")
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -410,6 +423,9 @@ def main(argv):
     # Update logging level based on config
     set_logging_level(logger, os.getenv("CLP_LOGGING_LEVEL"))
 
+    # Register the sigterm handler
```
```suggestion
    # Register the SIGTERM handler
```

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -522,6 +523,11 @@ def run_clp(
         return CompressionTaskStatus.FAILED, worker_output
 
 
+@signals.worker_shutdown.connect
+def worker_shutdown_handler(signal=None, sender=None, **kwargs):
+    logger.info("Worker shutdown signal received.")
```
kirkrodrigues: Similar to the scheduler log, I think this should be sufficient:

```suggestion
    logger.info("Shutdown signal received.")
```

components/clp-package-utils/clp_package_utils/scripts/stop_clp.py
```diff
@@ -34,10 +34,12 @@
...
 
 
-def stop_running_container(container_name: str, already_exited_containers: List[str], force: bool):
+def stop_running_container(
+    container_name: str, already_exited_containers: List[str], force: bool, timeout: int = 10
+):
     if is_container_running(container_name):
         logger.info(f"Stopping {container_name}...")
-        cmd = ["docker", "stop", container_name]
+        cmd = ["docker", "stop", "-t", str(timeout), container_name]
```
kirkrodrigues:
```suggestion
        cmd = ["docker", "stop", "--timeout", str(timeout), container_name]
```

We prefer long-form options.

PR #1170, opened 2025-08-06
feat(Presto): Generate split filter file from `set-up-config.sh` command

quinntaylormitchell:
# Description

At the moment, a CLP + Presto user must configure Presto to use CLP’s metadata database by opening  `split-filter.json` and adding a filter for each dataset they wish to query (see the [Presto setup instructions](https://docs.yscope.com/clp/main/user-guide/guides-using-presto.html#setting-up-presto) for more information). Furthermore, within each filter, the user must correctly replace the `<dataset>` and `<timestamp-key>` fields with the correct corresponding information. This process is error-prone, and it can become time-consuming if the user has compressed (and wishes to query) many datasets.

This PR pads this process for the user by generating the filters for the `split-filter.json` file from a new script called `generate-split-filter-file.py`, which is run from `set-up-config.sh`. The user now only has to run the `scripts/set-up-config.sh <clp-json-dir>` command specified in step 3 of the [Presto setup instructions](https://docs.yscope.com/clp/main/user-guide/guides-using-presto.html#setting-up-presto), and then they will be prompted to enter the timestamp key for each of their datasets they've compressed (i.e., each of the first-level subdirectories in `clp-package/var/data/archives`).

# Checklist

  breaking change.

# Validation performed

After starting the package, I performed the following two compressions:

```bash
./sbin/compress.sh --timestamp-key 'timestamp' --dataset 'postgresql' /home/quinnmitchell/logs/postgresql/postgresql.log \
&& ./sbin/compress.sh --timestamp-key 'timestamp' --dataset 'cockroachdb' /home/quinnmitchell/logs/cockroachdb/cockroach.node1.log
```

Then I ran `./scripts/set-up-config.sh /home/quinnmitchell/clp/build/clp-package` from `clp/tools/deployment/presto-clp`, and received the following output:

```bash
Installing required Python packages...
Requirement already satisfied: python-dotenv in ./scripts/.venv/lib/python3.10/site-packages (from -r /home/quinnmitchell/clp/tools/deployment/presto-clp/scripts/requirements.txt (line 1)) (1.1.1)
Requirement already satisfied: PyYAML in ./scripts/.venv/lib/python3.10/site-packages (from -r /home/quinnmitchell/clp/tools/deployment/presto-clp/scripts/requirements.txt (line 2)) (6.0.2)
Generating config files corresponding to user-configured properties...
Generating split filter file for user-configured datasets...

Please enter the timestamp key that corresponds to each of your archived datasets.
Press <Enter> to accept the default key.

>>> cockroachdb [default key: timestamp]: timestamp
>>> postgresql [default key: timestamp]: timestamp
```

The `split-filter.json` that was generated from that command looks like this:

```json
{
  "clp.default.cockroachdb": [
    {
      "columnName": "timestamp",
      "customOptions": {
        "rangeMapping": {
          "lowerBound": "begin_timestamp",
          "upperBound": "end_timestamp"
        }
      },
      "required": false
    }
  ],
  "clp.default.postgresql": [
    {
      "columnName": "timestamp",
      "customOptions": {
        "rangeMapping": {
          "lowerBound": "begin_timestamp",
          "upperBound": "end_timestamp"
        }
      },
      "required": false
    }
  ]
}
```

and the output from running the `SHOW TABLES;` command from within the Presto CLI is as follows:

```SQL
    Table    
-------------
 cockroachdb 
 postgresql  
(2 rows)
```

## Summary by CodeRabbit

- **New Features**
  - Added a CLI to generate per-dataset split‑filter JSONs: discovers configured datasets, prompts for per‑dataset timestamp keys, validates inputs, logs progress, and writes a formatted output.

- **Chores**
  - Deployment setup now also generates a metadata‑filter configuration file alongside environment variables during setup.

---
  - https://app.asana.com/0/0/1212888478482885


quinntaylormitchell:
@coderabbitai review

quinntaylormitchell:
@coderabbitai review

haiqi96:
The PR claims that `This process is error-prone, and it can become time-consuming if the user has compressed (and wishes to query) many datasets.`
From high-level, however, I feel the process proposed in this PR is also very error prone. User will have to type dataset and timestamp key correctly one by one. and if they make one mistake, they can't rollback.

I feel the real issue is that the config file we ask user to edit is overly complicated. since it contains a lot of unrelated key-value pairs.

What if we let user specifies a simple yaml, or even plain text file with only `dataset:timestamp-key` pair, and parse that file instead?

quinntaylormitchell:
>User will have to type dataset and timestamp key correctly one by one

@haiqi96 fwiw, this PR proposes a process where the user only has to type out the timestamp; the dataset name is spoonfed. For example, if I had already compressed two datasets, `cockroachdb` and `postgresql`, then this is the prompt I'd get:

```text
Please enter the timestamp key that corresponds to each of your archived datasets.
Press <Enter> to accept the default key.

>>> cockroachdb [default key: timestamp]: timestamp
>>> postgresql [default key: timestamp]: timestamp
```

all I would have to type for each dataset is "timestamp", which the user will have to do anyway regardless of what method we use.

That being said, I do agree that the inability for the user to roll back is a downside of this method, that's a good point. And I think that your idea re. letting the user specify a simple yaml or something has merit. @kirkrodrigues what do you think?

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+    rangeMapping: RangeMapping
+    required: bool
+
+
+MetadataFilterDict = Dict[str, List[FilterRule]]
+
+
+def main(argv=None) -> int:
+    if argv is None:
+        argv = sys.argv
```
haiqi96: just looking at the code, I am not sure why we need this. Since this copied from the other script, I assume we have a valid reason.

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+    args_parser.add_argument(
+        "--clp-package-dir", help="CLP package directory.", required=True, type=Path
+    )
+    args_parser.add_argument(
+        "--output-file", help="Path for the metadata filter file.", required=True, type=Path
+    )
+
+    parsed_args = args_parser.parse_args(argv[1:])
+    clp_package_dir: Path = parsed_args.clp_package_dir.resolve()
+    json_output_file: Path = parsed_args.output_file
```
haiqi96: I know you copied this from the other script, but I still feel it is better to either not add resolve(), or add it to both of them.
```suggestion
    json_output_file: Path = parsed_args.output_file.resolve()
```

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+    ):
+        return 1
+
+    with open(json_output_file, "w") as json_file:
+        json.dump(metadata_filters, json_file, indent=2)
+
+    return 0
+
+
+def _get_dataset_names(clp_package_dir: Path, datasets: List[str]) -> bool:
```
haiqi96: just looking at the implementations. I feel the function can return an Optional[List[str]] instead.
if the dataset is somehow invalid, return None. I think this can make your code a little cleaner, and you don't need to predeclare `datasets: List[str] = []`

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+    parsed_args = args_parser.parse_args(argv[1:])
+    clp_package_dir: Path = parsed_args.clp_package_dir.resolve()
+    json_output_file: Path = parsed_args.output_file
+
+    datasets: List[str] = []
+    if not _get_dataset_names(clp_package_dir, datasets):
+        return 1
+
+    try:
+        timestamp_keys_by_dataset: Dict[str, str] = _prompt_timestamp_keys(datasets)
```
haiqi96: you don't need type annotation because `_prompt_timestamp_keys` has explicit return type.
```suggestion
        timestamp_keys_by_dataset = _prompt_timestamp_keys(datasets)
```

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+    :param metadata_filters:
+    :param datasets:
+    :param timestamp_keys: Mapping from dataset name -> timestamp key.
+    :param lower_bound_key:
+    :param upper_bound_key:
+    :param required:
+    :return: True on success, False on error.
+    """
+
+    if not datasets:
```
haiqi96: iirc, we prefer explicit check if condition is not already type bool (here python is doing an implicit conversion for you). 
```suggestion
    if len(datasets) == 0:
```

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+        logger.error("No datasets provided to _construct_metadata_filters.")
+        return False
+
+    missing = [d for d in datasets if d not in timestamp_keys]
+    if missing:
```
haiqi96: ditto

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+    :param upper_bound_key:
+    :param required:
+    :return: True on success, False on error.
+    """
+
+    if not datasets:
+        logger.error("No datasets provided to _construct_metadata_filters.")
+        return False
+
+    missing = [d for d in datasets if d not in timestamp_keys]
```
haiqi96: looking at your code, missing should always be empty?
I think there's some room to argue it's safer to have this check, but in that case I would prefer we directly throw because it means something is fundamentally wrong

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+    return data_time_pair
+
+
+def _construct_metadata_filters(
+    metadata_filters: MetadataFilterDict,
+    datasets: List[str],
+    *,
+    timestamp_keys: Dict[str, str],
+    lower_bound_key: str = "begin_timestamp",
+    upper_bound_key: str = "end_timestamp",
```
haiqi96: Instead of specifying those two params, can we pass in a RangeMapping directly?

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+    required: bool = False,
```
haiqi96: I might be missing context, do we need to specify True/False explicitly for certain case, or is it here for future.

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+                "lowerBound": lower_bound_key,
+                "upperBound": upper_bound_key,
+            }
+            rule: FilterRule = {
+                "columnName": timestamp_keys[dataset],
+                "rangeMapping": range_mapping,
+                "required": required,
+            }
+            metadata_filters[f"clp.default.{dataset}"] = [rule]
+    except Exception:
```
haiqi96: what exception do we expect here? timestamp_keys[dataset] shouldn't cause any exception since we have verified that the key exists in `timestamp_keys`.

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+        user_input = input(
+            f">>> {BOLD}{dataset}{RESET} [default key: {BOLD}{default_key}{RESET}]: "
+        )
+        key = user_input if user_input else default_key
+        data_time_pair[dataset] = key
+
+    return data_time_pair
+
+
+def _construct_metadata_filters(
```
haiqi96: same comment as _get_dataset_names, we can return Optional[MetadataFilterDict] (unless I missed a reason why we want to update metadata_filters by reference)

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+            }
+            rule: FilterRule = {
+                "columnName": timestamp_keys[dataset],
+                "rangeMapping": range_mapping,
+                "required": required,
+            }
+            metadata_filters[f"clp.default.{dataset}"] = [rule]
+    except Exception:
+        logger.exception("Unexpected error while constructing metadata filters.")
+        metadata_filters.clear()
```
haiqi96: This might become irrelevant if you take my previous suggestion. but we don't need to clear the dictionary since once the function return False, the caller will exit without using metadata_filters.

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+    dataset_names = sorted([p.name for p in archives_dir.iterdir() if p.is_dir()])
+    if not dataset_names:
+        logger.warning("No datasets found in %s. Did you compress any logs yet?", archives_dir)
+        return False
+
+    datasets[:] = dataset_names
+    return True
+
+
+def _prompt_timestamp_keys(
```
haiqi96: Unless you expect this function to be used at some other places with a default timestamp key, otherwise we should:
1. Define default_timestamp_key as a Final[str] constant at the top of the file.
2. enforce caller to pass in `default_key`, instead of allowing a default.

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+
+    BOLD = "\033[1m"
+    RESET = "\033[0m"
+
+    data_time_pair: Dict[str, str] = {}
+    for dataset in datasets:
+        user_input = input(
+            f">>> {BOLD}{dataset}{RESET} [default key: {BOLD}{default_key}{RESET}]: "
+        )
+        key = user_input if user_input else default_key
```
haiqi96: ditto, please do explitict type checking. I assume it should be `default_key if user_input is None else user_input` or `key = user_input if user_input is not None else default_key`

Or maybe it should check if len(user_input) != 0. either way, please use the explicit condition

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,190 @@
...
+
+    :param datasets:
+    :param default_key:
+    :return: Dictionary mapping dataset names to their timestamp keys.
+    """
+    print("\nPlease enter the timestamp key that corresponds to each of your archived datasets.")
+    print("Press <Enter> to accept the default key.\n")
+
+    BOLD = "\033[1m"
+    RESET = "\033[0m"
```
haiqi96:
```suggestion
    BOLD = "\033[1m"
    RESET = "\033[0m"
```

Kirk may suggest you to move them to the top of file. I feel it's fine to have them here, but please annotate them as Final[str] and move to the top of the function. (before print).

haiqi96:
Quick question; What will happen if user hit ctrl+c, or anything that forces the input() to terminate? Will script return a failure?

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
quinntaylormitchell: I think it lets you call `main` by itself if you were testing the function on its own (you could specify your own arguments), and then if you don't specify your own arguments, `argv = sys.argv` makes it so that the command line arguments are passed into `argv`. not super necessary as we probably won't be testing this by calling `main` by itself; I did just copy it from the other script, you're right

haiqi96: from what i understand, we shouldn't allow a script to be called both as <script>.py and <script>.main, since we are implying this script is both a module and a valid entry point, which is weird.

this is more like a comment for justifying why I don't try to suggest any change to this. Since it's also in another script, I would rather fix them all together in a separate PR (if we ever decide to change this)

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
quinntaylormitchell: even though it does feel better (I agree), the `set-up-config.sh` script which calls this code ensures that the `json_output_file` path is already resolved, but the `clp_package_dir` path is just user inputted, so it may be unresolved. maybe it's one of those things that we'd rather fix all together in a separate PR (like your comment on the `if argv is None:` section).

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
quinntaylormitchell: right on, thank you

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
haiqi96: Sure, sounds good

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
quinntaylormitchell: definitely

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
quinntaylormitchell: the latter; it's to match the format of the filter config [here](https://docs.yscope.com/clp/main/user-docs/guides-using-presto.html#setting-up-presto), and I'm not sure if that will change in the future. kinda violates the "don't add anything you don't need" rule though, so let me know if you think I should get rid of it (and just hard-code "False" into the `FilterRule`).

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
quinntaylormitchell: agreed!

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
quinntaylormitchell: very true, I will change it to throw

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
quinntaylormitchell: the default values are now all specified at the top of the file

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
quinntaylormitchell: true, I'll remove it

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
quinntaylormitchell: agreed!

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
quinntaylormitchell: I took your previous suggestion

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
quinntaylormitchell: default values are now defined at the top of the file

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
quinntaylormitchell: moved to the top of the file!

quinntaylormitchell:
>What will happen if user hit ctrl+c, or anything that forces the input() to terminate?

@haiqi96 this will happen:

```text
quinnmitchell@baker22:/home/quinnmitchell/clp/tools/deployment/presto-clp$ ./scripts/set-up-config.sh /home/quinnmitchell/clp/build/clp-package

... # other output not relevant to this

Please enter the timestamp key that corresponds to each of your archived datasets.
Press <Enter> to accept the default key.

>>> cockroachdb [default key: timestamp]: timestamp
>>> postgresql [default key: timestamp]: ^C2025-09-03T22:07:13.564 ERROR [generate-split-filter-file] Interrupted while collecting timestamp keys.
quinnmitchell@baker22:/home/quinnmitchell/clp/tools/deployment/presto-clp$ 
```

quinntaylormitchell:
@coderabbitai review

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,189 @@
...
+    metadata_filters: MetadataFilterDict = {}
+    if not _construct_metadata_filters(
+        metadata_filters, datasets, timestamp_keys=timestamp_keys_by_dataset
+    ):
+        return 1
+
+    with open(json_output_file, "w") as json_file:
+        json.dump(metadata_filters, json_file, indent=2)
+
+    return 0
```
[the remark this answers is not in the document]
quinntaylormitchell: the `set-up-config.sh` script which calls this code ensures that the `json_output_file` directory exists, so this is not strictly necessary

tools/deployment/presto-clp/scripts/generate-metadata-filter-file.py
```diff
@@ -0,0 +1,189 @@
...
+    args_parser = argparse.ArgumentParser(
+        description="Generates a metadata filter file for all user-configured datasets."
+    )
+    args_parser.add_argument(
+        "--clp-package-dir", help="CLP package directory.", required=True, type=Path
+    )
+    args_parser.add_argument(
+        "--output-file", help="Path for the metadata filter file.", required=True, type=Path
+    )
+
```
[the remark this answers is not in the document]
quinntaylormitchell: maybe in the future, but this script is probably not going to exist for a long time

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,169 @@
...
+SplitFilterDict = Dict[str, List[SplitFilterRule]]
+
+
+DEFAULT_TIMESTAMP_KEY: Final[str] = "timestamp"
+DEFAULT_RANGE_MAPPING: Final[RangeMapping] = {
+    "lowerBound": "begin_timestamp",
+    "upperBound": "end_timestamp",
+}
+DEFAULT_CUSTOM_OPTIONS: Final[CustomOptions] = {"rangeMapping": DEFAULT_RANGE_MAPPING}
+DEFAULT_REQUIRED: Final[bool] = False
```
haiqi96: This might be an overkill, but doesn't hurt to keep it here

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+BOLD: Final[str] = "\033[1m"
+RESET: Final[str] = "\033[0m"
```
haiqi96: how about 
```suggestion
ANSI_BOLD: Final[str] = "\033[1m"
ANSI_RESET: Final[str] = "\033[0m"
```

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,169 @@
...
+    if not archives_dir.exists():
+        logger.error("Archives directory does not exist: %s", archives_dir)
+        return 1
+
+    if not archives_dir.is_dir():
+        logger.error("Archives path is not a directory: %s", archives_dir)
+        return 1
+
+    datasets = _get_dataset_names(archives_dir)
+    if None == datasets:
```
haiqi96: iirc, using `is` better for None
```suggestion
    if datasets is None:
```

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,169 @@
...
+    Constructs a split filter for each dataset using a per-dataset timestamp key.
+
+    :param datasets:
+    :param timestamp_keys: Mapping from dataset name -> timestamp key.
+    :return: A SplitFilterDict containing all the SplitFilterRule objects for the JSON file
+    """
+
+    missing = [d for d in datasets if d not in timestamp_keys]
+    if 0 < len(missing):
+        raise ValueError()
```
haiqi96: we shouldn't abuse using Exceptions for error handling. Also, this contradicts with the return type (you don't return None anywhere)
Imo, we should log the missing keys, and return None.

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,169 @@
...
+class SplitFilterRule(TypedDict):
+    columnName: str
+    customOptions: CustomOptions
+    required: bool
+
+
+SplitFilterDict = Dict[str, List[SplitFilterRule]]
+
+
+DEFAULT_TIMESTAMP_KEY: Final[str] = "timestamp"
```
haiqi96: let's add #Constants in front of constant section. and I believe that constant should be in front of types.
you can refer to https://github.com/y-scope/clp/blob/main/components/clp-package-utils/clp_package_utils/general.py for order and format

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,169 @@
...
+DEFAULT_RANGE_MAPPING: Final[RangeMapping] = {
+    "lowerBound": "begin_timestamp",
+    "upperBound": "end_timestamp",
+}
+DEFAULT_CUSTOM_OPTIONS: Final[CustomOptions] = {"rangeMapping": DEFAULT_RANGE_MAPPING}
+DEFAULT_REQUIRED: Final[bool] = False
+
```
haiqi96, line 51: Don't think we need to empty lines here.

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    Prompt the user to provide the timestamp key for each dataset. If the user presses Enter, falls
+    back to `default_key`.
+
+    :param datasets:
+    :param default_key:
+    :return: Dictionary mapping dataset names to their timestamp keys.
+    """
+    print("\nPlease enter the timestamp key that corresponds to each of your archived datasets.")
+    print("Press <Enter> to accept the default key.\n")
```
haiqi96: Any chance we can use a single print for those two?

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,169 @@
...
+def _get_dataset_names(archives_dir: Path) -> Optional[List[str]]:
+    """
+    Return the names of first-level subdirectories in <clp-package-dir>/var/data/archives. Each
+    subdirectory name is treated as a dataset name.
+
+    :param archives_dir:
+    :return: List of dataset names. None if there are no datasets compressed.
+    """
+
+    return sorted([p.name for p in archives_dir.iterdir() if p.is_dir()])
```
haiqi96: what will happen if there's no dataset? it looks like the method will return an empty list. it will not return None

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,169 @@
...
+        timestamp_keys_by_dataset = _prompt_timestamp_keys(datasets)
+    except KeyboardInterrupt:
+        logger.error("Interrupted while collecting timestamp keys.")
+        return 1
+
+    try:
+        split_filters = _construct_split_filters(datasets, timestamp_keys_by_dataset)
+    except Exception:
+        logger.error("Missing timestamp key(s) for dataset(s).")
+        return 1
```
haiqi96: the right way to use Optional return should be:
```
    split_filters = _construct_split_filters(datasets, timestamp_keys_by_dataset)
    if split_filters is None:
        logger.error("Missing timestamp key(s) for dataset(s).")
        return 1
```

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,169 @@
...
+
+    with open(json_output_file, "w") as json_file:
```
haiqi96: it would be safer to ensure that json_output_file points to a valid file (i.e. not a file on an invalid path, or a directory)

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
haiqi96, line 1: High level: after making changes, please go through the docstrings and update if necessary.

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
quinntaylormitchell: I'll put `BOLD` and `RESET` in `#CONSTANTS` right under the imports, but the other constants (`DEFAULT_RANGE_MAPPING`, `DEFAULT_CUSTOM_OPTIONS`, etc.) need to go after the their type classes are defined. Is that OK?

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
quinntaylormitchell: will do!

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
quinntaylormitchell: sorry, what is this referring to?

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
quinntaylormitchell: I thought that we were supposed to use the `constant == variable` convention in `==` conditions?

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
quinntaylormitchell: agreed; the file should already exist as long as the user didn't delete it, but who knows what might happen.

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
quinntaylormitchell: fixed so that it now returns None

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
haiqi96: Only for None it is a special case.

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
haiqi96: Refer to define a variable for False. I think it's ok to leave it here

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,182 @@
...
+        return 1
+    if not archives_dir.is_dir():
+        logger.error("Archives path is not a directory: %s", archives_dir)
+        return 1
+    if not out_dir.exists():
+        logger.error("Output directory does not exist: %s", out_dir)
+        return 1
+    if not out_dir.is_dir():
+        logger.error("Output parent is not a directory: %s", out_dir)
+        return 1
```
haiqi96: We can have a util function for these at top of the file, so we don't duplicate the same logic twice.

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,182 @@
...
+    Constructs a split filter for each dataset using a per-dataset timestamp key.
+
+    :param datasets:
+    :param timestamp_keys: Mapping from dataset name -> timestamp key.
+    :return: A SplitFilterDict containing all the SplitFilterRule objects for the JSON file.
+    :return: None if there are any datasets that don't have an associated timestamp key.
+    """
+
+    missing = [d for d in datasets if d not in timestamp_keys]
+    if 0 < len(missing):
```
haiqi96: imo it's actuall very confusing to write if `0 < len(missing)`.
Two notes:
1. IIRC, We only need to put the constant on the write if it is a `==` and the rhs is a variable that you can assign value. in this case, you can write `if len(missing) > 0`.
2. I feel `if len(missing) != 0` is more straight forward.

haiqi96: Can we also add an error log message to print out which datasets are missing the timestamp key? would be good for debugging

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,182 @@
...
+    datasets: List[str],
+    timestamp_keys: Dict[str, str],
+) -> Optional[SplitFilterDict]:
+    """
+    Constructs a split filter for each dataset using a per-dataset timestamp key.
+
+    :param datasets:
+    :param timestamp_keys: Mapping from dataset name -> timestamp key.
+    :return: A SplitFilterDict containing all the SplitFilterRule objects for the JSON file.
+    :return: None if there are any datasets that don't have an associated timestamp key.
```
haiqi96: usually we put them in one line. you will need to run linter for my suggestion
```suggestion
    :return: A `SplitFilterDict` containing all the `SplitFilterRule` objects for the JSON file, or None 
    if there are any datasets that don't have an associated timestamp key.
```

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,182 @@
...
+        "\nPlease enter the timestamp key that corresponds to each of your archived datasets."
+        "\nPress <Enter> to accept the default key.\n"
+    )
+
+    data_time_pairs: Dict[str, str] = {}
+    for dataset in datasets:
+        user_input = input(
+            f">>> {ANSI_BOLD}{dataset}{ANSI_RESET} [default key: {ANSI_BOLD}{DEFAULT_TIMESTAMP_KEY}{ANSI_RESET}]: "
+        )
+        key = DEFAULT_TIMESTAMP_KEY if 0 == len(user_input) else user_input
```
haiqi96, line 159: iirc It's ok to write `len(user_input) == 0`, but you can keep what you have

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,182 @@
...
+    return datasets if datasets else None
+
+
+def _prompt_timestamp_keys(datasets: List[str]) -> Dict[str, str]:
+    """
+    Prompt the user to provide the timestamp key for each dataset. If the user presses Enter, falls
+    back to `DEFAULT_TIMESTAMP_KEY`.
+
+    :param datasets:
+    :return: Dictionary mapping dataset names to their timestamp keys.
```
haiqi96:
```suggestion
    :return: mapping of `dataset` -> `timestamp_keys`.
```

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,182 @@
...
+    :return: List of dataset names. None if there are no datasets compressed.
+    """
+
+    datasets = sorted([p.name for p in archives_dir.iterdir() if p.is_dir()])
+    return datasets if datasets else None
+
+
+def _prompt_timestamp_keys(datasets: List[str]) -> Dict[str, str]:
+    """
+    Prompt the user to provide the timestamp key for each dataset. If the user presses Enter, falls
```
haiqi96:
```suggestion
    Prompt the user to provide the timestamp key for each dataset. If the user doesn't provide one, falls
```

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+    Return the names of first-level subdirectories in <clp-package-dir>/var/data/archives. Each
+    subdirectory name is treated as a dataset name.
+
+    :param archives_dir:
+    :return: List of dataset names. None if there are no datasets compressed.
+    """
+
+    datasets = sorted([p.name for p in archives_dir.iterdir() if p.is_dir()])
+    return datasets if datasets else None
```
haiqi96: `if datasets` is an implicit check. Please be explicit about what you are checking.
I assume here it should be if `len(datasets) != 0`, but please double check

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
```diff
@@ -0,0 +1,182 @@
...
+    return 0
+
+
+def _get_dataset_names(archives_dir: Path) -> Optional[List[str]]:
+    """
+    Return the names of first-level subdirectories in <clp-package-dir>/var/data/archives. Each
+    subdirectory name is treated as a dataset name.
+
+    :param archives_dir:
+    :return: List of dataset names. None if there are no datasets compressed.
```
haiqi96:
```suggestion
    :return: List of dataset names. None if there are no directory under <clp-package-dir>/var/data/archives.
```

tools/deployment/presto-clp/scripts/generate-split-filter-file.py
quinntaylormitchell: definitely

PR #1176, opened 2025-08-08
feat: Add format information for float numbers to support lossless decompression.

anlowee:
# Description

TBD

# Checklist

  breaking change.

# Validation performed

### DIFF
--- a/components/core/src/clp_s/ColumnReader.cpp
+++ b/components/core/src/clp_s/ColumnReader.cpp
@@ -48,6 +48,7 @@ std::variant<int64_t, double, std::string, uint8_t> DeltaEncodedInt64ColumnReade
 
 void FloatColumnReader::load(BufferViewReader& reader, uint64_t num_messages) {
     m_values = reader.read_unaligned_span<double>(num_messages);
+    m_format = reader.read_unaligned_span<uint16_t>(num_messages);
 }
 
 void
@@ -74,8 +75,94 @@ void BooleanColumnReader::load(BufferViewReader& reader, uint64_t num_messages) 
 
 void
 FloatColumnReader::extract_string_value_into_buffer(uint64_t cur_message, std::string& buffer) {
-    buffer.append(std::to_string(m_values[cur_message]));
+    buffer.append(restore_format(cur_message));
 }
+
+std::string FloatColumnReader::restore_format(uint64_t cur_message) {
+    std::ostringstream oss;
+    const uint16_t significant_digits = (m_format[cur_message] >> float_format_encoding::cSignificantDigitsPos) & 0x1F;
+    oss << std::scientific << std::setprecision(significant_digits - 1);
+    if (m_format[cur_message] & (1 << float_format_encoding::cScientificExponentNotePos)) {
+        if (m_format[cur_message] & (1 << (float_format_encoding::cScientificExponentNotePos + 1))) {
+             oss << std::uppercase;
+        }
+        oss << m_values[cur_message];
+        auto formatted_double_str = oss.str();
+        const auto exp_pos = formatted_double_str.find_first_of("Ee");
+        const char maybe_sign = formatted_double_str[exp_pos + 1];
+        const uint16_t exp_digits = ((m_format[cur_message] >> float_format_encoding::cScientificExponentDigitsPos) & 0x03) + 1;
+        if (0b00 << float_format_encoding::cScientificExponentSignPos == (m_format[cur_message] & 0b11 << float_format_encoding::cScientificExponentSignPos)) {
+            if ('+' == maybe_sign || '-' == maybe_sign) {
+                formatted_double_str.erase(exp_pos + 1, 1);
+            }
+            if (exp_digits < (formatted_double_str.length() - exp_pos - 1)) {
+                formatted_double_str.erase(exp_pos + 1, (formatted_double_str.length() - exp_pos - 1) - exp_digits);
+            } else {
+                formatted_double_str.insert(exp_pos + 1, exp_digits - (formatted_double_str.length() - exp_pos - 1), '0');
+            }
+        } else {
+            if (exp_digits < (formatted_double_str.length() - exp_pos - 2)) {
+                formatted_double_str.erase(exp_pos + 2, (formatted_double_str.length() - exp_pos - 2) - exp_digits);
+            } else {
+                formatted_double_str.insert(exp_pos + 2, exp_digits - (formatted_double_str.length() - exp_pos - 2), '0');
+            }
+            if (0b01 << float_format_encoding::cScientificExponentSignPos == (m_format[cur_message] & 0b11 << float_format_encoding::cScientificExponentSignPos)) {
+                if (std::isdigit(maybe_sign)) {
+                    formatted_double_str.insert(exp_pos + 1, "+");
+                } else {
+                    formatted_double_str[exp_pos + 1] = '+';
+                }
+            } else if (0b10 << float_format_encoding::cScientificExponentSignPos == (m_format[cur_message] & 0b11 << float_format_encoding::cScientificExponentSignPos)) {
+                if (std::isdigit(maybe_sign)) {
+                    formatted_double_str.insert(exp_pos + 1, "-");
+                } else {
+                    formatted_double_str[exp_pos + 1] = '-';
+                }
+            }
+        }
+
+        return formatted_double_str;
+    }
+
+    // Convert the scientific notation to the standard decimal
+    oss << m_values[cur_message];
+    return scientific_to_decimal(oss.str());
+}
+
+std::string FloatColumnReader::scientific_to_decimal(
+    std::string_view scientific_notation) {
+    const auto sci_str = std::string(scientific_notation);
+    const size_t exp_pos = sci_str.find_first_of("e");
+    assert(std::string::npos != exp_pos);
+
+    // Split into mantissa and exponent parts
+    std::string mantissa_str = sci_str.substr(0, exp_pos);
+    const int exponent = std::stoi(sci_str.substr(exp_pos + 1));
+
+    // Remove the decimal point from the mantissa
+    const size_t dot_pos = mantissa_str.find('.');
+    std::string digits;
+    if (dot_pos != std::string::npos) {
+        digits = mantissa_str.substr(0, dot_pos) + mantissa_str.substr(dot_pos + 1);
+    } else {
+        digits = mantissa_str;
+    }
+
+    // Adjust position of decimal point based on exponent
+    const int decimal_pos = static_cast<int>(dot_pos) + exponent;
+
+    std::string result;
+    if (decimal_pos <= 0) {
+        result = "0." + std::string(-decimal_pos, '0') + digits;
+    } else if (decimal_pos < static_cast<int>(digits.size())) {
+        result = digits.substr(0, decimal_pos) + "." + digits.substr(decimal_pos);
+    } else {
+        result = digits + std::string(decimal_pos - digits.size(), '0');
+    }
+
+    return result;
+}
+
 
 std::variant<int64_t, double, std::string, uint8_t> BooleanColumnReader::extract_value(
         uint64_t cur_message
--- a/components/core/src/clp_s/ColumnReader.hpp
+++ b/components/core/src/clp_s/ColumnReader.hpp
@@ -143,8 +143,13 @@ public:
 
     void extract_string_value_into_buffer(uint64_t cur_message, std::string& buffer) override;
 
+    std::string restore_format(uint64_t cur_message);
+
 private:
     UnalignedMemSpan<double> m_values;
+    UnalignedMemSpan<uint16_t> m_format;
+
+    static std::string scientific_to_decimal(std::string_view scientific_notation);
 };
 
 class BooleanColumnReader : public BaseColumnReader {
--- a/components/core/src/clp_s/ColumnWriter.cpp
+++ b/components/core/src/clp_s/ColumnWriter.cpp
@@ -29,13 +29,79 @@ void DeltaEncodedInt64ColumnWriter::store(ZstdCompressor& compressor) {
 }
 
 size_t FloatColumnWriter::add_value(ParsedMessage::variable_t& value) {
-    m_values.push_back(std::get<double>(value));
-    return sizeof(double);
+    auto float_str = std::get<std::string>(value);
+
+    // Trim the raw string
+    float_str.erase(std::remove_if(float_str.begin(), float_str.end(), [](char c) {
+       return !std::isdigit(static_cast<unsigned char>(c)) && '.' != c && '+' != c && '-' != c && 'E' != c && 'e' != c;
+    }), float_str.end());
+
+    // If the raw JSON token is an illegal double number then here will throw
+    // exception
+    m_values.push_back(std::stod(float_str));
+
+    const auto dot_pos = float_str.find('.');
+    uint16_t format{0};
+
+    // Check whether it is the scientific; if so, if the exponent is E or e
+    size_t exp_pos = float_str.find_first_of("Ee");
+    if (std::string::npos != exp_pos) {
+        format |= 1 << float_format_encoding::cScientificExponentNotePos;
+        format |= ('E' == float_str[exp_pos]) << (float_format_encoding::cScientificExponentNotePos + 1);
+
+        // Check whether there is a sign for the exponent
+        if ('+' == float_str[exp_pos + 1]) {
+            format |= 1 << float_format_encoding::cScientificExponentSignPos;
+        } else if ('-' == float_str[exp_pos + 1]) {
+            format |= 1 << (float_format_encoding::cScientificExponentSignPos + 1);
+        }
+
+        // Set the number of exponent digits
+        int exp_digits = float_str.length() - exp_pos - 1;
+        if (false == std::isdigit(float_str[exp_pos + 1])) {
+            exp_digits--;
+        }
+        format |= (static_cast<uint16_t>(std::min(exp_digits - 1, 3)) & 0x03) << float_format_encoding::cScientificExponentDigitsPos;
+    } else {
+        exp_pos = float_str.length();
+    }
+
+    // According to the JSON grammar, there is no leading zeros for the integer
+    // part of a number, so we can check whether the first or the second (if the
+    // there is a sign) digit is 0 to know if the first non-zero number is in
+    // integer part.
+    size_t first_non_zero_frac_digit_pos = std::isdigit(float_str[0]) ? 0 : 1;
+    if (std::isdigit(float_str[0]) && '0' != float_str[0]) {
+        first_non_zero_frac_digit_pos = 0;
+    } else if (std::isdigit(float_str[1]) && '0' != float_str[1]) {
+        first_non_zero_frac_digit_pos = 1;
+    } else if (std::string::npos != dot_pos) {
+        for (size_t i = dot_pos + 1; i < float_str.length(); ++i) {
+            if (std::isdigit(float_str[i]) && '0' != float_str[i]) {
+                first_non_zero_frac_digit_pos = i;
+                break;
+            }
+        }
+    }
+
+    int significant_digits = exp_pos - first_non_zero_frac_digit_pos;
+    if (std::string::npos != dot_pos && first_non_zero_frac_digit_pos < dot_pos) {
+        significant_digits--;
+    }
+    const uint16_t compressed_significant_digits = static_cast<uint16_t>(std::min(significant_digits, 16)) & 0x1F;
+
+    format |= compressed_significant_digits << float_format_encoding::cSignificantDigitsPos;
+
+    m_format.push_back(format);
+    return sizeof(double) + sizeof(uint16_t);
 }
 
 void FloatColumnWriter::store(ZstdCompressor& compressor) {
-    size_t size = m_values.size() * sizeof(double);
-    compressor.write(reinterpret_cast<char const*>(m_values.data()), size);
+    assert(m_format.size() == m_values.size());
+    const auto values_size = m_values.size() * sizeof(double);
+    const auto format_size = m_format.size() * sizeof(uint16_t);
+    compressor.write(reinterpret_cast<char const*>(m_values.data()), values_size);
+    compressor.write(reinterpret_cast<char const*>(m_format.data()), format_size);
 }
 
 size_t BooleanColumnWriter::add_value(ParsedMessage::variable_t& value) {
--- a/components/core/src/clp_s/ColumnWriter.hpp
+++ b/components/core/src/clp_s/ColumnWriter.hpp
@@ -10,6 +10,7 @@
 #include "TimestampDictionaryWriter.hpp"
 #include "VariableEncoder.hpp"
 #include "ZstdCompressor.hpp"
+#include "FloatFormatEncoding.hpp"
 
 namespace clp_s {
 class BaseColumnWriter {
@@ -96,6 +97,7 @@ public:
 
 private:
     std::vector<double> m_values;
+    std::vector<uint16_t> m_format;
 };
 
 class BooleanColumnWriter : public BaseColumnWriter {
--- a/components/core/src/clp_s/FloatFormatEncoding.hpp
+++ b/components/core/src/clp_s/FloatFormatEncoding.hpp
@@ -1,0 +1,14 @@
+#ifndef CLP_S_FLOATFORMATENCODING_HPP
+#define CLP_S_FLOATFORMATENCODING_HPP
+
+#include <cstdint>
+namespace clp_s::float_format_encoding {
+
+constexpr uint16_t cScientificExponentNotePos = 14;
+constexpr uint16_t cScientificExponentSignPos = 12;
+constexpr uint16_t cScientificExponentDigitsPos = 10;
+constexpr uint16_t cSignificantDigitsPos = 5;
+
+} // namespace  clp_s::float_format_encoding
+
+#endif  // CLP_S_FLOATFORMATENCODING_HPP
--- a/components/core/src/clp_s/JsonParser.cpp
+++ b/components/core/src/clp_s/JsonParser.cpp
@@ -201,8 +201,8 @@ void JsonParser::parse_obj_in_array(ondemand::object line, int32_t parent_node_i
             case ondemand::json_type::number: {
                 ondemand::number number_value = cur_value.get_number();
                 if (true == number_value.is_double()) {
-                    double double_value = number_value.get_double();
-                    m_current_parsed_message.add_unordered_value(double_value);
+                    const auto double_value_str = cur_value.raw_json_token();
+                    m_current_parsed_message.add_unordered_value(double_value_str);
                     node_id = m_archive_writer
                                       ->add_node(node_id_stack.top(), NodeType::Float, cur_key);
                 } else {
@@ -279,8 +279,8 @@ void JsonParser::parse_array(ondemand::array array, int32_t parent_node_id) {
             case ondemand::json_type::number: {
                 ondemand::number number_value = cur_value.get_number();
                 if (true == number_value.is_double()) {
-                    double double_value = number_value.get_double();
-                    m_current_parsed_message.add_unordered_value(double_value);
+                    auto double_value_str = cur_value.raw_json_token();
+                    m_current_parsed_message.add_unordered_value(double_value_str);
                     node_id = m_archive_writer->add_node(parent_node_id, NodeType::Float, "");
                 } else {
                     int64_t i64_value;
@@ -407,8 +407,9 @@ void JsonParser::parse_line(ondemand::value line, int32_t parent_node_id, std::s
                                 ->ingest_timestamp_entry(m_timestamp_key, node_id, i64_value);
                     }
                 } else {
+                    const auto double_value_str = line.raw_json_token();
                     double double_value = line.get_double();
-                    m_current_parsed_message.add_value(node_id, double_value);
+                    m_current_parsed_message.add_value(node_id, double_value_str);
                     if (matches_timestamp) {
                         m_archive_writer
                                 ->ingest_timestamp_entry(m_timestamp_key, node_id, double_value);
--- a/components/core/src/clp_s/SchemaReader.cpp
+++ b/components/core/src/clp_s/SchemaReader.cpp
@@ -116,14 +116,14 @@ void SchemaReader::generate_json_string() {
                 auto name = m_global_schema_tree->get_node(column->get_id()).get_key_name();
                 m_json_serializer.append_key(name);
                 m_json_serializer.append_value(
-                        std::to_string(std::get<double>(column->extract_value(m_cur_message)))
+                        static_cast<FloatColumnReader*>(column)->restore_format(m_cur_message)
                 );
                 break;
             }
             case JsonSerializer::Op::AddFloatValue: {
                 column = m_reordered_columns[column_id_index++];
                 m_json_serializer.append_value(
-                        std::to_string(std::get<double>(column->extract_value(m_cur_message)))
+                        static_cast<FloatColumnReader*>(column)->restore_format(m_cur_message)
                 );
                 break;
             }
--- a/components/core/tests/test-clp_s-end_to_end.cpp
+++ b/components/core/tests/test-clp_s-end_to_end.cpp
@@ -20,21 +20,22 @@ constexpr std::string_view cTestEndToEndOutputDirectory{"test-end-to-end-out"};
 constexpr std::string_view cTestEndToEndOutputSortedJson{"test-end-to-end_sorted.jsonl"};
 constexpr std::string_view cTestEndToEndInputFileDirectory{"test_log_files"};
 constexpr std::string_view cTestEndToEndInputFile{"test_no_floats_sorted.jsonl"};
+constexpr std::string_view cTestEndToEndValidFloatFormatInputFile{"test_valid_float_format.jsonl"};
 
 namespace {
-auto get_test_input_path_relative_to_tests_dir() -> std::filesystem::path;
-auto get_test_input_local_path() -> std::string;
+auto get_test_input_path_relative_to_tests_dir(const std::string_view test_input_path) -> std::filesystem::path;
+auto get_test_input_local_path(const std::string_view test_input_path) -> std::string;
 auto extract() -> std::filesystem::path;
 void compare(std::filesystem::path const& extracted_json_path);
 
-auto get_test_input_path_relative_to_tests_dir() -> std::filesystem::path {
-    return std::filesystem::path{cTestEndToEndInputFileDirectory} / cTestEndToEndInputFile;
+auto get_test_input_path_relative_to_tests_dir(const std::string_view test_input_path) -> std::filesystem::path {
+    return std::filesystem::path{cTestEndToEndInputFileDirectory} / test_input_path;
 }
 
-auto get_test_input_local_path() -> std::string {
+auto get_test_input_local_path(const std::string_view test_input_path) -> std::string {
     std::filesystem::path const current_file_path{__FILE__};
     auto const tests_dir{current_file_path.parent_path()};
-    return (tests_dir / get_test_input_path_relative_to_tests_dir()).string();
+    return (tests_dir / get_test_input_path_relative_to_tests_dir(test_input_path)).string();
 }
 
 auto extract() -> std::filesystem::path {
@@ -59,7 +60,6 @@ auto extract() -> std::filesystem::path {
     std::filesystem::path extracted_json_path{cTestEndToEndOutputDirectory};
     extracted_json_path /= "original";
     REQUIRE(std::filesystem::exists(extracted_json_path));
-
     return extracted_json_path;
 }
 
@@ -83,7 +83,7 @@ void compare(std::filesystem::path const& extracted_json_path) {
     command = fmt::format(
             "diff --unified {} {}  > /dev/null",
             cTestEndToEndOutputSortedJson,
-            get_test_input_local_path()
+            get_test_input_local_path(cTestEndToEndInputFile)
     );
     result = std::system(command.c_str());
     REQUIRE((true == WIFEXITED(result)));
@@ -105,7 +105,7 @@ TEST_CASE("clp-s-compress-extract-no-floats", "[clp-s][end-to-end]") {
 
     REQUIRE_NOTHROW(
             std::ignore = compress_archive(
-                    get_test_input_local_path(),
+                    get_test_input_local_path(cTestEndToEndInputFile),
                     std::string{cTestEndToEndArchiveDirectory},
                     single_file_archive,
                     structurize_arrays,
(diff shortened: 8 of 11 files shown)


gibber9809 (commented):
One important comment -- this new float encoding should be a new `NodeType` w/ a new column reader and writer. We should avoid changing how the encoding for an existing `NodeType` works.

kirkrodrigues:
One thing we'll need to figure out is whether we should add support to KV-IR streams, since we use them to view logs in the log viewer. E.g., if the user's original JSON log contained a float that was printed with a specific format (e.g., "-0.0e-0"), when converted to IR, that format wouldn't be retained). This wouldn't be a breaking change, so it seems doable.

@LinZhihao-723 Any thoughts?

LinZhihao-723:
> One thing we'll need to figure out is whether we should add support to KV-IR streams, since we use them to view logs in the log viewer. E.g., if the user's original JSON log contained a float that was printed with a specific format (e.g., "-0.0e-0"), when converted to IR, that format wouldn't be retained). This wouldn't be a breaking change, so it seems doable.
> 
> @LinZhihao-723 Any thoughts?

From a format perspective, I think that should be feasible. We don't need to introduce any more types in the scheme tree, but add a CLP-encoded float type (so overall it works like the String type, where the encoded value can be a variable string or an encoded text AST). But we probably need to handle the case where the float can't be encoded into an 8-byte variable. In that case, we probably need to serialize it as a string.

From the serializer's perspective, it probably requires more work: currently the serializer only takes msgpack. To trigger the CLP-float encoding, we may need to serialize the float as a string (which will have performance overhead), or we define a customized type (for which I don't know the details of how to do this. Need time to figure out).

components/core/src/clp_s/ColumnWriter.cpp
```diff
@@ -46,6 +46,94 @@ void FloatColumnWriter::store(ZstdCompressor& compressor) {
...
+    }
+    assert(significant_digits > 0);
+    uint16_t const compressed_significant_digits
+            = static_cast<uint16_t>(std::min(significant_digits - 1, 15)) & 0x0F;
+
+    format |= compressed_significant_digits << float_format_encoding::cSignificantDigitsPos;
+
+    m_format.push_back(format);
+    return sizeof(double) + sizeof(uint16_t);
+}
```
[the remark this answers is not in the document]
anlowee: I intentionally wrote it that way to make it throw exception to terminate the decompression.

components/core/src/clp_s/SchemaTree.hpp
```diff
@@ -32,6 +32,7 @@ namespace clp_s {
 enum class NodeType : uint8_t {
     Integer,
     Float,
+    FormattedFloat,
```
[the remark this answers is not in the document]
anlowee: I intentionally to put it below Float to group them

components/core/src/clp_s/FloatFormatEncoding.hpp
```diff
@@ -0,0 +1,13 @@
...
+#define CLP_S_FLOATFORMATENCODING_HPP
+
+#include <cstdint>
+
+namespace clp_s::float_format_encoding {
+constexpr uint16_t cScientificExponentNotePos = 14;
+constexpr uint16_t cScientificExponentSignPos = 12;
+constexpr uint16_t cScientificExponentDigitsPos = 10;
+constexpr uint16_t cSignificantDigitsPos = 6;
+}  // namespace  clp_s::float_format_encoding
```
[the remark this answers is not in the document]
anlowee: I have a very detailed doc for the encoding in the docs, do I still need to write comment here?

components/core/src/clp_s/ColumnWriter.cpp
```diff
@@ -46,6 +46,98 @@ void FloatColumnWriter::store(ZstdCompressor& compressor) {
...
+            std::remove_if(
+                    float_str.begin(),
+                    float_str.end(),
+                    [](char c) {
+                        return !std::isdigit(static_cast<unsigned char>(c)) && '.' != c && '+' != c
+                               && '-' != c && 'E' != c && 'e' != c;
+                    }
+            ),
+            float_str.end()
+    );
```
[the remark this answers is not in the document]
anlowee: I already have assert(exp_pos + 1 < float_str.length()); why still need to add expo_pos + x checks in the subsequent if statemetns?

components/core/src/clp_s/SchemaTree.hpp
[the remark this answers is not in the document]
gibber9809: +1 to this rabbit comment @anlowee -- new nodetypes must always be added to the end of this enum right before Unknown.

Typical flow is add the type to the end of this enum, write encoding/decoding support, and bump the archive format patch version.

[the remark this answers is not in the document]
anlowee: Gotcha

components/core/src/clp_s/ColumnWriter.hpp
```diff
@@ -98,6 +99,24 @@ class FloatColumnWriter : public BaseColumnWriter {
...
+    ~FormattedFloatColumnWriter() override = default;
+
+    // Methods inherited from BaseColumnWriter
+    size_t add_value(ParsedMessage::variable_t& value) override;
+
+    void store(ZstdCompressor& compressor) override;
+
+private:
+    std::vector<double> m_values;
+    std::vector<uint16_t> m_format;
```
gibber9809:
```suggestion
    std::vector<uint16_t> m_formats;
```
I think its more consistent to also use the plural of format.

components/core/src/clp_s/ColumnWriter.cpp
```diff
@@ -46,6 +49,114 @@ void FloatColumnWriter::store(ZstdCompressor& compressor) {
...
```
gibber9809: Is this to deal with the fact that `raw_json_token` can leave some trailing whitespace [(per the docs)?](https://simdjson.org/api/2.0.0/classsimdjson_1_1_s_i_m_d_j_s_o_n___i_m_p_l_e_m_e_n_t_a_t_i_o_n_1_1ondemand_1_1value.html#ab3f9b17e853c3b7ed8cd7ca9bb9b05c6). If so I think it might make sense to either
1) Just reverse iterate to remove trailing whitespace here
2) Remove this string trimming stuff here and instead remove trailing whitespace when we first parse the values

I'd probably lean towards 2 so that the code in this function can be a bit simpler.

components/core/src/clp_s/ColumnWriter.cpp
```diff
@@ -46,6 +49,114 @@ void FloatColumnWriter::store(ZstdCompressor& compressor) {
...
+
+    // If the raw JSON token is an illegal double, this will throw an exception
+    m_values.push_back(std::stod(float_str));
```
gibber9809: This is fine as defensive programming, but just to confirm this shouldn't ever throw given the way we're currently parsing number values right? I would think that if simdjson claims something is a double it is guaranteed to be valid, but maybe that's not right.

components/core/src/clp_s/ColumnWriter.cpp
```diff
@@ -46,6 +49,114 @@ void FloatColumnWriter::store(ZstdCompressor& compressor) {
...
+
+    auto const dot_pos = float_str.find('.');
+    uint16_t format{0};
+
+    // Check whether it is the scientific; if so, if the exponent is E or e
```
gibber9809:
```suggestion
    // Check whether it is scientific; if so, whether the exponent is E or e
```

components/core/src/clp_s/ColumnWriter.cpp
```diff
@@ -46,6 +49,114 @@ void FloatColumnWriter::store(ZstdCompressor& compressor) {
     compressor.write(reinterpret_cast<char const*>(m_values.data()), size);
 }
 
+size_t FormattedFloatColumnWriter::add_value(ParsedMessage::variable_t& value) {
```
gibber9809, line 52: I think we might be able to simplify this function a lot by changing the calling convention s.t. we've already checked that `value` is a valid double. We have to confirm, but with the way we're using simdjson if  we reach here I think that that might already be the case.

I think this function would be much easier to read without the preprocessing at the start + asserts throughout.

components/core/src/clp_s/ColumnWriter.cpp
```diff
@@ -46,6 +49,114 @@ void FloatColumnWriter::store(ZstdCompressor& compressor) {
...
+                          == std::isdigit(
+                                  static_cast<unsigned char>(
+                                          float_str[first_non_zero_frac_digit_pos + 1]
+                                  )
+                          ));
+
+        // For "0.xxx", find first non-zero in fractional part
+        if (std::string::npos != dot_pos) {
+            for (size_t i = dot_pos + 1; i < exp_pos; ++i) {
+                if (std::isdigit(static_cast<unsigned char>(float_str[i])) && '0' != float_str[i]) {
```
gibber9809:
```suggestion
                if ('0' != float_str[i]) {
```
I'm pretty sure the `std::is_digit` check isn't needed here since we're bounding our iteration by `exp_pos`, but correct me if I'm wrong.

components/core/src/clp_s/FloatFormatEncoding.hpp
```diff
@@ -0,0 +1,25 @@
...
+// See docs/src/dev-guide/design-retain-float-format.md for the full layout.
+
+// Scientific notation marker (2 bits)
+constexpr uint16_t cScientificExponentNotePos = 14;
+// Exponent sign presence (2 bits)
+constexpr uint16_t cScientificExponentSignPos = 12;
+// Number of exponent digits (2 bits)
+constexpr uint16_t cScientificExponentDigitsPos = 10;
+// Number of significant digits (4 bits)
+constexpr uint16_t cSignificantDigitsPos = 6;
```
gibber9809: I think these names are a bit confusing. How about:
`cExponentNotationPos`
`cExponentSignPos`
`cNumExponentDigitsPos`
`cNumSignificantDigitsPos`

components/core/src/clp_s/ColumnWriter.cpp
gibber9809: Also realizing on a second read-through that this currently doesn't handle `NaN` correctly.

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -388,6 +406,8 @@ void JsonParser::parse_line(ondemand::value line, int32_t parent_node_id, std::s
                     // FIXME: should have separate integer and unsigned
                     // integer types to handle values greater than max int64
                     type = NodeType::Integer;
+                } else if (m_retain_float_format) {
+                    type = NodeType::FormattedFloat;
                 } else {
                     type = NodeType::Float;
```
gibber9809: May instead of an extra branch just do
```
} else {
    type = m_retain_float_format ? NodeType::FormattedFloat : NodeType::Float;
}
```
To me it reads weirdly to handle float in two separate branches.

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -408,7 +428,12 @@ void JsonParser::parse_line(ondemand::value line, int32_t parent_node_id, std::s
                     }
                 } else {
                     double double_value = line.get_double();
```
gibber9809: Sorry, not really related to your change, but I realized that this should be `number_value.get_double()`

components/core/src/clp_s/SchemaReader.cpp
```diff
@@ -127,6 +127,26 @@ void SchemaReader::generate_json_string() {
...
+                        )
+                );
+                break;
+            }
+            case JsonSerializer::Op::AddFormattedFloatValue: {
+                column = m_reordered_columns[column_id_index++];
+                m_json_serializer.append_value(
+                        static_cast<FormattedFloatColumnReader*>(column)->restore_format(
+                                m_cur_message
+                        )
```
gibber9809: Do you think its worth rewriting these as `m_json_serializer.append_value_from_column(column, m_cur_message)` instead of `m_json_serializer.append_value(...)`? I think it would make the code more re-usable if we add another kind of formatted float column in the future + leaves the door open for some optimizations since `append_value_from_column` calls `extract_string_value_into_buffer` internally.

components/core/src/clp_s/ColumnReader.hpp
```diff
@@ -147,6 +147,74 @@ class FloatColumnReader : public BaseColumnReader {
...
+     */
+    std::string restore_format(uint64_t cur_message) const;
+
+private:
+    static constexpr uint16_t cEmptyExponentSign = 0b00;
+    static constexpr uint16_t cPlusExponentSign = 0b01;
+    static constexpr uint16_t cMinusExponentSign = 0b10;
+
+    UnalignedMemSpan<double> m_values;
+    UnalignedMemSpan<uint16_t> m_format;
```
gibber9809:
```suggestion
    UnalignedMemSpan<uint16_t> m_formats;
```

components/core/src/clp_s/ColumnReader.hpp
```diff
@@ -147,6 +147,74 @@ class FloatColumnReader : public BaseColumnReader {
...
+     *
+     * @param cur_message
+     * @return The double value string formatted by the format information.
+     */
+    std::string restore_format(uint64_t cur_message) const;
+
+private:
+    static constexpr uint16_t cEmptyExponentSign = 0b00;
+    static constexpr uint16_t cPlusExponentSign = 0b01;
+    static constexpr uint16_t cMinusExponentSign = 0b10;
```
gibber9809: Since we have a header dedicated to float formatting stuff I think it would make sense to just move these sorts of constants to that header.

components/core/src/clp_s/ColumnReader.hpp
```diff
@@ -147,6 +147,74 @@ class FloatColumnReader : public BaseColumnReader {
...
+
+    UnalignedMemSpan<double> m_values;
+    UnalignedMemSpan<uint16_t> m_format;
+
+    bool has_exponent_sign(uint64_t cur_message, uint16_t sign) const;
+    bool has_scientific_notation(uint64_t cur_message) const;
+    bool is_uppercase_exponent(uint64_t cur_message) const;
+
+    uint16_t get_exponent_digits(uint64_t cur_message) const;
+    uint16_t get_significant_digits(uint64_t cur_message) const;
```
gibber9809: Same for these -- these sorts of utilities should probably be in the header dedicated to the float formatting stuff.

Also instead of `has_exponent_sign` we should probably just have a `get_exponent_sign` function and do comparisons directly -- the current approach is very hard to read (granted the current approach would be okay to read if the the function were named something like `has_matching_exponent_sign`).

components/core/tests/test-clp_s-search.cpp
```diff
@@ -116,14 +119,17 @@ void validate_results(
...
-    std::set<int64_t> expected_results_set{expected_results.begin(), expected_results.end()};
-    REQUIRE(results_set == expected_results_set);
-    REQUIRE(results.size() == expected_results.size());
+    if (expected_results.empty()) {
+        REQUIRE(results.empty());
+    } else {
+        std::set<int64_t> expected_results_set{expected_results.begin(), expected_results.end()};
+        REQUIRE(results_set == expected_results_set);
+        REQUIRE(results.size() == expected_results.size());
+    }
```
gibber9809: Why is this diff needed? It seems like the previous approach should also work for queries with no results.

components/core/tests/test-clp_s-end_to_end.cpp
```diff
@@ -117,3 +169,61 @@ TEST_CASE("clp-s-compress-extract-no-floats", "[clp-s][end-to-end]") {
 
     compare(extracted_json_path);
 }
+
+TEST_CASE("clp-s-compress-extract-valid-formatted-floats", "[clp-s][end-to-end]") {
```
gibber9809, line 216: It isn't too clear at a glance what the difference is between what this and the following `rounded-valid-formatted-floats` is testing. Maybe add comments to clarify?

components/core/src/clp_s/ColumnWriter.cpp
anlowee: As discussed with @kirkrodrigues , we don't support `NaN` or `Infinite` because simdjson doesn't support them.

components/core/src/clp_s/ColumnWriter.cpp
anlowee: true

components/core/src/clp_s/ColumnWriter.cpp
gibber9809: Fair, yeah actually now that I'm double checking the JSON spec NaN and infinity are [explicitly not allowed](https://datatracker.ietf.org/doc/html/rfc8259#section-6).

docs/src/dev-guide/design-retain-float-format.md
```diff
@@ -0,0 +1,95 @@
...
+- The double value in IEEE 754 64-bit format.
+- A 2-byte *format* field encoding the necessary output formatting information so that, upon
+  decompression, the value can be reproduced to match the original text as closely as possible
+  (see limitations below). Note that the remaining bits of the 2‑byte field are currently reserved.
+  Encoders must write them as 0, and decoders must ignore them (treat as “don’t care”) for forward
+  compatibility.
+
+```text
++-------------------------------------+------------------------+--------------------------+------------------------------------------------------+
+| Scientific Notation Marker (2 bits) | Exponent Sign (2 bits) | Exponent Digits (2 bits) | Digits from First Non-Zero to End of Number (4 bits) |
```
gibber9809: Sorry, thought about this a bit more -- I think this approach still runs into issues for numbers like `0.<large num zeroes><actual digits>e<whatever>` since we only record the number of digits in the exponent the actual exponent value and number of zeroes is ambiguous. I think we need to think a bit more about these edge cases before we finalize the design.

gibber9809: Can't think of a great way of dealing with this ambiguity without just using 9 bits to store the actual decimal exponent value.

gibber9809: Though then again I'm not really sure if there can be a large number of preceding zeroes in the scientific notation case. In general I think we'll just have to closely examine the spec/common implementations and decide what to support.

components/core/src/clp_s/ColumnWriter.cpp
anlowee: true, this is to apply a coderabbit comment before. We can remove it

components/core/src/clp_s/ColumnWriter.cpp
anlowee: That preprocessing is actuall necessary because float_str here is the raw json token and it seems contains the spaces unit the next field or the end of the json. For example `{ "normalStandard": 123456789.1234567 }` the `float_str` is `123456789.1234567 `.

gibber9809: Sorry I'd written some extra feedback about that specifically, but I don't see the comment so I must have deleted it. Yes `raw_json_token()` will yield extra whitespace up to the start of the next token. To deal with that I think it would be cleaner to just trim off that extra whitespace in the parsing code so that we don't need to do extra preprocessing inside of this function/in general I think we should be able to have a convention that the values passed to this function are valid floats.

components/core/src/clp_s/ColumnReader.hpp
anlowee: Marked. Will do this when apply the changes in the format design

docs/src/dev-guide/design-retain-float-format.md
anlowee: For `0.<large num zeroes><actual digits>e<whatever>` case it can handle (see the test case `{ "standardFloatWithLeadingZerosAndTrailingZeros": 0.00000000000000000000000000001234567891234500 }` in the `test_valid_formatted_float.jsonl`). When the number is not zero, we record up to 16 digits (will be 17 in the very soon) starting from the first non-zero digit, so preceding zeros won't be a problem.

components/core/tests/test-clp_s-search.cpp
anlowee: true, removed.

components/core/src/clp_s/ArchiveReader.cpp
```diff
@@ -247,6 +253,12 @@ void ArchiveReader::append_unordered_reader_columns(
             case NodeType::Float:
                 column_reader = new FloatColumnReader(column_id);
                 break;
+            case NodeType::FormattedFloat:
+                column_reader = new FormattedFloatColumnReader(column_id);
+                break;
+            case NodeType::DictionaryFloat:
+                column_reader = new DictionaryFloatColumnReader(column_id);
```
anlowee: Why no `m_var_dict` needed here?

components/core/src/clp_s/ColumnReader.hpp
```diff
@@ -147,6 +150,66 @@ class FloatColumnReader : public BaseColumnReader {
...
+     * @param cur_message
+     * @param buffer
+     *
+     */
+    void extract_string_value_into_buffer(uint64_t cur_message, std::string& buffer) override;
+
+private:
+    UnalignedMemSpan<double> m_values;
+    UnalignedMemSpan<float_format_t> m_formats;
+};
```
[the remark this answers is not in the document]
anlowee: Maybe consider replace that `uint16_t` with `float_format_t`

components/core/src/clp_s/ColumnWriter.cpp
```diff
@@ -46,6 +49,33 @@ void FloatColumnWriter::store(ZstdCompressor& compressor) {
     compressor.write(reinterpret_cast<char const*>(m_values.data()), size);
 }
 
+size_t FormattedFloatColumnWriter::add_value(ParsedMessage::variable_t& value) {
+    auto const [float_value, format]{std::get<std::pair<double, float_format_t>>(value)};
```
anlowee: Can we use `auto const&` here? It might cause a building issue for Velox with clp:
```
/home/yscope/actions-runner/_work/velox/velox/velox/_build/release/_deps/clp-src/components/core/src/clp/string_utils/../ffi/ir_stream/search/QueryHandlerImpl.hpp:468:21: note: use reference type to prevent copying
  468 |     for (auto const [node_id, token_it] : new_partial_resolutions) {
      |                     ^~~~~~~~~~~~~~~~~~~
      |                     &
```
https://github.com/y-scope/velox/actions/runs/17495530034/job/49695302630?pr=29

components/core/src/clp_s/ColumnWriter.cpp
```diff
@@ -46,6 +49,33 @@ void FloatColumnWriter::store(ZstdCompressor& compressor) {
...
+    compressor.write(reinterpret_cast<char const*>(m_values.data()), values_size);
+    compressor.write(reinterpret_cast<char const*>(m_formats.data()), format_size);
+}
+
+size_t DictionaryFloatColumnWriter::add_value(ParsedMessage::variable_t& value) {
+    clp::variable_dictionary_id_t id{};
+    m_var_dict->add_entry(std::get<std::string>(value), id);
+    m_var_dict_ids.push_back(id);
+    return sizeof(variable_dictionary_id_t);
+}
```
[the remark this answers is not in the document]
anlowee: Maybe consider this, since you are using `clp:: variable_dictionary_id_t` to declare the variable as well as `sizeof(clp::variable_dictionary_id_t)` below.

components/core/src/clp_s/ColumnReader.hpp
```diff
@@ -147,6 +150,66 @@ class FloatColumnReader : public BaseColumnReader {
...
+    std::variant<int64_t, double, std::string, uint8_t> extract_value(
+            uint64_t cur_message
+    ) override;
+
+    /**
+     * Appends the floating point value to the buffer in its original format by decoding the stored
+     * format information.
+     * @param cur_message
+     * @param buffer
+     *
```
anlowee:
```suggestion
     * Appends the floating point value to the buffer in its original format by decoding the stored
     * format information.
     *
     * @param cur_message
     * @param buffer
```

components/core/src/clp_s/ColumnReader.cpp
```diff
@@ -50,6 +56,11 @@ void FloatColumnReader::load(BufferViewReader& reader, uint64_t num_messages) {
     m_values = reader.read_unaligned_span<double>(num_messages);
 }
 
+void FormattedFloatColumnReader::load(BufferViewReader& reader, uint64_t num_messages) {
+    m_values = reader.read_unaligned_span<double>(num_messages);
+    m_formats = reader.read_unaligned_span<uint16_t>(num_messages);
```
anlowee: Maybe `float_format_t`

components/core/src/clp_s/FloatFormatEncoding.cpp
```diff
@@ -0,0 +1,322 @@
...
+                    first_non_zero_frac_digit_pos = i;
+                    break;
+                }
+            }
+        }
+    }
+
+    if (first_non_zero_frac_digit_pos >= exp_pos) {
+        return std::errc::protocol_not_supported;
+    }
```
anlowee: Is this check necessary? It seems the only chance that `first_non_zero_frac_digit_pos` gets updated is at L227, and `i` will be ensured to be less than `exp_pos`.

components/core/src/clp_s/FloatFormatEncoding.hpp
```diff
@@ -0,0 +1,57 @@
...
+// Bit positions for the 16-bit format descriptor.
+// See docs/src/dev-guide/design-retain-float-format.md for the full layout.
+
+// Scientific notation marker (2 bits)
+constexpr float_format_t cScientificNotationFlagPos = 14U;
+constexpr float_format_t cScientificNotationFlagMask = 0b11U << cScientificNotationFlagPos;
+constexpr float_format_t cScientificNotationEnabledBit = 0b01U << cScientificNotationFlagPos;
+constexpr float_format_t cScientificNotationLowerCaseEFlag = 0b01U << cScientificNotationFlagPos;
+constexpr float_format_t cScientificNotationUpperCaseEFlag = 0b11U << cScientificNotationFlagPos;
+// Exponent sign presence (2 bits)
```
anlowee, line 25:
```suggestion

// Exponent sign presence (2 bits)
```

components/core/src/clp_s/FloatFormatEncoding.hpp
```diff
@@ -0,0 +1,57 @@
...
+constexpr float_format_t cExponentSignFlagPos = 12U;
+constexpr float_format_t cExponentSignFlagMask = 0b11U << cExponentSignFlagPos;
+constexpr float_format_t cEmptyExponentSignFlag = 0b00U << cExponentSignFlagPos;
+constexpr float_format_t cPlusExponentSignFlag = 0b01U << cExponentSignFlagPos;
+constexpr float_format_t cMinusExponentSignFlag = 0b10U << cExponentSignFlagPos;
+// Number of exponent digits (2 bits)
```
anlowee, line 32:
```suggestion

// Number of exponent digits (2 bits)
```

components/core/src/clp_s/FloatFormatEncoding.hpp
```diff
@@ -0,0 +1,57 @@
...
+constexpr float_format_t cNumExponentDigitsPos = 10U;
+constexpr float_format_t cNumExponentDigitsMask = 0b11U << cNumExponentDigitsPos;
+// Number of significant digits (4 bits)
```
anlowee:
```suggestion

// Number of significant digits (5 bits)
```

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -41,6 +45,33 @@ using clp::ffi::KeyValuePairLogEvent;
 using clp::UtcOffset;
 
 namespace clp_s {
+namespace {
+/**
+ * Trims trailing whitespace off of a `string_view`. The returned `string_view` points to a subset
+ * of the input `string_view`.
+ * @param str
```
anlowee, line 52:
```suggestion
 *
 * @param str
```

components/core/tests/test-FloatFormatEncoding.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+        for (size_t i{0ULL}; i < cNumTests; ++i) {
+            auto const value{distribution(generator)};
+            auto const neg_value{-value};
+            auto const significant_digits{significant_digits_distribution(generator)};
+            std::vector<std::string> formats{
+                    fmt::format("{{:.{}e}}", significant_digits),
+                    fmt::format("{{:.{}E}}", significant_digits),
+                    fmt::format("{{:.{}f}}", significant_digits)
+            };
```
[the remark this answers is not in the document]
anlowee: Does this make sense?

components/core/tests/test-FloatFormatEncoding.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    SECTION("Test round trips for zeroes.") {
+        double const zero{0.0};
+        double const neg_zero{-zero};
+        for (size_t significant_digits{1ULL}; significant_digits < 17; ++significant_digits) {
+            std::vector<std::string> formats{
+                    fmt::format("{{:.{}e}}", significant_digits),
+                    fmt::format("{{:.{}E}}", significant_digits),
+                    fmt::format("{{:.{}f}}", significant_digits)
+            };
```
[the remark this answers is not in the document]
anlowee: Does this make sense?

components/core/tests/test-clp_s-end_to_end.cpp
```diff
@@ -85,7 +101,41 @@ void compare(std::filesystem::path const& extracted_json_path) {
...
+    REQUIRE((0 == result));
+    REQUIRE((false == std::filesystem::is_empty(cTestEndToEndOutputSortedJson)));
+
+    result = std::system("command -v diff >/dev/null 2>&1");
+    REQUIRE((0 == result));
+    command = fmt::format(
+            "diff --unified {} {}  > /dev/null",
+            cTestEndToEndExpectedOutputSortedFile,
+            cTestEndToEndOutputSortedJson
     );
```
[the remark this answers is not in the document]
anlowee: Seems correct.

components/core/tests/test_log_files/test_rounded_formatted_float.jsonl
```diff
@@ -0,0 +1,42 @@
...
+{ "negativeScientificNegativeExponentExceedingDigitLimitWithLeadingZeros": -1.234567891234567E-00008 }
+{ "scientificSignificandsExceedingPrecisionLimitWithTrailingZerosWithNegativeExponentExceedingDigitLimitWithLeadingZeros": 1.23456789123456700E-00008 }
+{ "negativeScientificSignificandsExceedingPrecisionLimitWithTrailingZerosWithPositiveExponentExceedingDigitLimitWithLeadingZerosWithPlus": -1.23456789123456700E+00008 }
+
+{ "scientificWithGreaterThanTenMantissa": 123456789.1234567E3 }
+{ "negativeScientificWithGreaterThanTenMantissa": -123456789.1234567E3 }
+{ "scientificWithGreaterThanTenMantissaExceedingPrecision": 123456789.123456789E3 }
+{ "negativeScientificWithGreaterThanTenMantissaExceedingPrecision": -123456789.123456789E3 }
+{ "scientificWithGreaterThanTenMantissa2": 12.34567891234567E01 }
+{ "-negativeScientificWithGreaterThanTenMantissa2": -12.34567891234567E01 }
```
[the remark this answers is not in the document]
anlowee: make sense

components/core/tests/test-clp_s-search.cpp
```diff
@@ -232,8 +234,47 @@
...
+            {R"aa(NOT formattedFloatValue: 0)aa", {0, 1, 2, 6, 7, 8, 9, 10, 11, 12}},
+            {R"aa(formattedFloatValue: 0)aa", {3, 4, 5}},
+            {R"aa(formattedFloatValue: 1e-16)aa", {6, 7}},
+            {R"aa(formattedFloatValue > 0.00)aa", {6, 7, 8, 9, 10, 11, 12}},
+            {R"aa(formattedFloatValue > 5000.000000000001)aa", {12}},
+            {R"aa(formattedFloatValue < 0.00 AND formattedFloatValue >= -0.01)aa", {1, 2}},
+            {R"aa(idx: 0 AND NOT formattedFloatValue: -1000.0)aa", {}},
+            {R"aa(msg: "xxx" AND formattedFloatValue: 3000.0)aa", {}},
+            {R"aa(msg: "xxx" OR formattedFloatValue: 3000.0)aa", {0, 9}}
+    };
```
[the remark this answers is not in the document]
anlowee: maybe these `{R"aa(formattedFloatValue: "0.0000000000000000E+000")aa", {3, 4, 5}},`
`{R"aa(formattedFloatValue: "-0.0000000000000000E+000")aa", {3, 4, 5}},`

components/core/tests/test_log_files/test_valid_formatted_float.jsonl
```diff
@@ -0,0 +1,40 @@
...
+{ "e0CornerCase3": 0.000E0 }
+{ "e0CornerCase4": -0.000E0 }
+{ "e0CornerCase5": 0E0 }
+{ "e0CornerCase6": 0E-0000 }
+{ "e0CornerCase7": -0E+0000 }
+{ "e0CornerCase8": 0.000E-0000 }
+{ "e0CornerCase9": -0.000E+0000 }
+{ "e0cornerCase10": 1.000E+0 }
+{ "e0cornerCase11": 1.000E-0 }
+{ "e0cornerCase12": 1.000E-00 }
```
anlowee, line 38: Maybe add to test 17 digits zero cases:
```
{ "mantissa0CornerCase1": 0.0000000000000000E0010 }
{ "mantissa0CornerCase2": -0.0000000000000000E0010 }
```

anlowee:
Finished the first round of review.

components/core/tests/test-clp_s-end_to_end.cpp
```diff
@@ -120,3 +210,89 @@ TEST_CASE("clp-s-compress-extract-no-floats", "[clp-s][end-to-end]") {
...
+            extracted_json_path
+    );
+}
+
+/**
+ * Test that floats that are over 17 digits of precision, or are 17 digits of precision or less
+ * but do not exactly represent the nearest decimal value for any ieee-754 binary64 at the given
+ * precision, can be retained accurately.
+ */
+TEST_CASE("clp-s-compress-extract-rounded-valid-formatted-floats", "[clp-s][end-to-end]") {
```
anlowee: Consider use another name instead of "rounded" here and the test file name (`test_rounded...`). Maybe `dictionarized-valid-formatted-floats`?

components/core/src/clp_s/ArchiveReader.cpp
gibber9809: Good catch.

components/core/src/clp_s/ColumnWriter.cpp
gibber9809: Sure

components/core/tests/test_log_files/test_valid_formatted_float.jsonl
gibber9809: Sure, I'll probably also add some some cases for max float and min subnormal float.

components/core/tests/test-clp_s-end_to_end.cpp
gibber9809: How about `clp-s-compress-extract-invalid-formatted-floats`?

components/core/tests/test_log_files/test_valid_formatted_float.jsonl
gibber9809: Actually I wouldn't expect these to round-trip correctly -- while it is technically equivalent to zero no standard marshalling implementation would use a non-zero exponent for zero values. We should probably explicitly check for this case in `get_float_encoding` and reject it.

components/core/src/clp_s/FloatFormatEncoding.cpp
gibber9809: You're right, this should always evaluate to true. I'll remove it.

components/core/tests/test-FloatFormatEncoding.cpp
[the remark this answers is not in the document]
gibber9809: Rabbit is wrong here because g is redundant when we're testing both 'e' and 'f' -- the general semantic of 'g' is that it will be equivalent to 'e' or 'f' depending on which would produce a shorter output (plus the semantic for number of significant digits is a bit different, but it doesn't matter for this particular test case).

components/core/tests/test_log_files/test_valid_formatted_float.jsonl
gibber9809: Added that check +the subnormal cases + modified your suggested case to just have E0000.

components/core/src/clp_s/ColumnWriter.cpp
```diff
@@ -46,6 +49,33 @@ void FloatColumnWriter::store(ZstdCompressor& compressor) {
...
+    auto const format_size = m_formats.size() * sizeof(float_format_t);
+    compressor.write(reinterpret_cast<char const*>(m_values.data()), values_size);
+    compressor.write(reinterpret_cast<char const*>(m_formats.data()), format_size);
+}
+
+size_t DictionaryFloatColumnWriter::add_value(ParsedMessage::variable_t& value) {
+    clp::variable_dictionary_id_t id{};
+    m_var_dict->add_entry(std::get<std::string>(value), id);
+    m_var_dict_ids.push_back(id);
+    return sizeof(variable_dictionary_id_t);
```
anlowee:
```suggestion
    return sizeof(clp::variable_dictionary_id_t);
```

anlowee: to make it consistent with three lines above

components/core/tests/test_log_files/test_invalid_formatted_float.jsonl
```diff
@@ -0,0 +1,43 @@
...
+{ "scientificNegativeExponentExceedingDigitLimitWithLeadingZeros": 1.234567891234567E-00008 }
+{ "negativeScientificNegativeExponentExceedingDigitLimitWithLeadingZeros": -1.234567891234567E-00008 }
+{ "scientificSignificandsExceedingPrecisionLimitWithTrailingZerosWithNegativeExponentExceedingDigitLimitWithLeadingZeros": 1.23456789123456700E-00008 }
+{ "negativeScientificSignificandsExceedingPrecisionLimitWithTrailingZerosWithPositiveExponentExceedingDigitLimitWithLeadingZerosWithPlus": -1.23456789123456700E+00008 }
+{ "scientificWithGreaterThanTenMantissa": 123456789.1234567E3 }
+{ "negativeScientificWithGreaterThanTenMantissa": -123456789.1234567E3 }
+{ "scientificWithGreaterThanTenMantissaExceedingPrecision": 123456789.123456789E3 }
+{ "negativeScientificWithGreaterThanTenMantissaExceedingPrecision": -123456789.123456789E3 }
+{ "scientificWithGreaterThanTenMantissa2": 12.34567891234567E01 }
+{ "-negativeScientificWithGreaterThanTenMantissa2": -12.34567891234567E01 }
```
anlowee:
```suggestion
{ "negativeScientificWithGreaterThanTenMantissa2": -12.34567891234567E01 }
```

anlowee:
LGTM, but I can't approve the PR created by me. @kirkrodrigues @LinZhihao-723 could either of you approve this PR?

PR #1179, opened 2025-08-08
feat(webui): Display search results in Presto UI

davemarco:
# Description

Draft

# Checklist

  breaking change.

# Validation performed

### DIFF
--- a/components/webui/client/src/pages/SearchPage/SearchControls/Presto/presto-search-requests.ts
+++ b/components/webui/client/src/pages/SearchPage/SearchControls/Presto/presto-search-requests.ts
@@ -2,6 +2,7 @@ import {
     type PrestoQueryJobCreationSchema,
     submitQuery,
 } from "../../../../api/presto-search";
+import useSearchStore from "../../SearchState/";
 
 
 /**
@@ -10,9 +11,11 @@ import {
  * @param payload
  */
 const handlePrestoQuerySubmit = (payload: PrestoQueryJobCreationSchema) => {
+    const store = useSearchStore.getState();
     submitQuery(payload)
         .then((result) => {
             const {searchJobId} = result.data;
+            store.updateSearchJobId(searchJobId);
             console.debug(
                 "Presto search job created - ",
                 "Search job ID:",
--- a/components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Presto/typings.tsx
+++ b/components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Presto/typings.tsx
@@ -1,0 +1,35 @@
+import {TableProps} from "antd";
+
+
+/**
+ * Structure of Presto search results data (dynamic properties).
+ */
+interface PrestoSearchResult {
+    _id: string;
+    row: Record<string, unknown>;
+}
+
+/**
+ * Columns configuration for Presto query engine (dynamic based on data).
+ *
+ * @param data Array of Presto search results
+ * @return
+ */
+const getPrestoSearchResultsTableColumns = (
+    data: PrestoSearchResult[]
+): NonNullable<TableProps<PrestoSearchResult>["columns"]> => {
+    if (0 === data.length || "undefined" === typeof data[0] || "undefined" === typeof data[0].row) {
+        return [];
+    }
+
+    return Object.keys(data[0].row)
+        .map((key) => ({
+            dataIndex: ["row",
+                key],
+            key: key,
+            title: key,
+        }));
+};
+
+export type {PrestoSearchResult};
+export {getPrestoSearchResultsTableColumns};
--- a/components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Presto/usePrestoSearchResults.ts
+++ b/components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Presto/usePrestoSearchResults.ts
@@ -1,0 +1,48 @@
+import MongoSocketCollection from "../../../../../api/socket/MongoSocketCollection";
+import {useCursor} from "../../../../../api/socket/useCursor";
+import useSearchStore, {SEARCH_STATE_DEFAULT} from "../../../SearchState/index";
+import {SEARCH_MAX_NUM_RESULTS} from "../typings";
+import {PrestoSearchResult} from "./typings";
+
+
+/**
+ * Custom hook to get Presto search results for the current searchJobId.
+ *
+ * @return
+ */
+const usePrestoSearchResults = () => {
+    const {searchJobId} = useSearchStore();
+
+    const searchResultsCursor = useCursor<PrestoSearchResult>(
+        () => {
+            // If there is no active search job, there are no results to fetch. The cursor will
+            // return null.
+            if (searchJobId === SEARCH_STATE_DEFAULT.searchJobId) {
+                return null;
+            }
+
+            console.log(
+                `Subscribing to updates to Presto search results with job ID: ${searchJobId}`
+            );
+
+            // Retrieve 1k most recent results.
+            const options = {
+                sort: [
+                    [
+                        "_id",
+                        "desc",
+                    ],
+                ],
+                limit: SEARCH_MAX_NUM_RESULTS,
+            };
+
+            const collection = new MongoSocketCollection(searchJobId.toString());
+            return collection.find({}, options);
+        },
+        [searchJobId]
+    );
+
+    return searchResultsCursor;
+};
+
+export {usePrestoSearchResults};
--- a/components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/SearchResultsVirtualTable.tsx
+++ b/components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/SearchResultsVirtualTable.tsx
@@ -1,0 +1,48 @@
+import {useEffect} from "react";
+
+import VirtualTable from "../../../../components/VirtualTable";
+import useSearchStore from "../../SearchState/index";
+import {
+    SearchResult,
+    searchResultsTableColumns,
+} from "./typings";
+import {useSearchResults} from "./useSearchResults";
+
+
+interface SearchResultsVirtualTableProps {
+    tableHeight: number;
+}
+
+/**
+ * Renders search results in a virtual table.
+ *
+ * @param props
+ * @param props.tableHeight
+ * @return
+ */
+const SearchResultsVirtualTable = ({tableHeight}: SearchResultsVirtualTableProps) => {
+    const {updateNumSearchResultsTable} = useSearchStore();
+    const searchResults = useSearchResults();
+
+    useEffect(() => {
+        const num = searchResults ?
+            searchResults.length :
+            0;
+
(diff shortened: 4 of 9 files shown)


components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Presto/usePrestoSearchResults.ts
```diff
@@ -0,0 +1,48 @@
...
+                sort: [
+                    [
+                        "_id",
+                        "desc",
+                    ],
+                ],
+                limit: SEARCH_MAX_NUM_RESULTS,
+            };
+
+            const collection = new MongoSocketCollection(searchJobId.toString());
```
hoophalab: we probably should remove `toString()` here and other places

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/SearchResultsVirtualTable.tsx
```diff
@@ -0,0 +1,48 @@
...
+ * @param props
+ * @param props.tableHeight
+ * @return
+ */
+const SearchResultsVirtualTable = ({tableHeight}: SearchResultsVirtualTableProps) => {
+    const {updateNumSearchResultsTable} = useSearchStore();
+    const searchResults = useSearchResults();
+
+    useEffect(() => {
+        const num = searchResults ?
```
hoophalab: How about organizing this folder a bit cleaner
1. move this `SearchResultsVirtualTable.tsx` and `typings.tsx` in to folder `SearchResultsVirtualTable`
2. rename `Presto` folder as `PrestoResultsVirtualTable`
3. rename `Presto/PrestoResultsVirtualTable.tsx` as `Presto/index.tsx`

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Presto/typings.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Generates dynamic columns configuration for Presto query engine.
+ *
+ * @param data Array of Presto search results
+ * @return
+ */
+const getPrestoSearchResultsTableColumns = (
+    data: PrestoSearchResult[]
+): NonNullable<TableProps<PrestoSearchResult>["columns"]> => {
```
hoophalab: this function probably should be in `utils.ts`? Or, we can just put it inside `PrestoResultsVirtualTable` if it doesn't exceeds line num limit.

components/webui/server/src/routes/api/presto-search/index.ts
```diff
@@ -115,6 +115,8 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
                 throw error;
             }
 
+            await mongoDb.createCollection(searchJobId);
```
hoophalab, line 171: I know this cannot happen in presto-client-node, but in case we change to a more async client, how about adding this comment:
```
// In the very unlikely situation where the first data arrives before we create the collection,
// MongoDB automatically creates a collection on the first insert if it doesn't exist.
// In this case, `createCollection` does nothing instead of throwing an error, which is what happens in pymongo.
```

components/webui/client/src/pages/SearchPage/SearchControls/Presto/presto-search-requests.ts
```diff
@@ -10,9 +11,11 @@ import {
  * @param payload
  */
 const handlePrestoQuerySubmit = (payload: PrestoQueryJobCreationSchema) => {
+    const store = useSearchStore.getState();
```
hoophalab:
```suggestion
    const {updateSearchJobId} = useSearchStore.getState();
```

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Presto/usePrestoSearchResults.ts
davemarco: ur right not sure why was there, perhaps was an int at once point

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/SearchResultsVirtualTable.tsx
davemarco: i did something slightly different since I want to keep the `Presto` folder. I am trying to keep a style where everything presto is in a `Presto` folder. But overall restructured per your comment

components/webui/server/src/routes/api/presto-search/index.ts
davemarco: I think with the current code the situation you describe is not possible. see line [here](https://github.com/y-scope/clp/blob/3a8caa000595220dd494cbfa7e7caf0d354788b5/components/webui/server/src/routes/api/presto-search/index.ts#L77). Like if the data comes first, then it will just exit early and print a log, like it wont create a collection. I dont think the comment is neccesary since if it is actually possible which i doubt, then it will print error log

davemarco (commented):
addresses comments, and added some constants where before we had hardcoded strings as quality improvement

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Presto/PrestoResultsVirtualTable/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Generates dynamic columns configuration for Presto query engine.
+ *
+ * @param data Array of Presto search results
+ * @return
+ */
+const getPrestoSearchResultsTableColumns = (
+    data: PrestoSearchResult[]
+): NonNullable<TableProps<PrestoSearchResult>["columns"]> => {
+    if (0 === data.length || "undefined" === typeof data[0] || "undefined" === typeof data[0].row) {
```
hoophalab: `.row` is not ideal. I prefer to remove `PRESTO_DATA_PROPERTY`. The correct handling of the search result type should be having a type in common, and when you need to rename it, the language server handles it correctly. Currently, it simply provides little help and adds the work of renaming

components/webui/server/src/routes/api/presto-search/index.ts
hoophalab: We should do better than only runtime correctness, but I'm fine if we don't add a comment here as it doesn't make the code better and we might forget to update comments in the future.

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Presto/PrestoResultsVirtualTable/utils.ts
hoophalab: ahh I see why you added PRESTO_DATA_PROPERTY, because we need to pass a string in `dataIndex: ["row",`. How about only using `PRESTO_DATA_PROPERTY` in `dataIndex` and leaving the type untouched? Also add a comment on the definition of `PRESTO_DATA_PROPERTY`.

davemarco: I got rid of the .row, and use PRESTO_DATA_PROPERTY there as well. I think since it will always be there, might as well keep it on the type. There is already a comment on the definition no? You get the idea right, like if the column in the database has a _id, then it would overwrite the mongo _id field, so now it wont do that. 

In short, i think it makes code better that just hardcoded "row", now we can change it easier in one spot

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Presto/PrestoResultsVirtualTable/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+    data: PrestoSearchResult[]
+): NonNullable<TableProps<PrestoSearchResult>["columns"]> => {
+    if (0 === data.length ||
+        "undefined" === typeof data[0] ||
+        "undefined" === typeof data[0][PRESTO_DATA_PROPERTY]
+    ) {
+        return [];
+    }
+
+    return Object.keys(data[0].row)
```
hoophalab, line 22:
``` suggestion
    return Object.keys(data[0][PRESTO_DATA_PROPERTY])
```

I'm still against `PRESTO_DATA_PROPERTY`. Imagine if the antd table doesn't require a string as `dataIndex`, you probably won't have `PRESTO_DATA_PROPERTY` and define a regular interface
```
interface PrestoSearchResult {
    _id: string;
    row: Record<string, unknown>;
}
```
and you dont need `obj[PRESTO_DATA_PROPERTY]` everytime.
Why do we change the type definition simply because of one library we use?

However, this is mainly your project and I respect your decision.  I have no issue if we simply fix this last `.row`

davemarco: okay i dont fully understand what you mean but lets discuss in meeting

davemarco: okay i removed per discussion

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Presto/PrestoResultsVirtualTable/typings.ts
```diff
@@ -1,9 +1,11 @@
+import type {PrestoRowObject} from "../../../../../../../../common/index.js";
+
+
 /**
  * Structure of dynamic Presto search results data.
  */
-interface PrestoSearchResult {
+interface PrestoSearchResult extends PrestoRowObject {
     _id: string;
```
hoophalab: sorry, but can we also move `interface PrestoSearchResult` to `common`?
Because it is the type of objects stored in the mongodb.
If someone needs to understand our types in the protocol between server and client, he/she only needs to take a look at the common folder

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Presto/PrestoResultsVirtualTable/typings.ts
```diff
@@ -0,0 +1,11 @@
+import type {PrestoRowObject} from "../../../../../../../../common/index.js";
+
+
+/**
+ * Structure of dynamic Presto search results data.
+ */
+interface PrestoSearchResult extends PrestoRowObject {
+    _id: string;
+}
```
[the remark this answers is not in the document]
hoophalab: @coderabbitai Can you create an issue on this?

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Presto/PrestoResultsVirtualTable/typings.ts
hoophalab: and there is a lint error in cicd

PR #1181, opened 2025-08-08
docs(package): Add user documents for retention control; Add garbage collector to the multi-node guide.

haiqi96:
# Description

As the title describes, this PR adds the user doc for retention control.
Note that one feature described in this PR depends on https://github.com/y-scope/clp/pull/1208

# Checklist

  breaking change.

# Validation performed

Locally built the docs and confirmed it looks as expected.

## Summary by CodeRabbit

* **Documentation**
  * Added a “Retention control” guide covering retention concepts, TTLs, sweep intervals, expiry rules, timezone handling, examples, race‑condition handling, observability, and recovery; linked it into the user‑guide and toctree.
  * Updated multi‑node docs to include a garbage‑collector controller and adjusted start order; replaced a Presto card with a Retention card and added a “Using Presto” card (grid now 4 items).
* **Chores**
  * Enabled dollarmath support in the documentation build.

### DIFF
--- a/docs/src/user-guide/guides-overview.md
+++ b/docs/src/user-guide/guides-overview.md
@@ -14,9 +14,9 @@ Using CLP to ingest logs from object storage and store archives on object storag
 
 :::{grid-item-card}
 :link: guides-retention
-Retention control
+Configuring retention periods
 ^^^
-How to configure retention control for CLP.
+How to configure retention periods for CLP archives and search results.
 :::
 
 :::{grid-item-card}
--- a/docs/src/user-guide/guides-retention.md
+++ b/docs/src/user-guide/guides-retention.md
@@ -1,4 +1,4 @@
-# Setting up data retention periods
+# Configuring retention periods
 
 CLP can automatically delete *archives* and/or *search results* once they're older than a configured
 retention period. This guide explains:


docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
+# Retention control in CLP
+
+CLP supports retention control to periodically delete outdated archives and search results to free
+storage space. Retention applies to both the local filesystem and object storage.
```
quinntaylormitchell:
```suggestion
CLP supports retention control to free up storage space by periodically deleting outdated archives
and search results. Retention applies to both the local filesystem and object storage.
```

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+
+This process is managed by background **garbage collector** jobs, which scan for and delete expired
+data based on configured retention settings in `etc/clp-config.yml`.
+
+:::{note}
+By default, retention control is disabled and data will be kept indefinitely in CLP.
```
quinntaylormitchell:
```suggestion
By default, retention control is disabled, and CLP retains data indefinitely.
```

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+:::
+
```
quinntaylormitchell, line 11:
```suggestion

---

```
add a line for readability

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+## Definitions
+
+This section defines the terms and criteria used by CLP to determine when data should be 
+deleted.
+
+### Terms
+
+- **Current Time (T):** The wall-clock time at which the garbage collector job is executed.
+- **Retention Period (TTL):** The configured duration after which data is considered expired.
```
quinntaylormitchell: For clarity: does Retention Period (TTL) indicate a span of time that the log should be retained for, or does it indicate some future date/time after which the log should not be retained?

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Expiry criteria
+
+- **Archive Expiry:**  
+  An archive is considered expired if its largest log message timestamp exceeds the retention
+  period:
```
quinntaylormitchell:
```suggestion
  An archive is considered expired if the archive's TTL has elapsed since its most recent log event,
  i.e. that the difference between T and `archive.largest_msg_ts` has surpassed TTL.
```
this seems more intuitive to me

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+- **Current Time (T):** The wall-clock time at which the garbage collector job is executed.
+- **Retention Period (TTL):** The configured duration after which data is considered expired.
+
+### Expiry criteria
+
+- **Archive Expiry:**  
+  An archive is considered expired if its largest log message timestamp exceeds the retention
+  period:  
+  ```text
+  archive.largest_msg_ts < T - TTL
```
quinntaylormitchell:
```suggestion
  if (T - archive.largest_msg_ts > TTL) then EXPIRED
```
more intuitive

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+  ```
+:::{note}
+Archives whose log messages do not contain timestamps are not subject to retention.
+:::
+
+:::{caution}
+Archive expiry is based on the `timestamp of the log messages`, not on when the archive is
+compressed. Compressing old logs with retention enabled could cause those archives to be deleted
+immediately after compression.
+:::
```
quinntaylormitchell:
```suggestion
  :::{note}
  Archives whose log messages do not contain timestamps are not subject to retention.
  :::
  
  :::{caution}
  Archive expiry is based on the **timestamp of the log messages**, not on when the archive is
  compressed. Compressing old logs with retention enabled could cause those archives to be deleted
  immediately after compression.
  :::
```
this should be indented

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+  period:  
+  ```text
+  archive.largest_msg_ts < T - TTL
+  ```
+:::{note}
+Archives whose log messages do not contain timestamps are not subject to retention.
+:::
+
+:::{caution}
+Archive expiry is based on the `timestamp of the log messages`, not on when the archive is
```
quinntaylormitchell:
```suggestion
Archive expiry is based on the **timestamp of the log messages**, not on when the archive is
```
this should be bolded rather than monospaced

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+compressed. Compressing old logs with retention enabled could cause those archives to be deleted
+immediately after compression.
+:::
+
+- **Search Result Expiry:** 
+  A search result is considered expired if it's finish time exceeds the retention period:
```
quinntaylormitchell:
```suggestion
  A search result is considered expired if the result's TTL has elapsed since the search was 
  completed, i.e. that the difference between T and `search_job.completion_time` has surpassed TTL.
```
intuitive (as above)

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+  ```text
+  search_job.finished_time < T - TTL.
```
quinntaylormitchell:
```suggestion
  if (T - search_job.completion_time > TTL) then EXPIRED
```
more intuitive (as above)

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+  ```
+
+
+## Configuration
+Retention setting can be configured in `etc/clp-config.yml`.
```
quinntaylormitchell:
```suggestion
Retention settings can be configured in `etc/clp-config.yml`.
```

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+compressed. Compressing old logs with retention enabled could cause those archives to be deleted
+immediately after compression.
+:::
+
+- **Search Result Expiry:** 
+  A search result is considered expired if it's finish time exceeds the retention period:
+  ```text
+  search_job.finished_time < T - TTL.
+  ```
+
```
quinntaylormitchell, line 43:
```suggestion

---
```
add line for readability

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Configuration
+Retention setting can be configured in `etc/clp-config.yml`.
+
+### Configure retention period
+To configure a retention period, update the `retention_period`
+key in `etc/clp-config.yml` with desired retention period in minutes.
```
quinntaylormitchell:
```suggestion
To configure a retention period, update the `retention_period` key in `etc/clp-config.yml` with the
desired retention period in minutes.
```
lint and added "the"

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+  # Other archive output specific settings
+
+  # Retention period for archives, in minutes. 
+  # Set to null to disable automatic deletion.
+  retention_period: 43200
+```
+
+### Configure sweep interval
+**Sweep interval** specifies the interval at which garbage collector jobs run to collect and delete
+expired data.
```
quinntaylormitchell:
```suggestion
**Sweep interval** specifies the time interval at which garbage collector jobs run to collect and
delete expired data.
```
add "time" and adjust line length accordingly

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+
+To configure a custom sweep frequency for different retention targets, set the fields under
+`garbage_collector.sweep_interval` in `etc/clp-config.yml`:
```
quinntaylormitchell:
```suggestion
To configure a custom sweep frequency for different retention targets, you can set the subfields
under `garbage_collector.sweep_interval` individually in `etc/clp-config.yml`. For example, to
configure a sweep interval of 15 minutes for search results but 3 hours (180 minutes) for archives,
enter the following:
```
modified for clarity

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+
+```yaml
+garbage_collector:
+  logging_level: "INFO"
+
+  # Interval (in minutes) at which garbage collector jobs run
+  sweep_interval:
+    archive: <sweep_interval>
+    search_result: <sweep_interval>
```
quinntaylormitchell:
```suggestion
    archive: 180
    search_result: 15
```
to match the example I suggested above

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+:::{note}
+If the retention period is set to `null`, the corresponding garbage collection task will not run 
+even if `sweep_interval` is configured.
+:::
+
+## Handle race condition
```
quinntaylormitchell:
```suggestion
## Handling data race conditions
```

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+Retention is designed to avoid the race condition where it deletes archives that might still be in
+use by active search jobs.
+CLP employs the following mechanisms to avoid race conditions:
```
quinntaylormitchell:
```suggestion
CLP's retention system is designed to avoid data race conditions that may arise from the deletion of
archives or search results that may still be in use by active jobs. CLP employs the following
mechanisms to avoid these conditions:
```

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+
+- If any query job is running, CLP conservatively calculates a **safe expiry timestamp** based on 
+  the earliest active search job. This ensures no archive that could be searched is deleted.
+
+- CLP will **not** search an archive once it is considered expired, even if it has not yet been
+  deleted by the garbage collector.
+
+:::{warning}
+A hanging search job will prevent CLP from deleting expired archives. 
+Restarting the query scheduler will mark such jobs as failed and allow retention to resume.
```
quinntaylormitchell:
```suggestion
Restarting the query scheduler will mark such jobs as failed and allow garbage collection to resume.
```

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+:::
+
+## Fault tolerance
+The garbage collector can resume execution from where it left off if a previous run fails. 
+This design ensures that CLP does not fall into an inconsistent state due to partial deletions.
+
+If the CLP package stops unexpectedly (for example, due to a host machine shutdown) when a garbage
+collection task was running, simply restart the package and the garbage collector will continue
```
quinntaylormitchell:
```suggestion
If the CLP package stops unexpectedly (for example, due to a host machine shutdown) while a garbage
collection task is running, simply restart the package and the garbage collector will continue 
```
verb tense/word choice

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,115 @@
...
+The garbage collector can resume execution from where it left off if a previous run fails. 
+This design ensures that CLP does not fall into an inconsistent state due to partial deletions.
+
+If the CLP package stops unexpectedly (for example, due to a host machine shutdown) when a garbage
+collection task was running, simply restart the package and the garbage collector will continue 
+from the point of failure.
+
+:::{note}
+During failure recovery, there may be a temporary period when an archive is deleted from the
+database but still exists on disk or in object storage. Once recovery completes, the physical
```
quinntaylormitchell:
```suggestion
During failure recovery, there may be a temporary period during which an archive no longer exists in
the database, but still exists on disk or in object storage. Once recovery is complete, the physical
```
clarity

quinntaylormitchell (changes requested):
Looking good @haiqi96! Let me know if there are any problems with my comments.

And in answer to your high-level question re. location in the docs, it seems logical to me to include the retention doc in the guides section, just as you've done it.

docs/src/user-guide/guides-retention.md
haiqi96: Merged this suggestions with the next one so there's no conflict.

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,122 @@
...
+
+If the CLP package stops unexpectedly (for example, due to a host machine shutdown) while a garbage
+collection task is running, simply restart the package and the garbage collector will continue 
+from the point of failure.
+
+:::{note}
+During failure recovery, there may be a temporary period during which an archive no longer exists in
+the database, but still exists on disk or in object storage. Once recovery is complete, the physical
+archive will also be deleted.
+:::
```
quinntaylormitchell: nit: put an empty line at the end of the file (not required but I think it's good form)

quinntaylormitchell (approved):
Great job @haiqi96. other than one nit comment re. adding a empty line at the end of `guides-retention.md`, I approve!

haiqi96:
Reminder to myself: add a note somewhere about UTC timestamp.

docs/src/user-guide/guides-multi-node.md
```diff
@@ -71,7 +72,9 @@ Running additional workers increases the parallelism of compression and search/a
     4. Set `archive_output.directory` to a directory on the distributed filesystem.
        * Ideally, the directory should be empty or should not yet exist (CLP will create it) since
          CLP will write several files and directories directly to the given directory.
-
+    5. (Optional) Configure retention periods for archives and search results. For details, see
+       [retention control](guides-retention) for details.
```
quinntaylormitchell:
```suggestion
    5. (Optional) Configure retention periods for archives and search results. See
       [retention control](guides-retention) for details.
```

docs/src/user-guide/guides-multi-node.md
```diff
@@ -175,6 +179,7 @@ cases, see the [SeaweedFS docs][seaweedfs-docs].
 
 [Docker]: https://docs.docker.com/engine/install/
 [docker-non-root]: https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user
+[Retention control]:
```
quinntaylormitchell:
```suggestion
```
not needed

docs/src/user-guide/guides-overview.md
```diff
@@ -25,4 +25,11 @@ Multi-node deployment
 ^^^
 How to deploy CLP across multiple nodes.
 :::
+
+:::{grid-item-card}
+:link: guides-retention
+Retention Control
+^^^
+How to configure retention control for CLP.
+:::
```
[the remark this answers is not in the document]
quinntaylormitchell: Agreed.

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+CLP supports retention control to free up storage space by periodically deleting outdated archives
+and search results. Retention applies to both the local filesystem and object storage.
+
+This process is managed by background **garbage collector** jobs, which scan for and delete expired
+data based on configured retention settings in `etc/clp-config.yml`.
+
+:::{note}
+By default, retention control is disabled, and CLP retains data indefinitely.
+:::
+
```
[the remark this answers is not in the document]
quinntaylormitchell: @haiqi96 I've given some more specific rewording suggestions in my comments; you can ignore this

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
+# Retention control in CLP
+
+CLP supports retention control to free up storage space by periodically deleting outdated archives
+and search results. Retention applies to both the local filesystem and object storage.
+
+This process is managed by background **garbage collector** jobs, which scan for and delete expired
+data based on configured retention settings in `etc/clp-config.yml`.
```
quinntaylormitchell:
```suggestion
CLP’s retention control system periodically removes outdated archives and search results. Retention
applies to both the local filesystem and object storage.

Background **garbage collector** jobs identify data that has exceeded its retention period and
delete it according to the configured specifications in `etc/clp-config.yml`.
```
clarity and flow

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+:::{note}
+By default, retention control is disabled, and CLP retains data indefinitely.
+:::
+
+---
+
+## Definitions
+CLP supports retention control for archives and search results.
```
quinntaylormitchell:
```suggestion

```
I think this is redundant

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+At a high level, CLP compares a data item's timestamp with the current time to determine whether
+it has expired, but the criteria differ slightly between archives and search results.
```
quinntaylormitchell:
```suggestion
it has expired. The criteria used to assess this expiration differs slightly between archives and
search results.
```
clarity and flow

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+This section explains the terms and criteria CLP uses to decide when data should be deleted.
```
quinntaylormitchell: I think that you should move this line so that it's the first thing under `## Definitions`.

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+- **Search result's timestamp (search_result.T):** The timestamp when a search result is inserted
+  into the results_cache.
+
+  :::{Note}
+  1. Archive's timestamp is based on the **timestamp of its log messages**, not on when the archive
+  is compressed. Compressing old logs with retention enabled could cause those archives to be
+  deleted immediately after compression.
+  
+  2. Archives whose log messages do not contain timestamps are not subject to retention.
+  :::
```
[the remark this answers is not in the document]
quinntaylormitchell: @haiqi96 I agree about the directive casing, and I also agree about the indentation (unless you wanted the note to be at the same indentation as the list; fwiw I like it better if the indentation is at the same level as the heading, as Rabbit suggests).

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Definitions
+CLP supports retention control for archives and search results.
+At a high level, CLP compares a data item's timestamp with the current time to determine whether
+it has expired, but the criteria differ slightly between archives and search results.
+
+This section explains the terms and criteria CLP uses to decide when data should be deleted.
+
+### Terms
+- **Current Time (T):** The present wall-clock time when data expiry is being evaluated.
```
quinntaylormitchell:
```suggestion
- **Current Time (T):** The current time (UTC), taken by a garbage collector when evaluating data
  expiration.
```
I think that using the phrase "wall-clock time" incorrectly indicates that the system time will be used, when in reality the UTC conversion of the wall-clock time will be used.

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+- **Retention Period (TTL):** The configured duration for which CLP retains data before it is
+  considered expired.
+- **Archive's timestamp (archive.T):** The most recent timestamp among all log messages
+  contained in the archive.
+- **Search result's timestamp (search_result.T):** The timestamp when a search result is inserted
+  into the results_cache.
```
quinntaylormitchell:
```suggestion
- **Archive timestamp (`archive.T`):** The most recent timestamp among all log messages held in the
  archive.
- **Search result timestamp (`search_result.T`):** The timestamp when a search result is inserted
  into the results_cache.
```
Using monospace for clarity; got rid of "'s" possessor as it sounds a bit awkward for a list

docs/src/user-guide/guides-retention.md
quinntaylormitchell:
```suggestion
:::{note}
Archives whose log messages do not contain timestamps are not subject to retention.
:::
```
ultimately it feels awkward to have two points in a note block. I think the second one should stay here, and the first one should be worked into definition 3 (see that comment for details).

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+it has expired, but the criteria differ slightly between archives and search results.
+
+This section explains the terms and criteria CLP uses to decide when data should be deleted.
+
+### Terms
+- **Current Time (T):** The present wall-clock time when data expiry is being evaluated.
+- **Retention Period (TTL):** The configured duration for which CLP retains data before it is
+  considered expired.
+- **Archive's timestamp (archive.T):** The most recent timestamp among all log messages
+  contained in the archive.
```
quinntaylormitchell:
```suggestion
- **Archive timestamp (`archive.T`):** The most recent log timestamp among all log messages held in
  the archive. Not related to the time at which the logs were compressed. Note that logs with old
  timestamps may be deleted immediately, depending on your retention settings.
```
this suggestion incorporates point 1. from the {note} (see comment below)

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+  deleted immediately after compression.
+  
+  2. Archives whose log messages do not contain timestamps are not subject to retention.
+  :::
+
+### Expiry criteria
+
+- **Archive Expiry:**  
+  An archive is considered expired if its retention period has elapsed since archive's timestamp,
+  i.e. that the difference between `T` and `archive.T` has surpassed `TTL`.
```
quinntaylormitchell:
```suggestion
  An archive is considered expired if its retention period has elapsed since its most recent
  timestamp, i.e., that the difference between `T` and `archive.T` has surpassed `TTL`.
```
improve flow slightly

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+  ```text
+  if (T - archive.T > TTL) then EXPIRED
+  ```
```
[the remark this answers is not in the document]
quinntaylormitchell: I like mine better; see my comment

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+  For example, if a compressed archive has `archive.T = 16:00` with a `TTL = 1 hour`, it will be
+  considered expired after `T = 17:00` since `T - 16:00 > 1:00` for all `T > 17:00`(for simplicity,
+  we omit dates and seconds from the timestamp).
```
quinntaylormitchell:
```suggestion
  For example, if a compressed archive has `archive.T = 16:00`, and `TTL = 1 hour`, it will be
  considered expired after `T = 17:00`, since `T - 16:00 > 1:00` for all `T > 17:00`.
```
small changes for readability; I also don't think you need the "for simplicity..." parenthetical.

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
```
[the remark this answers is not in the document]
quinntaylormitchell: incorporated into my comment; you can ignore

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+  :::{caution}
+  Retention control operates on **UTC timestamps** and assumes that archive timestamps are also in
+  UTC. Using archives with local timestamps can lead to an effective `TTL` that is different from 
+  the intended value.
```
quinntaylormitchell:
```suggestion
  Retention control assumes that archive timestamps are given in **UTC** time. Using retention
  control on archives with local (i.e., non-UTC) timestamps can lead to an effective `TTL` that is
  different from the intended value.
```
clarity

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+  In the example above, if the package operates in EST (UTC-4) and `archive.T = 16:00` is a local
+  timestamp, then at 16:30 local time, the garbage collector first converts 16:30 EST to 20:30 UTC
+  and the calculation would be `20:30 - 16:00 > 1:00`.
+  In this case, the archive would be determined to be expired and deleted, even though it has not 
+  actually reached its intended retention period.
```
quinntaylormitchell:
```suggestion
  In the example above, if the package operates on a system in EDT (UTC-4) and `archive.T = 16:00`
  is a local timestamp, then a garbage collection job operating at 16:30 local time will convert
  `16:30 EDT` to `20:30 UTC`, and the expiry calculation will be `20:30 - 16:00 > 1:00`. In this
  case, the archive would be considered expired, and would be deleted, even though it wouldn't have
  actually reached its intended retention period.
```
clarity and flow; also, EST is (UTC-5) and EDT is (UDT-4)

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+  To avoid this issue, either generate logs with UTC timestamps or adjust the retention period to
+  account for the offset:
+  `adjusted_retention_period = retention_period - UTC offset`
```
quinntaylormitchell:
```suggestion
  
  `adjusted_retention_period = retention_period - signed_UTC_offset`
```
add newline, underscore, and more-detailed name

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+  :::
```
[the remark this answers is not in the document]
quinntaylormitchell: rabbit's suggestion is incorrect. increasing by the absolute UTC offset will over-retain data in positive-offset timezones.

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+- **Search Result Expiry:** 
+  A search result is considered expired if its retention period has elapsed since the search was 
+  completed, i.e. that the difference between T and `search_result.T` has surpassed TTL.
+  ```text
+  if (T - search_result.T > TTL) then EXPIRED
+  ```
```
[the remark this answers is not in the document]
quinntaylormitchell: I agree with the use of monospace, but I still want the "that" in there.

(also, fwiw @coderabbitai, your title is pretty misleading; removing the word "that" and adding monospace formatting doesn't constitute "Tightening the grammar".)

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+---
+
+## Configuration
+CLP allows users to configure different **retention_periods** for different types of data. 
+Additionally, a custom **sweep_interval** can be set for each type of data, which determines how
+frequently its garbage collection job is executed. These settings can be configured in 
+`etc/clp-config.yml`. 
+
```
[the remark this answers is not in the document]
quinntaylormitchell: I agree with this; the structure of the sentence indicates you're referring to the general concept. If you want to refer to the actual fields called `retention_period` in `etc/clp-config.yml`, see my comment

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+  if (T - search_result.T > TTL) then EXPIRED
+  ```
+
+---
+
+## Configuration
+CLP allows users to configure different **retention_periods** for different types of data. 
+Additionally, a custom **sweep_interval** can be set for each type of data, which determines how
+frequently its garbage collection job is executed. These settings can be configured in 
+`etc/clp-config.yml`.
```
quinntaylormitchell:
```suggestion
CLP allows users to specify different retention periods for archival and search data in the
**`.retention_period`** fields of `etc/clp-config.yml`. Additionally, the frequency of garbage
collection job execution for each type of data can be configured by setting the appropriate
subfields in the **`garbage_collector.sweep_interval`** parameter.
```
clarity and flow

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Configuration
+CLP allows users to configure different **retention_periods** for different types of data. 
+Additionally, a custom **sweep_interval** can be set for each type of data, which determines how
+frequently its garbage collection job is executed. These settings can be configured in 
+`etc/clp-config.yml`. 
+
+### Configure retention period
+To configure a retention period, update the `retention_period` key in `etc/clp-config.yml` with the
+desired retention period in minutes.
```
quinntaylormitchell:
```suggestion
To configure a retention period, update the appropriate `.retention_period` key in
`etc/clp-config.yml` with the desired retention period in minutes.
```
adding the word "appropriate" because there are two `.retention_period` keys

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+For example, to configure an archive retention period of 30 days (43,200 minutes):
+```yaml
+archive_output:
+  # Other archive output specific settings
```
quinntaylormitchell:
```suggestion
  # Other archive_output settings
```

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+```yaml
+results_cache:
+  # Other results_cache settings
+
+  # Retention period for search results, in minutes. 
+  # Set to null to disable automatic deletion.
+  retention_period: 1440
+```
+
+### Configure sweep interval
```
quinntaylormitchell:
```suggestion
### Configure sweep interval

```

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+**Sweep interval** specifies the time interval at which garbage collector jobs run to collect and
+delete expired data.
```
quinntaylormitchell:
```suggestion
The **`garbage_collector.sweep_interval`** parameter specifies the time interval at which garbage
collector jobs run to collect and delete expired data.
```
specificity

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+  # Interval (in minutes) at which garbage collector jobs run
+  sweep_interval:
+    archive: 180
+    search_result: 15
+```
+
+:::{note}
+If the retention period is set to `null`, the corresponding garbage collection task will not run 
+even if `sweep_interval` is configured.
```
quinntaylormitchell:
```suggestion
If either of the `.retention_period` fields are set to `null`, the corresponding garbage collection
task will not run even if `garbage_collector.sweep_interval.<datatype>` is configured.
```
specificity

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+:::
+
+
```
quinntaylormitchell:
```suggestion

---

```
page line for navigability

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+## Internal
```
quinntaylormitchell:
```suggestion
## Internal

This section documents some of CLP’s internal behavior for retention and garbage collection.

```
there should be an intro here; not good form to have two levels of headings one after the other

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,161 @@
...
+Restarting the query scheduler will mark such jobs as failed and allow garbage collection to resume.
+:::
+
+### Fault tolerance
+The garbage collector can resume execution from where it left off if a previous run fails. 
+This design ensures that CLP does not fall into an inconsistent state due to partial deletions.
+
+If the CLP package stops unexpectedly (for example, due to a host machine shutdown) while a garbage
+collection task is running, simply restart the package and the garbage collector will continue 
+from the point of failure.
```
quinntaylormitchell:
```suggestion
If the CLP package stops unexpectedly while a garbage collection task is running (for example, due
to a host machine shutdown), simply restart the package and the garbage collector will continue from
the point of failure.
```
realized that it reads better to have the parenthetical right before the comma instead of in the middle of the clause

quinntaylormitchell (changes requested):
UTC stuff will work well. Let me know if any of my comments are unclear.

docs/src/user-guide/guides-retention.md
haiqi96: True

docs/src/user-guide/guides-retention.md
haiqi96: I feel this sounds like as long as one of the .retention_period is set to null, the garbage collection won't run? or maybe i misunderstood?

Edited: I updated it based on my understanding.

docs/src/user-guide/guides-retention.md
haiqi96: personally I would avoid EDT because it caused some confusions in the past. maybe I will use EST and offset everything by 5 instead,

docs/src/user-guide/guides-retention.md
haiqi96: @kirkrodrigues Any comment on the "for simplicity..." part?

kirkrodrigues: I think we should keep the parenthetical since some folks may get confused. But I think we should move it after the first timestamp.

docs/src/user-guide/guides-retention.md
haiqi96: I moved the first point to the archive timestamp with some repharsing, but it also looks very wordy. @kirkrodrigues do you have any suggestions?

docs/src/user-guide/guides-retention.md
haiqi96: since we define "archive timestamp" to be of latest message's one, I feel we don't need to explicity mention about the "most recent" again.

docs/src/user-guide/guides-retention.md
haiqi96: I incorporated part of your change. Given that this is high-level description, I feel that we don't need to mention the exact key in the config.

In my doc, i have been using "data" to refer "either archive or search result". I don't use the latter because later on when we add stream retention it would not require too much change.

Wonder is there any better name? Another option I can think of is "target", and maybe we can define it in the second section? @kirkrodrigues

docs/src/user-guide/guides-retention.md
haiqi96: I don't entirely agree with all proposed changes, given that I might not work on it again this weekend, leave for @kirkrodrigues to decide.

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,167 @@
...
+### Handling data race conditions
+CLP's retention system is designed to avoid data race conditions that may arise from the deletion of
+archives or search results that may still be in use by active jobs. CLP employs the following
+mechanisms to avoid these conditions:
+
+- If any query job is running, CLP conservatively calculates a **safe expiry timestamp** based on 
+  the earliest active search job. This ensures no archive that could be searched is deleted.
+
+- CLP will **not** search an archive once it is considered expired, even if it has not yet been
+  deleted by the garbage collector.
```
junhaoliao: Shall we add a "warning" for this?
Alternatively, can we mark the job with a non-successful status so that users can be aware that some archives are skipped? Otherwise, it would appear odd that we list a certain number of events on the WebUI ingestion page but none of them show up in the search results. (Or maybe we should also exclude the stats of those archives on the ingestion page?)

haiqi96: I feel we need to discuss the actual behavior so that we are on the same page, but I do think that "excluding the stats of those archives on the ingestion page" sounds like a valid suggestion.

kirkrodrigues: @davemarco @hoophalab How hard would it be to add this today?

haiqi96: From a high-level, here is what we need to implement
1. Webui should take in `archive_output.retention_period` in as an optional int. (can be null)
2. When fetching ingested log info, if the retention_period is not null
    - get the current UTC timestamp
    - reverse calculate the `lower_bound` = UTC_ts -  retention_period * 60 (coverting mins to secs)
    - Append a filter to the SQL query  `select <columns> from <existing conditions> and end_timestamp >= {archive_end_ts_lower_bound} OR end_timestamp = 0`

Similar code is implemented in [here](https://github.com/y-scope/clp/commit/70adb90cbc561b7480ea512d737d630449459efd#diff-98ed70652c7ba7a24bef1b43da9b1889415c74ac343b9fa2ee23a49070877aee) under the query_scheduler.py

davemarco: is it possible to add an expired bool to the metadata tables, then we can just query for expired?

haiqi96: maybe in the future release. I don't think it's at the level where we can finish planning, implementing and testing in one day though.

davemarco: I think we can get code in today, but someone will need to review and test. I guess its up to @kirk whether to put in this release, or fix after

kirkrodrigues: Can you put up the PR and then we can evaluate? Either way, we'll need this feature unless we switch to some new API server.

davemarco: @haiqi96 @kirkrodrigues it is actually slightly more complicated for the files table since multiple files may be part of the same archive and have different end timestamps. Is there an easy around this?

haiqi96: hmm, what if you do a join statement? 
like 
select * from File
join Archive_table on File.archive_id = Archive_table.id
Where Archive_table.end_ts > lower_bound?

Alternatively, you can execute two queries:
1. select archive_ids from Archive_table where Archive_table.end_ts > lower_bound
2. Select files from Files_table where Files_table.archive_id in archive_ids 

The first option could be easier to write though.

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,211 @@
...
+
+* [How retention works in CLP](#how-retention-works)
+* [How to configure retention](#retention-settings)
+* [Additional concerns worth noting](#additional-concerns)
+
+## How retention works
+
+To support retention periods, CLP (specifically, the garbage collector component) periodically scans
+for and deletes expired data (archives or search results). To understand the high-level algorithm,
+first consider the following definitions:
```
quinntaylormitchell:
```suggestion
To support retention periods, CLP's garbage collector component periodically scans for and deletes
expired data (archives or search results). To understand the high-level algorithm, first consider
the following definitions:
```

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,211 @@
...
+criteria shown in [Figure 1](#figure-1):
+
+(figure-1)=
+:::{card}
+
+$$data\_timestamp < current\_time - retention\_period$$
+
++++
+**Figure 1**: The criteria for determining whether a piece of data has expired and should be
+deleted (i.e., the data's expiry criteria).
```
quinntaylormitchell:
```suggestion
**Figure 1**: The criteria for determining whether a piece of data has expired and should be
deleted.
```

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,211 @@
...
+* If any query job is running, CLP conservatively calculates a **safe expiry timestamp** based on 
+  the earliest active search job. This ensures no archive which may be searched by the active job is
+  deleted.
+
+* CLP will **not** search an archive once it is considered expired, even if it has not yet been
+  deleted by the garbage collector.
+
+:::{warning}
+A hanging search job will prevent CLP from deleting expired archives. Restarting the query scheduler
+will mark such jobs as failed and allow garbage collection to resume.
```
haiqi96: If we are going to merge the scheduler change?
```suggestion
will mark such jobs as killed and allow garbage collection to resume.
```

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,211 @@
...
+particularly old logs that have already expired according to the expiry criteria, they will be
+deleted the next time the garbage collector runs.
+:::
+
+#### Handling log events with non-UTC timestamps
+
+If your log events use timestamps that **aren't** in the UTC time zone, you will need to adjust the
+configured retention period to ensure expired archives are deleted at the correct time. This is
+because CLP currently doesn't support parsing time zone information from timestamps that have it,
+whereas the garbage collector runs based on the UTC time zone.
```
quinntaylormitchell:
```suggestion
because CLP currently doesn't support parsing time zone information, and the garbage collector runs
based on the UTC time zone.
```

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,211 @@
...
+| $current\_time$     | The time at which the garbage collector is performing a check.                                                                                                             |
+| $data\_timestamp$   | The end of the time range for the data being evaluated for expiration (e.g., for an archive, this is the timestamp of the most recent log event contained in the archive). |
+
+When the garbage collector wakes up, it will scan for and delete any data that fits the expiry
+criteria shown in [Figure 1](#figure-1):
+
+(figure-1)=
+:::{card}
+
+$$data\_timestamp < current\_time - retention\_period$$
```
quinntaylormitchell: See high-level comment.

docs/conf/conf.py
```diff
@@ -24,6 +24,7 @@
 myst_enable_extensions = [
     "attrs_block",
     "colon_fence",
+    "dollarmath",
 ]
```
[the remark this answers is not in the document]
quinntaylormitchell: @kirkrodrigues I think it's fine the way it is, wdyt

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,211 @@
...
+For example, to configure a search result retention period of 1 day (1,440 minutes), use:
+
+```yaml
+results_cache:
+  # ... Other results_cache settings...
+
+  # Retention period for search results, in minutes. 
+  # Set to null to disable automatic deletion.
+  retention_period: 1440
+```
```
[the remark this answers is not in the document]
quinntaylormitchell: @kirkrodrigues add this if you think it's necessary, but I think it's fine the way it is

docs/conf/conf.py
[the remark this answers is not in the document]
kirkrodrigues: Yeah, it renders fine, so I think the rabbit is wrong.

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,211 @@
...
+
+:::{note}
+This is not the timestamp at which your logs were compressed. Therefore, if you compress
+particularly old logs that have already expired according to the expiry criteria, they will be
+deleted the next time the garbage collector runs.
+:::
+
+#### Handling log events with non-UTC timestamps
+
+If your log events use timestamps that **aren't** in the UTC time zone, you will need to adjust the
```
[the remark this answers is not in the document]
quinntaylormitchell: not needed

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,211 @@
...
+because CLP currently doesn't support parsing time zone information, and the garbage collector runs
+based on the UTC time zone.
+
+For example, let's say:
+
+* your log events use timestamps in the AWST timezone (UTC+8);
+* you set a retention period of 1 hour;
+* you have an archive with $data\_timestamp = 08:00$ AWST; and
+* the garbage collector runs at $current\_time = 09:01$ AWST.
+
```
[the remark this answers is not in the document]
quinntaylormitchell: @kirkrodrigues up to you, I think it's fine how it is. the australian people need representation

quinntaylormitchell (approved):
LGTM @kirkrodrigues!

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,227 @@
...
+  # Retention period for search results, in minutes. 
+  # Set to null to disable automatic deletion.
+  retention_period: 1440
+```
+
+By default, `results_cache.retention_period` is `60`, which means that search results will be
+retained for 60 minutes (1 hour).
+
+:::{note}
+When a user runs consecutive queries in the webui without refreshing the page, the results of one
```
haiqi96: how about? or maybe something like "current query"
```suggestion
When a user runs consecutive queries in the webui without refreshing the page, the results of the previous
```

docs/src/user-guide/guides-multi-node.md
```diff
@@ -71,6 +72,8 @@ Running additional workers increases the parallelism of compression and search/a
     4. Set `archive_output.directory` to a directory on the distributed filesystem.
        * Ideally, the directory should be empty or should not yet exist (CLP will create it) since
          CLP will write several files and directories directly to the given directory.
+    5. (Optional) Configure retention periods for archives and search results. See
+       [retention control](guides-retention) for details.
```
kirkrodrigues: This feels a bit out of scope of this doc, right?

docs/src/user-guide/index.md
```diff
@@ -63,6 +63,7 @@ guides-overview
 guides-using-object-storage/index
 guides-multi-node
 guides-using-presto
+guides-retention
```
kirkrodrigues: Can we move this above `guides-multi-node`? Same for the card in the overview.

docs/src/user-guide/guides-overview.md
```diff
@@ -13,10 +13,10 @@ Using CLP to ingest logs from object storage and store archives on object storag
 :::
 
 :::{grid-item-card}
-:link: guides-using-presto
-Using Presto with CLP
+:link: guides-retention
+Retention control
```
kirkrodrigues:
```suggestion
Configuring retention periods
```

docs/src/user-guide/guides-overview.md
```diff
@@ -13,10 +13,10 @@ Using CLP to ingest logs from object storage and store archives on object storag
...
 ^^^
-How to use Presto to query compressed logs in CLP.
+How to configure retention control for CLP.
```
kirkrodrigues:
```suggestion
How to configure retention periods for CLP archives and search results.
```

docs/src/user-guide/guides-retention.md
```diff
@@ -0,0 +1,227 @@
+# Setting up data retention periods
```
kirkrodrigues:
```suggestion
# Configuring retention periods
```

kirkrodrigues (approved):
Edited the PR title directly.

PR #1191, opened 2025-08-13
fix(webui): Log Presto cancel route output as object instead of string.

hoophalab:
# Description

Fastify logging accepts the first argument as an object not  a string.
As the title indicates, we're now logging the Presto cancel route output as an object.

This bug was discovered by @junhaoliao and can be reproduced by updating the webui server's packages to their latest versions. It didn't show previously because Pino could silently handling `obj` an unknown argument in the old version.

The old interface definition:
```
    interface LogFn {
    ...
        (obj: unknown, msg?: string, ...args: any[]): void;
    }
```

# Checklist

  breaking change.

# Validation performed

1. Search cancellation works exactly as before.

2. The package successfully builds with the updated dependencies.

## Summary by CodeRabbit

* **New Features**
  * No user-facing changes in this release.

* **Refactor**
  * Standardized structured logging for search cancellation events to improve consistency and traceability; no change to behaviour or API.

* **Chores**
  * Improved observability by switching to object-based log payloads for cancellation logs, aiding diagnostics and log aggregation without altering functionality.


hoophalab:
A bug of cancel button but not related to this PR: If the user click "Run" and then immediately "Cancel," the query won't actually stop. This might be because Presto's cancel route returns success even if queries aren't submitted yet.
This bug will probably be automatically fixed after we handle query status correctly in client in a future PR, but we may need to revisit the resetful api design especially if the user want to use api directly.
@coderabbitai can you create an issue on this?

junhaoliao (approved):
for the title, how about
```
fix(webui): Log Presto cancel route output as object instead of string.
```

PR #1198, opened 2025-08-13
feat(package): Add settings to configure webui Presto client (resolves #1259, resolves #1260).

### LINKED ISSUE #1259: Update ClpQueryEngine setting in start_clp.py server_settings_json_updates
## Description

The ClpQueryEngine setting was updated from "native" to "clp" in PR #1168 for both client and server settings.json files, but the corresponding configuration in `server_settings_json_updates` within `components/clp-package-utils/clp_package_utils/scripts/start_clp.py` also needs to be updated for consistency.

## Task

Update the `server_settings_json_updates` configuration in `components/clp-package-utils/clp_package_utils/scripts/start_clp.py` to change `ClpQueryEngine` from "native" to "clp".

## Context

- **PR**: #1168
- **Comment**: https://github.com/y-scope/clp/pull/1168#discussion_r2296622518
- **Requested by**: @junhaoliao

This ensures consistency across all configuration files in the codebase.

### LINKED ISSUE #1260: Configure Presto settings in start_clp.py server_settings_json_updates
## Description

The recent addition of Presto SQL query support in PR #1127 introduced new configuration settings in `components/webui/server/settings.json`:

- `ClpQueryEngine`: "native"
- `PrestoHost`: "localhost"  
- `PrestoPort`: 8889

These settings need to be properly configured in the `server_settings_json_updates` dictionary in `components/clp-package-utils/clp_package_utils/scripts/start_clp.py` to ensure consistency between the webui server configuration and the CLP startup script.

## Current Status

The settings are currently hardcoded in `settings.json` but are not dynamically configured through the startup script's configuration management system.

## Expected Behavior

The Presto-related settings should be configurable through the CLP configuration system and properly injected into the webui server settings during startup, similar to how other settings like database and MongoDB configurations are handled.

## References

- PR: https://github.com/y-scope/clp/pull/1127
- Comment: https://github.com/y-scope/clp/pull/1127#discussion_r2296622851
- Requested by: @junhaoliao

davemarco:
# Description

Adds host/port settings to start a presto client in webui.

Also added catalog,schema required for connector, but should not be configurable by users. Devs can change by modifying env variables if they want to use ui with a different connector.

Presto has fancier authentication methods, but will leave for a future PR.

Also added a user since neccesary i believe when running the webui in docker container. 

# Checklist

  breaking change.

# Validation performed

Tested package sets variables, and client starts succesfully

## Summary by CodeRabbit

- New Features
  - Support for a Presto-backed query engine; Web UI shows the selected engine and, when Presto is chosen, exposes Presto host/port plus catalog, schema and user options.

- Configuration
  - Added a Presto section to the sample config.
  - Added PRESTO_CATALOG, PRESTO_SCHEMA and USER environment variables with defaults.
  - Server and client settings now expose the CLP query engine selection; Presto host/port are set only when Presto is selected.

- Validation
  - Presto configuration is required when the Presto engine is selected.

### DIFF
--- a/components/clp-package-utils/clp_package_utils/scripts/compress.py
+++ b/components/clp-package-utils/clp_package_utils/scripts/compress.py
@@ -202,6 +202,12 @@ def main(argv):
         except Exception as e:
             logger.error(e)
             return -1
+
+        if parsed_args.timestamp_key is None:
+            logger.warning(
+                "`--timestamp-key` not specified. Events will not have assigned timestamps and can "
+                "only be searched from the command line without a timestamp filter."
+            )
     elif dataset is not None:
         logger.error(f"Dataset selection is not supported for storage engine: {storage_engine}.")
         return -1
--- a/components/core/src/clp_s/ArchiveReader.hpp
+++ b/components/core/src/clp_s/ArchiveReader.hpp
@@ -15,7 +15,6 @@
 #include "SchemaReader.hpp"
 #include "search/Projection.hpp"
 #include "TimestampDictionaryReader.hpp"
-#include "Utils.hpp"
 
 namespace clp_s {
 class ArchiveReader {
--- a/components/core/src/clp_s/DictionaryEntry.cpp
+++ b/components/core/src/clp_s/DictionaryEntry.cpp
@@ -12,7 +12,6 @@
 #include "../clp/ir/parsing.hpp"
 #include "../clp/ir/types.hpp"
 #include "../clp/type_utils.hpp"
-#include "Utils.hpp"
 
 using clp::EncodedVariableInterpreter;
 using clp::enum_to_underlying_type;
--- a/components/core/src/clp_s/DictionaryReader.hpp
+++ b/components/core/src/clp_s/DictionaryReader.hpp
@@ -14,7 +14,6 @@
 
 #include "ArchiveReaderAdaptor.hpp"
 #include "DictionaryEntry.hpp"
-#include "Utils.hpp"
 
 namespace clp_s {
 template <typename DictionaryIdType, typename EntryType>
--- a/components/core/src/clp_s/JsonParser.hpp
+++ b/components/core/src/clp_s/JsonParser.hpp
@@ -28,7 +28,6 @@
 #include "SchemaTree.hpp"
 #include "SchemaWriter.hpp"
 #include "TimestampDictionaryWriter.hpp"
-#include "Utils.hpp"
 #include "ZstdCompressor.hpp"
 
 namespace clp_s {
--- a/components/core/src/clp_s/Schema.hpp
+++ b/components/core/src/clp_s/Schema.hpp
@@ -8,7 +8,6 @@
 
 #include "SchemaTree.hpp"
 #include "TraceableException.hpp"
-#include "Utils.hpp"
 
 namespace clp_s {
 /**
--- a/components/core/src/clp_s/TimestampDictionaryWriter.cpp
+++ b/components/core/src/clp_s/TimestampDictionaryWriter.cpp
@@ -4,8 +4,6 @@
 #include <sstream>
 #include <string_view>
 
-#include "Utils.hpp"
-
 namespace clp_s {
 void TimestampDictionaryWriter::write_timestamp_entries(
         std::map<std::string, TimestampEntry> const& ranges,
--- a/components/core/src/clp_s/Utils.cpp
+++ b/components/core/src/clp_s/Utils.cpp
@@ -163,56 +163,6 @@ bool UriUtils::get_last_uri_component(std::string_view const uri, std::string& n
     return true;
 }
 
-bool StringUtils::get_bounds_of_next_var(string const& msg, size_t& begin_pos, size_t& end_pos) {
-    auto const msg_length = msg.length();
-    if (end_pos >= msg_length) {
-        return false;
-    }
-
-    while (true) {
-        begin_pos = end_pos;
-        // Find next non-delimiter
-        for (; begin_pos < msg_length; ++begin_pos) {
-            if (false == is_delim(msg[begin_pos])) {
-                break;
-            }
-        }
-        if (msg_length == begin_pos) {
-            // Early exit for performance
-            return false;
-        }
-
-        bool contains_decimal_digit = false;
-        bool contains_alphabet = false;
-
-        // Find next delimiter
-        end_pos = begin_pos;
-        for (; end_pos < msg_length; ++end_pos) {
-            char c = msg[end_pos];
-            if (clp::string_utils::is_decimal_digit(c)) {
-                contains_decimal_digit = true;
-            } else if (clp::string_utils::is_alphabet(c)) {
-                contains_alphabet = true;
-            } else if (is_delim(c)) {
-                break;
-            }
-        }
-
-        // Treat token as variable if:
-        // - it contains a decimal digit, or
-        // - it's directly preceded by an equals sign and contains an alphabet, or
-        // - it could be a multi-digit hex value
-        if (contains_decimal_digit
-            || (begin_pos > 0 && '=' == msg[begin_pos - 1] && contains_alphabet)
-            || could_be_multi_digit_hex_value(msg, begin_pos, end_pos))
-        {
-            break;
-        }
-    }
-
-    return (msg_length != begin_pos);
-}
-
 void StringUtils::escape_json_string(std::string& destination, std::string_view const source) {
     // Escaping is implemented using this `append_unescaped_slice` approach to offer a fast path
     // when strings are mostly or entirely valid escaped JSON. Benchmarking shows that this offers
--- a/components/core/src/clp_s/Utils.hpp
+++ b/components/core/src/clp_s/Utils.hpp
@@ -66,54 +66,6 @@ public:
 class StringUtils {
 public:
     /**
-     * Checks if character is a hexadecimal (base-16) digit
-     * @param c
-     * @return true if c is a hexadecimal digit, false otherwise
-     */
-    static inline bool is_delim(char c) {
-        return !(
-                '+' == c || ('-' <= c && c <= '9') || ('A' <= c && c <= 'Z') || '\\' == c
-                || '_' == c || ('a' <= c && c <= 'z')
-        );
-    }
-
-    /**
-     * Checks if the string could be a hexadecimal value
-     * @param str
-     * @param begin_pos
-     * @param end_pos
-     * @return true if str could be a hexadecimal value, false otherwise
-     */
-    static inline bool
-    could_be_multi_digit_hex_value(std::string const& str, size_t begin_pos, size_t end_pos) {
-        if (end_pos - begin_pos < 2) {
-            return false;
-        }
-
-        for (size_t i = begin_pos; i < end_pos; ++i) {
-            auto c = str[i];
-            if (false
-                == (('a' <= c && c <= 'f') || ('A' <= c && c <= 'F') || ('0' <= c && c <= '9')))
-            {
-                return false;
-            }
-        }
-
-        return true;
-    }
-
-    /**
-     * Returns bounds of next variable in given string
-     * A variable is a token (word between two delimiters) that contains numbers or is directly
-     * preceded by an equals sign
-     * @param msg
-     * @param begin_pos Begin position of last variable, changes to begin position of next variable
-     * @param end_pos End position of last variable, changes to end position of next variable
-     * @return true if a variable was found, false otherwise
-     */
-    static bool get_bounds_of_next_var(std::string const& msg, size_t& begin_pos, size_t& end_pos);
-
-    /**
      * Escapes a string according to JSON string escaping rules and appends the escaped string to
      * a buffer. The input string can be either ascii or UTF-8.
      *
--- a/components/core/src/clp_s/clp-s.cpp
+++ b/components/core/src/clp_s/clp-s.cpp
@@ -38,7 +38,6 @@
 #include "search/Projection.hpp"
 #include "search/SchemaMatch.hpp"
 #include "TimestampPattern.hpp"
-#include "Utils.hpp"
 
 using namespace clp_s::search;
 using clp_s::cArchiveFormatDevelopmentVersionFlag;
@@ -47,7 +46,6 @@ using clp_s::cEpochTimeMin;
 using clp_s::CommandLineArguments;
 using clp_s::KvIrSearchError;
 using clp_s::KvIrSearchErrorEnum;
-using clp_s::StringUtils;
 
 namespace {
 /**
--- a/components/core/src/clp_s/search/QueryRunner.cpp
+++ b/components/core/src/clp_s/search/QueryRunner.cpp
@@ -11,6 +11,7 @@
 #include "../../clp/Query.hpp"
 #include "../../clp/type_utils.hpp"
 #include "../SchemaTree.hpp"
+#include "../Utils.hpp"
 #include "ast/AndExpr.hpp"
 #include "ast/ColumnDescriptor.hpp"
 #include "ast/Expression.hpp"


davemarco:
@kirkrodrigues  do u want to review this or someone else?

components/package-template/src/etc/clp-config.yml
```diff
@@ -67,6 +67,10 @@
 #  port: 4000
 #  results_metadata_collection_name: "results-metadata"
 #
+#presto:
```
haiqi96: should we consider making the order to be consistent with the order we declare them in CLPConfig? (line 731)

components/webui/server/src/plugins/app/Presto.ts
```diff
@@ -7,6 +7,8 @@ import {
 import settings from "../../../settings.json" with {type: "json"};
 
 
+const CLP_PRESTO_USER = "clp-webui";
```
haiqi96: I am not familiar with the context, wonder if this value is only limited to webui.
If something else could use the value in the future, shall we consider moving it into clp-config.yml, and passin via settings?

components/package-template/src/etc/clp-config.yml
davemarco: sure but should garbage collector be higher up then? like above archive output settings?

components/webui/server/src/plugins/app/Presto.ts
davemarco: hmm i dont think anyone would really benefit from changing this. The username does nothing from a functional perspective, it is just metadata. Normally the library we are using defaults to using linux username automatically. This hardcode is basically a quick fix since when running in docker there is no username, and library we are using requires a user name. Maybe a better solution is to check process.env.USER, and if undefined, then set to hardcoded "clp-webui"?

components/package-template/src/etc/clp-config.yml
haiqi96: > sure but should garbage collector be higher up then? like above archive output settings?

True, the order isn't that simple.
Do we start presto as a component, or will it be something user configure themselves and only need to let CLP know the port?

components/webui/server/src/plugins/app/Presto.ts
haiqi96: Sure, checking process.env.USER sounds reasonable.

components/package-template/src/etc/clp-config.yml
davemarco: per kirk, plan is for user to start their own presto cluster, then the package/webui connects to it

haiqi96: How about this:
1. Let's make the declaration of Presto to be consistent in the clp-config.yml and clp_config.py -- since now class Presto is defined after class garbage collector in clp_config.py, let's keep the same order and move Presto lower in the clp-config.yml
2. for `presto: Presto = Presto()`. I would say let's move it to the block starting at [line734](https://github.com/y-scope/clp/blob/1fc6b20e1da43e3ebd8591b64a2c7721467e3bc8/components/clp-py-utils/clp_py_utils/clp_config.py#L734C5-L735C49), the reasoning is that the target in the first block are launchable components.

It looks like that there might be some inconsistent decisions made on where to put what, I would say let's leave those for future issues and PRs.

components/webui/server/src/plugins/app/Presto.ts
```diff
@@ -38,7 +39,7 @@ export default fp(
         const clientOptions: ClientOptions = {
             host: settings.PrestoHost,
             port: settings.PrestoPort,
-            user: CLP_PRESTO_USER,
+            user: process.env.USER || CLP_PRESTO_USER_FALLBACK,
```
haiqi96: @junhaoliao Defer to Junhao for review.

haiqi96 (approved):
The changes in python and yml looks good to me.
Leave one webui syntax for Junhao to review, since we need Junhao's approval anyway.

junhaoliao:
Let's merge main into this branch and add #1259 & #1260 to the PR title

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -715,6 +730,7 @@
     garbage_collector: GarbageCollector = GarbageCollector()
     credentials_file_path: pathlib.Path = CLP_DEFAULT_CREDENTIALS_FILE_PATH
 
+    presto: Presto = Presto()
```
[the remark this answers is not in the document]
junhaoliao: i agree with the Rabbit

components/webui/server/src/plugins/app/Presto.ts
```diff
@@ -36,6 +39,7 @@ export default fp(
         const clientOptions: ClientOptions = {
             host: settings.PrestoHost,
             port: settings.PrestoPort,
+            user: process.env.USER || CLP_PRESTO_USER_FALLBACK,
```
junhaoliao: do we want to use the fastify env plugin instead of directly reading it from `process.env`?

junhaoliao: That way, we don't have to configure the fallback with `||`

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -912,6 +912,9 @@ def start_webui(
         "ClientDir": str(container_webui_dir / "client"),
         "LogViewerDir": str(container_webui_dir / "yscope-log-viewer"),
         "StreamTargetUncompressedSize": container_clp_config.stream_output.target_uncompressed_size,
```
junhaoliao, line 890: off topic: @haiqi96 do you know why we read some of the settings from container_clp_config instead of clp_config?

junhaoliao (changes requested):

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
haiqi96: iirc, we once mounted host directories to different paths in the container (for example, the package directory on the host is mounted to `/opt/clp` in the container. The paths in clp_config are host path, whereas paths in container_clp_config are container paths.

That said, we implemented a change a while ago to let most (but not all) paths in the container to be the same as the host. So technically, we are now duplicating the entire config to just account for any path that's still different between the host and the container.

One possibility to improve this is to create a smaller class that only contains the different mounted path, and stick to the original clp_config (with everything loaded, such as password) for all other purposes. It's likely the CLPDockerMounts can already achieve something similar.

components/clp-py-utils/clp_py_utils/clp_config.py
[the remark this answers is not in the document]
davemarco: @junhaoliao @coderabbitai i dont know if this will work since it will still validate fields? Also I think better to keep validated as these inputs will be still be passed to webui, and unvalidated inputs may still cause problems if they are not used

components/webui/server/src/plugins/app/Presto.ts
davemarco: Sure, i made this change

components/webui/server/src/plugins/app/Presto.ts
```diff
@@ -34,8 +34,11 @@ export default fp(
         }
 
         const clientOptions: ClientOptions = {
+            catalog: fastify.config.PRESTO_CATALOG,
             host: settings.PrestoHost,
             port: settings.PrestoPort,
+            schema: fastify.config.PRESTO_SCHEMA,
+            user: fastify.config.USER,
         };
```
davemarco, line 43: settings are typicially set by package which we dont need here

components/webui/server/src/plugins/external/env.ts
```diff
@@ -37,6 +39,12 @@ const schema = {
             default: "localhost",
         },
 
+        // System
+        USER: {
+            type: "string",
+            default: "clp-webui",
+        },
+
```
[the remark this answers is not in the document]
davemarco: this may not be neccesary, i need to retest

davemarco:
@junhaoliao , I discovered we need more options after working with actual presto connector so added a couple things since your last review

Added catalog,schema required for connector, but should not be configurable by users. Devs can change by modifying env variables if they want to use ui with a different connector.

components/clp-py-utils/clp_py_utils/clp_config.py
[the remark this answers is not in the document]
junhaoliao: ah sorry for missing this - can we make `presto` an optional first? then if the user configures an object (which suggests they intend to use Presto), but forgot to update the query_engine, the `root_validator` can help catching that mistake.

e.g.,
```
presto: Optional[Presto] = None
```

since this is at the interface level, let's have @kirkrodrigues / @haiqi96 confirm before making the change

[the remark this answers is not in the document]
haiqi96: Hmm. this is actually very similar to `aws_config_directory`, where one optional field is only allowed when other field(s) satisfy a certain condition. So I think we should refer to 

Looking at `aws_config_directory`, we leave the validation outside of the sub-components by implementing a `validate_aws_config_dir()` under root config. I feel it makes more sense this way so that individual pydantic class doesn't need to know any global rule/constraint.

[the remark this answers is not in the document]
davemarco: hmm but then the defaults cannot be used? the localhost/port?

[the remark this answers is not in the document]
haiqi96: If necessary we can argue that, since presto is something user configures externally and CLP has no control over it, it's safer to not assume any default but let user explicitly configure.

If the default port is not available (i.e. no presto is running), what would be the behavior of starting a package? If the package will silently starts without a failure, it might be better to force user explicitly specify the port and host.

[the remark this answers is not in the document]
davemarco: Currently it fail silently until you run a query. 

I guess we can do that, but users may need to configure port twice. Once when they launch presto using our instructions (we need to add this - I set up an issue already) and once here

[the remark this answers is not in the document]
haiqi96: if they have to modify the clp-config.yml and update the query_engine to `presto` anyway, I feel it's not the end of the world to ask them also explicitly specify the presto port and host.

@junhaoliao how do you think?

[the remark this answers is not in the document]
davemarco: @haiqi96 @junhaoliao , i made change u both requested

davemarco:
@junhaoliao requesting rereview

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -645,6 +645,14 @@ def is_retention_period_configured(clp_config: CLPConfig) -> bool:
     return False
 
 
+def validate_presto_config(clp_config: CLPConfig) -> None:
+    if clp_config.package.query_engine == QueryEngine.PRESTO:
+        if clp_config.presto is None:
+            raise ValueError(
+                f"Presto config must be set when query_engine is `{clp_query_engine}`"
+            )
```
haiqi96: Nit: I got similar comments many times from Zhihao and Kirk, so let's go with less indent.
```suggestion
    if clp_config.package.query_engine == QueryEngine.PRESTO and clp_config.presto is None:
        raise ValueError(
            f"Presto config must be set when query_engine is `{clp_query_engine}`"
        )
```

components/package-template/src/etc/clp-config.yml
```diff
@@ -115,6 +115,11 @@
 #    archive: 60
 #    search_result: 30
 #
+## Presto client config
+#presto:
+#  host: "localhost"
+#  port: 8889
+#
```
haiqi96: iirc, the assumption is that we don't `show` non-default field. so we either don't hide the presto section, or explicitly assign it to null (if it works)
```suggestion
## Presto client config
#presto: null
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -777,6 +777,21 @@ def validate_logging_level(cls, field):
         return field
 
 
+class Presto(BaseModel):
+    host: str = "localhost"
+    port: int = 8889
```
haiqi96: Is there a way that we can use both default? I suppose you can't just have an "empty" presto like
```
presto:
# nothing

#next_component
archive:
   bla
```

@junhaoliao I feel that since we ask user to explicilty specify presto, it's better to not provide any default values. how do you think?

davemarco: may work with `{}`

components/package-template/src/etc/clp-config.yml
davemarco: how do u feel about changing comment `Presto client config. Set host and port of Presto cluster when using presto query engine.`

haiqi96: hmm, I would say what you suggest sounds reasonable I can't make a decision because we never had anything external like presto in the past. Can you discuss with Kirk to see which he prefers?

components/clp-py-utils/clp_py_utils/clp_config.py
haiqi96: I think we can go with either:

- remove default from clp_config.py:Presto
- assign Presto=null in clp_config.yml.

or

- Leave default in clp_config.py:Presto
- Leave commented prestor: port=<> host=<> in the clp_config.yml. and add comments suggested by Marco in https://github.com/y-scope/clp/pull/1198#discussion_r2301204405

While 1 seems to be consistent with what we have for `aws_config_directory`, 2 is easier for user. I don't have a strong preference between the 2, but if Marco prefers the 2nd approach, I am fine with it.

davemarco: okay i went with appraoch 1. eventually we will need docs for user for this. Not in this PR

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -911,6 +914,14 @@ def start_webui(
...
         server_settings_json_updates["StreamFilesS3Profile"] = None
 
+    query_engine = clp_config.package.query_engine
+    if QueryEngine.PRESTO == query_engine:
+        server_settings_json_updates["PrestoHost"] = clp_config.presto.host
+        server_settings_json_updates["PrestoPort"] = clp_config.presto.port
+    else:
+        server_settings_json_updates["PrestoHost"] = None
+        server_settings_json_updates["PrestoPort"] = None
+
```
davemarco, line 924: should not be none since already validated

components/clp-py-utils/clp_py_utils/clp_config.py
davemarco: went with 1

davemarco:
requesting rereview from @junhaoliao  as well

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -777,6 +777,18 @@ def validate_logging_level(cls, field):
...
+class Presto(BaseModel):
+    @validator("host")
+    def validate_host(cls, field):
+        _validate_host(cls, field)
+        return field
+
+    @validator("port")
+    def validate_port(cls, field):
+        _validate_port(cls, field)
+        return field
```
haiqi96: Similar to S3Credentials, we should still declare the field even if they don't have a default
```suggestion
class Presto(BaseModel):
    host: str
    port: int

    @validator("host")
    def validate_host(cls, field):
        _validate_host(cls, field)
        return field

    @validator("port")
    def validate_port(cls, field):
        _validate_port(cls, field)
        return field
```

haiqi96 (approved):
Looks good to me.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -803,6 +818,7 @@ class CLPConfig(BaseModel):
     garbage_collector: GarbageCollector = GarbageCollector()
     credentials_file_path: pathlib.Path = CLP_DEFAULT_CREDENTIALS_FILE_PATH
 
+    presto: Optional[Presto] = None
```
kirkrodrigues, line 828:
```suggestion
    presto: Optional[Presto] = None

```

davemarco:
@haiqi96 in meeting kirk asked to me use root validator in clp config instead. Rerequesting review.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -971,6 +971,17 @@ def dump_to_primitive_dict(self):
 
         return d
 
+    @root_validator(pre=True)
+    def validate_presto_config(cls, values):
+        package = values.get("package")
+        presto = values.get("presto")
+        query_engine = None
```
haiqi96:
```suggestion
        query_engine: Optional[QueryEngine] = None
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -975,7 +975,9 @@ def dump_to_primitive_dict(self):
     def validate_presto_config(cls, values):
         package = values.get("package")
         presto = values.get("presto")
-        query_engine = None
+        query_engine: Optional[QueryEngine] = None
+        # Root validator with `pre=True` is called before other validators, so package has not
+        # been validated as dict yet. Pydantic will throw error later if package is not a dict.
```
haiqi96:
```suggestion
        # Root validator with `pre=True` is called before other validators, so package has not
        # been validated yet. Pydantic will throw error later if package is not a valid Package object.
```
And sorry for one last comment:

what if we do 
```
if not isinstance(package, Package):
    return
query_engine = package.get("query_engine")
```

If package is an instance of Package, it must have query_engine key

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -955,6 +971,20 @@ def dump_to_primitive_dict(self):
 
         return d
 
+    @root_validator(pre=True)
```
kirkrodrigues, line 984:
```suggestion
    # We set `pre=True` so that we print errors about a mismatch between the query engine and presto
    # config before errors about the presto config itself.
    @root_validator(pre=True)
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -955,6 +971,20 @@ def dump_to_primitive_dict(self):
...
+        package = values.get("package")
+        presto = values.get("presto")
+        query_engine: Optional[QueryEngine] = None
+        # Root validator with `pre=True` is called before other validators, so package has not
+        # been validated yet. Pydantic will throw error later if package is not a valid Package object.
+        if not isinstance(package, Package):
+            return values
+        query_engine = package.get("query_engine")
+        if query_engine == QueryEngine.PRESTO and presto is None:
+            raise ValueError(f"Presto config must be set when query_engine is `{query_engine}`")
```
```suggestion
        if not isinstance(package, Package):
            # Skip validation since `package` is not a valid `Package` (Pydantic will validate it
            # later and throw an error).
            return values

        query_engine = package.get("query_engine")
        presto = values.get("presto")
        if query_engine == QueryEngine.PRESTO and presto is None:
            raise ValueError(f"`presto` config must be non-null when query_engine is"
                             f" `{query_engine}`")
```

kirkrodrigues (commented):
For the PR title, how about:

```
feat(package): Add settings to configure webui Presto client (resolves #1259, resolves #1260).
```

davemarco:
@kirkrodrigues - I tested this and it appears Presto Client only attempts connect when starting a query, not on startup. So it is okay as is

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -962,6 +979,24 @@ def dump_to_primitive_dict(self):
...
+        if not isinstance(package, Package):
+            # Skip validation since `package` is not a valid `Package` (Pydantic will validate it
+            # later and throw an error).
+            return values
+
+        query_engine = package.get("query_engine")
+        presto = values.get("presto")
+        if query_engine == QueryEngine.PRESTO and presto is None:
+            raise ValueError(
+                f"`presto` config must be non-null when query_engine is" f" `{query_engine}`"
```
```suggestion
                f"`presto` config must be non-null when query_engine is `{query_engine}`"
```

kirkrodrigues (approved):

junhaoliao (approved):

PR #1205, opened 2025-08-14
fix(package): Add garbage collector to valid components list.

haiqi96:
# Description

This PR fixes an issue introduced by https://github.com/y-scope/clp/pull/1035 where the garbage collector was missing from the components list.
 

# Checklist

  breaking change.

# Validation performed

Confirmed that both command launches garbage collector:
- sbin/start-clp.sh
- sbin/start-clp.sh garbage_collector


haiqi96:
@davemarco Can you please review this?

davemarco:
do we want this with presto? my instinct is yes so we should add to presto components as well

haiqi96:
> do we want this with presto? my instinct is yes so we should add to presto components as well

Thanks for raising this point. Updated the code to add the component to presto

davemarco:
Also just thought should probably be in controller components as well.

davemarco:
Lastly, does this garbage collector remove search results in mongoDB? The only difference for native vs. presto is that the searchJobID (collection name) are a hash instead of a counter. Will it still work?

haiqi96:
> Lastly, does this garbage collector remove search results in mongoDB? The only difference for native vs. presto is that the searchJobID (collection name) are a hash instead of a counter. Will it still work?

Oh GG, then i guess it won't work. we will need a regex to match the right collections

davemarco (dismissed):
Didnt test garbage collector, but the startup scripts now look good. I didnt retest start script after latest changes, but lmk if needed

kirkrodrigues (dismissed):
For the PR title, how about:

```
fix(clp-package): Start garbage collector as part of the "all" and "controller" targets; Validate retention periods aren't configured when using the Presto query engine.
```

kirkrodrigues (approved):

PR #1208, opened 2025-08-15
feat(package): Mark running jobs as failed when schedulers start.

haiqi96:
# Description

Currently, if a compression or search job stucks in running state, we don't have a way to recover them, and they will never change their status to succeed or failed. This could confuse user, and will also prevent the archives being properly removed based on the retention period.

This PR propose a change that removes any running jobs when scheduler restarts. When marking a running job as failed, we do the following:
1. Find all associated running tasks under the job and mark them as failed.
2. Update the duration to be 0.

Note that we directly mark job as failed instead of rerunning them because
- For compression job, rerunning a hanging job can cause same data being compressed multiple times.
- For search job, there's no point rerunning the job since very likely user has started another search.
In addition, if the hang is caused by some internal bugs, rerunning the job is very unlikely to resolve the issue.

# Checklist

  breaking change.

# Validation performed

- Manually modified the code to make compression job and search job stuck at running stage.
- made sure that running jobs are marked as failed when scheduler restarts.
Note, the QueryTask are not properly updated. Refer to https://github.com/y-scope/clp/pull/1208/files#r2279708983 for details.

## Summary by CodeRabbit

* **New Features**
  * Automatic cleanup of hanging compression and query jobs at scheduler startup.
  * System-wide recognition of a new "killed" terminal job state; tooling and schedulers stop polling when observed.
* **Bug Fixes**
  * Schedulers abort startup on cleanup failure with clear logs to avoid unstable runs.
* **Refactor**
  * Internal routing/constants harmonized for consistency (no behaviour changes).
* **UI**
  * Job status UI updated to display a new "killed" state.


components/job-orchestration/job_orchestration/scheduler/utils.py
```diff
@@ -0,0 +1,81 @@
...
+            WHERE status={task_status_running}
+            AND job_id IN ({job_id_placeholders_str})
+            """,
+            hanging_job_ids,
+        )
+
+        jobs_update_config = {"status": job_status_failed, "duration": 0}
+        field_set_expressions = [f"{k} = %s" for k in jobs_update_config.keys()]
+        if SchedulerType.COMPRESSION == scheduler_type:
+            field_set_expressions.append("update_time = CURRENT_TIMESTAMP()")
```
haiqi96, line 69: @kirkrodrigues This update_time has make things more complicated.
In the long term, it might be worthwhile adding the update time for query_job as well though.

components/job-orchestration/job_orchestration/scheduler/utils.py
```diff
@@ -0,0 +1,81 @@
...
+        tasks_table_name = COMPRESSION_TASKS_TABLE_NAME
+        task_status_running = CompressionTaskStatus.RUNNING
+        task_status_failed = CompressionTaskStatus.FAILED
+    elif SchedulerType.QUERY == scheduler_type:
+        jobs_table_name = QUERY_JOBS_TABLE_NAME
+        job_status_running = QueryJobStatus.RUNNING
+        job_status_failed = QueryJobStatus.FAILED
+        tasks_table_name = QUERY_TASKS_TABLE_NAME
+        task_status_running = QueryTaskStatus.RUNNING
+        task_status_failed = QueryTaskStatus.FAILED
```
haiqi96: This looks pretty ugly. I have discussed with Zhihao and we both agree that having a class managing the Tablename, job/task status might be the right way to go but it would require more refactoring.

For now, I still feel this is better than asking user to passin all 6 arguments by themselves.

components/job-orchestration/job_orchestration/scheduler/utils.py
```diff
@@ -0,0 +1,81 @@
...
+        job_id_placeholders_str = ",".join(["%s"] * len(hanging_job_ids))
+        db_cursor.execute(
+            f"""
+            UPDATE {tasks_table_name}
+            SET status={task_status_failed}, duration=0
+            WHERE status={task_status_running}
+            AND job_id IN ({job_id_placeholders_str})
+            """,
+            hanging_job_ids,
+        )
```
haiqi96: It seems we don't even update task status from pending to running, so this either won't do anything to task table.
or we hack it by setting `task_status_running = QueryTaskStatus.PENDING`.

Depends on when we will switch to spider, we can either fix the task status reporting properly, or live with it.

haiqi96:
Depending on whether Junhao can add a new Killed state handle to the webui or not, this PR still needs to be updated slightly. Nevertheless, it should be good for review.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -421,6 +426,14 @@ def main(argv):
     logger.info(f"Starting {COMPRESSION_SCHEDULER_COMPONENT_NAME}")
     sql_adapter = SQL_Adapter(clp_config.database)
 
+    try:
+        killed_jobs = kill_hanging_jobs(sql_adapter, SchedulerType.COMPRESSION)
+        if killed_jobs is not None:
+            logger.info(f"Killed {len(killed_jobs)} hanging compression jobs.")
```
haiqi96, line 432:
```suggestion
            logger.info(f"Killed {len(killed_jobs)} hanging query jobs.")
```

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -649,7 +649,7 @@ def start_compression_worker(
     mounts: CLPDockerMounts,
 ):
     celery_method = "job_orchestration.executor.compress"
-    celery_route = f"{QueueName.COMPRESSION}"
+    celery_route = f"{SchedulerType.COMPRESSION}"
```
haiqi96:
```suggestion
    celery_route = SchedulerType.COMPRESSION
```

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -674,7 +674,7 @@ def start_query_worker(
     mounts: CLPDockerMounts,
 ):
     celery_method = "job_orchestration.executor.query"
-    celery_route = f"{QueueName.QUERY}"
+    celery_route = f"{SchedulerType.QUERY}"
```
haiqi96:
```suggestion
    celery_route = SchedulerType.QUERY
```

components/job-orchestration/job_orchestration/scheduler/utils.py
```diff
@@ -0,0 +1,81 @@
...
+from job_orchestration.scheduler.constants import (
+    CompressionJobStatus,
+    CompressionTaskStatus,
+    QueryJobStatus,
+    QueryTaskStatus,
+    SchedulerType,
+)
+
+
+def kill_hanging_jobs(sql_adapter: SQL_Adapter, scheduler_type: str) -> Optional[List[str]]:
```
haiqi96:
```suggestion
def kill_hanging_jobs(sql_adapter: SQL_Adapter, scheduler_type: str) -> Optional[List[int]]:
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -421,6 +426,14 @@ def main(argv):
     logger.info(f"Starting {COMPRESSION_SCHEDULER_COMPONENT_NAME}")
     sql_adapter = SQL_Adapter(clp_config.database)
 
+    try:
+        killed_jobs = kill_hanging_jobs(sql_adapter, SchedulerType.COMPRESSION)
+        if killed_jobs is not None:
+            logger.info(f"Killed {len(killed_jobs)} hanging query jobs.")
```
```suggestion
            logger.info(f"Killed {len(killed_jobs)} hanging compression jobs.")
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -421,6 +426,14 @@ def main(argv):
...
+    except Exception:
+        logger.exception("Failed to kill hanging query jobs.")
```
```suggestion
        logger.exception("Failed to kill hanging compression jobs.")
```

kirkrodrigues (approved):

PR #1217, opened 2025-08-18
feat(clp-s): Add support for duplicate columns in projections.

wraymo:
# Description

This change adds support for duplicate columns in Projection.

When using CLP as a library, results may not always be serialized into a JSON string. In these cases, enforcing uniqueness is unnecessary and, in some scenarios, duplicate columns are expected.

For example, in the Velox–CLP connector, accessing the same nested field through both a UDF and a row type may produce two projection fields. Even if one of them is unused, we still need to allow duplicates.

Since the default value remains false, this change does not affect existing CLP behavior.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

- New Features
  - Added an option (off by default) to allow duplicate columns in projection results, enabling repeated fields to be preserved when desired.

- Bug Fixes
  - Prevents errors in workflows that intentionally produce repeated columns; such workflows now succeed when the option is enabled.

- Notes
  - Default behaviour unchanged for users who do not enable the duplicate-columns option.

### DIFF
--- a/components/core/src/clp_s/search/Projection.cpp
+++ b/components/core/src/clp_s/search/Projection.cpp
@@ -13,12 +13,13 @@ void Projection::add_column(std::shared_ptr<ast::ColumnDescriptor> column) {
     if (ProjectionMode::ReturnAllColumns == m_projection_mode) {
         throw OperationFailed(ErrorCodeUnsupported, __FILE__, __LINE__);
     }
-    if (m_selected_columns.end()
-        != std::find_if(
-                m_selected_columns.begin(),
-                m_selected_columns.end(),
-                [column](auto const& rhs) -> bool { return *column == *rhs; }
-        ))
+    if (false == m_allow_duplicate_columns
+        && m_selected_columns.end()
+                   != std::find_if(
+                           m_selected_columns.begin(),
+                           m_selected_columns.end(),
+                           [column](auto const& rhs) -> bool { return *column == *rhs; }
+                   ))
     {
         // no duplicate columns in projection
         throw OperationFailed(ErrorCodeBadParam, __FILE__, __LINE__);
--- a/components/core/src/clp_s/search/Projection.cpp
+++ b/components/core/src/clp_s/search/Projection.cpp
@@ -18,7 +18,7 @@ void Projection::add_column(std::shared_ptr<ast::ColumnDescriptor> column) {
                    != std::find_if(
                            m_selected_columns.begin(),
                            m_selected_columns.end(),
-                           [column](auto const& rhs) -> bool { return *column == *rhs; }
+                           [&column](auto const& rhs) -> bool { return *column == *rhs; }
                    ))
     {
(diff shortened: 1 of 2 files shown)


components/core/src/clp_s/search/Projection.hpp
```diff
@@ -86,6 +88,7 @@ class Projection {
     absl::flat_hash_set<int32_t> m_matching_nodes;
     std::vector<std::vector<int32_t>> m_ordered_matching_nodes;
     ProjectionMode m_projection_mode{ProjectionMode::ReturnAllColumns};
+    bool m_allow_duplicate_columns;
```
gibber9809:
```suggestion
    bool m_allow_duplicate_columns{false};
```

components/core/src/clp_s/search/Projection.hpp
```diff
@@ -30,7 +30,9 @@ class Projection {
                 : TraceableException(error_code, filename, line_number) {}
     };
 
-    explicit Projection(ProjectionMode mode) : m_projection_mode{mode} {}
+    explicit Projection(ProjectionMode mode, bool allow_duplicate_columns = false)
```
gibber9809, line 40: Probably need a brief docstring for this constructor.

components/core/src/clp_s/search/Projection.cpp
```diff
@@ -13,12 +13,13 @@ void Projection::add_column(std::shared_ptr<ast::ColumnDescriptor> column) {
...
-                m_selected_columns.end(),
-                [column](auto const& rhs) -> bool { return *column == *rhs; }
-        ))
+    if (false == m_allow_duplicate_columns
+        && m_selected_columns.end()
+                   != std::find_if(
+                           m_selected_columns.begin(),
+                           m_selected_columns.end(),
+                           [column](auto const& rhs) -> bool { return *column == *rhs; }
+                   ))
```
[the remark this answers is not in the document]
gibber9809: Since I'm asking for other small changes may as well also use `[&column]` here.

gibber9809 (changes requested):
Changes look good -- just a few nits.

components/core/src/clp_s/search/Projection.hpp
```diff
@@ -30,7 +30,16 @@ class Projection {
                 : TraceableException(error_code, filename, line_number) {}
     };
 
-    explicit Projection(ProjectionMode mode) : m_projection_mode{mode} {}
+    /**
+     * Constructs a Projection object with the specified mode and column handling behavior.
+     *
+     * @param mode Projection mode that determines how expressions are projected.
+     * @param allow_duplicate_columns Whether duplicate column descriptiors are permitted in the
```
gibber9809:
```suggestion
     * @param allow_duplicate_columns Whether duplicate column descriptors are permitted in the
```

gibber9809 (changes requested):

PR #1224, opened 2025-08-18
feat(webui): Move to success state (instead of failure state) after user-cancel action for Presto queries.

davemarco:
# Description

Cancelling a presto query creates triggers a failure state in the presto engine. These were being treated as errors in the webui, and a notification was exposed to user. Instead, to mirror clp/clp-s webui, we want to treat a user cancel as a non-error, and not show a notification or move to failure state. 

To achieve this, we just check if the error name is user cancel, and if so, ignore error.

# Checklist

  breaking change.

# Validation performed

Checked that no error notifications appeared when user cancels query

## Summary by CodeRabbit

- New Features
  - Improved handling of user-cancelled searches: cancellations are now treated as completed, avoiding failure states.

- Bug Fixes
  - Prevents error notifications when a search is cancelled by the user.
  - Ensures the interface no longer displays a failed state for cancelled operations.

- Notes
  - All other errors continue to show failure status and detailed notifications as before.


components/webui/client/src/pages/SearchPage/SearchState/useUpdateStateWithMetadata.ts
```diff
@@ -30,6 +35,13 @@ const useUiUpdateOnDoneSignal = () => {
                 updateSearchUiState(SEARCH_UI_STATE.DONE);
                 break;
             case PRESTO_SEARCH_SIGNAL.FAILED:
+                // Presto reports query cancellation as a failure, but we treat as a successful
+                // completion.
+                if (resultsMetadata.errorMsg === CANCEL_ERROR_NAME) {
+                    updateSearchUiState(SEARCH_UI_STATE.DONE);
+
+                    return;
+                }
```
hoophalab:
```suggestion
                // Presto reports query cancellation as a failure, but we treat as a successful
                // completion.
                if (resultsMetadata.errorName === PRESTO_CANCEL_ERROR_NAME) {
                    updateSearchUiState(SEARCH_UI_STATE.DONE);

                    break;
                }
```

1. We probably should use `break;` here
2. I guess we should test against `errorName`? My `resultsMetadata` output for a cancelled event is
```
Object { _id: "20250818_202701_00007_4zgje", errorMsg: "Query was canceled", errorName: "USER_CANCELED", lastSignal: "FAILED", queryEngine: "presto" }
```

components/webui/client/src/pages/SearchPage/SearchState/useUpdateStateWithMetadata.ts
```diff
@@ -11,6 +11,11 @@ import {SEARCH_UI_STATE} from "./typings";
 import {useResultsMetadata} from "./useResultsMetadata";
 
 
+/*
+* Presto error name for user cancellation
+*/
+const CANCEL_ERROR_NAME = "USER_CANCELED";
```
hoophalab:
```suggestion
const PRESTO_CANCEL_ERROR_NAME = "USER_CANCELED";
```

components/webui/client/src/pages/SearchPage/SearchState/useUpdateStateWithMetadata.ts
davemarco: oh yes that is true

kirkrodrigues (approved):

PR #1231, opened 2025-08-19
chore(package)!: Set `results_cache.retention_period` to 60min so that old search results aren't retained indefinitely by default.

haiqi96:
# Description

As discussed offline with @kirkrodrigues, it would make more sense to enable retention control for search results by default, since user only have the access to the search results generated by the latest query job.

We choose 60 minutes to be the retention period since
1. It's relatively small so stale search results won't be retained for too long.
2. 60 is 2x of the sweep interval (30). Note that it would not make much sense to have retention_period < sweep_interval. 2x is a reasonable factor for the use case.

# Checklist

  breaking change.

# Validation performed

Manually verified that garbage collector starts with the package and the following log is printed:
```
[INFO] Creating search-result-garbage-collector with retention period = 60 minutes
```

## Summary by CodeRabbit

- New Features
  - Results cache now defaults to a 60-minute retention period when not configured, automatically expiring older entries to manage storage and performance.

- Documentation
  - Updated sample configuration comments to reflect the new 60-minute default for results cache retention.

### DIFF
--- a/components/core/CMakeLists.txt
+++ b/components/core/CMakeLists.txt
@@ -166,10 +166,8 @@ if(CLP_NEED_BOOST)
 endif()
 
 if(CLP_NEED_CATCH2)
-    find_package(Catch2 REQUIRED)
-    if (Catch2_FOUND)
-        message(STATUS "Found Catch2 ${Catch2_VERSION}")
-    endif()
+    find_package(Catch2 3.8.0 REQUIRED)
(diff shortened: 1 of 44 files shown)


kirkrodrigues (approved):
For the PR title, how about:

```
chore(package): Set `results_cache.retention_period` to 60min so that old search results aren't retained indefinitely by default.
```

PR #1234, opened 2025-08-19
fix(webui): Expose rate limit configuration (fixes #1019); Increase rate limit to 1000 req/min (fixes #1020).

junhaoliao:
# Description

1. Expose rate limit configuration (fixes #1001 )
2. Increase rate limit to 1000 req/min (fixes #1020 )

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [ ] Necessary docs have been updated, OR no docs need to be updated.
  * To be added as #1233 is addressed

# Validation performed

1. Built and started the package
    ```
    task package
    cd build/clp-package/sbin
    ./start-clp.sh
    ```
2. Visited http://localhost:4000 in Microsoft Edge Version 139.0.3405.102 (Official build) (64-bit)
3. Clear browser cache for the site:
    1. `F12` to open Developer Console in the browser.
    2. Right-clicked on the "Refresh" button on the left of the browser address bar
    3. Clicked on "Empty cache and hard refresh"
4. Idled for 1 minute.
5. Kept refreshing the http://localhost:4000/ingest page at a rate of roughly twice a second. After roughly 30 seconds, observed "Rate limit exceeded, retry in 30 seconds".

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **New Features**
  * Added a configurable Web UI rate limit (RATE_LIMIT) with default 1000 and validation enforcing values > 0.

* **Chores**
  * Wired RATE_LIMIT into Web UI startup environment and runtime configuration; environment examples expanded.

* **Documentation**
  * Added a commented example for rate_limit in the config template; runtime schema default updated from 100 to 1000.

PR #1235, opened 2025-08-19
feat(webui): Add cache control for static files.

junhaoliao:
# Description

Add Cache-Control headers for static files to reduce network traffic
  - Configure no-cache for specific files (e.g., index.html, settings.json)
  - Set long-term caching for all other static files
  - Improve caching strategy for both log viewer and client directories

# Checklist

  breaking change.

# Validation performed

Below test was done without the changes in #1234 

## Stress test

1. Built and started the package
    ```
    task package
    cd build/clp-package/sbin
    ./start-clp.sh
    ```
2. Visited http://localhost:4000 in Microsoft Edge Version 139.0.3405.102 (Official build) (64-bit)
3. Clear browser cache for the site:
    1. `F12` to open Developer Console in the browser.
    2. Right-clicked on the "Refresh" button on the left of the browser address bar
    3. Clicked on "Empty cache and hard refresh"
4. Idled for 1 minute.
5. Kept refreshing the http://localhost:4000/ingest page at a rate of roughly twice a second. After roughly 30 seconds, observed "Rate limit exceeded, retry in 30 seconds".

## Inspect headers

Idled for 1 minute so that the rate limit recovered from the above stress test.

### The Web UI SPA

1. Opened Developer Console in Microsoft Edge Version 139.0.3405.102 (Official build) (64-bit). Switched to the "Network" tab in the Developer Console.
2. Visited http://localhost:4000/ingest 
3. Observed most requests were loaded from the cache:
   image
4. ... except requests for "ingest", "settings.json", "sql" (ignored "content-fontface-intercept.js" because that's caused by a Chrome extension i installed). Inspected the requests and observed "Cache-Control: public, max-age=0" was set for those.
5. Visited http://localhost:4000/ (a different path from http://localhost:4000/ingest though index.html is also served) and repeated Step 4 to ensure the no cache header was set.

### The log viewer SPA

1. Compressed sample logs: `clp-package/sbin/compress.sh ~/sample/hive-24hr`
2. Visited http://localhost:4000/search and performed a search with query string "1". Observed results were shown.
3. Clicked on any "Open file" link in any search result. The log viewer showed up.
4. Clicked the Refresh button on the left of the browser address bar to refresh the page.
5. Opened Developer Console. Switched to the "Network" tab in the Developer Console.
6. Observed most requests were loaded from the cache:
   image
7. ... except requests for "index.html?filePath=/streams/...", "dc4314df-6284-4555-824a-548de0932ebe_0_641.clp.zst" (ignored "content-fontface-intercept.js" because that's caused by a Chrome extension i installed). Inspected the requests and observed "Cache-Control: public, max-age=0" was set for those.

## Summary by CodeRabbit

* New Features
  * Enhanced cache-control for static assets: long-term caching for files like JS/CSS/images, with index.html and settings.json always served fresh.
  * Faster repeat page loads while ensuring users receive up-to-date app shell and configuration.

* Bug Fixes
  * Prevents stale content after deployments, reducing issues with outdated pages or settings.

---
  - https://app.asana.com/0/0/1212888468774326

### DIFF
--- a/components/webui/server/src/routes/static.ts
+++ b/components/webui/server/src/routes/static.ts
@@ -1,11 +1,44 @@
 import path from "node:path";
 import {fileURLToPath} from "node:url";
 
-import {fastifyStatic} from "@fastify/static";
+import fastifyStatic from "@fastify/static";
 import {FastifyPluginAsync} from "fastify";
 
 import settings from "../../settings.json" with {type: "json"};
 
+
+const CACHE_CONTROL_HEADER_KEY = "Cache-Control";
+const CACHE_CONTROL_HEADER_VALUE_NO_CACHE = "public, max-age=0";
+
+// Cache all other static files for 1 year without revalidation.
+const CACHE_CONTROL_HEADER_VALUE_LONG_TERM_CACHE = "public, max-age=31536000, immutable";
+
+
+/**
+ * Configures `Cache-Control` header for static files to reduce network traffic.
+ *
+ * @param res
+ * @param reqPath
+ * @param extraNoCachePaths
+ */
+const setCacheHeaders = (
+    res: Parameters<NonNullable<fastifyStatic.FastifyStaticOptions["setHeaders"]>>[0],
+    reqPath: string,
+    extraNoCachePaths: string[] = []
+): void => {
+    const noCachePaths = [
+        "/index.html",
+        ...extraNoCachePaths,
+    ];
+
+    if (noCachePaths.some((noCachePath) => reqPath.endsWith(noCachePath))) {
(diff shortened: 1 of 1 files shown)


davemarco (commented):
Should the cache control be setup in Nginx when it is deployed in production, rather than fastify?

junhaoliao:
> Should the cache control be setup in Nginx when it is deployed in production, rather than fastify?

Good point. do note adding cache control headers does help reduce load on the Fastify server in all cases. While it's technically possible, and probably most efficient, to handle this in any proxy server's configuration, doing so requires the maintainer to understand our webui's architecture, which is less convenient. our arch may also change over time

by setting the headers in the Fastify server itself, we make the caching hints more self-contained. If a maintainer wants to apply additional or alternative caching rules, they can still override these headers in their proxy configuration

PR #1236, opened 2025-08-19
feat(webui): Set number results using server metadata to display total result count for presto queries. 

davemarco:
# Description

Previous result count only considered the number of results in the table for presto queries. This PR allows the UI to also include the result count from server metadata. After this PR, the result count is no longer limited to 1k results, and will show the true amount of results from presto in query status header. The table is still limited to 1k results.

# Checklist

  breaking change.

# Validation performed

Ran long query and verified results >1000 in header

### DIFF
--- a/components/webui/client/src/pages/SearchPage/SearchQueryStatus/index.tsx
+++ b/components/webui/client/src/pages/SearchPage/SearchQueryStatus/index.tsx
@@ -24,7 +24,7 @@ const SearchQueryStatus = () => {
             {(searchUiState === SEARCH_UI_STATE.QUERYING ||
                 searchUiState === SEARCH_UI_STATE.DONE) && (
                 <Text type={"secondary"}>
-                    Search job #
+                    Search job id:
                     {searchJobId}
                     {" "}
                     found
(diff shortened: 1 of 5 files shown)


davemarco:
im still thinking about how to do that. Im not sure if we should do what was suggested, or just a icon with tooltip. Or just render some message at the bottom of the table.  Either way it will need to be conditionally rendered so may be much for this PR? 

Anyways we probably shouldnt merge this PR until after release, to prevent issues with release

components/webui/client/src/pages/SearchPage/SearchQueryStatus/index.tsx
```diff
@@ -24,7 +24,7 @@ const SearchQueryStatus = () => {
             {(searchUiState === SEARCH_UI_STATE.QUERYING ||
                 searchUiState === SEARCH_UI_STATE.DONE) && (
                 <Text type={"secondary"}>
-                    Search job #
+                    Search job id:
```
hoophalab: Does this read better?
```suggestion
                    Search job
                    {" "}
```
Or at least add a space after the colon

PR #1119, opened 2025-07-21
feat(webui): Migrate to npm workspace; Update taskfile and the package scripts accordingly.

davemarco:
# Description

Converts webui to npm workspace, and modifies taskfiles to deal with node modules in root and packages.

npm workspace allows us to install dependencies in common, like typebox, allowing for more shared types between client and server.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

Package starts, and webui can compress and search logs.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

- New Features
  - Added a shared web UI "common" package with build, lint, and workspace integration.

- Bug Fixes
  - Fixed server static path resolution to improve startup and asset serving reliability.

- Chores
  - Renamed/scoped packages, updated dependencies/devDependencies, removed legacy path aliases, adjusted TypeScript/Vite/ESLint configs, migrated imports to the new common package, and expanded build/packaging and checksum orchestration to include common node_modules.

PR #1242, opened 2025-08-20
fix(taskfile): Simplify `python-component` sources list to avoid go-task/task#2400 (fixes #872); Upgrade and pin `go-task` to v3.44.0 to avoid go-task/task#2401.

### LINKED ISSUE #872: `task package`  hangs with task >= 3.43.1
### Bug

When running `task package` with task version >= 3.43.1, the task command hangs indefinitely.
With some initial investigation, it seems the issue is caused by building any python components, but it's not clear which part of code has caused the issue.

The issue can be work around by using a lower version of task (3.42.0). But since we don't limit the task version in our user docs, we should not ignore this issue.

### CLP version

1f70d17463db65785ceb36eca8b8794e8953e81c

### Environment

Ubuntu 22.04. 

### Reproduction steps

`sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d 3.43.1`
`./bin/task clp-py-utils`

Note: clp-package-utils and job-orchestration also causes hang.

junhaoliao:
# Description

As the title suggests. See https://github.com/y-scope/clp/issues/872#issuecomment-3204227009 for the details of the "bug".

We previously locked Task to versions < 3.43.0 due to critical issues that caused build failures and hangs. Two specific issues were identified:
1. y-scope/clp#872 (reported to go-task/task#2400): Dynamic variables in sources caused glob expansion to hang due to unpopulated placeholders and symlink loops. This was particularly problematic with patterns like `{{.DYNAMIC_VARIABLE}}/**/*` which would expand to `/**/*` and cause infinite symlink traversal.
2. y-scope/clp-ffi-js#110 (reported to go-task/task#2401): Brace expansion was broken in v3.44.1 due to over-escaping of curly braces. The escape function introduced in commit go-task/task@d2bd834 incorrectly escaped `{` and `}` characters, breaking legitimate shell features like `file.{txt,md}` expansion.

This PR makes modification to the `taskfile.yaml`, upgrades the Task tool to version 3.44.0 and removes the previous version constraints that were in place due to critical bugs introduced in Task v3.43.0. The upgrade to 3.44.0 brings in various benefits such as unblocking use cases the leverage globing in sources files: https://github.com/y-scope/clp/pull/1119#discussion_r2277416748

# Checklist

  breaking change.

# Validation performed

1. Downloaded latest release (3.44.1) of go-task.
3. `which task` and `task --version` just to be sure the version was to be used.
4. in the project root:  `rm -rf build .task`
5. in the project root: `task` and observed the command to be completed in less than 3 minutes.

## Summary by CodeRabbit

- Chores
  - Standardized build tooling to Task 3.44.0 across Ubuntu, CentOS Stream 9, Manylinux 2.28, and MuslLinux 1.2 install scripts.
  - Simplified task sources configuration by broadening file matching and excluding build artefacts, removing redundant variables.
- Documentation
  - Updated requirements to specify Task 3.44.0 across README and developer docs.
  - Replaced prior version ranges and related notes with a single fixed version reference.


taskfile.yaml
```diff
@@ -534,18 +534,16 @@ tasks:
...
     sources:
       - "{{.G_BUILD_DIR}}/{{.COMPONENT}}_venv.md5"
-      - "{{.PACKAGE}}/**/*"
       - "{{.ROOT_DIR}}/requirements.txt"
       - "{{.TASKFILE}}"
       - "/etc/os-release"
-      - "pyproject.toml"
+      - "**/*"
+      - exclude: "dist/**/*"
     dir: "components/{{.COMPONENT}}"
```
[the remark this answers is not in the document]
junhaoliao: nothing you listed should exist

docs/src/dev-guide/building-package.md
```diff
@@ -13,8 +13,7 @@ prebuilt version instead, check out the [releases](https://github.com/y-scope/cl
 * Python 3.9 or newer
 * python3-dev
 * python3-venv (for the version of Python installed)
-* [Task] >= 3.40.0 and < 3.43.0
-  * We constrain the version due to unresolved [issues][clp-issue-872].
+* [Task] >= 3.40.0
```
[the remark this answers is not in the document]
junhaoliao: i see the merits of adding the latest tested version, but the 3.40.0 version already serves as a recommendation when developers care enough to look into this doc

hoophalab (commented):
lgtm

Validation: Build clp with task version 3.43.3. It no longer hangs.

For the title, how about:
fix(taskfiles): Simplify `python-component` sources list to avoid the dynamic variable bug in go-task (fixes #872).

hoophalab (dismissed):

junhaoliao:
we should also update https://github.com/y-scope/clp/blob/7aa86846e2823cf97e1c62d3965f70bf44799be2/components/core/tools/scripts/lib_install/ubuntu-jammy/install-prebuilt-packages.sh#L31-L42

taskfile.yaml
```diff
@@ -534,18 +534,16 @@ tasks:
     internal: true
     label: "{{.COMPONENT}}"
     vars:
-      PACKAGE:
-        sh: "echo {{.COMPONENT}} | tr - _"
       VENV_DIR: "{{.G_BUILD_DIR}}/{{.COMPONENT}}/venv"
     requires:
       vars: ["COMPONENT"]
     sources:
       - "{{.G_BUILD_DIR}}/{{.COMPONENT}}_venv.md5"
```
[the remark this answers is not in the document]
junhaoliao: i think you're right. since this is unrelated to #872 , could you create a separate issue for this?

junhaoliao:
> we should also update
> 
> https://github.com/y-scope/clp/blob/7aa86846e2823cf97e1c62d3965f70bf44799be2/components/core/tools/scripts/lib_install/ubuntu-jammy/install-prebuilt-packages.sh#L31-L42

addressed in 57e152a2a47519837fc0527e682181dd8169a1a9

components/core/tools/scripts/lib_install/ubuntu-jammy/install-prebuilt-packages.sh
```diff
@@ -29,14 +29,13 @@ DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
...
 task_pkg_path="$(mktemp -t --suffix ".deb" task-pkg.XXXXXXXXXX)"
 curl \
     --fail \
     --location \
     --output "$task_pkg_path" \
     --show-error \
-    "https://github.com/go-task/task/releases/download/v3.42.1/task_linux_${task_pkg_arch}.deb"
+    "https://github.com/go-task/task/releases/download/v3.44.0/task_linux_${task_pkg_arch}.deb"
 dpkg --install "$task_pkg_path"
 rm "$task_pkg_path"
```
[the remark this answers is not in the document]
junhaoliao: 3.44.1 actually causes an issue in another project https://github.com/y-scope/clp-ffi-js/issues/110 so let's avoid it for now

components/core/tools/scripts/lib_install/ubuntu-jammy/install-prebuilt-packages.sh
```diff
@@ -29,14 +29,13 @@ DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
   unzip
 
 # Install `task`
```
kirkrodrigues, line 31:
```suggestion
# Install `task`
# NOTE: We lock `task` to version 3.44.0 to avoid https://github.com/y-scope/clp/issues/872 and
# https://github.com/y-scope/clp-ffi-js/issues/110
```

components/core/tools/scripts/lib_install/ubuntu-jammy/install-prebuilt-packages.sh
kirkrodrigues, line 1: We should replicate these changes in the other container images.

docs/src/dev-guide/building-package.md
kirkrodrigues:
```suggestion
* [Task] 3.40.0
```

taskfile.yaml
```diff
@@ -545,7 +543,8 @@ tasks:
       - "{{.ROOT_DIR}}/requirements.txt"
```
kirkrodrigues, line 542: We need to remove the "{{.PACKAGE}}/**/*" line.

kirkrodrigues (changes requested):

components/core/tools/scripts/lib_install/ubuntu-jammy/install-prebuilt-packages.sh
junhaoliao: 3.44.0 does not avoid #872 (i.e., if we run 3.44.0 with our taskfile.yaml unmodified, the hang is still expected to occur. I'll submit an issue to https://github.com/go-task/task/ once i have the bandwidth.

how about just:
```suggestion
# Install `task`
# NOTE: We lock `task` to version 3.44.0 to avoid https://github.com/y-scope/clp-ffi-js/issues/110
```

docs/src/dev-guide/building-package.md
junhaoliao: as per the above installation script, shall we write
```suggestion
* [Task] 3.44.0
```

taskfile.yaml
junhaoliao: sorry this is a regression after merging main into branch

components/core/tools/scripts/lib_install/ubuntu-jammy/install-prebuilt-packages.sh
kirkrodrigues: Right. Remind me why we're locking to >= 3.44.0? What's wrong with 3.40+?

junhaoliao: 3.40 does not include this fix https://github.com/go-task/task/commit/a60c2ec3f8925f63ad1f3f4cb28cbbadeb720462 , which blocks use cases like https://github.com/y-scope/clp/pull/1119#discussion_r2277416748

I intended to upgrade task to the latest version; however, it was just found that the latest version 3.44.1 causes the another bug in https://github.com/y-scope/clp-ffi-js/issues/110 . 

do note that https://github.com/go-task/task/commit/a60c2ec3f8925f63ad1f3f4cb28cbbadeb720462 happens to also be the cause that the "dynamic variables do not get populated in `sources`" issue surfaced, which is why we should modify our taskfile.yaml anyways to mitigate the hang. 

(sorry... i can't find a better way to explain these. let me know if there's any confusion unresolved

junhaoliao: previously i intentionally skipped the y-scope/clp-ffi-js#110 issue from any docs here in CLP because CLP does not depend on clp-ffi-js. though now i think it's good to write why a specific version is used so we don't forget in the future when we are itchy upgrading to new versions

junhaoliao: slightly off topic: we could change this https://github.com/y-scope/clp-ffi-js/blob/59890ba9a09bb8205281911735eb48cebb5a5a7c/Taskfile.yml#L133-L135

```diff
-          "{{.G_CLP_FFI_JS_BUILD_DIR}}/{{.G_CLP_FFI_JS_TARGET_PREFIX}}{{.ITEM}}."{d.ts,js,wasm}
+          {{.G_CLP_FFI_JS_BUILD_DIR}}/{{.G_CLP_FFI_JS_TARGET_PREFIX}}{{.ITEM}}.{d.ts,js,wasm}
```

then the latest version of `task` 3.44.1 can also run the clp-ffi-js's `package` task just fine. apparently https://github.com/go-task/task/commit/d2bd834c81548f09462d0232c4e0525cbfe3abcc , which intrudocued some incorrect string escape handling, is the commit that broke clp-ffi-js's `package` task.

kirkrodrigues: Gotcha. These details should really be in the PR description, right? Otherwise reviewers (and developers) don't know why we're upgrading task nor why we're locking to a specific set of versions. Even if we're arbitrarily locking to the latest available version because we don't have time to test other versions, I think we should mention that in the PR description.

I think we should also mention the version bump in the PR title.

junhaoliao: right, sorry. the initial intention of the PR was to lift the version locking so developers can install any versions of `task` (which is sort of important). then the bug in ffi-js was found, and we had to lock the version again. then there were many above recent discoveries that i hadn't document yet

now the issues have been reported to go-task, and i have updated the PR description and title

junhaoliao: i kept the fix keyword in the title but let me know

- we could leave #872 opened until the issue is fixed in go-task, though technically #872 only reports the hang which is now really gone

kirkrodrigues (dismissed):
For the PR title, how about:

```
fix(taskfile): Simplify `python-component` sources list to avoid go-task/task#2400 (fixes #872); Upgrade and pin `go-task` to v3.44.0 to avoid go-task/task#2401.
```

kirkrodrigues (approved):

PR #1245, opened 2025-08-21
feat(kv-ir): Add support for duplicate columns in projections.

gibber9809:
# Description

This PR follows up on #1217 by also allowing projections in kv-ir search to specify duplicate columns, guarded by a flag. The general idea is that if a user is serializing to some non-JSON format they aren't limited to outputting each column a single time, and we run into such situations with the velox connector implementation.

To better support duplicate columns we also change the signature expected by `NewProjectedSchemaTreeNodeCallbackReq` to pass the user both the original key name as well as the index of the projected key within the original list of projections.

# Checklist

  breaking change.

# Validation performed

* Validated that existing unit tests pass.

## Summary by CodeRabbit

- **New Features**
  - Added an option to allow duplicate projected columns in search queries.
  - Extended the callback for newly resolved projections to include a projection index (path + index) for order-aware handling.

- **Documentation**
  - Updated parameter descriptions and examples to document the duplicate-allowance option and the extended callback signature.


components/core/src/clp/ffi/ir_stream/search/QueryHandlerImpl.cpp
```diff
@@ -182,20 +184,24 @@ auto preprocess_query(std::shared_ptr<Expression> query)
...
     std::unordered_set<std::string_view> unique_projected_columns;
     std::vector<std::shared_ptr<ColumnDescriptor>> projected_columns;
-    QueryHandlerImpl::ProjectionMap projected_column_to_original_key;
+    QueryHandlerImpl::ProjectionMap projected_column_to_original_key_and_index;
 
-    for (auto const& [key, types] : projections) {
-        if (unique_projected_columns.contains(key)) {
-            return ErrorCode{ErrorCodeEnum::DuplicateProjectedColumn};
+    for (size_t index = 0; index < projections.size(); ++index) {
+        auto const& [key, types] = projections[index];
```
gibber9809: If we were using c++23 I would use something like `std::views::enumerate`, but I couldn't think of a better way of doing this in c++20.

components/core/src/clp/ffi/ir_stream/search/QueryHandlerImpl.cpp
```diff
@@ -182,20 +184,24 @@ auto preprocess_query(std::shared_ptr<Expression> query)
...
                 QueryHandlerImpl::ProjectionMap>> {
     std::unordered_set<std::string_view> unique_projected_columns;
     std::vector<std::shared_ptr<ColumnDescriptor>> projected_columns;
-    QueryHandlerImpl::ProjectionMap projected_column_to_original_key;
+    QueryHandlerImpl::ProjectionMap projected_column_to_original_key_and_index;
 
-    for (auto const& [key, types] : projections) {
-        if (unique_projected_columns.contains(key)) {
-            return ErrorCode{ErrorCodeEnum::DuplicateProjectedColumn};
+    for (size_t index = 0; index < projections.size(); ++index) {
```
LinZhihao-723: Missing `<cstddef>` for `size_t`

LinZhihao-723: Nit:
```suggestion
    for (size_t index{0}; index < projections.size(); ++index) {
```

components/core/src/clp/ffi/ir_stream/search/QueryHandlerImpl.cpp
```diff
@@ -220,26 +226,29 @@ auto create_projected_columns_and_projection_map(
...
-        QueryHandlerImpl::ProjectionMap const& projected_column_to_original_key
+        QueryHandlerImpl::ProjectionMap const& projected_column_to_original_key_and_index
 )
         -> ystdlib::error_handling::Result<std::pair<
                 QueryHandlerImpl::PartialResolutionMap,
                 QueryHandlerImpl::PartialResolutionMap>> {
     QueryHandlerImpl::PartialResolutionMap auto_gen_namespace_partial_resolutions;
     QueryHandlerImpl::PartialResolutionMap user_gen_namespace_partial_resolutions;
-    for (auto const& [col, key] : projected_column_to_original_key) {
+    for (auto const& [col, key_and_index] : projected_column_to_original_key_and_index) {
```
LinZhihao-723: Since it's not used at all, how about:
```suggestion
    for (auto const& [col, _] : projected_column_to_original_key_and_index) {
```
This is totally optional. Apply it if this makes u think more readable.

components/core/src/clp/ffi/ir_stream/search/NewProjectedSchemaTreeNodeCallbackReq.hpp
```diff
@@ -27,7 +29,8 @@ concept NewProjectedSchemaTreeNodeCallbackReq = std::is_invocable_r_v<
         NewProjectedSchemaTreeNodeCallbackType,
         bool,
         SchemaTree::Node::id_t,
-        std::string_view>;
+        std::string_view,
+        size_t>;
```
LinZhihao-723: What do u think if we bind the index and the key into a pair?

components/core/src/clp/ffi/ir_stream/search/QueryHandlerImpl.hpp
```diff
@@ -117,7 +118,8 @@ class QueryHandlerImpl {
         clp_s::search::ast::DescriptorList::iterator m_next_token_it;
     };
 
-    using ProjectionMap = std::unordered_map<clp_s::search::ast::ColumnDescriptor*, std::string>;
+    using ProjectionMap = std::
+            unordered_map<clp_s::search::ast::ColumnDescriptor*, std::pair<std::string, size_t>>;
```
LinZhihao-723, line 122: A high-level question: the current implementation will create different column descriptors for the same key. What's the reason for not using the same descriptor to track all project columns? I can see the major concern would be that they might be projecting different types of node values. Are there any other concerns?

components/core/src/clp/ffi/ir_stream/search/QueryHandlerImpl.hpp
```diff
@@ -492,11 +498,14 @@ auto QueryHandlerImpl::handle_column_resolution_on_new_schema_tree_node(
     }
 
     auto* col{token_it.get_column_descriptor()};
-    if (m_projected_column_to_original_key.contains(col)) {
+    auto const original_key_and_index_it = m_projected_column_to_original_key_and_index.find(col);
+    if (m_projected_column_to_original_key_and_index.end() != original_key_and_index_it) {
+        auto const& [original_key, projected_index] = original_key_and_index_it->second;
```
LinZhihao-723, line 503:
```suggestion
    if (m_projected_column_to_original_key_and_index.contains(col)) {
        auto const& [original_key, projected_index]
                = m_projected_column_to_original_key_and_index.at(col);
```
I think our guideline prefers this way more than using `find`. With the compiler optimization it should be equivalent to the `find` form.

components/core/src/clp/ffi/ir_stream/search/QueryHandlerImpl.hpp
gibber9809: The major one is projections with the same key name but different types, as you say. Generally though it's also unclear to me that sharing column descriptors for keys with the same name would give us any measurable benefits, so I prefer the simpler solution of just not sharing them.

components/core/src/clp/ffi/ir_stream/search/QueryHandlerImpl.hpp
gibber9809: Are you sure? Older versions of g++ definitely don't optimize this away, and briefly playing around with this in godbolt recent versions of g++ still don't seem to optimize this away.

components/core/src/clp/ffi/ir_stream/search/NewProjectedSchemaTreeNodeCallbackReq.hpp
gibber9809: No strong preference either way myself -- if you prefer binding them into a pair I'll make the change.

components/core/src/clp/ffi/ir_stream/search/QueryHandlerImpl.hpp
LinZhihao-723: Let me double check.

LinZhihao-723: I think u're right. Tried both ways on the latest GCC and clang. Using `contains` + `at` involves two hash lookups, while using `find` + iterator comparison is only one. From the generated assembly, it seems like they can be manually optimized, as the first hash lookup dominates the second. But I guess it's also fair for the compiler not to aggressively look into this case as:
* The user-defined hash function may have side effects.
* `at` involves a throw on bad keys.

Anyway, I think the take-away is we should use `find` for better performance.

components/core/src/clp/ffi/ir_stream/search/QueryHandlerImpl.hpp
LinZhihao-723: Sure, make sense.

LinZhihao-723 (approved):
The PR title lgtm.

PR #1267, opened 2025-08-26
refactor(core): Use C++20 concepts to enforce requirements for logtype and variable dictionary entries, readers, and writers.

gibber9809:
# Description

This PR is the fourth in a series of PRs (#1163 for the third) that is focused on sharing core parsing and search code between clp and clp-s. This will be the final PR in this series for now, though we plan on doing some follow-up work to improve the shared subquery generation code.

The changes of this PR consist of
1. Defining an interface for variable and log type dictionary readers and writers as well as their entries as concepts in `DictionaryConcepts.hpp`.
2. Using those concepts in `EncodedVariableInterpreter.hpp` and `GrepCore.hpp` to ensure that callers pass dictionaries and dictionary entries that conform to the expected interface.

The previous PRs in this series mostly already modified clp and clp-s dictionary implementations to comply with the new interface, though we do make one small change to the clp-s dictionary reader templates to correct some Id types.

# Checklist

  breaking change.

# Validation performed

* Validated that tests still pass.

## Summary by CodeRabbit

- Refactor
  - Renamed public dictionary type aliases from entry_t to Entry.
  - Converted public templates to use concept-constrained parameters and rely on nested Entry types, improving compile-time type safety without behavioural changes.
  - Adjusted query and encoding APIs to use the updated type aliases and constraints.

- Chores
  - Updated build lists to include new concept headers across targets and unit tests to ensure consistent compilation.


components/core/src/clp/DictionaryConcepts.hpp
```diff
@@ -0,0 +1,262 @@
...
+    /**
+     * Parses the next variable from a message according to `ir::get_bounds_of_next_var`,
+     * constructing the constant part of the message's logtype in the processed range at the same
+     * time.
+     * @param msg The original log message.
+     * @param begin_pos_ref The beginning position of the last variable. This parameter is updated
+     * to the beginning position of the next variable.
+     * @param end_pos_ref The ending position of the last variable (exclusive). This parameter is
+     * updated to the ending position of the next variable (exclusive).
+     * @return Whether another variable was found or not.
```
LinZhihao-723:
```suggestion
     * @param parsed_var_ref Returns a view to the parsed variable, if one was found.
     * @return Whether a new variable was parsed.
```
- `parsed_var_ref` was missing.
- Nit: Using `parsed` to replace `found` in the return statement.

components/core/src/clp/DictionaryConcepts.hpp
```diff
@@ -0,0 +1,262 @@
...
+
+    /**
+     * Parses the next variable from a message according to `ir::get_bounds_of_next_var`,
+     * constructing the constant part of the message's logtype in the processed range at the same
+     * time.
+     * @param msg The original log message.
+     * @param begin_pos_ref The beginning position of the last variable. This parameter is updated
+     * to the beginning position of the next variable.
+     * @param end_pos_ref The ending position of the last variable (exclusive). This parameter is
+     * updated to the ending position of the next variable (exclusive).
```
LinZhihao-723:
```suggestion
     * @param begin_pos_ref The beginning position of the last variable. Returns the beginning
     * position of the next variable.
     * @param end_pos_ref The ending position of the last variable (exclusive). Returns the ending
     * position of the next variable (exclusive).
```
I think the standard is to use `Returns xxx` to annotate output reference variables.

components/core/src/clp/DictionaryConcepts.hpp
```diff
@@ -0,0 +1,262 @@
...
+
+    /**
+     * Gets the position and type of a variable placeholder in the logtype.
+     * @param placeholder_ix The index of the placeholder to get the info for.
+     * @param placeholder_ref The type of the placeholder at `placeholder_ix`.
+     * @return The placeholder's position in the logtype, or `SIZE_MAX` if `placeholder_ix` is out
+     * of bounds.
+     */
+    {
+        entry.get_placeholder_info(placeholder_ix, placeholder_ref)
```
LinZhihao-723: `placeholder_ix` -> `placeholder_idx`
We will stick to `idx` and drop the use `ix` for any new code.

components/core/src/clp/DictionaryConcepts.hpp
```diff
@@ -0,0 +1,262 @@
...
+     * the logtype.
+     */
+    {
+        entry.get_num_placeholders()
+    } -> std::same_as<size_t>;
+
+    /**
+     * Gets the position and type of a variable placeholder in the logtype.
+     * @param placeholder_ix The index of the placeholder to get the info for.
+     * @param placeholder_ref The type of the placeholder at `placeholder_ix`.
```
LinZhihao-723:
```suggestion
     * @param placeholder_ref Returns the type of the placeholder at `placeholder_ix`.
```
Same.

components/core/src/clp/DictionaryConcepts.hpp
```diff
@@ -0,0 +1,262 @@
...
+     * @return The placeholder's position in the logtype, or `SIZE_MAX` if `placeholder_ix` is out
+     * of bounds.
+     */
+    {
+        entry.get_placeholder_info(placeholder_ix, placeholder_ref)
+    } -> std::same_as<size_t>;
+
+    /**
+     * @return The dictionary Id for this logtype.
```
LinZhihao-723:
```suggestion
     * @return The dictionary ID for this logtype.
```

components/core/src/clp/DictionaryConcepts.hpp
```diff
@@ -0,0 +1,262 @@
...
+};
+
+/**
+ * Requirement for the variable dictionary entry interface.
+ * @tparam VariableDictionaryEntryType The type of the variable dictionary entry.
+ */
+template <typename VariableDictionaryEntryType>
+concept VariableDictionaryEntryReq = requires(VariableDictionaryEntryType entry) {
+    /**
+     * @return The dictionary Id for this variable.
```
LinZhihao-723:
```suggestion
     * @return The dictionary ID for this variable.
```

components/core/src/clp/DictionaryConcepts.hpp
```diff
@@ -0,0 +1,262 @@
...
+     * @param entries A hash set in which to store the found entries.
+     */
+    {
+        reader.get_entries_matching_wildcard_string(logtype, ignore_case, entries)
+    } -> std::same_as<void>;
+
+    requires std::
+            same_as<typename LogTypeDictionaryReaderType::dictionary_id_t, logtype_dictionary_id_t>;
+
+    requires std::
```
LinZhihao-723: The current implementation takes a template parameter `LogTypeDictionaryEntryType` and sets it with a default value, and later we have an assertion to check whether the argument is the default value. I think this gets a bit weird; like if we want them to be the same, then probably we just don't expose it as the default value. Maybe do the following to explicitly use `LogTypeDictionaryReader::entry_t`:
```C++
template <typename LogTypeDictionaryReaderType>
concept LogTypeDictionaryReaderReq = requires(
        LogTypeDictionaryReaderType reader,
        std::string_view logtype,
        bool ignore_case,
        std::unordered_set<typename LogTypeDictionaryReaderType::entry_t const*>& entries
) {
    requires LogTypeDictionaryEntryReq<typename LogTypeDictionaryReaderType::entry_t>;

    /**
     * Gets entries matching a given logtype.
     * @param logtype
     * @param ignore_case Whether the lookup should be case insensitive.
     * @return A vector of entries matching the given logtype.
     */
    {
        reader.get_entry_matching_value(logtype, ignore_case)
    } -> std::same_as<std::vector<typename LogTypeDictionaryReaderType::entry_t const*>>;

    /**
     * Gets entries matching a wildcard string.
     * @param logtype A wildcard search string.
     * @param ignore_case Whether the search should be case insensitive.
     * @param entries A hash set in which to store the found entries.
     */
    {
        reader.get_entries_matching_wildcard_string(logtype, ignore_case, entries)
    } -> std::same_as<void>;

    requires std::
            same_as<typename LogTypeDictionaryReaderType::dictionary_id_t, logtype_dictionary_id_t>;
};
```
Notice that:
* We should add a check to ensure the entry type matches the concept requirement.
* `entry_t` actually doesn't match our latest naming rule, as it's not a fundamental type. Shall we create an issue to rename it in another PR? (or maybe u can do it in this PR if the change isn't propagating too wild)

components/core/src/clp/DictionaryConcepts.hpp
```diff
@@ -0,0 +1,262 @@
...
+     */
+    {
+        reader.get_entry_matching_value(logtype, ignore_case)
+    } -> std::same_as<std::vector<LogTypeDictionaryEntryType const*>>;
+
+    /**
+     * Gets entries matching a wildcard string.
+     * @param logtype A wildcard search string.
+     * @param ignore_case Whether the search should be case insensitive.
+     * @param entries A hash set in which to store the found entries.
```
LinZhihao-723:
```suggestion
     * @param entries Returns all the matching entries.
```

components/core/src/clp/DictionaryConcepts.hpp
```diff
@@ -0,0 +1,262 @@
...
+concept LogTypeDictionaryReaderReq = requires(
+        LogTypeDictionaryReaderType reader,
+        std::string_view logtype,
+        bool ignore_case,
+        std::unordered_set<LogTypeDictionaryEntryType const*>& entries
+) {
+    /**
+     * Gets entries matching a given logtype.
+     * @param logtype
+     * @param ignore_case Whether the lookup should be case insensitive.
```
LinZhihao-723:
```suggestion
     * @param ignore_case Whether the search should be case insensitive.
```
Just to be consistent with the docstring below.

components/core/src/clp/DictionaryConcepts.hpp
```diff
@@ -0,0 +1,262 @@
...
+concept VariableDictionaryWriterReq = requires(
+        VariableDictionaryWriterType writer,
+        std::string_view value,
+        variable_dictionary_id_t& id_ref
+) {
+    /**
+     * Adds the given variable to the dictionary if it doesn't exist.
+     * @param value
+     * @param id_ref The Id of the variable matching the given entry.
+     * @return Whether this call resulted in inserting a new entry or not.
```
LinZhihao-723:
```suggestion
     * @param id_ref Returns the entry ID of the given variable.
     * @return Whether this call resulted in inserting a new entry.
```
We don't need to add "or not" in a "whether" return statement.

components/core/src/clp/DictionaryConcepts.hpp
```diff
@@ -0,0 +1,262 @@
...
+};
+
+/**
+ * Requirement for the variable dictionary reader interface.
+ * @tparam VariableDictionaryReaderType The type of the variable dictionary reader.
+ * @tparam VariableDictionaryEntryType The type of the entries in the variable dictionary reader.
+ */
+template <
+        typename VariableDictionaryReaderType,
+        typename VariableDictionaryEntryType = typename VariableDictionaryReaderType::entry_t>
```
LinZhihao-723: Same as the logtype reader:
- How about using `VariableDictionaryReaderType::entry_t` directly?
- We need to check it matches the variable entry requirement.

components/core/src/clp/DictionaryConcepts.hpp
```diff
@@ -0,0 +1,262 @@
...
+     * @return The value of the dictionary entry with the given Id.
+     */
+    {
+        reader.get_value(id)
+    } -> std::same_as<std::string const&>;
+
+    /**
+     * Gets entries matching a given variable value.
+     * @param variable The variable value to look up.
+     * @param ignore_case Whether the lookup should be case insensitive.
```
LinZhihao-723:
```suggestion
     * @param ignore_case Whether the search should be case insensitive.
```

components/core/src/clp/DictionaryConcepts.hpp
```diff
@@ -0,0 +1,262 @@
...
+     */
+    {
+        reader.get_entry_matching_value(variable, ignore_case)
+    } -> std::same_as<std::vector<VariableDictionaryEntryType const*>>;
+
+    /**
+     * Gets entries matching a given wildcard string.
+     * @param variable A wildcard search string.
+     * @param ignore_case Whether the search should be case insensitive.
+     * @param entries A hash set in which to store the found entries.
```
LinZhihao-723:
```suggestion
     * @param entries Returns all the matching entries.
```

components/core/src/clp/DictionaryConcepts.hpp
LinZhihao-723, line 1: I realized for dictionary readers, the signatures are almost the same for logtypes and variables. Do u think it's worth making a general dictionary reader requirement to include the common requirements?

components/core/src/clp/GrepCore.hpp
```diff
@@ -132,9 +135,10 @@ class GrepCore {
      * @return SubQueryMatchabilityResult::MayMatch
      */
     template <
-            typename LogTypeDictionaryReaderType,
-            typename VariableDictionaryReaderType,
-            typename LogTypeDictionaryEntryType = typename LogTypeDictionaryReaderType::entry_t>
+            LogTypeDictionaryReaderReq LogTypeDictionaryReaderType,
+            VariableDictionaryReaderReq VariableDictionaryReaderType,
+            LogTypeDictionaryEntryReq LogTypeDictionaryEntryType
+            = typename LogTypeDictionaryReaderType::entry_t>
```
LinZhihao-723: I think if we update the requirement to not explicitly take the entry type, we can only take the dictionary reader type as a template param.

components/core/src/clp/DictionaryConcepts.hpp
gibber9809: Ah, I started using `ix` over `idx` because @kirkrodrigues recently told me that `ix` is preferred. Are we changing that again?

kirkrodrigues: Uh, that might've been a miscommunication. `ix` is what we have in the codebase cause I used to use that for both iterators and indices, but `idx` is more conventional.

components/core/src/clp/DictionaryConcepts.hpp
gibber9809: Sure, I guess I'm fine with changing it to `EntryT` in this PR.

components/core/src/clp/GrepCore.hpp
gibber9809: Makes sense. Doing a similar simplification in `EncodedVariableInterpreter` as well.

components/core/src/clp/DictionaryConcepts.hpp
LinZhihao-723: To clarify: I think we should call it `Entry`. We only use `XxxType` suffix in a template parameter.

components/core/src/clp/DictionaryConcepts.hpp
gibber9809: I'm a bit on the fence about this one -- it will probably save a few lines of code but I think it might end up harder to read.

LinZhihao-723: kk sure~

LinZhihao-723 (dismissed):
For the PR title, how about:
```
refactor(core): Use C++20 concepts to enforce requirements for logtype and variable dictionary entries, readers, and writers.
```

PR #1273, opened 2025-08-28
feat(log-surgeon)!: Add support for a single capture group in a schema rule to have parity with the heuristic parser.

davidlion:
# Description

Previously, when using log surgeon for parsing the full match of a schema rule's regex pattern would be stored as a variable in CLP. This created differences from the heuristic parser's behaviour for certain cases.

For example, the heuristic's "equals" rule can be represented with the regex pattern `.*=(?<var>.*[a-zA-Z0-9].*)`. The heuristic parser will only store the `var` capture group as a variable (storing the prefix `.*=` as static text). When using log surgeon without capture groups this behaviour was not possible as we would store the full match (including the prefix `.*=`) as a variable.

This PR allows schema rules to contain up to 1 capture group. If a capture group is present only the capture's match will be stored as a variable and anything surrounding it will be stored as static text. In the case where the capture is repeated (e.g. `text(?<var>variable)+text)`) all repetitions will be stored together as a single variable.

# Checklist

  breaking change.

# Validation performed

Added unit tests for schema creation with single or multiple captures.
*TODO* Generated archives and matched their contents.

### DIFF
--- a/components/core/src/clp/streaming_archive/writer/Archive.cpp
+++ b/components/core/src/clp/streaming_archive/writer/Archive.cpp
@@ -2,14 +2,17 @@
 
 #include <sys/stat.h>
 
+#include <cstdint>
 #include <filesystem>
-#include <fstream>
-#include <iostream>
+#include <stdexcept>
+#include <string>
+#include <vector>
 
 #include <boost/asio.hpp>
 #include <boost/uuid/uuid.hpp>
 #include <boost/uuid/uuid_generators.hpp>
 #include <boost/uuid/uuid_io.hpp>
+#include <log_surgeon/Constants.hpp>
 #include <log_surgeon/LogEvent.hpp>
 #include <log_surgeon/LogParser.hpp>
 #include <nlohmann/json.hpp>
@@ -23,11 +26,7 @@
 
 using clp::ir::eight_byte_encoded_variable_t;
 using clp::ir::four_byte_encoded_variable_t;
-using log_surgeon::LogEventView;
-using std::list;
-using std::make_unique;
 using std::string;
-using std::unordered_set;
 using std::vector;
 
 namespace clp::streaming_archive::writer {
@@ -315,13 +314,13 @@ Archive::write_msg(epochtime_t timestamp, string const& message, size_t num_unco
     update_segment_indices(logtype_id, var_ids);
 }
 
-void Archive::write_msg_using_schema(LogEventView const& log_view) {
+void Archive::write_msg_using_schema(log_surgeon::LogEventView const& log_view) {
     epochtime_t timestamp = 0;
     TimestampPattern* timestamp_pattern = nullptr;
     auto const& log_output_buffer = log_view.get_log_output_buffer();
     if (log_output_buffer->has_timestamp()) {
-        size_t start;
-        size_t end;
+        size_t start{};
+        size_t end{};
         timestamp_pattern = (TimestampPattern*)TimestampPattern::search_known_ts_patterns(
                 log_output_buffer->get_mutable_token(0).to_string(),
                 timestamp,
@@ -360,7 +359,7 @@ void Archive::write_msg_using_schema(LogEventView const& log_view) {
     if (timestamp_pattern == nullptr) {
         start_pos = log_output_buffer->get_token(1).m_start_pos;
     }
-    uint32_t end_pos = log_output_buffer->get_token(log_output_buffer->pos() - 1).m_end_pos;
+    uint32_t const end_pos = log_output_buffer->get_token(log_output_buffer->pos() - 1).m_end_pos;
     if (start_pos <= end_pos) {
         num_uncompressed_bytes = end_pos - start_pos;
     } else {
@@ -369,7 +368,7 @@ void Archive::write_msg_using_schema(LogEventView const& log_view) {
     }
     for (uint32_t i = 1; i < log_output_buffer->pos(); i++) {
         log_surgeon::Token& token = log_output_buffer->get_mutable_token(i);
-        int token_type = token.m_type_ids_ptr->at(0);
+        auto const token_type{token.m_type_ids_ptr->at(0)};
         if (log_output_buffer->has_delimiters() && (timestamp_pattern != nullptr || i > 1)
             && token_type != static_cast<int>(log_surgeon::SymbolId::TokenUncaughtString)
             && token_type != static_cast<int>(log_surgeon::SymbolId::TokenNewline))
@@ -388,13 +387,13 @@ void Archive::write_msg_using_schema(LogEventView const& log_view) {
                 break;
             }
             case static_cast<int>(log_surgeon::SymbolId::TokenInt): {
-                encoded_variable_t encoded_var;
+                encoded_variable_t encoded_var{};
                 if (!EncodedVariableInterpreter::convert_string_to_representable_integer_var(
                             token.to_string(),
                             encoded_var
                     ))
                 {
-                    variable_dictionary_id_t id;
+                    variable_dictionary_id_t id{};
                     m_var_dict.add_entry(token.to_string(), id);
                     encoded_var = EncodedVariableInterpreter::encode_var_dict_id(id);
                     m_logtype_dict_entry.add_dictionary_var();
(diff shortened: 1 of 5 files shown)


components/core/src/clp/streaming_archive/writer/Archive.cpp
```diff
@@ -422,21 +421,68 @@ void Archive::write_msg_using_schema(LogEventView const& log_view) {
...
-                m_var_dict.add_entry(token.to_string(), id);
-                encoded_var = EncodedVariableInterpreter::encode_var_dict_id(id);
-                m_var_ids.push_back(id);
+                auto const& lexer{log_view.get_log_parser().m_lexer};
+                auto capture_ids{lexer.get_capture_ids_from_rule_id(token_type)};
+
+                // If the variable token contains capture groups, we break the token up by storing
+                // each capture as a variable and any substrings surrounding the capture as part of
+                // the logtype. If there are no capture groups the entire variable token is stored
+                // as a variable.
```
SharafMohamed: Flip the order of the sentences in the docstring to match the order of the cases in the code? I.e., the no capture case first, then the capture case.

components/core/src/clp/streaming_archive/writer/Archive.cpp
```diff
@@ -422,21 +421,68 @@ void Archive::write_msg_using_schema(LogEventView const& log_view) {
...
+                auto const register_ids{lexer.get_reg_ids_from_capture_id(capture_ids.value().at(0))};
+                if (false == register_ids.has_value()) {
+                    throw(std::runtime_error(
+                            "No register IDs found for variable's capture group. Full token: "
+                            + token.to_string()
+                    ));
+                }
+                auto const [start_reg_id, end_reg_id]{register_ids.value()};
+                auto const capture_start{token.get_reversed_reg_positions(start_reg_id).front()};
+                auto const capture_end{token.get_reversed_reg_positions(end_reg_id).back()};
```
SharafMohamed: I'm assuming the behavior we want is to treat multi-valued captures as a single variable. In that case the correct usage should be `.back()` for the start capture and `.front()` for the end capture (since the reg positions are reversed). 

Ideally we'd have a unit-test for this case, but I think multi-valued captures are currently a log-surgeon issue.

components/core/src/clp/streaming_archive/writer/Archive.cpp
```diff
@@ -422,21 +421,68 @@ void Archive::write_msg_using_schema(LogEventView const& log_view) {
...
+                auto const token_view{token.to_string_view()};
+                size_t token_pos{0};
```
SharafMohamed: Is it needed to track `token_pos`. Aren't the segments [0, `capture_start`), [`capture_start`, `capture_end`) and [`capture_end`, `token_view.length()`)?

components/core/tests/test_schema_files/multiple_capture_groups.txt
```diff
@@ -0,0 +1 @@
+multicapture:text(?<group0>var0)text(?<group1>var1)text
```
SharafMohamed, line 3: Technically lexers can exist with the schema as is, so I'm not entirely sure about this comment. But, if any `LogParser` object is created with this schema it will error as it lacks delimiters. So this schema file is malformed according to Log Surgeon.

components/core/tests/test_schema_files/single_capture_group.txt
```diff
@@ -0,0 +1 @@
+capture:text(?<group>var)text
```
SharafMohamed: Same as above.

components/core/src/clp/streaming_archive/writer/Archive.cpp
SharafMohamed: Perhaps the following is the most direct (and most log-surgeon independent) implementation. Manipulate token and call `to_string_view` as needed:
- Store `token_end_pos = token.m_end_pos`
- For the prefix static-text: Set `token.m_end_pos= capture_start`
- For the variable itself: Set `token.m_start_pos= capture_start` and `token.m_end_pos= capture_end`
- For the suffix static-text: Set `token.m_start_pos= capture_end` and `token.m_end_pos= token_end_pos `

This way CLP does not care about how the tokens wrap around is handled internally. Also nice for future-proofing in case LS changes its behavior.

We should also unit-test this if possible.

SharafMohamed: Also: for clarity, maybe format the code such that the prefix is a block with a newline after it, the variable is a block with a newline after it, and the suffix is a block with a newline after it.

components/core/src/clp/streaming_archive/writer/Archive.cpp
```diff
@@ -315,30 +316,140 @@ Archive::write_msg(epochtime_t timestamp, string const& message, size_t num_unco
...
+            );
+            m_var_ids.push_back(id);
+            m_encoded_vars.push_back(EncodedVariableInterpreter::encode_var_dict_id(id));
+            m_logtype_dict_entry.add_dictionary_var();
+
+            auto const after_capture{
+                    token_view.get_sub_token(capture_end_pos, token_view.get_end_pos())
+                            .to_string_view()
+            };
+            m_logtype_dict_entry.add_constant(after_capture, 0, after_capture.size());
```
gibber9809:
```suggestion
            m_logtype_dict_entry.add_constant(after_capture, 0, after_capture.size());
```
Sorry, decided to also read through the rest of the PR, and realized an issue here. It seems like `add_constant` won't do any escaping for variable placeholders, which means that this code could end up adding extra unescaped placeholders to the logtype if we have characters corresponding to variable placeholders in the static text here. The heuristic path seems to get around this since it doesn't use `add_constant`, and instead seems to do escaping inside of `parse_next_var`.

I think this may also apply to the `TokenNewline` and `TokenUncaughtString` cases, but I'm not as sure.

Generally seems like this is solved in other parts of the code by using `ir::append_constant_to_logtype` with an appropriate handler for escaping.

components/core/src/clp/streaming_archive/writer/Archive.cpp
```diff
@@ -315,30 +317,144 @@ Archive::write_msg(epochtime_t timestamp, string const& message, size_t num_unco
...
+                variable_dictionary_id_t id{};
+                m_var_dict.add_entry(token_view.to_string(), id);
+                m_var_ids.push_back(id);
+                m_encoded_vars.push_back(EncodedVariableInterpreter::encode_var_dict_id(id));
+                m_logtype_dict_entry.add_dictionary_var();
+                break;
+            }
+
+            auto capture_pos_result{
+                    log_view.get_capture_position(token_view, captures.value().at(0))
```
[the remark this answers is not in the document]
davidlion: Added a check with a log and break statement. This shouldn't ever happen, but is a possibility exposed by the log surgeon API.

components/core/src/clp/LogTypeDictionaryEntry.hpp
```diff
@@ -92,6 +92,11 @@ class LogTypeDictionaryEntry : public DictionaryEntry<logtype_dictionary_id_t> {
      */
     void add_escape();
 
+    /**
+     * Adds static text to the logtype escaping any placeholder characters.
```
gibber9809:
```suggestion
     * Adds static text to the logtype, escaping any placeholder characters.
```
Nit.

components/core/src/clp/LogTypeDictionaryEntryReq.hpp
```diff
@@ -74,6 +75,16 @@ concept LogTypeDictionaryEntryReq = requires(
      */
     { entry.add_dictionary_var() } -> std::same_as<void>;
 
+    /**
+     * Adds an escape character
```
gibber9809:
```suggestion
     * Adds an escape character.
```
Nit.

components/core/src/clp/LogTypeDictionaryEntryReq.hpp
```diff
@@ -74,6 +75,16 @@ concept LogTypeDictionaryEntryReq = requires(
...
+     */
+    { entry.add_escape() } -> std::same_as<void>;
+
+    /**
+     * Adds static text to the logtype escaping any placeholder characters.
```
gibber9809:
```suggestion
     * Adds static text to the logtype, escaping any placeholder characters.
```
Nit.

components/core/src/clp_s/DictionaryEntry.hpp
```diff
@@ -107,6 +107,11 @@ class LogTypeDictionaryEntry : public DictionaryEntry<clp::logtype_dictionary_id
      */
     void add_escape();
 
+    /**
+     * Adds static text to the logtype escaping any placeholder characters.
```
gibber9809:
```suggestion
     * Adds static text to the logtype, escaping any placeholder characters.
```
Nit.

gibber9809 (approved):
LGTM. For PR title I can't think of anything much better off the top of my head, but maybe `achieve parity` or `reach parity` instead of `have parity` would be better?

davidlion:
> LGTM. For PR title I can't think of anything much better off the top of my head, but maybe `achieve parity` or `reach parity` instead of `have parity` would be better?

Makes sense to me. Double checked for any other improvements with the big boss too.

PR #1277, opened 2025-08-29
fix(clp-s):  Handle pure wildcards and unexpected literal types correctly in `EvaluateTimestampIndex` (fixes #1096).

### LINKED ISSUE #1096: clp-s: `EvaluateTimestampIndex` pass doesn't handle matching against wildcards.
### Bug

Currently, the `EvaluateTimestampFilter` class does not properly handle matching against pure wildcards.

E.g. if "ts" is the authoritative timestamp then `ts: *` will crash because `EvaluateTimestampFilter` attempts to convert `*` to an integer.

The code has the following comment before it static casts the literal to `Integral`:
```
// this is safe after type narrowing because all DateType literals are either
// Integral or a derived class of Integral
```
The issue is that after type narrowing the literal is either Integral OR it is a pure wildcard stored as a string literal.

We should be able to fix this issue by adding a special case to the evaluation logic to check whether the literal predicate is a pure wildcard before attempting to convert it to an integer.

### CLP version

v0.4.0

### Environment

ubuntu jammy docker container

### Reproduction steps

* Compress a log like `{"ts": "123"}` with `--timestamp-key ts`
* Execute the search query `'ts: *'`

gibber9809:
# Description

This PR fixes several crashes in `EvaluateTimestampIndex` that were caused by the assumptions that:
1. Filters reaching the evaluation code would always have a literal
2. That literal would always be of integral type
which on the surface appear valid, because the pass is designed to run after type narrowing, meaning that the column must actually be able to match integral types.

These assumptions are untrue in the following scenarios:
1. The literal is a pure wildcard and can actually match any type (the literal is a `StringLiteral` in this case)
2. The literal is null and has been used in a clause such as `NOT ts: null`
3. The filter has no literal because it is an `EXISTS` or `NEXISTS` filter

The code has been updated to:
1. Not attempt to evaluate timestamp filters when the operation is `EXISTS`/`NEXISTS`
2. Not attempt to evaluate timestamp filters when there is no literal
3. Handle evaluation when the literal is a pure wildcard
4. Verify that the literal is actually an integral in the expected case

We also add a test case that exercises the behaviour that caused a crash prior to this PR.

# Checklist

  breaking change.

# Validation performed

* Manually validated that cases from #1096 no longer crash
* Added passing test case that exercises crashing behaviour from #1096 

## Summary by CodeRabbit

- **New Features**
  - Improved timestamp-based search evaluation with wildcard-aware matching and refined EXISTS/NEXISTS behaviour.
  - Test utilities accept an optional timestamp key (can be omitted) to customise archive compression inputs.

- **Bug Fixes**
  - Safer handling of missing/invalid filter operands; indeterminate cases now return Unknown.
  - More robust numeric evaluation and correct inversion of filter results.

- **Tests**
  - Updated tests to pass the new optional parameter and added cases for escaped/wildcard queries.

### DIFF
--- a/components/core/src/clp_s/search/EvaluateTimestampIndex.cpp
+++ b/components/core/src/clp_s/search/EvaluateTimestampIndex.cpp
@@ -120,15 +120,9 @@ EvaluatedValue EvaluateTimestampIndex::run(std::shared_ptr<Expression> const& ex
             }
             Integral64 integral = integral_literal->get();
             if (std::holds_alternative<int64_t>(integral)) {
-                ret = range_it->second->evaluate_filter(
-                        filter->get_operation(),
-                        std::get<int64_t>(integral)
-                );
+                ret = range_it->second->evaluate_filter(filter_op, std::get<int64_t>(integral));
             } else {
-                ret = range_it->second->evaluate_filter(
-                        filter->get_operation(),
-                        std::get<double>(integral)
-                );
+                ret = range_it->second->evaluate_filter(filter_op, std::get<double>(integral));
             }
 
             if (ret == EvaluatedValue::True) {
--- a/components/core/src/clp_s/search/EvaluateTimestampIndex.cpp
+++ b/components/core/src/clp_s/search/EvaluateTimestampIndex.cpp
@@ -65,6 +65,11 @@ EvaluatedValue EvaluateTimestampIndex::run(std::shared_ptr<Expression> const& ex
              range_it != m_timestamp_dict->tokenized_column_to_range_end();
              range_it++)
         {
+            // Don't attempt to evaluate the timestamp index against columns with wildcard tokens.
+            if (column->has_unresolved_tokens()) {
+                continue;
+            }
+
             std::vector<std::string> const& tokens = range_it->first;
             auto const& descriptors = column->get_descriptor_list();
             if (tokens.size() != descriptors.size()) {
@@ -72,8 +77,8 @@ EvaluatedValue EvaluateTimestampIndex::run(std::shared_ptr<Expression> const& ex
             }
 
             bool matched = true;
-            for (size_t i = 0; i < descriptors.size(); ++i) {
-                if (tokens[i] != descriptors[i].get_token() || descriptors[i].wildcard()) {
+            for (size_t i{0ULL}; i < descriptors.size(); ++i) {
+                if (tokens[i] != descriptors[i].get_token()) {
                     matched = false;
                     break;
                 }
--- a/components/core/src/clp_s/search/EvaluateTimestampIndex.cpp
+++ b/components/core/src/clp_s/search/EvaluateTimestampIndex.cpp
@@ -66,7 +66,7 @@ EvaluatedValue EvaluateTimestampIndex::run(std::shared_ptr<Expression> const& ex
              range_it++)
         {
             // Don't attempt to evaluate the timestamp index against columns with wildcard tokens.
-            if (column->has_unresolved_tokens()) {
+            if (column->is_unresolved_descriptor()) {
                 continue;
             }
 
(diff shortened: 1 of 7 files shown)


components/core/src/clp_s/search/EvaluateTimestampIndex.cpp
```diff
@@ -82,20 +87,40 @@ EvaluatedValue EvaluateTimestampIndex::run(std::shared_ptr<Expression> const& ex
...
-            Integral64 literal = std::static_pointer_cast<Integral>(filter->get_operand())->get();
-            if (std::holds_alternative<int64_t>(literal)) {
-                ret = range_it->second->evaluate_filter(
-                        filter->get_operation(),
-                        std::get<int64_t>(literal)
-                );
+            // Leave handling EXISTS/NEXISTS to schema matching.
+            auto const filter_op{filter->get_operation()};
+            if (FilterOperation::EXISTS == filter_op || FilterOperation::NEXISTS == filter_op) {
+                return EvaluatedValue::Unknown;
```
wraymo: Correct me if I'm wrong. `ts: *` is an existence check and we convert it from `EQ` to `EXISTS` before evaluating the timestamp filter. Why do we return `Unknown` here even when we find a match?

components/core/src/clp_s/search/EvaluateTimestampIndex.cpp
```diff
@@ -82,20 +87,40 @@ EvaluatedValue EvaluateTimestampIndex::run(std::shared_ptr<Expression> const& ex
...
+            }
+
+            auto const literal{filter->get_operand()};
+            // Defend against future FilterOperation types that don't take a literal.
+            if (nullptr == literal) {
+                return EvaluatedValue::Unknown;
+            }
+
+            // Handle the case where the literal is a pure wildcard.
+            if (literal->as_any(filter_op)) {
```
wraymo: And do we have still have an `EQ` case here if we it's already converted to `EXISTS`?

components/core/src/clp_s/search/EvaluateTimestampIndex.cpp
gibber9809: This is just evaluating against the timestamp index -- it won't tell us whether the timestamp exists on a given record/schema, so we leave that up to the schema matching pass.

components/core/src/clp_s/search/EvaluateTimestampIndex.cpp
gibber9809: We don't if we run this after the convert to exists pass. Thinking about this more though we should also return `Unknown` here and just leave it up to schema matching for the same reason I mention in my other comment.

Since that's the case I think I can just delete this code block and rely on following through to the `Unknown` case below.

gibber9809: Actually I can probably also delete the special case dedicated to exists/nexists and rely on the check for whether the filter has a literal -- going to do a pass and get rid of redundant special cases.

components/core/src/clp_s/search/EvaluateTimestampIndex.cpp
wraymo: Quick question: what happens if an archive contains some messages without timestamp fields? I think we simply skip recording timestamps for those messages? But if a query includes timestamp filters and none of the timestamped messages in that archive match (out of range), should we still scan the remaining messages that don’t have timestamp fields?

gibber9809: Yeah we currently don't record any timestamps and will skip them in a scan that contains a timestamp filter. It probably makes sense to add some options for users to get around this (e.g. assign timestamp based on surrounding log events at ingestion time, or potentially still search log events missing timestamps if the timestamp range index matches), but that's future work.

components/core/src/clp_s/search/EvaluateTimestampIndex.cpp
wraymo: Yeah, that's cool

wraymo (approved):

PR #1280, opened 2025-09-02
fix(webui): Defer query done status until all result data is delivered instead of relying on Presto's finished state.

davemarco:
# Description

I noticed that sometimes the state is updated to `FINISHED`, and the results are shown in green (DONE state), but some more results can still stream in. Not sure if always happens or just due to slow/overloaded machine i'm using. Nonetheless, effectively it appears the issue is that Presto completes the query and updates the state, but the UI has yet to dump all the results in Mongo. 

This PR fixes the issue by setting the done state only after all the data is recieved. 

# Checklist

  breaking change.

# Validation performed

### DIFF
--- a/components/webui/server/src/routes/api/presto-search/index.ts
+++ b/components/webui/server/src/routes/api/presto-search/index.ts
@@ -182,6 +182,20 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
                         },
                         success: () => {
                             request.log.info("Presto search succeeded");
+                            if (false === isResolved) {
+                                request.log.error(
+                                    "Presto query finished received before searchJobId was resolved; "
+                                );
+                                return;
+                            } else {
+                                searchResultsMetadataCollection.updateOne(
+                                    {_id: searchJobId},
+                                    {$set: {lastSignal: PRESTO_SEARCH_SIGNAL.DONE}}
+                                ).catch((err: unknown) => {
+                                    request.log.error(err, "Failed to update Presto metadata");
+                                });
+                            }
+
                         },
                         timeout: null,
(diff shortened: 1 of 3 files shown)


components/webui/server/src/routes/api/presto-search/index.ts
```diff
@@ -170,17 +170,23 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
                                 isResolved = true;
                                 resolve(queryId);
-                            } else {
-                                // Update metadata on subsequent calls
-                                searchResultsMetadataCollection.updateOne(
-                                    {_id: queryId},
-                                    {$set: {lastSignal: newState}}
-                                ).catch((err: unknown) => {
-                                    request.log.error(err, "Failed to update Presto metadata");
-                                });
```
hoophalab, line 180: state update is removed

davemarco:
> Alternatively, we can stick with the solution in the PR. If we do, many of the PRESTO_SEARCH_SIGNAL values will become redundant since the state update is removed. If we go this route, I suggest we simplify PRESTO_SEARCH_SIGNAL to just three values and we delete the state callback completely:

I prefer this working on it

components/webui/server/src/routes/api/presto-search/index.ts
davemarco: this is intentional

davemarco:
@hoophalab i made changes discussed

hoophalab:
how about this title:
```
fix(webui): Simplify Presto search signal; Defer query done status until all data is delivered; Reset the number of results to zero when submitting a new query.
```

PR #1283, opened 2025-09-04
feat(clp-package): Warn the user if they do not use the `--timestamp-key` flag when compressing with the `clp-s` storage engine

quinntaylormitchell:
# Description

When using the `clp-s` storage engine, users are currently not warned if they do not use the `--timestamp-key` flag in the compression command. This PR introduces a warning that notifies the user if they do not specify the `--timestamp-key` flag. This will create a better experience for users whose logs have a timestamp key and who have forgotten to specify it. The warning is phrased in such a way that users whose logs do _not_ have a timestamp key (or users who have otherwise _intentionally_ omitted the `--timestamp-key`) will not be deterred from using the `clp-s` storage engine to compress their logs.

The warning reads as follows:

```text
`--timestamp-key` not specified. Events will not have assigned timestamps and can only be searched from the command line without a timestamp filter.
```

The behaviour of compression scripts is otherwise unchanged, i.e., compression jobs will still proceed regardless of whether the user uses the `--timestamp-key` flag.

# Checklist

  breaking change.

# Validation performed

Compressed without using the `--timestamp-key` flag; received the warning correctly.

## Summary by CodeRabbit

- **New Features**
  - Displays a runtime warning during compression after dataset validation when no timestamp key is provided, advising use of the --timestamp-key flag.
  - No change to behaviour or outputs; the timestamp flag is included in the compression command only when explicitly supplied by the user.
  - Does not alter public interfaces or return values.

### DIFF
--- a/components/clp-package-utils/clp_package_utils/scripts/compress.py
+++ b/components/clp-package-utils/clp_package_utils/scripts/compress.py
@@ -205,7 +205,8 @@ def main(argv):
 
         if parsed_args.timestamp_key is None:
             logger.warning(
-                "If your logs have a timestamp field, you should use the --timestamp-key flag."
+                "`--timestamp-key` not specified. Events will not have assigned timestamps and can "
+                "only be searched from the command line without a timestamp filter."
             )
     elif dataset is not None:
         logger.error(f"Dataset selection is not supported for storage engine: {storage_engine}.")


components/clp-package-utils/clp_package_utils/scripts/compress.py
```diff
@@ -202,6 +202,12 @@ def main(argv):
         except Exception as e:
             logger.error(e)
             return -1
+
+        t_key = parsed_args.timestamp_key
+        if t_key is None:
+            logger.warning(
+                f"If your logs have a timestamp column, you should use the --timestamp-key flag."
+            )
```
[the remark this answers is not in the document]
quinntaylormitchell: technically, "" (i.e., the empty string) is still a valid key, so we don't need to account for that. I will remove the `f`-string and change "column".

components/clp-package-utils/clp_package_utils/scripts/compress.py
```diff
@@ -202,6 +202,11 @@ def main(argv):
         except Exception as e:
             logger.error(e)
             return -1
+
+        if parsed_args.timestamp_key is None:
+            logger.warning(
+                "If your logs have a timestamp field, you should use the --timestamp-key flag."
```
junhaoliao: i feel we can put some more emphasis on the consequences. how about
```suggestion
                "`--timestamp-key` not specified. Events will not have assigned timestamps and can "  
                "only be searched from the command line without a timestamp filter."
```

quinntaylormitchell: I agree!

junhaoliao (approved):

PR #1289, opened 2025-09-07
refactor(webui): Move shared types to common workspace. 

davemarco:
# Description

Moves common types into common workspace. Includes route types, and some other shared types like nullable. 

Imports keep common folder structure, instead of using barrel file which is discouraged.

I didnt move some stream metadata types since the types are not exactly the same on server and client. This is due to the types existing prior to new webui work. These types/code need to be refactored first, then can be moved.

# Checklist

  breaking change.

# Validation performed

UI loads. Query still sent

## Summary by CodeRabbit

- Bug Fixes
  - Improved reliability of cancelling and clearing searches by ensuring job IDs are handled consistently.

- Refactor
  - Centralized shared types, schemas and enums to unify data contracts across the app.
  - Broadened common package exports to allow direct subpath imports.

- Style
  - Simplified ESLint rules with pattern-based exceptions for TypeBox-related identifiers.

- Chores
  - Updated import paths across client and server to use the new common module subpaths.

### DIFF
--- a/components/core/src/clp/EncodedVariableInterpreter.hpp
+++ b/components/core/src/clp/EncodedVariableInterpreter.hpp
@@ -6,6 +6,8 @@
 #include <unordered_set>
 #include <vector>
 
+#include <string_utils/string_utils.hpp>
+
 #include "ffi/ir_stream/decoding_methods.hpp"
 #include "ir/LogEvent.hpp"
 #include "ir/types.hpp"
@@ -455,7 +457,8 @@ bool EncodedVariableInterpreter::encode_and_search_dictionary(
         add_float_var(logtype);
         sub_query.add_non_dict_var(encoded_var);
     } else {
-        auto const entries = var_dict.get_entry_matching_value(var_str, ignore_case);
+        auto const unescaped_var_string{string_utils::unescape_string(var_str)};
+        auto const entries = var_dict.get_entry_matching_value(unescaped_var_string, ignore_case);
         if (entries.empty()) {
             // Not in dictionary
             return false;
--- a/components/core/src/clp/string_utils/string_utils.cpp
+++ b/components/core/src/clp/string_utils/string_utils.cpp
@@ -3,6 +3,8 @@
 #include <algorithm>
 #include <charconv>
 #include <cstring>
+#include <string>
+#include <string_view>
 
 using std::string;
 using std::string_view;
@@ -186,6 +188,22 @@ string clean_up_wildcard_search_string(string_view str) {
     }
 
     return cleaned_str;
+}
+
+auto unescape_string(std::string_view str) -> std::string {
+    std::string unescaped_str;
+    bool escaped{false};
+    for (auto const c : str) {
+        if (escaped) {
+            unescaped_str.push_back(c);
+            escaped = false;
+        } else if ('\\' == c) {
+            escaped = true;
+        } else {
+            unescaped_str.push_back(c);
+        }
+    }
+    return unescaped_str;
 }
 
 bool wildcard_match_unsafe(string_view tame, string_view wild, bool case_sensitive_match) {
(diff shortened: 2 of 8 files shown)


components/webui/client/src/pages/SearchPage/SearchControls/search-requests.ts
```diff
@@ -40,7 +42,10 @@ const handleClearResults = () => {
     }
 
     clearQueryResults(
-        {searchJobId, aggregationJobId}
+        {
+            searchJobId: Number(searchJobId),
+            aggregationJobId: Number(aggregationJobId),
```
hoophalab, line 47: How about changing the type of `searchJobId` and `aggregationJobId` in zustand store to number? We can then delete type conversions in those places:

```
pages/SearchPage/SearchControls/SearchButton/CancelButton.tsx:                searchJobId: Number(searchJobId),
pages/SearchPage/SearchControls/SearchButton/CancelButton.tsx:                aggregationJobId: Number(aggregationJobId),
pages/SearchPage/SearchControls/search-requests.ts:            searchJobId: Number(searchJobId),
pages/SearchPage/SearchControls/search-requests.ts:            aggregationJobId: Number(aggregationJobId),
pages/SearchPage/SearchControls/search-requests.ts:            store.updateSearchJobId(searchJobId.toString());
pages/SearchPage/SearchControls/search-requests.ts:            store.updateAggregationJobId(aggregationJobId.toString());
```

components/webui/client/src/api/stream-files/index.ts
hoophalab, line 1: How about cleaning up this a bit?

If my understanding is correct, `submitExtractStreamJob` should take `Static<StreamFileExtractionSchema>` as props and return `Static<StreamFileMetadataSchema>`.

The following content in `client/src/typings/query.ts` is duplicated with `StreamFileMetadataSchema` in `server/src/typings/stream-files.ts`  (The `_id`  field is incorrect).
```
interface ExtractStreamResp {
    _id: string;
    begin_msg_ix: number;
    end_msg_ix: number;
    is_last_chunk: boolean;
    path: string;
    stream_id: string;
}
```

As a result, can we
1. move `StreamFileMetadataSchema` to `common`
2. use common types in `submitExtractStreamJob`'s prototype
3. delete duplicate `interface ExtractStreamResp`
4. delete  `server/src/typings/stream-files.ts`
(There is a `type StreamFilesCollection = Collection<StreamFileMetadata>;`, which is only used in two places. I guess we can delete that and simply write `Collection<StreamFileMetadata>` .)

components/webui/common/eslint.config.mjs
```diff
@@ -13,6 +13,27 @@ const EslintConfig = [
...
+                        "Type.Any",
+                        "Type.Enum",
+                        "Type.Integer",
+                        "Type.Literal",
+                        "Type.Null",
+                        "Type.Required",
+                        "Type.Union",
+                        "Value.Errors",
+                        "Value.Parse",
+                    ],
```
hoophalab:
```suggestion
                    capIsNewExceptionPattern: "(^Type\\..)|(^Value\\..)",
```
How about using this flag? If yes, we should also update `eslint.config.mjs` in client and server. Let me know if you feel this is a bad idea.

components/webui/client/src/pages/SearchPage/SearchControls/search-requests.ts
davemarco: we cant since also using searchJobId for presto. I thought about it tho

components/webui/client/src/api/stream-files/index.ts
davemarco: I wanted to do this in another PR, see description, but i think a good idea to fix these

components/webui/common/eslint.config.mjs
davemarco: I think good idea I will test

components/webui/client/src/api/stream-files/index.ts
davemarco: I thought about adding a new state for presto searchJob ID, but it will be a mess i think in alot of files

hoophalab: @coderabbitai Can you create an issue on this?

PR #1293, opened 2025-09-09
feat(webui): Set up ANTLR tooling for future client-side SQL parsing; Move clp-s parser generation to `codegen.yaml`; Remove unused client `package-lock.json`.

hoophalab:
# Description

1. Add the ANTLR dependency in the webui client.
2. Use the grammar file from the 0.293 release of the Presto repository
https://github.com/prestodb/presto/blob/release-0.293/presto-parser/src/main/antlr4/com/facebook/presto/sql/parser/SqlBase.g4
permalink: https://github.com/prestodb/presto/blob/2f48ff63a3d6d249b4062546003126ddbbb989ee/presto-parser/src/main/antlr4/com/facebook/presto/sql/parser/SqlBase.g4
3. Include the generated parser files directly in the codebase similar to #925.
4. Add an example validate function for SQL input.

# Checklist

  breaking change.

# Validation performed

`validate` detects incorrect sql queries.
`clp-s-generate-parsers` works the same as before.
`webui-generate-parser` generates the parser correctly.

## Summary by CodeRabbit

- New Features
  - Introduced SQL syntax validation in the web UI with case-insensitive parsing and clear line/column error messages.

- Chores
  - Added a new runtime dependency to support the parser.
  - Centralized and automated parser code generation via shared tasks.
  - Updated CI workflow to use the new codegen task.
  - Excluded generated files from linting to reduce noise.
  - Updated notices to include third‑party attribution.

### DIFF
--- a/taskfiles/codegen.yaml
+++ b/taskfiles/codegen.yaml
@@ -56,9 +56,9 @@ tasks:
           INCLUDE_PATTERNS: ["{{.OUTPUT_DIR}}"]
     generates: ["{{.CHECKSUM_FILE}}"]
     sources:
+      - "{{.ROOT_DIR}}/taskfile.yaml"
       - "{{.SRC_DIR}}/SqlBase.g4"
       - "{{.TASKFILE}}"
-      - "{{.ROOT_DIR}}/taskfile.yaml"
     dir: "{{.SRC_DIR}}"
     cmds:
       - "rm -rf '{{.OUTPUT_DIR}}'"


components/webui/package-lock.json
hoophalab, line 1: @davemarco `npm install -w client antlr4` only updates this `package-lock.json`. I guess `webui/client/package-lock.json` and `webui/server/package-lock.json` are no longer used after the workspace PR. Can we delete them?

davemarco: yes we should delete them. They were probably accidently added back when main was merged.

components/webui/client/package.json
```diff
@@ -8,7 +8,8 @@
     "lint:check": "eslint",
     "lint:fix": "eslint --fix",
     "start": "vite",
-    "guided": "VITE_GUIDED_DEV=true npm run start"
+    "guided": "VITE_GUIDED_DEV=true npm run start",
+    "generate-parser": "cd src/SqlParser && antlr4 -Dlanguage=TypeScript -o generated -no-listener -Xexact-output-dir SqlBase.g4 && find generated -type f -and -not -name '*.ts' -exec rm {} \\; && sed -i '1s/^/\\/\\/ @ts-nocheck\\n/' generated/*.ts"
```
davemarco: I feel like we should simplify this script. I think you can just ignore the typescript in the tsconfig file instead of injecting the ignore. Also why are we deleting these other files? Why not just leave them?

davemarco: Also maybe we should add a short readme into the sql Parser folder reminding people to install `pip3 install antlr4-tools.` Since this is needed?

davemarco: Also it may make sense to put this script in the readme, instead of package.json, since it isnt actually using the npm dependency. Its not like it will be run often

components/webui/client/src/SqlParser/index.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Validate a SQL string for syntax errors.
+ *
+ * @param sqlString
+ * @throws {SyntaxError} with line, column, and message details if a syntax error is found.
+ */
+const validate = (sqlString: string) => {
+    const syntaxErrorListener = new SyntaxErrorListener();
+    const lexer = new SqlBaseLexer(new CharStream(sqlString.toUpperCase()));
```
davemarco: can u use new UpperCaseCharStream

components/webui/client/src/SqlParser/index.ts
```diff
@@ -0,0 +1,46 @@
...
+    const parser = new SqlBaseParser(new CommonTokenStream(lexer));
+    parser.removeErrorListeners();
+    parser.addErrorListener(syntaxErrorListener);
```
davemarco: Does the lexer need its own error listerner. Or the parser alone is enough?

components/webui/client/src/SqlParser/index.ts
```diff
@@ -0,0 +1,46 @@
...
+import SqlBaseLexer from "./generated/SqlBaseLexer";
+import SqlBaseParser from "./generated/SqlBaseParser";
+
+
+class SyntaxError extends Error {
+}
+
+class SyntaxErrorListener<TSymbol> extends ErrorListener<TSymbol> {
+    // eslint-disable-next-line max-params, class-methods-use-this
+    override syntaxError (
```
davemarco: what is point of override?

components/webui/client/src/pages/SearchPage/SearchControls/Presto/SqlSearchButton/RunButton/index.tsx
davemarco, line 1: I dont think we need any changes to this file. I'm fine having the server validate the syntax for the freeform query. I think we should do another PR after this one to validate the select, where, and orderby syntax for guided mode. I think this PR can just be setting up the infra for the next PR.

components/webui/client/eslint.config.mjs
```diff
@@ -9,6 +9,7 @@ const EslintConfig = [
         ignores: [
             "dist/",
             "node_modules/",
+            "src/SqlParser/generated",
```
davemarco: maybe call this sql-parser since there is no class

components/webui/client/src/pages/SearchPage/SearchControls/Presto/SqlSearchButton/RunButton/index.tsx
davemarco: Like add rules to presto grammar for select, where, orderby
select example below

```
+// New: entry point for SELECT box parsing
+standaloneSelectList
+    : selectItem (',' selectItem)* EOF
+    ;
```

davemarco (commented):
Okay setup looks great. Added my comments. I think we will need another PR after this to do individual parsing.

Last thing is I think we need to copy the Presto [NOTICE](https://github.com/prestodb/presto/blob/master/NOTICES) file into our repo per license. And optionally give atttribution in there. 

Im not that familiar with the licensing stuff so maybe we ask kirk

components/webui/client/src/pages/SearchPage/SearchControls/Presto/SqlSearchButton/RunButton/index.tsx
davemarco: In the next PR, not here

components/webui/client/package.json
hoophalab: I guess `"exclude": ["...."]` won't work in tsconfig. Is any other config I'm not aware of?
Deleting unused files are what clps do for kql files, and it also has a similar script in the taskfile.

I guess a proper solution is to move this solution to taskfile?

components/webui/client/src/SqlParser/index.ts
hoophalab: I guess uppercasecharstream is defined in presto-ui. It's not part of antlr's library. On the other hand, touppercase is in js library and implemented in cpp by v8. So it should be faster?

components/webui/client/src/SqlParser/index.ts
hoophalab: I guess this overrides a method in ErrorListener, and we enabled this flag https://www.typescriptlang.org/tsconfig/#noImplicitOverride

components/webui/client/src/SqlParser/index.ts
hoophalab: nice catch! I wasn't able to inject a lexer error but I guess there must be some way to casue a lexer error

NOTICES
kirkrodrigues, line 1: The license states that this should be called `NOTICE`.

NOTICES
```diff
@@ -0,0 +1,3 @@
+The code for the t-digest was originally authored by Ted Dunning
+
+Adrien Grand contributed the heart of the AVLTreeDigest (https://github.com/jpountz)
```
kirkrodrigues: Are we using `t-digest` or `AVLTreeDigest` anywhere in the code we've copied from Presto? If not, I don't think we need the lines in this file:

> ... any Derivative Works that You distribute must include a readable copy of the attribution notices contained within such NOTICE file, excluding those notices that do not pertain to any part of the Derivative Works

```suggestion
`components/webui/client/sql-parser/SqlBase.g4` was authored by the contributors of https://github.com/prestodb/presto
```

NOTICES
kirkrodrigues, line 1: In the PR description, we should use a permalink to indicate what version of the file we copied.

components/webui/client/package.json
davemarco: for the // @ts-nocheck, I think you can make a new tsconfig like `tsconfig/tsconfig.generated.json` and have it include   "include": [
    "../src/sql-parser/generated/**/*"
  ],
  
  and have `tsconfig/tsconfig.app.json` exclude it. Then have the main tsconfig look like this
  ```
    "references": [
    {
      "path": "./tsconfig/tsconfig.app.json"
    },
    {
      "path": "./tsconfig/tsconfig.node.json"
    },
    {
      "path": "./tsconfig/tsconfig.generated.json"
    }
  ]
}
```

File: components/webui/client/src/SqlParser/index.ts (hunk 2 above)
Comment (davemarco) [reply components/webui/client/src/SqlParser/index.ts]: I googled this and supposedly they have that fancy method on purpose. Like ur way will change some of the tokens like `SELECT * FROM users WHERE name = 'hello';`. In ur way the token for 'hello' will be modified. In theres, during lexing it will show up at 'HELLO', but the token in the parse tree will remain 'HELLO'. If u want u can double check this, but i imagine they would have also just used the python uppercase if it didnt matter

**Review by davemarco — DISMISSED** [review]
code gen looks good to me. @kirkrodrigues can you just double check the checksum stuff in the taskfile. I dont really know well how that works, and would prefer if someone familiar with the taskfiles checked.

@hoophalab 
i added comments on the ts-ignore, as well as the uppercase.

Last thing - I think we should create an issue about adding workflow in the future like clp-s

**Comment by davemarco** [comment]
title:
`feat(webui): Set up ANTLR tooling for future client-side SQL parsing; Move clp-s parser generation to codegen.yaml; Remove unused client package-lock.json.`

File: components/webui/client/src/SqlParser/index.ts (hunk 2 above)
Comment (hoophalab) [reply components/webui/client/src/SqlParser/index.ts]: I see, you are right. nice catch

**Comment by hoophalab** [comment]
> Last thing - I think we should create an issue about adding workflow in the future like clp-s

@coderabbitai can you create an issue on creating a workflow similiar to clp-s

**Comment by hoophalab** [comment]
@davemarco We find it hard to make tsconfig ignore checks for specific files. Let's use sed to add a no-check comment. I applied other suggestions. Ready for your review.
@kirkrodrigues Ready for your review on the taskfiles.

File: taskfiles/codegen.yaml [hunk 7]
```diff
@@ -0,0 +1,79 @@
...
+        cmd: |-
+          rm -rf "{{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/{{.ITEM}}/generated/"* \
+          && java -jar {{.G_ANTLR_JAR_FILE}} \
+            -Dlanguage=Cpp \
+            -no-listener \
+            -visitor \
+            -package clp_s::search::{{.ITEM}}::generated \
+            -o clp_s/search/{{.ITEM}}/generated \
+            -Xexact-output-dir \
+            clp_s/search/{{.ITEM}}/*.g4 \
```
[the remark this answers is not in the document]
hoophalab: antlr always creates the folder

taskfiles/codegen.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - "{{.SRC_DIR}}/SqlBase.g4"
+    dir: "{{.SRC_DIR}}"
+    cmds:
+      - "rm -rf '{{.OUTPUT_DIR}}'"
+      - |-
+        java -jar {{.G_ANTLR_JAR_FILE}} \
+          -Dlanguage=TypeScript \
+          -no-listener \
+          -Xexact-output-dir SqlBase.g4 \
+          -o generated
```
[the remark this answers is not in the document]
hoophalab: antlr always creates the folder

junhaoliao:
```suggestion
      - >-
        java -jar {{.G_ANTLR_JAR_FILE}}
        -Dlanguage=TypeScript
        -no-listener
        -o generated
        -Xexact-output-dir SqlBase.g4
```

taskfiles/codegen.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - |-
+        find generated \
+          -type f -and -not -name "*.ts" \
+          -exec rm {} \;
```
junhaoliao:
```suggestion
      - >-
        find generated
        -type f -and -not -name "*.ts"
        -exec rm {} \;
```

taskfiles/codegen.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - for: ["generated/SqlBaseLexer.ts", "generated/SqlBaseParser.ts"]
```
junhaoliao: we don't impose very strict guideline for YAML array styles, but since this is technically a file list:
```suggestion
      - for:
          - "generated/SqlBaseLexer.ts"
          - "generated/SqlBaseParser.ts"
```

taskfiles/codegen.yaml
```diff
@@ -0,0 +1,82 @@
...
+            -no-listener \
+            -visitor \
+            -package clp_s::search::{{.ITEM}}::generated \
+            -o clp_s/search/{{.ITEM}}/generated \
+            -Xexact-output-dir \
+            clp_s/search/{{.ITEM}}/*.g4 \
+            && find "{{.G_CORE_COMPONENT_DIR}}/src/clp_s/search/{{.ITEM}}/generated" \
+              -type f -and -not \( -name "*.cpp" -o -name "*.h" \) \
+              -exec rm {} \;
+
```
junhaoliao:
```suggestion
```

taskfiles/codegen.yaml
```diff
@@ -0,0 +1,82 @@
...
+      SRC_DIR: "{{.G_WEBUI_SRC_DIR}}/client/src/sql-parser"
+      OUTPUT_DIR: "{{.G_WEBUI_SRC_DIR}}/client/src/sql-parser/generated"
+    deps:
+      - ":deps:antlr-jar"
+      - task: ":utils:checksum:validate"
+        vars:
+          CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
+          INCLUDE_PATTERNS: ["{{.OUTPUT_DIR}}"]
+    generates: ["{{.CHECKSUM_FILE}}"]
+    sources:
```
junhaoliao, line 58: @kirkrodrigues Though not consistently seen in all task files, is it expected that we add
```
- "{{.TASKFILE}}"
```

(and also this for any non-main taskfiles)
```
- "{{.ROOT_DIR}}/taskfile.yaml"
```

to any public tasks?

kirkrodrigues: Should apply to all tasks, not just public ones, right?

junhaoliao: @coderabbitai please create an issue to track the effort to consistently include `"{{.TASKFILE}}"` and/or `"{{.ROOT_DIR}}/taskfile.yaml"` as sources for all public tasks

taskfiles/codegen.yaml
```diff
@@ -0,0 +1,87 @@
...
+      - "{{.SRC_DIR}}/SqlBase.g4"
+      - "{{.TASKFILE}}"
+      - "{{.ROOT_DIR}}/taskfile.yaml"
```
junhaoliao:
```suggestion
      - "{{.ROOT_DIR}}/taskfile.yaml"
      - "{{.SRC_DIR}}/SqlBase.g4"
      - "{{.TASKFILE}}"
```

junhaoliao (approved):
let's fix one nit then this is good to go

PR #1298, opened 2025-09-10
docs(clp-s): Add documentation for the lossless float representation feature.

gibber9809:
# Description

This PR splits off the documentation changes from #1176 to slightly ease the review burden. These docs changes comprise:
* Describing how the lossless encoding of floating point numbers works to developers
* Telling users how to use the feature

# Checklist

  breaking change.

# Validation performed

* Validated that the modified docs render correctly

## Summary by CodeRabbit

- New Features
  - Added an option to retain original floating-point textual representations during compression; encoder prefers a compact encoded form and falls back to full-string storage when needed.

- Limitations
  - Not supported for KV‑IR inputs; numeric comparisons use the nearest IEEE‑754 binary64 value and may lose textual precision at query time.

- Documentation
  - Added user guidance, CLI flag docs, and a new design document describing retention strategies and behaviour.

---
  - https://app.asana.com/0/0/1212852560849193


docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,122 @@
...
+`DictionaryFloat`.
+
+## High-Level `FormattedFloat` Specification  
+
+Each `FormattedFloat` node contains:
+
+- The double value in IEEE-754 binary64 format.
+- A 2-byte little-endian *format* field encoding the necessary output formatting information so
+  that, upon decompression, the value can be decompressed exactly to the original text.
+  Note that the unused lowest 5 bits of the 2‑byte field are currently reserved.
```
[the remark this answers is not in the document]
anlowee: Agree

anlowee:
```suggestion

  Note that the unused lowest 5 bits of the 2‑byte field are currently reserved.
```

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,122 @@
...
+  Encoders must write them as 0, and decoders must ignore them (treat as “don’t care”) for forward
```
anlowee: Maybe concatenate this sentence to the last one (Note...)

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,122 @@
...
+* For non-scientific numbers we accept:
+  * Any number that has at most 16 digits after the first non-zero digit
+  * Or at most 1 zero before the decimal and 16 zeroes after the decimal, if the number is a zero
+* For scientific numbers we accept:
+  * Single digit numbers with no decimal, followed by an exponent
+  * Or numbers with **1** digit preceding the decimal and up to 16 digits following the
+    decimal, followed by an exponent
+  * Where zero can not be the digit before the decimal, unless the number is a zero
+  * And where the exponent is specified by `e` or `E` optionally followed by `+` or `-`
+  * With at most **4** exponent digits, which must be left-padded with `0`
```
anlowee: Not english expert, but the "must" sounds like it must have some padded zeros.
```suggestion
  * With at most **4** exponent digits, which can be left-padded with `0`
```

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,122 @@
...
+
+- `123456789.1234567000` → **19** (from first `1` to last `0`)
+- `1.234567890E16` → **10** (from first `1` to `0` before exponent)
+- `0.000000123000` → **6** (from first `1` to last `0`)
+- `0.00` → **3** (counts all zeros for zero value)
+
+Per the [JSON grammar][json_grammar], the integer part of a floating point number cannot be empty,
+so the minimum number of digits is **1**. To take advantage of this fact, we store this field as
+**actual number of non-zero digits to end of number - 1**; for the numeric value zero we store
+**actual number of digits - 1**.
```
anlowee: do we want to also mention that by json grammar, the decimal part of a floating point number cannot be empty either? (e.g., `1.`)

docs/src/user-docs/core-clp-s.md
```diff
@@ -76,6 +79,21 @@ AWS_ACCESS_KEY_ID='...' AWS_SECRET_ACCESS_KEY='...' \
     /mnt/logs/log1.json
 ```
 
+:::{tip}
+Use the `--retain-float-format` flag during compression. Internally, switch to using a different
+encoding approach for floating point numbers that always retains their original formats. For
+example, values like `1.000e+00` or `0.000000012300` will be decompressed unchanged.
+:::
+
+**Enable retaining float numbers' formats:**
```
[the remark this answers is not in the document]
anlowee: Partially agree. But we should not use `-` to make the term consistent.

docs/src/user-docs/core-clp-s.md
```diff
@@ -154,6 +172,8 @@ compressed data:**
 * The order of log events is not preserved.
 * The input directory structure is not preserved and during decompression all files are written to
   the same file.
+* When using the `--retain-float-format` flag:
+  * KV-IR inputs currently don't support preserving the original printed float formats.
```
anlowee, line 158: Do we want to point to other limitations in the way similar to the search syntax limitations below?

anlowee:
Finished the first round of the review.

docs/src/user-docs/core-clp-s.md
gibber9809: I guess it's probably worth mentioning that we treat everything as the nearest representable double to the string value at query time, so if you have some input that doesn't actually correspond to a double then you might get slightly non-intuitive results

docs/src/dev-docs/design-retain-float-format.md
gibber9809: I think it's fine not to mention it in this case since all we're trying to explain here is why we take 1-17 digits. We already mention near the start that we require our inputs to match the json floating point grammar as one of the necessary conditions.

docs/src/dev-docs/design-retain-float-format.md
gibber9809: Yeah I think that that makes more sense.

anlowee (dismissed):

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,122 @@
...
+
+Note that the unused lowest 5 bits of the 2‑byte field are currently reserved, encoders must write
+them as 0, and decoders must ignore them (treat as “don’t care”) for forward compatibility.
+
+```text
++-------------------------------------+------------------------+--------------------------+------------------------------------------------------+-------------------+
+| Scientific Notation Marker (2 bits) | Exponent Sign (2 bits) | Exponent Digits (2 bits) | Digits from First Non-Zero to End of Number (5 bits) | Reserved (5 bits) |
++-------------------------------------+------------------------+--------------------------+------------------------------------------------------+-------------------+
+MSB                                                                                                                                                                LSB
+```
```
kirkrodrigues: Can we use a bullet list instead and link to each subsection?

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,122 @@
...
+
+This counts the digits from the first non-zero digit up to the last digit of the integer or
+fractional part (excluding the exponent). Examples:
+
+- `123456789.1234567000` → **19** (from first `1` to last `0`)
+- `1.234567890E16` → **10** (from first `1` to `0` before exponent)
+- `0.000000123000` → **6** (from first `1` to last `0`)
+- `0.00` → **3** (counts all zeros for zero value)
+
+Per the [JSON grammar][json_grammar], the integer part of a floating point number cannot be empty,
```
kirkrodrigues: I think it would be helpful to kind of summarize what the JSON grammar really supports and what it doesn't.

kirkrodrigues (commented):
Can we restructure the doc and add a bit more motivation? I'm thinking something like:

* Motivation
    * Explain that a float value with the same binary encoding can have multiple string encodings, and show some examples.
    * Explain that to support, truly lossless decompression, we want to store not only the float’s value but the format as well.
    * Explain the complexities (e.g., IEEE-754 bin64 has its own limitations but also some values are intractable to retain the format for, etc.)
* Solution
  * Two encodings
  * When FormattedFloat cannot represent losslessly, we fall back to DictionaryFloat
* Overview of what formats we can represent and examples
* Specification

Overall, we want this doc to be a bit more readable for new developers and/or interested users.

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
+# What does it take to losslessly retain JSON floating-point numbers?
```
quinntaylormitchell:
```suggestion
# Lossless JSON float storage
```
The title appears as the clickable link in the left sidebar (toctree) and it looks odd because it's too long

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+
+Since our goal is to losslessly retain floating-point numbers that come from JSON input, it is worth
+taking a look at what kinds of floating-point numbers can appear in JSON.
+
+The [JSON specification][json_spec] treats fields matching the following grammar as number values:
```
quinntaylormitchell:
```suggestion
The [JSON specification][json_spec] treats fields that match the following grammar as number values:
```
nit: clarity

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+
+frac = decimal-point ( digit1-9 | zero )+
+
+int = zero | ( digit1-9 ( digit1-9 | zero )* )
+
+minus = '-'
+
+plus = '+'
+
+zero = '0'
```
quinntaylormitchell:
```suggestion
number        = [ minus ], int, [ frac ], [ exp ]

int           = zero | ( digit1_9, { digit } )
frac          = decimal-point, digit, { digit }
exp           = ( 'e' | 'E' ), [ minus | plus ], digit, { digit }

digit         = zero | digit1_9
digit1_9      = '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9'

decimal-point = '.'
minus         = '-'
plus          = '+'
zero          = '0'
```
I think the formatting of this is slightly confusing because it's difficult to decipher where the really important information is; in addition, the dependencies between the different declarations are unclear. however, your original version more closely follows the formatting in the JSON specification, so it's up to you.

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+- NaN and +/- Infinity are not allowed
+- The exponent must contain at least one digit
+- The fractional part of a number must contain at least one digit
+- The integer part of a number can only start with '0' if the entire integer part is '0'
+- The integer part of a number must contain at least one digit
+- Positive numbers cannot begin with an explicit '+'
+
+What it doesn't do is place any restrictions on how a given floating-point number should be written,
+or whether floating-point numbers have to correspond to values from a standard such as IEEE-754
+binary64.
```
quinntaylormitchell:
```suggestion
There are two important points to note about this grammar:
1. It doesn't place any restrictions on how a given floating-point number _should_ be written
2. It doesn't dictate whether or not floating-point numbers have to correspond to values from a
   standard such as IEEE-754 binary64.
```
the "what it doesn't do" sounded a little awkward, and enumerating the points will make them easier to refer to later on

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+
+Since the first point is less abstract, we'll explain it first. Say we're trying to represent the
+number `16` as a floating-point number using **3** digits of precision. We might write it as:
```
quinntaylormitchell:
```suggestion
Since point 1. is less abstract than point 2., we'll explain it first. Say we're trying to represent
the number `16` as a floating-point number using **3** digits of precision. We might write it as:
```
refer to the (new, enumerated) points specifically

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+- `0.160e2`; and
+- `0.00000000160e10`
+
+are both valid representations of `16` using 3 significant digits.
+
+Likewise, we can come up with infinite representations of `16` by choosing to represent arbitrarily
+many significant digits.
+
+The point about whether the original values correspond to IEEE-754 is a bit abstract, but is
+important for understanding our approach to losslessly storing floating-point numbers.
```
quinntaylormitchell:
```suggestion
Point 2. is a bit more abstract, but it is important for understanding our approach to losslessly
storing floating-point numbers.
```
referring to point 2. directly

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+
+It is probably easiest to show an example. Of the numbers:
```
quinntaylormitchell:
```suggestion
It's most easily demonstrated with an example. Of the numbers:
```
nit: specificity

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+Note that for scientific-notation representations of a number, we can shift the decimal and change
+the exponent arbitrarily to represent the same number in infinite possible ways. For example:
+
+- `0.160e2`; and
+- `0.00000000160e10`
+
+are both valid representations of `16` using 3 significant digits.
+
+Likewise, we can come up with infinite representations of `16` by choosing to represent arbitrarily
+many significant digits.
```
quinntaylormitchell: even though this is technically correct (16 and 16.000 are the same value), the amount of significant digits is never arbitrary, right? so I'm not sure if the essence of this point is really true.

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+- For any given number there are many possible representations (infinitely many in fact); and
+- Not all floating-point numbers that are valid JSON correspond to values from a standard like
+  IEEE-754
+
+In practice though, we know that most of the time we should be dealing with very standard
+machine-generated data. This means that most inputs _do_ correspond to IEEE-754 binary64
+floating-point numbers, and that of the infinitely many ways of representing a number only a few
+will be common.
+
+Our approach then, is to store most floating-point numbers as an IEEE-754 binary64 floating-point
```
quinntaylormitchell:
```suggestion
Our approach, then, is to store most floating-point numbers as an IEEE-754 binary64 floating-point
```
nit: comma

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+number alongside some formatting information, falling back to storing the number as a string when
+that doesn't work.
+
+# Retaining floating-point format information
```
quinntaylormitchell:
```suggestion
## Retaining floating-point format information
```
there should only be one `#` level header

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+
+## High-level `FormattedFloat` specification
+
+Each `FormattedFloat` node contains:
+
+- The double value in IEEE-754 binary64 format.
+- A 2-byte little-endian _format_ field encoding the necessary output formatting information so
+  that, upon decompression, the value can be decompressed exactly to the original text.
+
+Note that the unused lowest 5 bits of the 2‑byte field are currently reserved, encoders must write
```
quinntaylormitchell:
```suggestion
Note that the unused lowest 5 bits of the 2‑byte field are currently reserved. Encoders must write
```
nit: flow

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+From MSB to LSB, the 2-byte format field contains the following sections:
+
+- [Scientific notation marker](#scientific-notation-marker) (2 bits)
+- [Exponent sign](#exponent-sign) (2 bits)
+- [Exponent digits](#exponent-digits) (2 bits)
+- [Digits from first non-zero to end of number](#digits-from-first-non-zero-to-end-of-number) (5 bits)
+- Reserved for future use (5 bits)
+
+To clarify the floating-point formats that `FormattedFloat` can represent, we describe them in text
+here:
```
quinntaylormitchell:
```suggestion
The floating-point formats that `FormattedFloat` can represent are described below:
```
this reads as more direct

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+
+- For non-scientific numbers we accept:
```
quinntaylormitchell:
```suggestion
- Numbers not written in scientific notation are accepted if **either** of the following are true:
```
if you didn't mean "scientific notation", just ignore this, although the "either of the following are true" applies to your pending acceptance of the next comment (so don't ignore that part)

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+  - Any number that has at most 16 digits after the first non-zero digit
+  - Or at most 1 zero before the decimal and 16 zeroes after the decimal, if the number is a zero
+- For scientific numbers we accept:
```
quinntaylormitchell:
```suggestion
- Numbers written in scientific notation are accepted if all **four** of the following are true:
```
same as above

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+- [Exponent digits](#exponent-digits) (2 bits)
+- [Digits from first non-zero to end of number](#digits-from-first-non-zero-to-end-of-number) (5 bits)
+- Reserved for future use (5 bits)
+
+To clarify the floating-point formats that `FormattedFloat` can represent, we describe them in text
+here:
+
+- For non-scientific numbers we accept:
+  - Any number that has at most 16 digits after the first non-zero digit
+  - Or at most 1 zero before the decimal and 16 zeroes after the decimal, if the number is a zero
```
quinntaylormitchell:
```suggestion
  1. Any non-zero number with at most **17** digits starting at the first non-zero digit
  2. A zero written with at most **one** zero before the decimal point and at most **16** zeros
     after the decimal point
```
I think this works better as a numbered list; improved clarity

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+- For scientific numbers we accept:
+  - Single digit numbers with no decimal, followed by an exponent
+  - Or numbers with **1** digit preceding the decimal and up to 16 digits following the
+    decimal, followed by an exponent
+  - Where zero can not be the digit before the decimal, unless every digit in the number is zero
+  - And where the exponent is specified by `e` or `E` optionally followed by `+` or `-`
+  - With at most **4** exponent digits, which can be left-padded with `0`
```
quinntaylormitchell:
```suggestion
  1. The significand is either
      * a **single** digit with no decimal point, followed by an exponent
      * **one** digit before the decimal point, and up to **16** digits after the decimal point,
        followed by an exponent
  2. The digit before the decimal point cannot be zero unless every digit of the significand is zero
  3. The exponent uses `e` or `E`, optionally followed by `+` or `-`
  4. The exponent has at most **4** digits (left-padding with `0` is allowed)
```
this was getting a bit muddled; I hope this clears it up without changing the meaning

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+
+With the added restrictions that:
+
+- The floating-point number follows the JSON grammar for floating-point numbers.
+- There exists an IEEE-754 binary64 number for which the string is the closest decimal
+  representation at the given precision.
+
+These restrictions really correspond to "canonical" representations of floating-point numbers with
+up to 17 digits of precision. This means that our formatting scheme can always represent numbers
+produced by format specifiers such as '%f', '%e', and '%g', so long as they don't use too many
```
quinntaylormitchell:
```suggestion
produced by format specifiers such as `%f`, `%e`, and `%g`, so long as they don't use too many
```
nit: monospace

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+- `10`: `-`
+
+`11` is unused by the current implementation.
+
+For example, exponents of `0` may appear as `0`, `+0`, or `-0`, and these two bits can record the
+format correctly.
+
+### Exponent digits
+
+Since the maximum and minimum decimal exponents for a double, `308` and `-324` respectively, are
```
quinntaylormitchell:
```suggestion
Since the maximum and minimum decimal exponents for a double (`308` and `-324` respectively) are
```
nit: flow

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+both three digits, two bits are enough to represent the digit count. We allow up to 4 digits to
+support exponents left-padded with `0`.
+
+The stored value is **actual digits − 1**, since there is always at least one digit
+(e.g., `00` → 1 digit). The two-bit mapping is:
```
quinntaylormitchell:
```suggestion
The stored value is **one less than the number of actual digits**, since there is always at least
one digit (e.g., `00` → 1 digit). The two-bit mapping is:
```
clarity

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+
+This counts the digits from the first non-zero digit up to the last digit of the integer or
+fractional part (excluding the exponent). Examples:
+
+- `123456789.1234567000` → **19** (from first `1` to last `0`)
+- `1.234567890E16` → **10** (from first `1` to `0` before exponent)
+- `0.000000123000` → **6** (from first `1` to last `0`)
+- `0.00` → **3** (counts all zeros for zero value)
+
+Per the [JSON specification][json_grammar], the integer part of a floating-point number cannot be
```
quinntaylormitchell:
```suggestion
Per the [JSON specification][json_spec], the integer part of a floating-point number cannot be
```
link didn't exist

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+empty, so the minimum number of digits is **1**. To take advantage of this fact, we store this field
+as **actual number of non-zero digits to end of number - 1**; for the numeric value zero we store
+**actual number of digits - 1**.
```
quinntaylormitchell:
```suggestion
as **one less than the actual number of non-zero digits**; for the numeric value zero, we store **1**.
```
I think I grasped this and I hope the suggestion does not compromise the meaning! if it does, ignore

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+
+As well, according to IEEE-754, only 17 decimal significant digits are needed to represent all
+binary64 floating-point numbers without precision loss. As a result, we currently allow a maximum of
+**17 digits**. Because the stored value is **digits - 1** the maximum encoded value is 16, which
+requires 5 bits.
+
+We could support representing binary64 numbers with up to 32 significant digits, and we may choose
+to do so in the future, but this is explicitly not supported in the current version of the format.
+The rationale for not doing so now is that as the number of digits increases beyond 17, the
+likelihood that the number corresponds to a valid IEEE-754 binary64 float decreases.
```
quinntaylormitchell:
```suggestion
The rationale for not doing so now is that the likelihood that a number corresponds to a valid
IEEE-754 binary64 float decreases as the number of digits increases beyond 17.
```
flow

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,228 @@
...
+number alongside its IEEE-754 binary64 representation.
+- `DictionaryFloat`: we store the full string representation of the floating-point number in the
+variable dictionary, and encode numbers as their corresponding variable dictionary IDs.
+
+Generally we prefer storing floating-point numbers as `FormattedFloat` over `DictionaryFloat`
+because:
+
+- We can directly compare against the stored IEEE-754 binary64 float at query time instead of having
+to first parse the string representation of a floating-point number.
+- We avoid bloating the variable dictionary with non-repetitive floating-point strings.
```
quinntaylormitchell:
```suggestion
- `FormattedFloat` allows us to directly compare against the stored IEEE-754 binary64 float at query
  time instead of having to first parse the string representation of a floating-point number.
- Fewer `DictionaryFloat` encodings means that we can avoid bloating the variable dictionary with
  non-repetitive floating-point strings.
```
specificity

quinntaylormitchell (changes requested):
it's looking really good! easy to navigate, and informative without being overwhelming.

a number of comments to address; let me know if you need clarification.

high-level nit: I think that unordered lists should use `*` as the bullet instead of `-`; I didn't bother commenting every time, as everything renders correctly.

docs/src/dev-docs/design-retain-float-format.md
gibber9809: I'm using the meaning of "arbitrarily many" here that refers to the fact that we can make arbitrarily large choices tending to infinity. That meaning is pretty common in mathematics, but if you think its too unclear we can rephrase.

docs/src/dev-docs/design-retain-float-format.md
gibber9809: Directly edited your comment for capitalization, but I agree this is more clear

docs/src/dev-docs/design-retain-float-format.md
gibber9809: Directly edited your suggestion a bit since it should be "17 digits starting at" since that statement is inclusive vs the previous "16 digits after" which was exclusive. Besides that, I agree this is more clear.

docs/src/dev-docs/design-retain-float-format.md
gibber9809: Yeah the meaning is a little different and it isn't conveyed that well by the current text. I'll think about it a bit and make some edits here.

docs/src/dev-docs/design-retain-float-format.md
gibber9809: I do like some parts of how you've reorganized the grammar, so I'll make some changes, but I think I still want to keep it generally similar to the grammar in the JSON spec.

quinntaylormitchell: I like how you redid it, good good

docs/src/dev-docs/design-retain-float-format.md
quinntaylormitchell: I like your edits, good

docs/src/dev-docs/design-retain-float-format.md
```diff
@@ -0,0 +1,226 @@
...
+Note that for scientific-notation representations of a number, we can shift the decimal and change
+the exponent arbitrarily to represent the same number in infinite possible ways. For example:
+
+- `0.160e2`; and
+- `0.00000000160e10`
+
+are both valid representations of `16` using 3 significant digits.
+
+Likewise, we can come up with infinite representations of `16` by choosing to represent arbitrarily
+many significant digits.
```
quinntaylormitchell:
```suggestion
Likewise, we can come up with infinite representations of `16` (at different precisions) by
choosing to represent arbitrarily many significant digits.
```
as per our convo offline

quinntaylormitchell (dismissed):
other than the comment we discussed offline, lgtm!

quinntaylormitchell (dismissed):

PR #1299, opened 2025-09-11
feat(clp-package): Support the use of `decompress.sh` when using `clp-json`; document the usage of `decompress.sh` in the quick-start docs.

quinntaylormitchell:
# Description

This PR introduces support for decompression when using the `clp-json` package. Users can perform the decompression by using `decompress.sh` script:

```bash
./sbin/decompress.sh x [-d DIR] --dataset 'dataset-name'
```

where `'dataset-name'` (mandatory) is the name of the compressed dataset they would like to decompress (must be stored in `clp-package/var/data/archives`), and DIR (optional; default is `.`) is the path to the directory that they would like to store the uncompressed logs in.

The command uses `clp-s` under the hood, so decompression is subject to the same [restrictions and limitations](https://docs.yscope.com/clp/main/user-docs/core-clp-s.html#current-limitations) as exists when using the `clp-s` binary directly.

This PR also adds an explanation of how to use the `decompress.sh` command to both the `clp-text` and `clp-json` pages of the quick-start guide in the CLP docs.

# Checklist

  breaking change.

# Validation performed

Built the package, compressed a dataset, and decompressed it with `./sbin/decompress.sh`; the original and decompressed datasets were identical.

## Summary by CodeRabbit

* **New Features**
  * Extraction now branches by storage engine: one engine requires a mandatory --dataset while the other accepts explicit paths or a --files-from list. Added a --dataset option, storage-engine-specific validations, clearer error/success messages, and ensured temporary config cleanup on failures.

* **Documentation**
  * Added "Decompressing" subsections for JSON and text logs; new sections were duplicated in two places.

---
  - https://app.asana.com/0/0/1212888465368610


components/clp-package-utils/clp_package_utils/scripts/decompress.py
```diff
@@ -144,13 +135,42 @@ def handle_extract_file_cmd(
...
+        if parsed_args.dataset is None:
+            logger.error(
+                f"You must specify a dataset to decompress using the --dataset flag when using the "
+                f"{storage_engine} storage engine."
+            )
+            return -1
+        extract_cmd.extend(["--dataset", parsed_args.dataset])
+    else:
+        logger.error(f"Unsupported storage engine: {storage_engine}")
+        return -1
```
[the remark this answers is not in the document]
quinntaylormitchell: this section checks if the --dataset flag is used, not whether the dataset exists. I feel it is more appropriate to check that the dataset exists in the native script (as it currently is)

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
```diff
@@ -278,6 +279,69 @@ def handle_extract_file_cmd(
...
+    extraction_dir = pathlib.Path(parsed_args.extraction_dir)
+    if not extraction_dir.is_dir():
+        logger.error(f"extraction-dir ({extraction_dir}) is not a valid directory.")
+        return -1
+
+    target_dataset: str = parsed_args.dataset
+    if target_dataset is None:
+        logger.error(f"You must specify --dataset when using the {storage_engine} storage engine.")
+        return -1
+    archives_dir = clp_config.archive_output.get_directory()
```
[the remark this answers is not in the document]
quinntaylormitchell: extraction_dir already gets created in the non-native script, and the no paths check is performed there too.

docs/src/user-docs/quick-start/clp-json.md
```diff
@@ -180,6 +180,23 @@ searches are case-**sensitive** on the command line.
...
+
+- `DIR` is the path to the directory where you'd like the decompressed files to be stored. This
+  argument is optional; the default path is `.`.
+- `'dataset-name'` is the name of the dataset you'd like to decompress. This argument is mandatory.
+
+The decompression script will return you to the command line when your decompression job is
+finished.
+
+---
+
```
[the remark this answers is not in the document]
quinntaylormitchell: as this is a quick-start guide, I think it's reasonable to assume that the `clp-json` user won't know the `-f` flag even exists

docs/src/user-docs/quick-start/clp-json.md
```diff
@@ -180,6 +180,24 @@ searches are case-**sensitive** on the command line.
...
+- `x` selects the “extract” action. This argument is mandatory.
+- `DIR` is the path to the directory where you'd like the decompressed files to be stored. This
+  argument is optional; the default path is `.`.
+- `'dataset-name'` is the name of the dataset you'd like to decompress. This argument is mandatory.
+
+The decompression script will return you to the command line when your decompression job is
+finished.
+
+---
```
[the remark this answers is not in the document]
quinntaylormitchell: examples aren't provided for the other commands on the page, so I won't here either.

components/clp-package-utils/clp_package_utils/scripts/decompress.py
```diff
@@ -144,13 +135,42 @@
...
+            logger.error(
+                f"You cannot use the --dataset flag when using the {storage_engine} storage engine."
+            )
+            return -1
+        for path in parsed_args.paths:
+            extract_cmd.append(path)
+        if container_paths_to_extract_file_path:
+            extract_cmd.append("--files-from")
+            extract_cmd.append(container_paths_to_extract_file_path)
+    elif StorageEngine.CLP_S == storage_engine:
```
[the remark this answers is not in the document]
quinntaylormitchell: I'm just going to leave it like it was

components/clp-package-utils/clp_package_utils/scripts/decompress.py
```diff
@@ -144,21 +135,62 @@ def handle_extract_file_cmd(
...
-        extract_cmd.append(path)
-    if container_paths_to_extract_file_path:
-        extract_cmd.append("--input-list")
-        extract_cmd.append(container_paths_to_extract_file_path)
+
+    if StorageEngine.CLP == storage_engine:
+        # Use either file list or explicit paths; prohibit --dataset flag
+        if parsed_args.dataset is not None:
+            logger.error(
+                f"You cannot use the --dataset flag when using the {storage_engine} storage engine."
```
gibber9809:
```suggestion
                f"The --dataset flag cannot be used with the {storage_engine} storage engine."
```
"You ..." phrasing feels a bit unusual for application error messages. I've given suggestions here and a few other places to rephrase to a more imperative form.

components/clp-package-utils/clp_package_utils/scripts/decompress.py
```diff
@@ -144,21 +135,62 @@ def handle_extract_file_cmd(
...
+            extract_cmd.append(path)
+        if container_paths_to_extract_file_path:
+            extract_cmd.append("--files-from")
+            extract_cmd.append(container_paths_to_extract_file_path)
+    elif StorageEngine.CLP_S == storage_engine:
+        # Require --dataset flag; prohibit both file list and explicit paths
+        if parsed_args.files_from or parsed_args.paths:
+            logger.error(
+                f"You can't specify individual file paths or use the -f flag when decompressing "
+                f"with the {storage_engine} storage engine."
```
gibber9809:
```suggestion
                f"File paths cannot be specified when decompressing with the"
                f" {storage_engine} storage engine."
```

components/clp-package-utils/clp_package_utils/scripts/decompress.py
```diff
@@ -144,21 +135,62 @@ def handle_extract_file_cmd(
...
+            )
+            # Remove generated files
+            generated_config_path_on_host.unlink()
+            return -1
+        if parsed_args.dataset is None:
+            logger.error(
+                f"You must specify a dataset to decompress using the --dataset flag when using the "
+                f"{storage_engine} storage engine."
```
gibber9809:
```suggestion
                f"A dataset must be specified for decompression (using the --dataset flag) when using the"
                f" {storage_engine} storage engine."
```

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
```diff
@@ -278,6 +280,78 @@ def handle_extract_file_cmd(
...
+
+    # Validate extraction directory
+    extraction_dir = pathlib.Path(parsed_args.extraction_dir)
+    if not extraction_dir.is_dir():
+        logger.error(f"extraction-dir ({extraction_dir}) is not a valid directory.")
+        return -1
+
+    target_dataset: str = parsed_args.dataset
+    if target_dataset is None:
+        logger.error(f"You must specify --dataset when using the {storage_engine} storage engine.")
```
gibber9809:
```suggestion
        logger.error(f"The --dataset flag must be specified when using the {storage_engine} storage engine.")
```

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
```diff
@@ -278,6 +280,78 @@ def handle_extract_file_cmd(
...
+        logger.error(e)
+        return -1
+
+    # Construct dataset_dir path and validate
+    archives_dir = clp_config.archive_output.get_directory()
+    dataset_dir = archives_dir / target_dataset
+    if not dataset_dir.exists():
+        logger.error(
+            f"Dataset '{target_dataset}' not found in {archives_dir}. Are you sure it was "
+            "compressed?"
```
gibber9809:
```suggestion
            f"Dataset '{target_dataset}' not found in {archives_dir}."
```

Maybe just this? If we want to tell them to check if they compressed anything to the dataset maybe we could write it more like `Note: a dataset is only created in the package after at least one file is ingested into it.`

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
```diff
@@ -278,6 +280,78 @@ def handle_extract_file_cmd(
...
+        )
+        return -1
+
+    # fmt: off
+    extract_cmd = [
+        str(clp_home / "bin" / "clp-s"),
+        "x", str(dataset_dir), str(extraction_dir),
+    ]
```
gibber9809: Do we want to use the `--ordered` flag? I know it will end up creating a lot of of files instead of appending everything to the same file, but even still it might make the output more useful.

I'd also be fine with keeping this as-is for now, but worth considering.

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
```diff
@@ -278,6 +280,78 @@ def handle_extract_file_cmd(
...
+    # fmt: on
+
+    # Configure CLP metadata DB connection params.
+    extract_env = {
+        **os.environ,
+        CLP_DB_USER_ENV_VAR_NAME: clp_db_connection_params["username"],
+        CLP_DB_PASS_ENV_VAR_NAME: clp_db_connection_params["password"],
+    }
```
gibber9809:
```suggestion
```
clp-s currently doesn't use these environment variables.

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
```diff
@@ -278,6 +280,78 @@ def handle_extract_file_cmd(
...
+
+    proc = subprocess.Popen(extract_cmd, env=extract_env)
+    return_code = proc.wait()
+    if 0 != return_code:
+        logger.error(f"File extraction failed, return_code={return_code}")
+        return return_code
+
+    # logger.info(f"File extraction successful. Decompressed file written to {str(extraction_dir)}")
+
```
gibber9809:
```suggestion
```

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
```diff
@@ -278,6 +280,78 @@ def handle_extract_file_cmd(
...
+        CLP_DB_USER_ENV_VAR_NAME: clp_db_connection_params["username"],
+        CLP_DB_PASS_ENV_VAR_NAME: clp_db_connection_params["password"],
+    }
+
+    proc = subprocess.Popen(extract_cmd, env=extract_env)
+    return_code = proc.wait()
+    if 0 != return_code:
+        logger.error(f"File extraction failed, return_code={return_code}")
+        return return_code
+
```
gibber9809:
```suggestion
```
Feels like unnecessary whitespace.

gibber9809 (changes requested):
Nice work! Mostly looks good to me. My main high-level question is whether you think its worthwhile to add support for file-extraction from S3 to the clp-json in this PR?

To my memory, the only reason we explicitly disallow using S3 during file extraction for the clp-text package is that clp-text currently doesn't support S3 -- i.e. for clp-json the restriction is somewhat artificial.

We would need to change the file extraction code on the clp-json path to get the S3 archive paths for a dataset from the dataset, so it would add some complexity. If you think that that would be an easy thing to add to this PR I think it would be nice to include for completeness, but if not we can leave it for follow-up work.

components/clp-package-utils/clp_package_utils/scripts/decompress.py
quinntaylormitchell: completely agree, I will change it to passive voice like you suggest

components/clp-package-utils/clp_package_utils/scripts/decompress.py
quinntaylormitchell: agreed

components/clp-package-utils/clp_package_utils/scripts/decompress.py
quinntaylormitchell: agreed

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
quinntaylormitchell: agreed

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
quinntaylormitchell: agreed; your suggested change will be sufficient

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
quinntaylormitchell: I tried it with and without the `--ordered` flag when I was testing, and I found that when I used the flag, the output (decompressed) file gets named some long string (e.g., `6fbede15-b969-457e-bd50-57779cb88e0c_0_1000000.jsonl`); when I don't use the flag, the output file gets the same name as the dataset that's being decompressed (e.g., `postgresql`). I thought that the latter would be easier to understand for the user.

that being said, I don't think I fully understand what the `--ordered` flag is even doing; I'll leave it as-is for now, but I'll dm you for more info on what the flag is doing.

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
quinntaylormitchell: ah good to know, thank you!

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
quinntaylormitchell: missed that, thanks

components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
quinntaylormitchell: agreed

quinntaylormitchell (commented):
>support for file-extraction from S3 to the clp-json in this PR

@gibber9809 I think that we should leave this for follow-up work, the main reason being that I need to get more familiar with S3. I will dm you to talk it over!

components/clp-package-utils/clp_package_utils/scripts/decompress.py
```diff
@@ -91,15 +91,6 @@ def handle_extract_file_cmd(
...
 
-    storage_type = clp_config.archive_output.storage.type
-    storage_engine = clp_config.package.storage_engine
-    if StorageType.FS != storage_type or StorageEngine.CLP != storage_engine:
-        logger.error(
-            f"File extraction is not supported for archive storage type `{storage_type}` with"
-            f" storage engine `{storage_engine}`."
-        )
-        return -1
-
```
gibber9809: Sorry, missed this in the first round -- we should keep this error message and just remove the `or StorageEngine.CLP != storage_engine` condition. We should still notify users that the S3 storage type doesn't have support for file extraction.

components/clp-package-utils/clp_package_utils/scripts/decompress.py
```diff
@@ -144,21 +135,62 @@ def handle_extract_file_cmd(
         "-d", str(container_extraction_dir),
     ]
     # fmt: on
+
+    storage_engine = clp_config.package.storage_engine
+
```
gibber9809:
```suggestion
```
We can delete this since we're adding back the deleted code for the error log above.

gibber9809 (changes requested):
Sorry -- a few small things I missed in the first round. Otherwise LGTM.

components/clp-package-utils/clp_package_utils/scripts/decompress.py
quinntaylormitchell: quite right, good catch

gibber9809 (dismissed):
LGTM. For PR title I think we just need to add a period:
```
feat(clp-package): Support the use of `decompress.sh` when using `clp-json`; document the usage of `decompress.sh` in the quick-start docs.
```

components/clp-package-utils/clp_package_utils/scripts/decompress.py
```diff
@@ -144,21 +144,60 @@ def handle_extract_file_cmd(
...
+        return -1
 
     cmd = container_start_cmd + extract_cmd
 
     proc = subprocess.run(cmd)
     ret_code = proc.returncode
     if 0 != ret_code:
-        logger.error("file extraction failed.")
+        logger.error("File extraction failed.")
         logger.debug(f"Docker command failed: {shlex.join(cmd)}")
```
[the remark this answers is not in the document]
quinntaylormitchell: this is outside the scope of the pr.

PR #1304, opened 2025-09-16
feat(webui): Add support for building queries from guided SQL input boxes.

hoophalab:
# Description

1. Add a `buildSqlQuery` function to the `sql-parser` that takes SQL query component strings as arguments and returns a valid SQL query.
2. Add testing input boxes. @davemarco said he will replace those testing input boxes with real ones when the UI is done.

sqlbuilder

# Checklist

  breaking change.

# Validation performed

1. Run a few SQL queries.
2. The generated SQL queries execute correctly. The webui shows the results from Presto.

## Summary by CodeRabbit

- **New Features**
  - Added an experimental SQL testing UI in guided Presto mode to compose SELECT, FROM, WHERE, ORDER BY, LIMIT, log the generated SQL and submit queries.
  - Exposed a new buildSearchQuery utility for assembling and validating generated SQL before execution; validation failures show contextual errors.

- **Refactor**
  - Separated parser creation from validation to improve parsing reliability and support guided query features.


components/webui/client/src/sql-parser/index.ts
```diff
@@ -46,24 +57,170 @@
...
+                // eslint-disable-next-line new-cap
+                children.push(ctx.ORDER(), ctx.BY(), sortItemList);
+            }
+            if (null !== limitValue) {
+                // eslint-disable-next-line new-cap
+                children.push(ctx.LIMIT(), limitValue);
+            }
+            ctx.children = children;
+        };
+    }
```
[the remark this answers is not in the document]
hoophalab: There seems no easy way to construct a parsing tree w/o splice sub parsing trees from different parser instances, unless we implement all syntax tree facilities like [here](https://github.com/y-scope/clp/tree/main/components/core/src/clp_s/search/ast), which is an overkill for what we are going to do in webui.

components/webui/client/src/pages/SearchPage/SearchControls/index.tsx
```diff
@@ -37,9 +38,12 @@ const SearchControls = () => {
     }
 
     return (
-        <form onSubmit={handleSubmit}>
-            {controls}
-        </form>
+        <>
+            <form onSubmit={handleSubmit}>
```
davemarco, line 42: Move this to Guided Controls

components/webui/client/src/sql-parser/SqlBase.g4
davemarco, line 1: You may need to modify top of file per license

davemarco (commented):
Great that it's works!

Overall i have a high level questions about this comment from the rabbit - https://github.com/y-scope/clp/pull/1304#discussion_r2350389800

What is wrong with this approach, it seems more intuitive and simpler than the template approach? There is also no setting children directly. 

Open to template approach, if you have good reason for it. 

Also i think you should add more comments to parser code.

components/webui/client/src/pages/SearchPage/SearchControls/index.tsx
hoophalab: We cannot because the testing component uses form and a form cannot be nested in another form. Since it's temporarily, I guess it's ok.

hoophalab:
> Overall i have a high level questions about this comment from the rabbit - https://github.com/y-scope/clp/pull/1304#discussion_r2350389800
> What is wrong with this approach, it seems more intuitive and simpler than the template approach? There is also no setting children directly.
> Open to template approach, if you have good reason for it.

I guess both works for our use but template might be more structured? Ideally, we should be able to construct syntax trees directly using components, (e.g. `new Query(new QuerySpecification(columns, relations), booleanExpr)`), but unfortunately ANTLR doesn't support doing this on a parsing tree. So the template is a workaround.

I'm open to both. @junhaoliao

junhaoliao:
> > Overall i have a high level questions about this comment from the rabbit - [#1304 (comment)](https://github.com/y-scope/clp/pull/1304#discussion_r2350389800)
> > What is wrong with this approach, it seems more intuitive and simpler than the template approach? There is also no setting children directly.
> > Open to template approach, if you have good reason for it.
> 
> I guess both works for our use but template might be more structured? Ideally, we should be able to construct syntax trees directly using components, (e.g. `new Query(new QuerySpecification(columns, relations), booleanExpr)`), but unfortunately ANTLR doesn't support doing this on a parsing tree. So the template is a workaround.
> 
> I'm open to both. @junhaoliao

Let's pull in some ANTLR experts for this discussion. @gibber9809 @LinZhihao-723 @kirkrodrigues 

From the solutions that were presented (templating vs string assembly), I believe the templating approach is easier to maintain. if we hardcode with string assembly method, we lost the AST abstraction. the code might become more complex if the query grammar grows (e.g., GROUP BY, HAVING, etc..)

it does seem that building fragment ASTs for `SELECT`, `FROM`, `WHERE`, etc can be a better approach, though rejected at https://github.com/y-scope/clp/pull/1304#discussion_r2350410187

by the way, do you think this can be a better interface than buildSearchQuery(...)?:
```
const sql = queryBuilder()
  .select(selectCtx)
  .from(fromCtx)
  .where(optionalWhereCtx)
  .orderBy(optionalOrderCtx)
  .limit(optionalLimitCtx)
  .toSQL();
```
or can we mimic the interfaces in this library https://www.npmjs.com/package/node-sql-parser ?

LinZhihao-723:
> > > Overall i have a high level questions about this comment from the rabbit - [#1304 (comment)](https://github.com/y-scope/clp/pull/1304#discussion_r2350389800)
> > > What is wrong with this approach, it seems more intuitive and simpler than the template approach? There is also no setting children directly.
> > > Open to template approach, if you have good reason for it.
> > 
> > 
> > I guess both works for our use but template might be more structured? Ideally, we should be able to construct syntax trees directly using components, (e.g. `new Query(new QuerySpecification(columns, relations), booleanExpr)`), but unfortunately ANTLR doesn't support doing this on a parsing tree. So the template is a workaround.
> > I'm open to both. @junhaoliao
> 
> Let's pull in some ANTLR experts for this discussion. @gibber9809 @LinZhihao-723 @kirkrodrigues
> 
> From the solutions that were presented (templating vs string assembly), I believe the templating approach is easier to maintain. if we hardcode with string assembly method, we lost the AST abstraction. the code might become more complex if the query grammar grows (e.g., GROUP BY, HAVING, etc..)
> 
> it does seem that building fragment ASTs for `SELECT`, `FROM`, `WHERE`, etc can be a better approach, though rejected at [#1304 (comment)](https://github.com/y-scope/clp/pull/1304#discussion_r2350410187)
> 
> by the way, do you think this can be a better interface than buildSearchQuery(...)?:
> 
> ```
> const sql = queryBuilder()
>   .select(selectCtx)
>   .from(fromCtx)
>   .where(optionalWhereCtx)
>   .orderBy(optionalOrderCtx)
>   .limit(optionalLimitCtx)
>   .toSQL();
> ```
> 
> or can we mimic the interfaces in this library https://www.npmjs.com/package/node-sql-parser ?

I'm not sure if I fully understand what Marco proposed, but I think the correct way to do it here is:
* Use a format string to construct a complete SQL query.
* Formally define the grammar to parse the query as a single tree.

Q: Why parsing a complete SQL query string?
A: A few advantages:
* For completeness: the parsing rule will be a complete subset of SQL query (but not for its components).
* Grammar-level validation: the validation/parsing is now at the grammar level, end-to-end from a string to the parsed tree.
* Future proving:
  * Easier to extend the grammar to support more SQL features.
  * Easier to extend so that users can input a complete SQL query string.
  
Q: Does it mean we need to maintain the format string carefully to concat subcomponents/clusters into a complete SQL query string?
A: Yes. And it's not duplicated work.
* The grammar should define the standard ordering for main clusters (like `SELECT` before `FROM` before `WHERE`...).
* Before we support arbitrary SQL query inputs, we carefully construct the query from clusters to follow the standard, which guarantees that the standard ordering is followed.
* After adding support for arbitrary SQL query inputs, the grammar is still the checker to ensure user inputs follow the standard cluster ordering.

davemarco:
@LinZhihao-723 I think good approach, but how do we send syntax error to user? I guess the errors would be for the full format string, and not the individual input, which may be fine?

LinZhihao-723:
> @LinZhihao-723 I think good approach, but how do we send syntax error to user? I guess the errors would be for the full format string, and not the individual input, which may be fine?

Yeah. Maybe we can do a preview of the generated query string in the UI. And reporting errors to that generated query string would make more sense.

hoophalab:
1. @junhaoliao I had another offline discussion with Zhihao and we settled with this very simple format + validation approach.

2. 
> by the way, do you think this can be a better interface than buildSearchQuery(...)?:

tbh, I don't really like Java-style builders -- they add many boilerplates to the code. I made a few changes using optional fields. Can you do a review again?

components/webui/client/src/sql-parser/index.ts
```diff
@@ -46,24 +46,86 @@
...
+    if ("undefined" !== typeof booleanExpression) {
+        sqlString += ` WHERE ${booleanExpression}`;
+    }
+    if ("undefined" !== typeof sortItemList) {
+        sqlString += ` ORDER BY ${sortItemList}`;
+    }
+    if ("undefined" !== typeof limitValue) {
+        sqlString += ` LIMIT ${limitValue}`;
+    }
+
```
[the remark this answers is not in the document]
hoophalab: the devs shouldn't pass an empty/trim-empty string. They should simply omit the value

davemarco:
> Yeah. Maybe we can do a preview of the generated query string in the UI. And reporting errors to that generated query string would make more sense.

Okay i think eventually we can change the error so its shows the full format string

hoophalab:
> > Yeah. Maybe we can do a preview of the generated query string in the UI. And reporting errors to that generated query string would make more sense.
> 
> Okay i think eventually we can change the error so its shows the full format string

Or we can still add standaloneXXX back and validate individual components before formatting?

components/webui/client/src/pages/SearchPage/SearchControls/Presto/BuildSqlTestingInputs.tsx
```diff
@@ -0,0 +1,75 @@
...
+    Type,
+} from "@sinclair/typebox";
+import {Value} from "@sinclair/typebox/value";
+
+import {buildSearchQuery} from "../../../../sql-parser";
+import {handlePrestoQuerySubmit} from "./presto-search-requests";
+
+
+// eslint-disable-next-line no-warning-comments
+// TODO: Replace this with actual SQL inputs
```
davemarco:
```suggestion
// TODO: Replace this file with actual SQL inputs
```

components/webui/client/src/pages/SearchPage/SearchControls/index.tsx
davemarco: I think this is okay, since it still wont render without guided which is default false

components/webui/client/src/sql-parser/index.ts
```diff
@@ -46,24 +46,85 @@ class UpperCaseCharStream extends CharStream {
...
+ * Validate a SQL string for syntax errors.
+ *
+ * @param sqlString
+ * @throws {SyntaxError} with line, column, and message details if a syntax error is found.
+ */
+const validate = (sqlString: string) => {
+    buildParser(sqlString).singleStatement();
+};
+
+interface BuildSearchQueryProps {
```
davemarco, line 77: move to typings.ts?

davemarco: we could also maybe move the ascii types

davemarco (dismissed):
Some small comments. I am ready to approve after this

davemarco:
@hoophalab 
> Or we can still add standaloneXXX back and validate individual components before formatting?

if the errors look weird in testing, maybe we add this as a precheck in a later PR?

components/webui/client/src/sql-parser/index.ts
hoophalab: We usually keep function/component props interface in the original file?

PR #1306, opened 2025-09-16
fix(webui): Prevent the presto progress bar from overflowing its parent (fixes #1305).

### LINKED ISSUE #1305: Progress bar overflows window width
A progress bar overflow issue has been reported where the progress bar extends beyond the window width when it appears.

**Problem Description:**
The progress bar UI element overflows the window width, causing a horizontal layout issue.

**Context:**
- Reported during review of PR #1248 (https://github.com/y-scope/clp/pull/1248)
- Unclear if this PR is the root cause, but investigation is needed
- Issue affects the user interface responsiveness

**References:**
- PR: https://github.com/y-scope/clp/pull/1248
- Comment: https://github.com/y-scope/clp/pull/1248#issuecomment-

**Reporter:** @junhaoliao

hoophalab:
# Description

This PR fixes #1305.

The progress bar on the search page appears to be wider than the content layout. This is because the progress bar is using `position: absolute`, which makes it "float" on the search page rather than being part of the document flow. Because of this, this sizes it based on the entire page's width instead of its parent container (antd Layout).

To fix this, I created a new container with `position: relative` and a height of 0. This new container will not take up any space on the page, but it will serve as a reference point for the absolutely positioned progress bar. By placing the progress bar inside this new container, it will now correctly use the layout's width instead of the page's.

If you have a better way to approach this (e.g. using `position: fixed`), please let me know.

# Checklist

  breaking change.

# Validation performed

The progress bar no longer overflows.

## Summary by CodeRabbit

* Style
  * Updated the progress bar on the Presto search page to use fixed positioning, keeping it visible as you scroll through results.
  * Prevents the indicator from shifting or being obscured on longer pages and across different screen sizes, improving continuity and feedback during operations.
  * Provides a more stable layout and smoother visual behaviour, reducing flicker and optimizing readability during active searches.


components/webui/client/src/pages/SearchPage/Presto/ProgressBar/index.module.css
```diff
@@ -1,3 +1,9 @@
+.prestoProgressContainer {
+    /* A `position: relative` container is needed for the `prestoProgress` element to have the correct
+    width when using `position: absolute`. */
+    position: relative;
+}
+
 .prestoProgress {
     position: absolute;
```
junhaoliao: as discussed offline (and i tested), only changing this to `position: static;` should solve the issue, without causing other visual defects

hoophalab: static

yeah, but the page is moved down a bit when using static.

As suggested in the PR description, I find that using `position: fixed` can solve this issue, but I don't feel this is the correct use case for `fixed`. Any better ideas?

junhaoliao: maybe `sticky` could work? or maybe it has the same behaviour as `static` or `fixed`?

`fixed` might not be too bad, provided that it's good to always show the progress even if users scroll the page (i prefer this actually)

alternatively, for situations like components **slightly** moves after another component is dynamically mounted, we can:
1. keep that dynamic component always mounted
2. instead of letting the state to control whether the component is mounted, toggle the CSS visibility of the component between "hidden" and "visible"

hoophalab: I also tried visibility/opacity before, but both of them isn't updated as quick as display. So when the progress bar resets to zero, the bar is still displayed, and there is an animation of 100% -> 0%. I will use fixed.

PR #1312, opened 2025-09-18
feat(clp-s): Automatically infer the file types of inputs to compression.

gibber9809:
# Description

This PR changes the compression flow of clp-s to automatically infer the file-types of files passed to compression. We do this by iteratively:
1. Wrapping the input stream in a buffered reader
2. Scanning the first few buffered bytes to deduce the file type
3. If the content type is Zstd we wrap the buffered reader in a zstd decompressor and go back to step 1
4. Otherwise if the content is KV-IR or JSON we start parsing the input as the deduced content type

This allows us to get rid of the `--file-type` argument, as well as add support for new input types such as JSON compressed inside a Zstd stream. This approach should also make it easy to add support for ingesting input compressed by other compression algorithms such as LZMA in the future.

The bulk of these changes are split between:
1. A new `try_deduce_reader_type` function in `InputConfig` (and related utilities)
2. A refactor of `JsonParser` such that ingestion is always invoked by the `parsed` -- this comprises most of the diff, but really mostly involves moving around existing code

# Checklist

  breaking change.

# Validation performed

* Validated that all tests pass
* Manually validated that zstd compressed jsonl files can now be ingested

## Summary by CodeRabbit

- New Features
  - Automatic input-type detection for JSON, KV‑IR and Zstd-compressed inputs; removed explicit CLI file-type option and its accessor.

- Refactor
  - Unified ingestion flow with streaming readers, improved error handling, dynamic archive splitting and per-record metadata.

- Documentation
  - Removed --file-type from docs and examples; documented Zstd-compressed inputs.

- Tests
  - Test helpers and test cases updated to match automatic detection and simplified signatures.

- Chores
  - Build updated to include streaming-compression and reader components.


components/core/src/clp_s/InputConfig.cpp
```diff
@@ -163,6 +195,72 @@ auto try_create_network_reader(std::string_view const url, NetworkAuthOption con
...
+           || (0
+               == std::memcmp(
+                       peek_buf,
+                       clp::ffi::ir_stream::cProtocol::EightByteEncodingMagicNumber,
+                       clp::ffi::ir_stream::cProtocol::MagicNumberLength
+               ));
+}
+
+auto could_be_json(char const* peek_buf, size_t peek_size) -> bool {
+    for (size_t i{0ULL}; i < peek_size; ++i) {
```
wraymo, line 235: What if we have more spaces than `peek_size` and do you think the UTF‑8 BOM comment from coderabbit make sense?

components/core/src/clp_s/InputConfig.cpp
```diff
@@ -163,6 +195,72 @@ auto try_create_network_reader(std::string_view const url, NetworkAuthOption con
...
+        if (std::isspace(static_cast<int>(peek_buf[i]))) {
+            continue;
+        }
+
+        if ('{' == peek_buf[i]) {
+            return true;
+        } else {
+            return false;
+        }
```
wraymo:
```suggestion
        return '{' == peek_buf[i];
```

components/core/src/clp_s/InputConfig.cpp
```diff
@@ -163,6 +195,72 @@ auto try_create_network_reader(std::string_view const url, NetworkAuthOption con
...
+               ));
+}
+
+auto could_be_json(char const* peek_buf, size_t peek_size) -> bool {
+    for (size_t i{0ULL}; i < peek_size; ++i) {
+        if (std::isspace(static_cast<int>(peek_buf[i]))) {
+            continue;
+        }
+
+        if ('{' == peek_buf[i]) {
```
wraymo: How do we handle arrays or integers?

components/core/src/clp_s/InputConfig.cpp
```diff
@@ -95,6 +101,32 @@ auto get_archive_id_from_path(Path const& archive_path, std::string& archive_id)
...
+ * Peeks the first few bytes of input from a reader and tries to deduce the type of the underlying
+ * content.
+ * @param reader
+ * @return The deduced type, or `FileType::Unknown` if the type could not be deduced.
+ */
+auto peek_start_and_deduce_type(std::shared_ptr<clp::BufferedReader>& reader) -> FileType;
+
+/**
+ * Checks if an input contains Zstd data, based on the first few bytes of data from the input.
+ * @return Whether the input could be Zstd.
```
wraymo, line 117: Should we add parameter descriptions for these three methods?

components/core/src/clp_s/InputConfig.hpp
```diff
@@ -121,6 +124,18 @@ get_input_archives_for_raw_path(std::string_view const path, std::vector<Path>&
  */
 [[nodiscard]] auto try_create_reader(Path const& path, NetworkAuthOption const& network_auth)
         -> std::shared_ptr<clp::ReaderInterface>;
+
+/**
+ * Tries to deduce the underlying file-type of the file opened by `reader`, and return a
```
wraymo:
```suggestion
 * Tries to deduce the underlying file-type of the file opened by `reader`, and returns a
```

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -498,97 +496,317 @@ bool JsonParser::parse() {
...
-                    "Encountered error - {} - while trying to parse {} after parsing 0 bytes",
-                    simdjson::error_message(json_file_iterator.get_error()),
+        auto const [nested_readers, file_type] = try_deduce_reader_type(reader);
+        bool ingestion_successful{};
+        switch (file_type) {
+            case FileType::Json:
+                ingestion_successful = ingest_json(nested_readers.back(), path, archive_creator_id);
+                break;
+            case FileType::KeyValueIr:
+                ingestion_successful = ingest_kvir(nested_readers.back(), path, archive_creator_id);
```
wraymo, line 506: Since we only need the last one, is it required to return a vector?

components/core/src/clp_s/JsonParser.hpp
```diff
@@ -63,16 +61,10 @@ class JsonParser {
...
+     * @return Whether the input was parsed successfully.
      */
-    [[nodiscard]] bool parse();
-
-    /**
-     * Parses the Key Value IR Stream and stores the data in the archive.
-     * @return whether the IR Stream was parsed successfully
-     */
-    [[nodiscard]] auto parse_from_ir() -> bool;
+    [[nodiscard]] auto parse() -> bool;
```
wraymo: Not sure if we should also update this method name.

components/core/src/clp_s/InputConfig.cpp
```diff
@@ -175,4 +273,56 @@ auto try_create_reader(Path const& path, NetworkAuthOption const& network_auth)
         return nullptr;
     }
 }
+
+[[nodiscard]] auto try_deduce_reader_type(std::shared_ptr<clp::ReaderInterface> reader)
+        -> std::pair<std::vector<std::shared_ptr<clp::ReaderInterface>>, FileType> {
+    constexpr size_t cFileReadBufferCapacity = 64 * 1024;  // 64 KB
+    constexpr size_t cMaxNestedFormatDepth = 5;
```
wraymo, line 283: why 5?

wraymo (commented):
Great PR! Just had a few small comments.

components/core/src/clp_s/InputConfig.cpp
gibber9809: It's a somewhat arbitrary choice -- I'm just trying to avoid cases where someone maliciously or accidentally gives us a file that we interpret as being wrapped in many layers of compression and wasting compute/memory. I figure 5 is more than enough for real cases, especially for now.

components/core/src/clp_s/JsonParser.cpp
gibber9809: Unfortunately yes -- Since `zstd::Decompressor` only takes a reference to a reader instead of a `shared_ptr` if we return a decompressor as the last reader  alone (as we would do for kv-ir inputs) the previous readers would get deallocated. Returning a vector helps avoid this case. It can also be useful for example if we need to close multiple readers in the hierarchy before we exit (which I should be doing for the decompressors, but forgot about, so I'll add that to the PR).

components/core/src/clp_s/InputConfig.cpp
gibber9809: JSON doesn't allow BOM, and the current code should already reject BOM, so that part is fine. I'll double check about the negative characters though.

gibber9809: If we have more spaces than peek size I'm personally fine treating the input as not JSON -- for reference the peek size should 64KB unless the input is smaller, so the input has to be really malformed for this to not work.

components/core/src/clp_s/InputConfig.cpp
gibber9809: Currently we end up treating them as "Unknown" content type, and the user will see an error log like "could not deduce file-type for file xyz" before we exit. Since we don't actually support parsing JSON types besides object at the root right now I think this is fine -- we can change this check later if we start supporting other types.

components/core/src/clp_s/JsonParser.hpp
gibber9809: Sure, I can rename `parse` to `ingest`

components/core/src/clp_s/InputConfig.cpp
```diff
@@ -325,4 +328,15 @@ auto try_create_reader(Path const& path, NetworkAuthOption const& network_auth)
...
     return {{}, FileType::Unknown};
 }
+
+void close_nested_readers(std::vector<std::shared_ptr<clp::ReaderInterface>> const& readers) {
+    for (auto rit = readers.rbegin(); readers.rend() != rit; ++rit) {
+        if (auto decompressor
+            = std::dynamic_pointer_cast<clp::streaming_compression::Decompressor>(*rit);
+            nullptr != decompressor)
+        {
+            decompressor->close();
```
wraymo, line 338: Do we need to explicitly close other readers?

gibber9809: All of the ones we currently support don't even have a close method. Instead they do cleanup in the destructor conditional on being open.

wraymo (approved):

docs/src/user-docs/core-clp-s.md
```diff
@@ -14,12 +14,12 @@ Usage:
 * `archives-dir` is the directory that archives should be written to.
 * `input-path` is a filesystem path or URL to either:
   * a new-line-delimited JSON (ndjson) log file;
-  * a KV-IR file; or
+  * a KV-IR file;
+  * either of these file-types compressed by Zstd; or
```
kirkrodrigues:
```suggestion
  * a new-line-delimited JSON (ndjson) log file (may be Zstd-compressed);
  * a KV-IR file (may be Zstd-compressed);
```

kirkrodrigues (commented):
For the PR title, how about:

```
feat(clp-s): Automatically infer the file types of inputs to compression.
```

kirkrodrigues (approved):
Deferring to @wraymo's review.

PR #1315, opened 2025-09-19
feat(webui): Inject time range into guided UI search query; Generate timeline bucket query.

hoophalab:
# Description

1. Add a function to construct timeline bucket queries.
2. Check if the boolean expression references a timestamp column in the search query.
3. Add a WHERE filter based on a specified timestamp range in the search query.
4. Add a few testing input boxes.

timestampsql

# Checklist

  breaking change.

# Validation performed

1. Referencing the timestamp column in the WHERE clause of the query is causing an error (shown in the console).
2. The shown data in the result table is within the correct time range.
3. The timeline buckets are stored in MongoDB in the same the format as CLP aggregation task.

## Summary by CodeRabbit

- New Features
  - Runs a timeline aggregation query alongside the main search when submitting.
  - Added inputs for database name, start/end timestamps and timestamp key.
  - Searches now include time-range filtering and timeline bucketed aggregation.

- Improvements
  - Separate submission and logging for search and timeline queries with clearer error feedback.
  - Validation applied to both search and timeline queries to surface invalid SQL earlier.

### DIFF
--- a/components/clp-package-utils/pyproject.toml
+++ b/components/clp-package-utils/pyproject.toml
@@ -3,7 +3,7 @@ requires-python = ">= 3.9"
 
 [tool.poetry]
 name = "clp-package-utils"
-version = "0.5.1-dev"
+version = "0.5.2-dev"
 description = "Utilities for the CLP package."
 authors = ["YScope Inc. <dev@yscope.com>"]
 readme = "README.md"
--- a/components/clp-py-utils/pyproject.toml
+++ b/components/clp-py-utils/pyproject.toml
@@ -3,7 +3,7 @@ requires-python = ">= 3.9"
 
 [tool.poetry]
 name = "clp-py-utils"
-version = "0.5.1-dev"
+version = "0.5.2-dev"
 description = "Utilities for other Python packages in CLP."
 authors = ["YScope Inc. <dev@yscope.com>"]
 readme = "README.md"
--- a/components/core/src/clp/version.hpp
+++ b/components/core/src/clp/version.hpp
@@ -2,7 +2,7 @@
 #define CLP_VERSION_HPP
 
 namespace clp {
-constexpr char cVersion[] = "0.5.1-dev";
+constexpr char cVersion[] = "0.5.2-dev";
 }  // namespace clp
 
 #endif  // CLP_VERSION_HPP
--- a/components/core/src/glt/version.hpp
+++ b/components/core/src/glt/version.hpp
@@ -2,7 +2,7 @@
 #define GLT_VERSION_HPP
 
 namespace glt {
-constexpr char cVersion[] = "0.5.1-dev";
+constexpr char cVersion[] = "0.5.2-dev";
 }  // namespace glt
 
 #endif  // GLT_VERSION_HPP
--- a/components/job-orchestration/pyproject.toml
+++ b/components/job-orchestration/pyproject.toml
@@ -3,7 +3,7 @@ requires-python = ">= 3.9"
 
 [tool.poetry]
 name = "job-orchestration"
-version = "0.5.1-dev"
+version = "0.5.2-dev"
 description = "Scheduler and worker cluster for CLP's distributed architecture."
 authors = ["YScope Inc. <dev@yscope.com>"]
 readme = "README.md"
--- a/taskfile.yaml
+++ b/taskfile.yaml
@@ -32,7 +32,7 @@ vars:
   G_UTILS_TASKFILE: "{{.ROOT_DIR}}/tools/yscope-dev-utils/exports/taskfiles/utils/utils.yaml"
 
   # Versions
-  G_PACKAGE_VERSION: "0.5.1-dev"
+  G_PACKAGE_VERSION: "0.5.2-dev"
 
   # Build parameters
   # NOTE: Defaulting to an empty string is safe since CMake ignores an empty string.


components/webui/client/src/sql-parser/index.ts
```diff
@@ -46,6 +47,33 @@ class UpperCaseCharStream extends CharStream {
...
+            timestampKey,
+
+            // quotedIdentifier
+            `"${timestampKey}"`,
+
+            // backQuotedIdentifier
+            `\`${timestampKey}\``,
+        ]);
+
+        this.visitColumnReference = (ctx: ColumnReferenceContext) => {
```
davemarco: I think more idiotmatic to override the visitor directly?   override visitColumnReference(ctx: ColumnReferenceContext): void {

components/webui/client/src/sql-parser/index.ts
```diff
@@ -46,6 +47,33 @@ class UpperCaseCharStream extends CharStream {
...
+            if (timestampIdentifiers.has(ctx.identifier().getText())) {
+                this.hasTimestamp = true;
+            }
+        };
+    }
+}
```
davemarco: Also do we need this for t."timestamp"  override visitDereference(ctx: DereferenceContext): void {

components/webui/client/src/sql-parser/index.ts
```diff
@@ -91,40 +125,138 @@ interface BuildSearchQueryProps {
...
+    startTimestamp,
+    endTimestamp,
+    bucketCount,
+    timestampKey,
+}: BuildTimelineQueryProps) => {
+    // Converting float to decimal can have precision errors.
+    const startTimestampFixed = startTimestamp.toFixed(8);
+    const endTimestampFixed = endTimestamp.toFixed(8);
+    const queryString = `WITH filtered AS (
+    SELECT to_unixtime(${timestampKey}) AS timestamp
```
davemarco: I think it may be easier to read if we simplify the sql query, and just fix the bins and other issues in javascript in the server? What do u think? Also i think we still need to include the where expression here

components/webui/client/src/pages/SearchPage/SearchControls/Presto/BuildSqlTestingInputs.tsx
```diff
@@ -47,14 +61,33 @@ const BuildSqlTestingInputs = () => {
...
+                });
 
-                const sqlString = buildSearchQuery(props);
-                console.log(`SQL: ${sqlString}`);
-                handlePrestoQuerySubmit({queryString: sqlString});
+                console.log(`Timeline SQL: ${timelineQueryString}`);
+                submitQuery({queryString: timelineQueryString})
+                    .then((result) => {
+                        const {searchJobId} = result.data;
+                        console.log(`searchJobId: ${searchJobId}`);
```
davemarco: I guess this now aggregationJobId

components/webui/client/src/sql-parser/index.ts
davemarco: If this dosent make sense lmk

davemarco (changes requested):
timestamp key checker looks good. I have some slight concerns the sql is too complex, and we can maybe just do it in javascript

components/webui/client/src/sql-parser/SqlBase.g4
```diff
@@ -30,6 +33,10 @@ standaloneRoutineBody
     : routineBody EOF
     ;
 
+standaloneBooleanExpression
```
davemarco: Also how do u feel about having a second .g4 file, which imports the untouched grammer?

components/webui/client/src/sql-parser/index.ts
hoophalab: 1. I feel the only things that can be moved to javascript are `bucket_timestamp` and the final left join. The other query all operates on the whole dataset in time range, which should be executed in presto for better performance.

```
    SELECT
        index + 1 AS index,
        ${startTimestampFixed}
        + (index*(${endTimestampFixed}-${startTimestampFixed})/${bucketCount}) AS timestamp
    FROM
        UNNEST(sequence(0, ${bucketCount - 1}, 1)) AS t(index)
``` 
-->

```
SELECT *
FROM
    UNNEST(sequence(1,10,1) AS t(index),
    UNNEST(array [1600000, 1600010, ..., 1600090]) AS t(timestamp)
```
This replacement probably doesn't worth it.

2. the where expression is already included in `filtered AS ...` ?

davemarco: We cant just do this? 

```
SELECT
  width_bucket(to_unixtime(${timestampKey}), ${start}, ${end}, ${bucketCount}) AS idx,
  COUNT(*) AS cnt
FROM ${databaseName}
WHERE to_unixtime(${timestampKey}) >= ${start}
  AND to_unixtime(${timestampKey}) <  ${end}   
GROUP BY 1
ORDER BY 1;
```
and add the missing 0 buckets after?

davemarco: I just worry if someone has to modify the original query, they would have trouble

hoophalab: this could work, but it doesn't append begin `timestamp` to the result. Doing that in javascript can be even harder to read than a standard relational left join xD
https://github.com/y-scope/clp/blob/774a106cfe106de92be6efd0518e31a13ee211a4/components/core/src/reducer/ServerContext.cpp#L172C21-L172C44

I can merge the first three `with ... as ...`  to what you suggested, if we feel unnamed nested sql are actually easier to read than named sql. But they indeed take less line and are less scary

davemarco: i was thinking u just recalculate the timestamps based on the ranges in js? something like this? 
```
type Row = { idx: number; cnt: number };

function buildTimelineFromRows(
  rows: Row[],
  start: number,
  end: number,
  bucketCount: number
) {
  const width = (end - start) / bucketCount;

  // Start with all zeros
  const buckets = Array.from({ length: bucketCount }, (_, i) => ({
    index: i + 1,                           // 1..bucketCount
    timestamp: start + i * width,           // left edge (x-axis)
    count: 0,
  }));

  // Fill counts for buckets that appeared
  for (const { idx, cnt } of rows) {
    const i = idx - 1;
    if (i >= 0 && i < bucketCount) {
      buckets[i].count = cnt;
    }
  }

  return buckets; // [{index,timestamp,count}, ...]
}
```

hoophalab: yeah, but js code takes more lines and is harder to read than relational operations. Of course, this is just a personal preference

davemarco: kk just update to what u want and i will look

hoophalab:
@davemarco I've made the changes we discussed. The current code doesn't work when the timestamp key contains special chars such as `-`, but let me confirm with the other people of what is the correct way to reference such keys in WHERE. Other than this issue, the PR should be good to be merged.

components/webui/client/src/sql-parser/index.ts
```diff
@@ -100,31 +106,107 @@ const buildSearchQuery = ({
...
     sortItemList,
     limitValue,
+    startTimestamp,
+    endTimestamp,
+    timestampKey,
 }: BuildSearchQueryProps): string => {
-    let sqlString = `SELECT ${selectItemList} FROM ${relationList}`;
+    let queryString = `SELECT ${selectItemList} FROM ${relationList}
+WHERE to_unixtime(${timestampKey}) BETWEEN ${startTimestamp} AND ${endTimestamp}`;
+
```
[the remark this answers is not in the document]
hoophalab: to be fixed later

components/webui/client/src/sql-parser/index.ts
```diff
@@ -100,31 +106,107 @@ const buildSearchQuery = ({
...
+    timestampKey,
+}: BuildTimelineQueryProps) => {
+    const step = (endTimestamp - startTimestamp) / bucketCount;
+    const timestamps = Array.from(
+        {length: bucketCount},
+        (_, i) => startTimestamp + (i * step)
+    );
+
+    const queryString = `WITH buckets AS (
+    SELECT
```
davemarco, line 160: i think we should clean up indentation on query

components/webui/client/src/sql-parser/index.ts
```diff
@@ -100,31 +106,107 @@ const buildSearchQuery = ({
...
     } catch (err: unknown) {
-        throw new Error(`The constructed SQL is not valid: ${sqlString}`, {cause: err});
+        throw new Error(`The constructed SQL is not valid: ${queryString}`, {cause: err});
     }
 
-    return sqlString;
+    return queryString;
+};
+
+interface BuildTimelineQueryProps {
```
davemarco: i still think we should move types to typings.ts

davemarco (approved):
looks good 
`feat(webui): Inject time range into guided UI search query;Generate timeline bucket query.`

junhaoliao (approved):
deferring to @davemarco 's review

PR #1318, opened 2025-09-21
feat(clp-package): Add Spider and its dependencies to the build.

sitaowang1998:
# Description

This PR
* Adds `yscope-spider-py` as a dependency of `job-orchestration`.
* Changes to install c++ libraries in `build/deps/cpp`.
* Adds tasks to install `mariadb-connector-cpp` and `spider`.

Notice that `spider` and `mariadb-connector-cpp` are both Ubuntu only tasks because of the build issue of `mariadb-connector-cpp` on other platforms and that we don't need to use `spider` on other platforms.

# Checklist

  breaking change.

# Validation performed

* [x] `spider` C++ package installed successfully in Ubuntu 24.04. 
* [x] Running `deps:core` and then `deps:spider` or vice versa does not build libraries again in Ubuntu 24.04.
* [x] `job-orchestration` builds.
* [x] GitHub workflows pass.

## Summary by CodeRabbit

* New Features
  * Added Spider build/orchestration tasks and a Spider build-dir variable; added yscope-spider-py dependency.

* Refactor
  * Migrated public/config namespace from “core” to “cpp” (paths, checksum handling, CMake settings, tasks and artifact outputs).

* Chores
  * Standardized parallelism env var to CLP_CPP_MAX_PARALLELISM_PER_BUILD_TASK across workflows and tasks.
  * Tightened job-orchestration Python constraint to >=3.9,<4.0.


kirkrodrigues:
> Before we proceed: do we need a new namespace under components? I thought spider related subcomponents can also be located in the current `job-orchestration`: the current job orchestration already contains non-celery related components. But as I'm also new to the package side of thing, let's wait for @junhaoliao @kirkrodrigues to confirm.

Yeah, name-wise, I think it makes more sense to be under `job-orchestration`. I guess Sitao was making it a separate "component" since the package will only run one or the other (job-orchestration or spider-orchestration)?

sitaowang1998:
I have two reasons to use a separate "component":
1. The task function name is already occupied. I can go with `spider_compress` though.
2. It will be easier to do the switch later if we have a component that does not have `celery` dependency.

LinZhihao-723:
> > Before we proceed: do we need a new namespace under components? I thought spider related subcomponents can also be located in the current `job-orchestration`: the current job orchestration already contains non-celery related components. But as I'm also new to the package side of thing, let's wait for @junhaoliao @kirkrodrigues to confirm.
> 
> Yeah, name-wise, I think it makes more sense to be under `job-orchestration`. I guess Sitao was making it a separate "component" since the package will only run one or the other (job-orchestration or spider-orchestration)?

The problem is that we actually can't fully get rid of job-orchestration: we still need search components and the garbage collector.

LinZhihao-723:
> I have two reasons to use a separate "component":
> 
> 1. The task function name is already occupied. I can go with `spider_compress` though.
> 2. It will be easier to do the switch later if we have a component that does not have `celery` dependency.

For 1: I don't get what u mean by "task function name".
For 2: as explained in https://github.com/y-scope/clp/pull/1318#issuecomment-3322362318, I think we can't fully get rid of `celery` now... we'd also need the garbage collector.

sitaowang1998:
For 1, the task function in celery is already named `compress`, so we have to choose another name.
For 2, I am not ditching the whole code space, but rather copy around the code we need and structure them better. However, keeping two copies of code is hard to maintain. This is rather a judgement call.

LinZhihao-723:
> For 1, the task function in celery is already named `compress`, so we have to choose another name. For 2, I am not ditching the whole code space, but rather copy around the code we need and structure them better. However, keeping two copies of code is hard to maintain. This is rather a judgement call.

For 1: How about we rename them into `celery_compress` and `spider_compress`?
For 2: Briefly looked into your fork. How about we layout things in the following way:
```
- job-orchistration
  - executor
    - celery_compress
    - spider_compress
  - scheduler
    - celery_compress
    - spider_compress
```
This will rename the existing `compress` to `celery_compress`.
@kirkrodrigues does this plan work for u?

taskfiles/lint.yaml
LinZhihao-723, line 1: Should be reverted.

components/job-orchestration/pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [project]
-requires-python = ">= 3.9"
+requires-python = ">= 3.9,<4.0"
```
LinZhihao-723, line 2: Just curious, is this necessary?

components/job-orchestration/pyproject.toml
```diff
@@ -21,6 +21,7 @@ pika = "^1.3.2"
 pydantic = "^1.10.13"
 pymongo = "^4.7.2"
 PyYAML = "^6.0.1"
+yscope-spider-py="~0.1.0"
```
LinZhihao-723: Can we install the exact `0.1.0` version? We should explicitly switch to a newly released version when needed.

taskfile.yaml
```diff
@@ -22,6 +22,7 @@ vars:
   # Build paths
   G_BUILD_DIR: "{{.ROOT_DIR}}/build"
   G_CORE_COMPONENT_BUILD_DIR: "{{.G_BUILD_DIR}}/core"
+  G_SPIDER_BUILD_DIR: "{{.G_BUILD_DIR}}/spider"
```
LinZhihao-723: Can u move it down before `G_WEBUI_BUILD_DIR` to follow the lexical order?

taskfile.yaml
```diff
@@ -206,9 +207,30 @@ tasks:
       - task: "utils:cmake:build"
         vars:
           BUILD_DIR: "{{.G_CORE_COMPONENT_BUILD_DIR}}"
-          JOBS: "{{.G_CORE_MAX_PARALLELISM_PER_BUILD_TASK}}"
+          JOBS: "{{.G_CPP_MAX_PARALLELISM_PER_BUILD_TASK}}"
           TARGETS: ["clg", "clo", "clp", "clp-s", "indexer", "reducer-server"]
 
+  spider:
```
LinZhihao-723: How about `spider-build`?

taskfile.yaml
```diff
@@ -206,9 +207,30 @@ tasks:
...
+    deps:
+      - "deps:cpp"
+    platforms: ["linux"]
+    status:
+      - >-
+        ! grep -q '^ID=ubuntu' /etc/os-release
```
LinZhihao-723:
```suggestion
        source /etc/os-release && "$ID" != "ubuntu"
```
How about loading the release properly?

taskfile.yaml
```diff
@@ -206,9 +207,30 @@ tasks:
...
+          CMAKE_PACKAGE_NAME: "spider"
+          TAR_SHA256: "43478b6da99aa2df50c960573ec68c6fcf67cdfbab39b5e314676218aa93b959"
+          TAR_URL: "https://github.com/y-scope/spider/archive/6a7bad3.tar.gz"
+          WORK_DIR: "{{.G_SPIDER_BUILD_DIR}}"
+          CMAKE_JOBS: "{{.G_CPP_MAX_PARALLELISM_PER_BUILD_TASK}}"
+          CMAKE_GEN_ARGS:
+            - "-C {{.G_DEPS_CPP_CMAKE_SETTINGS_DIR}}/all.cmake"
+            - "-DCMAKE_BUILD_TYPE=Release"
+            - "-DCMAKE_INSTALL_MESSAGE=LAZY"
+            - "-DSPIDER_BUILD_TESTING=OFF"
```
LinZhihao-723: You may want to double-check this:
- This task doesn't do buffering. Rerunning it will trigger a new build. In this case, it's probably better to make it internal and trigger it by other tasks with `once`.
- I don't see anything installed after running this task even with the following message:

```
task: [utils:cmake:install:/home/lzh/clp/build/spider/spider-build-/home/lzh/clp/build/spider/spider-install-[]] cmake --install "/home/lzh/clp/build/spider/spider-build" --prefix "/home/lzh/clp/build/spider/spider-install"
-- Install configuration: "Release"
```

taskfiles/deps/main.yaml
```diff
@@ -7,24 +7,30 @@ includes:
...
   # Library names
+  G_ABSL_LIB_NAME: "absl"
+  G_ANTLR_RUNTIME_LIB_NAME: "antlr4-runtime"
   G_BOOST_LIB_NAME: "Boost"
   G_FMT_LIB_NAME: "fmt"
   G_GSL_LIB_NAME: "Microsoft.GSL"
+  G_MARIADB_CONNECTOR_CPP_LIB_NAME: "mariadb-connector-cpp"
+  G_MSGPACK_CXX_LIB_NAME: "msgpack-cxx"
+  G_SPDLOG_LIB_NAME: "spdlog"
+  G_YSTDLIB_LIB_NAME: "ystdlib"
```
LinZhihao-723: Why we need these?
- The existing ones are made variables because they're referenced in more than one place. The ones u added are not.
- If we need this refactoring, I'd prefer to put them into a separate PR (and technically we should also put `CORE` -> `CPP` refactoring into a separate PR. But let's keep them in this PR for this time)

taskfiles/deps/main.yaml
```diff
@@ -337,6 +345,25 @@ tasks:
           TARBALL_SHA256: "537512904744b35e232912055ccf8ec66d768639ff3abe5788d90d792ec5f48b"
           TARBALL_URL: "https://github.com/lz4/lz4/releases/download/v1.10.0/lz4-1.10.0.tar.gz"
 
+  mariadb-connector-cpp:
+    internal: true
+    run: "once"
+    platforms: ["linux"]
+    status:
+      - >-
+        ! grep -q '^ID=ubuntu' /etc/os-release
```
LinZhihao-723: Same.

taskfile.yaml
LinZhihao-723: And actually, why do we not build for Ubuntu? Or do u want to build for Ubuntu only (according to the PR description)? That said, I believe it's worth adding a comment explaining what this behavior is and why.

taskfiles/deps/main.yaml
```diff
@@ -337,6 +345,25 @@ tasks:
...
+    cmds:
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-DCMAKE_POSITION_INDEPENDENT_CODE=ON"
```
LinZhihao-723: Why do we need PIC?

taskfiles/deps/main.yaml
```diff
@@ -337,6 +345,25 @@ tasks:
...
+            - "-DUSE_SYSTEM_INSTALLED_LIB=ON"
+            - "-DINSTALL_LAYOUT=RPM"
```
LinZhihao-723: According to the PR description, if this is only for ubuntu, why the install layout is `RPM`?

taskfiles/deps/main.yaml
```diff
@@ -337,6 +345,25 @@ tasks:
...
+          LIB_NAME: "{{.G_MARIADB_CONNECTOR_CPP_LIB_NAME}}"
+          TARBALL_URL: "https://github.com/mariadb-corporation/mariadb-connector-cpp/archive/refs/\
+            tags/1.1.5.tar.gz"
```
LinZhihao-723, line 410: Let's make the version a variable like `msgpack-cxx`.

taskfiles/deps/utils.yaml
LinZhihao-723, line 1: It seems like if I run `task deps:cpp` twice the tasks won't be skipped; cmake will still be triggered but return immediately as all targets are cached and buffered. It's the same in our latest merged commit. Just double checked whether this is an expected behavior or a bug @kirkrodrigues .

components/job-orchestration/pyproject.toml
sitaowang1998: Yes. Without it when running `poetry lock`, an error message shows:
```
The current project's supported Python range (>=3.9) is not compatible with some of the required packages Python requirement:
  - msgpack-types requires Python <4.0,>=3.8, so it will not be satisfied for Python >=4.0
```

taskfile.yaml
sitaowang1998: This is not only a build. This task downloads `spider`, generates the build files and then builds the project.

taskfile.yaml
sitaowang1998: I think you mean `source /etc/os-release && [[ $ID != ubuntu ]]`.

components/job-orchestration/pyproject.toml
LinZhihao-723: I see, sure.

taskfile.yaml
sitaowang1998: 1. The task buffers the `spider` package it downloads. `cmake` can discovers that the source and build files are untouched and skip the actual build. I ran this task twice, and observe that no new thing is built.
2. This is intended. Currently `spider` hasn't setup install targets yet, and the `cmake install` inside the `yscope-dev-utils:install-remote-tar` will just not install anything. This behavior is fine for now and also works when we later support install targets in `spider`.

taskfiles/deps/main.yaml
sitaowang1998: This is no longer needed. I'll remove them.

taskfile.yaml
LinZhihao-723: How about moving it to `taskfiles/main.yaml`?

taskfile.yaml
LinZhihao-723: Yeah. I think if the goal is to test whether it's ubuntu (need ur confirmation in another comment), we can use the following for clarity:
```suggestion
        source /etc/os-release && [[ "$ID" == "ubuntu" ]]
```

LinZhihao-723: Actually reading more about `status` I'm afriad it won't do the work as u want: https://taskfile.dev/docs/guide. In my local dev env, no matter `!=` or `==` I use the task will still execute.

sitaowang1998: The intent is to build only for Ubuntu. The `status` is to skip the task if all its tests are true. Thus, we need to make the test command evaluate to false on Ubuntu only. I'll add a comment on both location.

sitaowang1998: I am sure the `status` work. At least the original version successfully skip files in GitHub workflows for platforms other than Ubuntu.

LinZhihao-723: Can u check this task will actually be skipped for non-ubuntu linux? From what I've tried `status` doesn't work in that way.

sitaowang1998: For example, the [CentOS workflow log](https://github.com/y-scope/clp/actions/runs/17956520407/job/51069387321?pr=1318#step:4:87) shows that `mariadb-connector-cpp` task is already up-to-date and thus skipped.

LinZhihao-723: Hmmm, that's weird. It doesn't work well on my local WSL. Let me check.

taskfiles/deps/main.yaml
sitaowang1998: Because other install layouts don't work with the [`FindMariaDBClientCpp.cmake`](https://github.com/y-scope/spider/blob/main/cmake/Modules/FindMariaDBClientCpp.cmake). I tried to tweak the `cmake` file to make it work with other layouts, but failed.

taskfiles/deps/main.yaml
sitaowang1998: No. Will remove it.

taskfile.yaml
LinZhihao-723: Alright I think I get what happened. The deps will still execute whether or not this task is skipped.
That said, for mariadb connection installation, I'm ok to use `status` since we can safely skip it.
For spider, let's use `preconditions` instead: https://taskfile.dev/docs/guide. This allows us to not proceed with any jobs that depend on the spider.

sitaowang1998: `preconditions` is not the right tool. It crash the workflow if the test fails. From the doc:

>Unlike status, which will skip a task if it is up to date and continue executing tasks that depend on it, a precondition will fail a task, along with any other tasks that depend on it.

This is not what we want. Unfortunately we have to use `status` as a workaround.

taskfiles/deps/main.yaml
sitaowang1998: I don't think a version is necessary. `msgpack-cxx` and `mongocxx` have version variables because the version appears twice in the url.

taskfiles/deps/utils.yaml
sitaowang1998: The only exception is the newly added `mariadb-connector-cpp` which rebuilds every time the the generate step modifies the source tree.

LinZhihao-723 (changes requested):
Also, we need to update the doc. Check this PR: https://github.com/y-scope/clp/pull/1229/files

taskfiles/deps/main.yaml
LinZhihao-723: Sure.

taskfile.yaml
LinZhihao-723: Why do u want a task depends on spider to keep running if we can't even download and install spider?

sitaowang1998: We want the task to be skipped instead of failing. Failing a task will fail the GitHub workflow. Since `mariadb-connector-cpp` is the only one dependency that is used by `spider` but not `core`, we can just skip `mariadb-connector-cpp` and `spider`.

LinZhihao-723: - I'm ok to skip `mariadb-connector-cpp`, as I suggested above. For this task, we should use `status`.
- I'm saying we should fail `spider` and any dependent task if it's not on Ubuntu, so we should use `precondition`. Why should the workflow succeed if it depends on the spider, but the spider can't be built?

taskfiles/deps/main.yaml
LinZhihao-723: As discussed offline, let's try `DEB` since it worked well locally on my WSL.

taskfile.yaml
sitaowang1998: As discussed offline, we will skip `mariadb-connector-cpp` and fails `spider` on platforms other than Ubuntu. We neither support nor need to build spider on other platforms.

taskfiles/deps/main.yaml
sitaowang1998: Confirm that `DEB` works on Ubuntu 24.04 as well. Will change the layout to `DEB`.

taskfiles/deps/main.yaml
```diff
@@ -337,6 +388,20 @@ tasks:
           TARBALL_SHA256: "537512904744b35e232912055ccf8ec66d768639ff3abe5788d90d792ec5f48b"
           TARBALL_URL: "https://github.com/lz4/lz4/releases/download/v1.10.0/lz4-1.10.0.tar.gz"
 
+  mariadb-connector-cpp:
```
LinZhihao-723, line 391:
```suggestion
  mariadb-connector-cpp:
    platforms: ["linux"]
    preconditions:
      # NOTE: The MariaDB connector is only required for building Spider, which is only supported on
      # Ubuntu. `INSTALL_LAYOUT` is currently hardcoded to "DEB". To support other Linux distros:
      # - Update the precondition accordingly.
      # - Make `INSTALL_LAYOUT` configurable.
      - >-
        source /etc/os-release && [[ "$ID" == "ubuntu" ]]
```
Let's add this to avoid any potential misuse of this task.

taskfiles/deps/main.yaml
```diff
@@ -88,6 +88,56 @@ tasks:
       - task: "zlib"
       - task: "zstd"
 
+  spider:
+    deps:
+      - "spider-dep"
+    platforms: ["linux"]
+    preconditions:
+      # We do not support building `spider` on platforms other than Ubuntu.
```
LinZhihao-723:
```suggestion
      # NOTE: Spider is only supported on Ubuntu at present.
```
We don't need \` for spider since it's not a symbol. It's like we directly refer to clang-format, not \`clang-format\`

taskfiles/deps/main.yaml
LinZhihao-723: I didn't find a way to test whether it's debian compatible or not. I will defer this to future PRs when needed.

taskfiles/deps/utils.yaml
LinZhihao-723: As discussed offline we're fine with this for now.

.github/workflows/clp-artifact-build.yaml
```diff
@@ -354,12 +354,13 @@ jobs:
...
             || (github.event_name != 'pull_request' && github.ref == 'refs/heads/main')}}
           run_command: >-
-            CLP_CORE_MAX_PARALLELISM_PER_BUILD_TASK=$(getconf _NPROCESSORS_ONLN) task deps:core
+            CLP_CPP_MAX_PARALLELISM_PER_BUILD_TASK=$(getconf _NPROCESSORS_ONLN) task deps:core
             && python3 /mnt/repo/components/core/tools/scripts/utils/build-and-run-unit-tests.py
             ${{matrix.use_shared_libs == true && '--use-shared-libs' || ''}}
             --source-dir /mnt/repo/components/core
             --build-dir /mnt/repo/components/core/build
             --num-jobs $(getconf _NPROCESSORS_ONLN)
+            && CLP_CPP_MAX_PARALLELISM_PER_BUILD_TASK=$(getconf _NPROCESSORS_ONLN) task deps:spider
```
LinZhihao-723: This is wrong. We should not modify this workflow. Instead, add `deps:spider` into `package` task, and rely on the package-image workflow to build the whole package (which will trigger spider build).

LinZhihao-723 (approved):
For the PR title, how about:
```
feat(clp-package): Add Spider and its dependencies to the build.
```

taskfile.yaml
```diff
@@ -121,6 +122,7 @@ tasks:
       - "core"
       - "clp-package-utils"
       - "clp-py-utils"
+      - "deps:spider"
```
junhaoliao, line 125: we should let `deps:spider` compute a checksum file and let add the checksum file to the `package` task's `source` list; otherwise, the `package` might not be re-run if there are changes in the artifacts generated.

alternatively, we can simply add `spider_scheduler` and `spider_worker` to the `package` task's `source` list.

taskfile.yaml
junhaoliao, line 1: (first reported by @sitaowang1998 )

it was found that `libmariadbcpp.so`, which both `spider_scheduler` and `spider_worker` dynamically link to, are not added into the package execution environment (i.e., the `clp-package` Docker execution image)

PR #1329, opened 2025-09-24
refactor(clp-s): Improve coding style and fix clang-tidy violations in `restore_encoded_float`.

LinZhihao-723:
# Description

As the title suggests.

# Checklist

  breaking change.

# Validation performed

* [ ] Ensure all workflows passed.
* [x] Ensure existing unit tests all passed.

### DIFF
--- a/components/core/src/clp_s/FloatFormatEncoding.cpp
+++ b/components/core/src/clp_s/FloatFormatEncoding.cpp
@@ -265,65 +265,67 @@ auto restore_encoded_float(double value, float_format_t format)
     std::ostringstream oss;
     oss.imbue(std::locale::classic());
     auto const num_significant_digits{get_num_significant_digits(format)};
-    oss << std::scientific << std::setprecision(num_significant_digits - 1);
-    if (has_scientific_notation(format)) {
-        if (is_uppercase_exponent(format)) {
-            oss << std::uppercase;
-        }
+    oss << std::scientific << std::setprecision(static_cast<int>(num_significant_digits) - 1);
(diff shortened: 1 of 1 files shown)


anlowee (dismissed):

components/core/src/clp_s/FloatFormatEncoding.cpp
```diff
@@ -265,65 +265,67 @@ auto restore_encoded_float(double value, float_format_t format)
...
+            );
         }
 
         return formatted_double_str;
     }
 
-    // Convert the scientific notation to the standard decimal
-    oss << value;
-    return scientific_to_decimal(oss.str());
+    if (num_exp_digits < (formatted_double_str.length() - exp_pos - 2)) {
```
gibber9809: While we're at it, could we re-order this if/else block and the following if/else block? I think the code will be much easier to understand if it first ensure that there is a +/- sign in this path before fixing up the number of exponent digits. I think this current code actually doesn't always work if there isn't a +/- sign when we hit this code block -- though given the fuzzing maybe the cases it doesn't work for don't happen in practice?

At any rate, it would be simpler to first guarantee that `e|E` is followed by `+|-` before trying to adjust the number of exponent digits.

gibber9809 (changes requested):

components/core/src/clp_s/FloatFormatEncoding.cpp
LinZhihao-723: Make sense

PR #1331, opened 2025-09-25
fix(taskfile): Make `deps:core` and `deps:spider-dep` write into different top-level CMake configuration files (fixes #1330).

### LINKED ISSUE #1330: Dependency build failure because of error in `build/deps/cpp/cmake-settings/all.cmake`.
### Bug

Running `task` sometimes throws an error:
```
loading initial cache file /workspace/clp/build/deps/cpp/cmake-settings/all.cmake                                                                                                                                     
CMake Error at /workspace/clp/build/deps/cpp/cmake-settings/all.cmake:9:
  Parse error.  Expected a newline, got identifier with text "include".
```

### CLP version

41f5c3c585256b88a57379f222630d45e6de2005

### Environment

Ubuntu 22.04

### Reproduction steps

Delete `build/deps/cpp/cmake-settings/all.cmake` and rerun `task`.

sitaowang1998:
# Description

## Problem
Both `deps:core` and `deps:spider-dep` write to `build/deps/cpp/cmake-setting/all.cmake`. When running in parallel, the content of `all.cmake` is corrupted, and both `core` and `deps:spider` fail to load the corrupted cmake file.

## Solution
This PR solves the problem by splitting the cmake setting file into two. `deps:core` writes to `core-all.cmake` and `deps:spider-dep` writes to `spider-all.cmake`, avoiding concurrent writes to same file. `core` and `deps:spider` now consume different cmake setting files.

# Checklist

  breaking change.

# Validation performed

* [x] Running `task` no longer triggers errors.
* [ ] GitHub workflows pass.

## Summary by CodeRabbit

- Chores
  - Standardized build configuration by switching to component-specific CMake settings for core and Spider.
  - Ensured the chosen settings file is propagated through dependency generation and installation steps to improve reproducibility.
  - Aligned dependency installation and task execution to use consistent presets, reducing configuration drift across environments.
  - No user-facing functionality changes; improves reliability and predictability of builds.

### DIFF
--- a/components/core/CMakeLists.txt
+++ b/components/core/CMakeLists.txt
@@ -123,7 +123,7 @@ if(PROJECT_IS_TOP_LEVEL)
     # NOTE: We mark the file optional since it's not required if the user happens to have the
     # dependencies installed already.
     set(CLP_CORE_DEPS_DIR "${CMAKE_SOURCE_DIR}/../../build/deps/cpp")
-    include("${CLP_CORE_DEPS_DIR}/cmake-settings/all.cmake"
+    include("${CLP_CORE_DEPS_DIR}/cmake-settings/core-all.cmake"
             OPTIONAL
             RESULT_VARIABLE CLP_DEPS_SETTINGS_FILE_PATH
     )
--- a/components/core/CMakeLists.txt
+++ b/components/core/CMakeLists.txt
@@ -123,7 +123,7 @@ if(PROJECT_IS_TOP_LEVEL)
     # NOTE: We mark the file optional since it's not required if the user happens to have the
     # dependencies installed already.
     set(CLP_CORE_DEPS_DIR "${CMAKE_SOURCE_DIR}/../../build/deps/cpp")
-    include("${CLP_CORE_DEPS_DIR}/cmake-settings/core-all.cmake"
+    include("${CLP_CORE_DEPS_DIR}/cmake-settings/all-core.cmake"
             OPTIONAL
             RESULT_VARIABLE CLP_DEPS_SETTINGS_FILE_PATH
     )
--- a/taskfiles/deps/main.yaml
+++ b/taskfiles/deps/main.yaml
@@ -46,7 +46,7 @@ tasks:
       - task: "yscope-dev-utils:cmake:install-deps-and-generate-settings"
         vars:
           CMAKE_SETTINGS_DIR: "{{.G_DEPS_CPP_CMAKE_SETTINGS_DIR}}"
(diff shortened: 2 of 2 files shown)


LinZhihao-723 (changes requested):
Overall idea for splitting into different meta-configs lgtm. The only thing is that shall we rename the setting files like:
- `core-all.cmake` -> `core.cmake`
- `spider-all.cmake` -> `spider.cmake`

This is more concise, and the full path `cmake-settings` already indicates they're setting configs. The concern for doing this would be later if we need spider as a dependency for core, then `spider.cmake` can be confusing as it may need to serve for the config for core's spider dependency, not a meta-config. To solve that, we can put the dependency-generated cmake into an inner directory inside `cmake-settings`, so that only meta-config at the top-level.

@kirkrodrigues Does this make sense to u?

kirkrodrigues:
> Overall idea for splitting into different meta-configs lgtm. The only thing is that shall we rename the setting files like:
> 
>     * `core-all.cmake` -> `core.cmake`
> 
>     * `spider-all.cmake` -> `spider.cmake`
> 
> 
> This is more concise, and the full path `cmake-settings` already indicates they're setting configs. The concern for doing this would be later if we need spider as a dependency for core, then `spider.cmake` can be confusing as it may need to serve for the config for core's spider dependency, not a meta-config. To solve that, we can put the dependency-generated cmake into an inner directory inside `cmake-settings`, so that only meta-config at the top-level.
> 
> @kirkrodrigues Does this make sense to u?

Do we just want to create that inner folder now? Maybe `cmake-settings/combined`?

LinZhihao-723:
> > Overall idea for splitting into different meta-configs lgtm. The only thing is that shall we rename the setting files like:
> > ```
> > * `core-all.cmake` -> `core.cmake`
> > 
> > * `spider-all.cmake` -> `spider.cmake`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is more concise, and the full path `cmake-settings` already indicates they're setting configs. The concern for doing this would be later if we need spider as a dependency for core, then `spider.cmake` can be confusing as it may need to serve for the config for core's spider dependency, not a meta-config. To solve that, we can put the dependency-generated cmake into an inner directory inside `cmake-settings`, so that only meta-config at the top-level.
> > @kirkrodrigues Does this make sense to u?
> 
> Do we just want to create that inner folder now? Maybe `cmake-settings/combined`?

Sure. @sitaowang1998 What do u think? Can u try if this works?

PR #1335, opened 2025-09-25
feat(docker): Generate unique tag for each clp-package dev image build to avoid conflicts on shared hosts.

junhaoliao:
# Description

Developer of the CLP project may be sharing the same machine or run multiple builds on the machine. Previously, in the clp-package image build script, we always tag the image builds with tag "clp-package:dev", which causes any older build, whether owned by the current user or not, to be deleted.

This PR adds the `--iidfile` to the `docker build` (buildx) command so that a unique image SHA256 ID is written to the specified location. Then the `docker-image:package` copies the generated `image.id` file into the package. In a future PR, the package start script will read this `image.id` file and use it for CLP service executions.

# Checklist

  breaking change.

# Validation performed

This ran successfully:
```
junhao@GIGABYTE:~/workspace/clp$ task docker-images:package
...
junhao@GIGABYTE:~/workspace/clp$ cat build/clp-package/image.id 
sha256:39c43e97507ef73cfe06e1441fe1487a590f04d786ac065dd500d2e64b96df01
```

## Summary by CodeRabbit

* Chores
  * Improved container image build workflow with automatic cleanup of previous images and temporary files to avoid leftover artifacts.
  * Builds now capture and persist image IDs, then auto-tag images with a dev user plus a short generated ID for easier identification.
  * When built inside a Git repository, images include revision and source labels for better traceability.

* Documentation
  * Added Docker prerequisites and a reference link to Docker installation documentation.


junhaoliao:
@kirkrodrigues if you think it makes sense to, in the same PR i can also remove `tools/docker-images/clp-execution-base-ubuntu-jammy` and update https://github.com/y-scope/clp/blob/173755f51399d786c6a1d36bb502e969fa25ccd8/components/clp-py-utils/clp_py_utils/clp_config.py#L932-L947 

---

we also discussed about cleaning up the image build process (that we don't have to copy everything into clp-package first; instead, we should specify the `build/<component>` paths in `COPY` commands in the Dockerfile and just directly copy from the `build/<component>`). while possible to be done in the same PR, i think it's cleaner to be done is a separate PR. if you agree, we can add an issue for that

taskfiles/docker-images.yaml
```diff
@@ -12,3 +12,4 @@ tasks:
       - ":package"
     cmds:
       - "./build.sh"
+      - "rsync -a 'image.id' '{{.G_PACKAGE_BUILD_DIR}}'"
```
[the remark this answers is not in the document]
junhaoliao: The `docker-images:package` task depends on the `:package` task, which already creates the G_PACKAGE_BUILD_DIR

tools/docker-images/clp-package/build.sh
```diff
@@ -4,11 +4,26 @@ set -eu
 set -o pipefail
 
 script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+iid_file="${script_dir}/image.id"
```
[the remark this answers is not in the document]
junhaoliao: it's not expected the user would invoke multiple image builds at the same time. @kirkrodrigues what do you think? shall we implement the rabbit's proposal and somehow guard against multiple concurrent builds?

tools/docker-images/clp-package/build.sh
```diff
@@ -4,11 +4,26 @@ set -eu
...
 repo_root=${script_dir}/../../../
 
+# Remove the previous image after the build to allow layer reuse.
```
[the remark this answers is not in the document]
junhaoliao: I mean to say:
1. we should remove the previous build to save space since nobody would use the old build once a new one is built
2. we should only do it after the current build is done; otherwise, it was found that it takes 6 more seconds on my computer to build the image

I can see "layer reuse" is not the most accurate wording. can you say "export reuse" or something similar?

tools/docker-images/clp-package/build.sh
```diff
@@ -4,11 +4,26 @@ set -eu
...
+    prev_image_id=$(cat "$iid_file")
+fi
+cleanup() {
+    if [[ -n "$prev_image_id" ]] && docker image inspect "$prev_image_id" >/dev/null 2>&1; then
+        echo "Removing previous image $prev_image_id"
+        docker image remove "$prev_image_id"
+    fi
+}
+trap cleanup EXIT
+
```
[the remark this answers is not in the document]
junhaoliao: i'm not sure how your proposal is cleaner. it also has so many nested-ifs, can you revisit?

tools/docker-images/clp-package/build.sh
```diff
@@ -4,11 +4,26 @@ set -eu
...
 build_cmd=(
     docker build
-    --tag "clp-package:dev"
+    --iidfile "$iid_file"
+    --tag "clp-package:dev-${USER}-$(date +%s)"
     "$repo_root"
     --file "${script_dir}/Dockerfile"
 )
```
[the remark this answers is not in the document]
junhaoliao: `docker build` is an alias of `docker buildx build` which uses the `BuildKit` as the builder. The switch happened in https://docs.docker.com/engine/release-notes/23.0/#2300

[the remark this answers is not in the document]
junhaoliao: @kirkrodrigues i realise we don't really mention any docker version in our dev docs (we do mention docker in the [userquick start guide](https://docs.yscope.com/clp/main/user-docs/quick-start/index.html#docker), though we don't say the version still). if that's not intentionally left out, maybe we should specify the version in our dev docs as well?

tools/docker-images/clp-package/build.sh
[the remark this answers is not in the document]
junhaoliao: i can't be totally sure if the benefits are clear to me, do you just hate the `trap` in the original code?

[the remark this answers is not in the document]
junhaoliao: i can't be sure if you understand the code correctly. will let @kirkrodrigues decide

taskfiles/docker-images.yaml
```diff
@@ -12,3 +12,4 @@ tasks:
       - ":package"
     cmds:
       - "./build.sh"
+      - "rsync -a '{{.G_BUILD_DIR}}/clp-package-image.id' '{{.G_PACKAGE_BUILD_DIR}}'"
```
kirkrodrigues: * Do we need `rsync` or would `cp` be sufficient?
* Why copy it into the package?

tools/docker-images/clp-package/build.sh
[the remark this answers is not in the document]
kirkrodrigues: I don't think it's necessary right now.

tools/docker-images/clp-package/build.sh
```diff
@@ -5,10 +5,25 @@ set -o pipefail
...
+        docker image remove "$prev_image_id"
+    fi
+}
+trap cleanup EXIT
 
 build_cmd=(
     docker build
-    --tag "clp-package:dev"
+    --iidfile "$iid_file"
+    --tag "clp-package:dev-${USER}-$(date +%s)"
```
kirkrodrigues: If we tag with the current timestamp, then identical image builds will result in two different tags pointing at the same image. Besides the clutter, that also means that the image can't be removed without a force flag. Instead, can we replace the timestamp with the image ID? I know that will make it harder to tell what's a newer image based on the tag, but that information is available from Docker since it keeps track of the date an image was built.

tools/docker-images/clp-package/build.sh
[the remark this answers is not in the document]
kirkrodrigues: I think the rabbit is correct.

* If the user kills the script before it completes, the existing image will be deleted. Is that intentional?
* If the current and previous image are identical, their IDs will be identical (since the ID is a sha256sum of the image), so on exit, the image will be deleted.
  * This doesn't currently happen since our builds are not exactly reproducible yet, but that's something we should work towards.

tools/docker-images/clp-package/build.sh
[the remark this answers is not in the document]
kirkrodrigues: Yeah, we should specify the version we require. I just don't know what version that is right now.

tools/docker-images/clp-package/build.sh
```diff
@@ -5,10 +5,25 @@ set -o pipefail
 
 script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 repo_root=${script_dir}/../../../
+iid_file="${repo_root}/build/clp-package-image.id"
+
+# Remove the previous image to save space, but only after the build so its contents can be reused.
+prev_image_id=""
+if [[ -f "$iid_file" ]]; then
+    prev_image_id=$(cat "$iid_file")
```
kirkrodrigues:
```suggestion
    prev_image_id=$(<"$iid_file")
```

tools/docker-images/clp-package/build.sh
```diff
@@ -5,10 +5,25 @@ set -o pipefail
...
+fi
+cleanup() {
```
kirkrodrigues: How about `remove_prev_image`?

tools/docker-images/clp-package/build.sh
```diff
@@ -5,10 +5,25 @@ set -o pipefail
...
+    if [[ -n "$prev_image_id" ]] && docker image inspect "$prev_image_id" >/dev/null 2>&1; then
+        echo "Removing previous image $prev_image_id"
```
kirkrodrigues:
```suggestion
        echo "Removing previous image $prev_image_id."
```

tools/docker-images/clp-package/build.sh
kirkrodrigues: Can we move this function to the top of the file, just after the `set` commands? That makes it easier to read the control flow imo.

kirkrodrigues:
> @kirkrodrigues if you think it makes sense to, in the same PR i can also remove `tools/docker-images/clp-execution-base-ubuntu-jammy`

Let's do it in another PR.

> we also discussed about cleaning up the image build process (that we don't have to copy everything into clp-package first; instead, we should specify the `build/<component>` paths in `COPY` commands in the Dockerfile and just directly copy from the `build/<component>`). while possible to be done in the same PR, i think it's cleaner to be done is a separate PR. if you agree, we can add an issue for that

Yeah, let's do it in another PR.

junhaoliao:
> If the user kills the script before it completes, the existing image will be deleted. Is that intentional?

yes, i think we want to invalidate the older image as soon as one requests to build a new one. otherwise, if the build fails and people are unaware of the failure status, they might still use the older image, thinking that's the latest

tools/docker-images/clp-package/build.sh
junhaoliao: > If we tag with the current timestamp, then identical image builds will result in two different tags pointing at the same image. Besides the clutter, that also means that the image can't be removed without a force flag.

not sure if this behaviour is possible. if i understand correctly, every time we run `docker build` a new image is built (we can tell by the different sha256 image id)

junhaoliao: maybe i understood it wrong. do you mind giving more steps on how this can be reproduced?

tools/docker-images/clp-package/build.sh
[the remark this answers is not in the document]
junhaoliao: it might be time consuming to figure out the exact version for the feature / syntax we use. maybe we can just take the latest version and put that into the docs? at least we can be sure that's a version we have validated

tools/docker-images/clp-package/build.sh
junhaoliao: my initial implementation of the unique id part was done with `uuidgen | head -c 4` though i think the chance of collision can be high.

later i replaced that with the timestamp because it's less likely a user can build multiple build in a second. (can probably change it to nanoseconds

taskfiles/docker-images.yaml
junhaoliao: * `cp` should work as well. i used `rsync` because it seems we don't use `cp` in any of our taskfiles, even for single file transfers. i vauguely remember it was some sort of convention we agreed on though i might have false memory
* The package start script needs to read this image id before deciding which container to use for launching package components. I am thinking that reading the id from `clp_home_directory/image.id` is better than `clp_home_directory/../clp-package-image.id`. though i don't really have a preference.

kirkrodrigues: > i vauguely remember it was some sort of convention we agreed on though i might have false memory

Hm, I don't really remember that, but I guess `rsync` can be more useful in terms of producing an exact copy with the same permissions, etc.

> The package start script needs to read this image id before deciding which container to use for launching package components.

That would be necessary in the next PR, right? Or is it necessary in this PR?

tools/docker-images/clp-package/build.sh
kirkrodrigues: > if i understand correctly, every time we run docker build a new image is built (we can tell by the different sha256 image id)

That's because `task package`'s output is not 100% reproducible---i.e., some files will change slightly. If you run the `build.sh` script alone twice, it should generate an image with the same ID.

tools/docker-images/clp-package/build.sh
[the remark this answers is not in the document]
kirkrodrigues: Generally, we support up to 2 years of backwards compatibility with OSes. Perhaps we can do the same for Docker?

kirkrodrigues (commented):
Responses.

kirkrodrigues:
> yes, i think we want to invalidate the older image as soon as one requests to build a new one. otherwise, if the build fails and people are unaware of the failure status, they might still use the older image, thinking that's the latest

Fair enough.

taskfiles/docker-images.yaml
junhaoliao: > > The package start script needs to read this image id before deciding which container to use for launching package components.
>
> That would be necessary in the next PR, right? Or is it necessary in this PR?

Right, the purpose is not immediately clear in this current PR but that will be necessary in the PR where we change the execution container selection. let's remove the line for now to help avoid confusion:
```suggestion
```

tools/docker-images/clp-package/build.sh
junhaoliao: actually i do mean running the `build.sh` gives a different sha256 id every time:
```bash
junhao@ASUS-X870E:~/workspace/2-clp/tools/docker-images/clp-package$ ./build.sh && echo '---' && cat /home/junhao/workspace/2-clp/build/clp-package-image.id && echo
[+] Building 0.9s (12/12) FINISHED                                                                                                                                                                                   docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                                                           0.0s
 => => transferring dockerfile: 585B                                                                                                                                                                                           0.0s 
 => [internal] load metadata for docker.io/library/ubuntu:jammy                                                                                                                                                                0.1s 
 => [internal] load .dockerignore                                                                                                                                                                                              0.0s 
 => => transferring context: 2B                                                                                                                                                                                                0.0s 
 => [base 1/6] FROM docker.io/library/ubuntu:jammy@sha256:4e0171b9275e12d375863f2b3ae9ce00a4c53ddda176bd55868df97ac6f21a6e                                                                                                     0.0s 
 => => resolve docker.io/library/ubuntu:jammy@sha256:4e0171b9275e12d375863f2b3ae9ce00a4c53ddda176bd55868df97ac6f21a6e                                                                                                          0.0s 
 => [internal] load build context                                                                                                                                                                                              0.6s 
 => => transferring context: 2.24MB                                                                                                                                                                                            0.6s
 => CACHED [base 2/6] WORKDIR /root                                                                                                                                                                                            0.0s
 => CACHED [base 3/6] COPY ./tools/docker-images/clp-package/setup-scripts ./setup-scripts                                                                                                                                     0.0s 
 => CACHED [base 4/6] RUN ./setup-scripts/install-prebuilt-packages.sh     && rm -rf ./setup-scripts/                                                                                                                          0.0s 
 => CACHED [base 5/6] RUN apt-get clean     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*                                                                                                                                   0.0s 
 => CACHED [base 6/6] COPY ./build/clp-package /opt/clp                                                                                                                                                                        0.0s 
 => CACHED [stage-1 1/1] COPY --from=base / /                                                                                                                                                                                  0.0s 
 => exporting to image                                                                                                                                                                                                         0.1s 
 => => exporting layers                                                                                                                                                                                                        0.0s 
 => => exporting manifest sha256:eb17ac6fd6ea60ce4eb1839a54f0bbc939b0d3dad201c2d9547d0b0ffd7fe2bf                                                                                                                              0.0s 
 => => exporting config sha256:41a6f9aae1dc947a3deeafcec0e82ab2891f2cb55ff0f9a3a0492cd5defd0739                                                                                                                                0.0s 
 => => exporting attestation manifest sha256:5cbe3ae1c541bb2237dbbb721e8635c31449bf9ef334bcbb691e98d4b882a168                                                                                                                  0.0s 
 => => exporting manifest list sha256:31e3696ede8077959e0dfe249f7be8caa42c12d2d351e0a93ba88f080b883e36                                                                                                                         0.0s 
 => => naming to docker.io/library/clp-package:dev-junhao-1759114001                                                                                                                                                           0.0s 
 => => unpacking to docker.io/library/clp-package:dev-junhao-1759114001                                                                                                                                                        0.0s 
Removing previous image sha256:754d92f3198297d74cae9ef9698f2d5f046392d6408158c4860ce820c7b08b17.                                                                                                                                    
Untagged: clp-package:dev-junhao-1759113990
Deleted: sha256:754d92f3198297d74cae9ef9698f2d5f046392d6408158c4860ce820c7b08b17
---
sha256:31e3696ede8077959e0dfe249f7be8caa42c12d2d351e0a93ba88f080b883e36
junhao@ASUS-X870E:~/workspace/2-clp/tools/docker-images/clp-package$ ./build.sh && echo '---' && cat /home/junhao/workspace/2-clp/build/clp-package-image.id && echo
[+] Building 0.9s (12/12) FINISHED                                                                                                                                                                                   docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                                                           0.0s
 => => transferring dockerfile: 585B                                                                                                                                                                                           0.0s 
 => [internal] load metadata for docker.io/library/ubuntu:jammy                                                                                                                                                                0.1s 
 => [internal] load .dockerignore                                                                                                                                                                                              0.0s 
 => => transferring context: 2B                                                                                                                                                                                                0.0s 
 => [base 1/6] FROM docker.io/library/ubuntu:jammy@sha256:4e0171b9275e12d375863f2b3ae9ce00a4c53ddda176bd55868df97ac6f21a6e                                                                                                     0.0s 
 => => resolve docker.io/library/ubuntu:jammy@sha256:4e0171b9275e12d375863f2b3ae9ce00a4c53ddda176bd55868df97ac6f21a6e                                                                                                          0.0s 
 => [internal] load build context                                                                                                                                                                                              0.6s 
 => => transferring context: 2.24MB                                                                                                                                                                                            0.6s
 => CACHED [base 2/6] WORKDIR /root                                                                                                                                                                                            0.0s
 => CACHED [base 3/6] COPY ./tools/docker-images/clp-package/setup-scripts ./setup-scripts                                                                                                                                     0.0s 
 => CACHED [base 4/6] RUN ./setup-scripts/install-prebuilt-packages.sh     && rm -rf ./setup-scripts/                                                                                                                          0.0s 
 => CACHED [base 5/6] RUN apt-get clean     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*                                                                                                                                   0.0s 
 => CACHED [base 6/6] COPY ./build/clp-package /opt/clp                                                                                                                                                                        0.0s 
 => CACHED [stage-1 1/1] COPY --from=base / /                                                                                                                                                                                  0.0s 
 => exporting to image                                                                                                                                                                                                         0.1s 
 => => exporting layers                                                                                                                                                                                                        0.0s 
 => => exporting manifest sha256:eb17ac6fd6ea60ce4eb1839a54f0bbc939b0d3dad201c2d9547d0b0ffd7fe2bf                                                                                                                              0.0s 
 => => exporting config sha256:41a6f9aae1dc947a3deeafcec0e82ab2891f2cb55ff0f9a3a0492cd5defd0739                                                                                                                                0.0s 
 => => exporting attestation manifest sha256:262a3a97f6ac6ebe1264878a49027dd9382532eda8aa5b8386a8496acb684c5c                                                                                                                  0.0s 
 => => exporting manifest list sha256:fda46ace138ddb9700a1dbdaf26bbde1577dd1e15f7159d05858e9c6a8d7b245                                                                                                                         0.0s 
 => => naming to docker.io/library/clp-package:dev-junhao-1759114003                                                                                                                                                           0.0s 
 => => unpacking to docker.io/library/clp-package:dev-junhao-1759114003                                                                                                                                                        0.0s 
Removing previous image sha256:31e3696ede8077959e0dfe249f7be8caa42c12d2d351e0a93ba88f080b883e36.                                                                                                                                    
Untagged: clp-package:dev-junhao-1759114001
Deleted: sha256:31e3696ede8077959e0dfe249f7be8caa42c12d2d351e0a93ba88f080b883e36
---
sha256:fda46ace138ddb9700a1dbdaf26bbde1577dd1e15f7159d05858e9c6a8d7b245
junhao@ASUS-X870E:~/workspace/2-clp/tools/docker-images/clp-package$ ./build.sh && echo '---' && cat /home/junhao/workspace/2-clp/build/clp-package-image.id && echo
[+] Building 0.9s (12/12) FINISHED                                                                                                                                                                                   docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                                                           0.0s
 => => transferring dockerfile: 585B                                                                                                                                                                                           0.0s 
 => [internal] load metadata for docker.io/library/ubuntu:jammy                                                                                                                                                                0.1s 
 => [internal] load .dockerignore                                                                                                                                                                                              0.0s 
 => => transferring context: 2B                                                                                                                                                                                                0.0s 
 => [base 1/6] FROM docker.io/library/ubuntu:jammy@sha256:4e0171b9275e12d375863f2b3ae9ce00a4c53ddda176bd55868df97ac6f21a6e                                                                                                     0.0s 
 => => resolve docker.io/library/ubuntu:jammy@sha256:4e0171b9275e12d375863f2b3ae9ce00a4c53ddda176bd55868df97ac6f21a6e                                                                                                          0.0s 
 => [internal] load build context                                                                                                                                                                                              0.6s 
 => => transferring context: 2.24MB                                                                                                                                                                                            0.6s
 => CACHED [base 2/6] WORKDIR /root                                                                                                                                                                                            0.0s
 => CACHED [base 3/6] COPY ./tools/docker-images/clp-package/setup-scripts ./setup-scripts                                                                                                                                     0.0s 
 => CACHED [base 4/6] RUN ./setup-scripts/install-prebuilt-packages.sh     && rm -rf ./setup-scripts/                                                                                                                          0.0s 
 => CACHED [base 5/6] RUN apt-get clean     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*                                                                                                                                   0.0s 
 => CACHED [base 6/6] COPY ./build/clp-package /opt/clp                                                                                                                                                                        0.0s 
 => CACHED [stage-1 1/1] COPY --from=base / /                                                                                                                                                                                  0.0s 
 => exporting to image                                                                                                                                                                                                         0.1s 
 => => exporting layers                                                                                                                                                                                                        0.0s 
 => => exporting manifest sha256:eb17ac6fd6ea60ce4eb1839a54f0bbc939b0d3dad201c2d9547d0b0ffd7fe2bf                                                                                                                              0.0s 
 => => exporting config sha256:41a6f9aae1dc947a3deeafcec0e82ab2891f2cb55ff0f9a3a0492cd5defd0739                                                                                                                                0.0s 
 => => exporting attestation manifest sha256:e1d31873b43c7ec6bcdeb4112671c6daf21b02b4019c1d40b6de746717c6d816                                                                                                                  0.0s 
 => => exporting manifest list sha256:b3a0bfc056f542a868cd1845256543dd585864fe0ca8d934f6a398bc07b52151                                                                                                                         0.0s 
 => => naming to docker.io/library/clp-package:dev-junhao-1759114005                                                                                                                                                           0.0s 
 => => unpacking to docker.io/library/clp-package:dev-junhao-1759114005                                                                                                                                                        0.0s 
Removing previous image sha256:fda46ace138ddb9700a1dbdaf26bbde1577dd1e15f7159d05858e9c6a8d7b245.                                                                                                                                    
Untagged: clp-package:dev-junhao-1759114003
Deleted: sha256:fda46ace138ddb9700a1dbdaf26bbde1577dd1e15f7159d05858e9c6a8d7b245
---
sha256:b3a0bfc056f542a868cd1845256543dd585864fe0ca8d934f6a398bc07b52151
```

junhaoliao: discussed offline - it appears that the above script generates a different hash every time only when Docker Desktop Windows + WSL is used. with a native Docker Engine installed inside WSL, the hash is consistent across docker build if the content inside `build/clp-package` is unchanged

---

a few facts:
1. if we run `docker build` without a `--tag` parameter, in the build logs it would appear that we are building an image with tag `moby-dangling`.
2. provided an id of an image, it's possible to add tags to the image afterwards
3. there's no atomic way to build an image, get the id of the built image, and add tag to the image with the id included in the tag name

---

how about this:
1. we read the image id file from `build/clp-package-image.id` into `$prev_image_id`, as we currently do.
2. we call `mktemp` to get a temporary file path, assigned to `$temp_iid_file`
3. `docker build --iidfile "$temp_iid_file" "$repo_root" --file "${script_dir}/Dockerfile"` (note we don't supply tags here. the build logs will show we're building for project `moby-dangling` but i guess we can live with that, so long the image is eventually tagged)
4. we `cat` the `$temp_iid_file` and store the value at `$new_iid`. fallback to an empty string if `cat` fails (the failure can possibly be caused by a failed docker build?)
5. if the `$new_iid` is non empty, truncate the `$new_iid` to 4 characters and tag the new image with `clp-package:dev-${USER}-${short_id}`
6. before the script exits, the `trap xxx EXIT` function is called. inside:
   1. if `$prev_image_id` differs from `$new_iid` (note: this is always true if `$new_iid` is empty), we `docker image remove $prev_image_id`.

kirkrodrigues: At step (3), it's possible two concurrent builds on the system create an image with tag `moby-dangling`, right? I think that will still succeed, but then after step (5) wouldn't we end up with two tags for the same image? I guess we need a step to remove the `moby-dangling` tag for the particular image we created.

junhaoliao: let me implement the proposal and monkey test a bit to see if there are issues. i could be wrong about how `docker image remove` works (or confused by the different behaviours in Docker Desktop)

junhaoliao: 1. it appears that the `moby-dangling` isn't a real project / tag --- Running `docker image ls` doesn't show the `moby-dangling` on the listed image, whether we manfully tag the image with our own tag or not.
2. the `moby-dangling` project only shows up in the build logs with Docker Desktop.

i have tested the script that's been pushed and didn't find any awkward issue, either with Docker Desktop or native Docker.

tools/docker-images/clp-package/build.sh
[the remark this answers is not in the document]
junhaoliao: https://containerd.io/releases/ 's LTS version is now `1.7` and it appears that the last retrievable version is `1.7.18` which was released on July 3, 2024 - https://download.docker.com/linux/ubuntu/dists/jammy/pool/stable/amd64/

i installed these versions on my machine and tested things fine:
```sh
sudo apt-get install -y \
  docker-ce=5:27.0.3-1~ubuntu.22.04~jammy \
  docker-ce-cli=5:27.0.3-1~ubuntu.22.04~jammy \
  containerd.io=1.7.18-1 \
  docker-buildx-plugin=0.15.1-1~ubuntu.22.04~jammy \
  docker-compose-plugin=2.28.1-1~ubuntu.22.04~jammy
```

tools/docker-images/clp-package/build.sh
```diff
@@ -3,12 +3,32 @@
...
 
+remove_prev_image_and_temp_file() {
+    if [[ -n "$prev_image_id" && "$prev_image_id" != "$new_image_id" ]]; then
+        if docker image inspect "$prev_image_id" >/dev/null 2>&1; then
+            echo "Removing previous image $prev_image_id."
+            docker image remove "$prev_image_id"
+        fi
+    fi
+    rm -f "$temp_iid_file"
+}
```
junhaoliao, line 14: this might be cleaner. though i'm not sure what bash syntaxes / conventions do we prefer
```suggestion
remove_temp_file_and_prev_image() {
    rm -f "$temp_iid_file"

    [[ -z "$prev_image_id" || "$prev_image_id" == "$new_image_id" ]] && return

    if docker image inspect "$prev_image_id" >/dev/null 2>&1; then
        echo "Removing previous image $prev_image_id."
        docker image remove "$prev_image_id"
    fi
}
```

junhaoliao:
> > @kirkrodrigues if you think it makes sense to, in the same PR i can also remove `tools/docker-images/clp-execution-base-ubuntu-jammy`
> 
> Let's do it in another PR.

This is added as part of #1357 , which add the mechanism to select the image build by the unique id generated in this current PR #1335 .

> > we also discussed about cleaning up the image build process (that we don't have to copy everything into clp-package first; instead, we should specify the `build/<component>` paths in `COPY` commands in the Dockerfile and just directly copy from the `build/<component>`). while possible to be done in the same PR, i think it's cleaner to be done is a separate PR. if you agree, we can add an issue for that
> 
> Yeah, let's do it in another PR.

This is strictly optimization and isn't blocking the Docker Compose migration, so this is of lower priority at the moment. @coderabbitai create an issue to track the request so we won't forget

tools/docker-images/clp-package/build.sh
```diff
@@ -3,17 +3,38 @@
...
 set -o pipefail
 
+remove_temp_file_and_prev_image() {
+    rm -f "$temp_iid_file"
+
+    [[ -z "$prev_image_id" || "$prev_image_id" == "$new_image_id" ]] && return
+
+    if docker image inspect "$prev_image_id" >/dev/null 2>&1; then
+        echo "Removing previous image $prev_image_id."
+        docker image remove "$prev_image_id"
```
junhaoliao, line 13: The image might fail to be removed because some running container might still be referencing it, which causes a non-zero exit code in the script.

When the image fails to be removed, we should either exit gracefully (with 0) with a warning message printed, or the caller should gracefully handle the non-exit code. Currently, the caller (the docker-images.yaml taskfile) doesn't handle this --- when the script fails, any commands following the script call (e.g., to copy the unique image id file into `build/clp-package`) would fail to execute.

tools/docker-images/clp-package/build.sh
```diff
@@ -22,3 +43,13 @@ then
 fi
 
 "${build_cmd[@]}"
+
+if [[ -s "$temp_iid_file" ]]; then
+    new_image_id="$(<"$temp_iid_file")"
+    echo "$new_image_id" > "$iid_file"
+
+    user="${USER:-$(id -un 2>/dev/null || whoami 2>/dev/null || echo unknown)}"
```
junhaoliao, line 51: There's one edge case we missed to handle -

When the username cannot be found on the system (e.g., the username is not specified in `/etc/passwd`), the user id would still be printed by `id -un` command. However, since the `id -un` command exits with non-zero code, the other command would continue to run, which may cause line breaks to show up in the middle of the `user` string. e.g.,

```
id -un: Prints `123\n`; Exits with non-0
whoami: Prints nothing; Exits with non-9
echo unknown: Prints "unknown"

then `user` becomes: `123\nunknown`
```

PR #1336, opened 2025-09-25
feat(webui): Provide improved calculation of time range for queries:

hoophalab:
- Treat the "All Time" time range as spanning the time range of existing archives.
- Determine the time range when the submit button is clicked.
- Include logs compressed without a timestamp key.

# Description

From the issue page:
Currently, the "All Time" time range in the search UI defaults to [UNIX epoch, current time + 1 year]. Thus, when a user enters a query and the results are displayed, they are usually shown as a single graph near the end of the timeline.
Instead, it would be better to treat "All Time" as spanning the time range of the existing archives.

When a query is run for "All Time", this PR fetches the time range of the existing archives in clp-database; translates that into a query for the time range of the existing archives.

clpsalltime

# Checklist

  breaking change.

# Validation performed

Search for `*` on the search page with the "all time" range option selected:

1. clp-text: The timeline's time range is the range of all logs.
2. clp-json: The timeline's time range is the range of all logs in the dataset.
3. clp-json: Different datasets have different time ranges.
4. logs w/o timestamp key are shown

## Summary by CodeRabbit

- New Features
  - “All time” now auto-detects earliest/latest logs via a backend fetch; user-visible warnings appear if dataset is missing or fetch fails.
  - Submit button now computes and submits time-range bounds (or nulls for All time) and shows inline message feedback on errors.

- Bug Fixes
  - Improved submit flow resilience with explicit error handling and fallback to a default time range.

- Refactor
  - Preset selections no longer auto-update the visible range (CUSTOM still opens the picker); time-range defaults consolidated.

---
  - https://app.asana.com/0/0/1210486680627843

### DIFF
--- a/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
+++ b/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
@@ -36,7 +36,7 @@ from job_orchestration.scheduler.job_config import (
     PathsToCompress,
     S3InputConfig,
 )
-from job_orchestration.scheduler.scheduler_data import CompressionTaskResult
+from job_orchestration.scheduler.task_result import CompressionTaskResult
 
 
 def update_compression_task_metadata(db_cursor, task_id, kv):
--- a/components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
+++ b/components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
@@ -10,7 +10,6 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional, Set
 
 import brotli
-import celery
 import msgpack
 from clp_package_utils.general import CONTAINER_INPUT_LOGS_ROOT_DIR
 from clp_py_utils.clp_config import (
@@ -30,8 +29,9 @@ from clp_py_utils.compression import validate_path_and_get_info
 from clp_py_utils.core import read_yaml_config_file
 from clp_py_utils.s3_utils import s3_get_object_metadata
 from clp_py_utils.sql_adapter import SQL_Adapter
-from job_orchestration.executor.compress.celery_compress import compress
 from job_orchestration.scheduler.compress.partition import PathsToCompressBuffer
+from job_orchestration.scheduler.compress.task_manager.celery_task_manager import CeleryTaskManager
+from job_orchestration.scheduler.compress.task_manager.task_manager import TaskManager
 from job_orchestration.scheduler.constants import (
     CompressionJobStatus,
     CompressionTaskStatus,
@@ -45,7 +45,6 @@ from job_orchestration.scheduler.job_config import (
 )
 from job_orchestration.scheduler.scheduler_data import (
     CompressionJob,
-    CompressionTaskResult,
 )
 from job_orchestration.scheduler.utils import kill_hanging_jobs
 from pydantic import ValidationError
@@ -217,6 +216,7 @@ def search_and_schedule_new_tasks(
     db_conn,
     db_cursor,
     clp_metadata_db_connection_config: Dict[str, Any],
+    task_manager: TaskManager,
 ):
     """
     For all jobs with PENDING status, splits the job into tasks and schedules them.
@@ -224,6 +224,7 @@ def search_and_schedule_new_tasks(
     :param db_conn:
     :param db_cursor:
     :param clp_metadata_db_connection_config:
+    :param task_manager:
     """
     global scheduled_jobs
 
@@ -356,7 +357,6 @@ def search_and_schedule_new_tasks(
             tag_ids = [tags["tag_id"] for tags in db_cursor.fetchall()]
             db_conn.commit()
 
-        task_instances = []
         for task_idx, task in enumerate(tasks):
             db_cursor.execute(
                 f"""
@@ -369,12 +369,9 @@ def search_and_schedule_new_tasks(
             db_conn.commit()
             task["task_id"] = db_cursor.lastrowid
             task["tag_ids"] = tag_ids
-            task_instances.append(compress.s(**task))
-        tasks_group = celery.group(task_instances)
+        result_handle = task_manager.submit(tasks)
 
-        job = CompressionJob(
-            id=job_id, start_time=start_time, async_task_result=tasks_group.apply_async()
-        )
+        job = CompressionJob(id=job_id, start_time=start_time, result_handle=result_handle)
         db_cursor.execute(
             f"""
             UPDATE {COMPRESSION_TASKS_TABLE_NAME}
@@ -384,13 +381,6 @@ def search_and_schedule_new_tasks(
         db_conn.commit()
 
         scheduled_jobs[job_id] = job
-
-
-def get_results_or_timeout(result):
-    try:
-        return result.get(timeout=0.1)
-    except celery.exceptions.TimeoutError:
-        return None
 
 
 def poll_running_jobs(db_conn, db_cursor):
@@ -407,14 +397,13 @@ def poll_running_jobs(db_conn, db_cursor):
         error_message = ""
 
         try:
-            returned_results = get_results_or_timeout(job.async_task_result)
+            returned_results = job.result_handle.get_result()
             if returned_results is None:
                 continue
 
             duration = (datetime.datetime.now() - job.start_time).total_seconds()
             # Check for finished jobs
             for task_result in returned_results:
-                task_result = CompressionTaskResult.model_validate(task_result)
                 if task_result.status == CompressionTaskStatus.SUCCEEDED:
                     logger.info(
                         f"Compression task job-{job_id}-task-{task_result.task_id} completed in"
@@ -497,6 +486,8 @@ def main(argv):
     logger.info(f"Starting {COMPRESSION_SCHEDULER_COMPONENT_NAME}")
     sql_adapter = SQL_Adapter(clp_config.database)
 
+    task_manager = CeleryTaskManager()
+
     try:
         killed_jobs = kill_hanging_jobs(sql_adapter, SchedulerType.COMPRESSION)
         if killed_jobs is not None:
@@ -521,6 +512,7 @@ def main(argv):
                         db_conn,
                         db_cursor,
                         clp_metadata_db_connection_config,
+                        task_manager,
                     )
                 poll_running_jobs(db_conn, db_cursor)
                 time.sleep(clp_config.compression_scheduler.jobs_poll_delay)
--- a/components/job-orchestration/job_orchestration/scheduler/compress/task_manager/celery_task_manager.py
+++ b/components/job-orchestration/job_orchestration/scheduler/compress/task_manager/celery_task_manager.py
@@ -1,0 +1,27 @@
+from __future__ import annotations
+
+from typing import Any
+
+import celery
+from job_orchestration.executor.compress.celery_compress import compress
+from job_orchestration.scheduler.compress.task_manager.task_manager import TaskManager
+from job_orchestration.scheduler.task_result import CompressionTaskResult
+
+
+class CeleryTaskManager(TaskManager):
+
+    class ResultHandle(TaskManager.ResultHandle):
+        def __init__(self, celery_result: celery.result.GroupResult) -> None:
+            self._celery_result: celery.result.GroupResult = celery_result
+
+        def get_result(self, timeout: float = 0.1) -> list[CompressionTaskResult] | None:
+            try:
+                results = self._celery_result.get(timeout=timeout)
+                return [CompressionTaskResult.model_validate(res) for res in results]
+            except celery.exceptions.TimeoutError:
+                return None
+
+    def submit(self, task_params: list[dict[str, Any]]) -> TaskManager.ResultHandle:
+        task_instances = [compress.s(**params) for params in task_params]
+        task_group = celery.group(task_instances)
+        return CeleryTaskManager.ResultHandle(task_group.apply_async())
--- a/components/job-orchestration/job_orchestration/scheduler/compress/task_manager/spider_task_manager.py
+++ b/components/job-orchestration/job_orchestration/scheduler/compress/task_manager/spider_task_manager.py
@@ -1,0 +1,47 @@
+from __future__ import annotations
+
+import json
+from typing import Any
+
+import spider_py
+from job_orchestration.executor.compress.spider_compress import compress
+from job_orchestration.scheduler.compress.task_manager.task_manager import TaskManager
+from job_orchestration.scheduler.task_result import CompressionTaskResult
+from job_orchestration.utils.spider_utils import int8_list_to_utf8_str, utf8_str_to_int8_list
+from spider_py.client.job import Job
+
+
+class SpiderTaskManager(TaskManager):
+
+    class ResultHandle(TaskManager.ResultHandle):
+        def __init__(self, spider_job: Job) -> None:
+            self._spider_job: Job = spider_job
+
+        def get_result(self, timeout: float = 0.1) -> list[CompressionTaskResult] | None:
+            job_results = self._spider_job.get_results()
+            if job_results is None:
+                return None
+            return [
+                CompressionTaskResult.model_validate_json(int8_list_to_utf8_str(task_result))
+                for task_result in job_results
+            ]
+
+    def __init__(self, storage_url: str) -> None:
+        self._driver = spider_py.Driver(storage_url)
+
+    def submit(self, task_params: list[dict[str, Any]]) -> TaskManager.ResultHandle:
+        job = spider_py.group(
+            [compress for _ in range(len(task_params))],
+        )
+        job_args = []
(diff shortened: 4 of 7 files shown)


components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/timeRangeSql.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+const buildClpsTimeRangeSql = (datasetName: string): string => {
+    return `SELECT
+  MIN(${CLP_ARCHIVES_TABLE_COLUMN_NAMES.BEGIN_TIMESTAMP}) AS begin_timestamp,
+  MAX(${CLP_ARCHIVES_TABLE_COLUMN_NAMES.END_TIMESTAMP}) AS end_timestamp
+FROM ${settings.SqlDbClpTablePrefix}${datasetName}_${SqlTableSuffix.ARCHIVES}`;
+};
+
+export {
+    buildClpsTimeRangeSql, buildClpTimeRangeSql,
```
junhaoliao:
```suggestion
    buildClpsTimeRangeSql,
    buildClpTimeRangeSql,
```

components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/index.tsx
```diff
@@ -24,26 +32,74 @@ import styles from "./index.module.css";
...
+                begin_timestamp: Nullable<number>;
+                end_timestamp: Nullable<number>;
+            }[]
+        >(sql);
+        const [timestamps] = resp.data;
+        if ("undefined" === typeof timestamps ||
+            null === timestamps.begin_timestamp ||
+            null === timestamps.end_timestamp
+        ) {
+            throw new Error();
```
junhaoliao:
```suggestion
            throw new Error("Unable to get All Time range");
```
alternatively, use the `message` API from Antd to display an error message in the UI?

hoophalab: there is an automatic fallback in the button handling function

hoophalab: this error is to tell the caller that time range isn't available.

junhaoliao: sorry i just read the handling again and think we don't have to throw here. can we change the return type of fetchAllTimeTimeRange() to be `Promise<Nullable<[Dayjs, Dayjs]>>`, return null when the fields are not found, then the caller can just check for null instead of catching the error?

components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/index.tsx
```diff
@@ -24,26 +32,74 @@ import styles from "./index.module.css";
...
      * Submits search query.
      */
-    const handleSubmitButtonClick = useCallback(() => {
+    const handleSubmitButtonClick = useCallback(async () => {
+        let updatedTimeRange: [Dayjs, Dayjs];
+        if (timeRangeOption === TIME_RANGE_OPTION.ALL_TIME) {
+            try {
+                updatedTimeRange = await fetchAllTimeTimeRange();
+            } catch (err: unknown) {
+                console.error("Cannot fetch all time time range.", err);
```
junhaoliao: i think it's better to show it in the UI for such errors, have we consider Antd's `message` API?

components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/index.tsx
```diff
@@ -85,9 +146,16 @@ const SubmitButton = () => {
...
+                disabled={
+                    isQueryStringEmpty ||
+                    isNoDatasetsAndClpS ||
+                    searchUiState === SEARCH_UI_STATE.QUERY_ID_PENDING
+                }
+                onClick={() => {
+                    handleSubmitButtonClick().catch((err: unknown) => {
+                        throw err;
+                    });
+                }}
```
junhaoliao, line 141: mind explaining why this change is needed?

components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/index.tsx
```diff
@@ -24,26 +32,74 @@ import styles from "./index.module.css";
...
+            sql = buildClpsTimeRangeSql(selectDataset ?? "default");
+        }
+        const resp = await querySql<
+            {
+                begin_timestamp: Nullable<number>;
+                end_timestamp: Nullable<number>;
+            }[]
+        >(sql);
+        const [timestamps] = resp.data;
+        if ("undefined" === typeof timestamps ||
```
junhaoliao: by the way, for validation, do you know how i can reproduce this missing timestamp begin /end response?

components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/index.tsx
hoophalab: because the handlesubmitbuttonclicked is changed to an async, and eslint reports an error if we pass a function with a return value to onClick

components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/index.tsx
hoophalab: This code is copied from the ingestion page and ultimately comes from the old webui.

There is no way to reproduce them in runtime as far as I know, but the undefined check is to satisfy typescript type checks

components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/index.tsx
hoophalab: yeah, that could also work, but I don't feel we should use null value to represent error.
e.g. when there can be an error in cpp, we won't change the return type to std::optional, but instead, we probably simply throw an error or wrap it in a proper boost::result type

also, axios can throw an error on error http status code. To properly implement the idea of null, we need to wrap the whole function in try catch

hoophalab: I'm open if we believe null is better here

davemarco:
Do you think we should update TIME_RANGE_OPTION_DAYJS_MAP in /home/davidmawk/clp/components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/utils.tsx to call the fetch timestamp function? Then we could upgrade the whole UI, so that it calls the function in TIME_RANGE_OPTION_DAYJS_MAP when we submit a query. Right now the timestamps get outdated, but we could potentially fix this for all options, not just allTime

components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/index.tsx
```diff
@@ -14,36 +18,93 @@ import {computeTimelineConfig} from "../../../SearchResults/SearchResultsTimelin
...
     const queryString = useSearchStore((state) => state.queryString);
     const selectDataset = useSearchStore((state) => state.selectDataset);
-    const updateCachedDataset = useSearchStore((state) => state.updateCachedDataset);
+    const updateCachedDataset = useSearchStore(
+        (state) => state.updateCachedDataset,
+    );
+    const [messageApi, contextHolder] = message.useMessage();
+
+    const fetchAllTimeTimeRange = useCallback(async (): Promise<
+        [Dayjs, Dayjs]
```
davemarco: i think can be moved to another files and imported

hoophalab: sure

hoophalab:
> Do you think we should update TIME_RANGE_OPTION_DAYJS_MAP in /home/davidmawk/clp/components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/utils.tsx to call the fetch timestamp function? Then we could upgrade the whole UI, so that it calls the function in TIME_RANGE_OPTION_DAYJS_MAP when we submit a query. Right now the timestamps get outdated, but we could potentially fix this for all options, not just allTime

sure I'll do that

components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/index.tsx
```diff
@@ -24,26 +31,50 @@ import styles from "./index.module.css";
...
      */
-    const handleSubmitButtonClick = useCallback(() => {
+    const handleSubmitButtonClick = useCallback(async () => {
+        let newTimeRange: [Dayjs, Dayjs];
+        if (timeRangeOption !== TIME_RANGE_OPTION.CUSTOM) {
+            try {
+                newTimeRange = await TIME_RANGE_OPTION_DAYJS_MAP[timeRangeOption](selectDataset);
+            } catch {
+                messageApi.warning(
+                    'Cannot fetch the time range for "All Time". Fallback to the unix epoch.',
```
davemarco: now that we changed to all inputs, this could maybe be innacurate even tho unliekly. I think cannot fetch timerange is fine for log. Also i dont think we need `TIME_RANGE_SINCE_UNIX_EPOCH` here (we can do this in mapping function, i think we can just return and submit nothing if query fails since there is log

components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/index.tsx
```diff
@@ -54,21 +85,35 @@ const SubmitButton = () => {
             ignoreCase: false === queryIsCaseSensitive,
             queryString: queryString,
             timeRangeBucketSizeMillis: newTimelineConfig.bucketDuration.asMilliseconds(),
-            timestampBegin: timeRange[0].valueOf(),
-            timestampEnd: timeRange[1].valueOf(),
+            ...(
+                timeRangeOption === TIME_RANGE_OPTION.ALL_TIME ?
+                    {
+                        timestampBegin: null,
```
davemarco, line 91: Why not just use newTimeRange and leave code as it was?

components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/timeRangeSql.ts
davemarco, line 1: i think u can just name sql.ts 

if u want fancier name i think time-range-sql.ts is convention

components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/utils.tsx
```diff
@@ -23,70 +42,95 @@ enum TIME_RANGE_OPTION {
...
         dayjs().utc()
             .startOf("year"),
         dayjs().utc(),
     ],
-    [TIME_RANGE_OPTION.ALL_TIME]: () => [
-        dayjs(0).utc(),
-        dayjs().utc()
-            .add(1, "year"),
-    ],
+    [TIME_RANGE_OPTION.ALL_TIME]: async (selectDataset) => {
```
davemarco: I still think you can put this as a function in another file, then just call it here. Then it will also be easier to use getState, and u can just pull in the dataset and not pass the string dependency here

components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/utils.tsx
```diff
@@ -23,70 +42,95 @@ enum TIME_RANGE_OPTION {
...
+        let sql: string;
+        if (CLP_STORAGE_ENGINES.CLP === SETTINGS_STORAGE_ENGINE) {
+            sql = buildClpTimeRangeSql();
+        } else {
+            sql = buildClpsTimeRangeSql(selectDataset ?? "default");
```
davemarco: maybe we exit and log if the dataset is null

components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/utils.tsx
```diff
@@ -23,70 +42,95 @@ enum TIME_RANGE_OPTION {
...
+        const resp = await querySql<
+            {
+                begin_timestamp: Nullable<number>;
+                end_timestamp: Nullable<number>;
+            }[]
+        >(sql);
+        const [timestamps] = resp.data;
+        if ("undefined" === typeof timestamps ||
+              null === timestamps.begin_timestamp ||
+              null === timestamps.end_timestamp
```
davemarco: Do u know what happens if there is no timestamp key. If it just returns null, then there shouldnt be an error?  I think here we can use your TIME_RANGE_SINCE_UNIX_EPOCH, if the result is nulls, just use this. instead of throwing error, and catching later and setting to TIME_RANGE_SINCE_UNIX_EPOCH

components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/utils.tsx
```diff
@@ -1,9 +1,28 @@
...
+import {
+    buildClpsTimeRangeSql,
+    buildClpTimeRangeSql,
+} from "./timeRangeSql";
+
 
 dayjs.extend(utc);
 
+
+const TIME_RANGE_SINCE_UNIX_EPOCH: [Dayjs, Dayjs] = [
```
davemarco: how about DEFAULT_TIME_RANGE, and rename current default time range to DEFAULT_TIME_RANGE_OPTION?

components/webui/common/src/schemas/search.ts
```diff
@@ -20,8 +20,10 @@ const QueryJobCreationSchema = Type.Object({
     ignoreCase: Type.Boolean(),
     queryString: StringSchema,
     timeRangeBucketSizeMillis: Type.Integer(),
-    timestampBegin: Type.Integer(),
-    timestampEnd: Type.Integer(),
+    timestampBegin: Type.Union([Type.Null(),
```
davemarco, line 23: do we need this change, if we get rid of submitting null? unless of course we keep it

components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/index.tsx
davemarco: is this something do to with no timestamp key?

hoophalab: this is to do third item. We pass null in order to include logs compressed without a timestamp key.

components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/utils.tsx
hoophalab: > Do u know what happens if there is no timestamp key.
It just returns null?

axios can also throw errors. We need to try-catch later anyway.
I guess `null === timestamps.begin_timestamp` probably should throw an error because the sql returns an unexpected result

components/webui/common/src/schemas/search.ts
hoophalab: yeah, this is to include data w/o timestampkey

components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/index.tsx
hoophalab: Showing a notification is junhao's idea, see previous comments.

I changed the notification text

components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/utils.tsx
davemarco: i think if we are saying we can now search logs with no timestamps. then this is default behaviour, so like user shouldnt get error. I think if default behaviour, it shouldnt throw error. I still think better to just check if null === timestamps.begin_timestamp || null === timestamps.begin_timestamp return default?

components/webui/client/src/pages/SearchPage/SearchControls/SearchButton/SubmitButton/index.tsx
davemarco: if u take my other comment. I think you can remove the fallback part from notification, and then just exit and not submit anything. I think the notification is fine since this is serious error and it means something is broken

davemarco (changes requested):
reponded to ur question

davemarco:
`feat(webui): Provide improved calculation of time range for queries:`

davemarco (approved):

junhaoliao (approved):
nice job. deferring to @davemarco 's review

PR #1338, opened 2025-09-25
refactor(clp-package): Upgrade Pydantic to V2; Update existing code to match V2 syntax.

junhaoliao:
# Description

WIP - the PR is created for diff purpose

# Checklist

  breaking change.

# Validation performed


components/clp-package-utils/poetry.lock
junhaoliao, line 1: generate by updating the `pyproject.toml` then running `poetry update pydantic`

components/clp-py-utils/poetry.lock
junhaoliao, line 1: generate by updating the `pyproject.toml` then running `poetry update pydantic`

components/job-orchestration/poetry.lock
junhaoliao, line 1: generate by updating the `pyproject.toml` then running `poetry update pydantic`

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -124,28 +131,30 @@ class Package(BaseModel):
     storage_engine: str = "clp"
     query_engine: str = "clp"
 
-    @validator("storage_engine")
-    def validate_storage_engine(cls, field):
-        if field not in VALID_STORAGE_ENGINES:
+    @field_validator("storage_engine")
```
Eden-D-Zhang, line 134: The Pydantic `field_validator` has four modes (`after`, `before`, `plain`, `wrap`) and defaults to `after`. It might be worth including `mode='after'` so that it is more clear.
```suggestion
    @field_validator("storage_engine", mode='after')
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -295,53 +307,59 @@ class CompressionScheduler(BaseModel):
...
+        return value
 
-    @validator("port")
-    def validate_port(cls, field):
-        if not field > 0:
-            raise ValueError(f"{field} is not greater than zero")
-        return field
+    @field_validator("port")
+    @classmethod
+    def validate_port(cls, value):
```
Eden-D-Zhang, line 345: Unless the upper bound is intentionally left unchecked, the existing `_validate_port` should be used here.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -385,28 +404,32 @@ class Reducer(BaseModel):
...
+
+    @field_validator("logging_level")
+    @classmethod
+    def validate_logging_level(cls, value):
+        _validate_logging_level(cls, value)
+        return value
+
+    @field_validator("base_port")
+    @classmethod
+    def validate_base_port(cls, value):
```
Eden-D-Zhang, line 433: Same with `_validate_port` goes here.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -504,11 +533,12 @@ class AwsAuthentication(BaseModel):
...
     credentials: Optional[S3Credentials] = None
 
-    @root_validator(pre=True)
-    def validate_authentication(cls, values):
-        auth_type = values.get("type")
-        profile = values.get("profile")
-        credentials = values.get("credentials")
+    @model_validator(mode="before")
+    @classmethod
+    def validate_authentication(cls, data):
```
Eden-D-Zhang, line 567: Other uses of `model_validator` have parameter (`self`) instead of (`cls`, `data`). Consider updating this, using `self.<field>` instead of `data.get("<field>")` for retrieving variables.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -532,42 +562,45 @@ class S3Config(BaseModel):
...
     type: Literal[StorageType.FS.value] = StorageType.FS.value
     directory: pathlib.Path
 
-    @validator("directory")
-    def validate_directory(cls, field):
-        if "" == field:
+    @field_validator("directory")
+    @classmethod
+    def validate_directory(cls, value):
+        if "" == value:
```
Eden-D-Zhang: Is this comparison valid? Directory is pathlib Path, make sure this is checking against empty Path instead of empty string.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -577,29 +610,31 @@ class S3Storage(BaseModel):
     s3_config: S3Config
     staging_directory: pathlib.Path
 
-    @validator("staging_directory")
-    def validate_staging_directory(cls, field):
-        if "" == field:
+    @field_validator("staging_directory")
+    @classmethod
+    def validate_staging_directory(cls, value):
+        if "" == value:
```
Eden-D-Zhang: Same here, double check empty Path comparison.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -657,41 +692,47 @@ class ArchiveOutput(BaseModel):
...
-        return field
+        return value
 
-    @validator("compression_level")
-    def validate_compression_level(cls, field):
-        if field < 1 or 19 < field:
+    @field_validator("compression_level")
+    @classmethod
+    def validate_compression_level(cls, value):
+        if value < 1 or 19 < value:
```
Eden-D-Zhang:
```suggestion
        if 1 > value or 19 < value:
```

components/clp-py-utils/clp_py_utils/clp_config.py
Eden-D-Zhang, line 717: I noticed the `SweepInterval` class was changed to use `Field()`. It might make sense to do the same for the `target_<*>_size` fields here. i.e `target_archive_size: int = Field(default = 256 * 1024 * 1024, gt = 0)`. Same goes for `StreamOutput`.

components/job-orchestration/job_orchestration/scheduler/scheduler_data.py
```diff
@@ -16,22 +16,23 @@
...
 class CompressionJob(BaseModel):
     id: int
     start_time: datetime.datetime
-    async_task_result: Any
+    async_task_result: Optional[Any] = None
 
 
 class CompressionTaskResult(BaseModel):
     task_id: int
     status: int
```
Eden-D-Zhang, line 30:
```suggestion
    status: CompressionTaskStatus
```

components/job-orchestration/job_orchestration/scheduler/job_config.py
Eden-D-Zhang:
```suggestion
    empty_directories: typing.Optional[typing.List[str]] = None
```

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -117,7 +117,7 @@ def handle_job(sql_adapter: SQL_Adapter, clp_io_config: ClpIoConfig, no_progress
     ) as db_cursor:
         try:
             compressed_clp_io_config = brotli.compress(
-                msgpack.packb(clp_io_config.dict(exclude_none=True, exclude_unset=True)), quality=4
+                msgpack.packb(clp_io_config.model_dump(exclude_none=True, exclude_unset=True)), quality=4
```
Eden-D-Zhang: Wrap this line for length?

components/clp-py-utils/clp_py_utils/clp_config.py
Eden-D-Zhang: This would apply to all `field_validator`, not sure if this is worth doing but thought I should mention it.

junhaoliao: Good point. For most fields, I think it's fine to assume we always validate the parsed value unless specified otherwise. it's better to omit the mode imo

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: good catch. it should be use the helper

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: The variable naming conventions are taken from the Pydantic docs: https://docs.pydantic.dev/latest/concepts/validators/#model-validators

That why `self` is used with `mode="after"` is because the type of the `self` argument can be assumed after parsing is complete. For the `mode="before"` validators, we can't assume the type on `data` (most of the time it's just a dict) because some of the fields might not be generated yet for example

components/job-orchestration/job_orchestration/scheduler/job_config.py
junhaoliao: (not directly related)

@kirkrodrigues not sure if it was discussed previously - did we intentionally do qualified import (module import) or maybe we should do direct import for conciseness in such typing info? how often do we expect those names to encounter conflicts?

kirkrodrigues: Don't think it was intentional to keep it this way. It was probably just used once at the time of writing and then it wasn't updated in future PRs to be a direct import.

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: alternatively, i think we can declare the port field with `Field(gt=0, lt=2**16)`. maybe in a separate PR

components/clp-py-utils/clp_py_utils/clp_config.py
Eden-D-Zhang: I see, makes sense.

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: I thought about removing the validator and instead write:
```
class FsStorage(BaseModel):
    # ...
    directory: pathlib.Path = Field(min_length=1, description="directory cannot be empty")
```
however that `min_length=1` check didn't raise any alert on an `""` value because the type of `directory` is `pathlib.Path` rather than a string

i rewrote the validator with `mode="before"` instead

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: took a similar solution - discussed at https://github.com/y-scope/clp/pull/1338#discussion_r2383350645

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: our coding guidelines do prefer YODA; however, variable assignment isn't possible in if-conditions in Python, and the linters only have rules to disallow YODA rather than enforcing usage of YODA. For consistency, it was decided that we should avoid YODA in Python so we don't necessary have two ways to write the same code

---

@LinZhihao-723 i vaguely remember we were discussing the linter check that disallows YODA in Python. do you know why it's not caught here?

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: This is definitely worth considering. This PR has been huge and blocking another team. Since the improvement is non-functional, do you mind if we do it in a separate PR to clean up all such validators?

components/job-orchestration/job_orchestration/scheduler/scheduler_data.py
junhaoliao: @jackluo923 do you remember why we added the below validator? i.e., in what case do we expect below to throw?
```
    @field_validator("status")
    @classmethod
    def valid_status(cls, field):
        supported_status = [CompressionTaskStatus.SUCCEEDED, CompressionTaskStatus.FAILED]
        if field not in supported_status:
            raise ValueError(f'must be one of the following {"|".join(supported_status)}')
        return field
```

Since the `CompressionTaskResult` is not user-facing, maybe we can just remove the validator and type `status` as `CompressionTaskStatus` as @Eden-D-Zhang sugguested?

components/clp-py-utils/clp_py_utils/clp_config.py
Eden-D-Zhang: Yeah a separate PR definitely makes more sense. Should we create an issue for it?

junhaoliao: sure. @coderabbitai please help create an issue

components/clp-py-utils/clp_py_utils/clp_config.py
kirkrodrigues: Those linting rules aren't turned on for the python modules in the package due to time constraints.

junhaoliao: let's create an issue to track the effort of enabling those rules @coderabbitai

components/job-orchestration/job_orchestration/scheduler/job_config.py
junhaoliao: @coderabbitai create an issue to track the request to remove the `typing` qualififer

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -177,26 +186,35 @@ class Database(BaseModel):
...
-    def validate_database_host(cls, field):
-        if "" == field:
+    @field_validator("host")
+    @classmethod
+    def validate_host(cls, value):
+        if "" == value:
             raise ValueError("database.host cannot be empty.")
-        return field
+        return value
+
```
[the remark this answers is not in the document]
junhaoliao: This will be refactored in #1342

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -385,28 +415,31 @@ class Reducer(BaseModel):
...
+    def validate_base_port(cls, value):
+        _validate_port(cls, value)
+        return value
+
+    @field_validator("upsert_interval")
+    @classmethod
+    def validate_upsert_interval(cls, value):
+        if not value > 0:
+            raise ValueError(f"{value} is not greater than zero")
+        return value
```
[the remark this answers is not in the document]
junhaoliao: This should be refactored in #1342

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -532,42 +591,44 @@ class S3Config(BaseModel):
...
-    @validator("directory")
-    def validate_directory(cls, field):
-        if "" == field:
-            raise ValueError("directory cannot be empty")
-        return field
+    @field_validator("directory", mode="before")
+    @classmethod
+    def validate_directory(cls, value):
+        _validate_directory(value)
+        return value
```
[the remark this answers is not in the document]
junhaoliao: It's confirmed that errors are not reported on the default values, likely due to that the validator is specified in the base class and the default values of type `Path` are assigned in the subclasses

components/job-orchestration/job_orchestration/scheduler/scheduler_data.py
jackluo923: I don't know. This is where I tried to be safe, thrown an error just in case.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -577,29 +638,30 @@ class S3Storage(BaseModel):
...
         if not hasattr(s3_config, "key_prefix"):
             raise ValueError("s3_config must have field key_prefix")
         key_prefix = s3_config.key_prefix
         if "" == key_prefix:
             raise ValueError("s3_config.key_prefix cannot be empty")
         if not key_prefix.endswith("/"):
             raise ValueError('s3_config.key_prefix must end with "/"')
-        return values
+
+        return self
```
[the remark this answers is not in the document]
junhaoliao: @Eden-D-Zhang do you recall why we didn't directly specify a `@field_validator("key_prefix")` inside `S3Config`? like i see the other fields at `S3Config` have validators

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -981,21 +1048,15 @@ def dump_to_primitive_dict(self):
...
-    def validate_presto_config(cls, values):
-        package = values.get("package")
-        if not isinstance(package, Package):
-            # Skip validation since `package` is not a valid `Package` (Pydantic will validate it
-            # later and throw an error).
-            return values
-
-        query_engine = package.get("query_engine")
-        presto = values.get("presto")
+    @model_validator(mode="after")
```
junhaoliao, line 1060: cc @davemarco / @hoophalab for awareness of the change

components/job-orchestration/job_orchestration/scheduler/scheduler_data.py
junhaoliao: Let's leave this unchanged in this PR. @coderabbitai create an issue to track the effort of whether it's possible to rewrite the class without pydantic.BaseModel . note in the issue that we should validate throughly to ensure no regression is caused by lifting the checks / typing

components/clp-py-utils/clp_py_utils/clp_config.py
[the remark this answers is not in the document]
Eden-D-Zhang: I don’t remember exactly why off the top of my head, but I do remember having to change it to a ‘root_validator’ because it wasn’t working as a standard one. I can look back at the original PR to double check.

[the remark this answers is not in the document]
Eden-D-Zhang: So I changed `S3InputConfig` in `job_config.py` to inherit `S3Config` (to cut down on repeated code I guess?). Because the `S3InputConfig` could have an empty `key_prefix` and be valid (I believe the key prefixes are relative to the URL provided, so compressing a single file would have no relative prefix), the existing `@validator("key_prefix")` was failing valid configs. So I moved the validation of the key prefix to `S3Storage` since `S3Config` only needs to validate `key_prefix` when it's under `S3Storage`. 

I think we agreed then that that was preferable, but I'm not sure I agree now.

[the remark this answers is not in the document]
junhaoliao: I see. In some recent offline discussions, we discussed that the s3 input handling might need to differ from s3 output handling, so we might want to remove the inheritance relationship from the input config. And then it can be possible to move the validator inside the s3 config class.

Before we (properly design and) implement the new s3 input handling, I think it’s fine to keep those models untouched. So it still makes a lot of sense if we want to keep the validation happen only inside S3Storage. Still, I think it’s possible to replace the model_validator with a field_validator on S3Storage ‘s s3_config field. Let me try if that works and update here

components/clp-py-utils/clp_py_utils/clp_config.py
davemarco: this probably will break the original desired behaviour. read the comment above

junhaoliao: i tried for hours to switch branches and come up with different inputs, but i found no behavioural difference with or without `pre=True`. i can't be sure if i understand the `We set pre=True ... presto config itself` comment correctly. (it seems the comment was proposed by @kirkrodrigues ?) 

could you share some config example to help understand the difference?

---

i did dig up the conversations in #1198 and found this from Haiqi:

> https://github.com/y-scope/clp/pull/1198#pullrequestreview-3161740271
> One note to add:
The reason we put pre=true in root_validator is to prevent validating Presto config when it is not allowed.
> 
> For example, if user specifies a Presto config with invalid port when query_engine is NOT presto. Without pre=true, the script will first complain about invalid port, and only if user fixes the port, it will then let user know that presto is not allowed when query_engine is not presto.

so i tested with this `clp-config.yaml`
```yaml
package:
  storage_engine: "clp"
  query_engine: "presto"

presto:
  host: 1
  port: -1
```
1. with the existing code on `main@a040dfb4f4a14ef22f135dac13d90ff2b6984e07`:
    ```
    junhao@ASUS-X870E:~/workspace/2-clp/build/clp-package$ ./sbin/start-clp.sh 
    2025-09-28T21:48:13.377 ERROR [start_clp] Failed to load config.
      File "/home/junhao/workspace/2-clp/build/clp-package/lib/python3/site-packages/clp_package_utils/scripts/start_clp.py", line 1185, in main
        clp_config = load_config_file(config_file_path, default_config_file_path, clp_home)
      File "/home/junhao/workspace/2-clp/build/clp-package/lib/python3/site-packages/clp_package_utils/general.py", line 422, in load_config_file
        clp_config = CLPConfig.parse_obj(raw_clp_config)
      File "pydantic/main.py", line 532, in pydantic.main.BaseModel.parse_obj
      File "pydantic/main.py", line 347, in pydantic.main.BaseModel.__init__
    pydantic.error_wrappers.ValidationError: 2 validation errors for CLPConfig
    package -> __root__
      query_engine 'presto' is only compatible with storage_engine 'clp-s'. (type=value_error)
    presto -> port
      Presto.port is not within valid range 0-65535. (type=value_error)
    ```
2. I modified the code to use `@root_validator` without `pre=True`:
    ```patch
    diff --git a/components/clp-py-utils/clp_py_utils/clp_config.py b/components/clp-py-utils/clp_py_utils/clp_config.py
    --- a/components/clp-py-utils/clp_py_utils/clp_config.py	(revision a040dfb4f4a14ef22f135dac13d90ff2b6984e07)
    +++ b/components/clp-py-utils/clp_py_utils/clp_config.py	(date 1759110179774)
    @@ -989,7 +989,7 @@
     
         # We set `pre=True` so that we print errors about a mismatch between the query engine and presto
         # config before errors about the presto config itself.
    -    @root_validator(pre=True)
    +    @root_validator
         def validate_presto_config(cls, values):
             package = values.get("package")
             if not isinstance(package, Package):
    @@ -997,7 +997,7 @@
                 # later and throw an error).
                 return values
     
    -        query_engine = package.get("query_engine")
    +        query_engine = package.query_engine
             presto = values.get("presto")
             if query_engine == QueryEngine.PRESTO and presto is None:
                 raise ValueError(
    ```
    ->
    ```
    junhao@ASUS-X870E:~/workspace/2-clp/build/clp-package$ ./sbin/start-clp.sh 
    2025-09-28T21:50:54.323 ERROR [start_clp] Failed to load config.
      File "/home/junhao/workspace/2-clp/build/clp-package/lib/python3/site-packages/clp_package_utils/scripts/start_clp.py", line 1185, in main
        clp_config = load_config_file(config_file_path, default_config_file_path, clp_home)
      File "/home/junhao/workspace/2-clp/build/clp-package/lib/python3/site-packages/clp_package_utils/general.py", line 422, in load_config_file
        clp_config = CLPConfig.parse_obj(raw_clp_config)
      File "pydantic/main.py", line 532, in pydantic.main.BaseModel.parse_obj
      File "pydantic/main.py", line 347, in pydantic.main.BaseModel.__init__
    pydantic.error_wrappers.ValidationError: 2 validation errors for CLPConfig
    package -> __root__
      query_engine 'presto' is only compatible with storage_engine 'clp-s'. (type=value_error)
    presto -> port
      Presto.port is not within valid range 0-65535. (type=value_error)
    ```
3. with the code on this PR branch where i changed it to `@model_validate(mode="after)` with Pydantic V2, which is similar to `@root_validate` in Pydantic V1: https://github.com/y-scope/clp/blob/2ca7cb95ecc25ca9ecd844177d267df54153cc74/components/clp-py-utils/clp_py_utils/clp_config.py
    ```
    junhao@ASUS-X870E:~/workspace/2-clp/build/clp-package$ ./sbin/start-clp.sh 
    2025-09-28T22:00:01.268 ERROR [start_clp] Failed to load config.
      File "/home/junhao/workspace/2-clp/build/clp-package/lib/python3/site-packages/clp_package_utils/scripts/start_clp.py", line 1183, in main
        clp_config = load_config_file(config_file_path, default_config_file_path, clp_home)
      File "/home/junhao/workspace/2-clp/build/clp-package/lib/python3/site-packages/clp_package_utils/general.py", line 422, in load_config_file
        clp_config = CLPConfig.model_validate(raw_clp_config)
      File "/home/junhao/workspace/2-clp/build/clp-package/lib/python3/site-packages/pydantic/main.py", line 705, in model_validate
        return cls.__pydantic_validator__.validate_python(
    pydantic_core._pydantic_core.ValidationError: 2 validation errors for CLPConfig
    package
      Value error, query_engine 'presto' is only compatible with storage_engine 'clp-s'. [type=value_error, input_value={'storage_engine': 'clp',...query_engine': 'presto'}, input_type=dict]
        For further information visit https://errors.pydantic.dev/2.11/v/value_error
    presto.port
      Value error, Presto.port is not within valid range 0-65535. [type=value_error, input_value=-1, input_type=int]
        For further information visit https://errors.pydantic.dev/2.11/v/value_error
    ```

The query_engine mismatch error was reported in all above cases.

junhaoliao: in case we no longer need to replicate the previous handling, i have removed the outdated comment in 167f2a65ce2e2fde88e81915b70633ff40e8b2ac. If `mode="before"` is indeed needed, i can revert and update the comment.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -504,11 +562,12 @@ class AwsAuthentication(BaseModel):
...
 
-    @root_validator(pre=True)
-    def validate_authentication(cls, values):
-        auth_type = values.get("type")
-        profile = values.get("profile")
-        credentials = values.get("credentials")
+    @model_validator(mode="before")
+    @classmethod
+    def validate_authentication(cls, data):
+        auth_type = data.get("type")
```
gibber9809, line 573:
```suggestion
        if not isinstance(data, dict):
            raise ValueError(f"authentication config expects to be initialized from a dict, but received {type(data).__name__}.")
        auth_type = data.get("type")
```
The [pydantic docs](https://docs.pydantic.dev/latest/concepts/validators/#model-validators) mention that "before" validators need to deal with data that can be an arbitrary python object. I think it makes sense to check that `data` is actually a dict before calling `get` to avoid weird errors here.

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -66,8 +66,8 @@ def get_credential_env_vars(auth: AwsAuthentication) -> Dict[str, str]:
     :raise: ValueError if `auth.type` is not a supported type or fails to authenticate with the
     given `auth`.
     """
-    env_vars: Optional[Dict[str, str]] = None
-    aws_credentials: Optional[S3Credentials] = None
+    env_vars: Optional[Dict[str, str]]
+    aws_credentials: Optional[S3Credentials]
```
gibber9809, line 70: Just for my own understanding, why do we remove the explicit `= None` for these optional arguments here, but add it at `clp_config.py:538`? Also seems like there are several other places with explicit `= None` that we've left untouched.

gibber9809 (changes requested):
Nice work! Just a few minor comments/clarifications, then this change should be good to go.

components/clp-py-utils/clp_py_utils/s3_utils.py
junhaoliao: > why do we remove the explicit = None for these optional arguments here

the `None` value is never used because new values will be assigned by one of the if-branches. the "unused variable error" is actually marked by PyCharm

>  add it at clp_config.py:538

Pydantic V2 requires that all fields with optional types be assigned `None` values at their declarations inside the models.

> Also seems like there are several other places with explicit `= None` that we've left untouched.

Likely in some later branches the variable is untouched, so the `None` value can be used.

components/clp-py-utils/clp_py_utils/clp_config.py
Eden-D-Zhang: FYI if applying the suggested diff, "authentication" has a little spelling mistake.

Eden-D-Zhang (commented):
No comments from me, just existing suggestion from Devin.

gibber9809 (approved):
LGTM! I'm fine with the PR title as well.

components/job-orchestration/job_orchestration/scheduler/job_config.py
```diff
@@ -84,7 +84,8 @@ class SearchJobConfig(QueryJobConfig):
     network_address: typing.Optional[typing.Tuple[str, int]] = None
     aggregation_config: typing.Optional[AggregationConfig] = None
 
-    @validator("network_address")
+    @field_validator("network_address")
+    @classmethod
     def validate_network_address(cls, field):
```
junhaoliao, line 89: this is purely stylistic - we should change the second argument's name (`field`) to `value` to match the convention in the latest Pydantic docs

similarly, we should do the same for `valid_status` in `components/job-orchestration/job_orchestration/scheduler/scheduler_data.py`

components/job-orchestration/job_orchestration/scheduler/scheduler_data.py
```diff
@@ -79,25 +80,25 @@ def get_config(self) -> QueryJobConfig:
...
     def get_config(self) -> QueryJobConfig:
         return self.search_config
 
-    class Config:  # To allow asyncio.Task and asyncio.Queue
-        arbitrary_types_allowed = True
-
 
 class QueryTaskResult(BaseModel):
     status: QueryTaskStatus
     task_id: str
```
junhaoliao, line 102: Pydantic V2 seems to enforce stricter type checks, and creating a `QueryTaskResult` with the `task_id` being an `int` actually causes failure in IR extraction jobs (i.e., the "Link to log-viewer" feature in the webui).

@coderabbitai file an issue to track the bug

PR #1340, opened 2025-09-26
feat(job-orchestration): Add Spider compression task; Move Celery compression task into a dedicated submodule.

sitaowang1998:
# Description

This PR
* Separates celery compression task into separate file.
* Adds spider compression task.

# Checklist

  breaking change.

# Validation performed
* [x] Celery compression works end-to-end.
* [x] GitHub workflows pass.

## Summary by CodeRabbit

- **New Features**
  - Compression exposed as a Celery worker task with graceful shutdown logging.
  - Spider-compatible compression endpoint and payload encoding/decoding helpers for interoperability.

- **Refactor**
  - Core compression logic decoupled from worker bindings; logger is now passed through execution paths.

- **Chores**
  - Task routing and scheduler wiring updated to use the new worker-oriented compression endpoints.


components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
LinZhihao-723, line 1: This file should be called `spider_compress.py`. Please double-check the diff before sending the review request.

components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
```diff
@@ -0,0 +1,37 @@
...
+
+# Setup logging
+logger = get_logger("spider_compression_scheduler")
+
+
+def compress(
+    _: TaskContext,
+    job_id: int,
+    task_id: int,
+    tag_ids: list[int] | None,
```
LinZhihao-723: Since the package is running on Python3.9, I think we should use `Union` or `Optional` instead.

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -523,20 +518,14 @@ def run_clp(
...
 
-@signals.worker_shutdown.connect
-def worker_shutdown_handler(signal=None, sender=None, **kwargs):
-    logger.info("Shutdown signal received.")
-
-
-@app.task(bind=True)
-def compress(
-    self: Task,
+def general_compress(
```
LinZhihao-723: How about: `compression_entry_point`

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -523,20 +518,14 @@ def run_clp(
...
     job_id: int,
     task_id: int,
     tag_ids,
     clp_io_config_json: str,
     paths_to_compress_json: str,
     clp_metadata_db_connection_config,
+    logger,
```
LinZhihao-723, line 528: Alright, maybe let's keep this logger passing around for now. But we should open an issue to manage this logger properly. I think the proper implementation could be: instead of entering the compressor from an entry point function, we create a class called `ClpCompressor`. This object takes the logger as a class member, which allows sharing the logger among all helper functions (and presumably we need to make the current helpers private member functions).

components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
```diff
@@ -0,0 +1,37 @@
+from clp_py_utils.clp_logging import get_logger
+from job_orchestration.executor.compress.compression_task import general_compress
+from spider_py import TaskContext
+
+# Setup logging
+logger = get_logger("spider_compression_scheduler")
+
+
+def compress(
+    _: TaskContext,
```
LinZhihao-723, line 13: Are we planning to use this variable in the future PRs?

components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
```diff
@@ -0,0 +1,37 @@
...
+    job_id: int,
+    task_id: int,
+    tag_ids: list[int] | None,
+    clp_io_config_json: str,
+    paths_to_compress_json: str,
+    clp_metadata_db_connection_config: dict[str, bool | int | str | None],
+) -> dict[str, int | float | str | None]:
```
LinZhihao-723:
```suggestion
) -> dict[str, Any]:
```
I don't think we need to go into the granularity of what values can be serialized by `CompressionTaskResult.dict()` method. Using `Any` should be fine.
Similar for `clp_metadata_db_connection_config`

components/job-orchestration/job_orchestration/executor/compress/spider_comrpess.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    Compress files using the general compression function.
+    :param _:
+    :param job_id:
+    :param task_id:
+    :param tag_ids:
+    :param clp_io_config_json: A Json string representing `ClpIoConfig`.
+    :param paths_to_compress_json: A Json string representing `PathsToCompress`.
+    :param clp_metadata_db_connection_config: A dict representing `Database`.
+    :return: A dict representing `CompressionTaskResult`.
```
LinZhihao-723:
```suggestion
    Compresses files using the general compression entry point.
    :param _: Spider's task context. Not used in the function.
    :param job_id:
    :param task_id:
    :param tag_ids:
    :param clp_io_config_json: A JSON string representation of
        `job_orchestration.scheduler.constants.ClpIoConfig`.
    :param paths_to_compress_json: A JSON string representation of
        `job_orchestration.scheduler.constants.PathToCompress`.
    :param clp_metadata_db_connection_config: A dict representation of
        `clp_py_utils.clp_config.Database`.
    :return: A dict representation of `job_orchestration.scheduler.constants.CompressionTaskResult`.
```
* `Compresses` not `Compress`.
* `JSON` not `Json`.
* Without including the namespace, these symbols are super confusing, especially for `Database` (as it's not a database but a database config).
* `A xxx representation of xxx` is clearer in this context, as they're the serialized form of the original object.

components/job-orchestration/job_orchestration/executor/compress/celery_compress.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@signals.worker_shutdown.connect
+def worker_shutdown_handler(signal=None, sender=None, **kwargs):
+    logger.info("Shutdown signal received.")
+
+
+@app.task(bind=True)
+def compress(
+    _: Task,
```
LinZhihao-723: What's the issue of using `self` as what we originally did?

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
LinZhihao-723, line 1: Can u post a draft PR for the actual integration? Without seeing how things get integrated, it's a little hard to determine whether this PR is the best practice.

LinZhihao-723 (changes requested):

components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
sitaowang1998: How about `from __future__ import annotations`?

components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
sitaowang1998: Maybe. Depend on how we plan the failure handling.

components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
sitaowang1998: Unfortunately `Spider`'s task requires explicit return type. Use of `Any` does not work.

components/job-orchestration/job_orchestration/executor/compress/celery_compress.py
sitaowang1998: Just that it is not used.

components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
LinZhihao-723: Wait, does Spider task even allow union types?

components/job-orchestration/job_orchestration/executor/compress/celery_compress.py
LinZhihao-723: Yeah I understand, but I suggest to explicitly use `self` as the official doc do: https://docs.celeryq.dev/en/stable/userguide/tasks.html#bound-tasks.
This method is bound as a member method of `Task` in the runtime, I think.

components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
LinZhihao-723: Sure. Just make sure it's compatible with Python 3.9.

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
LinZhihao-723: @coderabbitai Can u create a feature request?

components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
sitaowang1998: No. Now I remember why I was using json before.

components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    return json.load(compression_entry_point(
+        job_id,
+        task_id,
+        tag_ids,
+        clp_io_config_json,
+        paths_to_compress_json,
+        json.dumps(clp_metadata_db_connection_config_json),
+        logger,
+    ))
```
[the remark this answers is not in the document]
LinZhihao-723: I think Rabitt is right: `load` is misused here: https://docs.python.org/3/library/json.html#json.load
We should use `dumps` instead.

components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
```diff
@@ -0,0 +1,57 @@
...
+logger = get_logger("spider_compression_scheduler")
+
+
+def convert_to_str(byte_list: list[Int8]) -> str:
+    """
+    Converts a list of Int8 to a UTF-8 string.
+    :param byte_list: A list of Int8 representing bytes.
+    :return: A UTF-8 string.
+    """
+    return bytes(int(byte) for byte in byte_list).decode("utf-8")
```
LinZhihao-723: - Is this function properly named and located? It is a private function, right?
- Shouldn't we use \`Int8\`?
- "A list of Int8 representing bytes" is not descriptive. I think if the param name + type annotation are already self-explanatory, we can omit the long-text description.

However, please pay particular attention to the first two bullet points, as these types of mistakes are repeated too often. Maybe use a checklist to ensure you go through all these yourself before requesting reviews.

components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
LinZhihao-723, line 1: @coderabbitai review

components/job-orchestration/job_orchestration/utils/spider_utils.py
LinZhihao-723, line 1: @coderabittai review

LinZhihao-723, line 1: @coderabbitai review

LinZhihao-723:
@coderabbitai review

LinZhihao-723:
@coderabbitai review

components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
```diff
@@ -42,9 +41,7 @@ def compress(
             [int(tag_id) for tag_id in tag_ids],
             int8_list_to_utf8_str(clp_io_config_json),
             int8_list_to_utf8_str(paths_to_compress_json),
-            Database.model_validate(
-                json.loads(int8_list_to_utf8_str(clp_metadata_db_connection_config_json))
-            ),
+            json.loads(int8_list_to_utf8_str(clp_metadata_db_connection_config_json)),
```
LinZhihao-723, line 44: @sitaowang1998 `compression_entry_point` requires the database config to be a dictionary. I believe this was a mistake.

LinZhihao-723 (approved):
- Tested that the celeray compression worked.
- Can't test Spider compression since the scheduler side is not ready.

PR #1344, opened 2025-09-26
feat: Set up the Cargo workspace for Rust components:

LinZhihao-723:
* Add a task to install and upgrade the Rust toolchain.
* Add tasks to lint and test the Cargo workspace.
* Add a new GitHub workflow to run linters and testers for the Cargo workspace.

# Description

This PR sets up the Cargo workspace for adding Rust components, which includes:

* Add a skeleton workspace with a single project `log-ingestor`.
* Add tasks to:
  * Install Rust toolchains
  * Lint the workspace using rustfmt and clippy
  * Test the workspace
* Add workflow to lint and test the workspace.

# Checklist

  breaking change.

# Validation performed

* [ ] Ensure all workflows passed.
* [x] Ensure the toolchain can be properly set up in a local dev environment.

## Summary by CodeRabbit

- New Features
  - Introduced an initial Rust-based log ingestor component (early preview).

- Documentation
  - Added README outlining the log ingestor at a high level.

- Chores
  - Expanded automated linting to include JavaScript, Python, YAML, and Rust.
  - Set up CI for Rust linting and tests across multiple OS environments.
  - Added automated Rust toolchain setup and workspace configuration.
  - Adjusted Rust lint configuration and updated ignore rules for local formatting files.

### DIFF
--- a/components/clp-py-utils/clp_py_utils/clp_config.py
+++ b/components/clp-py-utils/clp_py_utils/clp_config.py
@@ -835,6 +835,12 @@ class CLPConfig(BaseModel):
 
     _os_release_file_path: pathlib.Path = PrivateAttr(default=OS_RELEASE_FILE_PATH)
 
+    @validator("aws_config_directory")
+    def expand_profile_user_home(cls, value: Optional[pathlib.Path]):
+        if value is not None:
+            value = value.expanduser()
+        return value
+
     def make_config_paths_absolute(self, clp_home: pathlib.Path):
         if StorageType.FS == self.logs_input.type:
             self.logs_input.make_config_paths_absolute(clp_home)
@@ -923,7 +929,9 @@ class CLPConfig(BaseModel):
                     "aws_config_directory must be set when using profile authentication"
                 )
             if not self.aws_config_directory.exists():
-                raise ValueError("aws_config_directory does not exist")
+                raise ValueError(
+                    f"aws_config_directory does not exist: '{self.aws_config_directory}'"
+                )
         if not profile_auth_used and self.aws_config_directory is not None:
             raise ValueError(
                 "aws_config_directory should not be set when profile authentication is not used"


.github/workflows/clp-rust-checks.yaml
LinZhihao-723, line 1: Another option is to call this `clp-cargo-workspace-ci`.

taskfiles/lint.yaml
```diff
@@ -25,17 +26,12 @@ tasks:
       - task: "check-cpp-format"
       - task: "check-cpp-static-full"
 
-  check-no-cpp:
```
LinZhihao-723, line 28: As discussed offline, I removed this and made it verbose in the lint workflow.

taskfiles/toolchains.yaml
LinZhihao-723, line 1: Need to verify:
* We're ok with this `toolchains` namespace
* We're ok with this `toolchains:rust`'s task name.
* The installation steps (which should be ok from what I've tried).

LinZhihao-723 (commented):
@kirkrodrigues let a few comments for u to confirm.

.github/workflows/clp-rust-checks.yaml
kirkrodrigues: How about `clp-rust-checks.yaml`? Matches up with the task names.

taskfiles/toolchains.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+tasks:
+  rust:
+    run: "once"
+    cmds:
+      # Installs Rust toolchains using rustup. If the toolchains are already installed, this task
+      # will ensure they are updated to the latest versions.
+      - |-
+        curl --proto "=https" --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+        . "$HOME/.cargo/env"
```
kirkrodrigues, line 12: On a fresh system without rust installed, the rust linting task fails since this line doesn't update the path of the current shell (until the user logs in again). I think the reason this works in the workflow is because rust is already installed on the GitHub runner.

taskfiles/toolchains.yaml
```diff
@@ -0,0 +1,12 @@
+version: "3"
+
+tasks:
+  rust:
+    run: "once"
+    cmds:
+      # Installs Rust toolchains using rustup. If the toolchains are already installed, this task
+      # will ensure they are updated to the latest versions.
+      - |-
+        curl --proto "=https" --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
kirkrodrigues: * Why not add `--default-toolchain none` like the docs suggest?
* We usually prefer using long-form options rather than short-forms (unless that affects compatibility on macOS).

.github/workflows/clp-lint.yaml
```diff
@@ -43,6 +43,14 @@ jobs:
         name: "Install coreutils (for md5sum)"
         run: "brew install coreutils"
 
-      - name: "Run lint task"
+      - name: "Run lint task: js"
```
kirkrodrigues:
```suggestion
      - name: "Lint .js files"
```

Ditto below if you agree.

kirkrodrigues (commented):
Didn't do a complete review, just posting some comments based on skimming.

taskfiles/toolchains.yaml
LinZhihao-723: Sure. This command was basically copied from https://rust-lang.org/tools/install/, lol. I can expand it with full terms.

taskfiles/toolchains.yaml
LinZhihao-723: I think this might be that task execution starts a new shell, and the current shell just never gets updated.
It might be worth sourcing the environment before any `cargo` command executes.

LinZhihao-723: Successfully reproduced with a fresh Ubuntu container. The problem should be fixed now by adding `. "$HOME/.cargo/env"` before calling `cargo`.

taskfiles/toolchains.yaml
LinZhihao-723: I replaced everything except `-s and -y` in `sh` in the latest commit.

.github/workflows/clp-lint.yaml
LinZhihao-723: Shall I include `.ts`? Like `Lint .js/.ts files`?

Cargo.toml
```diff
@@ -0,0 +1,5 @@
+[workspace]
+members = [
+    "components/log-ingestor"
+]
+resolver = "3"
```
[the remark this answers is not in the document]
LinZhihao-723: U're outdated. It's been using 3 now, lol

taskfiles/toolchains.yaml
```diff
@@ -0,0 +1,13 @@
...
+  rust:
+    run: "once"
+    cmds:
+      # Installs Rust toolchains using rustup. If the toolchains are already installed, this task
+      # will ensure they are updated to the latest versions.
+      - |-
+        curl --proto "=https" --tlsv1.2 --silent --show-error --fail https://sh.rustup.rs \
+          | sh -s -- --default-toolchain none -y
+        . "$HOME/.cargo/env"
+        rustup toolchain install nightly --allow-downgrade --component clippy,rustfmt
```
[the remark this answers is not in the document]
LinZhihao-723: It worked. Using `,` to split the components is the right way to go.

taskfiles/lint.yaml
```diff
@@ -786,6 +783,68 @@ tasks:
...
+        vars:
+          RUST_CLIPPY_FLAGS: "--fix --allow-dirty --allow-staged"
+
+  rust-lint-configs:
+    internal: true
+    run: "once"
+    cmd: |
+      cp \
+        {{.ROOT_DIR}}/tools/yscope-dev-utils/exports/lint-configs/.rustfmt.toml \
+        {{.ROOT_DIR}}/.rustfmt.toml
```
sitaowang1998, line 826: Symbolic link would be better unless you are sure we will have different version of `.rustfmt.toml` soon.

sitaowang1998 (commented):
I think it is about time we setup lint for `toml` files like the one in [spider](https://github.com/y-scope/spider/pull/176), probably in another PR.

.github/workflows/clp-lint.yaml
junhaoliao: Both including ".ts" and excluding ".ts" make sense. At the end of the day, Typescript is just some type checking construct built on top of JavaScript. like TS code is transpiled into JS before it can be executed.

i think it's fine to mention just ".js"

taskfiles/lint.yaml
LinZhihao-723: David, Kirk, and I were having the discussion offline. We decided not to use symlinks anymore, as they can confuse some IDEs. We will directly copy the files instead.

LinZhihao-723:
> I think it is about time we setup lint for `toml` files like the one in [spider](https://github.com/y-scope/spider/pull/176), probably in another PR.

Sure. I can do it in another PR.

sitaowang1998 (approved):

PR #774, opened 2025-03-17
feat(clp-package): Add option for specifying archive compression level and update max compression level to 19.

Eden-D-Zhang:
# Description

This PR adds the `compression_level` option to `archive_output` in `clp-config.yml`. Compression level can be set to an integer value from 1 (lowest compression ratio, highest speed) to 19 (highest compression ratio, lowest speed).

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

- Compressed logs at different levels on both clp and clp-s.
- Verified compression ratio and speed change according to selected level.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

- **New Features**
  - A new compression-level setting has been introduced, allowing users to control archive compression with values ranging from 1 (fast/low compression) to 19 (slow/high compression), with a default of 3.
  - Job orchestration commands now support specifying the compression level, enabling greater flexibility in processing.
  - Configuration documentation has been enhanced to outline the new compression level option and its value range.

PR #1356, opened 2025-09-30
refactor(clp-package): Migrate config types and validation to Pydantic V2 (fixes #1342).

### LINKED ISSUE #1342: Refactor validators to use Field() instead of @field_validator for simple gt/lt constraints
## Description

As suggested in #1338, we should refactor validators that only check simple constraints (like ) to use Pydantic's  declaration instead of  decorators for consistency and cleaner code.

## Background

During the Pydantic v2 upgrade in #1338, the  class was updated to use  with constraints:
```python
archive: int = Field(default=60, gt=0)
search_result: int = Field(default=30, gt=0)
```

However, other classes like  and  still use  decorators for similar simple validations.

## Proposed Changes

Refactor the following fields to use  with appropriate constraints:

### ArchiveOutput class
- `target_archive_size: int = Field(default=256 * 1024 * 1024, gt=0)`
- `target_dictionaries_size: int = Field(default=32 * 1024 * 1024, gt=0)`
- `target_encoded_file_size: int = Field(default=256 * 1024 * 1024, gt=0)`
- `target_segment_size: int = Field(default=256 * 1024 * 1024, gt=0)`
- `compression_level: int = Field(default=3, ge=1, le=19)`
- `retention_period: Optional[int] = Field(default=None, gt=0)`

### StreamOutput class
- `target_uncompressed_size: int = Field(default=128 * 1024 * 1024, gt=0)`

This would make the code more consistent and reduce boilerplate validator methods.

## References

- Original discussion: https://github.com/y-scope/clp/pull/1338#discussion_r2382791246
- PR #1338: https://github.com/y-scope/clp/pull/1338

## Files to modify
- `components/clp-py-utils/clp_py_utils/clp_config.py`

junhaoliao:
# Description

This pull request migrates the type system and validation in `clp-py-utils/clp_config.py` and related modules to use Pydantic V2's new `Annotated` syntax and stronger type checking. Key changes include:
- Replacing legacy validator functions with Pydantic `Field` constraints and `Annotated` type aliases like `NonEmptyStr`, `PositiveInt`, `PositiveFloat`, `Host`, `Port`, and `ZstdCompressionLevel`.
- Introducing strict enum for `DatabaseEngine` using `KebabCaseStrEnum`.
- Adding `LoggingLevel` `Literal` which replaces the previous logging level validation method with a map.
- Refactoring constructor arguments and class attributes to these stricter types, improving type safety across all config objects.
- Removing legacy field_validator and model_validator code that can now be handled by Pydantic's built-in validation.
- Updating all classes (Database, Package, CompressionScheduler, QueryScheduler, Redis, ResultsCache, etc.) to use new type aliases and enums.
- Adjusting utility and dump methods to work with the new type structure.

# Checklist

  breaking change.

# Validation performed

1. To test the new logging level validation / setting utilities:
   1. Set logging_level to "DEBUG" in all applicable components in clp-config.yml. Started the package successfully and observed in the `clp-compression_scheduler` docker container logs that debug logs were printed. e.g.,
    ```
    2025-09-30 06:20:48,808 compression_scheduler [DEBUG] Search and schedule new tasks
    2025-09-30 06:20:48,808 compression_scheduler [DEBUG] Poll running jobs
    ```
    2. Set all `logging_level`s to "INVALID" in all applicable components and observed the package failed to start due to validation error.
2. Set out of bound / invalid values, ran `start-clp.sh` and observed the script fails with validation errors:
1. `NonEmptyStr`: `""`
2. `PositiveFloat`: `0`, `-1`, `-0.1`, false. (Also tried `1` which is an integer and the input was considered valid, which is expected)
3. `PositiveInt`: `0`, `-1`, `-0.1`, false
4. `Host`: `""`
5. `Port`: `0`, `-1`, `65536`. (also tried `1023` with non-root user - the script was able to start but port binding fails due to lack of permission)
6. `ZstdCompressionLevel`: `0`, `20`

## Summary by CodeRabbit

- New Features
  - Added PRESTO query engine support; new DatabaseEngine enum and typed aliases (Host, Port, NonEmptyStr, PositiveInt, ZstdCompressionLevel).
  - Configs now serialize enums and typed fields to primitive values; several config fields updated (database, S3, archive/stream outputs, web UI, redis, presto).

- Refactor
  - Simplified logging: validate level as a literal and set directly; invalid inputs default to INFO.
  - Removed legacy logging helpers and many in-model validators in favour of type-based constraints.


components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -99,12 +99,26 @@
 CLP_QUEUE_PASS_ENV_VAR_NAME = "CLP_QUEUE_PASS"
 CLP_REDIS_PASS_ENV_VAR_NAME = "CLP_REDIS_PASS"
 
+# Generic types
+NonEmptyStr = Annotated[str, Field(min_length=1)]
+PositiveFloat = Annotated[float, Field(gt=0)]
+PositiveInt = Annotated[int, Field(gt=0)]
+# Specific types
+Host = NonEmptyStr
```
junhaoliao: The alias is not strictly necessary in this PR, but they serve as markers for the host type. in a future PR we can replace it with `DomainStr`: https://github.com/pydantic/pydantic-extra-types/blob/main/pydantic_extra_types/domain.py

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -99,12 +99,26 @@
...
+Port = Annotated[int, Field(gt=0, lt=2**16)]
+ZstdCompressionLevel = Annotated[int, Field(ge=1, le=19)]
```
junhaoliao, line 109: cc @Eden-D-Zhang @kirkrodrigues 

technically, zstd [compression levels from 1 up to ZSTD_maxCLevel(),
  which is currently 22](https://facebook.github.io/zstd/zstd_manual.html?utm_source=chatgpt.com#:~:text=compression%20levels%20from%201%20up%20to%20ZSTD_maxCLevel()%2C%0A%20%20which%20is%20currently%2022) i believe we can also use `0` to specify a "no compression" level.

since this PR is strictly about pydantic validation refactoring. the range is directly inherited from #774 . if we want to update the range, we can file an issue and do it in a separate PR

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -173,49 +163,25 @@ def validate_query_engine_package_compatibility(self):
 
         return self
 
+    def dump_to_primitive_dict(self):
+        d = self.model_dump()
+        d["storage_engine"] = d["storage_engine"].value
+        d["query_engine"] = d["query_engine"].value
```
junhaoliao, line 169: pydantic by default doesn't know how to serialize these values, so we have to add this

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -99,12 +99,26 @@
...
+
 
 class StorageEngine(KebabCaseStrEnum):
     CLP = auto()
     CLP_S = auto()
 
 
+class DatabaseEngine(KebabCaseStrEnum):
+    MARIADB = auto()
+    MYSQL = auto()
```
junhaoliao, line 119: i can't be sure if it's a good idea to write `MARIA_DB` and `MY_SQL` so i didn't

junhaoliao: actually, we shouldn't. or the enum values would become `maria-db` and `my-sql`

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -265,7 +231,9 @@ def get_clp_connection_params_and_type(self, disable_localhost_socket_connection
         return connection_params_and_type
 
     def dump_to_primitive_dict(self):
-        return self.model_dump(exclude={"username", "password"})
+        d = self.model_dump(exclude={"username", "password"})
+        d["type"] = d["type"].value
```
junhaoliao, line 235: pydantic by default doesn't know how to serialize type `DatabaseEngine` so we have to add this

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -287,107 +255,35 @@ def load_credentials_from_env(self):
...
-
-    @field_validator("port")
-    @classmethod
-    def validate_port(cls, value):
-        _validate_port(cls, value)
-        return value
+    host: Host = "localhost"
+    port: Port = 7000
+    jobs_poll_delay: PositiveFloat = 0.1  # seconds
+    num_archives_to_search_per_sub_job: PositiveInt = 16
```
junhaoliao, line 267: should an upper bound be added (in a future PR)?

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -173,49 +163,25 @@ def validate_query_engine_package_compatibility(self):
...
-    ssl_cert: Optional[str] = None
+    type: DatabaseEngine = DatabaseEngine.MARIADB
+    host: Host = "localhost"
+    port: Port = 3306
+    name: NonEmptyStr = "clp-db"
+    ssl_cert: Optional[NonEmptyStr] = None
     auto_commit: bool = False
     compress: bool = True
 
     username: Optional[str] = None
```
junhaoliao, line 182: i didn't change this to `Optional[NonEmptyStr]` because it's possible for a mysql db username to be an empty string (though usually it's not a good idea)

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -799,73 +543,27 @@ def dump_to_primitive_dict(self):
...
+    rate_limit: PositiveInt = 1000
 
 
 class SweepInterval(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
-    archive: int = Field(default=60, gt=0)
-    search_result: int = Field(default=30, gt=0)
+    archive: PositiveInt = 60
+    search_result: PositiveInt = 30
```
junhaoliao, line 556: this should be a better syntax to specify shared types, and then default values can be specifically assigned to individual fields

components/clp-py-utils/clp_py_utils/clp_logging.py
```diff
@@ -25,17 +26,10 @@ def get_logger(name: str):
...
-
 def set_logging_level(logger: logging.Logger, level: str):
-    if not is_valid_logging_level(level):
+    if level not in get_args(LoggingLevel):
         logger.warning(f"Invalid logging level: {level}, using INFO as default")
         logger.setLevel(logging.INFO)
         return
-    logger.setLevel(LOGGING_LEVEL_MAPPING[level])
+
+    logger.setLevel(level)
```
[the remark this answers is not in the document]
junhaoliao: the `setLevel` method accepts both str and int logging levels. when the argument is a str, the method tries to map it to some numeric logging value before setting the number, so we should not need to do such mapping in our code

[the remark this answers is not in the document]
junhaoliao: https://docs.python.org/3/library/logging.html#logging.Logger.setLevel

> Changed in version 3.2: The level parameter now accepts a string representation of the level such as ‘INFO’ as an alternative to the integer constants such as [INFO](https://docs.python.org/3/library/logging.html#logging.INFO). Note, however, that levels are internally stored as integers, and methods such as e.g. [getEffectiveLevel()](https://docs.python.org/3/library/logging.html#logging.Logger.getEffectiveLevel) and [isEnabledFor()](https://docs.python.org/3/library/logging.html#logging.Logger.isEnabledFor) will return/expect to be passed integers.

components/clp-py-utils/clp_py_utils/clp_config.py
Eden-D-Zhang: I feel like making `Host` its own type just obscures the constraints since it gives less information in the name than `NonEmptyStr`, but the same requirements. If we want to use `DomainStr`, why not do `host: DomainStr = "localhost"`?

As it is, reading `host: Host = "localhost"` creates the need to look at the definition of `Host` to understand what values are legitimate.

Eden-D-Zhang (commented):
One question about the `Host` alias, otherwise looks good.

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: I renamed the `Host` type to `DomainStr` and added a TODO comment about plan for replacement. (I'm avoiding installing a new dependency in this PR which may bloat the changed lines.)

Let me know if i misunderstood your comment

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -98,12 +98,27 @@
...
 CLP_REDIS_PASS_ENV_VAR_NAME = "CLP_REDIS_PASS"
 
+# Generic types
+NonEmptyStr = Annotated[str, Field(min_length=1)]
+PositiveFloat = Annotated[float, Field(gt=0)]
+PositiveInt = Annotated[int, Field(gt=0)]
+# Specific types
+# TODO: Replace this with pydantic_extra_types.domain.DomainStr.
+DomainStr = NonEmptyStr
+Port = Annotated[int, Field(gt=0, lt=2**16)]
```
sitaowang1998, line 108: Should we avoid the lower range ports 0-1023 that are reserved for reserved for well-known applications?
```suggestion
Port = Annotated[int, Field(gt=1023, lt=2**16)]
```

sitaowang1998 (commented):
Should we include the annotated serialization in this PR or a later one?

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: If my understanding is correct, there's nothing functionally special about port 1-1023 in user applications, you can bind to them with root privileges. The restriction is a security convention to avoid malicious applications binding to the well-know ports, not a protocol-level limitation. That is, if users run docker with root permissions, they can still bind to those ports.

If users who don't have root permissions attempt binding to such ports, they will be stopped when `docker` is run. If we restrict the ports at config validation stage, users with root permissions would not be able to bind to those ports because we set the restrictions too tight. Therefore, I prefer keeping the definition `0-2**16` unchanged.

junhaoliao:
> Should we include the annotated serialization in this PR or a later one?

Do you mean the serialization improvements that you proposed last Friday? i think we can submit separately to avoid this PR getting too large and hard to review

sitaowang1998:
> > Should we include the annotated serialization in this PR or a later one?
> 
> Do you mean the serialization improvements that you proposed last Friday? i think we can submit separately to avoid this PR getting too large and hard to review

That's what I refer to. Separate PR is good.

sitaowang1998 (approved):

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: @coderabbitai file an issue for the type replacement with `pydantic-extra-types`'s `DomainStr`

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: @coderabbitai file an issue for the zstd compression level range update

PR #1357, opened 2025-09-30
feat(clp-package): Select execution container from image id or version file.

junhaoliao:
# Description

This PR adds the mechanism to the package init scripts to select the clp-package image built inside the project directory.

As a result, the ["clp-execution-base-ubuntu-jammy"](https://github.com/y-scope/clp/tree/main/tools/docker-images/clp-execution-base-ubuntu-jammy) image will no longer be in use, and its build script and related GH workflow can be removed.

# Checklist

  breaking change.

# Validation performed

### DIFF
--- a/components/clp-py-utils/clp_py_utils/clp_config.py
+++ b/components/clp-py-utils/clp_py_utils/clp_config.py
@@ -83,14 +83,13 @@ QUERY_TASKS_TABLE_NAME = "query_tasks"
 COMPRESSION_JOBS_TABLE_NAME = "compression_jobs"
 COMPRESSION_TASKS_TABLE_NAME = "compression_tasks"
 
-OS_RELEASE_FILE_PATH = pathlib.Path("etc") / "os-release"
-
 CLP_DEFAULT_CREDENTIALS_FILE_PATH = pathlib.Path("etc") / "credentials.yml"
 CLP_DEFAULT_DATA_DIRECTORY_PATH = pathlib.Path("var") / "data"
 CLP_DEFAULT_DATASET_NAME = "default"
 CLP_METADATA_TABLE_PREFIX = "clp_"
+CLP_PACKAGE_IMAGE_ID_PATH = pathlib.Path("clp-package-image.id")
 CLP_SHARED_CONFIG_FILENAME = ".clp-config.yml"
-
+CLP_VERSION_FILE_PATH = pathlib.Path("VERSION")
 
 # Environment variable names
 CLP_DB_USER_ENV_VAR_NAME = "CLP_DB_USER"
@@ -902,7 +901,8 @@ class CLPConfig(BaseModel):
     logs_directory: pathlib.Path = pathlib.Path("var") / "log"
     aws_config_directory: Optional[pathlib.Path] = None
 
-    _os_release_file_path: pathlib.Path = PrivateAttr(default=OS_RELEASE_FILE_PATH)
+    _image_id_path: pathlib.Path = PrivateAttr(default=CLP_PACKAGE_IMAGE_ID_PATH)
+    _version_file_path: pathlib.Path = PrivateAttr(default=CLP_VERSION_FILE_PATH)
 
     @field_validator("aws_config_directory")
     @classmethod
@@ -919,7 +919,8 @@ class CLPConfig(BaseModel):
         self.stream_output.storage.make_config_paths_absolute(clp_home)
         self.data_directory = make_config_path_absolute(clp_home, self.data_directory)
         self.logs_directory = make_config_path_absolute(clp_home, self.logs_directory)
-        self._os_release_file_path = make_config_path_absolute(clp_home, self._os_release_file_path)
+        self._image_id_path = make_config_path_absolute(clp_home, self._image_id_path)
+        self._version_file_path = make_config_path_absolute(clp_home, self._version_file_path)
 
     def validate_logs_input_config(self):
         logs_input_type = self.logs_input.type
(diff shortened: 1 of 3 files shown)


components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -1012,17 +1013,14 @@ def load_execution_container_name(self):
...
-            self.execution_container = (
-                f"clp-execution-x86-{os_release['ID']}-{os_release['VERSION_CODENAME']}:main"
-            )
-        else:
-            raise NotImplementedError(
-                f"Unsupported OS {os_release['ID']} in {OS_RELEASE_FILE_PATH}"
-            )
+        if self._image_id_path.exists():
+            with open(self._image_id_path) as image_id_file:
+                self.execution_container = image_id_file.read().strip()
```
junhaoliao: we may change this attribute name to `package_container` though i feel the name `execution_container` is more intuitive to those who are unfamiliar with the history of the execution container

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -83,14 +83,13 @@
 COMPRESSION_JOBS_TABLE_NAME = "compression_jobs"
 COMPRESSION_TASKS_TABLE_NAME = "compression_tasks"
 
-OS_RELEASE_FILE_PATH = pathlib.Path("etc") / "os-release"
-
 CLP_DEFAULT_CREDENTIALS_FILE_PATH = pathlib.Path("etc") / "credentials.yml"
 CLP_DEFAULT_DATA_DIRECTORY_PATH = pathlib.Path("var") / "data"
 CLP_DEFAULT_DATASET_NAME = "default"
 CLP_METADATA_TABLE_PREFIX = "clp_"
+CLP_PACKAGE_IMAGE_ID_PATH = pathlib.Path("clp-package-image.id")
```
kirkrodrigues: How about `CLP_PACKAGE_CONTAINER_IMAGE_ID_PATH`? Similarly, how about `_container_image_id_path`?

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -1012,17 +1013,14 @@ def load_execution_container_name(self):
...
-        else:
-            raise NotImplementedError(
-                f"Unsupported OS {os_release['ID']} in {OS_RELEASE_FILE_PATH}"
-            )
+        if self._image_id_path.exists():
+            with open(self._image_id_path) as image_id_file:
+                self.execution_container = image_id_file.read().strip()
 
-        self.execution_container = "ghcr.io/y-scope/clp/" + self.execution_container
+        if self.execution_container == "":
```
kirkrodrigues: `self.execution_container` is `None` by default which means this branch won't get triggered unless the image ID file is empty, right? It also feels a little inconsistent to validate that the image ID file isn't empty, but we don't do the same for the version file.

I think we should expect both of them to have content and let Python throw an exception if they don't. So I would write this as:

```python
        if self._image_id_path.exists():
            with open(self._image_id_path) as image_id_file:
                self.execution_container = image_id_file.read().strip()
        else:
            with open(self._version_file_path) as version_file:
                clp_package_version = version_file.read().strip()
            self.execution_container = f"ghcr.io/y-scope/clp/clp-package:{clp_package_version}"
```

components/clp-py-utils/clp_py_utils/clp_config.py
kirkrodrigues: How about `container_image_ref`?

taskfiles/docker-images.yaml
```diff
@@ -12,3 +12,4 @@ tasks:
       - ":package"
     cmds:
       - "./build.sh"
+      - "rsync -a '{{.G_BUILD_DIR}}/clp-package-image.id' '{{.G_PACKAGE_BUILD_DIR}}'"
```
kirkrodrigues:
```suggestion
      - "rsync --archive '{{.G_BUILD_DIR}}/clp-package-image.id' '{{.G_PACKAGE_BUILD_DIR}}'"
```

kirkrodrigues (commented):
Can we get rid of the logic to build execution images [here](https://github.com/y-scope/clp/blob/d25f16b1442d54db08165832cedb37b4f2f71f97/.github/actions/clp-build-runtime-image/action.yaml#L56-L59)?

.github/actions/clp-build-runtime-image/action.yaml
```diff
@@ -52,15 +52,8 @@ runs:
...
-        if [[ "${{inputs.image_type}}" == "execution" ]]; then
-          platform="${{inputs.platform_id}}-${{inputs.platform_version_id}}"
-          dockerfile_path="$base_path/clp-execution-base-$platform/Dockerfile"
-          image_name="clp-${{inputs.image_type}}-$platform"
-        else
-          dockerfile_path="$base_path/clp-package/Dockerfile"
-          image_name="clp-${{inputs.image_type}}"
-        fi
+        dockerfile_path="$base_path/clp-package/Dockerfile"
+        image_name="clp-${{inputs.image_type}}"
```
kirkrodrigues: We can now get rid of `inputs.image_type`, right?

junhaoliao: i hesitated, thinking we might still need `inputs.image_type` to build a different minimal worker image for multi-node deployments. since that isn't coming any time soon and `inputs.image_type` is unused for now, i agree we should remove

junhaoliao:
@kirkrodrigues shall we remove the `os-release` copy from the package, or do you think that's worth keeping?

kirkrodrigues:
> @kirkrodrigues shall we remove the `os-release` copy from the package, or do you think that's worth keeping?

I think we can remove it.

taskfile.yaml
```diff
@@ -113,7 +113,6 @@ tasks:
       - "{{.G_CORE_COMPONENT_BUILD_DIR}}/indexer"
       - "{{.G_CORE_COMPONENT_BUILD_DIR}}/reducer-server"
       - "{{.TASKFILE}}"
-      - "/etc/os-release"
```
junhaoliao, line 116: A different distro might cause the output of `npm ci` to be different though most of the time i don't think it would cause issues in the Node.js dependencies

kirkrodrigues (approved):
For the PR title, how about:

```
feat(clp-package): Replace package execution container with `clp-package`; Change default taskfile task to `docker-images:package`.
```

junhaoliao:
This PR changes the default task to be `docker-images:package`, which requires the `docker` binary (and also the builds plugin) to be available in the build environment. For those who build inside containers, we may need to create a new build container image or add Docker to the existing clp-core-dependencies-x86-ubuntu-jammy image. 

@coderabbitai create an issue to track the request

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -1007,22 +1012,18 @@ def validate_aws_config_dir(self):
...
             )
 
-    def load_execution_container_name(self):
-        if self.execution_container is not None:
+    def load_container_image_ref(self):
+        if self.container_image_ref is not None:
             # Accept configured value for debug purposes
             return
 
-        os_release = dotenv_values(self._os_release_file_path)
```
junhaoliao, line 1015: we should have removed the `dotenv_values` import and related dependency installation in this PR. @coderabbitai file an issue

PR #1361, opened 2025-10-01
feat(webui): Set up guided UI for query submission; Remove testing inputs.

davemarco:
# Description

Connects to new UI components to zustand states, and sets up the new components and run button to submit injected queries to the server. 

Does not include timeline support

# Checklist

  breaking change.

# Validation performed

Queries run using new guided ui

## Summary by CodeRabbit

* **New Features**
  * Added dedicated Guided and Freeform Run buttons with validation, disabled states and tooltips.
  * Guided controls (Select, From, Where, Order By, Limit, Timestamp Key) now sync with the Presto search state.

* **Style**
  * Reduced max width of the time range picker for a more compact layout.

* **Chores**
  * Removed the experimental SQL testing panel and its export to simplify the Search controls.


components/webui/server/settings.json
```diff
@@ -18,7 +18,7 @@
     "StreamFilesS3PathPrefix": null,
     "StreamFilesS3Profile": null,
 
-    "ClpQueryEngine": "clp",
+    "ClpQueryEngine": "clp-s",
```
hoophalab:
```suggestion
    "ClpQueryEngine": "clp",
```

components/webui/client/src/sql-parser/typings.ts
```diff
@@ -1,9 +1,9 @@
 interface BuildSearchQueryProps {
     selectItemList: string;
     databaseName: string;
-    booleanExpression?: string | undefined;
-    sortItemList?: string | undefined;
-    limitValue?: string | undefined;
+    booleanExpression: string;
+    sortItemList: string;
+    limitValue: number;
```
hoophalab: It feels a bit unnatural to force someone to pass an empty string for a clause they don't want. I'd prefer to keep the input props as they are and use this simple utility function in `GuidedRunButton` to handle optional values . junhao even wanted to implement a builder for this function https://github.com/y-scope/clp/pull/1304#issuecomment-3299977020

Also, since the limit value can be `ALL`, I think it's best to keep it as a string and just convert any number inputs to a string within `GuidedRunButton`.

```
            // eslint-disable-next-line no-undefined
            const getUndefinedIfEmpty = (str: string) => str === "" ? undefined : str
            const queryString = buildSearchQuery({
                booleanExpression: getUndefinedIfEmpty(where.trim()),
                databaseName: from,
                endTimestamp: endTimestamp.unix(),
                limitValue: limit.toString()
                selectItemList: select.trim(),
                sortItemList: getUndefinedIfEmpty(orderBy.trim()),
                startTimestamp: startTimestamp.unix(),
                timestampKey: timestampKey,
            });
```

(We could come up with a better name for `getUndefinedIfEmpty`)

davemarco: i see ur point but the inputs will naturally provide empty string and not undefined? Also user cant input 'ALL' for limit since we are restricting them to numbers. Like the max they can put is 1000, which makes sense since this is all we want them to see anyways

hoophalab: but the sql builder isn't limited to our inputs. It's a standalone component which might be used by the others

davemarco: okay i changed. linter would not accept ur exact approach, but did not something similiar with undefined

davemarco:
@hoophalab rereview

junhaoliao (approved):

PR #1369, opened 2025-10-02
fix(clp-s): Simplify timestamp range index evaluation code; Fix conversion utility used to compare float AST literals to integer values (fixes #1375).

### LINKED ISSUE #1375: clp-s: Incorrect rounding in `double_as_int` when converting floating-point numbers into integers for comparison against integer values.
### Bug

The `double_as_int` function is used in the AST code to convert floating-point literals into integers such that they can be compared against integer values.

Nominally this function is meant to do the following sorts of conversions:

`1 < 1.1` -> `1 < 2`
`1 > 0.9` -> `1 > 0`

Where depending on the kind of operations being performed we may need to compare the integer value against the floor or the ceiling of the integer literal.

Unfortunately, the `double_as_int` function has a bug where it will always return the _floor_ of the floating-point literal for all operations besides `==` and `!=`.

This results in in correct behaviour in the following situations:

1. When filtering against a field like `{"n": 1}` the query `n < 1.1` will erroneously _not_ match the record, and the query `n >= 1.1` erroneously _will_ match the record.

2. When filtering against the timestamp range index when it contains a range like `[1.1, 1.5]` a query like `timestamp < 1.4` will erroneously _not_ match the archive's time range. (This is because the timestamp range index evaluation code has some behaviour where it will convert float literals to integers before converting them back to floats to compare against a double-encoded time range, which is nominally correct, but overly complicated).

### CLP version

v0.5.1

### Environment

ubuntu focal

### Reproduction steps

Simplest reproduction is:

1. Ingest `{"n": 1}`
2. Observe that the query `n < 1.1` returns no results

gibber9809:
# Description

This PR fixes a bug in the `double_as_int` utility that the AST code uses to convert floating point literals into integers such that they can be compared against integer values.

Nominally this code is meant to do the following sorts of conversions:

`1 < 1.1` -> `1 < 2`
`1 > 0.9` -> `1 > 0`

Where depending on the kind of operation being performed, we may have to take the floor or ceiling of a given floating-point value in order to ensure correct comparison against an integer value.

Unfortunately, this conversion code had a bug causing the returned integer to always be the floor of the floating-point number for all operations besides `==` and `!=`.

We discovered this bug while investigating a case where the timestamp range index was incorrectly not being matched for certain queries -- the issue was that for archives with double encoded epoch range [a, b] we always tried to convert a literal `c` in a query like `timestamp < c` into an integer before evaluating the timestamp range index, so even if `a < c` was true the bug in the AST code would end up turning  this comparison into `a < floor(c)` which _may not_ be true.

Without the AST bug the timestamp range index evaluation code would technically be correct, but the way it was written (to allow float and integer literals to be compared with double-encoded and integer-encoded timestamp ranges interchangeably) was unnecessarily complex.

As a result, this PR:
* Fixes the AST float conversion bug
* Adds a search unit test covering the float conversion bug
* Simplifies the timestamp range index evaluation code by always comparing integer-encoded ranges with integer interpretations of literals and double-encoded ranges with double interpretations of literals
* Adds search unit tests dedicated to timestamp filtering

# Checklist

  breaking change.

# Validation performed

* Added dedicated unit tests for search against the timestamp column, some of which fail before this change
* Added test case exercising floating-point conversions that fail before this change

## Summary by CodeRabbit

- New Features
  - Improved timestamp search with support for both floating-point and epoch-based formats.
  - Stricter parsing aligned to data encoding for more accurate comparisons and range queries.

- Bug Fixes
  - Fixed potential mis-evaluation in timestamp comparisons caused by switch fall-through.
  - Standardized handling when timestamp literals don’t match the expected format, yielding consistent results.

- Tests
  - Added comprehensive tests for float and integer timestamp searches with new datasets.
  - Expanded general search dataset to cover additional scenarios.


components/core/tests/test-clp_s-search.cpp
```diff
@@ -289,3 +293,60 @@ TEST_CASE("clp-s-search-formatted-float", "[clp-s][search]") {
...
+        CAPTURE(query);
+        REQUIRE_NOTHROW(search(query, false, expected_results));
+    }
+}
+
+TEST_CASE("clp-s-search-epoch-timestamp", "[clp-s][search]") {
+    std::vector<std::pair<std::string, std::vector<int64_t>>> queries_and_results{
+            {R"aa(timestamp < 1759417024400)aa", {0, 1, 2}},
+            {R"aa(timestamp > 1759417023100)aa", {0, 1, 2}},
+            {R"aa(timestamp > 1759417024000)aa", {0, 1, 2}},
```
wraymo, line 330: Should we add a test for comparing timestamp with a floating point value?

components/core/tests/test_log_files/test_search.jsonl
```diff
@@ -11,3 +11,4 @@
 {"idx": 10, "ambiguous_varstring": "abcde"}
 {"idx": 11, "ambiguous_varstring": "ae"}
 {"idx": 12, "ambiguous_varstring": "a*e"}
+{"idx": 13, "one": 1}
```
wraymo, line 14: What's this case for?

wraymo (commented):
Thanks for the PR! Just two questions

components/core/tests/test_log_files/test_search.jsonl
gibber9809: The AST bug I mentioned -- it is exercised with the `one < 1.1 AND one > 0.9 AND one: 1.0` test case.

components/core/tests/test-clp_s-search.cpp
gibber9809: Sure, I can add one.

wraymo (approved):

kirkrodrigues:
Nice PR description. Can we file an issue for the bug (it's user-facing, right?) and refer to it in the title? Otherwise, it's a bit awkward to refer users to the PR, and the title itself doesn't (or can't) indicate what the actual bug was from a user perspective (which, in turn, makes writing release notes a bit harder).

PR #1377, opened 2025-10-03
refactor(taskfile): Split main dependency Taskfile into category-specific modules; Migrate `antlr-runtime` and `utfcpp` to use `utils:install-remote-cmake-lib`.

Bill-hbrhbr:
# Description

To address #1477, we split each library install task into its own file inside `taskfiles/deps/installs`. In addition, we also group tasks for each major target into respective files. So far the major targets are `core`, `log-viewer` and `spider`.

Other than the structural copy-paste refactoring, this PR also includes several one big and several small adjustments.

Big adjustment:
- **`antlr-runtime`** has been modified to be installable via `utils:install-remote-cmake-lib`. Instead of extracting the downloaded tar and removing the external layers outside the Cpp directory, we directly specify the directory where `CMakelists.txt` lives.

Small adjustments:
- `deps:core` and `deps:spider-dep` no longer resets the cmake settings directory, which is already done by `yscope-dev-utils:cmake:install-deps-and-generate-settings`
- The **`deps:`** sections of `deps:<target>-all-parallel` now iterate over list variables.
  - Easier to add common arguments to deps tasks in the future.
  - Easier to perform additional actions on libraries before/after the actual lib install tasks.
  - Similar adjustments have been made to **`deps:`** sections of lib install tasks that depend on other libraries.
- The **`utils:init`** task has been moved into `main.yaml`.
- Tasks prefixed with **`utils:`** and **`yscope-dev-utils:`** inside some taskfiles are now prefixed with a leading colon (`:`) to align with updated namespaces.

# Checklist

  breaking change.

# Validation performed

* [X] Existing workflows pass after refactorizations.

## Summary by CodeRabbit

* **Chores**
  * Reorganized and centralized dependency orchestration for more reliable initialisation and checksum handling.
  * Modularised workflows to enable parallel per-dependency generation/install flows and clearer platform checks.
  * Added automated install workflows for numerous third‑party libraries and tools, with integrity verification and reusable install steps.

---
  - https://app.asana.com/0/0/1212888467155538


taskfiles/deps/log-viewer.yaml
```diff
@@ -0,0 +1,17 @@
+version: "3"
+
+vars:
+  G_DEPS_LOG_VIEWER_CHECKSUM_FILE: "{{.G_DEPS_DIR}}/log-viewer.md5"
+
+tasks:
+  default:
+    deps:
+      - task: ":init"
```
Bill-hbrhbr:
```suggestion
```

Bill-hbrhbr:
```suggestion
```

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,365 @@
...
+    cmds:
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-DBUILD_SHARED_LIBS={{ if .BUILD_SHARED_LIBS }}ON{{ else }}OFF{{ end }}"
+            - >-
+              {{ join " " .COMMON_CMAKE_GEN_ARGS }}
+          LIB_NAME: "LibLZMA-{{ if .BUILD_SHARED_LIBS }}shared{{ else }}static{{ end }}"
+          TARBALL_SHA256: "{{.TARBALL_SHA256}}"
+          TARBALL_URL: "{{.TARBALL_URL}}"
```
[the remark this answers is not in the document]
Bill-hbrhbr: In the end we use
```
    cmds:
      - >-
        cmake
        -S "{{.SOURCE_DIR}}"
        -B "{{.BUILD_DIR}}"
        {{- range .EXTRA_ARGS}}
        {{.}}
        {{- end}}
```
so it ends up being one giant string anyway.
Our method works. In fact, list expansion that you provide doesn't work with go task template and yaml grammar

taskfile.yaml
```diff
@@ -127,7 +127,7 @@ tasks:
       - "clp-mcp-server"
       - "clp-package-utils"
       - "clp-py-utils"
-      - "deps:spider"
+      - "deps:spider:spider-dep"
```
Bill-hbrhbr:
```suggestion
      - "deps:spider"
```

davidlion:
If I understand correctly the scenario is that:

* In the package spider and clp are two separate projects/entities.
* The tasks changes from spider and clp in #1318 called their dependencies tasks as if they were separate projects.
* The `install-deps-and-generate-settings` task is made to be called once `CMAKE_SETTINGS_FILE`.
    * If `install-deps-and-generate-settings` is called from multiple projects with the same `CMAKE_SETTINGS_DIR` and different `CMAKE_SETTINGS_FILE`s any shared deps will be handled correctly (but the files may include unnecessary deps from other projects).
* The dependency tasks for spider were added as if spider was a shared dependency of clp (in order to share dependencies between the projects).
    * This is the root of the problem.

Is it correct to say the fix in this PR is splitting `all.cmake` into two files (basically changing 2 lines) and all other changes are organizational/visual improvements? If my understanding is correct, we should separate the fix into its own PR as this PR is currently both a `fix` and a `refactor`.

====

(Assuming my understanding above is correct.)
We should open an issue in dev-utils to track/remember this problem (add better support for dependencies shared between multiple projects aka multiple `install-deps-and-generate-settings` calls). However, I think a proper solution for shared-deps in non-trivial and the solution in this PR is probably fine in the few cases we have multiple projects inside one repo.

====

On a side note, in the future only request a single reviewer at a time to reduce confusion and prevent duplicate comments. The yscope-maintainers group automatically notifies a bunch of people, but you probably want to talk to people to figure out the order of reviews (if you want multiple) or who has time.

sitaowang1998:
@coderabbit Open a PR in `yscope-dev-utils` repo to better support dependency sharing between multiple projects.

sitaowang1998:
@coderabbit. Sorry I meant opening an issue

sitaowang1998:
> Is it correct to say the fix in this PR is splitting all.cmake into two files (basically changing 2 lines) and all other changes are organizational/visual improvements? If my understanding is correct, we should separate the fix into its own PR as this PR is currently both a fix and a refactor.

This is already done in #1331. This PR is just a refactor of the taskfile structure with some minor fixes.

The main problem is that `install-deps-and-generate-settings` removes and recreates the `CMAKE_SETTING_DIR`. This leads to concurrency issue when both `core` and `spider-deps` executes this task, which might crash the workflow.

The problem of including unnecessary settings in combined cmake setting files is not a big issue, as it does not affect the correctness of the build system.

taskfiles/deps/main.yaml
```diff
@@ -1,639 +1,55 @@
 version: "3"
 
 includes:
-  utils: "./utils.yaml"
+  # Main targets
```
Bill-hbrhbr:
```suggestion
  # Main project targets
```

taskfiles/deps/main.yaml
```diff
@@ -1,639 +1,55 @@
...
+  core: "core.yaml"
+  spider: "spider.yaml"
+  log-viewer: "log-viewer.yaml"
+
+  # Utils
+  utils: "utils.yaml"
   yscope-dev-utils: "../../tools/yscope-dev-utils/exports/taskfiles/utils/utils.yaml"
 
+  # Individual component tasks
```
Bill-hbrhbr:
```suggestion
  # Individual library component installation tasks
```

taskfiles/deps/misc-libs.yaml
```diff
@@ -0,0 +1,110 @@
...
+            - "iostreams"
+            - "process"
+            - "program_options"
+            - "regex"
+            - "system"
+            - "url"
+          URL: "https://github.com/boostorg/boost/releases/download/boost-{{.VERSION}}/\
+          boost-{{.VERSION}}-b2-nodocs.tar.gz"
+          WORK_DIR: "{{.G_DEPS_CPP_DIR}}"
+
```
Bill-hbrhbr:
```suggestion
```

taskfiles/deps/spider.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+  spider:
+    deps:
+      - task: "spider-dep"
+    platforms: ["linux"]
+    preconditions:
+      # NOTE: Spider is only supported on Ubuntu at present.
+      - >-
+        source /etc/os-release && [[ "$ID" == "ubuntu" ]]
+    cmds:
```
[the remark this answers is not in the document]
Bill-hbrhbr: @sitaowang1998 Does this comment make sense to you?

Bill-hbrhbr (commented):
several nit picks to be merged together with future review address commits

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+# - the component depends on another via `<lib_name>_ROOT`.
+tasks:
+  absl:
+    internal: true
+    run: "once"
+    cmds:
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-DCMAKE_BUILD_TYPE=Release"
```
sitaowang1998: Let's explicitly set all test building to off for cmake dependency libraries.
```suggestion
          CMAKE_GEN_ARGS:
            - "-DABSL_BUILD_TESTING=OFF"
            - "-DCMAKE_BUILD_TYPE=Release"
```

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+
+  antlr-runtime:
+    internal: true
+    run: "once"
+    cmds:
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-DANTLR4_INSTALL=ON"
+            - "-DCMAKE_BUILD_TYPE=Release"
```
sitaowang1998:
```suggestion
            - "-DANTLR4_INSTALL=ON"
            - "-DBUILD_TESTING=OFF"
            - "-DCMAKE_BUILD_TYPE=Release"
```

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+
+  date:
+    internal: true
+    run: "once"
+    cmds:
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-DCMAKE_BUILD_TYPE=Release"
+            - "-DCMAKE_INSTALL_MESSAGE=LAZY"
```
sitaowang1998:
```suggestion
            - "-DCMAKE_INSTALL_MESSAGE=LAZY"
            - "-DENABLE_DATE_TESTING=OFF"
```

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+            - "-DFMT_DOC=OFF"
+            - "-DFMT_TEST=OFF"
+          LIB_NAME: "{{.G_FMT_LIB_NAME}}"
+          TARBALL_SHA256: "bc23066d87ab3168f27cef3e97d545fa63314f5c79df5ea444d41d56f962c6af"
+          TARBALL_URL: "https://github.com/fmtlib/fmt/archive/refs/tags/11.2.0.tar.gz"
+
+  liblzma:
+    internal: true
+    vars:
+      COMMON_CMAKE_GEN_ARGS:
```
sitaowang1998: `COMMON_CMAKE_GEN_ARGS` sounds like general arguments for all cmake libraries.
```suggestion
      LIBLZMA_CMAKE_GEN_ARGS:
```

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+        - "-DXZ_TOOL_XZDEC=OFF"
+      TARBALL_SHA256: "507825b599356c10dca1cd720c9d0d0c9d5400b9de300af00e4d1ea150795543"
+      TARBALL_URL: "https://github.com/tukaani-project/xz/releases/download/v5.8.1/xz-5.8.1.tar.gz"
+    run: "once"
+    deps:
+      - task: "liblzma-install"
+        vars:
+          BUILD_SHARED_LIBS: true
+          COMMON_CMAKE_GEN_ARGS:
+            ref: ".COMMON_CMAKE_GEN_ARGS"
```
sitaowang1998:
```suggestion
            ref: ".LIBLZMA_CMAKE_GEN_ARGS"
```

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+          TARBALL_SHA256: "{{.TARBALL_SHA256}}"
+          TARBALL_URL: "{{.TARBALL_URL}}"
+      - task: "liblzma-install"
+        vars:
+          BUILD_SHARED_LIBS: false
+          COMMON_CMAKE_GEN_ARGS:
+            ref: ".COMMON_CMAKE_GEN_ARGS"
+          TARBALL_SHA256: "{{.TARBALL_SHA256}}"
+          TARBALL_URL: "{{.TARBALL_URL}}"
```
sitaowang1998: Why add same task twice?

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+
+  liblzma-install:
+    internal: true
+    requires:
+      vars: ["BUILD_SHARED_LIBS", "COMMON_CMAKE_GEN_ARGS", "TARBALL_SHA256", "TARBALL_URL"]
```
sitaowang1998:
```suggestion
      vars: ["BUILD_SHARED_LIBS", "LIBLZMA_CMAKE_GEN_ARGS", "TARBALL_SHA256", "TARBALL_URL"]
```

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+    cmds:
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-DBUILD_SHARED_LIBS={{ if .BUILD_SHARED_LIBS }}ON{{ else }}OFF{{ end }}"
+            - >-
+              {{ join " " .COMMON_CMAKE_GEN_ARGS }}
```
sitaowang1998:
```suggestion
              {{ join " " .LIBLZMA_CMAKE_GEN_ARGS }}
```

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+        source /etc/os-release && [[ "$ID" == "ubuntu" ]]
+    run: "once"
+    cmds:
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-DCMAKE_BUILD_TYPE=Release"
+            - "-DCMAKE_INSTALL_MESSAGE=LAZY"
+            - "-DINSTALL_LAYOUT=DEB"
+            - "-DUSE_SYSTEM_INSTALLED_LIB=ON"
```
sitaowang1998:
```suggestion
            - "-DUSE_SYSTEM_INSTALLED_LIB=ON"
            - "-DWITH_UNIT_TESTS=OFF"
```

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+    internal: true
+    vars:
+      VERSION: "r4.1.1"
+    run: "once"
+    cmds:
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-DBUILD_SHARED_AND_STATIC_LIBS=ON"
+            - "-DBUILD_SHARED_LIBS_WITH_STATIC_MONGOC=ON"
```
sitaowang1998:
```suggestion
            - "-DBUILD_SHARED_LIBS_WITH_STATIC_MONGOC=ON"
            - "-DENABLE_TESTS=OFF"
```

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+      - task: "boost"
+    cmds:
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-C {{.G_DEPS_CPP_CMAKE_SETTINGS_DIR}}/{{.G_BOOST_LIB_NAME}}.cmake"
+            - "-DCMAKE_BUILD_TYPE=Release"
+            - "-DCMAKE_INSTALL_MESSAGE=LAZY"
+            - "-DCMAKE_POLICY_DEFAULT_CMP0074=NEW"
+            - "-DMSGPACK_BUILD_DOCS=OFF"
```
sitaowang1998:
```suggestion
            - "-DMSGPACK_BUILD_DOCS=OFF"
            - "-DMSGPACK_BUILD_TESTS=OFF"
```

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+          TARBALL_URL: "https://github.com/nlohmann/json/archive/refs/tags/v3.11.3.tar.gz"
+
+  simdjson:
+    internal: true
+    run: "once"
+    cmds:
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-DCMAKE_BUILD_TYPE=Release"
```
sitaowang1998:
```suggestion
          CMAKE_GEN_ARGS:
            - "-DBUILD_TESTING=OFF"
            - "-DCMAKE_BUILD_TYPE=Release"
```

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-C {{.G_DEPS_CPP_CMAKE_SETTINGS_DIR}}/{{.G_FMT_LIB_NAME}}.cmake"
+            - "-DCMAKE_BUILD_TYPE=Release"
+            - "-DCMAKE_INSTALL_MESSAGE=LAZY"
+            - "-DCMAKE_POLICY_DEFAULT_CMP0074=NEW"
+            - "-DSPDLOG_BUILD_EXAMPLE=OFF"
+            - "-DSPDLOG_BUILD_EXAMPLE_HO=OFF"
+            - "-DSPDLOG_FMT_EXTERNAL=ON"
```
sitaowang1998:
```suggestion
            - "-DSPDLOG_BUILD_EXAMPLE=OFF"
            - "-DSPDLOG_BUILD_EXAMPLE_HO=OFF"
            - "-DSPDLOG_BUILD_TESTS=OFF"
            - "-DSPDLOG_BUILD_TESTS_HO=OFF"
            - "-DSPDLOG_FMT_EXTERNAL=ON"
```

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+
+  utfcpp:
+    internal: true
+    run: "once"
+    cmds:
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-DCMAKE_BUILD_TYPE=Release"
+            - "-DCMAKE_INSTALL_MESSAGE=LAZY"
```
sitaowang1998:
```suggestion
            - "-DCMAKE_INSTALL_MESSAGE=LAZY"
            - "-DUTF8CPP_ENABLE_TESTS=OFF"
```

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+    run: "once"
+    cmds:
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-DBUILD_SHARED_LIBS=ON"
+            - "-DCMAKE_BUILD_TYPE=Release"
+            - "-DCMAKE_INSTALL_MESSAGE=LAZY"
+            - "-DCMAKE_INSTALL_PREFIX={{.G_DEPS_CPP_DIR}}/{{.LIB_NAME}}-install"
+            - "-DZLIB_BUILD_EXAMPLES=OFF"
```
sitaowang1998:
```suggestion
            - "-DZLIB_BUILD_EXAMPLES=OFF"
            - "-DZLIB_BUILD_TESTING=OFF"
```

taskfiles/deps/main.yaml
```diff
@@ -1,639 +1,55 @@
...
-# NOTE: For dependencies built using CMake, we set CMP0074 to NEW whenever:
-# - a component's minimum required CMake version is less than 3.27 (where CMP0074 defaults to OLD);
-#   and
-# - the component depends on another via `<lib_name>_ROOT`.
 tasks:
   default:
     deps:
       - "core"
+      - "spider"
       - "log-viewer"
```
sitaowang1998: Sort tasks in lexical order. @LinZhihao-723 Could you confirm if we should keep `spider` as a default task?
```suggestion
      - "log-viewer"
      - "spider"
```

taskfiles/deps/spider.yaml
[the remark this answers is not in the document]
sitaowang1998: It is ok to use bash only commands. No one is really using the plain old sh, not even the GitHub workflows.

sitaowang1998 (changes requested):
The PR description is much border than what this PR is targeting. Maybe we could open an issue in clp to track the problem, which links the issue and PRs in `yscope-dev-utils` and this PR.

taskfiles/deps/main.yaml
LinZhihao-723: Yeah we should. If `log-viewer` is then Spider should also be.

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+          BUILD_SHARED_LIBS: true
+          COMMON_CMAKE_GEN_ARGS:
+            ref: ".COMMON_CMAKE_GEN_ARGS"
+          TARBALL_SHA256: "{{.TARBALL_SHA256}}"
+          TARBALL_URL: "{{.TARBALL_URL}}"
+      - task: "liblzma-install"
+        vars:
+          BUILD_SHARED_LIBS: false
+          COMMON_CMAKE_GEN_ARGS:
+            ref: ".COMMON_CMAKE_GEN_ARGS"
```
Bill-hbrhbr:
```suggestion
            ref: ".LIBLZMA_CMAKE_GEN_ARGS"
```

taskfiles/deps/cmake-libs.yaml
Bill-hbrhbr: LibLZMA doesn't support building both static and shared artifacts

taskfiles/deps/cmake-libs.yaml
Bill-hbrhbr: `BUILD_TESTING` does not exist in `simdjson`.

taskfiles/deps/cmake-libs.yaml
Bill-hbrhbr: `BUILD_TESTING` does not exist in `antlr4-runtime`.

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,388 @@
...
+    cmds:
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-C {{.G_DEPS_CPP_CMAKE_SETTINGS_DIR}}/{{.G_FMT_LIB_NAME}}.cmake"
+            - "-DCMAKE_BUILD_TYPE=Release"
+            - "-DCMAKE_INSTALL_MESSAGE=LAZY"
+            - "-DCMAKE_POLICY_DEFAULT_CMP0074=NEW"
+            - "-DSPDLOG_BUILD_EXAMPLE=OFF"
+            - "-DSPDLOG_BUILD_EXAMPLE_HO=OFF"
```
Bill-hbrhbr:
```suggestion
            - "-DSPDLOG_BUILD_EXAMPLE=OFF"
            - "-DSPDLOG_BUILD_TESTS=OFF"
            - "-DSPDLOG_BUILD_TESTS_HO=OFF"
```

taskfiles/deps/cmake-libs.yaml
Bill-hbrhbr: `SPDLOG_BUILD_EXAMPLE_HO` is already `OFF` so we don't need it anymore

taskfiles/deps/cmake-libs.yaml
sitaowang1998: `BUILD_TESTING` is a variable in `CTest`. If a project use `enable_testing`, we can set `BUILD_TESTING` to turn off the test build. See https://cmake.org/cmake/help/latest/variable/BUILD_TESTING.html.

taskfiles/deps/cmake-libs.yaml
Bill-hbrhbr: `UTF8CPP_ENABLE_TESTS` is not in `utfcpp` `v4.0.6` but in `v4.0.8`

taskfiles/deps/cmake-libs.yaml
Bill-hbrhbr: `ZLIB_BUILD_TESTING` is not present in `zlib-1.3.1`

taskfiles/deps/cmake-libs.yaml
sitaowang1998: Look like no way to stop test from building in `v4.0.6`. Maybe we can leave this here with a comment in case we bump the version later.

taskfiles/deps/cmake-libs.yaml
sitaowang1998: In `1.3.1`, `zlib` uses `enable_testing`. However, it does not build test if example build if off. Same with `utf8`, let's leave it here with a comment.

Bill-hbrhbr:
@sitaowang1998 @kirkrodrigues 

To provide some context: 
Kirk and I discussed when to add flags/options to the library CMake *generate* step.  
Our criteria for adding a flag are:
- The flag exists in `CMakeLists.txt` (or a file it includes).  
- We intend to change its default value (e.g., `ON → OFF` or `OFF → ON`).  
- **Exceptions (always include):**
  - `-DCMAKE_BUILD_TYPE=Release`
  - `-DCMAKE_INSTALL_MESSAGE=LAZY`

Many libraries expose test-related flags, but most are already `OFF`. To recognize Sitao’s effort
collecting them, I added a commit that includes all valid suggestions for our current lib versions:
- https://github.com/y-scope/clp/pull/1377/commits/5a2150cef9640ebc1cb15001d0fd9a5a96059356

**Policy options we can adopt going forward:**
1. **Make `-DBUILD_TESTING=OFF` universal**, and turn off all related testing flags.  
2. **Not universal, but always disable testing flags** when present, regardless of their defaults.  
   _This is what the code does now after the commit above._  
3. **Only disable testing flags that default to `ON`.**  
   _This is what `main` currently does._

Thoughts or preferences?

Bill-hbrhbr:
As a side note, I'm trying to avoid scope creep in this PR. We should do as much copy-paste and moving code around as possible. 

But I guess I did modify the install methods of two libraries (`utfcpp`, `antlr4-runtime`) to shorten the future PR pipeline length.

sitaowang1998:
> **Policy options we can adopt going forward:**
> 
>     1. **Make `-DBUILD_TESTING=OFF` universal**, and turn off all related testing flags.
> 
>     2. **Not universal, but always disable testing flags** when present, regardless of their defaults.
>        _This is what the code does now after the commit above._
> 
>     3. **Only disable testing flags that default to `ON`.**
>        _This is what `main` currently does._

For 1. I think we can only apply `-DBUILD_TESTING=OFF` to the libraries where this is the one and only way to turn off test building.
I am ok with either 2 or 3.

kirkrodrigues:
> For 1. I think we can only apply `-DBUILD_TESTING=OFF` to the libraries where this is the one and only way to turn off test building. I am ok with either 2 or 3.

I prefer to stick with 3 since it's the clearest to reason about.

taskfiles/deps/cmake-libs.yaml
Bill-hbrhbr: As discussed in all-hands we are switching back to the source version.

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,368 @@
...
+    internal: true
+    platforms: ["linux"]
+    preconditions:
+      # NOTE: The MariaDB connector is only required for building Spider, which is only supported on
+      # Ubuntu. `INSTALL_LAYOUT` is currently hardcoded to "DEB". To support other Linux distros:
+      # - Update the precondition accordingly.
+      # - Make `INSTALL_LAYOUT` configurable.
+      - >-
+        source /etc/os-release && [[ "$ID" == "ubuntu" ]]
+    run: "once"
```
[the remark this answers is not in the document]
Bill-hbrhbr: we don't care

taskfiles/deps/cmake-libs.yaml
Bill-hbrhbr: I've found a better flag: `-DANTLR_BUILD_CPP_TESTS=OFF`

taskfiles/deps/cmake-libs.yaml
```diff
@@ -0,0 +1,395 @@
...
+            - "-C {{.G_DEPS_CPP_CMAKE_SETTINGS_DIR}}/{{.G_BOOST_LIB_NAME}}.cmake"
+            - "-DCMAKE_BUILD_TYPE=Release"
+            - "-DCMAKE_INSTALL_MESSAGE=LAZY"
+            - "-DCMAKE_POLICY_DEFAULT_CMP0074=NEW"
+            - "-DMSGPACK_BUILD_DOCS=OFF"
+            - "-DMSGPACK_CXX20=ON"
+          LIB_NAME: "msgpack-cxx"
+          TARBALL_SHA256: "7504b7af7e7b9002ce529d4f941e1b7fb1fb435768780ce7da4abaac79bb156f"
+          TARBALL_URL: "https://github.com/msgpack/msgpack-c/releases/download/cpp-{{.VERSION}}\
+          /msgpack-cxx-{{.VERSION}}.tar.gz"
```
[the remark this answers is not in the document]
Bill-hbrhbr: This flag is off by default.

taskfiles/deps/installs/log-surgeon.yaml
```diff
@@ -0,0 +1,27 @@
...
+      - task: ":utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-C {{.G_DEPS_CPP_CMAKE_SETTINGS_DIR}}/{{.G_FMT_LIB_NAME}}.cmake"
+            - "-C {{.G_DEPS_CPP_CMAKE_SETTINGS_DIR}}/{{.G_GSL_LIB_NAME}}.cmake"
+            - "-DCMAKE_BUILD_TYPE=Release"
+            - "-DCMAKE_INSTALL_MESSAGE=LAZY"
+            - "-Dlog_surgeon_BUILD_TESTING=OFF"
+          DEPS:
+            ref: ".DEPS"
```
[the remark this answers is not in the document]
Bill-hbrhbr: this is how the project is written. Same for ystdlib

PR #1385, opened 2025-10-06
feat(clp-s): Implement part of the parsing side of the new `timestamp_parser` utility.

gibber9809:
# Description

This PR implements part of the parsing side of a new timestamp utility that will allow us to consistently parse timestamps into a known precision, support timezones, and ease maintenance of timestamp patterns.

This initial PR includes all of the necessary boilerplate for the parsing side (docstring for the parsing function, error codes + error category), implements about half of the planned format specifiers, and adds basic tests to ensure that all of the format specifiers that have been implemented so far can correctly parse the kind of content that they're supposed to accept.

Tests that ensure that whole timestamps are parsed correctly will be added in a future PR.

For more context on the project, the full design doc is available upon request.

# Checklist

  breaking change.

# Validation performed

* Added test case to validate that all of the implemented format specifiers correctly parse the content that they're supposed to.


components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
```diff
@@ -0,0 +1,84 @@
+#ifndef CLP_S_TIMESTAMPPARSER_HPP
+#define CLP_S_TIMESTAMPPARSER_HPP
```
```suggestion
#ifndef CLP_S_TIMESTAMP_PARSER_TIMESTAMPPARSER_HPP
#define CLP_S_TIMESTAMP_PARSER_TIMESTAMPPARSER_HPP
```
Also need to update the comment in `#endif`

components/core/src/clp_s/timestamp_parser/ErrorCode.hpp
```diff
@@ -0,0 +1,21 @@
+#ifndef CLP_S_TIMESTAMPPARSERERRORCODE_HPP
+#define CLP_S_TIMESTAMPPARSERERRORCODE_HPP
```
LinZhihao-723: Same, this seems out dated.

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
```diff
@@ -0,0 +1,84 @@
...
+ * @param pattern A timestamp pattern made up of literals, format specifiers, and potentially CAT
+ *                sequences.
+ * @param generated_pattern A buffer where a newly-generated timestamp pattern can be written, if
+ *                          necessary.
+ * @return On success:
+ *              - The timestamp in epoch nanoseconds.
+ *              - A string_view of the timestamp pattern that corresponds to the timestamp.
+ *                      - Lifetime is least of `pattern` and `generated_pattern`.
+ *         On error:
+ *              - A `clp_s::timestamp_parser::ErrorCode`.
```
LinZhihao-723: Haven't reviewed the content yet. But:
- We don't need to indent when wrapping to a new line (unless we recently udpated the guideline that I wasn't aware of)
- We have a guideline to document result return type. Here's an example that also returns a pair: https://github.com/y-scope/clp/blob/2e5ab12b00f7a2d8b413dbb860c7950e6da3234d/components/core/src/clp/ffi/ir_stream/ir_unit_deserialization_methods.hpp#L32
In general the rule for result return is:
`A result containing xxxx on success, or an error code indicating the failure:\n List all possible error codes`, or use `A void result containing xxx on success` if the return type is `void`.

Can u go through other docstrings to ensure these are all applied?

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+/**
+ * Converts a padded decimal integer string to an integer.
+ * @param str Substring containg the padded decimal integer string.
+ * @param padding_character Padding character which may prefix the integer.
+ * @param value The integer value, returned by reference.
+ * @return Whether conversion was successful.
+ */
+[[nodiscard]] auto
+convert_padded_string_to_number(std::string_view str, char padding_character, int& value) -> bool;
```
LinZhihao-723: We should modernize this by returning `std::optional<int>`.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+constexpr int cMaxParsedYear = 9999;
+constexpr int cMinTwoDigitYear = 0;
+constexpr int cTwoDigitYearOffsetBoundary = 69;
+constexpr int cMaxTwoDigitYear = 99;
+constexpr int cTwoDigitYearLowOffset = 1900;
+constexpr int cTwoDigitYearHighOffset = 2000;
+
+constexpr int cDefaultYear = 1970;
+constexpr int cDefaultMonth = 1;
+constexpr int cDefaultDay = 1;
```
```suggestion
constexpr int cMinParsedDay{1};
constexpr int cMaxParsedDay{31};
constexpr int cMinParsedMonth{1};
constexpr int cMaxParsedMonth{12};
constexpr int cMinParsedYear{0};
constexpr int cMaxParsedYear{9999};
constexpr int cMinTwoDigitYear{0};
constexpr int cTwoDigitYearOffsetBoundary{69};
constexpr int cMaxTwoDigitYear{99};
constexpr int cTwoDigitYearLowOffset{1900};
constexpr int cTwoDigitYearHighOffset{2000};

constexpr int cDefaultYear{1970};
constexpr int cDefaultMonth{1};
constexpr int cDefaultDay{1};
```

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+auto convert_padded_string_to_number(std::string_view str, char padding_character)
+        -> std::optional<int> {
+    if (str.empty()) {
+        return std::nullopt;
+    }
+
+    // Leave at least one character for parsing to ensure we actually parse number content.
+    size_t i{};
+    for (; i < (str.size() - 1) && padding_character == str[i]; ++i) {}
```
LinZhihao-723: Nit: I understand we've done in-bound check already, but shall we use `.at(i)` to access the char to resolve the clang-tidy warning (might not show up with the current clang-tidy version)?

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+/**
+ * Finds the first matching prefix from a list of candidates.
+ * @tparam CandidateArrayType
+ * @param str Substring with a prefix potentially matching one of the candidates.
+ * @param candidates Candidate prefixes that will be matched against `str`.
+ * @return A result containing the index of the matching prefix in the candidates array, or
+ * `clp_s::timestamp_parser::ErrorCodeEnum::IncompatibleTimestampPattern` on failure.
+ */
+template <typename CandidateArrayType>
```
LinZhihao-723: Do we need a template parameter?
- If to use a template, how about using range?
- If not, we can use a span.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+[[nodiscard]] auto convert_padded_string_to_number(std::string_view str, char padding_character)
+        -> std::optional<int>;
+
+/**
+ * Finds the first matching prefix from a list of candidates.
+ * @tparam CandidateArrayType
+ * @param str Substring with a prefix potentially matching one of the candidates.
+ * @param candidates Candidate prefixes that will be matched against `str`.
+ * @return A result containing the index of the matching prefix in the candidates array, or
+ * `clp_s::timestamp_parser::ErrorCodeEnum::IncompatibleTimestampPattern` on failure.
```
```suggestion
 * @return A result containing the index of the matching prefix in the candidates array, or an error
 * code indicating the failure:
 * - ErrorCodeEnum::IncompatibleTimestampPattern if no candidates match the prefix of `str`.
```
This matches how we document a result:
- You don't need to use \` to quote the error code enum.
- You also don't need to specify the full namespace. `ErrorCodeEnum::IncompatibleTimestampPattern` is already a valid symbol under the current namespace, and it can be resolved by the IDE (at least by CLion, lol).

But beyond this, does it make more sense to return an optional index instead? I understand the current implementation would make it easier to use in the parser function, but from this helper's perspective it might be more general to only return an optional and let the caller to decide what to do if the index isn't found. Up to u to decide.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+           std::string_view{"Sep"},
+           std::string_view{"Oct"},
+           std::string_view{"Nov"},
+           std::string_view{"Dec"}};
+
+/**
+ * Converts a padded decimal integer string to an integer.
+ * @param str Substring containg the padded decimal integer string.
+ * @param padding_character Padding character which may prefix the integer.
+ * @return The integer value on success, or `std::nullopt` on failure.
```
```suggestion
 * @return The integer value on success
 * @return std::nullopt otherwise.
```
This matches what we do for std::optional return. You don't need to quote `std::nullopt` since it's considered as a return value but not a symbol.

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
```diff
@@ -0,0 +1,92 @@
...
+ *     conflicting information about the actual date (e.g., if the parsed day of the week doesn't
+ *     match up with the rest of the timestamp information).
+ *   - `FormatSpecifierNotImplemented` if the pattern contains format specifiers that have not been
+ *     implemented yet.
+ */
+[[nodiscard]] auto parse_timestamp(
+        std::string_view timestamp,
+        std::string_view pattern,
+        std::string& generated_pattern
+) -> ystdlib::error_handling::Result<std::pair<epochtime_t, std::string_view>, ErrorCode>;
```
LinZhihao-723: I think the convention is to return a `std::error_code` so u don't need to specify the error type (the default is `std::error_code`). This makes it easier to propagate errors to the upper level callers, since it doesn't need to care which error code type it is.
If you see problems of can't get things compiled, try:
```c++
return ErrorCode{ErrorCodeEnum::XXX};
```

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+    size_t pattern_idx{};
+    size_t timestamp_idx{};
+
+    int parsed_year{cDefaultYear};
+    int parsed_month{cDefaultMonth};
+    int parsed_day{cDefaultDay};
+    std::optional<int> day_of_week_idx;
+
+    bool date_type_representation{false};
+    bool const number_type_representation{false};
```
LinZhihao-723, line 141: I guess in later PRs we will need this variable to be mutable?

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+            escaped = true;
+            continue;
+        }
+        if (false == escaped) {
+            if (pattern[pattern_idx] == timestamp[timestamp_idx]) {
+                ++timestamp_idx;
+                continue;
+            }
+            return ErrorCodeEnum::IncompatibleTimestampPattern;
+        }
```
```suggestion
        if (false == escaped) {
            if ('\\' == pattern[pattern_idx]) {
                escaped = true;
                continue;
            }
            if (pattern[pattern_idx] == timestamp[timestamp_idx]) {
                ++timestamp_idx;
                continue;
            }
            return ErrorCodeEnum::IncompatibleTimestampPattern;
        }
```
Should be equivalent?

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+        escaped = false;
+        switch (pattern[pattern_idx]) {
+            case 'y': {  // Zero-padded 2-digit year in century.
+                constexpr size_t cFieldLength = 2;
```
```suggestion
                constexpr size_t cFieldLength{2};
```

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+                if (timestamp_idx + cFieldLength > timestamp.size()) {
+                    return ErrorCodeEnum::IncompatibleTimestampPattern;
+                }
+
+                auto const two_digit_year_option{convert_padded_string_to_number(
```
LinZhihao-723: I think the convention is to call it `optional_two_digit_year`

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+                    parsed_year = two_digit_year + cTwoDigitYearLowOffset;
+                } else {
+                    parsed_year = two_digit_year + cTwoDigitYearHighOffset;
+                }
+                timestamp_idx += cFieldLength;
+                date_type_representation = true;
+                break;
+            }
+            case 'Y': {  // Zero-padded 4-digit year.
+                constexpr size_t cFieldLength = 4;
```
```suggestion
                constexpr size_t cFieldLength{4};
```

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+                if (timestamp_idx + cFieldLength > timestamp.size()) {
+                    return ErrorCodeEnum::IncompatibleTimestampPattern;
+                }
+
+                auto const parsed_year_option{convert_padded_string_to_number(
```
LinZhihao-723: Same

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+                        timestamp.substr(timestamp_idx),
+                        cAbbreviatedMonthNames
+                ))};
+                parsed_month = static_cast<int>(month_idx) + 1;
+                timestamp_idx += cAbbreviatedMonthNames.at(month_idx).length();
+                date_type_representation = true;
+                break;
+            }
+            case 'm': {  // Zero-padded month.
+                constexpr size_t cFieldLength = 2;
```
```suggestion
                constexpr size_t cFieldLength{2};
```

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+                if (timestamp_idx + cFieldLength > timestamp.size()) {
+                    return ErrorCodeEnum::IncompatibleTimestampPattern;
+                }
+
+                auto const parsed_month_option{convert_padded_string_to_number(
```
LinZhihao-723: Same

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+                if (parsed_month < cMinParsedMonth || parsed_month > cMaxParsedMonth) {
+                    return ErrorCodeEnum::IncompatibleTimestampPattern;
+                }
+
+                timestamp_idx += cFieldLength;
+                date_type_representation = true;
+                break;
+            }
+            case 'd': {  // Zero-padded day in month.
+                constexpr size_t cFieldLength = 2;
```
```suggestion
                constexpr size_t cFieldLength{2};
```

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+                if (timestamp_idx + cFieldLength > timestamp.size()) {
+                    return ErrorCodeEnum::IncompatibleTimestampPattern;
+                }
+
+                auto const parsed_day_option{convert_padded_string_to_number(
```
LinZhihao-723: Same

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+                if (parsed_day < cMinParsedDay || parsed_day > cMaxParsedDay) {
+                    return ErrorCodeEnum::IncompatibleTimestampPattern;
+                }
+
+                timestamp_idx += cFieldLength;
+                date_type_representation = true;
+                break;
+            }
+            case 'e': {  // Space-padded day in month.
+                constexpr size_t cFieldLength = 2;
```
```suggestion
                constexpr size_t cFieldLength{2};
```

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+                if (timestamp_idx + cFieldLength > timestamp.size()) {
+                    return ErrorCodeEnum::IncompatibleTimestampPattern;
+                }
+
+                auto const parsed_day_option{convert_padded_string_to_number(
```
LinZhihao-723: Same

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+                }
+                return ErrorCodeEnum::IncompatibleTimestampPattern;
+            }
+            default:
+                return ErrorCodeEnum::InvalidTimestampPattern;
+        }
+    }
+
+    // Do not allow trailing unmatched content.
+    if (false == (pattern_idx == pattern.size() && timestamp_idx == timestamp.size())) {
```
```suggestion
    if (pattern_idx != pattern.size() || timestamp_idx != timestamp.size())) {
```
Nit: easier to read?

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+    // timestamps.
+    if (date_type_representation && number_type_representation) {
+        return ErrorCodeEnum::InvalidTimestampPattern;
+    }
+
+    if (number_type_representation) {
+        return ErrorCodeEnum::FormatSpecifierNotImplemented;
+    }
+
+    auto year_month_day{date::year(parsed_year) / parsed_month / parsed_day};
```
```suggestion
    auto const year_month_day{date::year(parsed_year) / parsed_month / parsed_day};
```

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+    if (false == year_month_day.ok()) {
+        return ErrorCodeEnum::InvalidDate;
+    }
+
+    auto time_point = date::sys_days(year_month_day) + std::chrono::hours(0)
```
```suggestion
    auto const time_point = date::sys_days(year_month_day) + std::chrono::hours(0)
```

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+        std::string_view pattern,
+        [[maybe_unused]] std::string& generated_pattern
+) -> ystdlib::error_handling::Result<std::pair<epochtime_t, std::string_view>, ErrorCode> {
+    size_t pattern_idx{};
+    size_t timestamp_idx{};
+
+    int parsed_year{cDefaultYear};
+    int parsed_month{cDefaultMonth};
+    int parsed_day{cDefaultDay};
+    std::optional<int> day_of_week_idx;
```
LinZhihao-723: `optional_day_of_week_idx`?

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+                                                   .weekday_indexed()
+                                                   .weekday()
+                                           - date::Sunday)
+                                                  .count()};
+        if (actual_day_of_week_idx != day_of_week_idx.value()) {
+            return ErrorCodeEnum::InvalidDate;
+        }
+    }
+
+    epochtime_t epoch_nanoseconds{
```
```suggestion
    epochtime_t const epoch_nanoseconds{
```

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
```diff
@@ -0,0 +1,92 @@
...
+ *   - An `std::string_view` of the timestamp pattern that corresponds to the timestamp.
+ *     - The lifetime of the `std::string_view` is the least of `pattern` and `generated_pattern`.
+ * - The possible error codes from `clp_s::timestamp_parser::ErrorCodeEnum`:
+ *   - `InvalidTimestampPattern` if the pattern is illegal, per the specification.
+ *   - `IncompatibleTimestampPattern` if the pattern is not able to exactly consume the timestamp.
+ *   - `InvalidDate` if parsing was successful, but some components of the timestamp offer
+ *     conflicting information about the actual date (e.g., if the parsed day of the week doesn't
+ *     match up with the rest of the timestamp information).
+ *   - `FormatSpecifierNotImplemented` if the pattern contains format specifiers that have not been
+ *     implemented yet.
```
```suggestion
 * @param timestamp
 * @param pattern A timestamp pattern made up of literals, format specifiers, and potentially CAT
 * sequences.
 * @param generated_pattern A buffer where a newly-generated timestamp pattern can be written, if
 * necessary.
 * @return A result containing a pair, or an error code indicating the failure:
 * - The pair:
 *   - The timestamp in epoch nanoseconds.
 *   - An `std::string_view` of the timestamp pattern that corresponds to the timestamp.
 *     - The lifetime of the `std::string_view` is the least of `pattern` and `generated_pattern`.
 * - The possible error codes:
 *   - ErrorCodeEnum::InvalidTimestampPattern if the pattern is illegal, per the specification.
 *   - ErrorCodeEnum::IncompatibleTimestampPattern if the pattern is not able to exactly consume the
 *     timestamp.
 *   - ErrorCodeEnum::InvalidDate if parsing was successful, but some components of the timestamp
 *     offer conflicting information about the actual date (e.g., if the parsed day of the week
 *     doesn't match up with the rest of the timestamp information).
 *   - ErrorCodeEnum::FormatSpecifierNotImplemented if the pattern contains format specifiers that
 *     have not been implemented yet.
```
- I think we don't need to indent multi-line docstring.
- Same, we should document the error code in the above way.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -0,0 +1,382 @@
...
+           std::string_view{"Jul"},
+           std::string_view{"Aug"},
+           std::string_view{"Sep"},
+           std::string_view{"Oct"},
+           std::string_view{"Nov"},
+           std::string_view{"Dec"}};
+
+/**
+ * Converts a padded decimal integer string to an integer.
+ * @param str Substring containg the padded decimal integer string.
```
```suggestion
 * @param str Substring containing the padded decimal integer string.
```

components/core/src/clp_s/timestamp_parser/test/test_TimestampParser.cpp
LinZhihao-723, line 1: Can u also address clang-tidy violations in this test file?

LinZhihao-723 (changes requested):
Reviewed implementation.
Haven't reviewed:
- The test case content.
- CMake file changes.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
gibber9809: Yes -- just const here to keep clang-tidy happy. I wanted to have the variable even if it isn't used yet so that I could write the error handling for cases when the user tries to mix format specifiers for representing the timestamp as a date or number.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
gibber9809: Agreed.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
gibber9809: I think since we're keeping it in the empty namespace, and as far as I can tell the current implementation should cover all of the cases we'll need for parsing, we should just stick with the current implementation since it is a bit easier to use.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
gibber9809: I think we should be able to make that switch, yeah. Will try using range.

gibber9809: Actually, I'll probably go with span.

PR #1388, opened 2025-10-06
feat(mcp-server): Add ClpConnector for query job submission and result retrieval.

All-less:
# Description

This PR adds the connector between MCP server and CLP package. It enables the MCP to submit queries and read results.

# Checklist

  breaking change.

# Validation performed

```
cd clp
task package

cd components/clp-mcp-server
uv run pytest
```

## Summary by CodeRabbit

- New Features
  - Added a connector to submit time‑ranged searches, track job status, wait for completion with optional timeouts, and retrieve cached results (max 1,000).
  - Introduced environment-driven defaults for DB credentials and results‑cache configuration.
  - Standardized job statuses/types and a 1s polling interval.

- Tests
  - Added async tests for validation, status transitions, failure handling, and result retrieval.

- Chores
  - Updated runtime/dev dependencies and lint/test settings.

### DIFF
--- a/components/webui/client/src/components/SqlEditor/index.module.css
+++ b/components/webui/client/src/components/SqlEditor/index.module.css
@@ -4,4 +4,5 @@
     min-width: 0;
     width: 100%;
     transition: border-color 0.2s, box-shadow 0.2s;
+    overflow: hidden;
 }
--- a/components/webui/client/src/components/SqlInput/index.tsx
+++ b/components/webui/client/src/components/SqlInput/index.tsx
@@ -47,7 +47,7 @@ const SqlInput = (props: SqlEditorProps) => {
                 overviewRulerBorder: false,
                 overviewRulerLanes: 0,
                 padding: {
-                    top: 4,
+                    top: 6,
                     bottom: 4,
                 },
                 renderLineHighlight: "none",


components/clp-mcp-server/clp_mcp_server/settings.py
```diff
@@ -0,0 +1,13 @@
+"""Settings for CLP MCP server."""
+
+import os
+
+CLP_DB_SERVICE_NAME = os.environ.get("CLP_DB_SERVICE_NAME", "db")
```
junhaoliao: except the credentials, is it possible to accept the rest of these by a config file? (you may look at how `job_orchestration.reducer.reducer` accepts such config file for example)

components/clp-mcp-server/tests/test_clp_connector.py
```diff
@@ -0,0 +1,109 @@
+"""Tests for clp_connector.py."""
```
junhaoliao, line 1:
```suggestion
```

junhaoliao: ditto for the other files

components/clp-mcp-server/pyproject.toml
```diff
@@ -8,6 +8,9 @@ requires-python = ">=3.10"
 dependencies = [
     "click>=8.3.0",
     "fastmcp>=2.12.4",
+    "pymongo>=4.15.1",
```
junhaoliao, line 11: possible to use the same version as in https://github.com/y-scope/clp/blob/ba34c279d30206c123cb0ab27e16c3863e67bd44/components/clp-package-utils/pyproject.toml#L17 ?

Alternatively, we can upgrade the versions in the existing `pyproject.toml` files

That way, we don't necessary specify (and install) two different versions of a same dependency in the package.

components/clp-mcp-server/pyproject.toml
```diff
@@ -8,6 +8,9 @@ requires-python = ">=3.10"
...
+    "aiomysql>=0.2.0",
+    "msgpack>=1.1.1",
```
junhaoliao, line 13: ditto - let's try to use the same version across the CLP Python packages

https://github.com/y-scope/clp/blob/ba34c279d30206c123cb0ab27e16c3863e67bd44/components/clp-package-utils/pyproject.toml#L14

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,188 @@
+"""CLPConnector: A class to interact with the CLP database and results cache."""
+
+import asyncio
+
+import aiomysql
```
junhaoliao, line 6: instead of using this library directly, is it possible to use this adapter which handles connections to both mariadb / mysql?

https://github.com/y-scope/clp/blob/c9854b3d5436dc7e712b186fa4b9feaa6ff6a50b/components/clp-py-utils/clp_py_utils/sql_adapter.py

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,188 @@
...
+            "user": CLP_DB_USER,
+            "password": CLP_DB_PASS,
+            "db": CLP_DB_NAME,
+        }
+
+    async def submit_query(self, query: str, begin_ts: int, end_ts: int) -> str:
+        """
+        Submits a query to the CLP database and returns the ID of the query.
+
+        Args:
```
junhaoliao: for consistency, is it possible to write reST docstrings?

https://peps.python.org/pep-0287/
->
https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#field-lists

components/clp-mcp-server/clp_mcp_server/constants.py
junhaoliao, line 1: now that the pydantic incompatibility issue has been resolved, i believe it's possible to install the mcp server in the same environment as other CLP Python packages (which should already be the case?), and import some of these constants from the job_orchestration package?

components/clp-mcp-server/tests/test_clp_connector.py
All-less: This is to make the linter stop complaining. Should I configure the linter to bypass this rule?

components/clp-mcp-server/clp_mcp_server/clp_connector.py
All-less: The problem with`sql_adapter` is its synchronous design. The FastMCP framework (which our MCP server is built on) is async by design, and that's why I used `aiomysql` here.

That said, in the long run, I am all for a common connection library/layer for all CLP projects, but that's beyond the scope of this PR.

components/clp-mcp-server/pyproject.toml
All-less: PyMongo 4.7.2 does not have the `AsyncMongoClient` that I used in this project. Perhaps we can let the maintainer of `clp-package-utils` upgrade the version there?

components/clp-mcp-server/clp_mcp_server/constants.py
All-less: There are several implications if we want to directly import the constants (which, I agree, is the correct thing to do). The TLDR is that it breaks the unit tests, and the current build tool does not support it.

1. I see that current CLP Python packages are mostly managed as internal packages and not published into a registry. They reference each other by implicitly assuming they are installed in the same environment (e.g., the `clp-package-utils` directly uses `clp_py_utils` without specifying it as a dependency). This will work for `clp-mcp-server` during the runtime, but it will break the unit tests during development time. The tests are run in a standalone environment with no `job-orchestration` installed.

2. When I tried to add `job-orchestration` package as a dependency, it turned out to have a different build system from `clp-mcp-server`:  the former is using `poetry` and the latter `hatchling`. This conflict makes the package unable to be built.

Therefore, below are some of the possible workarounds, but they are all beyond the scope of this PR.
1. Remove the unit tests.
2. Change how the Python packages are managed.
3. Change the build system.

components/clp-mcp-server/clp_mcp_server/settings.py
All-less: It is a better idea, but I think it will involve some efforts from @20001020ycx 

Overall, we want to add a `config` parameter to the MCP server so that the user can pass in the path to the configuration file. Then the server populates a `clp_config` object and uses it to initialize the `CLPConnector` class. I will change the class constructor so that it reads those configurations directly from `clp_config`.

I'll talk to Chenxing to figure out if we want to do it in this PR or next.

20001020ycx: I haven't got the chance to look at the details of this pr, but while I was drafting the first pr, I think there's definitely some part of this pr can be applied to docker compose set up. I will add myself as the reviewer and we can work together on this pr then for the required docker compose change.

components/clp-mcp-server/tests/test_clp_connector.py
junhaoliao: sorry this linter rule is new to me. please ignore my comment

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,188 @@
...
+
+        :param query_id: The ID of the query.
+        :type query_id: str
+
+        :raises aiomysql.Error: If there is an error connecting to or querying MariaDB.
+        :raises ValueError: When the query is not found.
+
+        :return: The status of the query.
+        :rtype: int
+
```
junhaoliao: does the MCP server strictly need this rtype (and the parameter types in the docstrings of below functions) if we have specified the correct typing hints in the function prototype? if not, i feel we can omit such from the docstrings so they don't go out of sync with the type hints

components/clp-mcp-server/clp_mcp_server/settings.py
20001020ycx: Okay, I think I know what Junhao is suggesting, and I think it should be simple to implement, and it is more elegant. Here is the proposed change for which I will make directly on this pr with a separate commit:

1. make CLP_DB_PASS and CLP DB_USER as environment variables in docker compose
2. add infrastructure for making clp-config.yml as part of the argument for invoking mcp server. This way, you can access all names by simply reading the config file.

@All-less can you confirm this? If you are okay, I should be able to get something running in 30 mins

components/clp-mcp-server/pyproject.toml
20001020ycx, line 1: For the updated pyproject.toml, can you run `uv lock`. With the uv lock as the dependency management tool, we need to run `uv lock` every time the pyproject file is changed

20001020ycx, line 1: We also have other linting requirements specified in uv file, can you run `uv run ruff check .` under the clp-mcp-server folder to make sure the linting requirement is respected?

components/clp-mcp-server/clp_mcp_server/constants.py
20001020ycx, line 1: can you also backport the constant defined in server.py into this folder if constants.py is needed, this is subject to the discussion above.

components/clp-mcp-server/pyproject.toml
```diff
@@ -8,6 +8,9 @@ requires-python = ">=3.10"
 dependencies = [
     "click>=8.3.0",
     "fastmcp>=2.12.4",
+    "pymongo==4.15.1",
```
20001020ycx: let's always use the latest dependency with the same formatting, i.e >= version number

components/clp-mcp-server/clp_mcp_server/settings.py
All-less: The plan sounds good to me.

components/clp-mcp-server/pyproject.toml
All-less: Yes, it passed all the linting rules.

components/clp-mcp-server/pyproject.toml
All-less: I don't think this is the formatting issue. It's a decision whether you want to pin down a dependency version or continuously support more updated versions. I think @junhaoliao might have some opinions about this.

components/clp-mcp-server/clp_mcp_server/clp_connector.py
All-less: I've removed the types from docstrings.

components/clp-mcp-server/pyproject.toml
All-less: Updated.

components/clp-mcp-server/clp_mcp_server/constants.py
All-less: As we have discussed, the next PR will include the creation of `clp_config`. I think it will be a more suitable time to move the constants.

junhaoliao: Thanks for the detailed explanations! 

> Change the build system.

I believe this is the way to go - For the other Python components `clp-package-utils`, `clp-py-utils` and `job-orchestration`, we should migrate from poetry to uv. @coderabbitai create an issue to track this.

For now, i agree duplicating the definitions in `clp-mcp-server` is a proper approach.

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,173 @@
...
+    QueryJobType,
+    SEARCH_MAX_NUM_RESULTS,
+)
+from .settings import (
+    CLP_DB_PASS,
+    CLP_DB_USER,
+)
+
+
+class CLPConnector:
```
junhaoliao: Let's rename this as `ClpConnector`.

---

I understand the existing code does not consistently follow the convention (e.g, `CLPConfig` should really be `ClpConfig`). We aim to rectify those inconsistencies soon. @coderabbitai please create an issue to track the rectification

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,173 @@
...
+    """A connector class to interact with the CLP database and results cache."""
+
+    def __init__(self, clp_config: Any) -> None:
+        """Initializes the CLPConnector with MongoDB and MariaDB configurations."""
+        mongo_url = f"mongodb://{clp_config.results_cache.host}:{clp_config.results_cache.port}/"
+        self.mongo_client = AsyncMongoClient(mongo_url)
```
junhaoliao: `mongo_client` doesn't seem used anywhere else. do we have a plan to add teardown method in `ClpConnector`? if not, i believe we can keep `mongo_client` local and do no need to expose it as a public instance attribute

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,173 @@
...
+        self.results_cache = self.mongo_client[clp_config.results_cache.db_name]
```
junhaoliao: It seems `.results_cache` is not access anywhere outside of the class, except in the test cases. Can we rename it as a private attribute? (e.g. `_results_cache`)

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # Configuration to be used in `aiomysql.connect` to MariaDB.
+        self.db_conf = {
```
junhaoliao: ditto - i believe we can rename it as `_db_conf`

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,173 @@
...
+
+        :param query_id: The ID of the query.
+        :param timeout: Maximum time to wait in seconds, or None for no timeout.
+
+        :raises aiomysql.Error: If there is an error connecting to or querying MariaDB.
+        :raises ValueError: When the query is not found.
+        :raises RuntimeError: When the query fails or is cancelled.
+
+        """
+        waiting_states = {QueryJobStatus.PENDING, QueryJobStatus.RUNNING, QueryJobStatus.CANCELLING}
```
junhaoliao, line 125: (just a recommendation - feel free to reject)

would it be better to declare the states in a type definition source file (i.e., constants.py) so that whenever we add new status codes, it's less likely we would forget updating the `waiting_states` if they do need to be updated?

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,173 @@
...
+        :param begin_ts: The beginning timestamp of the query range.
+        :param end_ts: The end timestamp of the query range.
+
+        :raises ValueError: If ``end_ts`` is smaller than ``begin_ts``.
+        :raises aiomysql.Error: If there is an error connecting to or querying MariaDB.
+        :raises pymongo.errors.PyMongoError: If there is an error interacting with MongoDB.
+        :raises Exception: For any other unexpected errors.
+
+        :returns: The ID assigned to the query.
+
```
junhaoliao: they are just aliases of each other but we prefer non -s form
```suggestion
        :param query: The query string.
        :param begin_ts: The beginning timestamp of the query range.
        :param end_ts: The end timestamp of the query range.
        :raise ValueError: If ``end_ts`` is smaller than ``begin_ts``.
        :raise aiomysql.Error: If there is an error connecting to or querying MariaDB.
        :raise pymongo.errors.PyMongoError: If there is an error interacting with MongoDB.
        :raise Exception: For any other unexpected errors.
        :return: The ID assigned to the query.
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        Reads the job status of a query.
+
+        :param query_id: The ID of the query.
+
+        :raises aiomysql.Error: If there is an error connecting to or querying MariaDB.
+        :raises ValueError: When the query is not found.
+
+        :return: The status of the query.
+
```
junhaoliao:
```suggestion
        :param query_id: The ID of the query.
        :raise aiomysql.Error: If there is an error connecting to or querying MariaDB.
        :raise ValueError: When the query is not found.
        :return: The status of the query.
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        Waits for the query to complete with an optional timeout.
+
+        :param query_id: The ID of the query.
+        :param timeout: Maximum time to wait in seconds, or None for no timeout.
+
+        :raises aiomysql.Error: If there is an error connecting to or querying MariaDB.
+        :raises ValueError: When the query is not found.
+        :raises RuntimeError: When the query fails or is cancelled.
+
```
junhaoliao:
```suggestion
        :param query_id: The ID of the query.
        :param timeout: Maximum time to wait in seconds, or None for no timeout.
        :raise aiomysql.Error: If there is an error connecting to or querying MariaDB.
        :raise ValueError: When the query is not found.
        :raise RuntimeError: When the query fails or is cancelled.
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,173 @@
...
+                raise RuntimeError(err_msg)
+
+    async def read_results(self, query_id: str) -> list[dict]:
+        """
+        Reads the results of a query.
+
+        :param query_id: The ID of the query.
+
+        :return: A list of result documents.
+
```
junhaoliao:
```suggestion
        :param query_id: The ID of the query.
        :return: A list of result documents.
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,173 @@
...
+        :param query_id: The ID of the query.
+        :param timeout: Maximum time to wait in seconds, or None for no timeout.
+
+        :raises aiomysql.Error: If there is an error connecting to or querying MariaDB.
+        :raises ValueError: When the query is not found.
+        :raises RuntimeError: When the query fails or is cancelled.
+
+        """
+        waiting_states = {QueryJobStatus.PENDING, QueryJobStatus.RUNNING, QueryJobStatus.CANCELLING}
+        start_time = asyncio.get_event_loop().time()
```
junhaoliao: how about extracting an event_loop variable to avoid getting it in the while
```suggestion
        event_loop = asyncio.get_event_loop()
        start_time = event_loop.time()
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,173 @@
...
+                raise RuntimeError(err_msg)
+            elif status == QueryJobStatus.CANCELLED:
+                err_msg = f"Query job with ID {query_id} was cancelled."
+                raise RuntimeError(err_msg)
+            elif status == QueryJobStatus.KILLED:
+                err_msg = f"Query job with ID {query_id} was killed."
+                raise RuntimeError(err_msg)
+            else:
+                err_msg = f"Query job with ID {query_id} has unknown status {status}."
+                raise RuntimeError(err_msg)
```
junhaoliao, line 139:
```suggestion
        while True:
            status = await self.read_job_status(query_id)
            if status in waiting_states:
                await asyncio.sleep(POLLING_INTERVAL_SECONDS)
                if timeout and (event_loop.time() - start_time) > timeout:
                    err_msg = f"Timeout waiting for query job {query_id} to complete."
                    raise TimeoutError(err_msg)
                continue
            
            if status == QueryJobStatus.SUCCEEDED:
                break # or maybe we can just `return`?
            
            # shall we extract an `error_states` const like the waiting_states?
            if status in {QueryJobStatus.FAILED, QueryJobStatus.CANCELLED, QueryJobStatus.KILLED}:
                err_msg = f"Query job {query_id} ended with status {status.name}.")
                raise RuntimeError(err_msg)

            err_msg = f"Query job {query_id} has unknown status {status}."
            raise RuntimeError(err_msg)
```

junhaoliao: or we could get rid of the error state check and just write once the status isnt `QueryJobStatus.SUCCEEDED`
```py
err_msg = f"Query job {query_id} ended with status {status.name}.")
raise RuntimeError(err_msg)
```

components/clp-mcp-server/pyproject.toml
```diff
@@ -65,3 +72,8 @@ isort.order-by-type = false
 [tool.ruff.format]
 docstring-code-format = true
 docstring-code-line-length = 100
+
+[tool.ruff.lint.per-file-ignores]
+    "tests/test_clp_connector.py" = ["INP001", "E402"]
+    "tests/*.py" = ["S101"]
+    "clp_mcp_server/clp_connector.py" = ["I001"]
```
junhaoliao: 1. let's make sure files are terminated with an empty LF
2. it's always worth mentioning why we want to supress some linter rule (so in the future developers better understand if a rule should / should not be lifted)
3. why do we want to supress E402 and I001?
```suggestion
"tests/**" = [
  "INP001",  # Allow implicit namespace package for tests
  "S101",  # Allow use of `assert` (security warning)
]
"tests/test_clp_connector.py" = [
  "E402"
]
"clp_mcp_server/clp_connector.py" = [
  "I001"  
]

```

components/clp-mcp-server/tests/test_clp_connector.py
```diff
@@ -0,0 +1,132 @@
...
+
+    mock_docs = [{"_id": "1"}, {"_id": "2"}, {"_id": "3"}]
+    mock_collection = AsyncMock()
+
+    async def async_gen() -> AsyncIterator[dict]:
+        for doc in mock_docs:
+            yield doc
+
+    mock_collection.find = MagicMock(return_value=async_gen())
+    connector.results_cache = { "12": mock_collection }
```
junhaoliao: instead of monkey patching, would `patch.object` be less error prone? e.g.,
```
@pytest.mark.asyncio
async def test_read_results_returns_docs() -> None:
    """Tests reading results returns expected documents."""
    clp_config = gen_mock_clp_config()

    # Prepare mock Mongo collection and async generator
    mock_docs = [{"_id": "1"}, {"_id": "2"}, {"_id": "3"}]
    mock_collection = AsyncMock()

    async def async_gen() -> AsyncIterator[dict]:
        for doc in mock_docs:
            yield doc

    mock_collection.find = MagicMock(return_value=async_gen())

    with patch.object(clp_connector.ClpConnector, "_results_cache", {"12": mock_collection}):
        connector = clp_connector.ClpConnector(clp_config)
        results = await connector.read_results("12")

    assert results == mock_docs
```

components/clp-mcp-server/tests/test_clp_connector.py
```diff
@@ -0,0 +1,132 @@
+"""Tests for clp_connector.py."""
+from collections.abc import AsyncIterator
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from clp_mcp_server import clp_connector
```
junhaoliao: possible to write this instead?

```py
from clp_mcp_server.clp_connector import CLPConnector, QueryJobStatus
```

components/clp-mcp-server/tests/test_clp_connector.py
```diff
@@ -0,0 +1,132 @@
...
+    class MockDatabase:
+        host = "localhost"
+        port = 3306
+        name = "mock_clp_db"
+
+    class MockCLPConfig:
+        results_cache = MockResultsCache()
+        database = MockDatabase()
+
+    return MockCLPConfig()
```
junhaoliao: how about

```suggestion
@pytest.fixture
def mock_clp_config():
    return SimpleNamespace(
        results_cache=SimpleNamespace(host="localhost", port=27017, db_name="mock_results_cache"),
        database=SimpleNamespace(host="localhost", port=3306, name="mock_clp_db"),
    )
```

then every test can just use the fixture from the args. e.g., 
```
@pytest.mark.asyncio
async def test_submit_query_invalid_timestamps(mock_clp_config) -> None:
    """Tests submitting a query with invalid timestamps."""
    connector = clp_connector.CLPConnector(mock_clp_config)
    with pytest.raises(ValueError, match="smaller than"):
        await connector.submit_query("test", 100, 50)
```

components/clp-mcp-server/tests/test_clp_connector.py
junhaoliao: also, is it possible to inline the generator? e.g.,
```
mock_collection.find = MagicMock(return_value=(doc async for doc in mock_docs))
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
All-less: It's a good idea, but it's better to do it in `job_orchestration`. Let's wait until [the build system conflict](https://github.com/y-scope/clp/pull/1388#discussion_r2414832019) is solved.

components/clp-mcp-server/pyproject.toml
```diff
@@ -65,3 +72,14 @@ isort.order-by-type = false
...
+[tool.ruff.lint.per-file-ignores]
+    "tests/test_clp_connector.py" = [
+        "INP001",  # Allow implicit namespace package for tests
+        "S101",  # Allow "assert" in test cases
+    ]
+    "*.py" = [
+        "I001", # Allow import statements to be grouped into standard library, third-party 
+                # packages, and local application/library specific imports, as recommended 
+                # by PEP 8.
+    ]
```
junhaoliao:
```suggestion
    ]

```

components/clp-mcp-server/tests/test_clp_connector.py
```diff
@@ -0,0 +1,116 @@
...
+        await connector.wait_query_completion("fail_id")
+
+
+@pytest.mark.asyncio
+async def test_read_results_returns_docs(mock_clp_config: Any) -> None:
+    """Tests reading results returns expected documents."""
+    mock_docs = [{"_id": "1"}, {"_id": "2"}, {"_id": "3"}]
+    mock_collection = AsyncMock()
+
+    mock_collection.find = MagicMock(return_value=(doc async for doc in mock_docs))
```
junhaoliao: 1. let's group the declarations
2. my apology for suggesting the `async for` approach. it doesn't work because Python would expect `mock_docs` to be a async iterable but a list isn't one. let's add this helper at the end of the file:
```
from collections.abc import AsyncGenerator, Iterable
from typing import Any, TypeVar
...

T = TypeVar("T")
async def _aiter(it: Iterable[T]) -> AsyncGenerator[T, None]:
    """
    Yields items from an iterable asynchronously.

    :param it:
    :yield: Items from the iterable.
    """
    for i in it:
        yield i
```

```suggestion
    mock_collection.find = MagicMock(return_value=_aiter(mock_docs))
```

components/clp-mcp-server/pyproject.toml
junhaoliao: i think it's okay to only enforce a minimum version rather than locking to a specific one

components/clp-mcp-server/pyproject.toml
```diff
@@ -8,6 +8,9 @@ requires-python = ">=3.10"
 dependencies = [
     "click>=8.3.0",
     "fastmcp>=2.12.4",
+    "pymongo==4.15.1",
+    "aiomysql>=0.2.0",
+    "msgpack>=1.0.7",
```
junhaoliao: my apology for missing the fact that the mcp server is set up as a separate project from the other package components (which is really the right thing to do). in that case, we don't have to keep the dependencies version the same as the existing packages'
```suggestion
    "pymongo>=4.15.3",
    "aiomysql>=0.2.0",
    "msgpack>=1.1.2",
```

components/clp-mcp-server/pyproject.toml
junhaoliao: i removed the rule suppression and ran the linter locally. it appears that the order the linter enforces should be correct?
```diff
-from typing import Any
 from types import SimpleNamespace
+from typing import Any
 from unittest.mock import AsyncMock, MagicMock, patch
```

if so, let's remove the suppression
```suggestion
```

components/clp-mcp-server/tests/test_clp_connector.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@pytest.fixture
+def mock_clp_config() -> Any:
+    """Provides a mock CLP configuration for testing."""
+    return SimpleNamespace(
+        results_cache=SimpleNamespace(host="localhost", port=27017, db_name="mock_results_cache"),
+        database=SimpleNamespace(host="localhost", port=3306, name="mock_clp_db"),
+    )
+
```
junhaoliao, line 24: PEP8
```suggestion


```

components/clp-mcp-server/tests/test_clp_connector.py
```diff
@@ -0,0 +1,116 @@
...
+        await connector.submit_query("test", 100, 50)
+
+
+@pytest.mark.asyncio
+async def test_read_job_status_not_found(mock_clp_config: Any) -> None:
+    """Tests reading job status for a non-existent query."""
+    connector = ClpConnector(mock_clp_config)
+    mock = AsyncMock(side_effect=ValueError("not found"))
+    with patch.object(connector, "read_job_status", mock), \
+        pytest.raises(ValueError, match="not found"):
```
junhaoliao:
```suggestion
    with (patch.object(connector, "read_job_status", mock),
          pytest.raises(ValueError, match="not found")):
```

components/clp-mcp-server/tests/test_clp_connector.py
```diff
@@ -0,0 +1,116 @@
...
+    (QueryJobStatus.FAILED, RuntimeError),
+    (QueryJobStatus.CANCELLED, RuntimeError),
+    (QueryJobStatus.KILLED, RuntimeError),
+    (999, RuntimeError),  # unknown status
+])
+async def test_wait_query_completion_failure_cases(
+        fail_status: QueryJobStatus,
+        exc_type: Exception,
+        mock_clp_config: Any
+    ) -> None:
```
junhaoliao: sorry for missing this in the last review:
1. the typing of `exc_type` should be the type of `Exception` rather than an Exception instance.
2. the indent seems inconsistent
```suggestion
async def test_wait_query_completion_failure_cases(
    fail_status: QueryJobStatus,
    exc_type: type[Exception],
    mock_clp_config: Any
) -> None:
```

components/clp-mcp-server/tests/test_clp_connector.py
```diff
@@ -0,0 +1,116 @@
...
+async def test_read_results_returns_docs(mock_clp_config: Any) -> None:
+    """Tests reading results returns expected documents."""
+    mock_docs = [{"_id": "1"}, {"_id": "2"}, {"_id": "3"}]
+    mock_collection = AsyncMock()
+
+    mock_collection.find = MagicMock(return_value=(doc async for doc in mock_docs))
+
+    with patch.object(ClpConnector, "_results_cache", {"12": mock_collection}):
+        connector = ClpConnector(mock_clp_config)
+        results = await connector.read_results("12")
```
junhaoliao: sorry for not proofreading my previous suggestion
```suggestion
    connector = ClpConnector(mock_clp_config)
    with patch.object(connector, "_results_cache", {"12": mock_collection}):
        results = await connector.read_results("12")
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,162 @@
...
+
+        :param query: The query string.
+        :param begin_ts: The beginning timestamp of the query range.
+        :param end_ts: The end timestamp of the query range.
+        :raise ValueError: If ``end_ts`` is smaller than ``begin_ts``.
+        :raise aiomysql.Error: If there is an error connecting to or querying MariaDB.
+        :raise pymongo.errors.PyMongoError: If there is an error interacting with MongoDB.
+        :raise Exception: For any other unexpected errors.
+        :return: The ID assigned to the query.
+
```
junhaoliao: 1. list `:return` before `:raise`
2. no need to insert empty line at the end of the docstring block
```suggestion
        :param query: The query string.
        :param begin_ts: The beginning timestamp of the query range.
        :param end_ts: The end timestamp of the query range.
        :return: The ID assigned to the query.
        :raise ValueError: If ``end_ts`` is smaller than ``begin_ts``.
        :raise aiomysql.Error: If there is an error connecting to or querying MariaDB.
        :raise pymongo.errors.PyMongoError: If there is an error interacting with MongoDB.
        :raise Exception: For any other unexpected errors.
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,162 @@
...
+    async def read_job_status(self, query_id: str) -> int:
+        """
+        Reads the job status of a query.
+
+        :param query_id: The ID of the query.
+        :raise aiomysql.Error: If there is an error connecting to or querying MariaDB.
+        :raise ValueError: When the query is not found.
+        :return: The status of the query.
+
```
junhaoliao:
```suggestion
        :param query_id: The ID of the query.
        :return: The status of the query.
        :raise aiomysql.Error: If there is an error connecting to or querying MariaDB.
        :raise ValueError: When the query is not found.
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,162 @@
...
+    async def wait_query_completion(self, query_id: str, timeout: float | None = None) -> None:
+        """
+        Waits for the query to complete with an optional timeout.
+        :param query_id: The ID of the query.
+        :param timeout: Maximum time to wait in seconds, or None for no timeout.
+        :raise aiomysql.Error: If there is an error connecting to or querying MariaDB.
+        :raise ValueError: When the query is not found.
+        :raise RuntimeError: When the query fails or is cancelled.
+
```
junhaoliao:
```suggestion
        Waits for the query to complete with an optional timeout.

        :param query_id: The ID of the query.
        :param timeout: Maximum time to wait in seconds, or None for no timeout.
        :raise aiomysql.Error: If there is an error connecting to or querying MariaDB.
        :raise ValueError: When the query is not found.
        :raise RuntimeError: When the query fails or is cancelled.
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,162 @@
...
+            "errorMsg": None,
+            "errorName": None,
+            "lastSignal": "resp-querying",
+            "queryEngine": "clp",
+        }
+        await self._results_cache["results-metadata"].insert_one(results_metadata_doc)
+
+        return query_id
+
+    async def read_job_status(self, query_id: str) -> int:
```
junhaoliao:
```suggestion
    async def read_job_status(self, query_id: str) -> QueryJobStatus:
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,162 @@
...
+        start_time = event_loop.time()
+
+        while True:
+            status = await self.read_job_status(query_id)
+            if status in waiting_states:
+                await asyncio.sleep(POLLING_INTERVAL_SECONDS)
+                if timeout and (event_loop.time() - start_time) > timeout:
+                    err_msg = f"Timeout waiting for query job with ID {query_id} to complete."
+                    raise TimeoutError(err_msg)
+            elif status == QueryJobStatus.SUCCEEDED:
```
junhaoliao: no need to use `elif` since the above block would interrupt the control flow anyways
```suggestion
            if status == QueryJobStatus.SUCCEEDED:
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,162 @@
...
+                break
+            elif status in error_states:
```
junhaoliao: ditto - the control flow would have been interrupted by the bodies of above if-blocks
```suggestion
            if status in error_states:
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,162 @@
...
+                err_msg = f"Query job with ID {query_id} ended in status {status.name}."
+                raise RuntimeError(err_msg)
+            else:
+                err_msg = f"Query job with ID {query_id} has unknown status {status}."
+                raise RuntimeError(err_msg)
```
junhaoliao: ditto - the control flow would have been interrupted by the bodies of above if-blocks
```suggestion

            err_msg = f"Query job with ID {query_id} has unknown status {status}."
            raise RuntimeError(err_msg)
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,162 @@
...
+
+    async def read_results(self, query_id: str) -> list[dict]:
+        """
+        Reads the results of a query.
+
+        :param query_id: The ID of the query.
+        :return: A list of result documents.
+
+        """
```
junhaoliao, line 153:
```suggestion
        """
        Reads the results of a query.

        :param query_id: The ID of the query.
        :return: A list of result documents.
        """
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -0,0 +1,162 @@
...
+        :param timeout: Maximum time to wait in seconds, or None for no timeout.
+        :raise aiomysql.Error: If there is an error connecting to or querying MariaDB.
+        :raise ValueError: When the query is not found.
+        :raise RuntimeError: When the query fails or is cancelled.
+
+        """
+        waiting_states = {QueryJobStatus.PENDING, QueryJobStatus.RUNNING, QueryJobStatus.CANCELLING}
+        error_states = {QueryJobStatus.FAILED, QueryJobStatus.CANCELLED, QueryJobStatus.KILLED}
+        event_loop = asyncio.get_event_loop()
+        start_time = event_loop.time()
```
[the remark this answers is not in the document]
junhaoliao: i agree with the rabbit

components/clp-mcp-server/tests/test_clp_connector.py
```diff
@@ -0,0 +1,129 @@
...
+) -> None:
+    """Tests waiting for a query that ends in failure, cancellation, or unknown status."""
+    connector = ClpConnector(mock_clp_config)
+    connector.read_job_status = AsyncMock(return_value=fail_status)
+    with pytest.raises(exc_type):
+        await connector.wait_query_completion("fail_id")
+
+
+T = TypeVar("T")
+async def _aiter(it: Iterable[T]) -> AsyncGenerator[T, None]:
```
junhaoliao: let's move this helper to the end of the source file (no need to forward declare)

components/clp-mcp-server/tests/test_clp_connector.py
junhaoliao: @All-less did the linter complain about this?

components/clp-mcp-server/clp_mcp_server/clp_connector.py
All-less: `elif` can avoid some unnecessary comparisons.

components/clp-mcp-server/clp_mcp_server/clp_connector.py
All-less: `elif` makes the whole flow easier to read and understand.

components/clp-mcp-server/clp_mcp_server/clp_connector.py
All-less: I think you have mistakenly assumed there's a `continue` after L135. The current structure is more like a `switch-case`.

junhaoliao: Right, sorry for the mistake. it's fine to keep the existing code then.

if we really want to avoid the `switch-case` structure, i guess we can write
```py
        while True:
            if timeout is not None and (event_loop.time() - start_time) > timeout:
                err_msg = f"Timeout waiting for query job with ID {query_id} to complete."
                raise TimeoutError(err_msg)

            status = await self.read_job_status(query_id)
            if status == QueryJobStatus.SUCCEEDED:
                break
            if status in error_states:
                err_msg = f"Query job with ID {query_id} ended in status {status.name}."
                raise RuntimeError(err_msg)
            if status in error_states:
                err_msg = f"Query job with ID {query_id} ended in status {status.name}."
                raise RuntimeError(err_msg)
            if status not in waiting_states:
                err_msg = f"Query job with ID {query_id} has unknown status {status}."
                raise RuntimeError(err_msg)

            await asyncio.sleep(POLLING_INTERVAL_SECONDS)
```

i do think it's worth checking the status right after the `.sleep()` before raising a TimeoutError, because a query may happen to be completed in the wait period. in that case, we may want to extract a helper to do the checks. 
that said, what we're discussing is about edge-case handling and should not block subsequence PRs. we can create an issue and get that fixed in another PR if you agree

junhaoliao (dismissed):
there's one more suggestion (sorry for not raising it earlier), which is optional to be applied

for the title, how about:
```
feat(mcp-server): Add ClpConnector for query job submission and result retrieval.
```

PR #1393, opened 2025-10-07
feat(webui): Apply border radius to internal monaco editor elements; Improve vertical text alignment. 

davemarco:
# Description

Previously border radius only applied to outer border, but monaco had additional elements that had no radius, which looked a but weird on rendering. Also, the text in the single-line editor alignment was off, this PR aligns closer to the centre.

# Checklist

  breaking change.

# Validation performed


components/webui/client/src/components/SqlEditor/index.module.css
```diff
@@ -5,3 +5,9 @@
     width: 100%;
     transition: border-color 0.2s, box-shadow 0.2s;
 }
+
+/* Apply border radius to Monaco editor internal elements */
+.editor :global(.monaco-editor) {
```
hoophalab: Not sure whether it's a good idea, but can we name this style as `.monacoEditor` and add a `className={styles["monacoEditor"] || ""}` to the `Editor` component in `components/webui/client/src/components/SqlEditor/index.tsx`? I feel this is more conventional when using module css?

components/webui/client/src/components/SqlEditor/index.module.css
```diff
@@ -5,3 +5,9 @@
...
+    border-radius: var(--ant-border-radius);
+    overflow: hidden  !important;
```
hoophalab:
```suggestion
    overflow: hidden !important;
```

components/webui/client/src/components/SqlEditor/index.module.css
davemarco: im not fully clear about what ur suggesting. like do u just want to rename .editor .monaco-editor for both styles

hoophalab: I could be wrong, but I want to not use `:global` selector in css. Because in an unlikely chance where they change the name, or we change how we import monaco, this class name could change.

Instead, we should be able to simply add a class name to the main component. This doesn't do the same thing as `:global(.monaco-editor)`, but because of the "overflow" property, it fixes the issue on my machine. Let me know if this causes issue.

```
            <Editor
                className={styles["monacoEditor"] || ""} <---------
                language={"sql"}
                loading={
                    <div
                        style={{
                            backgroundColor: token.colorBgContainer,
                            height: "100%",
                            width: "100%",
                        }}/>
                }
                theme={disabled ?
                    "disabled-theme" :
                    "default-theme"}
                onMount={handleEditorDidMount}
                {...editorProps}/>
```

davemarco: ur right it seems just changing the overflow on top level component does fix the issue without the border radius applied to lower level component. I just left the class name as editor since this is the sql editor class

hoophalab:
We probably need to change the title. How about this title:
```
feat(webui): Apply `overflow: hidden` to `SqlEditor` to fix border radius. Improve vertical text alignment.
```
Does ```Apply `overflow: hidden` to `SqlEditor` to fix border radius. ``` sound a bit weird? better suggestions?

PR #1398, opened 2025-10-09
feat(webui): Remove limit from guided UI and set default limit in guided queries to 1000 results.

davemarco:
# Description

Removes limit, and reallocates space in grid to fill new empty space

# Checklist

  breaking change.

# Validation performed

Ran query

## Summary by CodeRabbit

- Refactor
  - Simplified Presto guided search by removing the per-query Limit control; queries now run with a standard limit of 1000 results.
  - Streamlined state usage for improved performance without changing behaviour.

- Style
  - Updated layout of Presto guided controls for clearer arrangement of fields (Where now spans two columns; Timestamp Key and Order positions adjusted).
  - Improved automatic row sizing for a more consistent UI.

### DIFF
--- a/components/clp-py-utils/clp_py_utils/clp_config.py
+++ b/components/clp-py-utils/clp_py_utils/clp_config.py
@@ -1,7 +1,7 @@
 import os
 import pathlib
 from enum import auto
-from typing import Any, Literal, Optional, Set, Union
+from typing import Annotated, Any, Literal, Optional, Set, Union
 
 from dotenv import dotenv_values
 from pydantic import (
@@ -14,7 +14,7 @@ from pydantic import (
 )
 from strenum import KebabCaseStrEnum, LowercaseStrEnum
 
-from .clp_logging import get_valid_logging_level, is_valid_logging_level
+from .clp_logging import LoggingLevel
 from .core import (
     get_config_value,
     make_config_path_absolute,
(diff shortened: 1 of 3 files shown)


components/webui/client/src/pages/SearchPage/SearchControls/Presto/presto-guided-search-requests.ts
```diff
@@ -72,13 +72,11 @@ const buildPrestoGuidedQueries = (timeRange: [Dayjs, Dayjs]) => {
 
     const trimmedWhere = where.trim();
     const trimmedOrderBy = orderBy.trim();
-    const limitString = String(limit);
 
     const searchQueryString = buildSearchQuery({
         ...(trimmedWhere && {booleanExpression: trimmedWhere}),
         databaseName: from,
         endTimestamp: endTimestamp,
-        ...(limitString && {limitValue: limitString}),
```
hoophalab, line 81: I feel we should prepare sql-parser for other use cases. How about adding `const DEFAULT_SEARCH_LIMIT = 1000;` to this file and pass ` DEFAULT_SEARCH_LIMIT` here?
```
limitValue: DEFAULT_SEARCH_LIMIT,
```

this way we can keep sqlparser generic

davemarco: but there is no limit string anymore since we removed it

components/webui/client/src/sql-parser/typings.ts
```diff
@@ -21,6 +21,7 @@ interface BuildTimelineQueryProps {
     timestampKey: string;
 }
 
+
```
hoophalab:
```suggestion
```

PR #1401, opened 2025-10-09
feat(clp-mcp-server): Introduce session-based MCP capabilities:

20001020ycx:
- Add `get_instructions` MCP tool call to retrieve the system prompt.
- Add `get_nth_page` MCP tool call to retrieve paginated query results cached from the previous search.
- Introduce `SessionManager` for managing asynchronous session-level states.

# Description

In our open-source release of the clp-mcp-server, we support 4 APIs: `search_kql_query`, `search_kql_query_with_timestamp`, `get_nth_page`, and `get_instructions`. In this PR, we implement the utility API: `get_nth_page` and `get_instructions`. For an overview of all four APIs, please refer to the internal documentation of the first MCP server release.

The documentation of `get_nth_page` and `get_instructions` are listed below:

## `get_nth_page(unsigned: page-idx)`

Description:
Retrieves the n-th page of a paginated response with the paging metadata from the previous query. This tool is used when a query’s response is too long and LLM only wishes to examine a specific subset of the results.

Inputs:
page-idx: the index of the page LLM wants to look into

Outputs:
The set of log messages corresponding to the requested page index.

Errors:
- “Error: No previous paginated response in this session”: If no query with a paginated response was run before running “get_nth_page” in this session then this error message is returned to the Agent. 
- “Error: page-idx is out of bounds” If page-idx is greater than num pages returned by a search query

## `get_instructions()`

Description:
Returns a pre-defined “system prompt” that guides the LLM’s behavior.

The instruction covers:
1. What outputs should be produced, and the desired output style.
2. How to use the provided tools.
3. Specific fine-tuning for the LLM (e.g., avoid premature conclusions, ask LLM for more input when necessary).
This function must be invoked before any other tool. 

Inputs:
empty

Outputs:
“system prompt” mentioned in Description.

## Implementation Detail

get_nth_page: 
- Implementation: The MCP server caches only the results of the most recent query, with a maximum of 1,000 log entries stored in memory.
- Scalability: If multiple users issue get_nth_page requests, the MCP server maintains an array of paginated responses. These are indexed by a session argument, which uniquely identifies each conversation between a user and the MCP server. The MCP server shall support multiple concurrent clients, where each client only makes synchronous API calls. 

get_instructions: 
- The MCP server stores a constant string containing the “system prompt”, it returns the string as the response when this tool is invoked by LLM. 
- In the tool description, we will mention that this should be the first tool to call in a session.
- In addition, we will implement an explicit static check to ensure this tool is called before any other tool. This can be done by using a session specific global flag which is set to true by the first tool call. If the first tool call is not `get_instructions` then we return an error to the AI saying that it should call get_instructions first.

# Checklist

  breaking change.

# Validation performed

1. **Unit Tests**: `uv run --group dev pytest tests/test_session_manager.py -v`.
2. **Lint Tests**: `uv run ruff check clp/components/clp-mcp-server` && `task lint:check-py`.
3. **Manual Integration Tests**: `task package` && `python3 -m clp_mcp_server.clp_mcp_server --host 127.0.0.1 --port 8000`. Connected a Claude Desktop agent to the running MCP server and issued a get_instruction query, successfully receiving the predefined system prompt. You can see the full end-to-end result using this [link](https://claude.ai/share/87003011-df5f-4051-9da8-0fcd0b659676)
Note: manual integration testing for get_nth_page was skipped due to the missing substring_search implementation; its correctness is verified through unit tests instead.
Screenshot 2025-10-09 at 10 41 42 PM

# Future work

To keep this PR focused, we are intentionally deferring the following non-trivial tasks:
1. *Prompt Pruning*: Author high-quality docstrings for each MCP API and refine the server-level system prompt. The goal is not only to make human maintainers understandable, but more importantly, to maximize LLM effectiveness when invoking the APIs.
5. *Integration Tests*: The validation performed is only on unit tests, plus the limited manual end-to-end testing performed by the PR author. Full integration testing is incomplete due to the missing substring search implementation. The AI team is actively developing a comprehensive integration test suite, which will be the immediate next step following the first release.
6. *GitHub Workflow on API Documentation*:  Because API docstrings directly impact LLM performance, clarity is the top priority. We will add a CI workflow that automatically generates and updates API documentation whenever docstrings change or new APIs are added.

## Summary by CodeRabbit

* **New Features**
  * In-app system instructions (KQL guidance) and a tool to fetch specific paginated result pages.

* **Improvements**
  * Per-session paginated caching with configurable TTL, page size and cache limits; session lifecycle management with periodic cleanup and async tooling; consolidated server settings/constants.

* **Tests**
  * Extensive unit tests covering pagination, session lifecycle, expiry, concurrency and edge cases.

* **Chores**
  * Dependency and lint config updates; tests package init and README testing instructions.

### DIFF
--- a/components/core/src/clp_s/SchemaReader.cpp
+++ b/components/core/src/clp_s/SchemaReader.cpp
@@ -63,7 +63,7 @@ SchemaReader::load(std::shared_ptr<char[]> stream_buffer, size_t offset, size_t 
     }
 }
 
-void SchemaReader::generate_json_string() {
+auto SchemaReader::generate_json_string(uint64_t message_index) -> std::string {
     m_json_serializer.reset();
     m_json_serializer.begin_document();
     size_t column_id_index = 0;
@@ -100,14 +100,14 @@ void SchemaReader::generate_json_string() {
                 auto name = m_global_schema_tree->get_node(column->get_id()).get_key_name();
                 m_json_serializer.append_key(name);
                 m_json_serializer.append_value(
-                        std::to_string(std::get<int64_t>(column->extract_value(m_cur_message)))
+                        std::to_string(std::get<int64_t>(column->extract_value(message_index)))
                 );
                 break;
             }
             case JsonSerializer::Op::AddIntValue: {
                 column = m_reordered_columns[column_id_index++];
                 m_json_serializer.append_value(
-                        std::to_string(std::get<int64_t>(column->extract_value(m_cur_message)))
+                        std::to_string(std::get<int64_t>(column->extract_value(message_index)))
                 );
                 break;
             }
@@ -116,14 +116,14 @@ void SchemaReader::generate_json_string() {
                 auto name = m_global_schema_tree->get_node(column->get_id()).get_key_name();
                 m_json_serializer.append_key(name);
                 m_json_serializer.append_value(
-                        std::to_string(std::get<double>(column->extract_value(m_cur_message)))
+                        std::to_string(std::get<double>(column->extract_value(message_index)))
                 );
                 break;
             }
             case JsonSerializer::Op::AddFloatValue: {
                 column = m_reordered_columns[column_id_index++];
                 m_json_serializer.append_value(
-                        std::to_string(std::get<double>(column->extract_value(m_cur_message)))
+                        std::to_string(std::get<double>(column->extract_value(message_index)))
                 );
                 break;
             }
@@ -131,12 +131,12 @@ void SchemaReader::generate_json_string() {
                 column = m_reordered_columns[column_id_index++];
                 auto name = m_global_schema_tree->get_node(column->get_id()).get_key_name();
                 m_json_serializer.append_key(name);
-                m_json_serializer.append_value_from_column(column, m_cur_message);
+                m_json_serializer.append_value_from_column(column, message_index);
                 break;
             }
             case JsonSerializer::Op::AddFormattedFloatValue: {
                 column = m_reordered_columns[column_id_index++];
-                m_json_serializer.append_value_from_column(column, m_cur_message);
+                m_json_serializer.append_value_from_column(column, message_index);
                 break;
             }
             case JsonSerializer::Op::AddBoolField: {
@@ -144,7 +144,7 @@ void SchemaReader::generate_json_string() {
                 auto name = m_global_schema_tree->get_node(column->get_id()).get_key_name();
                 m_json_serializer.append_key(name);
                 m_json_serializer.append_value(
-                        std::get<uint8_t>(column->extract_value(m_cur_message)) != 0 ? "true"
+                        std::get<uint8_t>(column->extract_value(message_index)) != 0 ? "true"
                                                                                      : "false"
                 );
                 break;
@@ -152,7 +152,7 @@ void SchemaReader::generate_json_string() {
             case JsonSerializer::Op::AddBoolValue: {
                 column = m_reordered_columns[column_id_index++];
                 m_json_serializer.append_value(
-                        std::get<uint8_t>(column->extract_value(m_cur_message)) != 0 ? "true"
+                        std::get<uint8_t>(column->extract_value(message_index)) != 0 ? "true"
                                                                                      : "false"
                 );
                 break;
@@ -161,12 +161,12 @@ void SchemaReader::generate_json_string() {
                 column = m_reordered_columns[column_id_index++];
                 auto name = m_global_schema_tree->get_node(column->get_id()).get_key_name();
                 m_json_serializer.append_key(name);
-                m_json_serializer.append_value_from_column_with_quotes(column, m_cur_message);
+                m_json_serializer.append_value_from_column_with_quotes(column, message_index);
                 break;
             }
             case JsonSerializer::Op::AddStringValue: {
                 column = m_reordered_columns[column_id_index++];
-                m_json_serializer.append_value_from_column_with_quotes(column, m_cur_message);
+                m_json_serializer.append_value_from_column_with_quotes(column, message_index);
                 break;
             }
             case JsonSerializer::Op::AddArrayField: {
@@ -174,7 +174,7 @@ void SchemaReader::generate_json_string() {
                 m_json_serializer.append_key(
                         m_global_schema_tree->get_node(column->get_id()).get_key_name()
                 );
-                m_json_serializer.append_value_from_column(column, m_cur_message);
+                m_json_serializer.append_value_from_column(column, message_index);
                 break;
             }
             case JsonSerializer::Op::AddNullField: {
@@ -190,6 +190,7 @@ void SchemaReader::generate_json_string() {
     }
 
     m_json_serializer.end_document();
+    return m_json_serializer.get_serialized_string();
 }
 
 bool SchemaReader::get_next_message(std::string& message) {
@@ -200,9 +201,7 @@ bool SchemaReader::get_next_message(std::string& message) {
     if (false == m_serializer_initialized) {
         initialize_serializer();
     }
-    generate_json_string();
-
-    message = m_json_serializer.get_serialized_string();
+    message = generate_json_string(m_cur_message);
 
     if (message.back() != '\n') {
         message += '\n';
@@ -223,8 +222,7 @@ bool SchemaReader::get_next_message(std::string& message, FilterClass* filter) {
             if (false == m_serializer_initialized) {
                 initialize_serializer();
             }
-            generate_json_string();
-            message = m_json_serializer.get_serialized_string();
+            message = generate_json_string(m_cur_message);
 
             if (message.back() != '\n') {
                 message += '\n';
@@ -256,8 +254,7 @@ bool SchemaReader::get_next_message_with_metadata(
             if (false == m_serializer_initialized) {
                 initialize_serializer();
             }
-            generate_json_string();
-            message = m_json_serializer.get_serialized_string();
+            message = generate_json_string(m_cur_message);
 
             if (message.back() != '\n') {
                 message += '\n';
--- a/components/core/src/clp_s/SchemaReader.hpp
+++ b/components/core/src/clp_s/SchemaReader.hpp
@@ -155,6 +155,13 @@ public:
     uint64_t get_num_messages() const { return m_num_messages; }
 
     /**
+     * Generates a JSON string from the encoded columns
+     * @param message_index The index of the message to generate the JSON string for.
+     * @return The generated JSON string
+     */
+    [[nodiscard]] auto generate_json_string(uint64_t message_index) -> std::string;
+
+    /**
      * Gets next message
      * @param message
      * @return true if there is a next message
@@ -196,6 +203,11 @@ public:
      * @param filter
      */
     void initialize_filter_with_column_map(FilterClass* filter);
+
+    /**
+     * Initializes all internal data structures required to serialize records.
+     */
+    void initialize_serializer();
 
     /**
      * Marks a column as timestamp
@@ -298,16 +310,6 @@ private:
             int32_t next_root,
             std::vector<int32_t>& path_to_intersection
     );
-
-    /**
-     * Generates a json string from the extracted values
-     */
-    void generate_json_string();
-
-    /**
-     * Initializes all internal data structured required to serialize records.
-     */
-    void initialize_serializer();
 
     int32_t m_schema_id;
     uint64_t m_num_messages;


components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -0,0 +1,27 @@
+"""Constants for CLP MCP server."""
+
+
+class CLPMcpConstants:
```
LinZhihao-723: - I don't think we need this class as a namespace anymore. The module should already play the role of the namespace.
- The correct term should be `ClpMcpConstants`.

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -0,0 +1,27 @@
...
+        "specific as possible to minimize the number of log messages returned. "
+        "You should also consider the following guidelines when generating KQL queries: "
+        "- Use specific field names and values to narrow down the search. "
+        "- Avoid using wildcards (*) unless absolutely necessary, as they can lead to "
+        "large result sets. - Use logical operators (AND, OR, NOT) to combine multiple "
+        "conditions. - Consider the time range of the logs you are searching. If the "
+        "user specifies a time range, include it in the KQL query. - If the user query "
+        "is ambiguous or lacks detail, ask clarifying questions to better understand "
+        "their intent before generating the KQL query. - Always ensure that the "
+        "generated KQL query is syntactically correct and can be executed without errors. "
```
LinZhihao-723: I will probably help polish this prompt later, but a few style-wise suggestions:
- Can u ensure the line wrap happens when 100-char overflow is triggered? The current wrapping is kind of random.
- We use uaully leave the space to the new line. So, instead of `"hello "\n"world"`, we should do `"hello"\n" world"`.
- Do u think adding `\n` to the prompt would make it more readable? Or from AI's view it will be the same.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+        The server supports multiple concurrent clients, where each client only makes synchronous
+        API calls. This assumptions leads to the following design decisions:
+        sessions is a shared variable as there may be multiple session attached to the MCP server
+        session state is NOT a shared variable because each session is accessed by only one
+        connection at a time because API calls for a single session are synchronous.
+        """
+        self._sessions_lock = threading.Lock()
+        self.sessions: dict[str, SessionState] = {}
+        self._cleanup_thread = threading.Thread(target=self._cleanup_loop, daemon=True)
+        self._cleanup_thread.start()
```
LinZhihao-723: - This docstring should go to the class' docstring section, right? Otherwise, no one will find it as it's inside the constructor as an inline comment.
- Can u justify why there are two public member variables?
- If we do need both private and public members, can u ensure they're grouped together, and public members are defined before private ones?

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -0,0 +1,27 @@
+"""Constants for CLP MCP server."""
+
+
+class CLPMcpConstants:
+    """Constants for the CLP MCP Server."""
+
+    CLEAN_UP_SECONDS = 600  # 10 minutes
```
LinZhihao-723: How about `EXPEIRED_SESSION_SWEEP_INTERVAL_SECONDS`?

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+@dataclass(frozen=True)
+class QueryResult:
+    """Cached results from the previous query's response."""
+
+    total_results: list[str]
+    items_per_page: int
+
+    _total_pages: int = field(init=False, repr=False)
+
+    def __post_init__(self) -> None:
```
LinZhihao-723: I don't fully get it. Why do we need `__post_init__`? You can do it with `__init__`?

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+        """
+        Validates that cached results don't exceed MAX_CACHED_RESULTS.
+
+        :raise: ValueError if the number of cached results or items_per_page is invalid.
+        """
+        if len(self.total_results) > CLPMcpConstants.MAX_CACHED_RESULTS:
+            err_msg = (
+                f"QueryResult exceeds maximum allowed cached results: "
+                f"{len(self.total_results)} > {CLPMcpConstants.MAX_CACHED_RESULTS}. "
```
LinZhihao-723:
```suggestion
                f"QueryResult exceeds maximum allowed cached results:"
                f" {len(self.total_results)} > {CLPMcpConstants.MAX_CACHED_RESULTS}."
```
- That's what I mean by leading space instead of trailing space.
- No trailing space in the end.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+@dataclass(frozen=True)
+class QueryResult:
+    """Cached results from the previous query's response."""
+
+    total_results: list[str]
+    items_per_page: int
+
+    _total_pages: int = field(init=False, repr=False)
```
LinZhihao-723: `_num_pages`?

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+
+from .constants import CLPMcpConstants
+
+
+@dataclass(frozen=True)
+class QueryResult:
+    """Cached results from the previous query's response."""
+
+    total_results: list[str]
+    items_per_page: int
```
LinZhihao-723: `num_items_per_page`?

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+from paginate import Page
+
+from .constants import CLPMcpConstants
+
+
+@dataclass(frozen=True)
+class QueryResult:
+    """Cached results from the previous query's response."""
+
+    total_results: list[str]
```
LinZhihao-723: What do u mean by "total" here?

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+        )
+
+    def get_page(self, page_number: int) -> Page | None:
+        """
+        Gets a specific page from the cached response.
+
+        :param page_number: One-based indexing, e.g., 1 for the first page.
+        :return: Page object or None if page number is out of bounds.
+        """
+        if page_number > self._total_pages or page_number <= 0:
```
LinZhihao-723:
```suggestion
        if 0 <= page_number or self._total_pages < page_number:
```
This should look more natural.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+            "_total_pages",
+            (len(self.total_results) + self.items_per_page - 1) // self.items_per_page,
+        )
+
+    def get_page(self, page_number: int) -> Page | None:
+        """
+        Gets a specific page from the cached response.
+
+        :param page_number: One-based indexing, e.g., 1 for the first page.
+        :return: Page object or None if page number is out of bounds.
```
LinZhihao-723:
```suggestion
        Returns a page from the cached query results.

        :param page_number: The number of page to retrieve (one-based index; 1 is the first page).
        :return: A `Page` object for the specified page.
        :return: None `page_number` is out of bounds.
```

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+from typing import Any
+
+from paginate import Page
+
+from .constants import CLPMcpConstants
+
+
+@dataclass(frozen=True)
+class QueryResult:
+    """Cached results from the previous query's response."""
```
LinZhihao-723: Can u clarify what you mean by "the previous query's response"? I don't see where "the previous" comes from.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+        return Page(
+            self.total_results,
+            page=page_number,
+            items_per_page=self.items_per_page,
+        )
+
+
+@dataclass
+class SessionState:
+    """State of a single user session."""
```
LinZhihao-723:
```suggestion
    """Represents the state of a user session."""
```
Don't think we need "single" here.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+
+    session_id: str
+    items_per_page: int
```
LinZhihao-723: `num_items_per_page`?

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+    session_ttl_minutes: int
+    last_accessed: datetime = field(default_factory=lambda: datetime.now(timezone.utc))
+    cached_query_result: QueryResult | None = None
+    ran_instructions: bool = False
```
LinZhihao-723: How about `instructions_retrieved`?

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+        :param page_number: One-based indexing, e.g., 1 for the first page.
+        :return: On success, dictionary containing paged log entries and the paging metadata.
+        On error, dictionary with ``{"Error": "error message describing the failure"}``.
+        """
+        if self.cached_query_result is None:
+            return {"Error": "No previous paginated response in this session."}
+
+        page = self.cached_query_result.get_page(page_number)
+        if page is None:
+            return {"Error": "Page index is out of bounds."}
```
LinZhihao-723, line 135:
```suggestion
            return {"Error": "Page number is out of bounds."}
```
Shall we be consistent since we use "number" but not "index" for page access?

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+            "items": list(page),
+            "total_pages": page.page_count,
+            "total_items": page.item_count,
+            "items_per_page": page.items_per_page,
+            "has_next": page.next_page is not None,
+            "has_previous": page.previous_page is not None,
+        }
+
+    def is_expired(self) -> bool:
+        """:return: whether the session has expired."""
```
LinZhihao-723:
```suggestion
        """:return: Whether the session has expired."""
```

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,221 @@
...
+            total_results=results, items_per_page=self.items_per_page
+        )
+
+    def get_page_data(self, page_number: int) -> dict[str, Any]:
+        """
+        Gets page data and its metadata in a dictionary format.
+
+        :param page_number: One-based indexing, e.g., 1 for the first page.
+        :return: On success, dictionary containing paged log entries and the paging metadata.
+        On error, dictionary with ``{"Error": "error message describing the failure"}``.
```
LinZhihao-723: note to myself: rewrite this docstring.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: By index, I am referring to zero-based indexing, e.g. 0 means the first page. By number, I am referring to one-based indexing, e.g 1 means the first page. The reason why we have two indexing mechanisms is because the third-party library (Paginate) uses one-based indexing while our pre-defined documentaion assumes zero-based indexing.

In our documentation, we define AI to use zero-based index for indexing get_nth_page. The above error message is written to respect our documentation.

20001020ycx: In fact, I see where your concern comes from, this is at SessionState.get_page_data where the page_index is already converted to page_number. 

That said, the value I’m returning here is the actual error message defined in our documentation. This is the error message that will return back to the AI,  and for the sake of clarity on AI side, I think we should keep the error message as is.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: for bool, is it better with `is_ instructions_retrieved`

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: my oversight, this is the full response we get from the previous query, in the class QueryResult, we will paginate the result and feed back to AI rather than simply returning the full result.

As said, this should be updated with `cached_response`

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: see below comment

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: I pushed a version following your convention in `get_page` of QueryResult, let me know if there's any problem.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: Below is from the page note from the documentation of the get_nth_page:
> Simply checking the previous query’s response should be fine for our first release where we will only use synchronous tool calls. But once we decide to add support for asynchronous tool calls, we should augment this tool to take a query ID as an argument as well.

In another word, say the client issues three API calls: substring_search 1, substring_search 2, get_nth_page(2). get_nth_page only caches the result for substring_search 2 because synchronous call ensures it is the previous response. However, when we are talking about async mode, the concept of previous is hard to get, so we may use a query ID in the future. For now, AI team has no plan for async mode within a client (note, multi-user are async in the current implementation already), this is because we didn't see the benefit, at least on the AI diagnosis side of issuing async API tool calls

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
LinZhihao-723: If you just look at this single object, it doesn't have to know it's a cache entry. I think:
- How about renaming the class to `QueryResultPaginator`, which matches what u state.
- You can drop "total" then, and probably name it to `result_log_entries`. This is more explicit about what's stored inside this list (which tbh isn't clear from what's presented in the existing code).

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
LinZhihao-723: I get what u mean. However, this class, `QueryResult`, doesn't enforce anything to be exactly "the previous". It is the session manager and the state's responsibility to maintain what u said. That's where my confusion comes from, and I don't think u should mention "the previous".
Thinking this way, you can still maintain multiple query results using this class, right? It's just the session manager's decision to only keep the last one.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
LinZhihao-723: Sure, make sense.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
LinZhihao-723: Sure, I get what u mean. Then I think it makes more sense to take the page index as the parameter then. Moving the page number conversion inside this function would lead to less confusion, as it also shows what an index is. What do you think?

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
LinZhihao-723: If my understanding was correct in the other comment, I believe `_num_pages` is the proper name to use. We usually use `num_xxx` to indicate the total number of xxx.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: wound't this revert the logic?

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: agreed, I will move the logic

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
LinZhihao-723: no my bad. It should be `page_numer <= 0` coming first.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: sorry, I misunderstood the comment, agreed

components/clp-mcp-server/clp_mcp_server/server/constants.py
20001020ycx: good to know about this!

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: This is actually a question to you. Functionality-wise, there's nothing limits us from using the _init_. So I wonder what is the most pythonic way of doing this?
I declare @dataclass(frozen=True) so __init__ is handled by python. Still, I need some code for init, that's why I throw it in the post_init. I learned/copied this from [integration test](https://github.com/y-scope/clp/blob/92ef5e122c9303d10879a65a0e52eb7833c9f39b/integration-tests/tests/utils/config.py)

LinZhihao-723: - I don't think using `dataclass` is absolutely required.
- The current implementation of using `__setattr__` is like a hack. `__post_init__` is supposed to be only for validation but not for creating a new member variable at runtime.

I'd say u should use `__init__` to make a normal class in this case.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: I see, agreed

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: adapted the second point. For the first point, how about `PaginatedQueryResult`, QueryResultPaginator sounds like a utility tool for pagination (the actual library I am exporting - Paginate) rather the class where we actually store the paginated result.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: - Right, my oversight, pushed with the suggested migration. But should I leave the explanation of why sessions is a shared variable while session state is not in the same place?
- `sessions` is made public for testing purposes, but `session_ttl_minutes` is not needed, will make it private. However, do you think if its a better practice to still make it private, but adding a `@property` annotation to make it readonly for tester? Or simplying adding an explicit accessor (I only have get_or_create, no get). Let me know what is the best practice, I will leave it to public for now.
- Done

components/clp-mcp-server/clp_mcp_server/server/constants.py
20001020ycx: - sure, I will update to make it more readable
- I think I get your point, submitted a version with all format fixed.

But I would say don't spend too much time on polishing the prompt, we will have a separate PR just to prune the documentation of the AI project once we have all the features in, see future work in the description.

junhaoliao: shall we use triple-quoted string instead to make this more readable to human? e.g.,
```py
SYSTEM_PROMPT = """You are an AI assistant that helps users query a log database using KQL (Kibana Query Language). 
When given a user query, you should generate a KQL query that accurately captures the user's intent. 
The KQL query should be as specific as possible to minimize the number of log messages returned.

You should also consider the following guidelines when generating KQL queries:
- Use specific field names and values to narrow down the search.
- Avoid using wildcards (*) unless absolutely necessary, as they can lead to large result sets.
- Use logical operators (AND, OR, NOT) to combine multiple conditions.
- Consider the time range of the logs you are searching. If the user specifies a time range, include it in the KQL query.
- If the user query is ambiguous or lacks detail, ask clarifying questions to better understand their intent before generating the KQL query.
- Always ensure that the generated KQL query is syntactically correct and can be executed without errors.
"""
```

components/clp-mcp-server/tests/test_session_manager.py
```diff
@@ -0,0 +1,260 @@
...
+        session.last_accessed = datetime.now(timezone.utc) - timedelta(minutes=61)
+        assert session.is_expired() is True
+
+
+class TestSessionManager:
+    """Unit tests for SessionManager class."""
+
+    def test_get_or_create_session(self) -> None:
+        """Validates session creation and retrieval."""
+        manager = SessionManager(session_ttl_minutes=TestConstants.SESSION_TTL_MINUTES)
```
junhaoliao: This seems repeated in many places. Have we considered creating a "fixture" to avoid duplicates? e.g.,
```py
@pytest.fixture
def active_session_manager() -> SessionManager:
    manager = SessionManager(session_ttl_minutes=TestConstants.SESSION_TTL_MINUTES)
    session = manager.get_or_create_session("test_session")
    session.is_instructions_retrieved = True
    return manager
```

then we can use the manager directly: 
```py
class TestSessionManager:
    def test_get_or_create_session(self, manager: SessionManager) -> None:
```

components/clp-mcp-server/tests/test_session_manager.py
```diff
@@ -0,0 +1,260 @@
...
+        session = SessionState(
+            session_id="test_session",
+            _num_items_per_page=TestConstants.ITEMS_PER_PAGE,
+            _session_ttl_minutes=TestConstants.SESSION_TTL_MINUTES
+        )
+
+        assert session.is_expired() is False
+
+        # Simulate old session
+        session.last_accessed = datetime.now(timezone.utc) - timedelta(minutes=61)
```
junhaoliao: why we don't use `SESSION_TTL_MINUTES`? if `61` has to be used - 

can you add extract the magic number into a constant and/or explain why `61` is used for those who are poor at math (like me)?

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,223 @@
...
+        first page can be retrieved.
+        :return: Dictionary with ``{"Error": "error message describing the failure"}`` if fails to
+        retrieve the first page.
+        """
+        session = self.get_or_create_session(session_id)
+        if session.is_instructions_retrieved is False:
+            return {
+                "Error": "Please call get_instructions() first "
+                "to understand how to use this MCP server."
+            }
```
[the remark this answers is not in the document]
junhaoliao: I agree with the rabbit

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,223 @@
...
+    def cache_query_result(
+        self,
+        results: list[str],
+    ) -> None:
+        """
+        Caches the latest query result of the session.
+
+        :param results: Complete log entries from previous query for caching.
+        """
+        self._cached_query_result = PaginatedQueryResult(
```
junhaoliao, line 92: Is per session concurrency possible? If so, do we have a potential data race here?

junhaoliao: If not, i think it's worth documenting this method / class is not thread-safe since it is used in a class `SessionManager` that claims to have concurrency support

junhaoliao: I might have misunderstood how the state objects work, but -

is it possible create an `SessionState` instance only when a `PaginatedQueryResult` is available? so we can get rid of the nullable property of `_cached_query_result`

components/clp-mcp-server/clp_mcp_server/server/constants.py
20001020ycx: I see, right, I will make the update

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: I see, done

components/clp-mcp-server/tests/test_session_manager.py
20001020ycx: This is for expiring some sessions for testing. I will add the magic number, good catch

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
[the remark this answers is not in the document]
20001020ycx: done!

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: Below is the concurrency model we assume for the session manager, this is documented in the class definition of SessionManager:
> Session manager for concurrent user sessions.
> `SessionManager` respects MCP Server Concurrency Model, that is:
> The server supports multiple concurrent clients, where each client only makes synchronous
> API calls.
> This model leads to the following design decisions:
> sessions is a shared variable as there may be multiple session attached to the MCP server
> session state is NOT a shared variable because each session is accessed by only one
> connection at a time because API calls for a single session are synchronous.

To answer your question, per session concurrency is possible if a user makes async call to our API suite. However, we dont see this behaviour in the market agent such as Claude and vscode. Therefore, for this version, we are assuming user won't make any async call.

<s>I also brought this up in today's stand up, after this pr, we will be adding a sequential queue to handle concurrent access per user.<s>

20001020ycx: For your second point, I fully agree and will add the explanation of the potential concurrency problem on SessionState object since everyone is asking this.

20001020ycx: As for the third point, I think the answer is no, we need a session for get_instructions, but we don't cache any response for this API at all. So a session can be created even without a cached response

components/clp-mcp-server/tests/test_session_manager.py
20001020ycx: Good suggestion, done!

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: After syncing up with Rishi, we decide in this pr, we will respect the above assumption. But in addition to the soft assumption, we will enforce a locking on the session state to make it thread safe. 

In the follow-up pr, we will provide the actual async support, where get_nth_page is indexed by an additional atomic counter to distinguish which API call's response the AI wishes to read

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,232 @@
...
+        """Updates the last accessed timestamp."""
+        self.last_accessed = datetime.now(timezone.utc)
+
+
+class SessionManager:
+    """
+    Session manager for concurrent user sessions.
+    `SessionManager` respects `FastMCP` asynchronous concurrency model that is built on Python's
+    asyncio runtime:
+    Asyncio achieves concurrency on a single thread by allowing tasks to yield control at `await`
```
20001020ycx: The explanation of the async concurrency within a single thread is duplicated in both SessionManger and SessionState. This does not seems to be a good practice. Maybe we can have a referral, I wonder what is the best practice? Should we move this to server.py where the mcp server is created?

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -0,0 +1,26 @@
...
+- Use specific field names and values to narrow down the search.
+- Avoid using wildcards (*) unless absolutely necessary, as they can lead to large result
+ sets.
+- Use logical operators (AND, OR, NOT) to combine multiple conditions.
+- Consider the time range of the logs you are searching. If the user specifies a time range,
+ include it in the KQL query.
+- If the user query is ambiguous or lacks detail, ask clarifying questions to better
+ understand their intent before generating the KQL query.
+- Always ensure that the generated KQL query is syntactically correct and can be executed
+ without errors.
```
LinZhihao-723: Are u sure this would work? The raw string will treat a newline as a line break, so a prompt sentence will be split into multiple lines.

components/clp-mcp-server/clp_mcp_server/server/server.py
```diff
@@ -31,22 +16,39 @@ def create_mcp_server() -> FastMCP:
...
+    @mcp.tool
+    def get_instructions(ctx: Context) -> str:
         """
-        Gets the MCP server's information.
+        Gets a pre-defined “system prompt” that guides the LLM behavior.
+        This function must be invoked before any other `FastMCP.tool`.
 
-        :return: The server's information with a list of capabilities.
+        :param ctx: The `FastMCP` context containing the metadata of the underlying MCP session.
+        :return: A string of “system prompt”.
```
LinZhihao-723: When u do copy-paste, pay attention to `"` as the source may use `“`.

components/clp-mcp-server/clp_mcp_server/server/server.py
```diff
@@ -31,22 +16,39 @@ def create_mcp_server() -> FastMCP:
...
+        """
+        Retrieves the n-th page of a paginated response with the paging metadata from the previous
+        query.
+
+        :param page_index: Zero-based index, e.g., 0 for the first page.
+        :param ctx: The `FastMCP` context containing the metadata of the underlying MCP session.
+        :return: Dictionary containing paged log entries and the paging metadata if the
+        page `page_index` can be retrieved.
+        :return: Dictionary with ``{"Error": "error message describing the failure"}`` if fails to
+        retrieve page `page_index`.
```
LinZhihao-723: I will rewrite some of the docstrings. However, you should indent multi-line return/param statements:
```suggestion
            retrieve page `page_index`.
```

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,232 @@
...
+        Initializes the PaginatedQueryResult.
+
+        :param result_log_entries: List of cached log entries to paginate.
+        :param num_items_per_page:
+        :raise: ValueError if the number of cached results or num_items_per_page is invalid.
+        """
+        if len(result_log_entries) > constants.MAX_CACHED_RESULTS:
+            err_msg = (
+                f"PaginatedQueryResult exceeds maximum allowed cached results:"
+                f" {len(result_log_entries)} > {constants.MAX_CACHED_RESULTS}."
```
LinZhihao-723:
```suggestion
                "PaginatedQueryResult exceeds maximum allowed cached results:"
                f" {len(result_log_entries)} > {constants.MAX_CACHED_RESULTS}."
```

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,232 @@
...
+            )
+            raise ValueError(err_msg)
+
+        if num_items_per_page <= 0:
+            err_msg = (
+                f"Invalid num_items_per_page: {num_items_per_page}, "
+                "it must be a positive integer. "
```
LinZhihao-723:
```suggestion
                f"Invalid num_items_per_page: {num_items_per_page}, it must be a positive integer."
```
Pay attention to these details:
- When wrapping new line, space should be in the next line.
- Trailing space should be removed.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def get_page(self, page_index: int) -> Page | None:
+        """
+        Returns a page from the cached query results.
+
+        :param page_index: The number of page to retrieve (zero-based index; 0 is the first page).
+        :return: A `Page` object for the specified page.
+        :return: None if `page_index` is out of bounds.
+        """
+        page_number = page_index + 1  # Convert zero-based to one-based
```
LinZhihao-723:
```suggestion
        # Convert zero-based to one-based
        page_number = page_index + 1
```
We prefer to put the inline comment in a separate line.

LinZhihao-723 (changes requested):
Let me know if u plan to rewrite the garbage collector using async.

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -0,0 +1,26 @@
...
+"""
```
LinZhihao-723: Refactored in the following way:

```suggestion
SYSTEM_PROMPT = (
"You are an AI assistant that helps users query a log database using KQL (Kibana Query Language)."
" You should generate a KQL query that accurately expresses the user's intent. The generated KQL"
" query should be as specific as possible to minimize the number of log messages returned.\n\n"
"You should consider the following guidelines to generate KQL queries efficiently:\n"
"- Use specific field names and values to narrow down the search.\n"
"- Avoid using wildcards (`*`) unless absolutely necessary, as they can lead to large result"
" sets.\n"
"- Use logical operators (`AND`, `OR`, `NOT`) to combine one or more key-value searches.\n"
"- Consider the time range of the logs you are searching. If the user specifies a time range,"
" include it in the KQL query.\n"
"- If the user query is ambiguous or lacks detail, ask clarifying questions to better understand"
" their intent before generating the KQL query.\n"
"- Always ensure that the generated KQL query is syntactically correct and can be executed without"
" errors."
)
```

A few questions about this prompt, may also need inputs from @kirkrodrigues and @junhaoliao.
- Since our KQL is slightly different from the standard version, shall we mention the syntax first? For example:
  - Use double quotes for string values.
  - Escape special characters.
- We didn't mention auto-generated keys. Would this make the agent only query user-generated keys, which may miss some important log-event-level metadata?
- We didn't mention file-level metadata query.
- When we say narrow down the timestamp, we didn't give rules to:
  - In what granularity should the timestamp be queried? We should maybe explain the timsetamp is Unix Epoch in milliseconds (or nanoseconds with Devin's latest change).
  - How to deal with the timezone? We don't have timezone information tracked in the archive now. Without some grace offsets applied, the agent could generate a query that doesn't fall into the actual expected timestamp range.

LinZhihao-723 (changes requested):

components/clp-mcp-server/clp_mcp_server/server/constants.py
kirkrodrigues: Agree with everything you mentioned except:

* Afaik, we aren't currently storing file-level metadata in clp-json (although clp-s supports storing it), so perhaps we should defer this.
* We can tell the LLM that we only want queries in UTC time, right?

Some additional points:

* I think it might also be useful to tell the LLM that we don't support fuzzy matches (otherwise it might expect that, for example, "system error" can match "system shutdown error").
* How does the LLM know what key to use for the timestamp?
* Do we really need the point about ensuring the KQL query is syntactically correct? Does it actually improve the accuracy of the generated query?

LinZhihao-723: iirc, we don't support parsing timestamps as UTC yet. When users give a timestamp range, LLM probably also doesn't know which timezone it should use to apply the offset.
I think a safe range should be: suppose `lower_bound` and `upper_bound` are given, we add a grace padding to search:
- `real_lower_bound` = `lower_bound` - 12hours
- `real_upper_bound` = `upper_bound` + 14hours

But this might make the range too wild. On the other hand, I'm not sure if AI can understand how to properly apply the offset if we give it a concrete timezone.

20001020ycx: Here's my opinion to each point you have mentioned:

- agreed, I wasn't aware of this, is there any documentation defining what specific KQL we are supporting?
- I don't have any background for auto-generated key? Are they just some other fields in the returned response? In this case, I think LLM may learn from the returned result right?
- In addition to what Kirk said about file-level metadata, the current suite of API does not take in any file-level metadata (e.g file name), so information about the file is not in the interest of the current release.
- I am having trouble understanding why timezone is important here. Say agent issues substring_seach ("error"), and in the receiving response, each log message is accompanied by a timestamp, say Toronto time or Unix epoch. LLM would learn what is the format here, and issue the next substring search with a timestamp that is around the point of interest. For example, if the timestamp is Unix epoch, and the error message that the LLM is interested in has a timestamp of Unix epoch 1234, then the next request LLM would issue is substring_search("some other related keyword", 1000, 1500), because it realizes a timestamp is Unix epoch rather a UTC timestamp.

kirkrodrigues: * https://docs.yscope.com/clp/main/user-docs/reference-json-search-syntax.html
* [Background](https://docs.yscope.com/clp/main/dev-docs/design-kv-ir-streams/auto-gen-kv-pairs). Auto-generated keys won't be returned in the response but they are query-able. Thinking about it, since we don't know what autogenerated keys the compressed logs will have, at this point, we probably can't say anything in the API. In the future, perhaps we can have better support.
* Regarding timezones, @LinZhihao-723 can you provide an example log input and output that demonstrates the issue? I'm a little unclear as well.

components/clp-mcp-server/clp_mcp_server/server/server.py
```diff
@@ -31,22 +16,39 @@ def create_mcp_server() -> FastMCP:
...
+        """
+        Retrieves the n-th page of a paginated response with the paging metadata from the previous
+        query.
+
+        :param page_index: Zero-based index, e.g., 0 for the first page.
+        :param ctx: The `FastMCP` context containing the metadata of the underlying MCP session.
+        :return: Dictionary containing paged log entries and the paging metadata if the
+            page `page_index` can be retrieved.
+        :return: Dictionary with ``{"Error": "error message describing the failure"}`` if fails to
+            retrieve page `page_index`.
```
LinZhihao-723: As this will be directly read by AI, shall we make it more concrete like:
```suggestion
        Retrieves the n-th page of a paginated response with the paging metadata from the previous
        query.

        :param page_index: Zero-based index, e.g., 0 for the first page.
        :param ctx: The `FastMCP` context containing the metadata of the underlying MCP session.
        :return: A dictionary containing the following key-value pairs on success:
            - "item": A list of log entries in the requested page.
            - "total_pages": Total number of pages available from the query as an integer.
            - "total_items": Total number of log entries available from the query as an integer.
            - "num_items_per_page": Number of log entries per page.
            - "has_next": Whether a page exists after the returned one.
            - "has_previous": Whether a page exists before the returned one.
        :return: A dictionary with the following key-value pair on failures:
            - "Error": An error message describing the failure.
```
This is assuming that AI will only read this function. If, AI is able to read the docstring of the actual function called, we should move this docstring to the implementation and make this docstring sth like "forwards xxx's return values."

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,232 @@
...
+from . import constants
+
+
+class PaginatedQueryResult:
+    """Paginates the cached log entries returned from a query's response."""
+
+    def __init__(self, result_log_entries: list[str], num_items_per_page: int) -> None:
+        """
+        Initializes the PaginatedQueryResult.
+
```
LinZhihao-723:
```suggestion
```

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,232 @@
...
+        self.result_log_entries = result_log_entries
+        self.num_items_per_page = num_items_per_page
+
+        self._num_pages = (len(result_log_entries) + num_items_per_page - 1) // num_items_per_page
+
+    def get_page(self, page_index: int) -> Page | None:
+        """
+        Returns a page from the cached query results.
+
```
LinZhihao-723:
```suggestion
```

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,232 @@
...
+    def get_page_data(self, page_index: int) -> dict[str, Any]:
+        """
+        Gets page data and its metadata in a dictionary format.
+
+        :param page_index: The number of page to retrieve (zero-based index; 0 is the first page).
+        :return: Dictionary containing paged log entries and the paging metadata if the
+            page `page_index` can be retrieved.
+        :return: Dictionary with ``{"Error": "error message describing the failure"}`` if fails to
+            retrieve page `page_index`.
```
LinZhihao-723:
```suggestion
        Retrieves the n-th page of a paginated response with the paging metadata from the previous
        query.

        NOTE: This docstring must be synchronized with `get_nth_page`'s MCP tool call.

        :param page_index: Zero-based index, e.g., 0 for the first page.
        :param ctx: The `FastMCP` context containing the metadata of the underlying MCP session.
        :return: A dictionary containing the following key-value pairs on success:
            - "item": A list of log entries in the requested page.
            - "total_pages": Total number of pages available from the query as an integer.
            - "total_items": Total number of log entries available from the query as an integer.
            - "num_items_per_page": Number of log entries per page.
            - "has_next": Whether a page exists after the returned one.
            - "has_previous": Whether a page exists before the returned one.
        :return: A dictionary with the following key-value pair on failures:
            - "Error": An error message describing the failure.
```

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,232 @@
...
+
+        return session.get_page_data(0)
+
+    def get_nth_page(self, session_id: str, page_index: int) -> dict[str, Any]:
+        """
+        Retrieves the n-th page of a paginated response with the paging metadata from the previous
+        query.
+
+        :param session_id: Unique identifier for the session.
+        :param page_index: The number of page to retrieve (zero-based index; 0 is the first page).
```
LinZhihao-723:
```suggestion
        :param session_id:
        :param page_index: Index of the page to retrieve (zero-based, e.g., 0 is the first page).
```

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,232 @@
...
+        :return: Forwards `SessionState.get_page_data`'s return values.
+        """
+        session = self.get_or_create_session(session_id)
+        if session.is_instructions_retrieved is False:
+            return self._GET_INSTRUCTIONS_NOT_RUN_ERROR.copy()
```
LinZhihao-723: It makes more sense to move this into the `SessionState`'s methods, right?

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,232 @@
...
+
+            for sid in expired_sessions:
+                del self.sessions[sid]
+
+    def get_or_create_session(self, session_id: str) -> SessionState:
+        """
+        Gets an existing session or creates a new one.
+
+        :param session_id: Unique identifier for the session.
+        :return: The SessionState object for the given session_id.
```
LinZhihao-723:
```suggestion
        :param session_id:
        :return: The `SessionState` object for the given `session_id`.
```

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,232 @@
...
+        :param session_id: Unique identifier for the session.
+        :param query_results: Complete log entries from previous query for caching.
+        :return: Dictionary containing the first page log entries and the paging metadata if the
+            first page can be retrieved.
+        :return: Dictionary with ``{"Error": "error message describing the failure"}`` if fails to
+            retrieve the first page.
+        """
+        session = self.get_or_create_session(session_id)
+        if session.is_instructions_retrieved is False:
+            return self._GET_INSTRUCTIONS_NOT_RUN_ERROR.copy()
```
LinZhihao-723: Same.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,232 @@
...
+    def cache_query_result(self, session_id: str, query_results: list[str]) -> dict[str, Any]:
+        """
+        Caches query results for a session and returns the first page and the paging metadata.
+
+        :param session_id: Unique identifier for the session.
+        :param query_results: Complete log entries from previous query for caching.
+        :return: Dictionary containing the first page log entries and the paging metadata if the
+            first page can be retrieved.
+        :return: Dictionary with ``{"Error": "error message describing the failure"}`` if fails to
+            retrieve the first page.
```
LinZhihao-723:
```suggestion
        :param session_id:
        :param query_results: Log entries from the query to cache.
        :return: Forwards `SessionState.get_page_data`'s return values.
```

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,232 @@
...
+            )
+            raise ValueError(err_msg)
+
+        if num_items_per_page <= 0:
+            err_msg = (
+                f"Invalid num_items_per_page: {num_items_per_page}, it must be a positive integer."
+            )
+            raise ValueError(err_msg)
+
+        self.result_log_entries = result_log_entries
```
LinZhihao-723: Make it private?

LinZhihao-723 (changes requested):
Haven't looked into the class-wise docstring. Will do a pass once ur async changes are pushed.

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,229 @@
...
+    Session manager for concurrent user sessions.
+    `SessionManager` respects `FastMCP` asynchronous concurrency model that is built on Python's
+    asyncio runtime:
+    Asyncio achieves concurrency on a single thread by allowing tasks to yield control at `await`
+    points. This means when a `mcp.tool` API calls executes an await expression, it gets suspended,
+    and the event loop executes the next `mcp.tool` API calls from received requests issued by
+    clients.
+    The operations on `sessions` which maintains concurrent user sessions are performed by
+    asynchronous tasks. The above concurrency model guarantees that the operations of `sessions`
+    are atomic because it cannot be accessed from two threads at the same time.
```
LinZhihao-723:
```suggestion
    Manages concurrent user sessions, following the same concurrency model as `FastMCP`.

    All data structures managed by this class are accessed exclusively by coroutines running
    within the same event loop in a single-thread context. This design implies:

    - No thread-level concurrency control (e.g., `threading.Lock`) is necessary, since operations
      occur in a single-threaded async context.
    - Async synchronization primitives (e.g., `asyncio.Lock`) are only necessary when an `await`
      occurs inside a critical section, to prevent interleaving of coroutine execution.
```

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,229 @@
...
+    """
+    Represents the state of a user session.
+    `SessionState` respects `FastMCP` asynchronous concurrency model that is built on Python's
+    asyncio runtime:
+    Asyncio achieves concurrency on a single thread by allowing tasks to yield control at `await`
+    points. This means when a `mcp.tool` API calls executes an await expression, it gets suspended,
+    and the event loop executes the next `mcp.tool` API calls from received requests issued by
+    clients.
+    Therefore, `SessionState` does not need thread-safe operations because a session cannot be
+    accessed from two threads at the same time.
```
LinZhihao-723:
```suggestion
    Represents the state of a user session, following the same concurrency model as
    `SessionManager`.

    NOTE: All methods are intended to be executed exclusively by coroutines within the same event
    loop in a single-threaded context.
```

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,229 @@
...
+    session_id: str
+    _num_items_per_page: int
+    _session_ttl_minutes: int
+
+    is_instructions_retrieved: bool = False
+    last_accessed: datetime = field(default_factory=lambda: datetime.now(timezone.utc))
+    _cached_query_result: PaginatedQueryResult | None = None
+
+    _GET_INSTRUCTIONS_NOT_RUN_ERROR: ClassVar[dict[str, str]] = {
+        "Error": "Please call get_instructions() first to understand how to use this MCP server."
```
LinZhihao-723:
```suggestion
        "Error": "Please call `get_instructions()` first to understand how to use this MCP server."
```

components/clp-mcp-server/README.md
LinZhihao-723, line 1: Added an instruction for test case.
@coerabbitai can u create an issue as a feature request to add tasks and workflows to run unit tests for all uv components?

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -0,0 +1,237 @@
...
+      occur in a single-threaded async context.
+    - Async synchronization primitives (e.g., `asyncio.Lock`) are only necessary when an `await`
+      occurs inside a critical section, to prevent interleaving of coroutine execution.
+    """
+
+    def __init__(self, session_ttl_seconds: int) -> None:
+        """:param session_ttl_seconds: Session time-to-live in seconds."""
+        self.sessions: dict[str, SessionState] = {}
+        self._session_ttl_seconds = session_ttl_seconds
+        self._cleanup_task: asyncio.Task | None = None
```
LinZhihao-723:
```suggestion
    def __init__(self, session_ttl_seconds: float) -> None:
        """:param session_ttl_seconds: Session time-to-live in seconds."""
        self.sessions: dict[str, SessionState] = {}
        self._session_ttl_seconds: float = session_ttl_seconds
        self._cleanup_task: asyncio.Task | None = None
```

components/clp-mcp-server/README.md
LinZhihao-723: @coderabbitai can u create an issue for us?

components/clp-mcp-server/clp_mcp_server/server/constants.py
LinZhihao-723: As discussed offline, we will leave the system prompt improvements for next PRs. @coderabbitai can u create two issues:
- The first one to keep track of updating this system prompt to cover the issues above.
- The second one to ask us to update CLP doc to state that the current CLP is timezoneless. Users should use timezoneless timestamp in queries.

PR #1404, opened 2025-10-09
feat(clp-package): Use CSafeLoader for faster YAML parsing in config loading (resolves #1402).

### LINKED ISSUE #1402: Improve YAML loading performance by using `CSafeLoader` instead of `safe_load`
### Request

(first reported by @Eden-D-Zhang )

While profiling YAML parsing inside the clp-package image, it was found that `yaml.safe_load()` is significantly slower than `yaml.load(..., Loader=yaml.CSafeLoader)`.

#### Test setup

Both methods were tested using the same `clp-config.yml` file within the clp-package container.
The Python environment is located at `/opt/clp/lib/python3/site-packages`.

#### Test commands

```sh
# CSafeLoader version
cat clp-config.yml | docker run -i --rm "$(docker images --format '{{.Repository}}:{{.Tag}}' clp-package | head -n1)" \
bash -c 'PYTHONPATH=/opt/clp/lib/python3/site-packages python3 -c "
import sys, time, yaml
s = sys.stdin.read()
t0 = time.perf_counter()
yaml.load(s, Loader=yaml.CSafeLoader)
t1 = time.perf_counter()
print(f\"Elapsed time (CSafeLoader): {t1 - t0:.5f}s\")"'

# safe_load version
cat clp-config.yml | docker run -i --rm "$(docker images --format '{{.Repository}}:{{.Tag}}' clp-package | head -n1)" \
bash -c 'PYTHONPATH=/opt/clp/lib/python3/site-packages python3 -c "
import sys, time, yaml
s = sys.stdin.read()
t0 = time.perf_counter()
yaml.safe_load(s)
t1 = time.perf_counter()
print(f\"Elapsed time (safe_load): {t1 - t0:.5f}s\")"'
```

#### Results

```
Elapsed time (CSafeLoader): 0.00152s
Elapsed time (safe_load):    0.00395s
```

CSafeLoader is roughly 2.6× faster in this test, with identical functional output.

### Possible implementation

We should switch YAML loading in our Python code (where PyYAML is used for config parsing) to use `yaml.CSafeLoader` when available, and fall back to `yaml.SafeLoader` otherwise.

A simple safe pattern is:

```py
import yaml

Loader = getattr(yaml, "CSafeLoader", yaml.SafeLoader)
with open("clp-config.yml", "r", encoding="utf-8") as f:
    config = yaml.load(f, Loader=Loader)
```

This preserves compatibility for environments where libyaml is not compiled while yielding a measurable performance improvement when it is.

junhaoliao:
# Description

As the benchmark in #1402 shows, the CSafeLoader is much faster than the default SafeLoader. This PR adds the logic to use CSafeLoader whenever available to speed up config parsing.

# Checklist

  breaking change.

# Validation performed

1. Built the package.
2. Uncommented the contents in `build/clp-package/etc/clp-config.yml`.
3. Started the package: `sbin/start-clp.sh`
4. Compressed sample logs: `sbin/compress.sh ~/samples/hive-24hr`
5. Performed a query in the webui and observed results listed.

## Summary by CodeRabbit

- Refactor
  - Improved YAML parsing for configuration and credential files by switching to a higher-performance loader, enhancing load speed and robustness without changing behaviour.
- Chores
  - Standardized loader selection across the application for consistent, explicit parsing.
  - No user action required; existing configurations and credentials continue to work as before.


components/clp-py-utils/clp_py_utils/core.py
```diff
@@ -3,6 +3,9 @@
 import yaml
 from yaml.parser import ParserError
 
+# Use `CSafeLoader` when available as it offers faster C-based parsing.
+YAML_LOADER = getattr(yaml, "CSafeLoader", yaml.SafeLoader)
```
junhaoliao: We can directly use `yaml.load(yaml_config_file, Loader=yaml.CSafeLoader)` below if we don't need to mind compatibility 
* outside the `clp-package` container
* or when the script is executed without the CLP package's `site-package`

junhaoliao: now i look at it again, i feel we can always assume the `CSafeLoader` to be available because the Python scripts are not supposed to be run elsewhere

Eden-D-Zhang (commented):
Changes look good provided it's guaranteed that `CSafeLoader` is available. Just for my understanding, could you show me where we include it?

junhaoliao:
> Changes look good provided it's guaranteed that `CSafeLoader` is available. Just for my understanding, could you show me where we include it?

sure. please see below for the explanation:

1. Why the C Extension (`libyaml`) is "optional":
   https://github.com/yaml/pyyaml/blob/d51d8a138f7230834fc6e95635ff09ebd329185f/setup.py#L350-L354
	* PyYAML optionally builds a C extension `_yaml` (used by `CSafeLoader` / `CSafeDumper`) for faster performance.
	* The extension links against the system `libyaml` (`libraries=['yaml']`).
	* A small C test (`LIBYAML_CHECK`) verifies that `<yaml.h>` and `libyaml` exist.
	* If it passes, `_yaml` is compiled; otherwise, PyYAML falls back to pure-Python.
	* No `libyaml` is bundled - it must come from system packages like `libyaml-dev`.

2. How we are using prebuilt libraries from wheels via `pip install`

	* When installing on Ubuntu Jammy which is our development base env (or most major platforms), pip downloads a "prebuilt wheel" (Python package bundle. e.g. `pyyaml-6.0.3-cp310-manylinux2014_x86_64.whl`) when we do `pip install pyyaml`
	* That wheel already contains the compiled `_yaml.cpython-310-x86_64-linux-gnu.so`, so no build step or `setup.py` logic runs.
	* You can confirm with:
	  ```
	  pip download pyyaml --only-binary=:all:
	  unzip pyyaml-*.whl
	  ls yaml/_yaml*.so
	  ```

Eden-D-Zhang (approved):

PR #1407, opened 2025-10-12
feat(job-orchestration): Allow compression jobs to specify S3 object keys via `S3InputConfig`.

LinZhihao-723:
# Description

NOTE: This is not a breaking change so any existing S3 compression should still work.

This was originally a part of #1383. However, we decided to remove the compression interface change and make a clean PR for updating the job input config.

This PR updates `S3InputConfig` to add support for accepting a list of S3 object keys. Before this PR, the input config can only take a key prefix and ingest all logs under this prefix. With this PR, we make it more flexible so that users can define a list of keys under the given prefix. Notice that these keys have the following constraints:

- All the keys must share the same key-prefix, as specified in the base `S3Config`.
- Duplicate keys are not allowed.

To get metadata for the compression job partition, we take advantage of the shared prefix and do the following:

- We sort keys in the lexical order.
- We use `HeadObject` API to get the metadata for the first object within the sorted key.
- We use `ListObject` API to get the metadata for the rest of keys, using the key-prefix for prefix filtering, and using the key of the first object (from last step) for `StartAfter` partition.

In this way, we obtain metadata in batches to improve performance.

# Checklist

  breaking change.

# Validation performed

I used this ingestor prototype for testing: https://github.com/LinZhihao-723/LogIngestorServer with the following steps:

- Build and start the CLP package from this branch (configured to use S3 input).
- Start the ingestor server with
  ```shell
  cargo run --release -- --db-url "mysql://clp-user:$CLP_DB_PASSWORD@localhost:3306/clp-db"
  ```
  Remember to replace the password with the actual password placeholder.
- Install the log-generation script from: https://github.com/LinZhihao-723/LogIngestorServer/tree/main/scripts
- Follow the README to create a generation job and upload logs to S3 by:
  ```shell
  python3 generate_and_upload.py \
    --s3-endpoint "https://s3.us-east-2.amazonaws.com" \
    --s3-key "$AWS_S3_KEY_ID" \
    --s3-secret "$AWS_S3_SECRET_KEY" \
    --s3-bucket "$BUCKET_NAME" \
    --num-files 10
  ```
  Remember to replace the credential and bucket placeholder by valid values (reviewer can use `lzh-test` bucket for testing. YScope developers should have a valid credential to access this bucket)
- The generation script will print an ingestion prefix. Use this prefix to create an scanner job:
  ```shell
  curl -v -u "$AWS_S3_KEY_ID:$AWS_S3_SECRET_KEY" \
  "http://127.0.0.1:8080/create?region=us-east-2&bucket=$BUCKET_NAME&dataset=test&key_prefix=$PREFIX"
  ```
  Remember to replace the credential, the bucket name, and the prefix with the valid values.
- Read the ingestor's logs to ensure log files are successfully created and scanned from S3.
- Read the ingestor's log to ensure there's exactly one compression job submitted.
- Use CLP's WebUI to check if the compression job finished successfully.

## Summary by CodeRabbit

- **New Features**
  - Job inputs can now optionally include an explicit list of S3 object keys.
  - Metadata retrieval supports both prefix-based discovery and targeted lookups by provided keys, with efficient pagination.

- **Bug Fixes**
  - Key-based lookups now surface clearer, descriptive errors for missing, duplicate or out-of-order keys.
  - Improved handling of directory-like keys and head-object errors.

- **Documentation**
  - Updated descriptions to reflect new retrieval modes, input validation rules, and error behaviours.

### DIFF
--- a/components/clp-py-utils/clp_py_utils/core.py
+++ b/components/clp-py-utils/clp_py_utils/core.py
@@ -56,7 +56,7 @@ def make_config_path_absolute(default_root: pathlib.Path, config_path: pathlib.P
 def read_yaml_config_file(yaml_config_file_path: pathlib.Path):
     with open(yaml_config_file_path, "r") as yaml_config_file:
         try:
-            config = yaml.safe_load(yaml_config_file)
+            config = yaml.load(yaml_config_file, Loader=yaml.CSafeLoader)
         except ParserError as ex:
             raise ValueError(f"Unable to parse configuration from {yaml_config_file_path}: {ex}")
     return config
--- a/tools/deployment/presto-clp/scripts/init.py
+++ b/tools/deployment/presto-clp/scripts/init.py
@@ -53,7 +53,7 @@ def main(argv=None) -> int:
         return 1
 
     with open(clp_config_file_path, "r") as clp_config_file:
-        clp_config = yaml.safe_load(clp_config_file)
+        clp_config = yaml.load(clp_config_file, Loader=yaml.CSafeLoader)
 
     env_vars: Dict[str, str] = {}
     if not _add_clp_env_vars(clp_config, clp_config_file_path, clp_package_dir, env_vars):
@@ -149,7 +149,7 @@ def _add_clp_env_vars(
         return False
 
     with open(credentials_file_path, "r") as credentials_file:
-        credentials = yaml.safe_load(credentials_file)
+        credentials = yaml.load(credentials_file, Loader=yaml.CSafeLoader)
 
     try:
         database_user = _get_required_config_value(


components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -254,36 +255,27 @@ def generate_s3_virtual_hosted_style_url(
...
     as `path` in FileMetadata.
 
     :param s3_input_config:
-    :return: List[FileMetadata] containing the object's metadata on success,
-    :raises: Propagates `boto3.client`'s exceptions.
-    :raises: Propagates `boto3.client.get_paginator`'s exceptions.
-    :raises: Propagates `boto3.paginator`'s exceptions.
+    :return: A list of `FileMetadata` containing the object's metadata on success.
+    :raises: Propagates `_s3_get_object_metadata_from_single_prefix`'s exceptions.
+    :raises: Propagates `_s3_get_object_metadata_from_keys`'s exceptions.
```
junhaoliao:
```suggestion
    :raise: Propagates `_s3_get_object_metadata_from_single_prefix`'s exceptions.
    :raise: Propagates `_s3_get_object_metadata_from_keys`'s exceptions.
```

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -392,3 +384,134 @@ def _gen_deletion_config(objects_list: List[str]):
...
+    """
+    Gets the metadata of all objects under the <`bucket`>/<`key_prefix`>.
+
+    :param s3_client:
+    :param bucket:
+    :param key_prefix:
+    :return: A list of `FileMetadata` containing the object's metadata on success.
+    :raises: Propagates `boto3.client`'s exceptions.
+    :raises: Propagates `boto3.client.get_paginator`'s exceptions.
+    :raises: Propagates `boto3.paginator`'s exceptions.
```
junhaoliao:
```suggestion
    :raise: Propagates `boto3.client`'s exceptions.
    :raise: Propagates `boto3.client.get_paginator`'s exceptions.
    :raise: Propagates `boto3.paginator`'s exceptions.
```

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -392,3 +384,134 @@ def _gen_deletion_config(objects_list: List[str]):
...
+
+    paginator = s3_client.get_paginator("list_objects_v2")
+    for page in paginator.paginate(Bucket=bucket, Prefix=key_prefix, StartAfter=first_key):
+        contents = page.get("Contents", None)
+        if contents is None:
+            continue
+
+        for obj in contents:
+            object_key = obj["Key"]
+            if object_key.endswith("/"):
```
junhaoliao: (let me know if i misunderstood the code)

my understanding is that such objects are empty placeholders that don't actually hold contents, right? However, if the user provides a list of such keys and the s3 bucket only contains such placeholder without actual content, below `ValueError(f"Cannot find following keys in the bucket `{bucket}`:\n{serialized_absent_keys}")` would be thrown, whose message is a bit inaccurate? I think we should either (preference from top to bottom):
1. Improve the handling so the behaviour is consistent with `_s3_get_object_metadata_from_single_prefix`, which returns an empty list when such key is specified.
2. OR Return an empty list even when such keys are specified (and such placeholders are found), which is consistent with the handling in `_s3_get_object_metadata_from_single_prefix`?
3. OR Return an empty list even when such keys are specified (no matter such objects are found or not). i.e., don't throw at all - which is even more consistent with the handling in `_s3_get_object_metadata_from_single_prefix`?
4. OR Improve the error message to make it more accurate.

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -392,3 +384,134 @@ def _gen_deletion_config(objects_list: List[str]):
...
+                # Skip any object that resolves to a directory-like path
+                continue
```
junhaoliao: this part of the logic seems a bit duplicated from `_s3_get_object_metadata_from_single_prefix`. is it possible to extract a helper function (as a generator like `_iter_s3_file_entries`) so we can just write the code as:
```
    for object_key, size in _iter_s3_file_entries(
        paginator.paginate(Bucket=bucket, Prefix=key_prefix, StartAfter=first_key)
    ):
        # We need to do both < and > checks since they are handled differently. Ideally, we can
        # do it with a single comparison. However, Python doesn't support three-way comparison.
        if object_key < next_key:
            continue
        if object_key > next_key:
            raise ValueError(f"Key `{next_key}` doesn't exist in the bucket `{bucket}`.")

        file_metadata_list.append(FileMetadata(Path(object_key), obj["Size"]))
        next_key = next(key_iterator, None)
        if next_key is None:
            # Early exit sine all keys have been found.
            return file_metadata_list
```

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -392,3 +384,134 @@ def _gen_deletion_config(objects_list: List[str]):
...
+    :raises: ValueError if any key in `keys` doesn't exist in the bucket.
+    :raises: Propagates `_s3_get_object_metadata_from_key`'s exceptions.
+    :raises: Propagates `boto3.client.get_paginator`'s exceptions.
+    :raises: Propagates `boto3.paginator`'s exceptions.
+    """
+    # Key validation
+    if len(keys) == 0:
+        raise ValueError("The list of keys is empty.")
+
+    keys.sort()
```
junhaoliao: this modifies the `keys` list in-place which is a side-effect? we should either document the side-effect, or preferably, just make a copy like `sorted_keys = sorted(keys)`?

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -392,3 +384,134 @@ def _gen_deletion_config(objects_list: List[str]):
             Bucket=s3_config.bucket,
             Delete=_gen_deletion_config(objects_to_delete),
         )
+
+
+def _s3_get_object_metadata_from_single_prefix(
+    s3_client: boto3.client, bucket: str, key_prefix: str
```
junhaoliao, line 390: (i understand the mistake is not introduced by the current PR)

`boto3.client` is a method rather than a type, right? to get boto3 s3 typing done right, i believe we want to use `boto3-stubs[s3]`, which is recommended by below StackOverflow responses, and also PyCharm as soon as i opened the source file. if you agree, let's ask the rabbit to create an issue and fix it in a future PR.

Ref: https://stackoverflow.com/questions/52087307/adding-type-hinting-to-functions-that-return-boto3-objects

components/clp-py-utils/clp_py_utils/s3_utils.py
junhaoliao: > my understanding is that such objects are empty placeholders that don't actually hold contents

shall we somehow document in any docstring about how we respect this convention? if we are extracting a helper, i guess we can just write a single inline comment right above the `if object_key.endswith("/")` check

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -392,3 +384,134 @@ def _gen_deletion_config(objects_list: List[str]):
...
+            continue
+
+        for obj in contents:
+            object_key = obj["Key"]
+            if object_key.endswith("/"):
+                # Skip any object that resolves to a directory-like path
+                continue
+
+            # We need to do both < and > checks since they are handled differently. Ideally, we can
+            # do it with a single comparison. However, Python doesn't support three-way comparison.
```
junhaoliao: > Python doesn't support three-way comparison.

... for strings?

junhaoliao:
```suggestion
```
the comment is a bit code/language-specific. removing the whole comment doesn't harm readability imo so maybe we can just remove

junhaoliao (changes requested):
the overall logic lgtm and is consistent with what we discussed offline. i posted some questions about refactoring and edge case handling

components/clp-py-utils/clp_py_utils/s3_utils.py
LinZhihao-723: I'd prefer to keep it as it introduces some inefficiency. If we need to reimplement this soon (assuming in Rust or other compiled lang), this is a hint for people to actually improve it. Otherwise, it's easy to be ignored.

components/clp-py-utils/clp_py_utils/s3_utils.py
LinZhihao-723: Sure. We should do a copy then.

components/clp-py-utils/clp_py_utils/s3_utils.py
LinZhihao-723: yeah sure. I didn't pay attention to this and just did copy-paste. That's a justification for keeping the comment below, lol

LinZhihao-723: Let's resolve this in another PR. @coderabbitai can u create an issue for us?

components/clp-py-utils/clp_py_utils/s3_utils.py
LinZhihao-723: Done. Tested on both ingestion path to ensure I didn't introduce any bug.

components/clp-py-utils/clp_py_utils/s3_utils.py
LinZhihao-723: I will add one in the helper.

LinZhihao-723: As discussed offline:
- For prefix-based ingestion, we will keep the behavior untouched, so any directory-like keys will be ignored.
- Error out on key-based ingestion, with a message indicating that a key ends with `/` is invalid.

PR #1415, opened 2025-10-13
fix(build): Use checksum to skip `mariadb-connector-cpp` rebuild.

### LINKED ISSUE #1412: deps:spider task performs unnecessary rebuilds due to mariadb-connector-cpp
## Problem

The `deps:spider` task rebuilds spider even when no source files have changed, causing unnecessary build time (~22 seconds per invocation).

## Root Cause

The cmake build for `mariadb-connector-cpp` modifies the source, causing it to rebuild every time the task is called. This triggers spider to rebuild as well.

## Proposed Solution

Use checksum utilities to prevent `mariadb-connector-cpp` from rerunning when its sources haven't actually changed.

## Context

- Parent issue: #1409
- Discussion: https://github.com/y-scope/clp/issues/1409
- Reported by: @junhaoliao
- Root cause identified by: @sitaowang1998

sitaowang1998:
# Description

## Root cause
As discussed in #1409 and later tracked in #1412, `mariadb-connector-cpp` rebuilds every time because the build modifies the source tree, so in every run we re-extract the directory and the files are updated so the build reruns.

## Solution
This PR fixes #1412 by
* Copies the extracted directory to src directory to avoid re-download and extraction of the library.
* Uses checksum utils from `yscope-dev-utils` to avoid re-execute `mariadb-connector-cpp`.

# Checklist

  breaking change.

# Validation performed

* [x] Rerun `task deps:spider` does not rebuilt `mariadb-connector-cpp` and `spider`.
* [x] GitHub workflows pass.

## Summary by CodeRabbit

- Chores
  - Added a new download-and-validate workflow for the MariaDB C++ connector, including checksum verification and generation.
  - Enabled incremental rebuilds by generating and tracking a checksum for the installed artifact.
  - Updated build to copy extracted sources into a dedicated source directory and use source-based CMake configuration.

- Refactor
  - Reworked dependency build flow and task orchestration for improved reliability and repeatability.

---
  - https://app.asana.com/0/0/1212888453704773

### DIFF
--- a/taskfiles/deps/main.yaml
+++ b/taskfiles/deps/main.yaml
@@ -390,6 +390,27 @@ tasks:
           TARBALL_SHA256: "537512904744b35e232912055ccf8ec66d768639ff3abe5788d90d792ec5f48b"
           TARBALL_URL: "https://github.com/lz4/lz4/releases/download/v1.10.0/lz4-1.10.0.tar.gz"
 
+  mariadb-connector-cpp-download-and-validate-checksum:
+    internal: true
+    cmds:
+      - task: "yscope-dev-utils:remote:download-and-extract-tar"
+        vars:
+          FILE_SHA256: "0e3dfe9f2bc3f7bb6f7c159009556290064a7c23402ea08019fa8aebfc3ff2c9"
+          OUTPUT_DIR: "{{.G_DEPS_CPP_DIR}}/mariadb-connector-cpp-extracted"
+          URL: "https://github.com/mariadb-corporation/mariadb-connector-cpp/archive/refs/tags/\
+            1.1.5.tar.gz"
+      # Uses checksum to skip mariadb-connector-cpp installation because the build updates the
+      # source directory and will always rebuild.
+      - task: "yscope-dev-utils:checksum:validate"
+        vars:
+          CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
+          INCLUDE_PATTERNS: &mariadb-cmake-include-patterns
+            - "{{.G_DEPS_CPP_DIR}}/mariadb-connector-cpp-build"
+            - "{{.G_DEPS_CPP_DIR}}/mariadb-connector-cpp-extracted"
+            - "{{.G_DEPS_CPP_DIR}}/mariadb-connector-cpp-extracted.tar.gz"
+            - "{{.G_DEPS_CPP_DIR}}/mariadb-connector-cpp-install"
+            - "{{.G_DEPS_CPP_DIR}}/mariadb-connector-cpp-src"
+
   mariadb-connector-cpp:
     platforms: ["linux"]
     preconditions:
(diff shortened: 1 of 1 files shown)


taskfiles/deps/main.yaml
```diff
@@ -390,6 +390,29 @@ tasks:
...
+          CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
+          INCLUDE_PATTERNS: &mariadb-cmake-include-patterns
+            - "{{.G_DEPS_CPP_DIR}}/mariadb-connector-cpp-build"
+            - "{{.G_DEPS_CPP_DIR}}/mariadb-connector-cpp-extracted"
+            - "{{.G_DEPS_CPP_DIR}}/mariadb-connector-cpp-extracted.tar.gz"
+            - "{{.G_DEPS_CPP_DIR}}/mariadb-connector-cpp-install"
+            - "{{.G_DEPS_CPP_DIR}}/mariadb-connector-cpp-src"
+
   mariadb-connector-cpp:
     platforms: ["linux"]
```
junhaoliao, line 420: since we have changed a great portion of this task - let's reorder these task attributes according to https://docs.yscope.com/dev-guide/contrib-guides-taskfiles.html#ordering-of-task-attributes

junhaoliao (commented):
the logic looks reasonable to me though the changes look a bit verbose. I vaguely remember discussing a similar approach with @davidlion though i don't know if there's any convention we should follow / any example we can refer to.

@davidlion / @Bill-hbrhbr would you mind guiding / completing this review as I observe?

Bill-hbrhbr:
I think the current checksum system is flawed and most libraries rebuild themselves even when nothing has changed.

The proper solution is to fix the checksum system used in `deps:utils:install-remote-cmake-lib`. The most glaring bug is that this task generates a checksum, but doesn't have a `generates:` section. The checksum is not being used for checking if the task is up-to-date.

There are also smaller problems such as I don't really get how the combined checksum file in `deps:utils:combine-cpp-checksum-files` is useful for preventing re-runs, since there's no way to validate this file. We might as well checksum the whole `clp/build/deps/cpp` directory. IMO we should remove this feature as properly checksuming individual libraries is enough for quick taskflow runs.

Bill-hbrhbr:
@junhaoliao Just to make sure I understand the problem root cause:
I ran `mariadb-connector-cpp` builds a couple times. The sha256sum of `<lib>-extracted` and `<lib>-install` didn't change over multiple runs. A proper solution just needs to checksum the `<lib>-install` directory to see if a re-run is necessary, so I don't think this lib is a special case scenario.

In summary, this PR seems to be too much effort to fix a single library while there's a more general and methodical solution.

taskfiles/deps/main.yaml
sitaowang1998: I have reordered the attributes, but some attributes are not recorded in the guideline, i.e. `internal` and `preconditions`.

sitaowang1998:
@Bill-hbrhbr makes some good points. I now switch to only checksum the `install` directory.
It is possible to use `deps:utils:install-remote-cmake-lib`, but the task need to support copying the extracted to another directory to avoid polluting the source tree and trigger a re-extract.
I also agree that the combined checksum seem to not work. Aside from the bug in it, the boost task currently only checksum the `generate` directory but not the final `install` directory. If we want to make use of the combined checksum, we need to fix the boost task first.

taskfiles/deps/main.yaml
davidlion: Fyi, those two attributes are in the list (unless someone updated it since your message).

davidlion:
> @Bill-hbrhbr makes some good points. I now switch to only checksum the `install` directory. It is possible to use `deps:utils:install-remote-cmake-lib`, but the task need to support copying the extracted to another directory to avoid polluting the source tree and trigger a re-extract. I also agree that the combined checksum seem to not work. Aside from the bug in it, the boost task currently only checksum the `generate` directory but not the final `install` directory. If we want to make use of the combined checksum, we need to fix the boost task first.

Fwiw https://github.com/y-scope/yscope-dev-utils/pull/81 and https://github.com/y-scope/yscope-dev-utils/pull/80 are directly what you're talking about, but have been needing reviewers for a while. Bill recently did a pass so hopefully we can land them soon.

Since this is an optimization, I think it makes sense to have those 2 PRs land and do this properly.

taskfiles/deps/main.yaml
sitaowang1998: My fault. Fixed.

PR #1417, opened 2025-10-14
refactor(clp-package): Simplify StrEnum and Path serialization via Annotated serializers.

sitaowang1998:
# Description

This PR adds pydantic annotated serializer for `pathlib.Path` and `StrEnum` to avoid adding custom model serialization for fields of these types.

# Checklist

  breaking change.

# Validation performed

* [x] Celery compression works.

## Summary by CodeRabbit

* **Improvements**
  * Standardized string-serialized enums and unified path handling across configuration for more consistent cross‑platform behaviour and stronger validation.

* **New Features**
  * Added explicit serialization utilities for enums and filesystem paths to stabilise exported/portable formats.

* **Refactor**
  * Simplified configuration serialization by removing legacy primitive‑dict export paths and updating initialization/validation semantics.

### DIFF
--- a/components/clp-py-utils/clp_py_utils/clp_config.py
+++ b/components/clp-py-utils/clp_py_utils/clp_config.py
@@ -743,7 +743,7 @@ class CLPConfig(BaseModel):
 class WorkerConfig(BaseModel):
     package: Package = Package()
     archive_output: ArchiveOutput = ArchiveOutput()
-    data_directory: pathlib.Path = CLPConfig().data_directory
+    data_directory: SerializablePath = CLPConfig().data_directory
 
     # Only needed by query workers.
     stream_output: StreamOutput = StreamOutput()


components/clp-py-utils/clp_py_utils/pydantic_serialization_utils.py
junhaoliao, line 1: For the file name, how about just `file_utils`? Though I understand the utils here likely are only compatible with Pydantic, the name "Pydantic" is still a bit implementation-specific which i feel can be omitted from the name.

junhaoliao: actually, can we also:
* Only export the lambda functions in `serialization_utils.py`: `serialize_path = lambda ...`
* Create the Pydantic types in `clp_config.py` (or move them into a separate `types.py` source file. maybe in a future refactoring PR) e.g.,
```py
# in clp_config.py
PathStr = Annotated[
    pathlib.Path,
    PlainSerializer(serialize_path)
]
```

so that we can decouples the serialization utils from Pydantic?

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -590,20 +568,18 @@ class CLPConfig(BaseModel):
...
+    data_directory: PathStr = pathlib.Path("var") / "data"
+    logs_directory: PathStr = pathlib.Path("var") / "log"
     aws_config_directory: Optional[pathlib.Path] = None
 
-    _container_image_id_path: pathlib.Path = PrivateAttr(
-        default=CLP_PACKAGE_CONTAINER_IMAGE_ID_PATH
-    )
-    _version_file_path: pathlib.Path = PrivateAttr(default=CLP_VERSION_FILE_PATH)
+    _container_image_id_path: PathStr = PrivateAttr(default=CLP_PACKAGE_CONTAINER_IMAGE_ID_PATH)
+    _version_file_path: PathStr = PrivateAttr(default=CLP_VERSION_FILE_PATH)
```
junhaoliao: (just to show we have considered -)

Since these are `PrivateAttr`, they won't serialize by default anyway. The type change is harmless though.

components/clp-py-utils/clp_py_utils/pydantic_serialization_utils.py
```diff
@@ -0,0 +1,8 @@
+import pathlib
+from typing import Annotated
+
+from pydantic import PlainSerializer
+
+StrEnumSerializer = PlainSerializer(lambda enum_value: enum_value.value)
+
+PathStr = Annotated[pathlib.Path, PlainSerializer(lambda path_value: str(path_value))]
```
junhaoliao: (i'm not sure if this is worth doing but we shall investigate)

can we embed the validator `_validate_directory` with `BeforeValidator()` into the `Annotation` as well?
https://docs.pydantic.dev/latest/api/functional_validators/#pydantic.functional_validators.BeforeValidator

components/clp-py-utils/clp_py_utils/pydantic_serialization_utils.py
sitaowang1998: I feel `TypeStr` is not a very intuitive name. Is there a better suggestion?

sitaowang1998: I am ok with another name, but `file_utils` is confusing as it is not even close to what this file is doing.

components/clp-py-utils/clp_py_utils/pydantic_serialization_utils.py
sitaowang1998: I tried. Somehow annotated `BeforeValidator` is not triggered. I haven't figured out why. Maybe we can ask coderabbit to open an issue to track this?

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -108,23 +110,36 @@
 Port = Annotated[int, Field(gt=0, lt=2**16)]
 ZstdCompressionLevel = Annotated[int, Field(ge=1, le=19)]
 
+StrEnumSerializer = PlainSerializer(serialize_enum)
+
+PathStr = Annotated[pathlib.Path, PlainSerializer(serialize_path)]
```
junhaoliao: shall we group these two with the above? Let's also place the Pydantic serializers at the top because the annotations *might* (though not currently) refer the serializers.

e.g.,
```py
# Serializers
StrEnumSerializer = PlainSerializer(serialize_str_enum)
# Generic types
...
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -232,7 +244,6 @@ def get_clp_connection_params_and_type(self, disable_localhost_socket_connection
 
     def dump_to_primitive_dict(self):
         d = self.model_dump(exclude={"username", "password"})
```
junhaoliao, line 245: in a separate PR, we shall investigate
specifying private attributes like `username` as https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.PrivateAttr
and
specifying secrets like `password` as https://docs.pydantic.dev/2.2/usage/types/secrets/

so we can completely get rid of such `dump_to_primitive_dict()` methods.

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: the annotated `pathlib.Path` is not a `str` so it might be misleading to name it `PathStr`. Do you think it would be better to name it as `SerializablePath`?

components/clp-py-utils/clp_py_utils/serialization_utils.py
```diff
@@ -0,0 +1,23 @@
+import pathlib
+
+from strenum import StrEnum
+
+
+def serialize_enum(enum_value: StrEnum) -> str:
```
junhaoliao: how about renaming as  `serialize_str_enum()`?

junhaoliao: accurately speaking, the argument here is a symbolic name of the enum ("enum member"): https://docs.python.org/3/library/enum.html#:~:text=An%20enumeration%3A,members%20by%20name

Shall we name the argument as `member` instead?

components/clp-py-utils/clp_py_utils/serialization_utils.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Serializes a StrEnum to its underlying value.
```
junhaoliao:
```suggestion
    Serializes a `strenum.StrEnum` member to its underlying value.
```

components/clp-py-utils/clp_py_utils/serialization_utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+    :param enum_value: A StrEnum instance.
```
junhaoliao:
```suggestion
    :param enum_value:
```
or if you agree with the proposed renaming
```suggestion
    :param member:
```

components/clp-py-utils/clp_py_utils/serialization_utils.py
```diff
@@ -0,0 +1,23 @@
...
+    :return: The underlying string value of the StrEnum.
```
junhaoliao:
```suggestion
    :return: The underlying string value of the enum member.
```

components/clp-py-utils/clp_py_utils/serialization_utils.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    return enum_value.value
+
+
+def serialize_path(path: pathlib.Path) -> str:
+    """
+    Serializes a pathlib.Path to its string representation.
```
junhaoliao:
```suggestion
    Serializes a `pathlib.Path` to its string representation.
```

components/clp-py-utils/clp_py_utils/serialization_utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+    :param path: A pathlib.Path instance.
```
junhaoliao:
```suggestion
    :param path:
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -590,20 +573,18 @@ class CLPConfig(BaseModel):
...
 
     presto: Optional[Presto] = None
 
     archive_output: ArchiveOutput = ArchiveOutput()
     stream_output: StreamOutput = StreamOutput()
-    data_directory: pathlib.Path = pathlib.Path("var") / "data"
-    logs_directory: pathlib.Path = pathlib.Path("var") / "log"
+    data_directory: PathStr = pathlib.Path("var") / "data"
+    logs_directory: PathStr = pathlib.Path("var") / "log"
     aws_config_directory: Optional[pathlib.Path] = None
```
```suggestion
    aws_config_directory: Optional[SerializablePath] = None
```
if it's not intentional to leave it unchanged?

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -590,20 +573,18 @@ class CLPConfig(BaseModel):
...
 
-    _container_image_id_path: pathlib.Path = PrivateAttr(
-        default=CLP_PACKAGE_CONTAINER_IMAGE_ID_PATH
-    )
-    _version_file_path: pathlib.Path = PrivateAttr(default=CLP_VERSION_FILE_PATH)
+    _container_image_id_path: PathStr = PrivateAttr(default=CLP_PACKAGE_CONTAINER_IMAGE_ID_PATH)
+    _version_file_path: PathStr = PrivateAttr(default=CLP_VERSION_FILE_PATH)
 
     @field_validator("aws_config_directory")
     @classmethod
```
junhaoliao, line 591: if we changed the above `aws_config_directory` type as `Optional[SerializablePath]`, we shall update below `value` type as `Optional[SerializablePath]` in below validator prototype for consistency

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -778,16 +752,6 @@ class WorkerConfig(BaseModel):
     stream_output: StreamOutput = StreamOutput()
     stream_collection_name: str = ResultsCache().stream_collection_name
 
-    def dump_to_primitive_dict(self):
-        d = self.model_dump()
-        d["archive_output"] = self.archive_output.dump_to_primitive_dict()
-
-        # Turn paths into primitive strings
-        d["data_directory"] = str(self.data_directory)
```
junhaoliao, line 786: do we need to update
```
data_directory: SerializablePath = CLPConfig().data_directory
```

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: I checked other references of `pathlib.Path` and believe they are fine to be left as `pathlib.Path` because those are used in public interfaces and the callers do not need to make the paths serializable.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -735,22 +716,15 @@ def get_runnable_components(self) -> Set[str]:
...
         }
         d = self.model_dump(exclude=custom_serialized_fields)
         for key in custom_serialized_fields:
             d[key] = getattr(self, key).dump_to_primitive_dict()
 
         # Turn paths into primitive strings
-        d["credentials_file_path"] = str(self.credentials_file_path)
-        d["data_directory"] = str(self.data_directory)
-        d["logs_directory"] = str(self.logs_directory)
         if self.aws_config_directory is not None:
```
junhaoliao: we should be able to remove this as well if the type of `aws_config_directory` is updated as suggested above?

junhaoliao (changes requested):
overall lgtm

for the title, how about:
```
refactor(clp-package): Simplify StrEnum and Path serialization via Annotated serializers.
```
The "A" in "Annotated" is intentionally capitalized because it's a Python feature

components/clp-py-utils/clp_py_utils/clp_config.py
sitaowang1998: I am a little confused about your suggestion. Isn't this self referencing?

components/clp-py-utils/clp_py_utils/clp_config.py
sitaowang1998: Then we have to manually serialize them in the `dump_to_primitive_dict`.

components/clp-py-utils/clp_py_utils/serialization_utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def serialize_str_enum(member: StrEnum) -> str:
+    """
+    Serializes a `strenum.StrEnum` member to its underlying value.
+
+    :param member:
+    :return: The underlying string value of the enum member.
+    """
+    return member.value
```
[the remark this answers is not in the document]
sitaowang1998: We omit the description of a parameter if it is obvious from the name.

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: In Line 746, `WorkerConfig`'s `data_directory` is still typed as `pathlib.Path`, which would mean that if anyone tries to call `.model_dump()` on the model and serialize to YAML, they wouldn't be able to do so

it's true that there isn't such logic at the moment, and we should probably deprecate the `WorkerConfig` model soon once we're done with refactoring the secrets / private fields in CLPConfig. that said, let's still change the type so we consistently use `SerializablePath` to annotate all `pathlib.Path` in all Pydantic models?

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: Right. Sorry if i may have confused you, i meant to say

> I checked other references of `pathlib.Path` and believe "all the other references" are fine to be left as `pathlib.Path`

-> the `value` argument in the validator of `aws_config_directory` is the ONLY argument whose type should be updated as `SerializablePath`
-> that is, let's not change the type from `pathlib.Path` to `SerializablePath` in any `def make_config_paths_absolute(self, clp_home: pathlib.Path)` or `def load_credentials_from_file(self, credentials_file_path: pathlib.Path)` prototypes

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -107,24 +111,34 @@
 DomainStr = NonEmptyStr
 Port = Annotated[int, Field(gt=0, lt=2**16)]
 ZstdCompressionLevel = Annotated[int, Field(ge=1, le=19)]
+SerializablePath = Annotated[pathlib.Path, PlainSerializer(serialize_path)]
```
junhaoliao, line 113:
```suggestion
SerializablePath = Annotated[pathlib.Path, PlainSerializer(serialize_path)]
ZstdCompressionLevel = Annotated[int, Field(ge=1, le=19)]
```

PR #1405, opened 2025-10-10
build(clp-package): Adopt uv + hatchling as the build and packaging backend for Python components (resolves #1396); Upgrade dependencies for Python components.

junhaoliao:
# Description

* For Python packages `clp-package-utils`, `clp-py-utils`, `job-orchestration`:
  * Migrate the package management system from poetry to uv
  * Migrate the build system from poetry to hatchling
* For Python packages `clp-package-utils`, `clp-py-utils`, `job-orchestration`, `integration-tests`:
* Upgrade dependencies in all Python components and make sure versions of common dependencies match.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

1. Built the package.
2. Uncommented the contents in `build/clp-package/etc/clp-config.yml`.
3. Started the package: `sbin/start-clp.sh`
4. Compressed sample logs: `sbin/compress.sh ~/samples/hive-24hr`
5. Performed a query in the webui and observed results listed.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

- Chores
  - Migrated packaging/build tooling across components, consolidated dependency declarations and standardized authors metadata; removed obsolete tooling entries.
  - Switched CI/task build references to a new component runner.

- Tests
  - Bumped integration test Python version to 3.10.

- Bug Fixes
  - Ensured Celery task registration and related job constants are accessible for runtime task registration.

- Style
  - Minor import ordering and whitespace cleanups for consistency.

PR #1423, opened 2025-10-15
feat(clp-package): Add Python profiling decorator for internal query performance measurement.

Eden-D-Zhang:
# Description

This pull request introduces a new lightweight profiling utility for CLP query execution based on PyInstrument and applies it to certain query execution functions. The profiling decorator generates HTML flame graphs and text summaries for performance analysis, with context extraction for job and task IDs.

Currently, profiling can only be enabled/disabled by modifying the return value of `is_profiling_enabled` in `profiling_utils.py`. Per-component config options should be added in a future PR.

* Added new `profiling_utils.py` module in `clp_py_utils` providing a `profile` decorator for both sync and async functions, automatic context extraction, and output of HTML and text profiling reports to a designated directory.
* Added `pyinstrument` as a dependency in `pyproject.toml` to support profiling features.
* Added `@profile` decorator to some `query_scheduler.py` functions.
* Added `@profile` decorator to `search` and `extract_stream` task functions.

Profiling output is written to `<CLP_LOGS_DIR>/profiles` as two files for each execution of a profiled function that are labeled with job, task, and timestamp in the filename.

* `.txt` file with call stack timing information:
image

* Interactive `.html` file with call stack and flame graph visualization:
image

# Checklist

  breaking change.

# Validation performed

Performed queries using the CLP package with profiling enabled and disabled.

## Summary by CodeRabbit

* **New Features**
  * Added decorator-based profiling across query tasks and scheduler entry points to capture runtime performance.
  * Profiling supports sync/async functions, extracts job/task context, skips nested profiling, and logs events.
  * Saves interactive HTML and plain-text profile outputs with metadata and timestamps for easier analysis.

* **Chores**
  * Added a runtime dependency for the profiling tool.

---
  - https://app.asana.com/0/0/1212888470600438


components/clp-py-utils/pyproject.toml
```diff
@@ -16,6 +16,7 @@ boto3 = "^1.35.81"
 mariadb = "~1.0.11"
 mysql-connector-python = "^8.2.0"
 pydantic = "^2.11.9"
+pyinstrument = "^5.0.0"
```
junhaoliao:
```suggestion
pyinstrument = "^5.1.1"
```

components/clp-py-utils/pyproject.toml
junhaoliao, line 1: also can we run `poetry lock` inside `components/clp-py-utils` to update the `poetry.lock` file?

components/clp-py-utils/pyproject.toml
junhaoliao: do we plan to include the profiling feature in the formal releases? if not, maybe we should put this into a
```
[dependency-groups]
dev = [
    ...
]
```

that said, the `pyinstrument` package is only ~270KB (https://pypi.org/project/pyinstrument/#files) so it shouldn't hurt to be bundled into the release.

junhaoliao (changes requested):
(not a formal review)

note we are migrating the build system from poetry to uv in #1405 so feel free to hold off on the dependency changes

components/clp-py-utils/clp_py_utils/profiling_utils.py
```diff
@@ -0,0 +1,263 @@
...
+
+logger = logging.getLogger(__name__)
+
+F = TypeVar("F", bound=Callable[..., Any])
+
+PROFILING_INTERVAL = 0.001
+
+
+def profile(
+    section_name: Optional[str] = None,
```
LinZhihao-723:
```suggestion
    section_name: str | None = None,
```
Do it for all occurrences.

components/clp-py-utils/clp_py_utils/profiling_utils.py
```diff
@@ -0,0 +1,263 @@
...
+    job_id_param: Optional[str] = None,
+    task_id_param: Optional[str] = None,
+) -> Callable[[F], F]:
+    """
+    Decorator for profiling function execution with automatic context extraction.
```
LinZhihao-723: `Decorator` to `Decorators`.
Do it for all docstrings.

components/clp-py-utils/clp_py_utils/profiling_utils.py
```diff
@@ -0,0 +1,263 @@
...
+
+    Output files are written to $CLP_LOGS_DIR/profiles/ (e.g., clp-package/var/log/query_worker/
+    profiles/).
+
+    :param section_name: Override for profile section name. If None, uses function name.
+    :param job_id_param: Parameter name to extract job_id from (default: "job_id").
+                         Can use dot notation for attributes, e.g., "job.id"
```
LinZhihao-723: WE don't indent to `:`; we should only indent for 4 spaces. @junhaoliao can u confirm just in case I'm outdated?

components/clp-py-utils/clp_py_utils/profiling_utils.py
```diff
@@ -0,0 +1,263 @@
...
+                try:
+                    result = await func(*args, **kwargs)
+                    return result
+                finally:
+                    profiler.stop()
+                    _save_profile(profiler, name, job_id, task_id)
+
+            return async_wrapper  # type: ignore
+
+        else:
```
LinZhihao-723: Don't add `else` after the previous `if` block unconditionally returns.

components/clp-py-utils/clp_py_utils/profiling_utils.py
```diff
@@ -0,0 +1,263 @@
...
+    task_id_param: Optional[str] = None,
+) -> Tuple[str, str]:
+    """
+    Extract job_id and task_id from function arguments.
+
+    :param func: The function being profiled
+    :param args: Positional arguments passed to the function
+    :param kwargs: Keyword arguments passed to the function
+    :param job_id_param: Name/path of the parameter containing job_id (default: "job_id").
+    :param task_id_param: Name/path of the parameter containing task_id (default: "task_id").
```
LinZhihao-723, line 135: Why not making the parameters to default these values instead of setting them optional?

components/clp-py-utils/clp_py_utils/profiling_utils.py
```diff
@@ -0,0 +1,263 @@
...
+    func: Callable,
+    args: tuple,
+    kwargs: dict,
+    job_id_param: Optional[str] = None,
+    task_id_param: Optional[str] = None,
+) -> Tuple[str, str]:
+    """
+    Extract job_id and task_id from function arguments.
+
+    :param func: The function being profiled
```
LinZhihao-723:
```suggestion
    :param func: The function being profiled.
```
Don't miss `.`. Do it for all docstrings.

components/clp-py-utils/clp_py_utils/profiling_utils.py
```diff
@@ -0,0 +1,263 @@
...
+
+    return job_id, task_id
+
+
+def _is_profiling_enabled() -> bool:
+    """
+    Check if profiling is enabled.
+    TODO: Add `CLPConfig` mechanism to enable/disable profiling for each component.
+
+    :return: False
```
LinZhihao-723:
```suggestion
    :return: Whether the profiler is enabled.
```

LinZhihao-723 (changes requested):
Not a full review either. Just catch some errors when taking a glance.
It would be better if we can add some instructions in the PR to indicate:
- How to turn on the profiler (seems like to tune the flag in the code level and rebuild the package)
- A step-by-step setup for testing, and what are the expected outputs?

components/clp-py-utils/clp_py_utils/profiling_utils.py
Eden-D-Zhang: In general, are there cases where do we prefer `Optional` or should it always be ` | None`?

components/clp-py-utils/pyproject.toml
```diff
@@ -21,4 +22,4 @@ dependencies = [
 
 [build-system]
 requires = ["hatchling"]
-build-backend = "hatchling.build"
+build-backend = "hatchling.build"
```
junhaoliao:
```suggestion
build-backend = "hatchling.build"

```

components/clp-py-utils/uv.lock
junhaoliao, line 1: It appears the revision number was somehow downgraded. can you check what version of `uv` do you have on your end? You might want to rerun the `components/core/tools/scripts/lib_install/pipx-packages/install-uv.sh` script to update the version.


Then can we restore the lock file back to the latest commit on main before performing the `uv lock`. like
```sh
cd <project-root>
git checkout 13f2cfd12550baf6c36d5fef0235d65bf41b4831 -- components/clp-py-utils/uv.lock
uv lock --no-upgrade
```

components/clp-py-utils/clp_py_utils/profiling_utils.py
```diff
@@ -0,0 +1,263 @@
...
+import os
+from pathlib import Path
+from typing import Any, Callable, Optional, Tuple, TypeVar
+
+from pyinstrument import Profiler
+
+logger = logging.getLogger(__name__)
+
+F = TypeVar("F", bound=Callable[..., Any])
+
```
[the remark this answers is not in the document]
junhaoliao: i agree with the rabbit

components/clp-py-utils/clp_py_utils/profiling_utils.py
```diff
@@ -0,0 +1,256 @@
...
+from pathlib import Path
+from typing import Any, Callable, Tuple, TypeVar
+
+from pyinstrument import Profiler
+
+logger = logging.getLogger(__name__)
+
+F = TypeVar("F", bound=Callable[..., Any])
+
+PROFILING_INTERVAL = 0.001
```
junhaoliao: how about

```suggestion
PROFILING_INTERVAL_SECONDS = 0.001
```

components/clp-py-utils/clp_py_utils/profiling_utils.py
```diff
@@ -0,0 +1,257 @@
...
+import inspect
+import logging
+import os
+from collections.abc import Callable
+from pathlib import Path
+from typing import Any, TypeVar
+
+from pyinstrument import Profiler
+
+logger = logging.getLogger(__name__)
```
sitaowang1998: Use the `get_logger` from `clp_logging.py`.

sitaowang1998 (commented):
Not a formal review. Need to setup logger correctly to use the uniform format.

PR #1430, opened 2025-10-16
fix(ci): Correctly update and restore cache for `lint:check-cpp-lint-static-full`'s generated files (fixes #1419):

### LINKED ISSUE #1419: ci: `lint:check-cpp-lint-static-full` cache not being updated
### Bug

The cache for `lint:check-cpp-lint-static-full`'s output is not being updated as can be seen from cache entries which show that they were last updated between 27 days and 6 months ago (see `gh cache list`).

This is leading to long workflow run times for both the [clp-core-build-macos](https://github.com/y-scope/clp/blob/66238292d4769a8c695a006b80f84cf211b3a31c/.github/workflows/clp-core-build-macos.yaml) and [clp-artifact-build](https://github.com/y-scope/clp/blob/66238292d4769a8c695a006b80f84cf211b3a31c/.github/workflows/clp-artifact-build.yaml) workflows since the they need to re-run clang-tidy on all C++ files that have been changed since the caches were last updated.

### CLP version

6623829

### Environment

GitHub runners:

* macos-14
* macos-15
* ubuntu-24.04

### Reproduction steps

Run the mentioned workflows.

kirkrodrigues:
- Save cache entries using unique key per entry.
- Restore latest entries using key prefix.
- Avoid using outputs from optionally-run `restore` task.

# Description

#815 added caching for the input and output of `lint:check-cpp-lint-static-full`, but two issues were preventing the cache from being updated correctly:

1. I misunderstood the docs for the cache action, so the cache was not being updated on every push to `main`. Specifically, cache entries are immutable for a given key, so every new entry needs to have a new key. I mistakenly set the cache key to be the same for every entry, so no new entries could be written.
2. In scheduled runs, the cache was not being saved since the key was empty. This is because the key was derived from the restore step, but scheduled runs don't run the restore step (intentionally).

This PR resolves the issues as follows:

1. When saving a cache entry, we now append a hash (of the entry's contents) to the key. When restoring the cache, we use a prefix of that key, and `actions/cache/restore` will automatically select the newest entry.

> [!NOTE]
> Although the `action/cache` [docs](https://github.com/actions/cache/blob/0057852bfaa89a56745cba8c7296529d2fc39830/README.md#cache-version) imply that cache entries are stored in different namespaces determined by the OS used, running
`gh api -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" /repos/y-scope/clp/actions/caches` shows that cache entries for different OSes have the same version. So we still must use a different cache key prefix per OS and build configuration.

2. For `action/cache/save`, we specify the cache key manually rather than trying to rely on the value from `action/cache/restore`.

This fix should reduce clp-core-build-macos workflow runtimes from 2-3 hours to ~20 minutes.

# Checklist

  breaking change.

# Validation performed

* Merged the change to `main` in my fork.
* Validated that the new cache entries were created successfully.
* Re-ran the clp-build-artifact and clp-macos-build workflows and validated that they utilized the cache and ran much more quickly.
* Changed a C++ file on this branch and validated that clang-tidy was only run on that file, while all others were cached.

> [!NOTE]
> clang-tidy violations are not being detected correctly, so we can't test if violations will be detected by the workflow. This will be resolved in another PR.

## Summary by CodeRabbit

* **Chores**
  * Improved CI workflow caching: adjusted restore/save behavior so scheduled runs perform full lint checks, and introduced OS/config-specific cache keys with hash-based inputs to keep restore/save prefixes in sync — reducing missed lint issues and improving build consistency and reliability across environments.


Bill-hbrhbr:
For title, how about:
```
fix(ci): Enable proper cache updates and reuse for C++ static linting task checksums (fixes #1419).
```

kirkrodrigues:
> For title, how about:
> 
> ```
> fix(ci): Enable proper cache updates and reuse for C++ static linting task checksums (fixes #1419).
> ```

How about a multi-line PR title & body:

```text
fix(ci): Correctly update and restore cache for `lint:check-cpp-lint-static-full`'s generated files (fixes #1419):
- Save cache entries using unique key per entry.
- Restore entries using key prefix.
- Don't use outputs from optionally run `restore` task.
```

kirkrodrigues:
> Other than we could update the pinned action sha to the latest v4.3.0 release, i don't have other comments. Logic looks correct but there's no way to test until we merge to `main`.
> 
> ```
> 0057852bfaa89a56745cba8c7296529d2fc39830
> ```

I'd prefer to do this in another PR so that future readers don't think this was a necessary change for this PR.

Bill-hbrhbr:
> > For title, how about:
> > ```
> > fix(ci): Enable proper cache updates and reuse for C++ static linting task checksums (fixes #1419).
> > ```
> 
> How about a multi-line PR title & body:
> 
> ```
> fix(ci): Correctly update and restore cache for `lint:check-cpp-lint-static-full`'s generated files (fixes #1419):
> - Save cache entries using unique key per entry.
> - Restore entries using key prefix.
> - Don't use outputs from optionally run `restore` task.
> ```

`Restore latest entries`...
`Avoid using attributes/outputs`

just some suggestions. otherwise lgtm

kirkrodrigues:
Bypassing the remaining workflows since they will take hours to run, they shouldn't fail based on tests on my fork, and the effort of running them will be wasted unless they're run on `main`.

PR #1434, opened 2025-10-17
feat(clp-s): Explicitly reject unstructured log inputs during compression.

gibber9809:
# Description

This PR adds support for detecting that an input might be unstructured logs during ingestion, and uses that support to explicitly reject unstructured log inputs during compression. This feature will also be used by the upcoming utility that converts unstructured logs to a structured representation for ingestion into clp-s.

The heuristic for detecting logtext is straightforwards -- we simply check if the first peeked chunk of the input stream corresponds to valid UTF-8, while accounting for the fact that the end of the buffer might contain truncated UTF-8.

Since a UTF-8 codepoint is at most 4 bytes, we know that we must scan at most 7 bytes to find the last complete UTF-8 codepoint in our peeked input. This is because in the worst case the stream terminates with a 4-byte codepoint followed by a 4-byte codepoint with its last byte truncated. The approach, then, is to scan the last 7 bytes of the peeked buffer to find the last complete codepoint and then validate that the entire buffer through that last complete codepoint corresponds to valid UTF-8. If there is not a complete codepoint in the last 7 bytes, we also reject the input without having to validate the whole buffer.

# Checklist

  breaking change.

# Validation performed

* Manually validated that various combinations of truncated UTF-8 are accepted as logtext
* Manually validated that various combinations of invalid UTF-8 are not accepted as logtext

## Summary by CodeRabbit

* **New Features**
  * Detects and classifies "logtext" input files during ingestion.

* **Bug Fixes / Behaviour**
  * Supplying a logtext file for direct ingestion now fails fast with a clear error and stops the write operation.

### DIFF
--- a/components/webui/client/src/sql-parser/index.ts
+++ b/components/webui/client/src/sql-parser/index.ts
@@ -119,7 +119,7 @@ WHERE to_unixtime(${timestampKey}) BETWEEN ${startTimestamp.unix()} AND ${endTim
     try {
         validate(queryString);
     } catch (err: unknown) {
-        throw new Error(`The constructed SQL is not valid: ${queryString}`, {cause: err});
+        console.error(`The constructed SQL is not valid: ${queryString}`, err);
     }
 
     return queryString;
@@ -185,7 +185,7 @@ ORDER BY timestamps.idx
     try {
         validate(queryString);
     } catch (err: unknown) {
-        throw new Error(`The constructed SQL is not valid: ${queryString}`, {cause: err});
+        console.error(`The constructed SQL is not valid: ${queryString}`, err);
     }
 
     return queryString;


components/core/src/clp_s/InputConfig.cpp
```diff
@@ -242,6 +253,49 @@ auto could_be_json(char const* peek_buf, size_t peek_size) -> bool {
...
+            mark_last_legal_character(0ULL);
+        }
+    }
+
+    // We were not able to find a complete UTF-8 character, so the input is definitely not UTF-8.
+    if (false == legal_last_byte_index.has_value()) {
+        return false;
+    }
+
+    return simdjson::validate_utf8(peek_buf, legal_last_byte_index.value() + 1ULL);
```
davidlion, line 312: We should try to follow the google and linux guidelines on comments. Generally speaking, the only time to have comments inline of code should be to explain some confusing behaviour (usually from an external library/dependency).

For comments applying to only a few lines usually it implies the naming is too confusing to read or the comment is just repeating the code. If it is not possible to making the naming in the code readable on its own, then we can add an implementation comment above the method/function definition. This puts all of the logic in one place so a reader can understand the implementation without needing to scan everything.

In this specific case, I think merging these comments together above the definition is fine.

(Sorry, I know most of the code generally doesn't follow this yet, but we need to start somewhere.)

davidlion (approved):

PR #1436, opened 2025-10-17
feat(clp-mcp-server): Add `search_by_kql` MCP tool call.

20001020ycx:
# Description

This PR implements search API for MCP server: `search_by_kql`. 
The documentation of `search_by_kql` is listed below:

## `search_by_kql(string: kql_query)`

Description:
Runs a search query that returns the 10 latest log messages that satisfy this KQL query (ordered by timestamp), the Agent can further access up to 1000 latest log messages that satisfy this KQL query using pagination. Pagination allows the Agent to query 10 log messages (one page) at a time using a page index. For example, the MCP agent can query log messages 100-110 using page index 10. The agent uses get_nth_page() to query a page from the search results.

Inputs:
kql_query: The query LLM wants to search for.

Outputs:
- The 10 newest log messages that satisfy the search query with their timestamp in datestring format.
- The pagination metadata for the query response.

Errors:
- Error: The query was CANCELLED
- Error: The query FAILED
- Error: The query was KILLED
- Error: The query job TIMED OUT

# Checklist

  breaking change.

# Validation performed

1. Unit Tests: `uv run --group dev pytest tests/server/test-utils.py -v.`
2. Lint Tests: `uv run ruff check clp/components/clp-mcp-server && task lint:check-py.`
3. Manual Integration Tests: Built using`task docker-images:package` with docker compose. Connected Claude Desktop agent to the running MCP server service in docker compose. Manually instructed LLM to perform a substring_search_query, the retruned result is consistent with manual grep, with the correct formatted date string. You can see the end-to-end result using this [link](https://claude.ai/share/59a6bb8b-dcc4-45c7-a42b-b7b32780d958)
Screenshot 2025-10-20 at 1 51 44 PM
Note: AI team is actively working on the automated integration test.

## Summary by CodeRabbit

* **New Features**
  * CLI config file support via new --config-path option
  * KQL search tool to run queries and return log entries

* **Enhancements**
  * Results post-processed: timestamps converted to ISO‑8601, sorted and formatted for display
  * New "N/A" placeholder for missing timestamps
  * Startup loads and validates configuration before server creation

* **Bug Fixes**
  * Docstrings updated to document TimeoutError and clarify error annotations

* **Tests**
  * Unit tests for filtering, timestamp handling and sorting

### DIFF
--- a/.github/workflows/clp-rust-checks.yaml
+++ b/.github/workflows/clp-rust-checks.yaml
@@ -2,7 +2,17 @@ name: "clp-rust-checks"
 
 on:
   pull_request:
+    paths: &monitored_paths
+      - ".cargo/**"
+      - "Cargo.lock"
+      - "Cargo.toml"
+      - ".github/workflows/clp-rust-checks.yaml"
+      - "components/clp-rust-utils/**"
+      - "components/log-ingestor/**"
+      - "taskfile.yaml"
+      - "taskfiles/**"
   push:
+    paths: *monitored_paths
   schedule:
     # Run daily at 00:15 UTC (the 15 is to avoid periods of high load)
     - cron: "15 0 * * *"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -332,7 +332,7 @@ dependencies = [
  "hyper-util",
  "pin-project-lite",
  "rustls 0.21.12",
- "rustls 0.23.33",
+ "rustls 0.23.34",
  "rustls-native-certs 0.8.2",
  "rustls-pki-types",
  "tokio",
@@ -1154,7 +1154,7 @@ dependencies = [
  "http 1.3.1",
  "hyper 1.7.0",
  "hyper-util",
- "rustls 0.23.33",
+ "rustls 0.23.34",
  "rustls-native-certs 0.8.2",
  "rustls-pki-types",
  "tokio",
@@ -1697,9 +1697,9 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.33"
+version = "0.23.34"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "751e04a496ca00bb97a5e043158d23d66b5aabf2e1d5aa2a0aaebb1aafe6f82c"
+checksum = "6a9586e9ee2b4f8fab52a0048ca7334d7024eef48e2cb9407e3497bb7cab7fa7"
 dependencies = [
  "aws-lc-rs",
  "once_cell",
@@ -2094,7 +2094,7 @@ version = "0.26.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1729aa945f29d91ba541258c8df89027d5792d85a8841fb65e8bf0f4ede4ef61"
 dependencies = [
- "rustls 0.23.33",
+ "rustls 0.23.34",
  "tokio",
 ]
 


components/clp-mcp-server/clp_mcp_server/clp_mcp_server.py
```diff
@@ -15,12 +19,19 @@
     "--host", type=str, default="127.0.0.1", help="The server's host address (default: 127.0.0.1)."
 )
 @click.option("--port", type=int, default=8000, help="The server's port number (default: 8000).")
-def main(host: str, port: int) -> None:
+@click.option(
+    "--config",
+    type=click.Path(exists=True),
+    default="/etc/clp-config.yml",
+    help="The path to server's configuration file (default: /etc/clp-config.yml).",
+)
```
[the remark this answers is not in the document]
20001020ycx: I want this brittle feature

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    :return: A list of strings containing only the `timestamp` (as a date string) and `message` of
+    a log entry.
+    """
+    filtered = []
+    for obj in query_results:
+        epoch = obj.get("timestamp")
+        try:
+            timestamp_str = convert_epoch_to_date_string(epoch)
+        except (TypeError, ValueError) as e:
+            logger.warning("Failed to convert epoch timestamp=%s to date string: %s", epoch, e)
```
20001020ycx:
```suggestion
            logger.warning("Failed to convert epoch timestamp=%s to date string: %s.", epoch, e)
```

components/clp-mcp-server/clp_mcp_server/server/server.py
```diff
@@ -67,4 +72,32 @@ def hello_world(name: str = "clp-mcp-server user") -> dict[str, Any]:
...
+    @mcp.tool
+    async def search_kql_query(kql_query: str, ctx: Context) -> dict[str, object]:
+        """
+        Searches the logs for the specified KQL query and returns the first page of the matching
+        result. The paginated results are ordered from latest to oldest, with each page containing
+        the paging metadata and 10 log messages along with their date string timestamps.
+
+        :param kql_query:
+        :param ctx: The `FastMCP` context containing the metadata of the underlying MCP session.
+        :return: Forwards `FastMCP.tool`''s `get_nth_page`'s return values on success:
```
20001020ycx:
```suggestion
        :return: Forwards `FastMCP.tool`'s `get_nth_page`'s return values on success:
```

components/clp-mcp-server/clp_mcp_server/server/server.py
```diff
@@ -67,4 +72,32 @@ def hello_world(name: str = "clp-mcp-server user") -> dict[str, Any]:
...
+    @mcp.tool
+    async def search_kql_query(kql_query: str, ctx: Context) -> dict[str, object]:
+        """
+        Searches the logs for the specified KQL query and returns the first page of the matching
+        result. The paginated results are ordered from latest to oldest, with each page containing
+        the paging metadata and 10 log messages along with their date string timestamps.
+
+        :param kql_query:
+        :param ctx: The `FastMCP` context containing the metadata of the underlying MCP session.
+        :return: Forwards `FastMCP.tool`'s `get_nth_page`'s return values on success:
```
20001020ycx:
```suggestion
        :return: Forwards `FastMCP.tool.get_nth_page`'s return values on success:
```

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -50,14 +52,14 @@ async def submit_query(self, query: str, begin_ts: int, end_ts: int) -> str:
         :raise pymongo.errors.PyMongoError: If there is an error interacting with MongoDB.
         :raise Exception: For any other unexpected errors.
         """
-        if end_ts < begin_ts:
+        if begin_ts is not None and end_ts is not None and end_ts < begin_ts:
```
LinZhihao-723, line 59: Can u update the docstring? The `end/begin_ts` is surrounded by double \` by accident

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -121,6 +123,7 @@ async def wait_query_completion(self, query_id: str, timeout: float | None = Non
         :raise aiomysql.Error: If there is an error connecting to or querying MariaDB.
         :raise ValueError: When the query is not found.
         :raise RuntimeError: When the query fails or is cancelled.
+        :raise TimeoutError: When the timeout is reached before the query completes.
```
LinZhihao-723: I'm noto sure if `When` is a good word here. Shall we use `If`?

components/clp-mcp-server/clp_mcp_server/clp_mcp_server.py
```diff
@@ -15,12 +19,19 @@
...
+def main(host: str, port: int, config: Path) -> None:
     """
     Runs the CLP MCP server with HTTP transport.
 
     :param host: The server's host address (IP address or hostname).
     :param port: The server's port number (1-65535).
+    :param config: The path to server's configuration file.
```
LinZhihao-723: How about `config` -> `config_path`? `config` sounds like an in-memory object.

components/clp-mcp-server/clp_mcp_server/clp_mcp_server.py
```diff
@@ -15,12 +19,19 @@
     "--host", type=str, default="127.0.0.1", help="The server's host address (default: 127.0.0.1)."
 )
 @click.option("--port", type=int, default=8000, help="The server's port number (default: 8000).")
-def main(host: str, port: int) -> None:
+@click.option(
+    "--config",
```
LinZhihao-723: How about `--config-path` for the same reason?

components/clp-mcp-server/clp_mcp_server/server/server.py
```diff
@@ -4,11 +4,14 @@
...
 
+from clp_mcp_server.clp_connector import ClpConnector
+
 from . import constants
 from .session_manager import SessionManager
+from .utils import filter_query_results, sort_query_results
 
 
-def create_mcp_server() -> FastMCP:
+def create_mcp_server(clp_config: Any) -> FastMCP:
```
LinZhihao-723: - U forgot to update the docstring.
- Shouldn't the type annotation be `CLPConfig`?

LinZhihao-723: Actually I think `ClpConnector`'s constructor is also wrong, lol
At least the docstring should be updated.

components/clp-mcp-server/clp_mcp_server/server/server.py
```diff
@@ -67,4 +72,32 @@ def hello_world(name: str = "clp-mcp-server user") -> dict[str, Any]:
...
+    @mcp.tool
+    async def search_kql_query(kql_query: str, ctx: Context) -> dict[str, object]:
+        """
+        Searches the logs for the specified KQL query and returns the first page of the matching
+        result. The paginated results are ordered from latest to oldest, with each page containing
+        the paging metadata and 10 log messages along with their date string timestamps.
+
+        :param kql_query:
+        :param ctx: The `FastMCP` context containing the metadata of the underlying MCP session.
+        :return: Forwards `FastMCP.tool.get_nth_page`'s return values on success:
```
LinZhihao-723:
```suggestion
        :return: A dictionary containing the following key-value pairs on success:
            - "items": A list of log entries in the requested page.
            - "num_total_pages": Total number of pages available from the query as an integer.
            - "num_total_items": Total number of log entries available from the query as an integer.
            - "num_items_per_page": Number of log entries per page.
            - "has_next": Whether a page exists after the returned one.
            - "has_previous": Whether a page exists before the returned one.
```
This isn't exactly correct. It doesn't forward `get_nth_page`, right?
I think it's better to explicitly write it out (to not confuse AI), and document in `get_page_data` (we documented in `NOTE` section that they should sync).

components/clp-mcp-server/clp_mcp_server/server/server.py
```diff
@@ -67,4 +72,32 @@ def hello_world(name: str = "clp-mcp-server user") -> dict[str, Any]:
             "status": "running",
         }
 
+    @mcp.tool
+    async def search_kql_query(kql_query: str, ctx: Context) -> dict[str, object]:
```
LinZhihao-723: I don't think "search a query" is a correct term. How about `run_kql_query`? @junhaoliao what do u think?

components/clp-mcp-server/clp_mcp_server/server/server.py
```diff
@@ -67,4 +72,32 @@ def hello_world(name: str = "clp-mcp-server user") -> dict[str, Any]:
...
+        """
+        Searches the logs for the specified KQL query and returns the first page of the matching
+        result. The paginated results are ordered from latest to oldest, with each page containing
+        the paging metadata and 10 log messages along with their date string timestamps.
```
LinZhihao-723: - I'm not sure if it's a good idea to document "10 log messages". It's a magic number now and it can be easily out-of-sync. If we document the returned kv-pair schema clearly, I don't think we need to explicitly mention how many results are returned in a page, right?
- "log message along with their date string timestamps" sounds like a "log event". In our terminology, we should just use `log event` to refer to messages + timestamps.

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def convert_epoch_to_date_string(epoch_ts: int) -> str:
+    """
+    :param epoch_ts: Unix epoch timestamp in milliseconds.
+    :return: ISO 8601 formatted date string with millisecond precision (YYYY-MM-DDTHH:mm:ss.fffZ).
+    :raise: TypeError if `epoch_ts` is None or not an int type.
+    :raise: ValueError if `epoch_ts` cannot be converted to a valid date string.
+    """
+    if epoch_ts is None:
```
LinZhihao-723: The input type is `int` so it should never be `None` right? Otherwise, it should be `int | None`.
I don't think we need to handle `None` inside this function tbh. The caller should make sure the input is a valid int.

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,66 @@
...
+        err_msg = "Epoch timestamp cannot be None."
+        raise TypeError(err_msg)
+
+    if not isinstance(epoch_ts, int):
+        err_msg = f"Object {type(epoch_ts).__name__} is not of type int."
```
LinZhihao-723:
```suggestion
        err_msg = f"Object `{type(epoch_ts).__name__}` is not of type `int`."
```

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,66 @@
...
+        raise TypeError(err_msg)
+
+    try:
+        epoch_seconds = epoch_ts / 1000.0
+        dt = datetime.fromtimestamp(epoch_seconds, tz=timezone.utc)
+        return dt.strftime("%Y-%m-%dT%H:%M:%S.%f")[:-3] + "Z"
```
LinZhihao-723: Do we have any library call that can do ISO format directly? This looks a bit hacky; but I don't know if this is the only option we have...

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    except (ValueError, OSError, OverflowError) as e:
+        err_msg = f"Invalid timestamp {epoch_ts}: {e}."
+        raise ValueError(err_msg) from e
+
+
+def filter_query_results(query_results: list[dict]) -> list[str]:
```
LinZhihao-723: `list[dict]` is not a proper type annotation: `dict` is like a template, u must fill it with concrete key type and value type. You should use `list[dict[str, Any]]`.

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    :param query_results: A list of dictionary containing log entries with its metadata.
```
LinZhihao-723: What do u mean by "its metadata"?
I feel like u could just say `A list of dictionary representing kv-pair log events.`

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    :return: A list of strings containing only the `timestamp` (as a date string) and `message` of
+    a log entry.
```
LinZhihao-723:
```suggestion
        a log event.
```
- Don't forget about indentation.
- Let's use `log event` from now to be consistent with our internal naming rule.

components/clp-mcp-server/clp_mcp_server/server/utils.py
LinZhihao-723: "filter" is not a correct term here. I think what this function does is `project_and_format`. What do you think? @junhaoliao

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    MongoDB.
+    :return: A sorted list of dictionary containing log entries with its metadata, ordered by epoch
+    from latest to oldest.
+    """
+
+    def _key(log_entry: dict) -> int:
+        ts = log_entry.get("timestamp")
+        return ts if isinstance(ts, int) else -1
+
+    return sorted(query_results, key=_key, reverse=True)
```
LinZhihao-723: Apply above comments to this one as well.

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,66 @@
...
+            logger.warning("Failed to convert epoch timestamp=%s to date string: %s.", epoch, e)
+            timestamp_str = "N/A"
+
+        message = obj.get("message", "")
+        filtered.append(f"timestamp: {timestamp_str}, message: {message}")
+
+    return filtered
+
+
+def sort_query_results(query_results: list[dict]) -> list[dict]:
```
LinZhihao-723: How about `sort_by_timestamp`?

LinZhihao-723: Do we have any concern to sort this list inline?
This might be a problem as it doubles the returned results in peak memory usage. This can be a problem if:
- The returned results have a large volume.
- There are many concurrent queries.

LinZhihao-723 (changes requested):
Haven't looked into the test cases.
Will go through docstrings once we all agreed on the API/method names.

components/clp-mcp-server/clp_mcp_server/server/server.py
20001020ycx: yeah, I think we might need some refactor after the release, but it is my bad for missing the docstring.

components/clp-mcp-server/clp_mcp_server/server/utils.py
20001020ycx: I thought Python's type is merely a hint, right? So if we handle this at the caller, wouldn't we make this repetitive for each callsites?

I think the same comment goes with the check with `if not isinstance(epoch_ts, int):`. If we ensure it is an int, then we dont need to dynamically check the type right?

Based on your comment and the actual use case, do you think it makes more sense if I declare the input type to be `Any`?

components/clp-mcp-server/clp_mcp_server/server/utils.py
20001020ycx: good catch, will note on this

components/clp-mcp-server/clp_mcp_server/server/utils.py
20001020ycx: This is a valid concern, and I think we should be memory efficient. A side note is that for any given request, the maximum returned log events are 1000. However, this is a soft contraint that may be violated in the future, so the bullet 1 can be true in the future.

Just to be cleear, I am not sure what do you mean by sort this list inline, do you mean code like below:
```
query_results.sort(key=_key, reverse=True)

return query_results
```

components/clp-mcp-server/clp_mcp_server/server/server.py
20001020ycx: I see, I will follow the same format as what you have wrote for get_neth_page.

This is a question for next pr where we implement substring_wz_timestamp, the type of the returned result is the same, how should I write the docstring there? Following your logic in not confusing AI, I think even though substring_wz_timestamp has the same return type with substring search, each return type should be listed explicitly.

components/clp-mcp-server/clp_mcp_server/server/server.py
20001020ycx: maybe `run_kql_query_search`? I think we should encode the purpose of why this tool call in the name so that AI can select which tool call is best for the current use case.

components/clp-mcp-server/clp_mcp_server/server/server.py
20001020ycx: - This is a good catch, in fact, the returned paging metadata already include the number of items per page, so I think 10 log messages can be omitted then.
- I see, this is very important to know, in this case, do you think the description of docstring can be written as:

> Searches the logs for the specified KQL query and returns the first page of the matching
> result, the paginated log events are ordered from latest to oldest.

components/clp-mcp-server/clp_mcp_server/server/utils.py
20001020ycx: my oversight, agreed, chatgpt didnt find it at the first place, lol

components/clp-mcp-server/clp_mcp_server/server/utils.py
20001020ycx: I think I am losing a bit of background here. What are kv-pair log events? Do you mean JSON-formatted log event?

This is what I mean by metadata, and what I mean by filtering:
The result read from MongoDB is this:
```
{
    "_id": "test000",
    "timestamp": None,
    "message": '{"pid":null,"tid":null,"message":"Log at epoch none"}\n',
    "orig_file_path": "/var/log/app.log",
    "archive_id": "abc123",
    "log_event_ix": 99,
}
```
What I mean by metadata are fields: _id, timestamp, message, orig_file_path, etc. For this release, we only care about timestamp and message, that's why we decided to filter other fields away.

But I agree metadata is vague in this docstring, does your suggested kv-pair log events capture the semantics of the above example I have? If yes, I don't have any problem.

components/clp-mcp-server/clp_mcp_server/server/server.py
LinZhihao-723: Yeah, I think that might be the best practice for now unless:
- This leads to too many tokens.
- It's proven that we don't need to be that verbose.

components/clp-mcp-server/clp_mcp_server/server/server.py
LinZhihao-723: how about `search_by_kql`?

components/clp-mcp-server/clp_mcp_server/server/server.py
LinZhihao-723: `the paginated log events are ordered from latest to oldest.` might be a bit weird. I will help polish it in the next round of review when we finalize the AIP name.

components/clp-mcp-server/clp_mcp_server/server/utils.py
LinZhihao-723: > I thought Python's type is merely a hint, right? So if we handle this at the caller, wouldn't we make this repetitive for each callsites?

* Yes, it is just a hint. But when the hint doesn't indicate the type can be optional, we don't assume it can be `None`.
* In general I would agree with u. But in this particular case, this function should only be responsible for converting an int input to a date string. It would be super weird to assume an optional int can be passed into this function without a caller-level check. Consider u write this in Rust/C++, would u make the function parameter to be `Option<int>`/`std::optional<int>` and always return an error when the given value is null, or just simply make the interface to take an `int`?

> I think the same comment goes with the check with if not isinstance(epoch_ts, int):. If we ensure it is an int, then we dont need to dynamically check the type right?

I agree. Normally, we shouldn't need this type of runtime check. You can rely on the linter to catch these type issues. I missed this in this round of review.

> Based on your comment and the actual use case, do you think it makes more sense if I declare the input type to be Any?

Absolutely no. You should not declare the input type to be `Any`. In that way, everything should be declared as `Any` and runtime type check should be enforced everywhere, and then tbh u should probably switch to some other typed language like Go or Rust, lol

components/clp-mcp-server/clp_mcp_server/server/utils.py
LinZhihao-723: * In the latest CLP's context, we call these dynamically structured log events "kv-pair log events".
* "metadata" is ambiguous in the context of kv-pairs inside a log event. We tried to avoid that.
* If this docstring is directly AI oriented, I'd consider to call it JSON-formatted log event, since I'm not sure whether we need to explain what is kv-pair log event to AI. But since this docstring is for our internal use, I'd say let's use "kv-pair log event".

components/clp-mcp-server/clp_mcp_server/server/utils.py
LinZhihao-723: Yes. We should document that the returned list is the input list, and the sorting happens in-place.

components/clp-mcp-server/clp_mcp_server/server/utils.py
20001020ycx: Got it, thanks for the best practice guide, got rid of the type check in the conversion function and added the check at callsite.

components/clp-mcp-server/clp_mcp_server/server/utils.py
20001020ycx: I see, good to know.

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -50,7 +50,7 @@ async def submit_query(
         :param begin_ts: The beginning timestamp of the query range.
         :param end_ts: The end timestamp of the query range.
         :return: The ID assigned to the query.
-        :raise ValueError: If ``end_ts`` is smaller than ``begin_ts``.
+        :raise ValueError: If `end_ts` is smaller than `begin_ts`.
         :raise aiomysql.Error: If there is an error connecting to or querying MariaDB.
         :raise pymongo.errors.PyMongoError: If there is an error interacting with MongoDB.
         :raise Exception: For any other unexpected errors.
```
junhaoliao: (not strictly related to the PR) I believe we should also document the `RuntimeError`

components/clp-mcp-server/clp_mcp_server/clp_mcp_server.py
```diff
@@ -52,7 +63,16 @@ def main(host: str, port: int) -> None:
         sys.exit(1)
 
     try:
-        mcp = create_mcp_server()
+        clp_config = CLPConfig.model_validate(read_yaml_config_file(config_path))
+    except ValidationError:
+        logger.exception("Configuration validation failed.")
+        sys.exit(1)
```
junhaoliao: shall we return an exit code in `main()` and rewrite the bottom part as
```
if __name__ == "__main__":
   sys.exit(main())
```
like other scripts

components/clp-mcp-server/clp_mcp_server/server/server.py
```diff
@@ -4,14 +4,18 @@
...
 
+from clp_mcp_server.clp_connector import ClpConnector
+
 from . import constants
 from .session_manager import SessionManager
+from .utils import filter_query_results, sort_by_timestamp
 
 
-def create_mcp_server() -> FastMCP:
+def create_mcp_server(config_path: Any) -> FastMCP:
```
junhaoliao: should the type of the `config_path` be `str` or `pathlib.Path`?

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,62 @@
...
+        dt = datetime.fromtimestamp(epoch_seconds, tz=timezone.utc)
+        return dt.isoformat(timespec="milliseconds").replace("+00:00", "Z")
+    except (ValueError, OSError, OverflowError) as e:
+        err_msg = f"Invalid timestamp {epoch_ts}: {e}."
+        raise ValueError(err_msg) from e
+
+
+def filter_query_results(query_results: list[dict[str, Any]]) -> list[str]:
+    """
+    :param query_results: A list of dictionary representing kv-pair log events.
```
20001020ycx:
```suggestion
    :param query_results: A list of dictionaries representing kv-pair log events.
```

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,62 @@
...
+
+        message = obj.get("message", "")
+        filtered.append(f"timestamp: {timestamp_str}, message: {message}")
+
+    return filtered
+
+
+def sort_by_timestamp(query_results: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    """
+    :param query_results: A list of dictionary representing kv-pair log events.
```
20001020ycx:
```suggestion
    :param query_results: A list of dictionaries representing kv-pair log events.
```

components/clp-mcp-server/clp_mcp_server/server/server.py
```diff
@@ -67,4 +73,38 @@ def hello_world(name: str = "clp-mcp-server user") -> dict[str, Any]:
             "status": "running",
         }
 
+    @mcp.tool
+    async def search_by_kql(kql_query: str, ctx: Context) -> dict[str, Any]:
```
junhaoliao, line 78: do models know what KQL is? whenever i google the first things that pop up are Kusto Query Language, which seems also related to log and telemetry data analysis

if not, would it be more clear to say "Kibana Query Language" in the comments?

components/clp-mcp-server/clp_mcp_server/server/utils.py
junhaoliao: or maybe simply `format(query_results)`

by the way, `list[dict]` can be a bit confusing. Shall we define annotate the `dict` type with a name and use the type name here?

components/clp-mcp-server/clp_mcp_server/server/server.py
20001020ycx: I can add this in the docstring, but we have informed LLM the context of kql in the system prompt

components/clp-mcp-server/clp_mcp_server/server/server.py
20001020ycx: This should be CLPConfig, I will change ClpConnector as well

components/clp-mcp-server/clp_mcp_server/server/utils.py
20001020ycx: type is already updated, so is this the final naming:
`def format(query_results: list[dict[str, Any]]) -> list[str]:`
@LinZhihao-723 for confirmation, I am okay with either, but I feel like it's losing the object of the verb

LinZhihao-723: I think `format` is ok. In the docstring, please document only timestamp and message will be used for formatting.

components/clp-mcp-server/clp_mcp_server/server/server.py
```diff
@@ -2,16 +2,21 @@
...
 
+from clp_mcp_server.clp_connector import ClpConnector
+
 from . import constants
 from .session_manager import SessionManager
+from .utils import format_query_results, sort_by_timestamp
 
 
-def create_mcp_server() -> FastMCP:
+def create_mcp_server(config_path: CLPConfig) -> FastMCP:
```
LinZhihao-723: Shouldn't this be `clp_config`?

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    except (ValueError, OSError, OverflowError) as e:
+        err_msg = f"Invalid timestamp {epoch_ts}: {e}."
+        raise ValueError(err_msg) from e
+
+
+def format_query_results(query_results: list[dict[str, Any]]) -> list[str]:
+    """
+    :param query_results: A list of dictionaries representing kv-pair log events.
+    :return: A list of strings only containing the formatted `timestamp` (as a date string) and
+        `message` of a log event.
```
LinZhihao-723:
```suggestion
    Formats the query results. For a log event to be formatted, it must contain the following
    kv-pairs:
    - "timestamp": An integer representing the epoch timestamp in milliseconds.
    - "message": A string representing the log message.

    The message will be formatted as `timestamp: <date string>, message: <message>`:

    :param query_results: A list of dictionaries representing kv-pair log events.
    :return: A list of strings representing formatted log events.
```

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    filtered = []
```
LinZhihao-723: Shouldn't we use `formatted_log_events`?

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,62 @@
...
+        epoch = obj.get("timestamp")
+        timestamp_str = TIMESTAMP_NOT_AVAILABLE
+
+        if isinstance(epoch, int):
+            try:
+                timestamp_str = convert_epoch_to_date_string(epoch)
+            except (TypeError, ValueError) as e:
+                logger.warning("Failed to convert epoch timestamp=%s to date string: %s.", epoch, e)
+
+        message = obj.get("message", "")
```
LinZhihao-723, line 50: I'm not sure if this would actually be useful. Shall we also log a warning and skip if there's no `message` field?

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,62 @@
...
+        filtered.append(f"timestamp: {timestamp_str}, message: {message}")
```
LinZhihao-723: Just curious: why don't we return structured log events directly, but with other fields stripped and timestamp formatted as date strings?

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,62 @@
...
+
+    return filtered
+
+
+def sort_by_timestamp(query_results: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    """
+    :param query_results: A list of dictionaries representing kv-pair log events.
+    :return: The same list, sorted in-place by epoch from latest to oldest.
```
LinZhihao-723:
```suggestion
    Sorts the query results in-place by timestamp in descending order (latest to oldest).

    NOTE:
    - Timestamp is expected to be an integer representing the epoch timestamp in milliseconds,
      stored under the `timestamp` key.
    - If `timestamp` is missing or not an integer, it is treated as the oldest possible timestamp.

    :param query_results: A list of dictionaries representing kv-pair log events.
    :return: The input list sorted in-place.
```

components/clp-mcp-server/clp_mcp_server/server/utils.py
```diff
@@ -0,0 +1,62 @@
...
+def convert_epoch_to_date_string(epoch_ts: int) -> str:
+    """
+    :param epoch_ts: Unix epoch timestamp in milliseconds.
+    :return: ISO 8601 formatted date string with millisecond precision (YYYY-MM-DDTHH:mm:ss.fffZ).
+    :raise: ValueError if `epoch_ts` cannot be converted to a valid date string.
+    """
+    try:
+        epoch_seconds = epoch_ts / 1000.0
+        dt = datetime.fromtimestamp(epoch_seconds, tz=timezone.utc)
+        return dt.isoformat(timespec="milliseconds").replace("+00:00", "Z")
```
LinZhihao-723, line 21: Just a sanity check: u tested this would work, right?

components/clp-mcp-server/clp_mcp_server/server/server.py
```diff
@@ -67,4 +74,37 @@ def hello_world(name: str = "clp-mcp-server user") -> dict[str, Any]:
             "status": "running",
         }
 
+    @mcp.tool
+    async def search_by_kql(kql_query: str, ctx: Context) -> dict[str, Any]:
+        """
+        Searches the logs for the specified Kibana Query Language (KQL) query and returns the first
+        page of the matching result, the paginated log events are ordered from latest to oldest.
```
LinZhihao-723:
```suggestion
        Searches log events that match the given Kibana Query Language (KQL) query. The resulting
        events are ordered by timestamp in descending order (latest to oldest), cached for
        subsequent pagination, and returned with the first page of results.
```

components/clp-mcp-server/clp_mcp_server/server/server.py
20001020ycx: rabbit also catched this, already made the update, lol

components/clp-mcp-server/clp_mcp_server/server/utils.py
20001020ycx: yes, unit test has covered this conversion

components/clp-mcp-server/tests/server/test_utils.py
```diff
@@ -0,0 +1,149 @@
...
+            '{"pid":null,"tid":null,'
+            '"message":"Log at epoch none"}\n'
+        ),
+    ]
+
+    def test_invalid_timestamp_type(self):
+        """Validates the handling of noninteger timestamp types."""
+        result = format_query_results(self.INVALID_TYPE_ENTRIES)
+
+        assert result == self.INVALID_TYPE_EXPECTED
```
LinZhihao-723: Can we rename everything to `EXPECTED_XXX`?

components/clp-mcp-server/clp_mcp_server/server/utils.py
20001020ycx: sure

components/clp-mcp-server/tests/server/test_utils.py
```diff
@@ -0,0 +1,149 @@
...
+
+        assert result == self.INVALID_VALUE_EXPECTED
+    
+    def test_missing_timestamp_and_message(self):
+        """Validates the handling of log entries without timestamp and message field."""
+        result = format_query_results(self.MISSING_TIMESTAMP_AND_MESSAGE_ENTRY)
+
+        assert result == self.MISSING_TIMESTAMP_AND_MESSAGE_EXPECTED
+
+    def test_sort_and_filter_query_results(self):
```
LinZhihao-723: Need to update the name for `filter`, lol

components/clp-mcp-server/tests/server/test_utils.py
```diff
@@ -0,0 +1,149 @@
...
+        """
+        Validates the post-processing functionality. We expect:
+        - Logs are sorted by timestamp from latest to oldest, with logs having invalid timestamps
+        appear at last.
+        - `timestamp` and `message` fields are correctly filtered.
```
LinZhihao-723: Personally I don't think this is necessary, lol

components/clp-mcp-server/clp_mcp_server/server/utils.py
20001020ycx: date string is for clear understanding of the time semantic because it is more natural lanuage process friendly compared to epoch.

Stripping other fields are my personal choice as I dont think they are useful and would consume extra token use.

LinZhihao-723 (approved):
For the PR title, how about:
```
feat(clp-mcp-server): Add `search_by_kql` MCP tool call.
```
- to make it concise
- to add the missing period

PR #1445, opened 2025-10-19
fix(clp-rust-utils): Use AWS SDK default configuration with latest behavior version for S3 client.

LinZhihao-723:
# Description

In the Log Ingestor prototyping, we later realized that creating a new client without a default configuration would cause connection loss very frequently when accessing the AWS-hosted S3 service. This is because that a client configured without the default config doesn't set HTTP behavior, timeouts, and credentials resolution.

This PR fixes this problem by introducing AWS SDK's "latest" behavior version as the default config.

# Checklist

  breaking change.

# Validation performed

* [x] Tested in the prototype repo to ensure clients with the default config are stable to use.

## Summary by CodeRabbit

* **Refactor**
  * Updated AWS SDK integration and dependencies to use current defaults.
  * S3 client creation converted to an asynchronous flow for better startup handling.
  * Improved S3 endpoint compatibility by enforcing path-style addressing.
  * Credentials handling updated for clearer provider identification and more reliable configuration.

### DIFF
--- a/.github/workflows/clp-artifact-build.yaml
+++ b/.github/workflows/clp-artifact-build.yaml
@@ -463,8 +463,10 @@ jobs:
         run: "chown $(id -u):$(id -g) -R ."
         shell: "bash"
 
-      # NOTE: We don't use the cache for scheduled runs so that they run lint:check-cpp-static-full
-      # on all files.
+      # NOTE: We don't restore the cache for `schedule` runs so that below,
+      # `lint:check-cpp-static-full` runs on all files. If we don't do this periodically,
+      # `lint:check-cpp-static-full` could miss issues in files that haven't changed but depend on
+      # files which have changed.
       - if: "'schedule' != github.event_name"
         name: "Restore lint:check-cpp-static-full cache"
         id: "cache-restore-lint-check-cpp-static-full"
@@ -475,9 +477,10 @@ jobs:
             .task/checksum/utils-cpp-lint-clang-tidy-*
             build/lint-clang-tidy
 
-          # NOTE: We use a per-OS cache since different OSes may trigger different clang-tidy
-          # violations.
-          key: "main-branch-ubuntu-jammy-lint:check-cpp-static-full"
+          # NOTE: This key must be kept in-sync with the key prefix in the `actions/cache/save` step
+          # below (we can't use the output of the current step since `schedule` workflow runs don't
+          # run this step).
+          key: "lint:check-cpp-static-full-on-ubuntu-24.04"
 
       - uses: "./.github/actions/run-on-image"
         env:
@@ -505,7 +508,15 @@ jobs:
             .task/checksum/lint-check-cpp-static-full
             .task/checksum/utils-cpp-lint-clang-tidy-*
             build/lint-clang-tidy
-          key: "${{steps.cache-restore-lint-check-cpp-static-full.outputs.cache-primary-key}}"
+
+          # NOTE: This key prefix (without the hash) must be kept in-sync with the
+          # `actions/cache/restore` step above.
+          key: >-
+            lint:check-cpp-static-full-on-ubuntu-24.04-${{hashFiles(
+              '.task/checksum/lint-check-cpp-static-full',
+              '.task/checksum/utils-cpp-lint-clang-tidy-*',
+              'build/lint-clang-tidy/**/*'
+            )}}
 
   package-image:
     name: "package-image"
--- a/.github/workflows/clp-core-build-macos.yaml
+++ b/.github/workflows/clp-core-build-macos.yaml
@@ -94,8 +94,10 @@ jobs:
           --num-jobs $(getconf _NPROCESSORS_ONLN)
           --test-spec "~[Stopwatch]"
 
-      # NOTE: We don't use the cache for scheduled runs so that they run lint:check-cpp-static-full
-      # on all files.
+      # NOTE: We don't restore the cache for `schedule` runs so that below,
+      # `lint:check-cpp-static-full` runs on all files. If we don't do this periodically,
+      # `lint:check-cpp-static-full` could miss issues in files that haven't changed but depend on
+      # files which have changed.
       - if: "'schedule' != github.event_name"
         name: "Restore lint:check-cpp-static-full cache"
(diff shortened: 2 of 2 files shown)


components/clp-rust-utils/src/s3/client.rs
```diff
@@ -23,7 +30,8 @@ pub fn create_new_client(
         "clp-user",
     );
     let region = Region::new(region_id.to_owned());
-    let config = Builder::new()
+    let base_config = aws_config::defaults(BehaviorVersion::latest()).load().await;
+    let config = Builder::from(&base_config)
```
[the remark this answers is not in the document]
LinZhihao-723: None of these errors are returned in this function, right?

components/clp-rust-utils/src/s3/client.rs
```diff
@@ -6,10 +7,16 @@ use secrecy::{ExposeSecret, SecretString};
...
+///
+/// * The client is configured using the latest AWS SDK behavior version.
+/// * The client enforces path-style addressing.
+///
 /// # Returns
+///
 /// A newly created S3 client.
 #[must_use]
-pub fn create_new_client(
+pub async fn create_new_client(
```
junhaoliao, line 19: changed as `async` which is an internal breaking change -

i confirmed that there's no usage of the function at the moment and therefore there's no caller to be updated, so the change looks correct

seeing `Client::from_conf()` might panic, i'm not sure if we should wrap things and let this return a `Result` though

components/clp-rust-utils/src/s3/client.rs
```diff
@@ -6,10 +7,16 @@ use secrecy::{ExposeSecret, SecretString};
 
 /// Creates a new S3 client.
```
junhaoliao, line 8: shall we clarify the client created here is not thread-safe? 

https://docs.aws.amazon.com/sdk-for-rust/latest/dg/concurrency.html
> The AWS SDK for Rust doesn't provide concurrency control

components/clp-rust-utils/src/s3/client.rs
```diff
@@ -23,7 +30,8 @@ pub fn create_new_client(
         "clp-user",
```
junhaoliao: i feel this `provider_name` should be named as `clp-credentials-provider` or `clp-provider` to avoid confusion

junhaoliao: or simply `"manual"`?

components/clp-rust-utils/Cargo.toml
```diff
@@ -4,5 +4,6 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
+aws-config = "1.8.8"
 aws-sdk-s3 = { version = "1.106.0", features = ["behavior-version-latest"] }
```
junhaoliao: do we still need to set `"behavior-version-latest"` here?

junhaoliao: like, i believe we previously had to manually set this feature flag otherwise `Client::from_conf()` would panic? 

https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/struct.Client.html#:~:text=This%20method%20will%20panic%20if%20no%20BehaviorVersion%20is%20provided.%20If%20you%20experience%20this%20panic%2C%20set%20behavior_version%20on%20the%20Config%20or%20enable%20the%20behavior%2Dversion%2Dlatest%20Cargo%20feature.
> This method will panic if no BehaviorVersion is provided. If you experience this panic, set behavior_version on the Config _**or**_ enable the behavior-version-latest Cargo feature.

according to the docs, i guess we only need to do one of those?

components/clp-rust-utils/src/s3/client.rs
LinZhihao-723: Is the provider even used? I thought it's just an internal debugging flag

components/clp-rust-utils/src/s3/client.rs
LinZhihao-723: I don't think that's necessary. We should only document things that are thread-safe: 99% of utils will be thread unsafe. If not documented, callers should not make assumptions for thread safety

components/clp-rust-utils/Cargo.toml
LinZhihao-723: Hmmm, let me double check why I need it. We were using Client::new previously

LinZhihao-723: oh mb, yeah u're right. Should be safe to remove.

components/clp-rust-utils/src/s3/client.rs
LinZhihao-723: Looked into the panic section for `from_config`: https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/struct.Client.html#panics
It looks like these panics happen only when the given configuration is invalid. The failure should be deterministic as it doesn't involve any network operations or credential validations. Using the default config + latest behavior version should be safe. On the other hand, if this API is supposed to return a resolvable error, it should be designed to return a Result already like other AWS APIs. So I think it should be ok to not return a Result and just let it panic. Let me know if you still want this to return a Result.

components/clp-rust-utils/src/s3/client.rs
LinZhihao-723: Updated to `clp-credentials-provider`.

components/clp-rust-utils/src/s3/client.rs
junhaoliao: sure, i think it's okay to keep the return type of `create_new_client` not `Result`.

> On the other hand, if this API is supposed to return a resolvable error, it should be designed to return a Result already like other AWS APIs.

Just to confirm, if there are issues in the connection parameters like malformed credentials or region code, they eventually would be reported in `Result`s by the other Rust SDK APIs that query the AWS APIs, right? Like we can create a `Client` with malformed connection parameters but there would be issues only when we list buckets with the client?

junhaoliao (approved):

components/clp-rust-utils/src/s3/client.rs
LinZhihao-723: Yeah, those errors are not handled by the `Client` object creation.

PR #1452, opened 2025-10-20
fix(job-orchestration): Support Spider job returns a single value.

### LINKED ISSUE #1449: Compression failed for job with one task.
### Bug

Compression orchestration using Spider fails with job with only one task.

### CLP version

11d73fb

### Environment

Docker image.

### Reproduction steps

https://github.com/junhaoliao/clp/pull/4. Run on a small input log.

sitaowang1998:
# Description

## Background
Spider returns a tuple when there is multiple return values and return as-is for a single return value, following the convention of Python function. The `spider_task_manager` right now assumes that the return value is always a tuple, causing the job with only one return value to fail.

## Solution
This PR fixes #1449 by checking whether return value is a tuple and handling single return value.

# Checklist

  breaking change.

# Validation performed
* [x] Compression works for jobs with single and multiple return values.
* [ ] GitHub workflows pass.

### DIFF
--- a/.github/workflows/clp-artifact-build.yaml
+++ b/.github/workflows/clp-artifact-build.yaml
@@ -463,8 +463,10 @@ jobs:
         run: "chown $(id -u):$(id -g) -R ."
         shell: "bash"
 
-      # NOTE: We don't use the cache for scheduled runs so that they run lint:check-cpp-static-full
-      # on all files.
+      # NOTE: We don't restore the cache for `schedule` runs so that below,
+      # `lint:check-cpp-static-full` runs on all files. If we don't do this periodically,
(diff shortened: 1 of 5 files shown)


LinZhihao-723 (changes requested):
Hmmm, what if we update Spider to return a tuple for single return as well, lol

sitaowang1998:
I'm not sure about breaking change in Spider right now. @kirkrodrigues What's your opinion?

kirkrodrigues:
> I'm not sure about breaking change in Spider right now. @kirkrodrigues What's your opinion?

We haven't made any official releases of Spider, right? So it should be fine to make breaking changes if we need to (if that's what you're asking).

sitaowang1998:
Closed as we will update `spider-py` to always return tuple.

PR #1454, opened 2025-10-20
feat(clp-mcp-server): Add log viewer links to query results for displaying in LLM output.

All-less:
# Description

This PR adds integration of Log Viewer into the MCP server. Specifically, it inserts links to the Log Viewer in the result returned by search API and instructs LLM to display the log messages as clickable links. As a result, the users can click the message in the diagnosis result and be redirected to the Log Viewer. 

# Checklist

  breaking change.

# Validation performed

```
cd components/clp-mcp-server
uv run pytest
```

## Summary by CodeRabbit

* **New Features**
  * Search queries now accept optional begin/end timestamps.
  * Search results include direct links to corresponding log entries in the web interface.

* **Bug Fixes**
  * Timestamp validation only triggers when both begin and end are provided.
  * Default dataset selection uses the configured default.
  * Returned search results no longer expose internal ID fields; status messages show readable names.

* **Documentation**
  * System guidance updated to render log messages as clickable hyperlinks.

* **Tests**
  * Added test verifying link field generation in search results.

### DIFF
--- a/components/job-orchestration/job_orchestration/executor/compress/celery_compress.py
+++ b/components/job-orchestration/job_orchestration/executor/compress/celery_compress.py
@@ -19,7 +19,7 @@ def compress(
     self: Task,
     job_id: int,
     task_id: int,
-    tag_ids,
+    tag_ids: list[int],
     clp_io_config_json: str,
     paths_to_compress_json: str,
     clp_metadata_db_connection_config,
--- a/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
+++ b/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
@@ -71,7 +71,7 @@ def update_tags(
     table_prefix: str,
     dataset: Optional[str],
     archive_id: str,
-    tag_ids: List[int],
+    tag_ids: list[int],
 ) -> None:
     db_cursor.executemany(
         f"""
@@ -87,10 +87,10 @@ def update_job_metadata_and_tags(
     job_id: int,
     table_prefix: str,
     dataset: Optional[str],
-    tag_ids: List[int],
+    tag_ids: list[int],
     archive_stats: Dict[str, Any],
 ) -> None:
-    if tag_ids is not None:
+    if len(tag_ids) > 0:
         update_tags(db_cursor, table_prefix, dataset, archive_stats["id"], tag_ids)
     increment_compression_job_metadata(
         db_cursor,
@@ -312,7 +312,7 @@ def run_clp(
     logs_dir: pathlib.Path,
     job_id: int,
     task_id: int,
-    tag_ids,
+    tag_ids: list[int],
     paths_to_compress: PathsToCompress,
     sql_adapter: SQL_Adapter,
     clp_metadata_db_connection_config,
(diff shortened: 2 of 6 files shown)


components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -50,14 +54,14 @@ async def submit_query(self, query: str, begin_ts: int, end_ts: int) -> str:
...
-        if end_ts < begin_ts:
+        if begin_ts is not None and end_ts is not None and end_ts < begin_ts:
             err_msg = f"end_ts {end_ts} is smaller than begin_ts {begin_ts}."
             raise ValueError(err_msg)
 
         job_config = msgpack.packb(
             {
                 "begin_timestamp": begin_ts,
-                "dataset": None,
+                "dataset": "default",
```
junhaoliao: 1. i believe a future plan is to also support specifying non-default dataset names when creating the archive?
2. can we avoid hardcoding this? With #1405 i believe we can directly import `CLP_DEFAULT_DATASET_NAME` from `clp_py_utils.clp_config`

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -155,6 +162,13 @@ async def read_results(self, query_id: str) -> list[dict]:
         results = []
 
         async for doc in collection.find({}, limit=SEARCH_MAX_NUM_RESULTS):
+            doc["link"] = (
+                f"{self._webui_addr}/streamFile?type=json"
+                f'&streamId={doc["archive_id"]}'
+                "&dataset=default"
```
junhaoliao: ditto - we shall avoid hardcoding "default"

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -37,7 +37,11 @@ def __init__(self, clp_config: Any) -> None:
             "db": clp_config.database.name,
         }
 
-    async def submit_query(self, query: str, begin_ts: int, end_ts: int) -> str:
+        self._webui_addr = f"http://{clp_config.webui.host}:{clp_config.webui.port}"
```
junhaoliao, line 41: In production environments, deployers often host the WebUI on `localhost` and use a reverse proxy such as Nginx or Apache for load balancing and HTTPS termination. Therefore, the "_webui_addr" generated directly from the config might not always be reachable. (You may also see how the WebUI tries to use relative paths when generating such links to avoid referencing the configured webui host and port directly. )
That said, I believe the changes in this PR is a reasonable compromise of what's available in the Package. @coderabbitai please create a feature request to allow configuration of a WebUI ("public" / "deployment") address for link generations when the Package is deployed in such setups.

All-less:
> the changes look reasonable to me and the added test looks good. please see the other comments
> 
> for the tile, how about:
> 
> ```
> feat(clp-mcp-server): Add log viewer links to query results to restore log context.
> ```

In the context of MCP server, "restore log context" is too vague a description. We could use the following title to emphasize the purpose.

```
feat(clp-mcp-server): Add log viewer links to query results for displaying in LLM output.
```

PR #1465, opened 2025-10-22
feat(webui): Disable guided inputs while query running; Replace redundant "from" state with "dataset" state from native UI.

davemarco:
# Description

Sets disabled states while querying for guided components. Remove from state and use dataset state instead, since the from is not neccesary.

# Checklist

  breaking change.

# Validation performed

Components become disabled

## Summary by CodeRabbit

* **New Features**
  * Search controls (Select, Where, Order By, and query buttons) now auto-disable while queries are pending or running to prevent edits during active searches.

* **Refactor**
  * Dataset selection state moved to a shared store; several guided controls no longer bind to local state and render as uncontrolled inputs.
  * Layout tweak: select control sizing changed to use flexible growth within its container.


hoophalab (approved):
looks good to me.
Validations: the inputs are disabled before query completes

hoophalab:
We probably need to fix the dataset name overflow issue before the release. This feature is required by a user.
overflow

Since we are already modifying guided inputs in this PR, how about patching it quickly?

The following change fixes the issue -- any better ideas?
```diff
+++ b/components/webui/client/src/pages/SearchPage/SearchControls/Presto/GuidedControls/index.module.css
@@ -31,7 +31,8 @@
 }
 
 .widthSelect {
-    width: 100%;
+    width: 0;
+    flex-grow: 1;
 }
```

davemarco:
thanks - I made ur change

junhaoliao (approved):

PR #1473, opened 2025-10-22
feat(taskfile): Verify `Cargo.lock` without updating dependencies (resolves #1471).

### LINKED ISSUE #1471: CI: Verify Cargo.lock is in sync with Cargo.toml without updating dependencies
Objective
- Investigate and implement a non-mutating verification that ensures Rust Cargo.lock is in sync with Cargo.toml across the workspace, preserving deterministic CI runs and builds.

Motivation
- Avoid non-deterministic behaviour caused by implicit lockfile updates during CI, and surface drift between manifests and the lockfile early.

Requirements
- The CI step must fail if Cargo.lock is out-of-date or inconsistent with the manifest(s), without modifying any files.
- Must work across the full Rust workspace (e.g., components/clp-rust-utils, components/log-ingestor).
- Should be fast and safe to run in CI without producing build artefacts.

Approaches to evaluate (non-exhaustive)
- Use `cargo metadata --locked` at the workspace root (and/or per member) to error when the lockfile is stale.
- Use `cargo fetch --locked` to validate resolution without compiling.
- Consider `cargo check --locked -q` as a fallback if needed.
- Explore `cargo tree --locked` for dependency resolution verification without mutation.

Definition of Done
- A dedicated step in .github/workflows/clp-rust-checks.yaml performs this verification and deterministically fails when Cargo.lock is not in sync, without altering repository files.
- Developer docs (e.g., CONTRIBUTING.md or Taskfile) updated to explain how to run the check locally.

Backlinks
- PR: https://github.com/y-scope/clp/pull/1466
- Comment permalink: https://github.com/y-scope/clp/pull/1466#issuecomment-3434190282
- Related: #1463, #1464, PR #1448

Assignee
- @junhaoliao

LinZhihao-723:
# Description

This PR updates Rust's lock file check to use  `cargo fetch --locked` to fetch and resolve dependencies without updating the dependencies.
Doc: https://doc.rust-lang.org/cargo/commands/cargo-fetch.html#manifest-options

# Checklist

  breaking change.

# Validation performed

* [x] Ensure workflows pass.
* [x] Tested locally that with one dependency library's version outdated:
  - The previous check will fail as it's using `cargo update --locked`
  - The current check won't fail as it can resolve the dependencies that match `Cargo.toml` properly.

## Summary by CodeRabbit

* **Chores**
  * Updated dependency handling to use a dedicated fetch step that runs in locked mode for more reproducible builds.
  * Introduced a dedicated fetch task and a flag variable to control fetch behaviour, and updated the Rust check workflow to use this fetch step.


taskfiles/deps/lock.yaml
```diff
@@ -28,3 +28,16 @@ tasks:
     cmd: |-
       . "$HOME/.cargo/env"
       cargo update {{.CARGO_UPDATE_FLAGS}}
+
+  # Runs `cargo fetch` in the root Rust workspace directory with the specified flags.
```
junhaoliao: (please discard the comment - i did some research and found the purpose of `cargo fetch` and `cargo update` are quite different）

originally, i was trying to propose
```
  # Runs `cargo fetch` in the root Rust workspace directory with the specified flags.
  # Similar to `cargo-workspace-update`, but without updating dependencies.
```

taskfiles/deps/lock.yaml
```diff
@@ -28,3 +28,16 @@ tasks:
...
+  #
+  # @param {string} CARGO_FETCH_FLAGS The flags to pass to the `cargo fetch` command.
+  cargo-workspace-fetch:
```
junhaoliao: let's alphabetize the internal tasks

junhaoliao (commented):
the pr title lgtm

PR #1476, opened 2025-10-23
feat(clp-package)!: Add `compress-from-s3` Python and shell scripts for S3-object and S3-key-prefix compression; Restrict `compress.py` to accept local file paths only.

Eden-D-Zhang:
# Description

This PR adds a new `compress-from-3.sh` CLI script that enables compressing logs directly from S3 object storage. The script provides two modes of operation:

- s3-object: Compress specific S3 objects by providing their full URLs. Supports multiple objects that must be in the same region and bucket.
- s3-key-prefix: Compress all objects under a given S3 key prefix.

Changes:
- Added `compress-from-s3` script to handle S3 log ingestion separate from filesystem ingestion.
- Modified existing `compress` script to ingest exclusively from the filesystem.
- Modified `native` `compress.py` script to handle new S3 ingestion options.

# Checklist

  breaking change.

# Validation performed
- Verified that when `logs_input` is configured to `fs`:
  - The `compress-from-s3.sh` script fails to run.
  - `compress.sh` ingests from the filesystem successfully.
- Verified that when `logs_input` is configured to `s3`:
  - The `compress.sh` script fails to run.
  - `compress-from-s3.sh s3-object` successfully ingests one or more S3 URLs (that share a bucket and prefix).
  - `compress-from-s3.sh s3-key-prefix` successfully ingests all objects under the key prefix in an S3 URL.

### `compress-from-s3.py`:

From cli directly:
- Single object URL → success
- Multiple object URLs with a shared prefix → success
- Multiple object URLs without a shared prefix → fail
- Single prefix URL → success
- Multiple prefix URLs → fail
- Single prefix URL but it’s an object → success

From input file:
- Single object URL → success
- Multiple object URLs with a shared prefix → success
- Multiple object URLs without a shared prefix → fail
- Single prefix URL → success
- Multiple prefix URLs → fail
- Multiple prefix URLs → fail
- Single prefix URL but it’s an object → success
- Empty input file (for both object compression and key-prefix compression) → fail

Additional cases:
- From both positional and input file → fail
- Given both an input file and positional args → fail
- Given both `s3-object` and `s3-key-prefix` → fail (treated as previous)

### `compress.py` (for both `clp` and `clp-s` storage engine):

From cli directly:
- Single path → success
- Multiple path → success

From input file:
- Single path → success
- Multiple path → success
- No path → fail

Additional cases:
- From both positional and input file → fail

## Summary by CodeRabbit

* **New Features**
  * S3-based log compression with object and key-prefix modes and a new command to run it.
  * Lightweight wrapper script to invoke the S3 compression workflow.
  * New CLI option to explicitly choose FS or S3 input mode.

* **Refactor**
  * Simplified filesystem compression flow with unified FS validation and mounts.
  * Centralized S3 handling into a dedicated S3 compression path.

* **Bug Fixes**
  * Improved error handling, exit codes and more reliable cleanup of temporary files.

### DIFF
--- a/components/clp-mcp-server/clp_mcp_server/clp_mcp_server.py
+++ b/components/clp-mcp-server/clp_mcp_server/clp_mcp_server.py
@@ -2,16 +2,20 @@
 
 import ipaddress
 import logging
+import os
 import socket
 import sys
 from pathlib import Path
 
 import click
-from clp_py_utils.clp_config import CLPConfig
+from clp_py_utils.clp_config import CLPConfig, MCP_SERVER_COMPONENT_NAME
+from clp_py_utils.clp_logging import get_logger, get_logging_formatter, set_logging_level
 from clp_py_utils.core import read_yaml_config_file
 from pydantic import ValidationError
 
 from .server import create_mcp_server
+
+logger = get_logger(MCP_SERVER_COMPONENT_NAME)
 
 
 @click.command()
@@ -34,10 +38,12 @@ def main(host: str, port: int, config_path: Path) -> int:
     :param config_path: The path to server's configuration file.
     :return: Exit code (0 for success, non-zero for failure).
     """
-    logging.basicConfig(
-        level=logging.INFO, format="%(asctime)s - %(name)s - %(levelname)s - %(message)s"
-    )
-    logger = logging.getLogger(__name__)
+    # Setup logging to file
+    log_file_path = Path(os.getenv("CLP_LOGS_DIR")) / "mcp_server.log"
+    logging_file_handler = logging.FileHandler(filename=log_file_path, encoding="utf-8")
+    logging_file_handler.setFormatter(get_logging_formatter())
+    logger.addHandler(logging_file_handler)
+    set_logging_level(logger, os.getenv("CLP_LOGGING_LEVEL"))
 
     exit_code = 0
 
--- a/components/clp-mcp-server/clp_mcp_server/server/server.py
+++ b/components/clp-mcp-server/clp_mcp_server/server/server.py
@@ -4,6 +4,8 @@ from typing import Any
 
 from clp_py_utils.clp_config import CLPConfig
 from fastmcp import Context, FastMCP
+from starlette.requests import Request
+from starlette.responses import PlainTextResponse
 
 from clp_mcp_server.clp_connector import ClpConnector
 
@@ -160,5 +162,15 @@ def create_mcp_server(clp_config: CLPConfig) -> FastMCP:
             return {"Error": str(e)}
 
         return await _execute_kql_query(ctx.session_id, kql_query, begin_ts, end_ts)
+
+    @mcp.custom_route("/health", methods=["GET"])
+    async def health_check(_request: Request) -> PlainTextResponse:
+        """
+        Health check endpoint.
+
+        :param _request: An HTTP request object.
+        :return: A plain text response indicating server is healthy.
+        """
+        return PlainTextResponse("OK")
 
     return mcp
--- a/components/clp-mcp-server/clp_mcp_server/server/session_manager.py
+++ b/components/clp-mcp-server/clp_mcp_server/server/session_manager.py
@@ -83,11 +83,14 @@ class SessionState:
         """
         :param results: Log entries from the query to cache.
         :return: Forwards `SessionState.get_page_data`'s return values.
-        :return: _GET_INSTRUCTIONS_NOT_RUN_ERROR if `get_instructions` has not been called in this
-            session.
+        :return: A dictionary with the following key-value pair on failures:
+            - "Error": An error message describing the failure.
         """
         if self._is_instructions_retrieved is False:
             return self._GET_INSTRUCTIONS_NOT_RUN_ERROR.copy()
+
+        if len(results) == 0:
+            return {"Error": "No log events found matching the KQL query."}
 
         self._cached_query_result = PaginatedQueryResult(
             log_entries=results, num_items_per_page=self._num_items_per_page
--- a/components/clp-mcp-server/clp_mcp_server/server/utils.py
+++ b/components/clp-mcp-server/clp_mcp_server/server/utils.py
@@ -83,8 +83,9 @@ def format_query_results(query_results: list[dict[str, Any]]) -> list[str]:
     kv-pairs:
     - "timestamp": An integer representing the epoch timestamp in milliseconds.
     - "message": A string representing the log message.
+    - "link": A string representing the link to open the log viewer displaying the message.
 
-    The message will be formatted as `timestamp: <date string>, message: <message>`:
+    The message will be formatted as `timestamp: <date string>, message: <message>, link: <link>`.
 
     :param query_results: A list of dictionaries representing kv-pair log events.
     :return: A list of strings representing formatted log events.
@@ -105,7 +106,9 @@ def format_query_results(query_results: list[dict[str, Any]]) -> list[str]:
             logger.warning("Empty message attached to a log event: %s.", obj)
             continue
 
-        formatted_log_events.append(f"timestamp: {timestamp_str}, message: {message}")
+        link = obj["link"]
+
+        formatted_log_events.append(f"timestamp: {timestamp_str}, message: {message}, link: {link}")
 
     return formatted_log_events
 
--- a/components/clp-mcp-server/tests/server/test_utils.py
+++ b/components/clp-mcp-server/tests/server/test_utils.py
@@ -13,6 +13,8 @@ from clp_mcp_server.server.utils import (
 class TestUtils:
     """Test suite for utility functions."""
 
+    LINK = "http://localhost:4000/"
+
     # Error Messages:
     INVALID_DATE_STRING_ERROR = "Invalid date string"
     INVALID_DATE_STRING_FORMAT_ERROR = "Timestamp must end with 'Z' to indicate UTC."
@@ -24,20 +26,23 @@ class TestUtils:
         {
             "timestamp": None,
             "message": '{"message":"Log with None timestamp"}\n',
+            "link": LINK
         },
         {
             "timestamp": "1729267200000",  # str instead of int
             "message": '{"message":"Log with str timestamp"}\n',
+            "link": LINK
         },
         {
             "timestamp": 1729267200000.0,  # float instead of int
             "message": '{"message":"Log with float timestamp"}\n',
+            "link": LINK
         },
     ]
     EXPECTED_INVALID_TYPE = [
-        'timestamp: N/A, message: {"message":"Log with None timestamp"}\n',
-        'timestamp: N/A, message: {"message":"Log with str timestamp"}\n',
-        'timestamp: N/A, message: {"message":"Log with float timestamp"}\n',
+        f'timestamp: N/A, message: {{"message":"Log with None timestamp"}}\n, link: {LINK}',
+        f'timestamp: N/A, message: {{"message":"Log with str timestamp"}}\n, link: {LINK}',
+        f'timestamp: N/A, message: {{"message":"Log with float timestamp"}}\n, link: {LINK}'
     ]
 
     # Test case: invalid timestamp values.
@@ -45,33 +50,44 @@ class TestUtils:
         {
             "timestamp": 9999999999999999,
             "message": '{"message":"Log with overflow timestamp"}\n',
+            "link": LINK
         },
         {
             "timestamp": -9999999999999999,
             "message": '{"message":"Log with negative overflow timestamp"}\n',
+            "link": LINK
         },
     ]
     EXPECTED_INVALID_VALUE = [
-        'timestamp: N/A, message: {"message":"Log with overflow timestamp"}\n',
-        'timestamp: N/A, message: {"message":"Log with negative overflow timestamp"}\n',
+        (
+            f'timestamp: N/A, message: {{"message":"Log with overflow timestamp"}}\n,'
+            f' link: {LINK}'
+        ),
+        (
+            f'timestamp: N/A, message: {{"message":"Log with negative overflow timestamp"}}\n,'
+            f' link: {LINK}'
+        )
     ]
 
     # Test case: missing timestamp and message fields.
     MISSING_TIMESTAMP_AND_MESSAGE_ENTRY = [
         {
             "_id": "test001",
+            "link": LINK
         },
         {
             "_id": "test002",
             "message": '{"message":"Log with no timestamp"}\n',
+            "link": LINK
         },
         {
             "_id": "test003",
             "timestamp": 0,
+            "link": LINK
         }
     ]
     EXPECTED_MISSING_TIMESTAMP_AND_MESSAGE = [
-        'timestamp: N/A, message: {"message":"Log with no timestamp"}\n',
+        f'timestamp: N/A, message: {{"message":"Log with no timestamp"}}\n, link: {LINK}',
     ]
 
     # Testing basic functionality.
@@ -83,6 +99,7 @@ class TestUtils:
             "orig_file_path": "/var/log/app.log",
             "archive_id": "abc123",
             "log_event_ix": 99,
+            "link": LINK
         },
         {
             "_id": "test001",
@@ -91,6 +108,7 @@ class TestUtils:
             "orig_file_path": "/var/log/app.log",
             "archive_id": "abc123",
             "log_event_ix": 100,
+            "link": LINK
         },
         {
             "_id": "test002",
@@ -102,6 +120,7 @@ class TestUtils:
             "orig_file_path": "/var/log/app.log",
             "archive_id": "abc124",
             "log_event_ix": 101,
+            "link": LINK
         },
         {
             "_id": "test003",
@@ -113,6 +132,13 @@ class TestUtils:
             "orig_file_path": "/var/log/app.log",
             "archive_id": "abc125",
             "log_event_ix": 102,
+            "link": (
+                "http://localhost:4000/streamFile"
+                "?dataset=default"
+                '&type=json'
+                "&streamId=abc125"
+                "&logEventIdx=102"
+            ),
         },
     ]
 
@@ -120,22 +146,28 @@ class TestUtils:
         (
             'timestamp: 2024-10-18T16:00:00.123Z, message: '
             '{"ts":1729267200123,"pid":1234,"tid":5678,'
-            '"message":"Log with millisecond precision"}\n'
+            '"message":"Log with millisecond precision"}\n, '
+            "link: http://localhost:4000/streamFile"
+            "?dataset=default"
+            '&type=json'
+            "&streamId=abc125"
+            "&logEventIdx=102"
+            
         ),
         (
-            'timestamp: 2024-10-18T16:00:00.000Z, message: '
-            '{"ts":1729267200000,"pid":1234,"tid":5678,'
-            '"message":"Log with zero milliseconds"}\n'
+            f'timestamp: 2024-10-18T16:00:00.000Z, message: '
+            f'{{"ts":1729267200000,"pid":1234,"tid":5678,'
+            f'"message":"Log with zero milliseconds"}}\n, link: {LINK}'
         ),
         (
-            'timestamp: 1970-01-01T00:00:00.000Z, message: '
-            '{"ts":0,"pid":null,"tid":null,'
-            '"message":"Log at epoch zero"}\n'
+            f'timestamp: 1970-01-01T00:00:00.000Z, message: '
+            f'{{"ts":0,"pid":null,"tid":null,'
+            f'"message":"Log at epoch zero"}}\n, link: {LINK}'
         ),
         (
-            'timestamp: N/A, message: '
-            '{"pid":null,"tid":null,'
-            '"message":"Log at epoch none"}\n'
+            f'timestamp: N/A, message: '
+            f'{{"pid":null,"tid":null,'
+            f'"message":"Log at epoch none"}}\n, link: {LINK}'
         ),
     ]
 
--- a/components/clp-mcp-server/tests/test_session_manager.py
+++ b/components/clp-mcp-server/tests/test_session_manager.py
@@ -27,6 +27,7 @@ class TestConstants:
     # Number of logs tested in unit test and its expected page counts
     EXPECTED_NUM_LOG_ENTRIES = 25
     EXPECTED_NUM_PAGES = 3
+    EMPTY_LOG_ENTRIES = 0
 
     # 0.5 second for fast expiration tests
     FAST_SESSION_TTL_SECONDS = 0.5
@@ -41,6 +42,7 @@ class TestConstants:
     EXCEEDS_MAX_CACHED_RESULTS_ERR = "exceeds maximum allowed cached results"
     INVALID_NUM_ITEMS_PER_PAGE_ERR = "must be a positive integer"
     GET_INSTRUCTIONS_NOT_CALLED_ERR = "Please call `get_instructions()`"
+    NO_RESULTS_FOUND_IN_KQL_QUERY_ERR = "No log events found matching the KQL query."
     NO_PREVIOUS_PAGINATED_RESPONSE_ERR = "No previous paginated response in this session."
     PAGE_INDEX_OUT_OF_BOUNDS_ERR = "Page index is out of bounds."
 
@@ -115,6 +117,28 @@ class TestPaginatedQueryResult:
 class TestSessionState:
     """Unit tests for SessionState class."""
 
+    def test_error_handling(self) -> None:
+        """Validates error handling of `SessionState`."""
+        session = SessionState(
+            _num_items_per_page=TestConstants.ITEMS_PER_PAGE,
+            _session_id=TestConstants.TEST_SESSION_ID,
+            _session_ttl_seconds=TestConstants.SESSION_TTL_SECONDS
+        )
+
+        first_page = session.cache_query_result_and_get_first_page(
+            TestConstants.create_log_messages(TestConstants.EXPECTED_NUM_LOG_ENTRIES)
+        )
+
+        assert TestConstants.GET_INSTRUCTIONS_NOT_CALLED_ERR in first_page["Error"]
+        page_data = session.get_page_data(1)
+        assert TestConstants.GET_INSTRUCTIONS_NOT_CALLED_ERR in page_data["Error"]
+
+        _ = session.get_instructions()
+        first_page = session.cache_query_result_and_get_first_page(
+            TestConstants.create_log_messages(TestConstants.EMPTY_LOG_ENTRIES)
+        )
+        assert TestConstants.NO_RESULTS_FOUND_IN_KQL_QUERY_ERR in first_page["Error"]
+
     def test_get_page_data(self) -> None:
         """Validates pagination functionality is respecting the defined dictionary format."""
         session = SessionState(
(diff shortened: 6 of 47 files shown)


components/clp-package-utils/clp_package_utils/scripts/compress.py
```diff
@@ -188,10 +168,20 @@ def main(argv):
...
-    except:
+    except Exception:
         logger.exception("Failed to load config.")
         return -1
 
+    # Validate logs_input type is FS
+    if clp_config.logs_input.type != StorageType.FS:
+        logger.error(
+            "Filesystem compression requires logs_input.type to be '%s', but configured type is '%s'. "
+            "For S3 compression, use compress-from-s3.sh instead.",
```
LinZhihao-723:
```suggestion
            "Filesystem compression expects `logs_input.type` to be `%s`, but `%s` is found. For S3"
            " compression, use `compress-from-s3.sh` instead.",
```
- We should use \` instead ' for symbols.
- Space should start in the new line.

components/clp-package-utils/clp_package_utils/scripts/compress.py
```diff
@@ -257,16 +240,25 @@ def main(argv):
...
+        try:
+            generated_config_path_on_host.unlink()
+        except Exception:
+            logger.debug(
+                "Failed to remove generated config file: %s", generated_config_path_on_host
+            )
+        try:
+            container_logs_list_path.unlink()
+        except Exception:
+            logger.debug("Failed to remove logs list file: %s", container_logs_list_path)
```
LinZhihao-723: Do we need these debug logs?

components/clp-package-utils/clp_package_utils/scripts/compress.py
```diff
@@ -30,43 +30,40 @@
...
-        with open(container_logs_list_path, "w") as container_logs_list_file:
-            container_logs_list_file.write(f"{parsed_args.paths[0]}\n")
-
-    else:
-        raise ValueError(f"Unsupported input type: {input_type}.")
+    """
+    Generates logs list file for filesystem input.
+
+    :param container_logs_list_path: Path to write logs list
+    :param parsed_args: Parsed command-line arguments
```
LinZhihao-723:
```suggestion
    :param container_logs_list_path: Path to write logs list.
    :param parsed_args: Parsed command-line arguments.
```
Pleiase don't forget adding period for complete docstring lines.

components/clp-package-utils/clp_package_utils/scripts/compress.py
```diff
@@ -30,43 +30,40 @@
...
+                        resolved_path.anchor
+                    )
+                    container_logs_list_file.write(f"{mounted_path}\n")
+
+        for path in parsed_args.paths:
+            resolved_path = pathlib.Path(path).resolve()
+            mounted_path = CONTAINER_INPUT_LOGS_ROOT_DIR / resolved_path.relative_to(
+                resolved_path.anchor
+            )
+            container_logs_list_file.write(f"{mounted_path}\n")
```
LinZhihao-723:
```suggestion
        if host_logs_list_path is None:
            for path in parsed_args.paths:
                resolved_path = pathlib.Path(path).resolve()
                mounted_path = CONTAINER_INPUT_LOGS_ROOT_DIR / resolved_path.relative_to(
                    resolved_path.anchor
                )
                container_logs_list_file.write(f"{mounted_path}\n")
            return

        with open(host_logs_list_path, "r") as host_logs_list_file:
            for line in host_logs_list_file:
                stripped_path_str = line.rstrip()
                if "" == stripped_path_str:
                    # Skip empty paths
                    continue
                resolved_path = pathlib.Path(stripped_path_str).resolve()
                mounted_path = CONTAINER_INPUT_LOGS_ROOT_DIR / resolved_path.relative_to(
                    resolved_path.anchor
                )
                container_logs_list_file.write(f"{mounted_path}\n")
```
We should rewrite it this way:
- This way leads to less indentation-level. We should always try to reduce indentation levels.
- Since host logs list and positional paths are mutually exclusive options, we can do an early return.

components/clp-package-utils/clp_package_utils/scripts/compress.py
```diff
@@ -257,16 +240,25 @@ def main(argv):
...
+        return ret_code
+    finally:
+        try:
+            generated_config_path_on_host.unlink()
+        except Exception:
+            logger.debug(
+                "Failed to remove generated config file: %s", generated_config_path_on_host
+            )
+        try:
+            container_logs_list_path.unlink()
```
LinZhihao-723: Seems like we didn't remove this file in previous (and technically we also don't need to). Shall we just keep this file untouched in this PR? @kirkrodrigues What do u think?

LinZhihao-723 (changes requested):
Reviewed host `compress.py`.

components/clp-package-utils/clp_package_utils/scripts/compress.py
kirkrodrigues: If the compression job succeeds, I think it's fine to remove. If it doesn't, then it's good to keep it as a log.

LinZhihao-723: @Eden-D-Zhang can u do a conditional removal?

components/clp-package-utils/clp_package_utils/scripts/compress_from_s3.py
```diff
@@ -0,0 +1,299 @@
...
+from clp_py_utils.clp_config import (
+    CLP_DB_PASS_ENV_VAR_NAME,
+    CLP_DB_USER_ENV_VAR_NAME,
+    CLP_DEFAULT_DATASET_NAME,
+    StorageEngine,
+    StorageType,
+)
+
+from clp_package_utils.general import (
+    CLP_DEFAULT_CONFIG_FILE_RELATIVE_PATH,
```
LinZhihao-723: This is not from `clp_package_utils.general`. It should be from `clp_py_utils.clp_config`, right?

components/clp-package-utils/clp_package_utils/scripts/compress_from_s3.py
```diff
@@ -0,0 +1,299 @@
...
+        validate_and_load_db_credentials_file(clp_config, clp_home, False)
+    except Exception:
+        logger.exception("Failed to load config.")
+        return -1
+
+    # Validate logs_input type is S3
+    if clp_config.logs_input.type != StorageType.S3:
+        logger.error(
+            "S3 compression requires logs_input.type to be '%s', but configured type is '%s'. "
+            "Please update your clp-config.yml.",
```
LinZhihao-723:
```suggestion
            "S3 compression expects `logs_input.type` to be `%s`, but `%s` is found. Please update"
            " `clp-config.yml`.",
```
To make it consistent with my previous comment.

components/clp-package-utils/clp_package_utils/scripts/compress.py
Eden-D-Zhang: Was a coderabbit suggestion

components/clp-package-utils/clp_package_utils/scripts/compress.py
```diff
@@ -30,43 +30,40 @@
...
-                container_logs_list_file.write(f"{mounted_path}\n")
-
-    elif InputType.S3 == input_type:
-        with open(container_logs_list_path, "w") as container_logs_list_file:
-            container_logs_list_file.write(f"{parsed_args.paths[0]}\n")
-
-    else:
-        raise ValueError(f"Unsupported input type: {input_type}.")
+    """
+    Generates logs list file for filesystem input.
```
LinZhihao-723:
```suggestion
    Generates logs list file for the native compression script.
```

components/clp-package-utils/clp_package_utils/scripts/compress_from_s3.py
```diff
@@ -0,0 +1,299 @@
...
+    subcommand: str,
+    container_url_list_path: pathlib.Path,
+    parsed_args: argparse.Namespace,
+) -> None:
+    """
+    Generates URL list file for native script.
+
+    :param subcommand: 's3-object' or 's3-key-prefix'
+    :param container_url_list_path: Path to write URL list
+    :param parsed_args: Parsed command-line arguments
```
LinZhihao-723:
```suggestion
    Generates URL list file for the native compression script.

    :param subcommand: `s3-object` or `s3-key-prefix`.
    :param container_url_list_path: Path to write URL list.
    :param parsed_args: Parsed command-line arguments.
```
Similarly:
- Don't forget the period.
- Use \`, not '

components/clp-package-utils/clp_package_utils/scripts/compress_from_s3.py
```diff
@@ -0,0 +1,299 @@
...
+        if parsed_args.inputs_from is not None:
+            with open(parsed_args.inputs_from, "r") as input_file:
+                for line in input_file:
+                    stripped_url = line.strip()
+                    if "" == stripped_url:
+                        continue
+                    url_list_file.write(f"{stripped_url}\n")
+        else:
+            for url in parsed_args.inputs:
+                url_list_file.write(f"{url}\n")
```
LinZhihao-723:
```suggestion
        if parsed_args.inputs_from is None:
            for url in parsed_args.inputs:
                url_list_file.write(f"{url}\n")
            return

        with open(parsed_args.inputs_from, "r") as input_file:
            for line in input_file:
                stripped_url = line.rstrip()
                if "" == stripped_url:
                    continue
                url_list_file.write(f"{stripped_url}\n")
```

components/clp-package-utils/clp_package_utils/scripts/compress_from_s3.py
```diff
@@ -0,0 +1,299 @@
...
+    url_list_path: pathlib.Path,
+) -> List[str]:
+    """
+    Generates command to run native compress script.
+
+    :param parsed_args: Parsed arguments
+    :param dataset: Dataset name
+    :param config_path: Path to config file (in container)
+    :param url_list_path: Path to URL list file (in container)
+    :return: Command list
```
LinZhihao-723:
```suggestion
    Generates command to run the native compression script.

    :param parsed_args:
    :param dataset:
    :param config_path: Path to the config file (in the container).
    :param url_list_path: Path to URL list file (in the container).
    :return: The generated command.
```

components/clp-package-utils/clp_package_utils/scripts/compress_from_s3.py
```diff
@@ -0,0 +1,299 @@
...
+
+    return compress_cmd
+
+
+def _validate_s3_object_args(
+    parsed_args: argparse.Namespace,
+    args_parser: argparse.ArgumentParser,
+) -> None:
+    """
+    Validates s3-object subcommand arguments.
```
LinZhihao-723:
```suggestion
    Validates `s3-object` subcommand arguments.

    :param parsed_args:
    :param args_parser:
```

components/clp-package-utils/clp_package_utils/scripts/compress_from_s3.py
```diff
@@ -0,0 +1,299 @@
...
+    if len(parsed_args.inputs) > 0 and parsed_args.inputs_from is not None:
+        args_parser.error("URLs cannot be specified on the command line AND through a file.")
+
+
+def _validate_s3_key_prefix_args(
+    parsed_args: argparse.Namespace,
+    args_parser: argparse.ArgumentParser,
+) -> None:
+    """
+    Validates s3-key-prefix subcommand arguments.
```
LinZhihao-723:
```suggestion
    Validates `s3-key-prefix` subcommand arguments.

    :param parsed_args:
    :param args_parser:
```

components/clp-package-utils/clp_package_utils/scripts/compress_from_s3.py
```diff
@@ -0,0 +1,299 @@
...
+    object_parser.add_argument("inputs", metavar="URL", nargs="*", help="S3 object URLs")
+    object_parser.add_argument(
+        "--inputs-from", type=str, help="File containing S3 object URLs (one per line)"
+    )
+
+    prefix_parser = subparsers.add_parser(
+        "s3-key-prefix", help="Compress all objects under S3 key prefix"
+    )
+    prefix_parser.add_argument("inputs", metavar="URL", nargs="*", help="S3 prefix URL")
+    prefix_parser.add_argument("--inputs-from", type=str, help="File containing S3 prefix URL")
```
LinZhihao-723:
```suggestion
    object_parser = subparsers.add_parser(
        "s3-object", help="Compress specific S3 objects identified by their full URLs."
    )
    object_parser.add_argument("inputs", metavar="URL", nargs="*", help="S3 object URLs.")
    object_parser.add_argument(
        "--inputs-from", type=str, help="A file containing all S3 object URLs to compress."
    )

    prefix_parser = subparsers.add_parser(
        "s3-key-prefix", help="Compress all S3 objects under the key prefix."
    )
    prefix_parser.add_argument("inputs", metavar="URL", nargs="*", help="S3 prefix URL.")
    prefix_parser.add_argument(
        "--inputs-from", type=str, help="A file containing S3 key prefix to compress."
    )
```
- Polished the help notes a bit.
- Don't forget the period, lol

components/clp-package-utils/clp_package_utils/scripts/compress_from_s3.py
```diff
@@ -0,0 +1,299 @@
...
+            logger.warning(
+                "`--timestamp-key` not specified. Events will not have assigned timestamps and can "
+                "only be searched from the command line without a timestamp filter."
+            )
+    else:
+        logger.error(
+            f"S3 compression requires storage engine {StorageEngine.CLP_S}, "
+            f"but configured engine is {storage_engine}."
+        )
+        return -1
```
LinZhihao-723:
```suggestion
    if StorageEngine.CLP_S != storage_engine:
        logger.error(
            f"S3 compression requires storage engine {StorageEngine.CLP_S}, but configured engine"
            f" is {storage_engine}."
        )
        return -1

    # TODO: The following dataset validation is duplicated in `compress.py`. We should extract it
    # into a common utility function.
    dataset = CLP_DEFAULT_DATASET_NAME if dataset is None else dataset
    try:
        clp_db_connection_params = clp_config.database.get_clp_connection_params_and_type(True)
        validate_dataset_name(clp_db_connection_params["table_prefix"], dataset)
    except Exception as e:
        logger.error(e)
        return -1

    if parsed_args.timestamp_key is None:
        logger.warning(
            "`--timestamp-key` not specified. Events will not have assigned timestamps and can "
            "only be searched from the command line without a timestamp filter."
        )
```
- We don't need if-else here, right?
- Adding a TODO to track code duplication.

components/clp-package-utils/clp_package_utils/scripts/compress_from_s3.py
```diff
@@ -0,0 +1,299 @@
...
+
+    if parsed_args.subcommand == "s3-object":
+        _validate_s3_object_args(parsed_args, args_parser)
+    else:
+        _validate_s3_key_prefix_args(parsed_args, args_parser)
+
+    container_name = generate_container_name(str(JobType.COMPRESSION))
```
LinZhihao-723, line 269:
```suggestion
    # TODO: The following container setup code is duplicated in `compress.py`. We should extract it
    # into a common utility function.
    container_name = generate_container_name(str(JobType.COMPRESSION))
```
Similarly

components/clp-package-utils/clp_package_utils/scripts/compress_from_s3.py
```diff
@@ -0,0 +1,299 @@
...
+
+    object_parser = subparsers.add_parser("s3-object", help="Compress specific S3 objects")
+    object_parser.add_argument("inputs", metavar="URL", nargs="*", help="S3 object URLs")
+    object_parser.add_argument(
+        "--inputs-from", type=str, help="File containing S3 object URLs (one per line)"
+    )
+
+    prefix_parser = subparsers.add_parser(
+        "s3-key-prefix", help="Compress all objects under S3 key prefix"
+    )
```
LinZhihao-723: Since `"s3-object"` and `"s3-key-prefix"` are used in many places, and even in `native.compress.py`, I think we should make a constant in `general.py`. How about:
- `S3_OBJECT_COMPRESSION`
- `S3_KEY_PREFIX_COMPRESSION`

LinZhihao-723 (changes requested):
Reviewed `compress_form_s3.py`.

components/clp-package-utils/clp_package_utils/scripts/compress.py
LinZhihao-723: Prefer to defer it in future refactoring PR. We need a refactoring PR anyway to deduplicate two host-level compression scripts.

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -153,41 +156,152 @@ def _generate_clp_io_config(
...
+    return region_code, bucket, key_prefix
+
+
+def _get_aws_authentication_from_config(clp_config: CLPConfig) -> AwsAuthentication:
+    """
+    Get AWS authentication configuration.
+
+    :param clp_config: CLPConfig object.
+    :return: AwsAuthentication object.
+    :raise ValueError: If no authentication configured.
```
LinZhihao-723:
```suggestion
    Gets AWS authentication configuration.

    :param clp_config:
    :return: The AWS authentication configuration extracted from the CLP config.
    :raise ValueError: If no AWS authentication provided in `clp_config`.
```
- `Get` -> `Gets`
- `CLPConfig object` and `AwsAuthentication object` are meaningless: the type is already annotated in the function signature.

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -153,41 +156,152 @@ def _generate_clp_io_config(
...
+    """
+    if hasattr(clp_config, "logs_input") and hasattr(clp_config.logs_input, "type"):
+        if StorageType.S3 == clp_config.logs_input.type:
+            return clp_config.logs_input.aws_authentication
+
+    raise ValueError("No AWS authentication configured for logs_input.")
```
LinZhihao-723: However, iiuc, `.logs_input.type` must exist, as Pydantic will be responsible for input validation, right?

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -212,6 +326,13 @@ def main(argv):
         default=None,
         help="The dataset that the archives belong to.",
     )
+    args_parser.add_argument(
+        "--source",
```
LinZhihao-723: How about `input-type`? To match the `type` attribute in the config?

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -153,41 +156,152 @@ def _generate_clp_io_config(
...
+                    region_code=region_code,
+                    bucket=bucket,
+                    key_prefix=key_prefix,
+                    keys=None,
+                    aws_authentication=aws_authentication,
+                    timestamp_key=parsed_args.timestamp_key,
+                )
     else:
-        raise ValueError(f"Unsupported input type: {input_type}")
+        raise ValueError(f"Unsupported source type: {source}")
```
LinZhihao-723: How about rewriting in this way:
```python
    elif source != "s3":
        raise ValueError(f"Unsupported input type: `{source}`.")

    # Handle S3 inputs
    if len(logs_to_compress) < 2:
        raise ValueError("No URLs given.")

    aws_authentication = _get_aws_authentication_from_config(clp_config)

    s3_compress_subcommand = logs_to_compress[0]
    urls = logs_to_compress[1:]

    if s3_compress_subcommand == "s3-object":
        region_code, bucket, key_prefix, keys = _parse_and_validate_s3_object_urls(urls)
        return S3InputConfig(
            dataset=parsed_args.dataset,
            region_code=region_code,
            bucket=bucket,
            key_prefix=key_prefix,
            keys=keys,
            aws_authentication=aws_authentication,
            timestamp_key=parsed_args.timestamp_key,
        )
    elif s3_compress_subcommand == "s3-key-prefix":
        if len(urls) != 1:
            raise ValueError(f"s3-key-prefix requires exactly one URL, got {len(urls)}")
        region_code, bucket, key_prefix = _parse_s3_key_prefix_url(urls[0])
        return S3InputConfig(
            dataset=parsed_args.dataset,
            region_code=region_code,
            bucket=bucket,
            key_prefix=key_prefix,
            keys=None,
            aws_authentication=aws_authentication,
            timestamp_key=parsed_args.timestamp_key,
        )
    else:
        raise ValueError(f"Unsupported S3 compress subcommand: `{s3_compress_subcommand}`.")
```
- Check `len(logs_to_compress) < 2` since the list size should be at least 2.
- `subcommand` -> `s3_compress_subcommand`.
- With less indentation.

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -153,41 +156,152 @@ def _generate_clp_io_config(
...
 
 
 def _get_logs_to_compress(logs_list_path: pathlib.Path) -> List[str]:
-    # Read logs from the input file
+    """
+    Read logs or URLs from the input file.
```
LinZhihao-723:
```suggestion
    Reads logs or URLs from the input file.
```

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -153,41 +156,152 @@ def _generate_clp_io_config(
...
+
+    :param logs_list_path: Path to the list file.
```
LinZhihao-723:
```suggestion
    :param logs_list_path:
```

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -153,41 +156,152 @@ def _generate_clp_io_config(
...
+        keys = None
+    else:
+        key_prefix = os.path.commonprefix(keys)
+        if not key_prefix:
+            raise ValueError("No common prefix found among object keys")
+
+    return first_region, first_bucket, key_prefix, keys
+
+
+def _parse_s3_key_prefix_url(url: str) -> Tuple[str, str, str]:
```
LinZhihao-723: Do we need this? `parse_s3_url` should throw a similar error already, right?

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -153,41 +156,152 @@ def _generate_clp_io_config(
...
+    """
+    Parse single S3 URL for key-prefix mode.
+
+    :param url: S3 URL.
+    :return: Tuple of (region_code, bucket, key_prefix).
+    :raise ValueError: If parsing fails.
```
LinZhihao-723: fyi this docstring doesn't match our guideline either. Here's how I would write it
```suggestion
    Parses the given input as an S3 key prefix URL.

    :param url:
    :return: A tuple containing:
        - The region code.
        - The bucket.
        - The key prefix.
    :raise ValueError: If `parse_s3_url` fails.
```
Notice that this should be the correct way to document a tuple return.

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -153,41 +156,152 @@ def _generate_clp_io_config(
...
+    Parse and validate S3 object URLs.
+
+    Validates:
+    - All URLs have same region and bucket.
+    - No duplicate keys.
+    - For multiple URLs: non-empty common prefix exists.
+
+    :param urls: List of S3 URLs
+    :return: Tuple of (region_code, bucket, key_prefix, keys).
+             - For single URL: key_prefix is the full key, keys is None (more efficient).
```
LinZhihao-723: This is not efficient, but in fact, it will be confusing:
- It is not efficient as the compression scheduler will consider this as a prefix, and thus fall into the prefix-based ingestion.
- It implicitly changes key input to key prefix input, and thus leads to different handling on both the submission side and the compression scheduler side.
- It makes this function more complicated: now users need to understand why it could return a `None` list but still valid.

Therefore, I believe we should return a list of a single key.

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -153,41 +156,152 @@ def _generate_clp_io_config(
...
+             - For multiple URLs: key_prefix is the common prefix, keys are full keys.
+    :raises ValueError: If validation fails.
+    """
```
LinZhihao-723, line 240:
```suggestion
def _parse_and_validate_s3_object_urls(
    urls: List[str],
) -> Tuple[str, str, str, List[str]]:
    """
    Parses and validates S3 object URLs.

    The validation will ensure:
    - All URLs have the same region and bucket.
    - No duplicate keys among the URLs.
    - The URLs share a non-empty common prefix.

    :param urls: List of S3 URLs
    :return: A tuple containing:
        - The region code.
        - The bucket.
        - The common key prefix.
        - The list of keys.
    :raises ValueError: If the validation fails.
    """
```

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -153,41 +156,152 @@ def _generate_clp_io_config(
...
+    if len(urls) == 0:
+        raise ValueError("No URLs provided")
```
LinZhihao-723:
```suggestion
        raise ValueError("No URLs provided.")
```

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -153,41 +156,152 @@ def _generate_clp_io_config(
...
+
+    parsed_urls = []
+    for url in urls:
+        try:
+            region_code, bucket, key = parse_s3_url(url)
+            parsed_urls.append((region_code, bucket, key))
+        except ValueError as e:
+            raise ValueError(f"Failed to parse URL '{url}': {e}")
```
LinZhihao-723:
```suggestion
            raise ValueError(f"Failed to parse URL `{url}`: {e}")
```

components/clp-package-utils/clp_package_utils/scripts/native/compress.py
```diff
@@ -153,41 +156,152 @@ def _generate_clp_io_config(
...
+    :param urls: List of S3 URLs
+    :return: Tuple of (region_code, bucket, key_prefix, keys).
+             - For single URL: key_prefix is the full key, keys is None (more efficient).
+             - For multiple URLs: key_prefix is the common prefix, keys are full keys.
+    :raises ValueError: If validation fails.
+    """
+    if len(urls) == 0:
+        raise ValueError("No URLs provided")
+
+    parsed_urls = []
```
LinZhihao-723: Can u rewrite this function based on my prototype implementation here: https://github.com/y-scope/clp/pull/1383/files#diff-21d979ebe62533860b1718cdd22e3259efb2b6d14730d149aef95011f8484054 for the following reasons:
- We shouldn't buffer a list of tuples for validation; the stored region codes and bucket names are wasted.
- We should avoid using a list of tuples in this way. Tuple is hard to read.
- As said, we shouldn't treat a single URL specially. We should handle single/multiple URLs using the same logic. `os.path.commonprefix` should return the key itself if the input key list only contains one element.

components/clp-package-utils/clp_package_utils/scripts/compress_from_s3.py
```diff
@@ -0,0 +1,313 @@
...
+        try:
+            generated_config_path_on_host.unlink()
+        except Exception:
+            logger.debug(
+                "Failed to remove generated config file: %s", generated_config_path_on_host
+            )
+        try:
+            container_url_list_path.unlink()
+        except Exception:
+            logger.debug("Failed to remove URL list file: %s", container_url_list_path)
```
LinZhihao-723: As discussed in previous, we should update this:
- I don't think it's needed for adding this nested catch. We could let the script throw.
- We should only unlink the list path when the compression returns on success.

PR #1414, opened 2025-10-13
fix(docker): Include `libmariadbcpp.so` (required by `spider_scheduler` and `spider_worker`) in the `clp-package` image; Set `mariadb-connector-cpp` build type to `Release` (fixes #1410).

junhaoliao:
# Description

(Pending bug fixing / code refactoring changes in #1413 )

This PR includes the `libmariadbcpp.so` shared library into the `clp-package` Docker image. The `LD_LIBRARY_PATH` in the Docker image is also updated so binaries executions will look up the clp ld library path for dynamic libraries.

Also, to avoid bloating the `clp-package` image, the build type of `mariadb-connector-cpp` is changed to `Release` from the default `RelWithDebInfo`, which reduces the `libmariadbcpp.so`'s size from `23135472 bytes (23M)` to `2329080 bytes (2.3M)` which 10x size reduction.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

## Linkage fix
1. `task docker-images:package`.
2.  ```
	junhao@ASUS-X870E:~/workspace/5-clp/tools/docker-images/clp-package$ docker run -it --rm "$(docker images --format '{{.Repository}}:{{.Tag}}' \
	  | grep 'clp-package' | head -n1)" /bin/bash
	groups: cannot find name for group ID 1000
	I have no name!@53098ce876cf:/$ which spider_scheduler
	/opt/clp/bin/spider_scheduler
	I have no name!@53098ce876cf:/$ ldd /opt/clp/bin/spider_scheduler
	        linux-vdso.so.1 (0x00007ffcab0b7000)
	        libmariadbcpp.so => /opt/clp/lib/libmariadbcpp.so (0x00007d10b2966000)
	        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007d10b2738000)
	        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007d10b2651000)
	        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007d10b2631000)
	        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007d10b2408000)
	        /lib64/ld-linux-x86-64.so.2 (0x00007d10b2cae000)
[...]

PR #1479, opened 2025-10-23
refactor(deps): Document standard rules for writing lib install tasks and apply across all existing tasks:

Bill-hbrhbr:
- Modify `CMAKE_GEN_ARGS` of each task to align with the newly established conventions.
- Ensure all CLP targets link against static export targets.
- Remove redundant checksum validation from `antlr-jar`.
- Refactor `antlr-runtime` to build using `utils:install-remote-cmake-lib`.
- Make `spider` platform check POSIX-compatible.
- Replace multi-line `echo` with heredoc (`cat <<EOF`) for clearer text generation.

# Description

Due to various `CMAKE_GEN_ARGS` issues encountered while reviewing other PRs such as #1377 and #1414, we standardize CMake lib writing once and for all to cut down on future review/development time.

One important feature is that we enforce using static library artifacts and export targets whenever possible when linking against CLP targets.

Running:
```
find . -type f -name '*.so.*' -print
```
under `clp/build/deps/cpp` returns:
```
./boost-build/boost/bin.v2/libs/filesystem/config/gcc-11/release/x86_64/threading-multi/visibility-hidden/libhas_linkflag_no_undefined.so.1.87.0
./boost-build/boost/bin.v2/libs/filesystem/build/gcc-11/release/x86_64/threading-multi/visibility-hidden/libboost_filesystem.so.1.87.0
./boost-build/boost/bin.v2/libs/context/build/gcc-11/release/x86_64/threading-multi/visibility-hidden/libboost_context.so.1.87.0
./boost-build/boost/bin.v2/libs/random/build/gcc-11/release/x86_64/threading-multi/visibility-hidden/libboost_random.so.1.87.0
./boost-build/boost/bin.v2/libs/program_options/build/gcc-11/release/x86_64/threading-multi/visibility-hidden/libboost_program_options.so.1.87.0
./boost-build/boost/bin.v2/libs/system/build/gcc-11/release/x86_64/threading-multi/visibility-hidden/libboost_system.so.1.87.0
./boost-build/boost/bin.v2/libs/process/build/gcc-11/release/x86_64/boost.process.fs-boost/threading-multi/visibility-hidden/libboost_process.so.1.87.0
./boost-build/boost/bin.v2/libs/url/build/gcc-11/release/x86_64/threading-multi/visibility-hidden/libboost_url.so.1.87.0
./boost-build/boost/bin.v2/libs/atomic/build/gcc-11/release/x86_64/threading-multi/visibility-hidden/libboost_atomic.so.1.87.0
./boost-build/boost/bin.v2/libs/date_time/build/gcc-11/release/x86_64/threading-multi/visibility-hidden/libboost_date_time.so.1.87.0
./boost-build/boost/bin.v2/libs/iostreams/build/gcc-11/release/x86_64/threading-multi/visibility-hidden/libboost_iostreams.so.1.87.0
./boost-build/boost/bin.v2/libs/regex/build/gcc-11/release/x86_64/threading-multi/visibility-hidden/libboost_regex.so.1.87.0
./zlib-build/libz.so.1.3.1
./mongocxx-build/src/bsoncxx/libbsoncxx.so.4.1.1
./mongocxx-build/src/mongocxx/libmongocxx.so.4.1.1
./mongocxx-build/_deps/mongo-c-driver-build/src/libbson/libbson2.so.2.0.0
./mongocxx-build/_deps/mongo-c-driver-build/src/libmongoc/libmongoc2.so.2.0.0
./mongocxx-install/lib/libmongocxx.so.4.1.1
./mongocxx-install/lib/libbsoncxx.so.4.1.1
./mongocxx-install/lib/libmongoc2.so.2.0.0
./mongocxx-install/lib/libbson2.so.2.0.0
./zlib-install/lib/libz.so.1.3.1
./boost-install/lib/libboost_process.so.1.87.0
./boost-install/lib/libboost_system.so.1.87.0
./boost-install/lib/libboost_atomic.so.1.87.0
./boost-install/lib/libboost_url.so.1.87.0
./boost-install/lib/libboost_program_options.so.1.87.0
./boost-install/lib/libboost_iostreams.so.1.87.0
./boost-install/lib/libboost_filesystem.so.1.87.0
./boost-install/lib/libboost_context.so.1.87.0
./boost-install/lib/libboost_date_time.so.1.87.0
./boost-install/lib/libboost_regex.so.1.87.0
./boost-install/lib/libboost_random.so.1.87.0
```

* `Boost` is a special cases handled by `yscope-dev-utils`
* There is no way to disable shared artifacts building for `mongocxx` and `zlib`
  * For `mongocxx`, we link against the static export target `mongo::mongocxx_static`
  * We do not call `find_package()` for `zlib`. It will be used by `LibArchive` when we switch it to task-based installation.

# Checklist

  breaking change.

# Validation performed

* [X] Dep tasks run successfully and help build functionally correct binaries.

## Summary by CodeRabbit

* **Documentation**
  * Added a Dependency Taskfiles README with guidelines for dependency installation, preferred installer arguments, checksum guidance and tarball extraction naming.

* **Chores**
  * Centralized and simplified dependency installation flows with standardized installer arguments and unified build defaults (Release, skip tests/docs by default); many components moved to fixed-version tarballs.

* **Compatibility**
  * Simplified dependency discovery and build selection for several libraries; removed a custom LibLZMA discovery module and related public targets/variables, and standardized core CMake target selections for consistent linking.

---
  - https://app.asana.com/0/0/1212888489993931


Bill-hbrhbr:
@coderabbitai

give me a diff of each lib install task in the current version against the files in PR #1377, diff should include comments as well.

Bill-hbrhbr:
@coderabbitai Diff should include the comments inside the tasks. not your own comments.

taskfiles/deps/main.yaml
```diff
@@ -392,26 +330,29 @@ tasks:
...
     platforms: ["linux"]
     preconditions:
       # NOTE: The MariaDB connector is only required for building Spider, which is only supported on
       # Ubuntu. `INSTALL_LAYOUT` is currently hardcoded to "DEB". To support other Linux distros:
       # - Update the precondition accordingly.
       # - Make `INSTALL_LAYOUT` configurable.
       - >-
-        source /etc/os-release && [[ "$ID" == "ubuntu" ]]
-    internal: true
+        . /etc/os-release && [ "$ID" = "ubuntu" ]
```
sitaowang1998, line 319: Could you also update the similar line in spider task? I cannot comment on unchanged line.

taskfiles/deps/README.md
```diff
@@ -0,0 +1,38 @@
...
+- Use `deps:utils:install-remote-cmake-lib` and thus `yscope-dev-utils:cmake:install-remote-tar`
+  whenever applicable.
+
+- For ones that don't apply:
+  - Include `deps:utils:init` in the `deps:` section of the task.
+  - Briefly explain why if the library is installable via CMake.
+  - Verify that each library has proper checksum validation to check if it is up-to-date.
+  - Use `<lib>-extracted` as the directory name for tarball extractions.
+
+- Avoid parsing version numbers from download URLs unless the version is used elsewhere.
```
sitaowang1998:
```suggestion
- Avoid parsing version numbers in download URLs unless the version is used elsewhere.
```

taskfiles/deps/README.md
```diff
@@ -0,0 +1,38 @@
+# Dependency Taskfiles
+
+Follow the guidelines below when writing or updating dependency installation taskfiles.
+
+- Use `deps:utils:install-remote-cmake-lib` and thus `yscope-dev-utils:cmake:install-remote-tar`
+  whenever applicable.
```
sitaowang1998: Please clearly state that we prefer `deps:utils:install-remote-cmake-lib` to `yscope-dev-utils:cmake:install-remote-tar`.

taskfiles/deps/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+- For ones that don't apply:
+  - Include `deps:utils:init` in the `deps:` section of the task.
+  - Briefly explain why if the library is installable via CMake.
+  - Verify that each library has proper checksum validation to check if it is up-to-date.
```
sitaowang1998: Maybe we should add a section about how to properly use checksum validation with `yscope-dev-utils:checksum`.

sitaowang1998 (changes requested):
The code lgtm. The doc needs some more works.

taskfiles/deps/main.yaml
sitaowang1998: Also document miscellaneous changes in the PR description.

sitaowang1998 (dismissed):
LGTM. This PR just makes me think that we need some documents in `yscope-dev-utils` to explain how to use the tasks. Some logic, e.g. checksum validation should be in task dependency, is hard to understand at first glance. Do you think it worth the effort @kirkrodrigues ?

Bill-hbrhbr:
> LGTM. This PR just makes me think that we need some documents in `yscope-dev-utils` to explain how to use the tasks. Some logic, e.g. checksum validation should be in task dependency, is hard to understand at first glance. Do you think it worth the effort @kirkrodrigues ?

@davidlion In the near future, we could add some simple docs on how-to-use the checksum stuff.

taskfiles/deps/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+- Skip unit tests, examples, docs or unused binaries to speed up the install process.
+
+- Prefer disabling lib-specific testing flags (e.g., `CATCH_BUILD_TESTING=OFF`) rather than setting
+  the generic `BUILD_TESTING` to `OFF`.
+
+- **Lastly**, while satisfying the above requirements, remove redundant flags that merely restate
+  default values found in `CMakeLists.txt` and others source CMake files, with two exceptions:
+  - Set `CMAKE_BUILD_TYPE` to `Release` unless another build type is explicitly required.
+  - Set `CMAKE_INSTALL_MESSAGE` to `LAZY` to reduce log verbosity during installation.
```
LinZhihao-723:
```suggestion
Follow the guidelines below when writing or updating dependency installation taskfiles.

* Use one of the following tasks for installation, in descending order of preference:
  * `deps:utils:install-remote-cmake-lib`
  * `yscope-dev-utils:cmake:install-remote-tar`

* For special tasks that don't apply to the rule above:
  * Include `deps:utils:init` in the `deps` section of the task.
  * Briefly explain why if the library is installable via CMake.
  * Ensure each library installation includes checksum validation using the best effort.
    * Use of `yscope-dev-utils` tasks is encouraged, as most of them include proper checksum checks.
  * Use `<lib>-extracted` as the directory name for tarball extractions.

* Avoid parsing version numbers in download URLs unless the version is used elsewhere, as URL
  formats can change over time, making it more maintainable to store and use the full URL directly
  as plain text.

## CMake Generate Arguments

Apply the following guidelines to the `CMAKE_GEN_ARGS` section of tasks that call
`deps:utils:install-remote-cmake-lib`:

* Set `CMP0074` to `NEW` whenever:
  * The component's minimum required CMake version is less than 3.27 (where `CMP0074` defaults to
    `OLD`)
  * The component depends on another via `<lib_name>_ROOT`.

* Build both shared and static libraries when possible.

* Skip unit tests, examples, docs or unused binaries to speed up the installation process.

* Prefer disabling lib-specific testing flags (e.g., `CATCH_BUILD_TESTING=OFF`) rather than setting
  the generic `BUILD_TESTING` to `OFF`.

* **Lastly**, while satisfying the above requirements, remove redundant flags that merely restate
  default values found in `CMakeLists.txt` and others source CMake files, with two exceptions:
  * Set `CMAKE_BUILD_TYPE` to `Release` unless another build type is explicitly required.
  * Set `CMAKE_INSTALL_MESSAGE` to `LAZY` to reduce log verbosity during installation.
```
- We should use `*` for bullet points.
- "install" -> "installation"
- "best effort" -> "the best effort"

taskfiles/deps/main.yaml
```diff
@@ -95,7 +91,7 @@ tasks:
     preconditions:
       # NOTE: Spider is only supported on Ubuntu at present.
       - >-
-        source /etc/os-release && [[ "$ID" == "ubuntu" ]]
+        . /etc/os-release && [ "$ID" == "ubuntu" ]
```
LinZhihao-723, line 94: Just curious, any reason not to explicitly use `source`?

taskfiles/deps/main.yaml
```diff
@@ -540,34 +479,35 @@ tasks:
...
-      - "mkdir -p '{{.INSTALL_INCLUDE_DIR}}'"
-      - "rm -f '{{.INSTALL_SYMLINK}}'"
-      - "ln -s '{{.SRC_DIR}}' '{{.INSTALL_SYMLINK}}'"
       - |-
-        echo "set(CLP_SQLITE3_SOURCE_DIRECTORY \"{{.SRC_DIR}}\")" \
-        > "{{.G_DEPS_CPP_CMAKE_SETTINGS_DIR}}/{{.LIB_NAME}}.cmake"
-        echo "set(CLP_SQLITE3_INCLUDE_DIRECTORY \"{{.INSTALL_INCLUDE_DIR}}\")" \
-        >> "{{.G_DEPS_CPP_CMAKE_SETTINGS_DIR}}/{{.LIB_NAME}}.cmake"
+        mkdir -p "{{.INSTALL_INCLUDE_DIR}}"
+        ln -sfn "{{.SRC_DIR}}" "{{.INSTALL_SYMLINK}}"
```
LinZhihao-723, line 462: * I think we prefer to copy instead of using symlink.
* If we want to use symlink, can u expend `-sfn` with the long forms like `--symbolic --force --no-derefernece`? The guideline requires the use of the full term for clarity unless it has a macOS compatibility issue.

LinZhihao-723 (changes requested):

taskfiles/deps/main.yaml
Bill-hbrhbr: ln doesn't have long options on macos. Ref: https://ss64.com/mac/ln.html

taskfiles/deps/main.yaml
Bill-hbrhbr: To make the command poxis compatible

taskfiles/deps/main.yaml
```diff
@@ -510,7 +449,7 @@ tasks:
             - "-DCMAKE_INSTALL_MESSAGE=LAZY"
             - "-DCMAKE_POLICY_DEFAULT_CMP0074=NEW"
             - "-DSPDLOG_BUILD_EXAMPLE=OFF"
-            - "-DSPDLOG_BUILD_EXAMPLE_HO=OFF"
+            - "-DSPDLOG_BUILD_SHARED=ON"
```
LinZhihao-723: Do we still need to build shared libs?

taskfiles/deps/main.yaml
LinZhihao-723: Ok, let's use short terms then.

taskfiles/deps/main.yaml
```diff
@@ -626,7 +568,6 @@ tasks:
           CMAKE_GEN_ARGS:
             - "-DCMAKE_BUILD_TYPE=Release"
             - "-DCMAKE_INSTALL_MESSAGE=LAZY"
-            - "-DZSTD_BUILD_CONTRIB=OFF"
             - "-DZSTD_BUILD_PROGRAMS=OFF"
             - "-DZSTD_BUILD_SHARED=ON"
```
LinZhihao-723: Similarly, shall we remove building for shared?

components/core/CMakeLists.txt
```diff
@@ -301,20 +297,11 @@ if(CLP_NEED_ZSTD)
     # v1.4.8 is the lowest version available in the package managers of the OSes we support.
     find_package(zstd 1.4.8 REQUIRED)
     message(STATUS "Found zstd ${zstd_VERSION}")
-    if(CLP_USE_STATIC_LIBS)
-        set(zstd_TARGET zstd::libzstd_static)
-    else()
-        set(zstd_TARGET zstd::libzstd_shared)
-    endif()
+    # TODO: replace all occurrences of `zstd_TARGET` with `zstd::libzstd`.
```
LinZhihao-723, line 300: @coderabbitai Can you create an issue for this? We should update `zstd_TARGET` to `zstd::libzstd`.
@Bill-hbrhbr For this type of TODO, we should create an issue to track it.

taskfiles/deps/main.yaml
```diff
@@ -154,94 +150,43 @@ tasks:
...
-        vars:
-          CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
-          INCLUDE_PATTERNS: ["{{.INSTALL_PREFIX}}"]
+          CMAKE_SOURCE_DIR: "runtime/Cpp"
+          LIB_NAME: "antlr4-runtime"
+          TARBALL_SHA256: "9f18272a9b32b622835a3365f850dd1063d60f5045fb1e12ce475ae6e18a35bb"
+          TARBALL_URL: "https://github.com/antlr/antlr4/archive/refs/tags/\
+                        {{.G_ANTLR_VERSION}}.tar.gz"
 
   boost:
```
LinZhihao-723, line 185: Shall we create issues to update boost installation task to:
- Take a param for whether we should build share.
- Change the extract dir name to `boost-extracted` from `boost-generated` to match the new guideline.

taskfiles/deps/main.yaml
```diff
@@ -608,7 +528,6 @@ tasks:
       - task: "utils:install-remote-cmake-lib"
         vars:
           CMAKE_GEN_ARGS:
-            - "-DBUILD_SHARED_LIBS=ON"
```
LinZhihao-723, line 606: This option is `ON` by default: https://github.com/madler/zlib/blob/5a82f71ed1dfc0bec044d9702463dbdf84ea3b71/CMakeLists.txt#L23
You will observe that in the installed dir, shared libs are also installed.

LinZhihao-723 (changes requested):
Overall doc changes lgtm.

taskfiles/deps/README.md
junhaoliao, line 1: possible to move this to the `docs/src`? if the intent of adding a README file here is to increase visibility of the docs, i believe we can change this to a simple reference to the doc page on the docs site like https://github.com/y-scope/clp/blob/main/components/webui/README.md

PR #1480, opened 2025-10-23
feat(clp-package)!: Add support for specifying a temporary directory; Remove hack in `var/data` conditional mounts (resolves #1478).

### LINKED ISSUE #1478: Remove Docker Compose mount hack for archive/staged-archive directories
## Context

Currently, the Docker Compose configuration contains a hack to avoid Docker creating directories as root on the host when mounting archive and staged-archive directories.

The issue arises because only one of `CLP_ARCHIVE_OUTPUT_DIR_HOST` and `CLP_STAGED_ARCHIVE_OUTPUT_DIR_HOST` is set at a time, but since `./var/data` on the host is mounted into the container and both variables default to directories under `./var/data`, Docker would create the unset directory on the host as root if we naively mount with a default path under `/var/data`.

## Current Hack

When one of the variables is unset, we set the target for the corresponding mount to a path that's not under `/var/data` in the container to prevent Docker from creating it on the host.

See: https://github.com/y-scope/clp/blob/docker-compose/tools/deployment/package/docker-compose.base.yaml#L267-L283

## Proposed Solution

Accept a `var/tmp` directory config from the user and generate all temporary runtime data in that directory instead of the current `var/data` directory. This would eliminate the need for the `/var/data` bind mount and remove the hack.

## References

- PR #1178
- Comment: https://github.com/y-scope/clp/pull/1178#discussion_r2455414543
- Requested by: @junhaoliao

junhaoliao:
# Description

This change adds support for a configurable temporary directory used by CLP runtime
components and removes the prior conditional mount hack around `/var/data` in the Docker Compose package
orchestration.
This change is BREAKING because now temporary runtime data is stored in the configurable tmp_directory (default `var/tmp`, mounted as `/var/tmp` in the container) instead of the `data_directory` (`/var/data`), and `WorkerConfig` now uses `tmp_directory` for temp artifacts.

Summary of changes:
- `clp_py_utils.clp_config`
  - Introduced `CLP_DEFAULT_TMP_DIRECTORY_PATH` defaulting to `var/tmp`.
  - Added `tmp_directory` to `CLPConfig` with absolute-path transformation, `validate_tmp_dir()`,
    and container translation in `transform_for_container()`.
  - Updated `WorkerConfig` to use `tmp_directory` (replacing `data_directory`) for temp files.
- Exported `CLP_TMP_DIR_HOST` env for compose.
- job_orchestration/executor/compress
  - compression_task.py: Switched temp artifacts (e.g., `*-log-paths.txt`) to be created under
    `worker_config.tmp_directory`.
- package-template
  - clp-config.yml: Documented the new `tmp_directory` option and constraints.
- Docker Compose deployment
  - docker-compose.base.yaml: Removed conditional `/var/data` mount targets; added a straightforward
    bind mount for `${CLP_TMP_DIR_HOST:-./var/tmp} -> /var/tmp` and direct mounts for archives and
    staged archives. This eliminates the `/tmp` indirection hack.
  - Resolves #1478.

# Checklist

  breaking change.

# Validation performed

```
task
cd build/clp-package

./sbin/start-clp.sh
# Observed the package started successfully

docker compose config
# Observed the `/tmp` mount hack has been removed

./sbin/compress.sh ~/samples/hive-24hr
# Observed the compression job completed successfully

ls -l var/data
# Observed no `staged-archives` directory was listed

# Loaded the WebUI http://localhost:4000 and performed a search.
# The search completed successfully with results.
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable temporary directory for runtime data with validation and container mapping.

* **Configuration Changes**
  * New tmp_directory option with default path and safety guidance.
  * Orchestration now exposes a host TMP bind for containers and simplifies archive mounts.

* **Improvements**
  * Worker and task flows now use the tmp directory consistently and create/validate it at startup.


junhaoliao:
@sitaowang1998 could you help review this before passing it onto @kirkrodrigues ?

sitaowang1998 (commented):
Just from this PR, it looks like the temporary directory is used exclusively inside the worker container, so why do we need to mount it?

junhaoliao:
> Just from this PR, it looks like the temporary directory is used exclusively inside the worker container, so why do we need to mount it?

for the background, the compression scheduler write an input files catalog to a temp location and pass the location to the clp binaries:
https://github.com/y-scope/clp/blob/6abc934920b55da2de9c05ebf565cf96e47e6a45/components/job-orchestration/job_orchestration/executor/compress/compression_task.py#L139-L153

as for whether to write to `/var/tmp` or `/tmp`, it a matter of lifetime of the data: data in `/var/tmp` survives reboots while data in `/tmp` don't (on most POSIX compliant systems. i once worked with RedHat's in-vehicle OS and their `/tmp` is persistent lol).

it's true that we don't need persistence of such catalog across service (i.e., `compression-worker`) reboots. Then `/tmp` can be used instead of `/var/tmp`. Then we don't have to map such directories on the host. @kirkrodrigues what do you think?

sitaowang1998:
> for the background, the compression scheduler write an input files catalog to a temp location and pass the location to the clp binaries:

Correct me if I'm wrong, but I thinks it is the compression worker who writes the input files catalog to a temp location and then call clp binaries to read from the created file.

kirkrodrigues:
> it's true that we don't need persistence of such catalog across service (i.e., `compression-worker`) reboots. Then `/tmp` can be used instead of `/var/tmp`. Then we don't have to map such directories on the host. @kirkrodrigues what do you think?

It's useful for debugging when things go wrong, so I'd prefer to persist it for now.

PR #1481, opened 2025-10-23
feat(clp-package): Add MCP Server config interface and Docker Compose orchestration.

20001020ycx:
# Description

Adding the MCP server as a service in docker-compose.yaml.

- Add an mcp-server service using the shared CLP Docker image.
  - Register the mcp-server service under a non-default [profile](https://docs.docker.com/compose/how-tos/profiles/). This way mcp server service doesn’t always get launched by default unless the AI user explicitly instructs to do so.
- Configure environment variables, volumes, ip and port.
  -  Variables set by user configuration are ip and port, which are defined in clp-config.yml with the following format, with the default value shown below:
```
    mcp_server:
    host: "localhost"
    port: 8000
```
- Set the service to run the MCP server using python3 -m clp_mcp_server.clp_mcp_server
- Implement health monitoring for the MCP server by exposing a GET /health endpoint (e.g., http://0.0.0.0:8000/health) and configuring Docker Compose to poll it periodically via an HTTP GET. 
  - The mechanism is supported using “custom routes” documented by FastMCP:[ https://gofastmcp.com/servers/server#custom-routes](https://gofastmcp.com/servers/server#custom-routes)

# Checklist

  breaking change.

# Validation performed

1. Ran `task docker-images:package` to build the package.
2. Ran `./start-clp.sh`
3. Confirmed all containers have started successfully.
image
4. Connected Claude Desktop agent and issued a “hello world” query, receiving the expected greeting and server status response.
Screenshot 2025-10-23 at 11 59 41 AM

## Summary by CodeRabbit

* **New Features**
  * Added an MCP server with configurable host, port and logging level, file-based logging, and a /health endpoint.
  * CLI/config and runtime wiring to enable an "mcp" profile and inject MCP environment variables after validating configuration.

* **Documentation**
  * Architecture and deployment docs updated to include the MCP server and its service relationships.

* **Chores**
  * Deployment updates: compose service, healthcheck, volumes, and added curl to image setup.

### DIFF
--- a/components/package-template/src/etc/clp-config.yml
+++ b/components/package-template/src/etc/clp-config.yml
@@ -71,6 +71,7 @@
 #mcp_server:
 #  host: "localhost"
 #  port: 8000
+#  logging_level: "INFO"
 #
 ## Where archives should be output to
 #archive_output:


components/clp-mcp-server/clp_mcp_server/clp_mcp_server.py
```diff
@@ -34,10 +38,12 @@ def main(host: str, port: int, config_path: Path) -> int:
...
-    logging.basicConfig(
-        level=logging.INFO, format="%(asctime)s - %(name)s - %(levelname)s - %(message)s"
-    )
-    logger = logging.getLogger(__name__)
+    # Setup logging to file
+    log_file_path = Path(os.getenv("CLP_LOGS_DIR")) / "mcp_server.log"
+    logging_file_handler = logging.FileHandler(filename=log_file_path, encoding="utf-8")
+    logging_file_handler.setFormatter(get_logging_formatter())
+    logger.addHandler(logging_file_handler)
+    set_logging_level(logger, os.getenv("CLP_LOGGING_LEVEL"))
```
junhaoliao, line 46: in a future PR, we should extract the `configure_logger` util from `garbage_collector` into `clp-py-utils` so we don't have to repeat these logger init steps in many different places in the pacakge:
https://github.com/y-scope/clp/blob/6abc934920b55da2de9c05ebf565cf96e47e6a45/components/job-orchestration/job_orchestration/garbage_collector/utils.py#L21-L28

@coderabbitai create an issue to track the refactor

components/clp-mcp-server/clp_mcp_server/clp_mcp_server.py
```diff
@@ -2,17 +2,21 @@
...
 
 import click
 from clp_py_utils.clp_config import CLPConfig
+from clp_py_utils.clp_logging import get_logger, get_logging_formatter, set_logging_level
 from clp_py_utils.core import read_yaml_config_file
 from pydantic import ValidationError
 
 from .server import create_mcp_server
 
+logger = get_logger("mcp_server")
```
junhaoliao: possible to import `MCP_SERVER_COMPONENT_NAME` and write this instead?
```suggestion
logger = get_logger(MCP_SERVER_COMPONENT_NAME)
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -528,6 +530,35 @@ def _set_up_env_for_webui(self, container_clp_config: CLPConfig) -> EnvVarsDict:
 
         return env_vars
 
+    def _set_up_env_for_mcp_server(self) -> EnvVarsDict:
+        """
+        Sets up environment variables for the MCP server component.
```
junhaoliao: sorry again for missing this - 
```suggestion
        Sets up environment variables and directories for the MCP server component.
```

i'll correct the inconsistencies in the docstrings in the other methods (like `_set_up_env_for_reducer`)

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -528,6 +530,35 @@ def _set_up_env_for_webui(self, container_clp_config: CLPConfig) -> EnvVarsDict:
...
+
+        :return: Dictionary of environment variables necessary to launch the component.
+        """
+        component_name = MCP_SERVER_COMPONENT_NAME
+        logger.info(f"Setting up environment for {component_name}...")
+
+        logs_dir = self._clp_config.logs_directory / component_name
+        logs_dir.mkdir(parents=True, exist_ok=True)
+
+        validate_mcp_server_config(self._clp_config, logs_dir)
```
junhaoliao, line 546: `validate_mcp_server_config()` should check if the `logs_dir` "could be a directory" instead of checking whether the path is already a directory. see the comments at `validate_mcp_server_config()`.

then this should become
```suggestion
        logs_dir = self._clp_config.logs_directory / component_name
        validate_mcp_server_config(self._clp_config, logs_dir)
        logs_dir.mkdir(parents=True, exist_ok=True)
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -528,6 +530,35 @@ def _set_up_env_for_webui(self, container_clp_config: CLPConfig) -> EnvVarsDict:
...
+        # Logging config
+        env_vars |= {
+            "CLP_MCP_LOGGING_LEVEL": self._clp_config.mcp_server.logging_level,
+        }
+
+        # Connection config
+        env_vars |= {
+            "CLP_MCP_HOST": _get_ip_from_hostname(self._clp_config.mcp_server.host),
+            "CLP_MCP_PORT": str(self._clp_config.mcp_server.port),
+        }
```
junhaoliao: let's alphabetize the groups

```suggestion
        # Connection config
        env_vars |= {
            "CLP_MCP_HOST": _get_ip_from_hostname(self._clp_config.mcp_server.host),
            "CLP_MCP_PORT": str(self._clp_config.mcp_server.port),
        }

        # Logging config
        env_vars |= {
            "CLP_MCP_LOGGING_LEVEL": self._clp_config.mcp_server.logging_level,
        }
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -627,6 +658,8 @@ def start(self) -> None:
         cmd = ["docker", "compose", "--project-name", self._project_name]
         if deployment_type == DeploymentType.BASE:
             cmd += ["--file", "docker-compose.base.yaml"]
+        if len(self._clp_config.mcp_server.model_fields_set) > 0:
```
junhaoliao: as discussed at junhaoliao/clp#3 , can we add a Boolean flag like `enabled` in the `McpServer` config model so we can explicitly control whether the MCP server is launched or not? e.g.,

clp-config.yaml
```yaml
mcp_server:
  enabled: false
  host: "localhost"
  port: 8000
```

since this is interface-related, cc @kirkrodrigues for awareness

junhaoliao: The log in the setup method should also be different if the MCP server isn't launched. we can refer to the `_set_up_env_for_garbage_collector()` for the format of the log

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -627,6 +658,8 @@ def start(self) -> None:
...
+            cmd += ["--profile", "mcp"]
```
junhaoliao, line 664: (just to show we have considered the alternative - no action item is need for this comment)

This is fine. if we (ever) add more profiles, i think an alternative can be including this in the env var dict to be returned by `_set_up_env_for_garbage_collector()`:
```
# NOTE: for this to work, we also need to move the local declaration of `env_vars` into the constructor and name it `_env_vars`.
# A helper can also be added to help append the profile name
"COMPOSE_PROFILES": f'mcp {self._env_vars.get("COMPOSE_PROFILES", "")}'
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -584,6 +585,16 @@ def validate_webui_config(
     validate_port(f"{WEBUI_COMPONENT_NAME}.port", clp_config.webui.host, clp_config.webui.port)
 
 
+def validate_mcp_server_config(clp_config: CLPConfig, logs_dir: pathlib.Path):
+    if not logs_dir.exists():
+        error_msg = f"{MCP_SERVER_COMPONENT_NAME} logs directory at {logs_dir!s} is missing."
+        raise ValueError(error_msg)
+
```
junhaoliao: instead of checking whether `logs_dir` is already a directory, can we check if the path could be a directory (so later we can mkdir at the location) like other validators? maybe we can simply use the helper `_validate_log_directory`?

docs/src/dev-docs/design-deployment-orchestration.md
junhaoliao, line 1: cc @quinntaylormitchell / @kirkrodrigues for awareness / comments

tools/deployment/package/docker-compose.yaml
```diff
@@ -125,3 +125,42 @@ services:
       "--concurrency", "${CLP_REDUCER_CONCURRENCY:-1}",
       "--upsert-interval", "${CLP_REDUCER_UPSERT_INTERVAL:-100}"
     ]
+
+  mcp-server:
+    <<: *service_defaults
+    hostname: "mcp_server"
+    profiles:
+      - "mcp"
```
junhaoliao: since there's only one profile name assigned
```suggestion
    profiles: ["mcp"]
```

tools/deployment/package/docker-compose.yaml
```diff
@@ -125,3 +125,42 @@ services:
...
+        condition: "service_completed_successfully"
+      results-cache-indices-creator:
+        condition: "service_completed_successfully"
+    command: [
+      "python3", "-u",
+      "-m", "clp_mcp_server.clp_mcp_server",
+      "--host", "mcp_server",
+      "--port", "8000",
+      "--config-path", "/etc/clp-config.yml",
+    ]
```
junhaoliao: let's follow `query-scheduler` and move `healthcheck` to the bottom of the service

junhaoliao (changes requested):
overall lgtm. i've posted all comments

for the PR title, how about:
```
feat(clp-package): Add MCP Server config interface and Docker Compose orchestration.
```

components/clp-package-utils/clp_package_utils/controller.py
20001020ycx: Just posting my thoughts of why using the above approach; however, I am also okay with what Junhao is suggesting. This is what I had in [junhaoliao#3](https://github.com/junhaoliao/clp/pull/3)
> I feel like the enable flag is redundant. If the user has specified mcp server's configuration file, we can just automatically start it for them because that's the reason why they are specifying the mcp server config file. I feel this approach is more intuitive and easy to use.

20001020ycx: btw, this might be a conflict with what you have in the doc, @rishikeshdevsot for awareness

kirkrodrigues: We use presto: null for presto (instead of a boolean). Do you want to do the same here?

junhaoliao: i agree if we want to be consistent with the Presto config

though i realize in the package template clp-config.yml it's not obvious what config fields are available for the Presto config, and i have to look up the manual / check clp_config.py to figure out. would it be possible to use `null` to disable yet we can still somehow document what config fields are available in the template clp-config.yml?

20001020ycx: So just to be clear on the implementation side:
Below means mcp server is on, as long as one of the fields is filled bythe  user
```
mcp_server:
  host: "localhost"
  port: 8006
```
Below means mcp server is off:
mcp_server: null
```

Comment (rishikeshdevsot) [reply components/clp-package-utils/clp_package_utils/controller.py]: what if no option for `mcp_server` is passed in the config? error out? or assume don't start MCP server?

Comment (20001020ycx) [reply components/clp-package-utils/clp_package_utils/controller.py]: The proposed change by Kirk is implemented, I have tested the start-clp.sh respectice following convention:
Below means mcp server is on, as long as one of the fields is filled bythe user
```
mcp_server:
  host: "localhost"
  port: 8006
```
Below means mcp server is off:
`mcp_server: null`

Comment (20001020ycx) [reply components/clp-package-utils/clp_package_utils/controller.py]: @rishikeshdevsot if there's no other objection, can you update the documentation on your side?

Comment (rishikeshdevsot) [reply components/clp-package-utils/clp_package_utils/controller.py]: > what if no option for `mcp_server` is passed in the config? error out? or assume don't start MCP server?

what about this situation?

Comment (20001020ycx) [reply components/clp-package-utils/clp_package_utils/controller.py]: Of courrse, leaving the mcp_server commented out in the clp-config.yml would also make it off. Hope this is consistent with our existing practice!

Comment (rishikeshdevsot) [reply components/clp-package-utils/clp_package_utils/controller.py]: got it, yeah I am okay with this. I'll add this change to #1456

File: components/clp-package-utils/clp_package_utils/controller.py [hunk 11]
```diff
@@ -528,6 +530,34 @@ def _set_up_env_for_webui(self, container_clp_config: CLPConfig) -> EnvVarsDict:
...
         return env_vars
 
+    def _set_up_env_for_mcp_server(self) -> EnvVarsDict:
+        """
+        Sets up environment variables and directories for the MCP server component.
+
+        :return: Dictionary of environment variables necessary to launch the component.
+        """
+        component_name = MCP_SERVER_COMPONENT_NAME
+        logger.info(f"Setting up environment for {component_name}...")
```
junhaoliao, line 543: let's refer to the garbage collector for skipping the component creation

```suggestion
        if self._clp_config.mcp_server is None:
            logger.info(f"The MCP Server is not configured, skipping {component_name} creation...")
            return EnvVarsDict()
        logger.info(f"Setting up environment for {component_name}...")
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -750,6 +782,8 @@ def _set_up_env(self) -> None:
         env_vars |= self._set_up_env_for_query_worker(num_workers)
         env_vars |= self._set_up_env_for_reducer(num_workers)
         env_vars |= self._set_up_env_for_webui(container_clp_config)
+        if self._clp_config.mcp_server is not None:
+            env_vars |= self._set_up_env_for_mcp_server()
```
junhaoliao: see above suggestion for skipping the component creation
```suggestion
        env_vars |= self._set_up_env_for_mcp_server()
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -608,6 +617,7 @@ class CLPConfig(BaseModel):
     compression_worker: CompressionWorker = CompressionWorker()
     query_worker: QueryWorker = QueryWorker()
     webui: WebUi = WebUi()
+    mcp_server: Optional[McpServer] = None
     garbage_collector: GarbageCollector = GarbageCollector()
     credentials_file_path: SerializablePath = CLP_DEFAULT_CREDENTIALS_FILE_PATH
```
junhaoliao: let's group the optional configs together

```suggestion
    garbage_collector: GarbageCollector = GarbageCollector()
    credentials_file_path: SerializablePath = CLP_DEFAULT_CREDENTIALS_FILE_PATH

    mcp_server: Optional[McpServer] = None
```

components/package-template/src/etc/clp-config.yml
```diff
@@ -68,6 +68,10 @@
 #  results_metadata_collection_name: "results-metadata"
 #  rate_limit: 1000
 #
+# mcp_server:
+#   host: "localhost"
+#   port: 8000
```
junhaoliao:
```suggestion
#mcp_server:
#  host: "localhost"
#  port: 8000
```

junhaoliao (dismissed):

components/clp-package-utils/clp_package_utils/controller.py
kirkrodrigues: > would it be possible to use null to disable yet we can still somehow document what config fields are available in the template clp-config.yml?

Yeah, I get where you're coming from and I've thought about it before. But it can be tricky to document every possible parameter for the config file (e.g., we would need to document both fs and s3 input types and parameters, every possible credential type, etc.). Maybe the more long-term option is to have some kind of config generator?

components/package-template/src/etc/clp-config.yml
```diff
@@ -68,6 +68,11 @@
 #  results_metadata_collection_name: "results-metadata"
 #  rate_limit: 1000
 #
+#mcp_server:
+#  host: "localhost"
+#  port: 8000
+#  logging_level: "INFO"
```
junhaoliao, line 74: good catch

junhaoliao (approved):

components/package-template/src/etc/clp-config.yml
20001020ycx: credit goes to rabbit

components/clp-package-utils/clp_package_utils/controller.py
junhaoliao: > leaving the mcp_server commented out in the clp-config.yml would also make it off. Hope this is consistent with our existing practice

i was fine with the approach, until i just selected all and uncommented before starting the package. then i realize the uncommented config is not consistent with the default values defined in ClpConfig. maybe it's better to just write the default value in the template clp-config.yml?

@coderabbitai : create an issue to track the inconsistency.
[at] humans : feel free to object / propose ideas. we can continue the discussion in the issue to be created by the rabbit

tools/deployment/package/docker-compose.yaml
```diff
@@ -125,3 +125,41 @@ services:
       "--concurrency", "${CLP_REDUCER_CONCURRENCY:-1}",
       "--upsert-interval", "${CLP_REDUCER_UPSERT_INTERVAL:-100}"
     ]
+
+  mcp-server:
+    <<: *service_defaults
+    hostname: "mcp_server"
+    profiles: ["mcp"]
```
junhaoliao, line 132: RFC: Standardize optional service enable/disable mechanisms in Docker Compose

We currently have three different mechanisms for optionally enabling or disabling services in our Docker Compose setup:
1. Replica count toggling (0/1)
   * Example: Garbage Collector
   * Controlled via env var
   * `deploy.replicas = 1` or `0`
2. Profiles
   * Example: MCP server
   * Enabled via `--profile=mcp`
   * Requires controller logic to inject flags at startup
3. Separate overlay compose files
   * Example: Query services (and soon compression services via spider integration)
   * Additional `docker-compose-XYZ.yaml` extending the base
   * Enabled by passing `-f docker-compose-XYZ.yaml` at runtime
   * Used for toggling **groups of services** when the composition varies significantly

Each mechanism serves a slightly different use case, but the result is fragmented enable/disable logic, increasing maintenance burden and making controller code inconsistent.

## Problem

Because we use multiple patterns to control optional service activation:

1. The controller code becomes inconsistent.
   * Replica toggling uses environment setup functions (`CLP_GARBAGE_COLLECTOR_ENABLED` inside `_set_up_env_for_garbage_collector()`).
   * Profiles require injecting CLI flags in the `start()` path (`cmd += ["--profile", "mcp"])`, which is less isolated.
2. It's harder to reason about which services support which activation pattern.
3. Adding new optional services may copy-paste the existing inconsistency.

## Proposal

Adopt (1) and (3) as the two primary supported mechanisms, and phase out (2) profiles.
* Use (1) replica count toggling for individual services that are simply enabled or disabled.
* Use (3) overlay compose files for groups of services whose compositions vary significantly depending on the configuration (e.g., different schedulers with different service sets).

Remove or migrate existing usage of (2) profiles, since it introduces controller-side CLI logic and is less maintainable. i.e., Remove the MCP server’s profile-based activation and migrate it to mechanism (1) using an environment-variable-controlled replica count (e.g., `CLP_MCP_ENABLED` -> replicas 0 or 1).

@coderabbitai create an issue

PR #1489, opened 2025-10-23
fix(clp-mcp-server): Return an error when a KQL query has no results.

20001020ycx:
# Description

When no log events match a given KQL query, the MCP server currently returns `{"Error":"Page index is out of bounds."}`, which is confusing for both end users and LLMs. This PR adds an explicit check for this scenario and includes unit tests.

# Checklist

  breaking change.

# Validation performed

1. Unit Tests: `uv run --group dev pytest tests/test_session_manager.py -vv`
2. Lint Tests: `uv run ruff check clp/components/clp-mcp-server && task lint:check-py.`
3. Manual Integration Tests: Built using `task docker-images:package` with docker compose. Connected Claude Desktop agent to the running MCP server service in docker compose. Manually instructed LLM to perform a kql that returns empty result:
Screenshot 2025-10-23 at 9 38 16 PM

## Summary by CodeRabbit

* **Bug Fixes**
  * Queries that return no results now surface a clear error: "No log events found matching the KQL query." — prompting users to broaden their search when no log events match.

* **Tests**
  * Added tests validating error propagation before instructions are provided and when queries yield no results; removed an outdated no-instruction test and added related test constants.


components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -83,12 +83,18 @@ def cache_query_result_and_get_first_page(
         """
         :param results: Log entries from the query to cache.
         :return: Forwards `SessionState.get_page_data`'s return values.
-        :return: _GET_INSTRUCTIONS_NOT_RUN_ERROR if `get_instructions` has not been called in this
```
rishikeshdevsot: if you removing this error message, will you be handling the case where `get_instructions` has not been called another way? or does that case no longer happen?

components/clp-mcp-server/tests/test_session_manager.py
```diff
@@ -41,6 +42,7 @@ class TestConstants:
     EXCEEDS_MAX_CACHED_RESULTS_ERR = "exceeds maximum allowed cached results"
     INVALID_NUM_ITEMS_PER_PAGE_ERR = "must be a positive integer"
     GET_INSTRUCTIONS_NOT_CALLED_ERR = "Please call `get_instructions()`"
+    NO_RESULTS_FOUND_IN_KQL_QUERY = "No log events found matching the KQL query."
```
LinZhihao-723: `NO_RESULTS_FOUND_IN_KQL_QUERY_ERR`?

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
```diff
@@ -83,12 +83,18 @@ def cache_query_result_and_get_first_page(
...
+        :return: A dictionary with the following key-value pair on failures:
+            - "Error": An error message describing the failure.
         """
         if self._is_instructions_retrieved is False:
             return self._GET_INSTRUCTIONS_NOT_RUN_ERROR.copy()
 
+        if len(results) == 0:
+            return {
+                "Error": "No log events found matching the KQL query. Try broadening your search"
+                " criteria."
```
LinZhihao-723: Can we drop "Try broadening your search criteria."? Or AI must need this hint to proceed?

LinZhihao-723 (changes requested):

components/clp-mcp-server/clp_mcp_server/server/session_manager.py
20001020ycx: emmm, yeah, agreed, this error should be general enough. But didn't we already violate this generosity when designing this error message:
```
    _GET_INSTRUCTIONS_NOT_RUN_ERROR: ClassVar[dict[str, str]] = {
        "Error": "Please call `get_instructions()` first to understand how to use this MCP server."
    }
```

I think the immediate move is to follow what you have suggested. Then we should refactor in a way that these functions should just return some generic error code, then add another layer to add on the instruction/hint for AI to better understand.

LinZhihao-723: I think this example message is fine, since AI must first call `get_instruction`.
But I am not sure if the one I commented is the same. AI could terminate search instead of broading the query, right?

20001020ycx: Well, the decision is up to AI, no matter what we do, we are merely giving a hint. Even with get_instructions, simply based on the dostring in the mcp tool call get_instructions(), AI is able to invoke it most of the time as the first API call from our empirical experience.  Adding the static check is just for a sanity check.

Similar thing goes with this line, even with our current implementation, that the returned error message is confusing, AI will continue doing the search rather than stop from our empirical experience.

LinZhihao-723: Sure. But:
- Without calling `get_instructions`, we shouldn't proceed any tool calls.
- If a query returns nothing, broading the search might not be the action to take all the time.

LinZhihao-723 (approved):
For the PR title, how about:
```
fix(clp-mcp-server): Return an error when a KQL query has no results.
```

PR #1494, opened 2025-10-24
fix(clp-mcp-server): Refine system prompt to make LLMs' KQL query generation more CLP friendly.

20001020ycx:
# Description

With our current system prompt, LLM would generate kql such as  `kql_query`: `register OR registration OR registered`.
This would result `{"Error":"No log events found matching the KQL query."}` or `{"Error":"Query job with ID 15 ended in status FAILED."}` pending the exact query submitted. 

The reason why this happens is that the kql generated by LLM is not CLP-compatible; CLP expects a kql query that looks like this, following the convention specified in [here](https://docs.yscope.com/clp/main/user-docs/reference-json-search-syntax.html).
`kql_query`: `*register* OR *registration* OR *registered*`.

This PR refines the system prompt to ensure LLM understands the rule of CLP's variant of KQL, thereby improving the efficiency of the diagnosis.

# Checklist

  breaking change.

# Validation performed

We have carefully designed the system prompt and evaluated on two real-world cases: HDFS-4058 and Hadoop-10508. We have run each case three times, and the AI-generated kql reaches 100% accuracy (that is each query submitted wound't be killed or fail). HDFS-4058 makes 5 kql queries per time, and Hadoop-10508 makes 3 kql queires per time.

## Release Notes

* **Improvements**
  * Updated query guidance system to provide enhanced assistance and better support for CKQL queries with improved terminology and rule clarity.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "94fb8275041c72129eb51b7d0322c29b8387a0386127718b096429201a5d6ece"
+dependencies = [
+ "alloc-no-stdlib",
+]
+
+[[package]]
 name = "allocator-api2"
 version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -551,10 +566,37 @@ dependencies = [
 ]
 
 [[package]]
+name = "brotli"
+version = "8.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4bd8b9603c7aa97359dbd97ecf258968c95f3adddd6db2f7e7a5bef101c84560"
+dependencies = [
+ "alloc-no-stdlib",
+ "alloc-stdlib",
+ "brotli-decompressor",
+]
+
+[[package]]
+name = "brotli-decompressor"
+version = "5.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "874bb8112abecc98cbd6d81ea4fa7e94fb9449648c93cc89aa40c81c24d7de03"
+dependencies = [
+ "alloc-no-stdlib",
+ "alloc-stdlib",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "46c5e41b57b8bba42a04676d81cb89e9ee8e859a1a66f80a5a72e1cb76b34d43"
+
+[[package]]
+name = "byteorder"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
@@ -617,7 +659,13 @@ dependencies = [
  "aws-config",
  "aws-sdk-s3",
  "aws-sdk-sqs",
+ "brotli",
+ "hex",
+ "rmp-serde",
  "secrecy",
+ "serde",
+ "serde_json",
+ "thiserror",
 ]
 
 [[package]]
@@ -1508,6 +1556,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "paste"
+version = "1.0.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "57c0d7b74b563b49d38dae00a0c37d4d6de9b432382b2892f0574ddcae73fd0a"
+
+[[package]]
 name = "percent-encoding"
 version = "2.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1689,6 +1743,28 @@ dependencies = [
  "libc",
  "untrusted",
  "windows-sys 0.52.0",
+]
+
+[[package]]
+name = "rmp"
+version = "0.8.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "228ed7c16fa39782c3b3468e974aec2795e9089153cd08ee2e9aefb3613334c4"
+dependencies = [
+ "byteorder",
+ "num-traits",
+ "paste",
+]
+
+[[package]]
+name = "rmp-serde"
+version = "1.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "52e599a477cf9840e92f2cde9a7189e67b42c57532749bf90aea6ec10facd4db"
+dependencies = [
+ "byteorder",
+ "rmp",
+ "serde",
 ]
 
 [[package]]
@@ -1924,6 +2000,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde_json"
+version = "1.0.145"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "402a6f66d8c709116cf22f558eab210f5a50187f702eb4d7e5ef38d9a7f1c79c"
+dependencies = [
+ "itoa",
+ "memchr",
+ "ryu",
+ "serde",
+ "serde_core",
+]
+
+[[package]]
 name = "sha1"
 version = "0.10.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2040,6 +2129,26 @@ name = "synstructure"
 version = "0.13.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "728a70f3dbaf5bab7f0c4b1ac8d7ae5ea60a4b5549c8a5914361c99147a709d2"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "thiserror"
+version = "2.0.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f63587ca0f12b72a0600bcba1d40081f830876000bb46dd2337a3051618f4fc8"
+dependencies = [
+ "thiserror-impl",
+]
+
+[[package]]
+name = "thiserror-impl"
+version = "2.0.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3ff15c8ecd7de3849db632e14d18d2571fa09dfc5ed93479bc4485c7a517c913"
 dependencies = [
  "proc-macro2",
  "quote",
--- a/components/clp-package-utils/clp_package_utils/controller.py
+++ b/components/clp-package-utils/clp_package_utils/controller.py
@@ -743,8 +743,7 @@ class DockerComposeController(BaseController):
         logger.info(f"Starting CLP using Docker Compose ({deployment_type} deployment)...")
 
         cmd = ["docker", "compose", "--project-name", self._project_name]
-        if deployment_type == DeploymentType.BASE:
-            cmd += ["--file", "docker-compose.base.yaml"]
+        cmd += ["--file", self._get_docker_file_name()]
         if self._clp_config.mcp_server is not None:
             cmd += ["--profile", "mcp"]
         cmd += ["up", "--detach", "--wait"]
@@ -794,6 +793,15 @@ class DockerComposeController(BaseController):
         """
         # This will change when we move from single to multi-container workers. See y-scope/clp#1424
         return multiprocessing.cpu_count() // 2
+
+    def _get_docker_file_name(self) -> str:
+        """
+        :return: The Docker Compose file name to use based on the config.
+        """
+        deployment_type = self._clp_config.get_deployment_type()
+        if deployment_type == DeploymentType.BASE:
+            return "docker-compose-base.yaml"
+        return "docker-compose.yaml"
 
 
 def get_or_create_instance_id(clp_config: CLPConfig) -> str:
--- a/components/clp-py-utils/clp_py_utils/clp_config.py
+++ b/components/clp-py-utils/clp_py_utils/clp_config.py
@@ -48,8 +48,8 @@ COMPRESSION_JOBS_TABLE_NAME = "compression_jobs"
 COMPRESSION_TASKS_TABLE_NAME = "compression_tasks"
 
 # Paths
-CONTAINER_AWS_CONFIG_DIRECTORY = pathlib.Path("/") / ".aws"
 CONTAINER_CLP_HOME = pathlib.Path("/") / "opt" / "clp"
+CONTAINER_AWS_CONFIG_DIRECTORY = CONTAINER_CLP_HOME / ".aws"
 CONTAINER_INPUT_LOGS_ROOT_DIR = pathlib.Path("/") / "mnt" / "logs"
 CLP_DEFAULT_CONFIG_FILE_RELATIVE_PATH = pathlib.Path("etc") / "clp-config.yml"
 CLP_DEFAULT_CREDENTIALS_FILE_PATH = pathlib.Path("etc") / "credentials.yml"
--- a/components/clp-rust-utils/Cargo.toml
+++ b/components/clp-rust-utils/Cargo.toml
@@ -7,4 +7,12 @@ edition = "2024"
 aws-config = "1.8.8"
 aws-sdk-s3 = "1.106.0"
 aws-sdk-sqs = "1.86.0"
+brotli = "8.0.2"
+rmp-serde = "1.3.0"
 secrecy = { version = "0.10.3", features = ["serde"] }
+serde = { version = "1.0.228", features = ["derive"] }
+thiserror = "2.0.17"
+
+[dev-dependencies]
+hex = "0.4.3"
+serde_json = "1.0.145"
--- a/components/clp-rust-utils/src/clp_config.rs
+++ b/components/clp-rust-utils/src/clp_config.rs
@@ -1,0 +1,3 @@
+mod s3_config;
+
+pub use s3_config::*;
--- a/components/clp-rust-utils/src/clp_config/s3_config.rs
+++ b/components/clp-rust-utils/src/clp_config/s3_config.rs
@@ -1,0 +1,25 @@
+use serde::Serialize;
+
+/// Represents the configuration for connecting to an S3 bucket.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub struct S3Config {
+    pub bucket: String,
+    pub region_code: String,
+    pub key_prefix: String,
+    pub aws_authentication: AwsAuthentication,
+}
+
+/// An enum representing AWS authentication methods.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+#[serde(tag = "type")]
+pub enum AwsAuthentication {
+    #[serde(rename = "credentials")]
+    Credentials { credentials: AwsCredentials },
+}
+
+/// Represents AWS credentials.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub struct AwsCredentials {
+    pub access_key_id: String,
+    pub secret_access_key: String,
+}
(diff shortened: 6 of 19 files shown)


rishikeshdevsot:
@kirkrodrigues @gibber9809 this PR's objective is to explain to the LLM the difference between regular KQL and KQL supported by CLP using as little text/tokens as possible. So one of you should probably take a second look to ensure that the important points are covered. 

Some considerations:
1. We don't want to list all the [differences](https://docs.yscope.com/clp/main/user-docs/reference-json-search-syntax.html#differences-with-kql) because that's too much text and it will confuse the LLM. 
2. We also can't rely on AI following this link because not all agents will be given internet access and even with internet access, an agent following a link is hit or miss.

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -12,22 +12,27 @@
 
 # fmt: off
 SYSTEM_PROMPT = (
-"You are an AI assistant that helps users query a log database using KQL (Kibana Query Language)."
+"You are an AI assistant that helps users query a log database called CLP using KQL"
```
kirkrodrigues: Here's what I came up with:

You are an AI assistant that helps users query a log database, called CLP, by using a variant of KQL
(Kibana Query Language) that we'll call CLP-KQL.

You should generate a CLP-KQL query that accurately expresses the user's intent. The generated query
should be as specific as possible to minimize the number of log messages returned. When displaying
log messages, wrap them in hyperlinks with the `link` field from the search result.

Consider the following guidelines when generating CLP-KQL queries:
- To search for a substring (a partial match), use the wildcards `*` and `?`.
- To query a key or a value that contains multiple words separated by spaces, wrap the words in
  double quotation marks.
- In keys, to search for any of the literal characters `\`, `"`, `.`, `*`, `@`, `$`, `!`, and `#`, escape
  them with a backslash.
- In values, to search for any of the literal characters `\`, `"`, `?`, and `*`, escape them with a
  backslash.
- Consider using `search_by_kql_with_timestamp_range` to narrow the query to a specific time range.
- CLP doesn't support fuzzy matches.
- CLP doesn't support less/greater than comparisons on strings, IP addresses, or timestamps.
- If the user's prompt is vague, ask clarifying questions to better understand their intent before
  generating the CLP-KQL query.

20001020ycx: I don't have problems with the rest of the points except the first two because we have specifically refined those two in order to make AI generate correct kql.
We have iterated many versions and only the current version ensures AI could understand and follow the rule.  The takeaway from the experiment was that in order to mark a critical rule effective, we need to be:
1.  marking important rule with CRITICAL, 
2.  explaining the rule with a concrete example

kirkrodrigues: * Can you write a revision?
* Can we add a comment above the prompt explaining what characteristics need to be maintained based on your experiments? Otherwise the next person to edit this likely won't know either, lol.

20001020ycx: I agree with the second point and will add the rule as the comment right above the constant instantiation. But let's first make an agreement on the actual system prompt string.

Here is my revision:

You are an AI assistant that helps users query a log database, called CLP, by using a variant of KQL
(Kibana Query Language) that we'll call CLP-KQL.

You should generate a CLP-KQL query that accurately expresses the user's intent. The generated query
should be as specific as possible to minimize the number of log messages returned. When displaying
log messages, wrap them in hyperlinks with the link field from the search result.

Consider the following guidelines when generating CLP-KQL queries, and pay special attention to rules marked as CRITICAL - violation will cause query failure:

- CRITICAL: To search for a substring keyword (a partial match), you MUST use wildcards (* or ?) at the beginning and end of your keyword.
- CRITICAL: Use double quotation(") to wrap multiple keywords search that are separated by spaces. For example, to search for all logs containing: hello world, your query parameter should literally be: "*hello world*" (with the quotes as part of the search string wrapping around both wildcards and multiple words).
- In keys, to search for any of the literal characters \, ", ., *, @, $, !, and #, escape them with a backslash.
- In values, to search for any of the literal characters \, ", ?, and *, escape them with a backslash.
- Consider using search_by_kql_with_timestamp_range to narrow the query to a specific time range. Timestamps must follow the ISO 8601 UTC format (`YYYY-MM-DDTHH:mm:ss.fffZ`), where the trailing `Z` indicates UTC.
- CLP doesn't support fuzzy matches.
- CLP doesn't support less/greater than comparisons on strings, IP addresses, or timestamps.

20001020ycx: Here are my rationales behind my edits, a key point that I would like to emphasize is that we would like this system prompt to be AI friendly, even if this might not be stricitly following a documentation style for human readers

1. CRITICAL Rule Marking: LLMs do not guarantee enforcement of all system prompt rules—this was confirmed through empirical testing. Rules 1 and 2 are essential for generating correct CLP-KQL queries; without them, LLMs produce invalid syntax like message: hello world (missing wildcards and quotes). Marking these rules as "CRITICAL" significantly improves KQL success rate, and ensuring LLMs prioritize them even when other rules may be ignored.
2. Action item first and example showing: What I change in the second point was I move the action item for AI to the beginning of the sentence. I found that effective in the experiement. Same argument goes with example.
3. Definition of exact format of timestamp: query with timestamp that does not follows ISO 8601 in UTC format will be rejected, so we want to make sure AI generates the exact format
4. Removing your last point (also this is what I have in my original commit): i feel this is something good to have. However, I found that system prompt's length directly impacts LLM attention, as we don't want LLM loses its focus on CRITICAL points by some minor rules or reminder.

20001020ycx: The 2nd point mentioned by @kirkrodrigues is updated in the pr

kirkrodrigues: Responded in the design doc.

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,59 @@
...
-" Timestamps must follow the ISO 8601 UTC format (`YYYY-MM-DDTHH:mm:ss.fffZ`), where the trailing"
-" `Z` indicates UTC.\n"
-"- If the user query is ambiguous or lacks detail, ask clarifying questions to better understand"
-" their intent before generating the KQL query.\n"
-"- Always ensure that the generated KQL query is syntactically correct and can be executed without"
-" errors."
-)
+# Guidelines for Writing Effective System Prompts For LLM
+#
+# System prompts should be AI-friendly; while LLMs may not strictly enforce all rules, we
```
kirkrodrigues:
```suggestion
# System prompts should be LLM-friendly; while LLMs may not strictly enforce all rules, we
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,59 @@
...
+# empirically found the following practices effective for LLM to understand the listed rules:
```
kirkrodrigues:
```suggestion
# empirically found the following practices effective for LLM to understand the listed rules:
#
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,59 @@
...
+# 1. Provide concrete examples to explain the rule.
+# 2. Place critical rules at the beginning and mark them as "CRITICAL".
+# 3. Use action-first sentence structure (e.g., "Use X format" instead of "To do Y, use X format").
+# 4. Tell AI about any functional behaviour (like showing logs using LogViewer) very early
+#    in the prompt as the "job" of the agent.
+# 5. Use terse language and bullet points; avoid complex sentence structures.
+#    AI can fill in the gaps
+# 6. Let some instructions and details be implicit to avoid overwhelming AI.
```
kirkrodrigues:
```suggestion
# 3. Use action-first sentence structure (e.g., "Use X to do Y" instead of "To do Y, use X").
# 4. Specify any behaviour that the agent needs to perform (like formatting hyperlinks) very early
#    in the prompt.
# 5. Use terse language and bullet points; avoid complex sentence structures. LLMs can fill in the
#    gaps.
# 6. Let some instructions and details be implicit to avoid overwhelming the LLM.
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,59 @@
...
+# 7. Use the same example across different rules to maintain consistency.
+SYSTEM_PROMPT = """You are an AI assistant for querying the CLP log database
+ using CLP-KQL (CKQL).\n
```
kirkrodrigues:
```suggestion
SYSTEM_PROMPT = """You are an AI assistant for querying the CLP log database using CLP-KQL (CKQL).
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,59 @@
...
+
+Your job is to generate CKQL that faithfully expresses the user's intent and
+ show key logs to the user
+- Start broad to learn the schema/fields using wildcard searches like *, then
+ narrow the query to return a manageable result set.
+- When showing log messages or when the user wants to see log messages, provide
+ the hyperlink from the result's link field.\n\n
```
kirkrodrigues:
```suggestion
Your job is to generate CKQL that faithfully expresses the user's intent and show key logs to the
user:
- Start broad to learn the schema/fields using wildcard searches like *, then narrow the query to
  return a manageable result set.
- When showing log messages or when the user wants to see log messages, provide the hyperlink from
  the result's link field.
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,59 @@
...
+
+CKQL rules (read carefully; items marked CRITICAL will fail if violated):\n
+- CRITICAL -- Substrings: use wildcards for partial
+ matches -- * (any sequence), ? (single character).
+  Example:
+  request: *GET*\n
```
kirkrodrigues:
```suggestion
- CRITICAL -- Substrings: use wildcards for partial matches -- * (any sequence), ? (single
  character).\n
  Example:\n
  request: *GET*\n\n
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,59 @@
...
+
+- Combining conditions: use AND / OR (case-insensitive).
+  Example:
+  request: GET AND response: 400\n
```
kirkrodrigues:
```suggestion
- Combining conditions: use AND / OR (case-insensitive).\n
  Example:\n
  request: GET AND response: 400\n\n
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,59 @@
...
+
+- CRITICAL --  Multi-word text must be quoted: wrap multi-word
+ searches in double quotes.
+  Example:
+  request: "*GET wp-admin*"
+  (quotes and wildcards are required).
```
kirkrodrigues:
```suggestion
- CRITICAL --  Multi-word text must be quoted: wrap multi-word searches in double quotes.\n
  Example:\n
  request: "*GET wp-admin*"\n
  (quotes and wildcards are required).\n\n
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,59 @@
...
+
+
+- Escaping characters:
+  - In keys, use backslash to escape searching for any
+   of the literal characters: \\, ", ., *, @, $, !, #.
+  - In values, use backslash to escape
+   searching for any of the literal characters: \\, ", ?, *.
+  \n
```
kirkrodrigues:
```suggestion
- Escaping characters:\n
  - In keys, use backslash to escape searching for any of the literal characters: \\, ", ., *, @, $,
    !, #.\n
  - In values, use backslash to escape searching for any of the literal characters: \\, ", ?, *.\n\n
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,59 @@
...
+
+- Time range: use search_by_kql_with_timestamp_range
+ to constrain by time.\n
+
+- Unsupported: no fuzzy matches; no less/greater-than
+ comparisons on strings, IPs, or timestamps.\n
```
kirkrodrigues:
```suggestion
- Unsupported: no fuzzy matches; no less/greater-than comparisons on strings, IPs, or timestamps.
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
kirkrodrigues:
```suggestion
CKQL rules (read carefully; items marked CRITICAL will fail if violated):
- CRITICAL -- Substrings: use wildcards for partial matches -- * (any sequence), ? (single
  character).
  Example:
  request: *GET*
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
kirkrodrigues:
```suggestion
  request: GET AND response: 400
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,59 @@
...
+ matches -- * (any sequence), ? (single character).
+  Example:
+  request: *GET*\n
+
+- Combining conditions: use AND / OR (case-insensitive).
+  Example:
+  request: GET AND response: 400\n
+
+- CRITICAL --  Multi-word text must be quoted: wrap multi-word
+ searches in double quotes.
```
kirkrodrigues:
```suggestion
- CRITICAL --  Multi-word text must be quoted: wrap multi-word searches in double quotes.
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,59 @@
...
+  Example:
+  request: "*GET wp-admin*"
+  (quotes and wildcards are required).
+
+
```
kirkrodrigues:
```suggestion
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
kirkrodrigues:
```suggestion
  - In keys, use backslash to escape searching for any of the literal characters: \\, ", ., *, @, $,
    !, #.
  - In values, use backslash to escape searching for any of the literal characters: \\, ", ?, *.
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
kirkrodrigues:
```suggestion
- Unsupported: no fuzzy matches; no less/greater-than comparisons on strings, IPs, or timestamps.
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,59 @@
...
+  \n
+
+- Time range: use search_by_kql_with_timestamp_range
+ to constrain by time.\n
+
+- Unsupported: no fuzzy matches; no less/greater-than
+ comparisons on strings, IPs, or timestamps.\n
+"""
+
+
```
kirkrodrigues:
```suggestion
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,59 @@
...
+- Escaping characters:
+  - In keys, use backslash to escape searching for any
+   of the literal characters: \\, ", ., *, @, $, !, #.
+  - In values, use backslash to escape
+   searching for any of the literal characters: \\, ", ?, *.
+  \n
+
+- Time range: use search_by_kql_with_timestamp_range
+ to constrain by time.\n
```
kirkrodrigues:
```suggestion
- Time range: use search_by_kql_with_timestamp_range to constrain by time.
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
20001020ycx: I have a concern about using triple quotation marks for the system prompt. Triple quotes create a raw string that preserves all newlines, including those added solely for code formatting (to maintain 100 characters per line). This causes prompt sentences to be split across multiple lines unintentionally.

For example,  for a sentence defined like this:
> When showing log messages or when the user wants to see log messages, provide the hyperlink from
>   the result's link field.

will be interpreted by AI to be 
> When showing log messages or when the user wants to see log messages, provide the hyperlink from \n the result's link field.

which would alter the semantics of the system prompt due to this actual \n.

Therefore, should we revert back to double quotation and add \n explicitly?

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,49 @@
 SERVER_NAME = "clp-mcp-server"
 
 # fmt: off
+# System prompts should be LLM-friendly; while LLMs may not strictly enforce all rules, we
+# empirically found the following practices effective for LLM to understand the listed rules:
```
kirkrodrigues:
```suggestion
# empirically found the following practices effective for LLMs to understand the listed rules:
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,49 @@
...
+"- CRITICAL -- Substrings: use wildcards for partial matches -- * (any sequence), ? (single"
+" character).\n"
+"  Example:\n"
+"  request: *GET*\n"
+"\n"
+"- Combining conditions: use AND / OR (case-insensitive).\n"
+"  Example:\n"
+"  request: GET AND response: 400\n"
+"\n"
+"- CRITICAL --  Multi-word text must be quoted: wrap multi-word searches in double quotes.\n"
```
kirkrodrigues:
```suggestion
"- CRITICAL -- Multi-word text must be quoted: wrap multi-word searches in double quotes.\n"
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,49 @@
...
+#
+# 1. Provide concrete examples to explain the rule.
+# 2. Place critical rules at the beginning and mark them as "CRITICAL".
+# 3. Use action-first sentence structure (e.g., "Use X to do Y" instead of "To do Y, use X").
+# 4. Specify any behaviour that the agent needs to perform (like formatting hyperlinks) very early
+#    in the prompt.
+# 5. Use terse language and bullet points; avoid complex sentence structures. LLMs can fill in the
+#    gaps.
+# 6. Let some instructions and details be implicit to avoid overwhelming the LLM.
+# 7. Use the same example across different rules to maintain consistency.
```
kirkrodrigues, line 24:
```suggestion
# 7. Use the same example across different rules to maintain consistency.
# 8. Don't wrap text since the extra line breaks may influence the LLM's understanding.
```

components/clp-mcp-server/clp_mcp_server/server/constants.py
```diff
@@ -11,23 +11,49 @@
...
 SYSTEM_PROMPT = (
```
kirkrodrigues:
```suggestion
# System prompts should be LLM-friendly; while LLMs may not strictly enforce all rules, we
# empirically found the following practices effective for LLM to understand the listed rules:
#
# 1. Provide concrete examples to explain the rule.
# 2. Place critical rules at the beginning and mark them as "CRITICAL".
# 3. Use action-first sentence structure (e.g., "Use X to do Y" instead of "To do Y, use X").
# 4. Specify any behaviour that the agent needs to perform (like formatting hyperlinks) very early
#    in the prompt.
# 5. Use terse language and bullet points; avoid complex sentence structures. LLMs can fill in the
#    gaps.
# 6. Let some instructions and details be implicit to avoid overwhelming the LLM.
# 7. Use the same example across different rules to maintain consistency.
# fmt: off
SYSTEM_PROMPT = (
```

kirkrodrigues (commented):
For the PR title, how about:

```
fix(clp-mcp-server): Refine system prompt to make the LLM's KQL query generation more CLP friendly.
```

components/clp-mcp-server/pyproject.toml
kirkrodrigues, line 1: Do we need this change in this PR? It seems unrelated.

20001020ycx: we dont, this is fo the change of another pr, I accidentally merge it in, sorry I will fix it now

kirkrodrigues (approved):

PR #1500, opened 2025-10-25
fix(deployment): Update archives directory mount paths in Presto Docker Compose project to match the CLP package Compose project (fixes #1496).

### LINKED ISSUE #1496: Fix file mounting inconsistency between Presto and CLP Package Docker Compose projects
## Problem

The Presto Docker Compose project and the CLP Package Docker Compose project use different conventions for file mounting, which causes Presto queries to return zero rows.

### Current Behavior

- **Presto project (PR #1132):** Mounts the host's `/path/to/var/data/archives` directory (or `/path/to/var/data/staged-archives` when using S3 output) into the container **at the same absolute path** as on the host.
- **CLP Package project (PR #1178):** Always mounts the host's `/path/to/var/data/archives` directory to `/var/data/archives` in the container, and `/path/to/var/data/staged-archives` to `/var/data/staged-archives`.

### Root Cause

With the `clp-s` storage engine, when archives are compressed through the Package project, the metadata DB's `clp-datasets` table stores the `archive_storage_directory` field using container paths (e.g., `/var/data/archives`) instead of the corresponding host paths (e.g., `/path/to/var/data/archives`). This mismatch prevents the Presto coordinator from locating the archives, as they are not mounted under `/var/data/archives` in the Presto container.

### Impact

Presto queries return zero rows because the coordinator cannot locate the archives due to path mismatches.

### Historical Context

Before PR #1178, when the components were orchestrated via Python subprocess calls to the `docker` CLI, they followed the same mapping convention as the current Presto project.

## References

- PR #1132 (Presto Docker Compose setup)
- PR #1178 (CLP Package Docker Compose project)
- Original comment: https://github.com/y-scope/clp/pull/1132#discussion_r2462660309

junhaoliao:
# Description

This change standardizes archive-related mount points in the Presto Docker Compose setup to align with the CLP Package Compose conventions.

- Docker Compose (presto-clp):
  - Mount host archives to `/var/data/archives:ro` using `${CLP_ARCHIVES_DIR:-empty}`.
  - Mount staged archives to `/var/data/staged-archives:ro` using `${CLP_STAGED_ARCHIVES_DIR:-empty}`.
  - Introduce a named `empty` volume as a safe default so Compose can start even when env vars are unset.
- Init script:
  - For `s3` archive output storage, set `CLP_STAGED_ARCHIVES_DIR` based on the configured `staging_directory` (previously set `CLP_ARCHIVES_DIR`).

# Checklist

  breaking change.

# Validation performed

## fs output
1. Started both projects and compressed sample data:
	```
	cd <project-root>
	task
	
	cd build/clp-package
	
	# follow https://docs.yscope.com/clp/main/user-docs/guides-using-presto.html#setting-up-clp and configure `etc/clp-config.yaml`
	
	./sbin/start-clp.sh
	./sbin/compress.sh --timestamp-key=timestamp ~/samples/postgresql.jsonl
	
	
	cd <project-root>
	cd tools/deployment/presto-clp
	# follow https://docs.yscope.com/clp/main/user-docs/guides-using-presto.html#setting-up-presto and configure `coordinator/config-template/split-filter.json`
	./scripts/set-up-config.sh <project-root>/build/clp-package
	docker compose up --wait
	```
2. Accessed the presto web console at http://localhost:8889/ui/sql_client.html
   1. Set "Catalog" to `clp`, "Schema" to `default`
   2. Ran query:
      ```sql
      select timestamp, message from default limit 100
      ```
   3. Observed rows are returned in the results:
      image

## s3 output
1. Clean built the project.
2. Followed Step 1 in above "fs output" instructions to start both projects and compress sample data. When configuring `clp-config.yml`, also configure the archives output to use s3 output:
	```yaml
	archive_output:
	  storage:
	    type: "s3"
	    s3_config:
	      ...
	```
3. Followed Step 1 in above "fs output" instructions. Performed a query and observed results:
   image

## Summary by CodeRabbit

* **Chores**
  * Updated deployment to use conditional mounts with a dummy fallback volume for archives and staged archives, improving robustness when host paths are unset.
  * Adjusted environment variable for S3-based storage to use a staging-specific name so staged archive paths are consistently set for cloud deployments.

### DIFF
--- a/tools/deployment/presto-clp/docker-compose.yaml
+++ b/tools/deployment/presto-clp/docker-compose.yaml
@@ -43,9 +43,10 @@ services:
       - "presto"
 
 volumes:
-  coordinator-config:
   # Dummy volume to use when a bind mount is not desired.
   empty:
+
+  coordinator-config:
   worker-config:
 
 networks:


tools/deployment/presto-clp/docker-compose.yaml
```diff
@@ -34,15 +34,17 @@ services:
...
       - "./worker/config-template:/configs:ro"
       - "./worker/scripts:/scripts:ro"
-      - "${CLP_ARCHIVES_DIR}:${CLP_ARCHIVES_DIR}"
       - "worker-config:/opt/presto-server/etc"
     networks:
       - "presto"
 
 volumes:
   coordinator-config:
+  empty:
```
junhaoliao, line 47: This volume differs from the one used in the Package project, where a `tmpfs` mount with size 0 is configured.

In this case, the volume is mounted as read-only, so we don't need to account for any data written to the dummy empty volume.

kirkrodrigues: Perhaps we should add a comment for future editors?

kirkrodrigues (commented):
All good except for the question I posted.

For the PR title, how about:

```
fix(deployment): Update `/var/data` mount paths in Presto Docker Compose project to match the CLP package Compose project (fixes #1496).
```

tools/deployment/presto-clp/docker-compose.yaml
junhaoliao: one moment. i think it's worth marking the volume as readonly to avoid mistakes

junhaoliao: actually the `readonly` attribute can only be applied when defining the mount, not in the volume definition. changing the definition to `tmpfs` like the Package project should just work, but it will take time to re-validate. i think it's fine to keep it simple as currently is, so long it works and has been validated

tools/deployment/presto-clp/docker-compose.yaml
```diff
@@ -34,15 +34,18 @@ services:
...
       - "./worker/config-template:/configs:ro"
       - "./worker/scripts:/scripts:ro"
-      - "${CLP_ARCHIVES_DIR}:${CLP_ARCHIVES_DIR}"
       - "worker-config:/opt/presto-server/etc"
     networks:
       - "presto"
 
 volumes:
   coordinator-config:
+  # Dummy volume to use when a bind mount is not desired.
```
kirkrodrigues, line 46:
```suggestion

  # Dummy volume to use when a bind mount is not desired.
```

kirkrodrigues: Actually, one sec.

tools/deployment/presto-clp/docker-compose.yaml
```diff
@@ -34,15 +34,18 @@ services:
...
+  empty:
   worker-config:
```
kirkrodrigues:
```suggestion
  # Dummy volume to use when a bind mount is not desired.
  empty:

  coordinator-config:
  worker-config:
```

PR #1512, opened 2025-10-28
feat(clp-package)!: Containerize the scripts in `sbin` via Docker Compose; Remove redundant assets from the package (resolves #1358).

### LINKED ISSUE #1358: Optimize clp-package Docker build to avoid intermediate copying
## Description

Optimize the clp-package Docker image build process to avoid copying everything into the clp-package directory first. Instead, specify `build/<component>` paths directly in COPY commands in the Dockerfile.

## Background

Currently, the build process copies all components into clp-package before building the Docker image. This can be optimized by directly copying from `build/<component>` paths in the Dockerfile COPY commands.

## Priority

This is a low-priority optimization that is not blocking the Docker Compose migration.

## Related

- PR #1335: Generate unique identifier for each clp-package dev image build
- Comment: https://github.com/y-scope/clp/pull/1335#discussion_r
- Requested by: @junhaoliao

junhaoliao:
# Description

This change containerizes the package runtime, moving all user-facing CLI wrappers to run inside a Docker Compose–managed container. It introduces a unified environment bootstrap for consistent configuration and reduces package bloat by eliminating duplication of assets already included in the container image.

Summary of changes (2 <ins>**BREAKING**</ins> changes):
- Runtime execution
  - <ins>**BREAKING**</ins>: `${CLP_HOME}/var/data`, `${CLP_HOME}/var/log`, `${CLP_HOME}/var/tmp` will now be created with ownership `${USER}:docker` instead of `${USER}:<user's default group>`. 
  - Add `tools/deployment/package/docker-compose.runtime.yaml` defining the `clp-runtime` service.
    - Maps the running user's `HOME` into the runtime environment for `~` in `clp-config.yaml` to work.
  - New `components/package-template/src/sbin/.common-env.sh` that:
    - Resolves `CLP_HOME`, image ref (`CLP_PACKAGE_CONTAINER_IMAGE_REF`), and `CLP_FIRST_PARTY_SERVICE_UID_GID`.
    - Exposes `CLP_PWD_HOST`
    - Detects Docker Compose plugins & socket for in-container use.
- CLI wrappers
  - Update all `src/sbin/*.sh` wrappers (admin-tools and user tools) to `source` `.common-env.sh` and invoke:
    `docker compose -f "$CLP_HOME/docker-compose.runtime.yaml" run --rm clp-runtime python3 -m ...`
- Python scripts (package utils)
  - Remove `mounts.clp_home` from `necessary_mounts` in:
    - `archive_manager.py`, `compress.py`, `compress_from_s3.py`, `dataset_manager.py`, `decompress.py` (both handlers), `search.py`.
  - `decompress.py`: default `--extraction-dir` to `CLP_PWD_HOST` (falls back to `"."`).
- Build system
  - <ins>**BREAKING**</ins>: Remove all binaries from `clp-package/bin`.
  - Restructure `taskfile.yaml`:
    - New `python-libs` internal task producing `build/python-libs`.
    - `docker-images:package` now depends on image and copies settings/webui artifacts
    - `default` now runs task `package` again.
  - Rework `tools/docker-images/clp-package/Dockerfile` to assemble image directly from build outputs:
    - Copy core binaries, spider workers, Node 22, Python libs to `lib/python3/site-packages/`, and webui to `var/www/webui/`.
- Template/package layout
  - Package now ships `docker-compose.runtime.yaml` and runs without a locally managed `PYTHONPATH`.

Rationale:
- Standardize execution environment across hosts (Linux/WSL, and possibly Mac) via container.
- Eliminate brittle dependency on host Python layout by running modules within the packaged image.

# Checklist

  breaking change.

# Validation performed

Observed size reduction of the package
```
junhao@ASUS-X870E:~/workspace/6-clp/build/clp-package$ du -sh .
180K    .
```

---

1. `task` to build the package.
2. `cd build/clp-package/ && ./sbin/start-clp.sh`
3. `./sbin/compress.sh ~/samples/hive-24hr`
4. `./sbin/search.sh 1` which returned results
5. `cd  /tmp && ./sbin/decompress.sh x` which extracted the logs into `/tmp` with correct hierarchy set up.
6. Launched the webui, performed a search with string "1" and observed results returned; clicked the link in the results which restored the log event context in the log viewer without failure.

---

1. repeated above with clp-s

## Summary by CodeRabbit

* **New Features**
  * CLI tools now run inside a containerised runtime with a shared environment and improved host-bind handling.

* **Chores**
  * Packaging adds Python library artifacts and web UI assets; entry scripts source a common environment.
  * Runtime image now includes additional runtime components and web UI content.

* **Bug Fixes**
  * Improved host-path resolution for validations, mounts, file I/O and cleanup.

* **Documentation**
  * Quick‑start: removed Python install guidance and Docker Desktop host‑networking note.


components/clp-package-utils/clp_package_utils/scripts/archive_manager.py
```diff
@@ -220,7 +220,6 @@ def main(argv: List[str]) -> int:
     )
 
     necessary_mounts: List[Optional[DockerMount]] = [
-        mounts.clp_home,
```
junhaoliao, line 223: mounts like this have to be removed, or they conflict with the contents inside the docker image.

components/clp-package-utils/clp_package_utils/scripts/decompress.py
```diff
@@ -298,8 +298,9 @@ def main(argv):
     file_extraction_parser.add_argument(
         "-f", "--files-from", help="A file listing all files to extract."
     )
+    default_extraction_dir = pathlib.Path(os.environ.get("CLP_PWD_HOST", "."))
```
junhaoliao, line 313: if we don't explicitly set and read this, `.` resolves to `/opt/clp`

tools/deployment/package/docker-compose.runtime.yaml
junhaoliao, line 1: we don't really have to use Docker Compose for this container definition, though Docker Compose helps it keep structured and avoids duplicate code.

junhaoliao (commented):
to make sure if we don't miss resolving any paths and to avoid showing the prefixes in the error messages which may confuse users, we only
1. "resolve" any host path right before their consumptions (e.g., `.mkdir()`, `open()`, `.exists()`, `isdir()`)
2. "resolve" only the paths that can be outside of the CLP_HOME directory, since we map CLP_HOME into the runtime container anyways

components/clp-py-utils/clp_py_utils/core.py
```diff
@@ -62,6 +65,26 @@ def read_yaml_config_file(yaml_config_file_path: pathlib.Path):
...
+    Translates a host path to its container-mount equivalent.
+
+    :param path: The host path.
+    :return: The translated container path.
+    """
+    path = path.absolute()
+    resolved = CONTAINER_HOST_ROOT_DIR / path.relative_to("/")
+
+    try:
+        if resolved.is_symlink():
```
junhaoliao: this is needed because somehow my WSL Ubuntu has this
```
(clp) junhao@ASUS-X870E:~$ ls -l .aws
lrwxrwxrwx 1 junhao junhao 24 Aug 26 12:07 .aws -> /mnt/c/Users/junhao/.aws
```

`/mnt/c/Users/junhao/.aws` is certainly not a valid path in the runtime container, to access, the bootstrap scripts should use the prefix + `/mnt/c/Users/junhao/.aws`

tools/deployment/package/docker-compose.runtime.yaml
```diff
@@ -0,0 +1,19 @@
...
+    environment:
+      HOME: "${HOME}"
+      CLP_HOME: "${CLP_HOME}"
+      CLP_PWD_HOST: "${CLP_PWD_HOST:-${PWD}}"
+    volumes:
+      - "${CLP_DOCKER_PLUGIN_DIR}:/usr/local/lib/docker/cli-plugins:ro"
+      - "${CLP_DOCKER_SOCK_PATH}:/var/run/docker.sock"
+      - "${CLP_HOME}:${CLP_HOME}"
+      - "/:/mnt/host"
+      - "/usr/bin/docker:/usr/bin/docker:ro"
```
kirkrodrigues: We can use the extended syntax here, right?

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -749,7 +751,8 @@ def validate_aws_config_dir(self):
                 raise ValueError(
                     "aws_config_directory must be set when using profile authentication"
                 )
-            if not self.aws_config_directory.exists():
+            if not resolve_host_path(self.aws_config_directory).exists():
+                print(resolve_host_path(self.aws_config_directory))
```
kirkrodrigues: Intentional?

components/clp-py-utils/clp_py_utils/core.py
```diff
@@ -1,8 +1,11 @@
+import os
```
kirkrodrigues:
```suggestion
```

components/package-template/src/sbin/start-clp.sh
```diff
@@ -1,9 +1,10 @@
 #!/usr/bin/env bash
 
 script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
-package_root="$script_dir/.."
+common_env_path="$script_dir/.common-env.sh"
+source "$common_env_path"
```
kirkrodrigues, line 7:
```suggestion

# shellcheck source=.common-env.sh
source "$common_env_path"
```

kirkrodrigues (commented):
Not a complete review but some things I noticed while reading so far.

Some of the shell scripts have lines longer than 100 chars as well as shellcheck warnings. Can you address?

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: sorry, no

.github/workflows/clp-artifact-build.yaml
```diff
@@ -548,7 +548,8 @@ jobs:
             ${{needs.filter-relevant-changes.outputs.ubuntu_jammy_image_changed == 'false'
             || (github.event_name != 'pull_request' && github.ref == 'refs/heads/main')}}
           run_command: >-
-            CLP_CPP_MAX_PARALLELISM_PER_BUILD_TASK=$(getconf _NPROCESSORS_ONLN) task package
+            CLP_CPP_MAX_PARALLELISM_PER_BUILD_TASK=$(getconf _NPROCESSORS_ONLN)
+            task package-build-deps
```
kirkrodrigues, line 552: Maybe change the name of this workflow step to "Build the package without the package image"

components/clp-py-utils/clp_py_utils/core.py
```diff
@@ -3,6 +3,8 @@
 import yaml
 from yaml.parser import ParserError
 
+CONTAINER_HOST_ROOT_DIR = pathlib.Path("/") / "mnt" / "host"
```
kirkrodrigues: How about `CONTAINER_DIR_FOR_HOST_ROOT`?

components/clp-py-utils/clp_py_utils/core.py
```diff
@@ -62,6 +64,26 @@ def read_yaml_config_file(yaml_config_file_path: pathlib.Path):
     return config
 
 
+def resolve_host_path(path: pathlib.Path) -> pathlib.Path:
```
kirkrodrigues: How about `resolve_host_path_in_container`?

components/clp-py-utils/clp_py_utils/core.py
```diff
@@ -62,6 +64,26 @@ def read_yaml_config_file(yaml_config_file_path: pathlib.Path):
...
+    """
+    Translates a host path to its container-mount equivalent.
+
+    :param path: The host path.
+    :return: The translated container path.
```
kirkrodrigues:
```suggestion
    :return: The translated path.
```

Otherwise, it could be misinterpreted as the container path which has been translated.

components/clp-py-utils/clp_py_utils/core.py
```diff
@@ -62,6 +64,26 @@ def read_yaml_config_file(yaml_config_file_path: pathlib.Path):
...
+    """
+    path = path.absolute()
+    resolved = CONTAINER_HOST_ROOT_DIR / path.relative_to("/")
```
kirkrodrigues: How about `host_path` and `translated_path`?

components/clp-py-utils/clp_py_utils/core.py
```diff
@@ -62,6 +64,26 @@ def read_yaml_config_file(yaml_config_file_path: pathlib.Path):
...
+
+    try:
+        if resolved.is_symlink():
+            target_path = (resolved.parent / resolved.readlink()).resolve()
+            resolved = CONTAINER_HOST_ROOT_DIR / target_path.relative_to("/")
+    except OSError:
+        pass
```
kirkrodrigues, line 90: I don't really get this part. Can you explain how it works?

taskfile.yaml
```diff
@@ -249,6 +212,14 @@ tasks:
...
 
+  package-build-deps:
+    deps:
+      - "init"
+      - "core"
+      - "deps:spider"
+      - "python-libs"
+      - "webui"
+
   webui:
```
kirkrodrigues, line 222: Don't we need to move `NODE_ENV: "production"` from the `package` task?

taskfile.yaml
```diff
@@ -107,80 +110,40 @@ tasks:
...
-      - "{{.G_SPIDER_BUILD_DIR}}/spider-build/src/spider/spider_worker"
+      - "{{.G_BUILD_DIR}}/clp-package-image.id"
+      - "{{.G_WEBUI_BUILD_DIR}}/client/settings.json"
+      - "{{.G_WEBUI_BUILD_DIR}}/server/dist/settings.json"
       - "{{.TASKFILE}}"
-      - "components/clp-mcp-server/dist/*.whl"
-      - "components/clp-package-utils/dist/*.whl"
-      - "components/clp-py-utils/dist/*.whl"
-      - "components/job-orchestration/dist/*.whl"
-      - "components/package-template/src/**/*"
```
kirkrodrigues, line 126: Why wouldn't this still be a source?

taskfile.yaml
```diff
@@ -249,6 +212,14 @@ tasks:
       vars:
         COMPONENT: "{{.TASK}}"
 
+  package-build-deps:
+    deps:
+      - "init"
+      - "core"
+      - "deps:spider"
+      - "python-libs"
+      - "webui"
```
kirkrodrigues: Deps run in parallel, so we should alphabetize.

taskfile.yaml
```diff
@@ -249,6 +212,14 @@ tasks:
...
+
   webui:
     vars:
       CHECKSUM_FILE: "{{.G_BUILD_DIR}}/{{.TASK}}.md5"
```
kirkrodrigues:
```suggestion
      CHECKSUM_FILE: "{{.G_WEBUI_CHECKSUM_FILE}}"
```

kirkrodrigues (commented):
Some more comments.

taskfile.yaml
junhaoliao: right, we should

taskfile.yaml
junhaoliao: sorry, it should still be

components/clp-py-utils/clp_py_utils/core.py
junhaoliao: -> `resolved.readlink()` gives the symlink target
-> `resolved.parent` gives the parent of the symlink itself
the full path of the target should be `(resolved.parent / resolved.readlink()).resolve()` then

the code doesn't handle the case when symlink target is an absolute path, and i agree it's not so well documented. let's see if the rewritten code is better

components/package-template/src/sbin/.common-env.sh
```diff
@@ -0,0 +1,60 @@
...
+version_file="$CLP_HOME/VERSION"
+
+if [[ -f "$image_id_file" ]]; then
+    image_id="$(tr -d '[:space:]' < "$image_id_file")"
+    export CLP_PACKAGE_CONTAINER_IMAGE_REF="$image_id"
+elif [[ -f "$version_file" ]]; then
+    version="$(tr -d '[:space:]' < "$version_file")"
+    export CLP_PACKAGE_CONTAINER_IMAGE_REF="ghcr.io/y-scope/clp/clp-package:v$version"
+else
+    >&2 echo "Error: Neither clp-package-image.id nor VERSION file exists."
```
kirkrodrigues:
```suggestion
    >&2 echo "Error: Neither '${image_id_file}' nor '${version_file}' exist."
```

components/package-template/src/sbin/.common-env.sh
```diff
@@ -0,0 +1,60 @@
...
+    return 1
+fi
+
+uid="$(id --user 2>/dev/null || echo "1000")"
+gid="$(getent group docker | cut -d: -f3 2>/dev/null || echo "999")"
```
kirkrodrigues, line 27: Why 999?

components/package-template/src/sbin/.common-env.sh
```diff
@@ -0,0 +1,60 @@
...
+export CLP_FIRST_PARTY_SERVICE_UID_GID="$uid:$gid"
+
+CLP_PWD_HOST="$(pwd 2>/dev/null || echo "")"
+export CLP_PWD_HOST
+
+if [[ -z "${CLP_DOCKER_PLUGIN_DIR:-}" ]]; then
+    for dir in \
```
kirkrodrigues: How about `compose_plugin_dir` to avoid overlap with the command `dir`?

components/package-template/src/sbin/.common-env.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
+
+script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
```
kirkrodrigues:
```suggestion
script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
```

components/package-template/src/sbin/.common-env.sh
```diff
@@ -0,0 +1,60 @@
...
+    export CLP_HOME="$package_root"
+else
+    export CLP_HOME="$CLP_HOME"
+fi
+
+image_id_file="$CLP_HOME/clp-package-image.id"
+version_file="$CLP_HOME/VERSION"
+
+if [[ -f "$image_id_file" ]]; then
+    image_id="$(tr -d '[:space:]' < "$image_id_file")"
```
kirkrodrigues:
```suggestion
    image_id="$(tr -d '[:space:]' <"$image_id_file")"
```

components/package-template/src/sbin/.common-env.sh
```diff
@@ -0,0 +1,60 @@
...
+    export CLP_PACKAGE_CONTAINER_IMAGE_REF="$image_id"
+elif [[ -f "$version_file" ]]; then
+    version="$(tr -d '[:space:]' < "$version_file")"
```
kirkrodrigues:
```suggestion
    version="$(tr -d '[:space:]' <"$version_file")"
```

components/package-template/src/sbin/.common-env.sh
```diff
@@ -0,0 +1,60 @@
...
+
+        compose_plugin_path="$dir/docker-compose"
+        if [[ -f "$compose_plugin_path" ]]; then
+            export CLP_DOCKER_PLUGIN_DIR="$dir"
+            break
+        fi
+    done
+    if [[ -z "${CLP_DOCKER_PLUGIN_DIR:-}" ]]; then
+        >&2 echo "Warning: Docker plugin directory not found;" \
+             "Docker Compose may not work inside container."
```
kirkrodrigues:
```suggestion
            "Docker Compose may not work inside container."
```

kirkrodrigues (commented):
Reviewed the shell scripts.

components/package-template/src/sbin/.common-env.sh
junhaoliao: that's the GID of the `docker` group on one of my computers. i do notice the group numbers can be different, depending on whether i have had other services installed

components/clp-package-utils/clp_package_utils/scripts/archive_manager.py
```diff
@@ -281,7 +283,8 @@ def main(argv: List[str]) -> int:
         logger.debug(f"Docker command failed: {shlex.join(cmd)}")
 
     # Remove generated files
-    generated_config_path_on_host.unlink()
+    resolved_generated_config_path_on_host = resolve_host_path_in_container(generated_config_path_on_host)
```
kirkrodrigues:
```suggestion
    resolved_generated_config_path_on_host = resolve_host_path_in_container(
        generated_config_path_on_host
    )
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -166,8 +170,10 @@ def _set_up_env_for_queue(self) -> EnvVarsDict:
         logs_dir = self._clp_config.logs_directory / component_name
         validate_queue_config(self._clp_config, logs_dir)
 
-        logs_dir.mkdir(exist_ok=True, parents=True)
-        _chown_paths_if_root(logs_dir)
+        resolved_logs_dir = resolve_host_path_in_container(logs_dir)
+
```
kirkrodrigues:
```suggestion
```

For consistency.

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -426,8 +443,16 @@ def _set_up_env_for_webui(self, container_clp_config: CLPConfig) -> EnvVarsDict:
         server_settings_json_path = (
             self._clp_home / "var" / "www" / "webui" / "server" / "dist" / "settings.json"
         )
+        resolved_client_settings_json_path = resolve_host_path_in_container(
+            client_settings_json_path
+        )
+        resolved_server_settings_json_path = resolve_host_path_in_container(
+            server_settings_json_path
+        )
```
kirkrodrigues: Can we move these closer to their use?

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -544,7 +569,8 @@ def _set_up_env_for_mcp_server(self) -> EnvVarsDict:
 
         logs_dir = self._clp_config.logs_directory / component_name
         validate_mcp_server_config(self._clp_config, logs_dir)
-        logs_dir.mkdir(parents=True, exist_ok=True)
+        resolved_logs_dir = resolve_host_path_in_container(logs_dir)
```
kirkrodrigues, line 572:
```suggestion

        resolved_logs_dir = resolve_host_path_in_container(logs_dir)
```
For consistency.

components/clp-package-utils/clp_package_utils/general.py
kirkrodrigues: NTS:

        if config_file_path != default_config_file_path:
            # This is going to print the wrong path
            raise ValueError(f"Config file '{config_file_path}' does not exist.")

components/clp-py-utils/clp_py_utils/core.py
```diff
@@ -62,6 +64,33 @@ def read_yaml_config_file(yaml_config_file_path: pathlib.Path):
...
+
+    :param host_path: The host path.
+    :return: The translated path.
+    """
+    host_path = host_path.absolute()
+    translated_path = CONTAINER_DIR_FOR_HOST_ROOT / host_path.relative_to("/")
+
+    try:
+        if not translated_path.is_symlink():
+            return translated_path
```
kirkrodrigues, line 80:
```suggestion
            return translated_path

```

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -53,30 +53,38 @@ def main(argv):
     try:
         # Validate and load config file.
         config_file_path = pathlib.Path(parsed_args.config)
-        clp_config = load_config_file(config_file_path, default_config_file_path, clp_home)
+        clp_config = load_config_file(
```
kirkrodrigues, line 56: Inside `load_config_file`, we print one of the arguments in a log, which I think is wrong, right? I'm fine leaving it as is and opening an issue.

components/clp-package-utils/clp_package_utils/scripts/archive_manager.py
kirkrodrigues: Sorry, can you explain more?

components/package-template/src/sbin/.common-env.sh
kirkrodrigues: So should we pick another number or is 999 the first value that gets assigned to services?

tools/deployment/package/docker-compose.runtime.yaml
```diff
@@ -0,0 +1,31 @@
...
+  clp-runtime:
+    hostname: "clp_runtime"
+    image: "${CLP_PACKAGE_CONTAINER_IMAGE_REF:-clp-package}"
+    logging:
+      driver: "local"
+    stdin_open: true
+    tty: true
+    user: "${CLP_FIRST_PARTY_SERVICE_UID_GID:-1000:999}"
+    environment:
+      HOME: "${HOME:?Please set a value.}"
```
kirkrodrigues, line 14: Why do we need this?

tools/deployment/package/docker-compose.runtime.yaml
```diff
@@ -0,0 +1,31 @@
...
+      CLP_HOME: "${CLP_HOME:?Please set a value.}"
+      CLP_PWD_HOST: "${CLP_PWD_HOST:-${PWD}}"
+    volumes:
```
kirkrodrigues, line 18: Shall we organize these into groups? E.g., Docker-related mounts and other mounts?

tools/deployment/package/docker-compose.runtime.yaml
junhaoliao: we have to forward the running shell's HOME into the bootstrap container, or any `~` in pathlib.Path objects cannot be correctly resolved. e.g. the aws auth directory

components/clp-package-utils/clp_package_utils/scripts/archive_manager.py
junhaoliao: sorry, that comment was so unclear that it took me time to recall why it was done - 

on the host, clp_home is essentially the clp-package directory root. previously we were mounting the package root into the container so the python modules in `lib/python3/site-packages` can be accessed in the container. however, we have now removed such contents from the package (recall "now the package is only 180K"); if we mount an empty directory into `/opt/clp` in the container, anything in `/opt/clp` is essentially gone. we should just use the assets in the docker image and remove such mounts then

components/package-template/src/sbin/.common-env.sh
junhaoliao: the value may vary across various platforms - https://github.com/moby/moby/issues/21184 i think picking any has the same effects

on the other hand, `getent` is part of the glibc utils, so we expect it to be available in most distros (hopefully)

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
junhaoliao: Right, the log would be confusing to general users when `load_config_file` is used in the "bootstrap" scripts. I didn't modify `load_config_file` directly because the "native" scripts also uses `load_config_file` - it's possible to use a flag to determine whether the method is called in a bootstrap script or a native script though.

@coderabbitai please create an issue for the inaccurate log and we will fix it as we clean up the bootstrap / native scripts in a future PR

tools/deployment/package/docker-compose.runtime.yaml
```diff
@@ -0,0 +1,34 @@
...
+  clp-runtime:
+    hostname: "clp_runtime"
+    image: "${CLP_PACKAGE_CONTAINER_IMAGE_REF:-clp-package}"
+    logging:
+      driver: "local"
+    stdin_open: true
+    tty: true
+    user: "${CLP_FIRST_PARTY_SERVICE_UID_GID:-1000:999}"
+    environment:
+      HOME: "${HOME:?Please set a value.}"
```
kirkrodrigues, line 14:
```suggestion
      # NOTE: We forward "$HOME" into the container so that if the user specified "~" anywhere in
      # their config, the container can resolve it to the relevant path on the host. As a result,
      # the container should not rely on "$HOME" or "~" pointing to a path inside the container.
      HOME: "${HOME:?Please set a value.}"

```

kirkrodrigues (approved):
For the PR title, how about:

```
feat(clp-package)!: Containerize the scripts in `sbin` via Docker Compose; Remove redundant assets from the package (resolves #1358).
```

components/package-template/src/sbin/.common-env.sh
```diff
@@ -0,0 +1,60 @@
...
+        compose_plugin_path="$compose_plugin_dir/docker-compose"
+        if [[ -f "$compose_plugin_path" ]]; then
+            export CLP_DOCKER_PLUGIN_DIR="$compose_plugin_dir"
+            break
+        fi
+    done
+    if [[ -z "${CLP_DOCKER_PLUGIN_DIR:-}" ]]; then
+        echo >&2 "Warning: Docker plugin directory not found;" \
+            "Docker Compose may not work inside container."
+    fi
```
junhaoliao, line 49: Plugin upgrades that are installed through Docker Desktop on Windows seems to create symlinks in `$HOME/.docker/cli-plugins`. The symlinks can link to some Windows host path, such as `/mnt/c/Users/junhao/.docker/modules/cli-plugins-wsl/docker-compose`. If we don't map the link target's parent path into the container, the compose plugin would not be found in the container.

components/package-template/src/sbin/admin-tools/archive-manager.sh
```diff
@@ -1,9 +1,12 @@
 #!/usr/bin/env bash
```
junhaoliao, line 1: we should
```
set -o errexit
set -o nounset
set -o pipefail
```
or
```
set -eu
set -o pipefail
```

so the scripts early exit on any error or on any unset variables, and make a pipeline fail if any command in it fails

@coderabbitai create an issue to track the improvement

components/clp-package-utils/clp_package_utils/scripts/compress.py
junhaoliao, line 1: we need to make sure the `.resolve()` calls in this script can properly resolve relative and symlinks in the host path space; otherwise, the "mounted_path"s might point to some invalid file in the Package services containers

tools/deployment/package/docker-compose.runtime.yaml
junhaoliao, line 1: host network mode should be used or the start_clp.py script cannot properly check what IP addresses it can bind to on the host.

@coderabbitai file an issue

tools/deployment/package/docker-compose.runtime.yaml
```diff
@@ -0,0 +1,38 @@
+services:
+  clp-runtime:
+    hostname: "clp_runtime"
+    image: "${CLP_PACKAGE_CONTAINER_IMAGE_REF:-clp-package}"
+    logging:
+      driver: "local"
+    stdin_open: true
+    tty: true
```
junhaoliao, line 8: it's found that when requesting to detach stdin from the shell using `&`, no output is seen in the console. e.g.,
```
cd build/clp-package
./sbin/start-clp.sh

# this command prints compression speed to the console
./sbin/compress.sh ~/samples/hive-24hr

# with the `&` added, no command output is observed in the console
./sbin/compress.sh ~/samples/hive-24hr &
```

i believe we should simply remove `tty: true` and `stdin_open: true` from the `docker-compose.runtime.yaml` to avoid allocating a TTY. however, removing those lines have no effects - when `&` is used, there's still no output printed to the console.

It was later discovered that the `docker compose run` command **overrides** the `tty` and `stdin_open` options defined in service configurations ([source](https://github.com/docker/compose/blob/c37ede62dbb6485c4d5e297db9134631e9b2095e/cmd/compose/run.go#L88-L89)), even when neither the `--no-TTY` nor the `--interactive` flag is specified. In other words, if these flags are omitted, `docker compose run` will automatically set `tty: true` when it detects a TTY-capable terminal, and will always set `stdin_open: true`, regardless of the values defined in `docker-compose.yaml`. (The [official documentation](https://github.com/docker/compose/blob/c37ede62dbb6485c4d5e297db9134631e9b2095e/docs/reference/compose_run.md#L73) is unfortunately confusing and misleading on this behavior.)

then i think we should remove these two lines (rather than changing them to `false`) in docker-compose.runtime.yaml to avoid confusions, and add `--no-TTY` to all of the sbin scripts.

@coderabbitai create an issue to track the fix

components/package-template/src/sbin/.common-env.sh
junhaoliao, line 1: Ctrl + C does not kill the native containers which are launched inside the runtime containers, which may leave a lot of long running containers for example when compression job does not report a completion status

@coderabbitai report an issue

PR #1515, opened 2025-10-28
fix(clp-package): Update `mcp_server` config in `etc/clp-config.yml` to match defaults in `CLPConfig`.

rishikeshdevsot:
# Description

The current default configuration for `mcp-server` in `clp-config.yml` is:
```
#mcp-server:
# host: "localhost"
# port: 8000
# logging_level: "INFO"
```
which does not respect the default specification in `CLPConfig` which is `None`.
This PR changes the default to:
```
#mcp-server: null
```

making the default configuration for `mcp-server` consistent with the default configuration in `CLPConfig`.

The defaults of `etc/clp-config.yml` and `CLPConfig` need to be consistent, since a fully commented `etc/clp-config.yml` makes the CLP package use `CLPConfig`’s defaults
Thus, this PR makes CLP Package's behaviour when `etc/clp-config.yml` is uncommented (but unmodified)  to be same as when `etc/clp-config.yml` is left commented. 

# Checklist

  breaking change.

# Validation performed

## Release Notes

* **Chores**
  * Updated server configuration settings to disable a backend service component.

### DIFF
--- a/components/core/src/clp/EncodedVariableInterpreter.hpp
+++ b/components/core/src/clp/EncodedVariableInterpreter.hpp
@@ -9,6 +9,7 @@
 #include <string_utils/string_utils.hpp>
 
 #include "ffi/ir_stream/decoding_methods.hpp"
+#include "ir/EncodedTextAst.hpp"
 #include "ir/LogEvent.hpp"
 #include "ir/types.hpp"
 #include "LogTypeDictionaryEntryReq.hpp"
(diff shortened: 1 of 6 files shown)


rishikeshdevsot:
> 
> ```
> fix(clp-package): Update `mcp_server` config in `etc/clp-config.yml` to match defaults in `CLPConfig`.
> ```
> 
> Similarly, it's worth updating the PR description to indicate that the reason for the change is because `etc/clp-config.yml` didn't respect the defaults specified in `CLPConfig`. When the `etc/clp-config.yml` is fully commented, the CLP package will use the defaults from `CLPConfig`. When `etc/clp-config.yml` is uncommented, we want the CLP package's behaviour to remain identical (unless, of course, the user changes some config).

addressed.

PR #1522, opened 2025-10-29
refactor(deployment): Refactor Package Docker Compose files using service-level compositions.

sitaowang1998:
# Description

This PR improves docker compose file extensibility by:
- Moving all services' default definitions into a `docker-compose-all.yaml`.
- Using `extends` to include default services in the docker compose files.

This PR also includes minor improvements:
- Renames `docker-compose.base.yaml` to `docker-compose-base.yaml`.
- Uses a separate function in `controller.py` to find the docker file name.

# Checklist

  breaking change.

# Validation performed

* [x] Compression and search work on full deployment.
* [x] Containers up successfully on base deployment.
* [x] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Added query scheduler and query worker services for distributed query processing capabilities.
  * Added reducer service for result aggregation and processing.
  * Added MCP server service for enhanced protocol integrations.

* **Refactor**
  * Restructured deployment configuration for improved maintainability and modularity across services.


tools/deployment/package/docker-compose-base.yaml
junhaoliao, line 1: for these services, can we label with a profile in the "all.yaml"? like `profiles: ["base"]`

Then we don't need this "base.yaml". To launch bases services, we can just run `docker compose -f docker-compose.all.yaml --profile base up`

We should still keep the docker-compose.yaml file, there we reset the profiles of the services with the `!reset []` syntax, so we can just use `docker compose up` to launch the default set of services

tools/deployment/package/docker-compose.yaml
```diff
@@ -1,165 +1,80 @@
...
-  clp-config-readonly: &volume_clp_config_readonly
-    type: "bind"
-    source: "${CLP_LOGS_DIR_HOST:-./var/log}/.clp-config.yml"
-    target: "/etc/clp-config.yml"
-    read_only: true
-  clp-logs: &volume_clp_logs
-    type: "bind"
-    source: "${CLP_LOGS_DIR_HOST:-./var/log}"
-    target: "/var/log"
+volumes:
```
junhaoliao: Can we extract this into a docker-compose.volumes.yaml and let both "docker-compose.all.yaml" and "docker-compose.yaml" `include:` the volume yaml?

Then we don't have to duplicate this definition

junhaoliao (changes requested):

tools/deployment/package/docker-compose-base.yaml
junhaoliao: the reset syntax: https://docs.docker.com/reference/compose-file/merge/#reset-value

sitaowang1998: The whole point of this PR is to make the docker compose file easily extensible. Adding a base profile does not really help if we later introduces a new "profile". Then we need to introduce another profile, and maybe a new docker file as well with resets, making it hard to track how the variables are defined, and what docker compose file to use.

Using the extends approach, for each profile we have a single file listing all the services it needs and the override variables. This way is more structured than using the profile.

junhaoliao: right, that makes a lot of sense

tools/deployment/package/docker-compose.volume.yaml
```diff
@@ -0,0 +1,9 @@
+name: "clp-package-volumes"
+
```
junhaoliao: do we have to add the `name` since this file is not intended to be used as a complete project?
```suggestion
```

tools/deployment/package/docker-compose.utils.yaml
junhaoliao, line 1: (though it was me who proposed the `docker-compose.volume.yaml` name,)

do you think renaming this file as `volumes.yaml` or `common.yaml` would help clarify this yaml file is not intended to be used as a standalone docker compose project?

junhaoliao (approved):
the changes lgtm. i posted some stylistic comments

for the title, would this be better?
```
refactor(deployment): Refactor Package Docker Compose files using service-level compositions.
```

junhaoliao (approved):

PR #1523, opened 2025-10-29
feat(clp-rust-utils): Add `ClpIoConfig` and related inner configuration structs with Brotli-compressed MessagePack serialization support.

LinZhihao-723:
# Description

As required by CLP's compression job database, compression jobs must be submitted with `ClpIoConfig` to specify the input/output specification. This PR mirrors this config from `clp-py-utils`, with related inner configs defined. Since the config is serialized using msgpack and compressed using brotli, such a serialization method is implemented to support submission.

Notice that for inner configs, not all options are implemented. For example, the input type now only supports S3 as this Rust implementation will only be used for S3 compression for now. However, it is designed in a way so that other input types can be easily added. Same as S3 authentication configs.

The implementation should also support deserializing configs using Rust's serde framework. This feature can be added in future PRs (with validations added).

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests to ensure the serialized results match expectations.

## Summary by CodeRabbit

* **New Features**
  * Added S3 bucket and AWS credential configuration support.
  * Implemented Brotli-compressed MessagePack serialization for improved performance.
  * Extended configuration framework with comprehensive job input and output parameters.

* **Tests**
  * Added serialization tests validating Brotli+MessagePack output and JSON representation.

* **Bug Fixes**
  * Improved error reporting for serialization and I/O operations.

### DIFF
--- a/components/clp-rust-utils/src/job_config/clp_io_config.rs
+++ b/components/clp-rust-utils/src/job_config/clp_io_config.rs
@@ -9,7 +9,7 @@ pub struct ClpIoConfig {
     pub output: OutputConfig,
 }
 
-// An enum representing CLP input config.
+/// An enum representing CLP input config.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize)]
 #[serde(tag = "type")]
 pub enum InputConfig {


components/clp-rust-utils/src/clp_config/scheduler/clp_io_config.rs
```diff
@@ -0,0 +1,124 @@
...
+  "output": {
+    "tags": null,
+    "target_archive_size": 268435456,
+    "target_dictionaries_size": 33554432,
+    "target_encoded_file_size": 268435456,
+    "target_segment_size": 268435456,
+    "compression_level": 3
+  }
+}"#;
+        assert_eq!(expected_json, json_serialized);
```
hoophalab: Keys can be in a different order. How about converting `config` to `serde_json::Value` and set `expected_json` using json! literal/marco, and assert based on [`impl Eq for Value`](https://docs.rs/serde_json/latest/serde_json/value/enum.Value.html#impl-Eq-for-Value)? This way we should be able to compare json values directly rather than comparing strings.

components/clp-rust-utils/src/clp_config/scheduler/clp_io_config.rs
```diff
@@ -0,0 +1,124 @@
+use std::io::Write;
+
+use anyhow::Result;
+use brotli::CompressorWriter;
+use serde::Serialize;
+
+use crate::clp_config::scheduler::{InputConfig, OutputConfig};
+
+/// Represents CLP IO config.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
```
hoophalab: How about moving the contents `input_config.rs` and `output_config.rs` both in to `clp_io_config.rs`? I guess it's common to write a super large rust source file, and `InputConfig` and `OutputConfig` will never be used independently.

Same for aws structs in `s3_config.rs`.

components/clp-rust-utils/src/clp_config/scheduler.rs
```diff
@@ -0,0 +1,7 @@
+mod clp_io_config;
+mod input_config;
+mod output_config;
+
+pub use clp_io_config::*;
+pub use input_config::*;
+pub use output_config::*;
```
hoophalab: How about moving the contents `input_config.rs` and `output_config.rs` to `clp_io_config.rs`? `InputConfig` and `OutputConfig` will never be used independently.

I guess it's very common to write a super large rust source file, which keeps closely related types together rather than having them scattered across multiple modules.

We could also move `AwsAuthentication` to `s3_config.rs`.

components/clp-rust-utils/src/clp_config/scheduler/clp_io_config.rs
```diff
@@ -0,0 +1,124 @@
...
+        assert!(brotli_compressed_msgpack_result.is_ok());
+        let brotli_compressed_msgpack = brotli_compressed_msgpack_result.unwrap();
+        let expected = "1ba00100e4ffdf9f43284b650e496850ba5f1eeefb53844a05d074faa66eb23ebef2dc45638\
+                275e9c24cb3bccba29c9bfc9d95db42175d52eecc81793cb3bc3c4ed0bf604c56e5c9a24581d9e65080\
+                1fd7263a8fb774fa362adf02eecc5b9d99532b8be8be173f6b659a9538c6c56a15571bc9856e20d0267\
+                b1591599975a75cdeb2aea30b83c8b486f3a2b3a74b419d6f99db0742a1482603a9480912e1336f2780\
+                dd9c3391503a9205a89a755bfe2c0d3a6be4c98ef0489c0b7e7f2d50b85f8f6e671a54d5dc6fa16d1ac\
+                cbaaffc5c3f1fb140f21ba0dce6ff0e8bc5f2da3c58426a9947046ca3cf9a06c7c8219e25a6ad0c4c67\
+                b6aceb8c88c782293b";
+        assert_eq!(expected, hex::encode(brotli_compressed_msgpack));
```
hoophalab: purely nitpicking - feel free to keep it as is
```suggestion
        assert_eq!(hex::decode(expected).unwrap(), brotli_compressed_msgpack);
```

Is it better/faster to compare in vec?

components/clp-rust-utils/src/clp_config/scheduler/clp_io_config.rs
```diff
@@ -0,0 +1,124 @@
...
+    ///
+    /// * Forwards [`rmp_serde::to_vec_named`]'s errors on failure.
+    /// * Forwards [`std::io::Write::write_all`]'s errors on failure.
+    pub fn to_brotli_compressed_msgpack(&self) -> Result<Vec<u8>> {
+        let msgpack_data = rmp_serde::to_vec_named(self)?;
+        let mut brotli_compressor = CompressorWriter::new(Vec::new(), 4096, 5, 22);
+        brotli_compressor.write_all(&msgpack_data)?;
+        Ok(brotli_compressor.into_inner())
+    }
+}
```
hoophalab: This code is actually generic to everything that implements `serde::Serialize`, and we might need to create brotli compressed msgpack Vec from other serde structs in the future. e.g. [here](https://github.com/y-scope/clp/blob/56ebca1ea4dcf4bd50a20a35197dd7b502f42f86/components/job-orchestration/job_orchestration/scheduler/compress/partition.py#L86)
Also, the user of `ClpIoConfig` might not necessarily want to brotli compress it.

How about create a separate `brotli_msgpack.rs` and do something like this:

```
struct BrotliMsgpack {}

impl BrotliMsgpack {
    pub fn try_convert_to_brotli_compressed_msgpack<T: Serialize + ?Sized>(
        val: &T,
    ) -> Result<Vec<u8>> {
...
    }
}
```

or 
```
struct BrotliMsgpack {
    pub data: Vec<u8>,
}

impl BrotliMsgpack {
    pub fn try_from<T: Serialize + ?Sized>(val: &T) -> Result<Self> {
...
    }
}

```

components/clp-rust-utils/src/clp_config/scheduler/clp_io_config.rs
```diff
@@ -0,0 +1,124 @@
...
+                target_archive_size: 268_435_456,
+                target_dictionaries_size: 33_554_432,
+                target_encoded_file_size: 268_435_456,
+                target_segment_size: 268_435_456,
+            },
+        };
+
+        let brotli_compressed_msgpack_result = config.to_brotli_compressed_msgpack();
+        assert!(brotli_compressed_msgpack_result.is_ok());
+        let brotli_compressed_msgpack = brotli_compressed_msgpack_result.unwrap();
```
hoophalab: purely nitpicking - feel free to keep it as is

How about unwrapping w/o `assert is_some()`? The thread panics anyway.
```suggestion
        let brotli_compressed_msgpack = brotli_compressed_msgpack_result.unwrap();
```

components/clp-rust-utils/src/clp_config/scheduler/clp_io_config.rs
```diff
@@ -0,0 +1,124 @@
+use std::io::Write;
+
+use anyhow::Result;
```
hoophalab: I feel that it's generally not recommended to use `anyhow` in a library, as the user cannot handle different error types in distinct ways (without downcasting).

For example, if a function could return either a connection error or a Brotli compression error, the user might want to retry on the connection error and panic on the Brotli compression error. Using `anyhow::Result` discards this information entirely.

Instead, we could define a library-wide `enum Error { BrotliMsgpackSerialization }` and implement `std::error::Error` for this enum using derive marcos provided by `thiserror` crate.

hoophalab (commented):
Looks good to me in general. Some comments.

components/clp-rust-utils/src/clp_config/scheduler/clp_io_config.rs
LinZhihao-723: True, sure.
I remembered Rust JSON will preserve the key-value ordering. But I forgot it's only enabled if `preserve_order` is on: https://docs.rs/crate/serde_json/latest/features#preserve_order

components/clp-rust-utils/src/clp_config/scheduler/clp_io_config.rs
LinZhihao-723: When doing string comparison, the failure is easier to interpret: you can copy the expected/actual hex string to ChatGPT and ask it to translate to JSON. I'm not sure if it's that easy to do so with vector comparison.

components/clp-rust-utils/src/clp_config/scheduler/clp_io_config.rs
LinZhihao-723: Sure, that's a good head up, let me take a look.

components/clp-rust-utils/src/clp_config/scheduler/clp_io_config.rs
LinZhihao-723: kk sg. I used this in our private Rust ffi library.

components/clp-rust-utils/src/job_config/clp_io_config.rs
```diff
@@ -0,0 +1,44 @@
+use serde::Serialize;
+
+use crate::clp_config::S3Config;
+
+/// Represents CLP IO config.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub struct ClpIoConfig {
```
hoophalab, line 7:
```suggestion
pub struct IoConfig {
```

1. How about not using "Clp" in code at all except for the library names in `Cargo.toml`. Brand names such as `Clp` `MySql` `Mongo`  are usually only specified in the namespace.

For example, [mongodb](https://docs.rs/mongodb/latest/mongodb/struct.Client.html) and [aws s3](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/struct.Client.html) simply uses `Client` rather than `AwsClient`. 


2. Similarly, [aws_sdk_s3::config](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/config/index.html) is not called `aws_sdk_s3::aws_config`
We could also rename `src/clp_config` as `src/job_config` or simply `src/config`

If there are conflicts with other crates, the user can just use a fully specified name `clp::config:scheduler::io_config::IoConfig`

components/clp-rust-utils/src/serde.rs
hoophalab, line 1: Open to discussion:

Personally, I'm against using the name "utils". However, it doesn't make sense to create a named crate for each small shared functions. e.g. create `brotli-msgpack` crate only for this function...

Since we already call this crate as `clp-rust-utils`, then how about not creating another level of `utils` by putting `brotli_msgpack.rs` under `components/clp-rust-utils/src/` directly?

components/clp-rust-utils/tests/clp_config_test.rs
```diff
@@ -0,0 +1,91 @@
...
+    "target_encoded_file_size": 268435456,
+    "target_segment_size": 268435456,
+    "compression_level": 3
+  }
+}"#;
+    let actual: Value = serde_json::from_str(json_serialized.as_str())
+        .expect("The serialization result should be a valid JSON string.");
+    let expected: Value = serde_json::from_str(expected_json)
+        .expect("The expected serialization result should be a valid JSON string.");
+    assert_eq!(expected, actual);
```
hoophalab: Sorry, I didn't think about this last time. How about using `serde_json::json!`?
```suggestion
    let expected = serde_json::json!({
  "input": {
    "type": "s3",
    "bucket": "yscope",
    "region_code": "us-east-2",
    "key_prefix": "sample-logs/cockroachdb.clp.zst",
    "aws_authentication": {
      "type": "credentials",
      "credentials": {
        "access_key_id": "ACCESS_KEY_ID",
        "secret_access_key": "SECRET_ACCESS_KEY"
      }
    },
    "keys": null,
    "dataset": "test-dataset",
    "timestamp_key": "timestamp",
    "unstructured": false
  },
  "output": {
    "tags": null,
    "target_archive_size": 268435456,
    "target_dictionaries_size": 33554432,
    "target_encoded_file_size": 268435456,
    "target_segment_size": 268435456,
    "compression_level": 3
  }
});
    let actual: Value = serde_json::from_str(json_serialized.as_str())
        .expect("The serialization result should be a valid JSON string.");
    assert_eq!(expected, actual);
```

hoophalab (approved):
LGTM. Approved. There are some nitpicks. Feel free to keep them as-is.

components/clp-rust-utils/src/job_config/clp_io_config.rs
LinZhihao-723: I agree with all u said. However, this is to mirror Python's implementation [here](https://github.com/y-scope/clp/blob/0f24600c48770b5fa08ab5f494d3a0be9ff8f494/components/job-orchestration/job_orchestration/scheduler/job_config.py#L56). It would be better if we can rename these two as a pair in a future PR.

components/clp-rust-utils/src/serde.rs
LinZhihao-723: How about `serde`? We may need to add a deserialization method for `BrotliMsgpack` as well.

hoophalab: sure

components/clp-rust-utils/src/job_config/clp_io_config.rs
hoophalab: yeah... how about the second point, renaming the folder `clp-rust-utils/src/clp_config/` as `clp-rust-utils/src/job_config/`?

PR #1528, opened 2025-10-30
fix(clp-package): Use generic archive directory as mount target in `native/dataset_manager.py` to match path in metadata database (fixes #1526).

### LINKED ISSUE #1526: Archive deletion fails in dataset manager due to path mapping mismatch
## Description

The native dataset manager script (`native/dataset_manager.py`) is currently unable to remove archives because it hasn't been updated to use the new in-container path (`/var/data`) introduced in the Docker Compose migration.

## Root Cause

The script still relies on the previous direct host-to-container path mapping instead of the new container path structure.

## Related

- PR: https://github.com/y-scope/clp/pull/1178
- Comment: https://github.com/y-scope/clp/pull/1178#discussion_r2476634906
- Reported by: @junhaoliao

## Expected Fix

Update `native/dataset_manager.py` and related native scripts to use the correct in-container paths that align with the Docker Compose deployment model.

junhaoliao:
# Description

- Use explicit bind mount for FS archive output so the container sees the transformed path.
- Replace `mounts.archives_output_dir` with `DockerMount(DockerMountType.BIND, host_dir, container_dir)` derived from a deep-copied, container-transformed archive_output config.
- Simplify `_try_deleting_archives_from_fs` to take only `dataset_archive_storage_dir` and remove brittle top-level containment check.

# Checklist

  breaking change.

# Validation performed

```
task
cd build/clp-package

# configure storage and query engine as `clp-s` in etc/clp-config.yml

./sbin/start-clp.sh
./sbin/compress.sh --timestamp-key timestamp ~/samples/postgresql.jsonl 

./sbin/admin-tools/dataset-manager.sh del default
# 2025-10-30T07:20:27.248 INFO [dataset_manager] Deleted archives of dataset `default`.
# 2025-10-30T07:20:27.274 INFO [dataset_manager] Deleted dataset `default` from the metadata database
```

## Summary by CodeRabbit

* **Refactor**
  * Improved archive storage handling and container mount configuration for filesystem-based storage.
  * Simplified archive deletion logic and removed unnecessary validation checks.

### DIFF
--- a/components/clp-package-utils/clp_package_utils/controller.py
+++ b/components/clp-package-utils/clp_package_utils/controller.py
@@ -819,7 +819,7 @@ class DockerComposeController(BaseController):
         :return: Number of worker processes to run.
         """
         # This will change when we move from single to multi-container workers. See y-scope/clp#1424
-        return multiprocessing.cpu_count() // 2
+        return max(1, multiprocessing.cpu_count() // 2)
 
     def _get_docker_file_name(self) -> str:
         """


components/clp-package-utils/clp_package_utils/scripts/dataset_manager.py
```diff
@@ -136,7 +138,14 @@ def main(argv: List[str]) -> int:
         mounts.logs_dir,
     ]
     if clp_config.archive_output.storage.type == StorageType.FS:
-        necessary_mounts.append(mounts.archives_output_dir)
+        container_archive_output_config = container_clp_config.archive_output.model_copy(deep=True)
+        container_archive_output_config.storage.transform_for_container()
```
quinntaylormitchell, line 144: does this line really need to be here? if `container_clp_config` is the version of the config meant for use inside the container, then its `archive_output` paths should already be in container form (from when they get transformed [here](https://github.com/y-scope/clp/blob/f2da1dbc56533f8643aee569426a76c233b2c90d/components/clp-package-utils/clp_package_utils/general.py#L323)), so the paths in `container_archive_output_config` should already be transformed because it's a deep copy, right?

junhaoliao: right, we could have used `generate_docker_compose_container_config()` (NOTE: it's a different function than `generate_container_config()` which is the one currently used in `dataset_manager.py`) to generate a copy of the config and directly reference it, but to limit the memory impacts, though not significantly, i only make a copy of the `container_archive_output_config` and transform the copy.

`generate_container_config` will be deprecated when we move all docker orchestrions to Docker Compose so the changes here are temporary as well

let me know if my explanations should be clarified or if i misunderstood your comment

kirkrodrigues (approved):
For the PR title, how about:

```
fix(clp-package): Use generic archive directory as mount target in `native/dataset_manager.py` to match path in metadata database (fixes #1526).
```

PR #1531, opened 2025-10-30
fix(clp-mcp-server): Add missing `clp-py-utils` dependency needed for running unit tests.

20001020ycx:
# Description

`clp-mcp-server` relies on many constants, data structures, and util functions from clp_py_utils, e.g. `clp_py_utils.clp_config import CLPConfig`. However, this dependency is only resolved at run time after the entire clp package is built. This led to the problem when locally running `uv run --group dev pytest tests/* -vv`, the pytest fails because it cannot find the dependencies for imports from `clp_py_utils`. This PR adds support for this.

# Checklist

  breaking change.

# Validation performed

1. Being able to successfully run `uv run --group dev pytest tests/* -vv` locally without clp-package. 
2. Connecting Claude Code Agent to the mcp server in clp-package, and it returns a hello world message. The purpose of this test is to ensure there's no regression.

## Summary by CodeRabbit

* **Chores**
  * Added a local utility as an editable development dependency and configured local tooling to include it for development.
  * Introduced a new dependency group and reorganized dependency declarations to expand the default development groups for improved local setup and iterative work.

### DIFF
--- a/.github/workflows/clp-rust-checks.yaml
+++ b/.github/workflows/clp-rust-checks.yaml
@@ -25,13 +25,7 @@ concurrency:
 
 jobs:
   rust-checks:
-    strategy:
-      matrix:
-        os:
-          - "macos-15"
-          - "ubuntu-22.04"
-          - "ubuntu-24.04"
-    runs-on: "${{matrix.os}}"
+    runs-on: "ubuntu-22.04"
     steps:
       - uses: "actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683"
         with:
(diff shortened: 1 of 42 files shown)


components/clp-mcp-server/pyproject.toml
```diff
@@ -12,8 +12,12 @@ dependencies = [
     "msgpack>=1.1.2",
     "paginate>=0.5.7",
     "pymongo>=4.15.3",
+    "clp-py-utils",
```
junhaoliao: let's refer to the integration test to set up a `clp` group: https://github.com/y-scope/clp/blob/e2765115a99c1a754672c559cb2d44ab0f13f0ac/integration-tests/pyproject.toml#L19-L24

i.e.,
```
clp = [
    "clp-py-utils",
]
```

junhaoliao: then below we can add
```
[tool.uv]
default-groups = ["clp", "dev"]
```

junhaoliao (dismissed):

PR #1534, opened 2025-10-31
feat(api-server): Initial commit for CLP API server:

hoophalab:
# Description

This PR adds the API server. Only clp-json is supported for now.

1. Add the necessary serialization types for `clp-config.yml` and `credentials.yml` (This is moved to #1535)
2. Add a clp client which can be also used to create a cli app.
3. `/query` route: insert a compression job to jobs table
4. `/query_results` route: fetch results from the results_cache

# Checklist

  breaking change.

# Validation performed

1. start the package using

`CLP_HOME=<...>/build/clp-package RUST_LOG=trace cargo run --bin api_server`

or directly calling the binary in `/target/debug/`

`CLP_HOME=<...>/build/clp-package RUST_LOG=trace api_server`
2. Submit a query

```
POST http://localhost:3001/query

{"query_string": "pid:7891", "dataset": "default", "begin_timestamp": 1761923334000000}
```

Response: `{"query_results_uri":"/query_results/10"}`

3. Get results in the event stream format
`GET http://localhost:3001/query_results/10`
Response:
```
data: {...}

data: {...}
```

## Summary by CodeRabbit

* **New Features**
  * Added an API server with health, query submission and streaming (SSE) endpoints.
  * Added a client API to submit queries, poll status and stream results; query options include dataset, max results, timestamp range and case sensitivity.
  * Added YAML config loading utility.

* **Bug Fixes / Improvements**
  * Unified and improved error handling and reporting (includes YAML errors) and graceful shutdown.

* **Chores**
  * Updated AWS SDK versions and added serde_yaml dependency.

* **Other**
  * Extended config defaults (logs directory, polling/backoff and default max results).


components/api-server/src/client.rs
```diff
@@ -0,0 +1,256 @@
...
+    ///   `Killed`.
+    /// * Errors related to `self.get_status()` or `MongoDB` operations (e.g., connection, find).
+    /// * `ClientError::DecodeError`: If a retrieved document does not contain a "message" field, if
+    ///   the "message" field is not a BSON string, or if the JSON string fails to parse.
+    pub async fn fetch_results(
+        &self,
+        search_job_id: u64,
+    ) -> Result<impl Stream<Item = Result<serde_json::Value, ClientError>> + use<>, ClientError>
+    {
+        loop {
```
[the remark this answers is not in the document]
hoophalab: see https://blog.rust-lang.org/2024/09/05/impl-trait-capture-rules/

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,100 @@
...
+
+async fn query_results(
+    State(client): State<clp_client::Client>,
+    Path(search_job_id): Path<u64>,
+) -> Result<Sse<impl Stream<Item = Result<Event, HandlerError>>>, HandlerError> {
+    let stream = client.fetch_results(search_job_id).await?;
+    Ok(
+        Sse::new(stream.map(|res| Ok(Event::default().json_data(res?)?)))
+            .keep_alive(KeepAlive::default()),
+    )
```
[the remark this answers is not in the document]
hoophalab: client only returns ClientError, which packs serde_json error

hoophalab:
@junhaoliao All changes to `clp-rust-utils` are moved to #1535

components/api-server/src/client.rs
```diff
@@ -0,0 +1,256 @@
...
+    pub async fn connect(config: &Config, credentials: &Credentials) -> Result<Self, ClientError> {
+        let connect_options = sqlx::mysql::MySqlConnectOptions::new()
+            .host(&config.database.host)
+            .port(config.database.port)
+            .username(&credentials.database.user)
+            .password(&credentials.database.password)
+            .database(&config.database.name);
+        let sql_pool = sqlx::mysql::MySqlPoolOptions::new()
+            .connect_with(connect_options)
+            .await?;
```
LinZhihao-723: I will need this in log ingestor as well. Will extract it into clp-rust-utils based on #1535.

components/api-server/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "api-server"
+version = "0.1.0"
+edition = "2024"
+
+[lib]
+name = "clp_client"
+path = "src/lib.rs"
```
junhaoliao, line 8: question: do we have to define a `[lib]` in this Cargo project? as in, do we plan to reuse this lib in other projects at the moment?

junhaoliao: i guess it is useful if other Rust projects want to use the client without the HTTP server?

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,100 @@
...
+    routing::{get, post},
+};
+use clap::Parser;
+use clp_rust_utils::package_config;
+use futures::{Stream, StreamExt};
+use thiserror::Error;
+use tracing_subscriber::{self, prelude::*};
+
+#[derive(Parser)]
+#[command(version, about)]
```
junhaoliao: is the `about` needed here? i don't see a `description` field in `[package]` in `Cargo.toml`

maybe we can add a simple description into Cargo.toml?

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,100 @@
...
+struct Args {
+    #[arg(long)]
+    package_root: String,
```
junhaoliao: shall we call it `clp_home` to follow the convention in the CLP Package Python scripts?

---

in fact, we read such info from the env var `CLP_HOME` instead of from the command-line: https://github.com/y-scope/clp/blob/7c95e23fa5caf05ae3f3db56d90386f6fe868d3e/components/clp-package-utils/clp_package_utils/scripts/compress.py#L125

we do take `--config` from the arguments: https://github.com/y-scope/clp/blob/7c95e23fa5caf05ae3f3db56d90386f6fe868d3e/components/clp-package-utils/clp_package_utils/scripts/compress.py#L131-L137

were there reasons to deviate from the conventions used in the CLP Package Python scripts?

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,100 @@
...
+    let args = Args::parse();
+    let package_root = std::path::Path::new(&args.package_root);
+
+    let config_path = package_root.join(package_config::DEFAULT_CONFIG_FILE_PATH);
+    let config_string = std::fs::read_to_string(config_path)?;
+    let config: package_config::Config = serde_yaml::from_str(&config_string)?;
+
+    let credentials_path = package_root.join(package_config::DEFAULT_CREDENTIALS_FILE_PATH);
+    let credentials_string = std::fs::read_to_string(credentials_path)?;
+    let credentials: package_config::Credentials = serde_yaml::from_str(&credentials_string)?;
```
junhaoliao: would it be better to create helpers from those and maintain them in `clp-rust-utils`, so they become reusable for any project / binary that needs to parse clp config / credentials from files?

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,100 @@
...
+
+#[derive(Parser)]
+#[command(version, about)]
+struct Args {
+    #[arg(long)]
+    package_root: String,
+}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
```
junhaoliao, line 26: i suppose we use `anyhow::Result<()>` because we want the error type to be flexible?

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,100 @@
...
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::layer())
+        .with(tracing_subscriber::EnvFilter::from_default_env())
+        .init();
+
+    let args = Args::parse();
+    let package_root = std::path::Path::new(&args.package_root);
+
+    let config_path = package_root.join(package_config::DEFAULT_CONFIG_FILE_PATH);
+    let config_string = std::fs::read_to_string(config_path)?;
```
junhaoliao: if i'm not wrong, this is synchronous. shall we use `tokio::fs::read_to_string` instead?

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,100 @@
...
+enum HandlerError {
+    #[error("Internal server error")]
+    InternalServer,
+}
+trait IntoHandlerError {}
+impl IntoHandlerError for axum::Error {}
+impl IntoHandlerError for clp_client::ClientError {}
+impl<T: IntoHandlerError> From<T> for HandlerError {
+    fn from(_: T) -> Self {
+        Self::InternalServer
```
junhaoliao, line 127: is it intentional to swallow the error? shall we log the error somehow?

components/api-server/src/bin/api_server.rs
junhaoliao, line 26: shall we register the SIGTERM handler and perform graceful shutdowns? see https://docs.rs/axum/latest/axum/serve/struct.WithGracefulShutdown.html

components/api-server/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+
+[[bin]]
+name = "api_server"
+path  = "src/bin/api_server.rs"
+
+[dependencies]
+anyhow = "1.0"
+axum = { version = "0.8", features = ["json"] }
```
junhaoliao: shall we enable `tower-log` and do per request logging?

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,100 @@
...
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::layer())
+        .with(tracing_subscriber::EnvFilter::from_default_env())
+        .init();
+
+    let args = Args::parse();
+    let package_root = std::path::Path::new(&args.package_root);
+
```
junhaoliao, line 30: for the `?`s related to user actions like reading configs from the disk, shall we consider adding messages for the errors? maybe `anyhow::Context` can help?

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,100 @@
...
+    let config: package_config::Config = serde_yaml::from_str(&config_string)?;
+
+    let credentials_path = package_root.join(package_config::DEFAULT_CREDENTIALS_FILE_PATH);
+    let credentials_string = std::fs::read_to_string(credentials_path)?;
+    let credentials: package_config::Credentials = serde_yaml::from_str(&credentials_string)?;
+
+    let client = clp_client::Client::connect(&config, &credentials).await?;
+
+    let addr = format!("{}:{}", &config.api_server.host, &config.api_server.port);
+    let listener = tokio::net::TcpListener::bind(&addr).await?;
```
junhaoliao: (depending on the rationales, this might be trivial

shall we bind before creating the client (to connect to the db)? i think forgetting to shut down a previous server before starting a new one is a pretty common issue in development environments

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,100 @@
...
+        .route("/", get(root))
+        .route("/query", post(query))
+        .route("/query_results/{search_job_id}", get(query_results))
+        .with_state(client);
+    axum::serve(listener, app).await?;
+    Ok(())
+}
+
+async fn root() -> String {
+    "API server is running".to_owned()
```
junhaoliao, line 84: shall we (also?) bind this to a `/health` path? the `/health` path seems more conventional for checking health of a server

for the background, we need to do healthcheck in Docker Compose / K8s for automatic healing

components/api-server/src/bin/api_server.rs
junhaoliao: this seems a bit hard to read... 

can we extract
```
let event_stream = stream.map(|res| Ok(Event::default().json_data(res?)?));
```
then write it as
```
Ok(Sse::new(event_stream).keep_alive(KeepAlive::default()))
```

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,100 @@
...
+    let search_job_id = client.submit_query(query_config).await?;
+    let uri = format!("/query_results/{search_job_id}");
+    Ok(Json(serde_json::json!({"query_results_uri": uri})))
+}
+
+async fn query_results(
+    State(client): State<clp_client::Client>,
+    Path(search_job_id): Path<u64>,
+) -> Result<Sse<impl Stream<Item = Result<Event, HandlerError>>>, HandlerError> {
+    let stream = client.fetch_results(search_job_id).await?;
```
junhaoliao: see below comment for extracting `event_stream` -

would it be better to rename this as `results_stream`?

junhaoliao (changes requested):
reviewed the axum server code

i agree with the RESTful API endpoint designs in the code. as discussed offline, in a future PR, please consider integrating frameworks to generate OpenAPI specs to make the code more testable (maybe https://docs.rs/utoipa/latest/utoipa/ ?)

@LinZhihao-723 please help with the rest of the review as i observe and learn

components/api-server/Cargo.toml
LinZhihao-723: I think one of the reason to define it is to allow unused public symbols. It also helps when writing test cases.

components/api-server/src/client.rs
```diff
@@ -0,0 +1,265 @@
...
+            "SELECT status FROM `{QUERY_JOBS_TABLE_NAME}` WHERE id = ?"
+        ))
+        .bind(search_job_id)
+        .fetch_one(&self.sql_pool)
+        .await?;
+        let status: i32 = row.try_get("status")?;
+        Ok(QueryJobStatus::try_from(status)?)
+    }
+}
+
```
```suggestion
/// Defines the request configuration for submitting a search query.
```

components/api-server/src/client.rs
```diff
@@ -0,0 +1,265 @@
...
+            dataset: value.dataset,
+            query_string: value.query_string,
+            max_num_results: value.max_num_results,
+            begin_timestamp: value.begin_timestamp,
+            end_timestamp: value.end_timestamp,
+            ignore_case: value.ignore_case,
+            ..Default::default()
+        }
+    }
+}
```
LinZhihao-723: Can we move this type to the front of `Client`?

components/api-server/src/client.rs
```diff
@@ -0,0 +1,265 @@
...
+
+impl From<clp_rust_utils::Error> for ClientError {
+    fn from(value: clp_rust_utils::Error) -> Self {
+        match value {
+            clp_rust_utils::Error::MsgpackEncode(_) => Self::MalformedData,
+            clp_rust_utils::Error::Io(error) => error.into(),
+            clp_rust_utils::Error::Sqlx(error) => error.into(),
+        }
+    }
+}
```
LinZhihao-723: I think it's worth to make a dedicated file for errors to improve clarity. How about we move this into `client` module so the error is referred as `client::Error`?

components/api-server/src/client.rs
```diff
@@ -0,0 +1,265 @@
...
+
+    #[error("Query is not succeeded")]
+    QueryNotSucceeded,
+
+    #[error("IO error")]
+    Io(#[from] std::io::Error),
+
+    #[error("Malformed database")]
+    MalformedData,
+}
```
```suggestion
#[derive(Error, Debug)]
pub enum ClientError {
    #[error("`sqlx::Error`: {0}")]
    Sql(#[from] sqlx::Error),

    #[error("`mongodb::error::Error`: {0}")]
    Mongo(#[from] mongodb::error::Error),

    #[error("Query is not succeeded")]
    QueryNotSucceeded,

    #[error("`std::io::Error`: {0}")]
    Io(#[from] std::io::Error),

    #[error("Malformed data")]
    MalformedData,
}
```
- To be consistent with `clp-rust-utils` in terms of error definition.
- IO error needs `{0}`? Or u omitted it because the message will always be the same?
- "database" -> "data"

components/api-server/src/client.rs
```diff
@@ -0,0 +1,265 @@
...
+
+trait IsMalformedData {}
+impl IsMalformedData for serde_json::Error {}
+impl IsMalformedData for rmp_serde::encode::Error {}
+impl<Enum: TryFromPrimitive> IsMalformedData for num_enum::TryFromPrimitiveError<Enum> {}
```
LinZhihao-723: - Shall we move them below ClientError implementation?
- Adding empty lines to follow our standard:
```suggestion
/// Empty trait to mark errors that indicate malformed data.
trait IsMalformedData {}

impl IsMalformedData for serde_json::Error {}

impl IsMalformedData for rmp_serde::encode::Error {}

impl<Enum: TryFromPrimitive> IsMalformedData for num_enum::TryFromPrimitiveError<Enum> {}
```

components/api-server/src/client.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// # Errors
+    ///
+    /// Returns a `ClientError` if either the `MySQL` connection pool creation or
+    /// the `MongoDB` client initialization fails. This includes networking issues,
+    /// invalid credentials, or incorrect configuration settings.
+    ///
+    /// # Returns
+    ///
+    /// A `Result` which is `Ok(Self)` (the connected `Client` instance) on success,
+    /// or `Err(ClientError)` on failure.
```
```suggestion
    /// Factory method to create a new client with active connections to both MySQL and MongoDB
    /// databases.
    ///
    /// # Returns
    ///
    /// A newly created [`Client`] instance with active connections to both databases.
    ///
    /// # Errors
    ///
    /// Returns an error if:
    ///
    /// * Forwards [`create_mysql_pool`]'s errors on failure.
    /// * Forwards [`mongodb::Client::with_uri_str`]'s errors on failure.
```
To make it consistent with other docstrings. Key changes to follow our guidelines:
- Move `returns` section before `errors`.
- Drop the arg section. I think Rust APIs usually don't document the arguments to encourage self-explanatory code.
- Use `[]` to surround symbols. This allows IDEs to render a hyperline to the corresponding symbol.
- In `Returns` section, we only explain what's returned on success.
- In `Errors` section, we document any propagated errors using `Forwards xxx's return values` format.

components/api-server/src/client.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// A `Result` which is:
+    ///
+    /// * `Ok(u64)` - The unique ID of the newly created query job on success.
+    /// * `Err(ClientError)` - An error if the serialization fails or the database operation fails.
+    ///
+    /// # Errors
+    ///
+    /// This function can return a `ClientError` if:
+    /// * The `query_config` fails to be serialized into a `MessagePack` byte vector.
+    /// * The SQL execution against the database fails.
```
```suggestion
    /// Submits a search or aggregation query as a job.
    ///
    /// # Returns
    ///
    /// The unique ID of the newly created query job on success.
    ///
    /// # Errors
    ///
    /// Returns an error if:
    ///
    /// * Forwards [`rmp_serde::to_vec_named`]'s return values on failure.
    /// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
```
Same. We should avoid documenting types in the docstring as the type annotations should already indicate them.

components/api-server/src/client.rs
```diff
@@ -0,0 +1,265 @@
...
+    pub async fn submit_query(&self, query_config: QueryConfig) -> Result<u64, ClientError> {
+        let mut search_job_config: SearchJobConfig = query_config.into();
+        if search_job_config.dataset.is_none() {
+            search_job_config.dataset = match self.config.package.storage_engine {
+                StorageEngine::Clp => None,
+                StorageEngine::ClpS => Some("default".to_owned()),
+            }
+        }
+        if search_job_config.max_num_results == 0 {
+            search_job_config.max_num_results = 1000;
```
LinZhihao-723: Is this a hardcoded value? Do we want to allow a configurable default?

components/api-server/src/client.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// * `Err(ClientError)`: An error if the database query fails, the job ID is not found, or if
+    ///   the status value cannot be converted into a valid `QueryJobStatus`.
+    ///
+    /// # Errors
+    ///
+    /// This function can return a `ClientError` for reasons including:
+    /// * **Database Errors:** Failure to execute the SQL query (e.g., connection issues).
+    /// * **Row Not Found:** If no job with the given `search_job_id` exists.
+    /// * **Data Conversion Errors:** If the retrieved status integer is invalid or cannot be
+    ///   converted to `QueryJobStatus`.
```
```suggestion
    /// Retrieves the status of a previously submitted search job.
    ///
    /// # Returns
    ///
    /// The current status of the job on success.
    ///
    /// # Errors
    ///
    /// Returns an error if:
    ///
    /// * Forwards [`sqlx::query::Query::fetch_one`]'s return values on failure.
    /// * Forwards [`sqlx::Row::try_get`]'s return values on failure.
    /// * Forwards [`QueryJobStatus::try_from`]'s return values on failure.
```

components/api-server/src/client.rs
```diff
@@ -0,0 +1,265 @@
...
+                    break;
+                }
+                QueryJobStatus::Failed | QueryJobStatus::Cancelled | QueryJobStatus::Killed => {
+                    return Err(ClientError::QueryNotSucceeded);
+                }
+                QueryJobStatus::Running | QueryJobStatus::Pending | QueryJobStatus::Cancelling => {
+                    tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
+                }
+            }
+        }
```
LinZhihao-723, line 165:
```suggestion
        let mut sleep_ms: u64 = 100;
        let max_sleep_ms: u64 = 5000;
        loop {
            match self.get_status(search_job_id).await? {
                QueryJobStatus::Succeeded => {
                    break;
                }
                QueryJobStatus::Failed | QueryJobStatus::Cancelled | QueryJobStatus::Killed => {
                    return Err(ClientError::QueryNotSucceeded);
                }
                QueryJobStatus::Running | QueryJobStatus::Pending | QueryJobStatus::Cancelling => {
                    tokio::time::sleep(tokio::time::Duration::from_millis(sleep_ms)).await;
                    sleep_ms = std::cmp::min(sleep_ms.saturating_mul(2), max_sleep_ms);
                }
            }
        }
```
How about an exponential backoff?

components/clp-rust-utils/Cargo.toml
```diff
@@ -4,9 +4,9 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
-aws-config = "1.8.8"
-aws-sdk-s3 = "1.106.0"
-aws-sdk-sqs = "1.86.0"
+aws-config = "1.8.10"
+aws-sdk-s3 = "1.110.0"
+aws-sdk-sqs = "1.88.0"
```
LinZhihao-723, line 9: If u don't upgrade these dependencies would the workflow fail? (I should updated the workflow to allow legal but not latest locked versions...

components/api-server/src/client.rs
```diff
@@ -0,0 +1,265 @@
...
+
+        let mapped = cursor.map(|res| {
+            let doc = res?;
+            let Some(msg) = doc.get("message") else {
+                return Err(ClientError::MalformedData);
+            };
+            let mongodb::bson::Bson::String(message) = msg else {
+                return Err(ClientError::MalformedData);
+            };
+            let value: serde_json::Value = serde_json::from_str(message)?;
```
LinZhihao-723: iiuc, we deserialize the JSON string into a map and then serialize it back to a string in `api_server::query_results`. Can we directly return a string and serialize it as a raw JSON string then?

components/api-server/src/client.rs
```diff
@@ -0,0 +1,265 @@
...
+    ///   or if the job status is ultimately `Failed`, `Cancelled`, or `Killed`.
+    ///
+    /// # Errors
+    ///
+    /// This function can return the following errors wrapped in `ClientError`:
+    /// * `ClientError::QueryNotSucceeded`: If the search job status is `Failed`, `Cancelled`, or
+    ///   `Killed`.
+    /// * Errors related to `self.get_status()` or `MongoDB` operations (e.g., connection, find).
+    /// * `ClientError::DecodeError`: If a retrieved document does not contain a "message" field, if
+    ///   the "message" field is not a BSON string, or if the JSON string fails to parse.
```
```suggestion
    /// Asynchronously fetches the results of a completed search job.
    ///
    /// # Returns
    ///
    /// A stream of the job's results on success. Each item in the stream is a [`Result`] that:
    ///
    /// ## Returns
    ///
    /// A parsed JSON value representing a search result on success.
    ///
    /// ## Errors
    ///
    /// Returns an error if:
    ///
    /// * [`ClientError::MalformedData`] if a retrieved document does not contain a "message" field,
    ///   or if the "message" field is not a BSON string.
    /// * Forwards [`mongodb::error::Error`] produced by the MongoDB cursor item access.
    /// * Forwards [`serde_json::from_str`]'s return values on failure.
    ///
    /// # Errors
    ///
    /// Returns an error if:
    ///
    /// * Forwards [`Client::get_status`]'s return values on failure.
    /// * Forwards [`mongodb::Collection::find`]'s return values on failure.
```

LinZhihao-723 (changes requested):
Reviewed client implementation. Mostly style-related suggestions.

components/api-server/src/client.rs
junhaoliao: we have never allowed this value to be configurable in the webui server, though i think we should

either way, i think we should extract a constant from this to avoid magic numbers?

components/api-server/Cargo.toml
hoophalab: I feel another benefit is that a separate client library enables us to create an alternative to `search.py` -- simply create a clp_client, submit_query and print fetch results to stdout.

components/api-server/src/bin/api_server.rs
hoophalab: Yes, this is synchronous. I feel it's okay to be synchronous here, otherwise we need to create both sync and async version for the util function to read yaml file.

components/api-server/src/bin/api_server.rs
hoophalab: yes, basically we need to convert all errors into a tuple of status code + response body.

This swallow all possible errors in the program into `InternalServer`, and we return a 500 + empty body for `InternalServer`

components/api-server/src/client.rs
hoophalab: I'm a bit worried that `#[error(...)]`  actually becomes the string for `impl Display`, and we probably shouldn't expose namespace information `std::io::Error` in `Display`?

The latter two points are fixed.

components/clp-rust-utils/Cargo.toml
hoophalab: it doesn't fails the  workflow, but `lint:fix-rust` automatically updates the version

components/api-server/src/client.rs
hoophalab: I'm open to both approaches. just add my thoughts:

If we treat the client as a separate library, it might be more reasonable to return a serde_json::Value instead of a JSON string.

While this involves a performance overhead in the api_server binary(converting string -> JSON value -> string), it also offers the benefit of validating the string upon conversion.

components/api-server/src/client.rs
hoophalab: I'll add config fields to clp-config.yaml

components/api-server/src/bin/api_server.rs
hoophalab: I was only using `anyhow::Result<()>` to pretty print the error message, but the new version also uses anyhow to add context as you suggested.

components/api-server/src/client.rs
LinZhihao-723: I prefer to directly return the JSON string.  The overhead of parsing the string to construct a map is non-negligible, especially when the search returns too many results. If we want JSON validation, we could use validators to avoid constructing the map.

components/api-server/Cargo.toml
hoophalab: We actually already have per request logging by setting log level to `RUST_LOG=trace`.
```
... TRACE axum::serve: connection 127.0.0.1:54218 accepted
```
However, it just doesn't log the request json body, and we might need tower-log to log the request json body. I'm wondering do we really need this feature?

components/api-server/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+mongodb = "3.3.0"
+num_enum = "0.7.5"
+rmp-serde = "1.3.0"
+serde = { version = "1.0.228", features = ["derive"] }
+serde_json = "1.0.145"
+sqlx = { version = "0.8.6", features = ["runtime-tokio", "mysql"] }
+thiserror = "2.0.17"
+tokio = { version = "1.48.0", features = ["full"] }
+tracing = "0.1.41"
+tracing-subscriber = { version = "0.3.20", features = ["json", "env-filter"] }
```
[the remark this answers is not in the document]
hoophalab: see https://github.com/y-scope/clp/pull/1534#discussion_r2491919405

components/clp-rust-utils/src/error.rs
```diff
@@ -10,4 +10,7 @@ pub enum Error {
 
     #[error("`sqlx::Error`: {0}")]
     Sqlx(#[from] sqlx::Error),
+
+    #[error("`serde_yaml::Error`: {0}")]
+    SerdeYaml(#[from] serde_yaml::Error),
```
LinZhihao-723: Can we order this before `Sqlx` to follow the lexical order?

components/clp-rust-utils/src/serde/yaml_file.rs
```diff
@@ -0,0 +1,23 @@
+use serde::de::DeserializeOwned;
+
+use crate::Error;
+
+/// Reads a YAML configuration file from a specified path and deserializes its contents
+/// into the requested type `T`.
```
LinZhihao-723:
```suggestion
/// Reads a YAML file from a specified path and deserializes its contents into the requested type
/// `T`.
```
It doesn't necessarily to be a config file, right?

components/clp-rust-utils/src/serde/yaml_file.rs
```diff
@@ -0,0 +1,23 @@
...
+/// The deserialized value of type `T` on success.
+///
+/// # Errors
+///
+/// Returns an [`Error`] if:
+///
+/// * Forwards [`std::fs::File::open`]'s return values on failure (e.g., file not found, permission
+///   denied).
+/// * Forwards [`serde_yaml::from_reader`]'s return values on failure (e.g., invalid YAML format,
+///   data does not match the structure of type `T`).
```
LinZhihao-723:
```suggestion
/// * Forwards [`std::fs::File::open`]'s return values on failure.
/// * Forwards [`serde_yaml::from_reader`]'s return values on failure.
```
I don't think we need `e.g` section.

hoophalab:
@junhaoliao @LinZhihao-723 What do you think about renaming `/query` to `/submit_query`? I feel like `/query` on its own implies you should get the results back immediately, which might not be what's happening.

components/clp-rust-utils/src/serde/yaml.rs
LinZhihao-723, line 1: How about `yaml`?

LinZhihao-723:
> @junhaoliao @LinZhihao-723 What do you think about renaming `/query` to `/submit_query`? I feel like `/query` on its own implies you should get the results back immediately, which might not be what's happening.

I agree with `submit_query`.

components/api-server/src/client.rs
LinZhihao-723: I think it's fine to expose the source error via namespace annotations.

components/clp-rust-utils/src/serde/yaml.rs
hoophalab: hmm I feel yaml should be something like serde_yaml. It converts from a str in yaml format to a struct. What we actually do here is to add an extra layer of reading from file.

Maybe we shouldn't add this util function? It's a bit trival.

LinZhihao-723: So when u refer to this utils, it's like `clp_rust_utils::serde::yaml`, right? So it is `serde_yaml` in some sense, lol

components/api-server/src/client.rs
```diff
@@ -0,0 +1,210 @@
...
+            begin_timestamp: value.begin_timestamp,
+            end_timestamp: value.end_timestamp,
+            ignore_case: value.ignore_case,
+            ..Default::default()
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct Client {
```
LinZhihao-723, line 46: I guess u're using `Client` for CLP client. After further thinking this class isn't exactly a client...
How about `Middleware` suggested by @kirkrodrigues? And the lib should be named to `clp_middleware`.

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,126 @@
...
+        .route("/query", post(query))
+        .route("/query_results/{search_job_id}", get(query_results))
+        .with_state(client);
+
+    tracing::event!(tracing::Level::INFO, "Server started at {addr}");
+    axum::serve(listener, app)
+        .with_graceful_shutdown(async {
+            tokio::signal::ctrl_c()
+                .await
+                .expect("failed to listen for event");
```
LinZhihao-723, line 77:
```suggestion
                .expect("Failed to listen for event");
```

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,126 @@
...
+        let message = res?;
+        let trimmed_message = message.trim();
+        if trimmed_message.lines().count() != 1 {
+            return Err(HandlerError::InternalServer);
+        }
+        Ok(Event::default().data(trimmed_message))
+    });
+    Ok(Sse::new(event_stream).keep_alive(KeepAlive::default()))
+}
+
```
LinZhihao-723, line 111:
```suggestion

/// Generic errors for request handlers.
```

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,126 @@
...
+impl IntoHandlerError for clp_client::ClientError {}
+
+impl<T: IntoHandlerError> From<T> for HandlerError {
+    fn from(_: T) -> Self {
+        Self::InternalServer
+    }
+}
+
+impl IntoResponse for HandlerError {
```
LinZhihao-723, line 132:
```suggestion
/// Converts [`HandlerError`] into an HTTP response.
impl IntoResponse for HandlerError {
```

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,126 @@
...
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::layer())
+        .with(tracing_subscriber::EnvFilter::from_default_env())
+        .init();
+
+    let _ = Args::parse();
+    let home = std::env::var("CLP_HOME").context("Expect a CLP_HOME env variable")?;
```
```suggestion
    let home = std::env::var("CLP_HOME").context("Expect `CLP_HOME` env variable")?;
```

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,126 @@
...
+        .context("Cannot connect to CLP")?;
+
+    let app = Router::new()
+        .route("/", get(health))
+        .route("/health", get(health))
+        .route("/query", post(query))
+        .route("/query_results/{search_job_id}", get(query_results))
+        .with_state(client);
+
+    tracing::event!(tracing::Level::INFO, "Server started at {addr}");
```
```suggestion
    tracing::info!("Server started at {addr}");
```
How about using `info!` directly?

components/api-server/Cargo.toml
LinZhihao-723, line 1: We should add a README for this component. I'm ok to defer it to future PRs, but we should ask coderabbit to create an issue to track this.

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,126 @@
...
+#[derive(Parser)]
+#[command(version, about = "API Server for CLP.")]
+struct Args {}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::layer())
+        .with(tracing_subscriber::EnvFilter::from_default_env())
+        .init();
```
LinZhihao-723: How about:
```rust
use tracing_appender::rolling::{RollingFileAppender, Rotation};

/// Initializes non-blocking tracing with hourly log rotation.
///
/// # Returns
///
/// A guard to the tracing logging. It must be held until program exit to ensure flushing.
fn init_non_blocking_tracing(log_dir: &str) -> tracing_appender::non_blocking::WorkerGuard {
    let file_appender = RollingFileAppender::new(Rotation::HOURLY, log_dir, "api_server.log");
    let (non_blocking_writer, guard) = tracing_appender::non_blocking(file_appender);

    tracing_subscriber::fmt()
        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
        .with_ansi(false)
        .with_writer(non_blocking_writer)
        .init();

    guard
}
```
With `tracing-appender` to support:
- Writing logs into a file instead of the console.
- Setup hourly log rotation (or daily for now).
- Logging in non-blocking manner: https://docs.rs/tracing-appender/latest/tracing_appender/non_blocking/index.html

components/api-server/src/bin/api_server.rs
hoophalab: I guess the message in Result::expect usually starts with a lower case, because it is formatted after a colon, iirc.

components/api-server/Cargo.toml
hoophalab: @coderabbitai can you create an issue on this?

components/api-server/src/client.rs
hoophalab: Open to suggestions, but here's my initial thinking:

It seems there are two ways to look at it. If the library is only useful to the API server, then Middleware is a fitting name.

However, because we could easily create something similar to `search.py` using this library, we might also consider it an internal client, which would be faster than a typical API client if the user has direct access to the database and file system.

For example, the [SeaweedFS module](https://github.com/seaweedfs/seaweedfs/blob/084b377f8786e3a4d98e0763c3e83be104a9b65e/weed/wdclient/masterclient.go#L106) that discovers volume nodes and builds a direct connecting URL is also named wdclient.

components/api-server/Cargo.toml
junhaoliao: i guess that's to say we don't need those for now. i'm fine if we want to keep it given we should add the tests soon

components/clp-rust-utils/src/clp_config/package/config.rs
```diff
@@ -46,13 +59,19 @@ impl Default for Database {
 pub struct ApiServer {
     pub host: String,
     pub port: u16,
+    pub query_job_pool_initial_delay_ms: u64,
+    pub query_job_pool_max_delay_ms: u64,
+    pub default_max_num_results: u32,
 }
```
LinZhihao-723, line 84: Sorry for missing this in the previous round. How about:
```suggestion
#[derive(Clone, Debug, Deserialize, Eq, PartialEq)]
#[serde(default)]
pub struct ApiServer {
    pub host: String,
    pub port: u16,
    pub query_job_polling: QueryJobPollingConfig,
    pub default_max_num_query_results: u32,
}

#[derive(Clone, Debug, Deserialize, Eq, PartialEq)]
pub struct QueryJobPollingConfig {
    #[serde(rename = "inital_backoff")]
    pub initial_backoff_ms: u64,

    #[serde(rename = "max_backoff")]
    pub max_backoff_ms: u64,
}
```
- I guess u mean "poll" not "pool".
- How about making a nested setup to mirror what we do in garbage collect config (notice that the serialized form doesn't have the unit "ms" in the config name): https://github.com/y-scope/clp/blob/d7c65f5b24a6d73988d693490fe771b372f37d46/components/package-template/src/etc/clp-config.yml#L117
- "default_max_num_results" -> "default_max_num_query_results" to make it more clear what "result" it refers to.

LinZhihao-723 (changes requested):
One last comment missed from previous rounds. Otherwise lgtm.

LinZhihao-723 (approved):
How about using a multi-line commit message and PR title:
```
feat(api-server): Initial commit for CLP API server:
* Add CLP client to send queries and retrieve results from clp-json.
* Define the package-level config schema for the API server.
* Implement an API server to forward incoming query requests to the underlying CLP client.
```

components/api-server/src/bin/api_server.rs
```diff
@@ -0,0 +1,136 @@
...
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    let _ = Args::parse();
+    let home = std::env::var("CLP_HOME").context("Expect `CLP_HOME` env variable")?;
+    let home = std::path::Path::new(&home);
+
+    let config_path = home.join(package::DEFAULT_CONFIG_FILE_PATH);
+    let config: package::config::Config = yaml::from_path(&config_path).context(format!(
+        "Config file {} does not exist",
```
junhaoliao, line 33: it appears that the log message is not so accurate because YAML schema mismatch can also cause failure:

```
[2026-01-05  15:20:01]	(test-clp-api-server-5f57bdf49d-nz7wr wait-for-db-table-creator)	job.batch/test-clp-db-table-creator condition met
[2026-01-05  15:20:02]	(test-clp-api-server-5f57bdf49d-nz7wr wait-for-results-cache-indices-creator)	job.batch/test-clp-results-cache-indices-creator condition met
[2026-01-05  15:21:42]	(test-clp-api-server-5f57bdf49d-nz7wr api-server)	Error: Config file /etc/clp-config.yaml does not exist
[2026-01-05  15:21:42]	(test-clp-api-server-5f57bdf49d-nz7wr api-server)	
[2026-01-05  15:21:42]	(test-clp-api-server-5f57bdf49d-nz7wr api-server)	Caused by:
[2026-01-05  15:21:42]	(test-clp-api-server-5f57bdf49d-nz7wr api-server)	    0: `serde_yaml::Error`: stream_output: unknown variant `profile`, expected `credentials` at line 85 column 3
[2026-01-05  15:21:42]	(test-clp-api-server-5f57bdf49d-nz7wr api-server)	    1: stream_output: unknown variant `profile`, expected `credentials` at line 85 column 3
```

can we improve the log message? @coderabbitai create an issue to track, and assign to @hoophalab

PR #1535, opened 2025-10-31
feat(clp-rust-utils): Add serde-compatible mirror structs for package and search job configs based on existing Python definitions.

hoophalab:
# Description

1. Add the necessary ser/de types for clp-config.yml and credentials.yml (only used fields are added for now)
2. Add the necessary ser/de types for search jobs in the sql database

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

Those structs works correctly in #1534 

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **Chores**
  * Introduced search/query job configuration and status types to the crate, enabling structured job records and status values.
  * Expanded the configuration package with components for package, database, API server, results cache and credentials, including sensible defaults and deserializable types.
  * Added enum-serialization support as a dependency to improve serialized enum handling.

PR #1538, opened 2025-11-01
feat(clp-rust-utils): Add utility function to create MySQL connection pools from CLP-package's database config.

LinZhihao-723:
# Description

NOTE: This PR depends on #1535.

* As the title suggests.

# Checklist

  breaking change.

# Validation performed

* [ ] Tested in #1534.

## Summary by CodeRabbit

* **New Features**
  * Added MySQL database connectivity with configurable connection pool support for improved database integration.

* **Chores**
  * Refined error handling for database operations and reorganized module structure for better API clarity.

### DIFF
--- a/components/clp-rust-utils/src/sqs.rs
+++ b/components/clp-rust-utils/src/sqs.rs
@@ -1,3 +1,4 @@
 mod client;
+pub mod event;
 
 pub use client::create_new_client;
--- a/components/clp-rust-utils/src/sqs/event.rs
+++ b/components/clp-rust-utils/src/sqs/event.rs
@@ -1,0 +1,3 @@
+mod s3;
+
+pub use s3::*;
--- a/components/clp-rust-utils/src/sqs/event/s3.rs
+++ b/components/clp-rust-utils/src/sqs/event/s3.rs
@@ -1,0 +1,38 @@
+use serde::Deserialize;
+
+/// Represents an Amazon S3 event notification message.
+///
+/// This struct models the essential fields required to process a single record from an S3 event
+/// notification. Its schema follows the AWS S3 event notification format version 2.1.
+///
+/// For more information, see:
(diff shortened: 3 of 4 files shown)


components/clp-rust-utils/src/database/mysql.rs
```diff
@@ -0,0 +1,36 @@
...
+/// # Errors
+///
+/// Returns an error if:
+///
+/// * Forwards [`sqlx::mysql::MySqlPoolOptions::connect_with`]'s errors on failure.
+pub async fn create_mysql_pool(
+    config: &DatabaseConfig,
+    credentials: &DatabaseCredentials,
+    max_connections: u32,
+) -> Result<sqlx::MySqlPool, crate::Error> {
```
hoophalab, line 24: I'm wondering whether to return the raw `sqlx::Error` directly or wrap it in our `clp_rust_utils::Error` type. My thinking is that returning `sqlx::Error` directly might simplify error handling for the function user, as they won't have to unpack or convert from `clp_rust_utils::Error`, but wrapping it in `clp_rust_utils::Error` might be a more common practice. What are your thoughts?

```suggestion
) -> Result<sqlx::MySqlPool, sqlx::Error> {
```

hoophalab (commented):
We need to resolve conflicts, but it looks good overall.
I'll test the function after the conflicts are resolved.

components/clp-rust-utils/src/database/mysql.rs
LinZhihao-723: Since this is a high-level API, I suspect users will actually need to look into the error and do some special handling. Instead, they might just log the error and exit. In this case, they probably don't need to understand the details of the error. Thus, I prefer to return a general library error to (1) hide the error details, (2) allow users to handle it as a general library error.

hoophalab (approved):

PR #1543, opened 2025-11-03
feat(webui): Increase precision of time range in guided UI from seconds to milliseconds. 

davemarco:
# Description

I noticed a small issue where the total results displayed in ui was slightly lower than the amount ingested when running query with "all time". I believe this is due to a quirk of dayJS.unix() where it floors the timestamp to the nearest second. This PR replaces it with dayJS.valueOf() which includes milliseconds, and now the total results matches logs ingested.

# Checklist

  breaking change.

# Validation performed

All time shows 1M results instead of slightly less than 1M with "all time"

## Summary by CodeRabbit

* **Refactor**
  * Improved timestamp handling in timeline and search queries to increase precision and accuracy of time-based filtering and retrieval, leading to more consistent query results across time ranges.

---
  - https://app.asana.com/0/0/1211827241306476

### DIFF
--- a/components/clp-package-utils/clp_package_utils/general.py
+++ b/components/clp-package-utils/clp_package_utils/general.py
@@ -396,7 +396,6 @@ def generate_container_start_cmd(
         "--rm",
         "--network", "host",
         "-w", str(CONTAINER_CLP_HOME),
-        "-e", f"PYTHONPATH={clp_site_packages_dir}",
         "-u", f"{os.getuid()}:{os.getgid()}",
         "--name", container_name,
         "--log-driver", "local"


hoophalab (changes requested):
Looks good, but when zooming in to only one log event, the timeline shows a bar for that log event, but result table is empty. We probably also need to do the same thing for `buildSearchQuery` to fix this.
```diff
--- a/components/webui/client/src/sql-parser/index.ts
+++ b/components/webui/client/src/sql-parser/index.ts
@@ -104,7 +104,7 @@ const buildSearchQuery = ({
     timestampKey,
 }: BuildSearchQueryProps): string => {
     let queryString = `SELECT ${selectItemList} FROM ${databaseName}
-WHERE to_unixtime(${timestampKey}) BETWEEN ${startTimestamp.unix()} AND ${endTimestamp.unix()}`;
+WHERE to_unixtime(${timestampKey}) BETWEEN ${startTimestamp.valueOf() / MILLISECONDS_PER_SECOND} AND ${endTimestamp.valueOf() / MILLISECONDS_PER_SECOND}`;
```

davemarco:
Okay good catch, i made change there as well

PR #1552, opened 2025-11-05
feat(log-ingestor): Add `Listener` to asynchronously receive and buffer S3 object metadata for compression job submission.

LinZhihao-723:
# Description

As the title suggests. This PR introduces `Listener` to receive S3 object metadata sent from the ingestor front end. The received metadata will be buffered by the `Buffer` object and eventually submitted by invoking a submitter that implements `BufferSubmitter` (not implemented in this PR). The buffer will be submitted if:
* A configurable size threshold is reached when a new object is received by the listener.
* A configurable timeout is reached (no objects received within a certain amount of time).
* A cancel signal is received.

# Checklist

* [x] This is a breaking change, and that has been indicated in the PR title, OR this isn't a
  breaking change.

# Validation performed

* [x] Add test cases to ensure the size-based submission and timeout-based submission work as expected.
* [x] The entire architecture (despite the implementation details being different) has been tested in the log-ingestor prototype.

## Summary by CodeRabbit

* **New Features**
  * Added a compression module to buffer and submit S3 object metadata via a reusable buffer and background listener, with a default listener capacity and a lightweight object-metadata type.

* **Tests**
  * Added integration tests covering concurrent senders, size-triggered submits, timeout-driven submission, and end-to-end submission counts.


components/log-ingestor/Cargo.toml
```diff
@@ -4,4 +4,8 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
+anyhow = "1.0.100"
+async-trait = "0.1.89"
 clp-rust-utils = { path = "../clp-rust-utils" }
+tokio = { version = "1", features = ["rt-multi-thread", "macros", "time"] }
```
hoophalab: Shall we include minor and patch version?

components/log-ingestor/src/compression/listener.rs
```diff
@@ -0,0 +1,140 @@
...
+    ) -> Self {
+        let (sender, receiver) = mpsc::channel(channel_capacity);
+        let task = ListenerTask {
+            buffer,
+            timeout,
+            receiver,
+        };
+        let cancel_token = CancellationToken::new();
+        let child_cancel_token = cancel_token.child_token();
+        let handle = tokio::spawn(async move { task.run(child_cancel_token).await });
```
hoophalab: how about
```suggestion
        let handle = tokio::spawn(async move { task.run(cancel_token.clone()).await });
```

components/log-ingestor/src/compression/listener.rs
```diff
@@ -0,0 +1,140 @@
...
+                    // TODO: Log the cancellation event when the logger has been integrated.
+                    self.buffer.submit().await?;
+                    return Ok(());
+                }
+
+                // New object metadata received.
+                Some(object_metadata) = self.receiver.recv() => {
+                    self.buffer.add(object_metadata).await?;
+                }
+
```
hoophalab, line 67: shall we early exist when self.receiver.recv() returns None

components/log-ingestor/src/compression/buffer.rs
```diff
@@ -0,0 +1,94 @@
+use anyhow::Result;
```
hoophalab, line 1: I guess we use anyhow because this is an application?

components/log-ingestor/tests/test_compression_listener.rs
```diff
@@ -0,0 +1,128 @@
...
+    (0..count)
+        .map(|i| ObjectMetadata {
+            bucket: bucket_name.to_string(),
+            key: format!("object-{i}"),
+            size: TEST_OBJECT_SIZE,
+        })
+        .collect()
+}
+
+#[tokio::test]
```
hoophalab, line 60: how about
```suggestion
#[tokio::test(start_paused = true)]
```
and add `test-util` feature of tokio to Cargo.toml

This way, we don't need to actually wait 4 secs when running unit test.

hoophalab (commented):
Looks good to me. There are some minor comments in terms of code.

I'm still feeling a bit uneasy about how we'll handle data persistence. Right now, we essentially have two buffers: the channel buffer and our buffer. Anything submitted to the channel buffer won't be written to disk until `select! { receiver.recv()}` runs.

I'm wondering if we should wrap Buffer in an arc-mutex and have the sender call `Buffer::submit` directly. However, this lock-based design will hurt performance.

Alternatively, we could look into implementing a data-persistent channel. Or even further, we maybe should just use rabbitmq as a data-persistent channel, and split the log ingestor into two microservice -- one listen to sqs and write object metadata to rabbitmq, and another service consumes object metadata and actually submits the job.

Anyway, the current code is all good to be merged.

components/log-ingestor/Cargo.toml
LinZhihao-723: yeah my bad, auto-filled by copilot, lol

LinZhihao-723:
> Looks good to me. There are some minor comments in terms of code.
> 
> I'm still feeling a bit uneasy about how we'll handle data persistence. Right now, we essentially have two buffers: the channel buffer and our buffer. Anything submitted to the channel buffer won't be written to disk until `select! { receiver.recv()}` runs.
> 
> I'm wondering if we should wrap Buffer in an arc-mutex and have the sender call `Buffer::submit` directly. However, this lock-based design will hurt performance. Alternatively, we could look into implementing a data-persistent channel. Or even further, we maybe should just use rabbitmq as a data-persistent channel?
> 
> Anyway, the current code is all good to be merged.

In our planning, we can persist the state using the following scheme:
* The ingestion front end (scanner-based or SQS-based) will onboard the input file to our system by writing the file entry to a dedicated DB table in one transaction. This step is required (at least for the scanner-based ingestor) to recover from a failure. There should be a state column, and it is initialized to `Buffered`, meaning that it should be now in the buffered state.
* The submitter will submit the buffer to create a compression job whenever a flush is triggered. Once the compression job is created, we will update the state column of submitted files to `Submitted` with the compression job ID. Technically, we can do this in one transaction, so it is atomic: either a job has been created with relevant files marked as `Submitted`, or nothing's submitted and the state is still `Buffered`.
* When the compression job finishes, it needs to update the table to mark the state as `Ingested`.
* When recovering from a failure:
  * Any file/object that has been onboarded by the ingestor frontend will be in state `Buffered`. So we can just walk through the DB table to find these entries and put them into the desired buffer.
  * If the atomic transaction mentioned above for submission is not implemented: we can still determine whether a job is an orphan job by checking whether the job ID attempts to update `Ingested` matches the submitted job ID. The buffer can be re-submitted and in that case the orphan job created by the previously failed ingestor won't be able to update the input file table as (1) the submitted job ID doesn't match; or (2) the state in the table is not `Submitted`.
  
That said, we don't need to explicitly persist any internal states for either the channel or the buffer. We rely on the DB update policy to ensure data persistence.

components/log-ingestor/src/compression/buffer.rs
LinZhihao-723: Yeah, this buffer should only be used in ingestor so let's uyse `anyhow`.

components/log-ingestor/src/compression/listener.rs
LinZhihao-723: Actually this doesn't work as it will move `cancel_token` to the capture...
image

components/log-ingestor/tests/test_compression_listener.rs
LinZhihao-723: As discussed offline, we will not use this feature for this particular test because `start_paused` only guarantees ordering for `sleep` calls across different coroutines. `Listener`'s implementation is using `select!` in a long-running loop, which complicates the execution model, and we can't find documentations for this type of behavior. We prefer to use real-time testing for now.

components/log-ingestor/src/compression/listener.rs
LinZhihao-723: Done. Return an error when None is received as the channel shouldn't be closed before calling shutdown_and_join.

components/log-ingestor/src/compression/listener.rs
```diff
@@ -0,0 +1,151 @@
...
+        channel_capacity: usize,
+    ) -> Self {
+        let (sender, receiver) = mpsc::channel(channel_capacity);
+        let task = ListenerTask {
+            buffer,
+            timeout,
+            receiver,
+        };
+        let cancel_token = CancellationToken::new();
+        let child_cancel_token = cancel_token.child_token();
```
hoophalab:
```suggestion
        let child_cancel_token = cancel_token.clone();
```
I see... Will this work? If child_token() is cancelled, the parent token will not be cancelled. This doesn't sound like a feature we need.

hoophalab (approved):
lgtm. one nitpick

components/log-ingestor/src/compression/listener.rs
LinZhihao-723: thx. Good catch.

hoophalab (approved):

Issue #1541, opened 2025-11-03
Improve kv-ir deserialization speed


LinZhihao-723:
### Request

The current kv-ir deserialization speed is not ideal. From our profiling experiment, around 30-40% of time is spent to construct/destruct the variable dictionary list (which is parsed and stored here: https://github.com/y-scope/clp/blob/main/components/core/src/clp/ffi/ir_stream/decoding_methods.cpp#L388)
We should come up a workaround to avoid this memory allocation overhead.

### Possible implementation

We should improve how `EncodedTextAst` is implemented. Instead of storing var strings as a vector of strings, we can store all var strings and the logtype in a concat string. For example:
```
logtype: "id=%s, passwd=%s" (
vars: ["x", "y"]
```
We can store it as:
```
string_buffer (as a string): ["xyid=%s, passwd=%s"]
```
with a possition vector to track how to partition substrings:
```
pos = [0, 1, 2]
```
In this way, we don't need to allocate strings for each var string and the logtype, and improve the spatial locality. In the meantime, we still preserve the capability to randomly access var strings or the logtype from the string buffer.

An early-stage experiment shows that this implementation leads to a 1.67x speedup, tested on two datasets.

Milestones:
* [ ] Implement the string buffer (StringBlob).
* [ ] Use the string buffer to re-implement the encoded text AST.
* [ ] Implement decoding methods based on the string buffer implementation.
* [ ] Replace the existing AST implementation by the new encoded text AST implementation.

PR #1561, opened 2025-11-05
feat(ffi): Add new `EncodedTextAst` implementation that uses `StringBlob` to reduce allocation overhead.

LinZhihao-723:
# Description

This is one of PR series to address #1541.

This PR adds a new `EncodedTextAst` implementation, using `StringBlob` to store dictionary variables and logtypes. Besides storage, it implements a generic decoding method similar to [this](https://github.com/y-scope/clp/blob/ec509bf4a083cd1c7d9cee21501fed748645cc18/components/core/src/clp/ffi/ir_stream/decoding_methods.inc#L13) as a member method to allow text decoding. Similarly, it implements a `to_string` method similar to [this](https://github.com/y-scope/clp/blob/ec509bf4a083cd1c7d9cee21501fed748645cc18/components/core/src/clp/ir/EncodedTextAst.hpp#L47) to decode and unparse the encoded text AST to its string form.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests to cover basic decoding and error handling behaviors.

## Summary by CodeRabbit

* **New Features**
  * Added a type-safe API to parse and render encoded text with embedded integers, floats and dictionary placeholders, plus optional unescaping.
  * Added a method to append strings to the string storage used by the encoded-text API.

* **Tests**
  * Added comprehensive unit tests covering decoding, round-trip fidelity, boundary cases and explicit error scenarios across supported encoded-variable types.

* **Bug Fixes**
  * Clarified user-facing error messages for missing components and trailing-escape conditions.


components/core/src/clp/ffi/EncodedTextAst.hpp
```diff
@@ -0,0 +1,264 @@
...
+ * A parsed and encoded unstructured text string.
+ * @tparam encoded_variable_t The type of encoded variables in the string.
+ */
+template <ir::EncodedVariableTypeReq encoded_variable_t>
+class EncodedTextAst {
+public:
+    // Factory function
+    /**
+     * @param encoded_vars
+     * @param string_blob
```
gibber9809:
```suggestion
     * @param string_blob A string blob containing a list of dictionary variables followed by a logtype.
```
I think the detail is necessary since you'd have to read the implementation to figure this out otherwise.

components/core/src/clp/ffi/EncodedTextAst.hpp
```diff
@@ -0,0 +1,264 @@
...
+    size_t m_num_dict_vars;
+};
+
+template <ir::EncodedVariableTypeReq encoded_variable_t>
+template <bool unescape_logtype>
+[[nodiscard]] auto EncodedTextAst<encoded_variable_t>::decode(
+        EncodedTextAstConstantHandlerReq auto&& constant_handler,
+        EncodedTextAstIntVarHandlerReq<encoded_variable_t> auto&& int_var_handler,
+        EncodedTextAstFloatVarHandlerReq<encoded_variable_t> auto&& float_var_handler,
+        EncodedTextAstDictVarHandlerReq auto&& dict_var_handler
```
gibber9809: Why rvalue references here? I think passing callbacks by value is more common + it's what the previous implementation does.

components/core/src/clp/ffi/EncodedTextAst.hpp
```diff
@@ -0,0 +1,264 @@
...
+                    constant_handler(logtype.substr(
+                            next_static_text_begin_pos,
+                            curr_pos - next_static_text_begin_pos
+                    ));
+                    // Skip the escape character
+                    next_static_text_begin_pos = curr_pos + 1;
+                }
+
+                // The character after the escape character is static text (regardless of whether it
+                // is a variable placeholder), so increment cur_pos by 1 to ensure we don't process
```
gibber9809:
```suggestion
                // is a variable placeholder), so increment curr_pos by 1 to ensure we don't process
```

components/core/src/clp/ffi/EncodedTextAstError.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include <cstdint>
+
+#include <ystdlib/error_handling/ErrorCode.hpp>
+
+namespace clp::ffi {
+/**
+ * Error enums for `EncodedTextAst`.
+ */
+enum class EncodedTextAstErrEnum : uint8_t {
```
gibber9809:
```suggestion
enum class EncodedTextAstErrorCodeEnum : uint8_t {
```
To be more consistent with how we name other error code enums.

components/core/src/clp/ffi/EncodedTextAstError.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include <string>
+
+#include <ystdlib/error_handling/ErrorCode.hpp>
+
+using clp::ffi::EncodedTextAstErrEnum;
+using EncodedTextAstErrorCategory = ystdlib::error_handling::ErrorCategory<EncodedTextAstErrEnum>;
+
+template <>
+auto EncodedTextAstErrorCategory::name() const noexcept -> char const* {
+    return "clp::ffi::ir_stream::IrErrorCode";
```
gibber9809:
```suggestion
    return "clp::ffi::EncodedTextAstErrorCode";
```

components/core/src/clp/ffi/EncodedTextAstError.cpp
```diff
@@ -0,0 +1,29 @@
...
+auto EncodedTextAstErrorCategory::message(EncodedTextAstErrEnum error_enum) const -> std::string {
+    switch (error_enum) {
+        case EncodedTextAstErrEnum::MissingEncodedVar:
+            return "An encoded variable is missing from the `EncodedTextAst`";
+        case EncodedTextAstErrEnum::MissingDictVar:
+            return "A dictionary variable is missing from the `EncodedTextAst`";
+        case EncodedTextAstErrEnum::MissingLogtype:
+            return "The logtype is missing from the `EncodedTextAst`";
+        case EncodedTextAstErrEnum::UnexpectedTrailingEscapeCharacter:
+            return "Unexpected escape character without escaped value at the end of the logtype";
```
gibber9809, line 25:
```suggestion
            return "An encoded variable is missing from the `EncodedTextAst`.";
        case EncodedTextAstErrEnum::MissingDictVar:
            return "A dictionary variable is missing from the `EncodedTextAst`.";
        case EncodedTextAstErrEnum::MissingLogtype:
            return "The logtype is missing from the `EncodedTextAst`.";
        case EncodedTextAstErrEnum::UnexpectedTrailingEscapeCharacter:
            return "Unexpected escape character without escaped value at the end of the logtype.";
```
We end the error messages with periods for our other error categories, so editing this for consistency.

components/core/src/clp/ffi/EncodedTextAstError.hpp
```diff
@@ -0,0 +1,24 @@
...
+    MissingDictVar = 1,
+    MissingEncodedVar,
+    MissingLogtype,
+    UnexpectedTrailingEscapeCharacter,
+};
+
+using EncodedTextAstErr = ystdlib::error_handling::ErrorCode<EncodedTextAstErrEnum>;
+}  // namespace clp::ffi
+
+YSTDLIB_ERROR_HANDLING_MARK_AS_ERROR_CODE_ENUM(clp::ffi::EncodedTextAstErrEnum);
```
gibber9809:
```suggestion
using EncodedTextAstErrorCode = ystdlib::error_handling::ErrorCode<EncodedTextAstErrorCodeEnum>;
}  // namespace clp::ffi

YSTDLIB_ERROR_HANDLING_MARK_AS_ERROR_CODE_ENUM(clp::ffi::EncodedTextAstErrorCodeEnum);
```
Similarly, for consistency with other error codes.

components/core/src/clp/ffi/test/test_EncodedTextAst.cpp
```diff
@@ -0,0 +1,193 @@
...
+    REQUIRE(encode_message(input, logtype, encoded_vars, dict_var_bounds));
+
+    StringBlob string_blob;
+    // Load dict vars
+    for (size_t i{0}; i < dict_var_bounds.size(); i += 2) {
+        auto const begin{static_cast<size_t>(dict_var_bounds[i])};
+        auto const length{static_cast<size_t>(dict_var_bounds[i + 1]) - begin};
+        auto const dict_var{input.substr(begin, length)};
+        BufferReader reader{dict_var.data(), dict_var.size()};
+        REQUIRE_FALSE(string_blob.read_from(reader, length).has_value());
```
gibber9809: Do you think it makes sense to also add an `emplace_back(std::string_view)` signature to `StringBlob`? Besides tests I think it might also end up being useful for the encode path.

components/core/src/clp/ffi/test/test_EncodedTextAst.cpp
```diff
@@ -0,0 +1,193 @@
...
+        };
+        size_t var_ix{0};
+        text = "here is a string with a small int " + var_strs[var_ix++];
+        text += " and a medium int " + var_strs[var_ix++];
+        text += " and a very large int " + var_strs[var_ix++];
+        text += " and a small float " + var_strs[var_ix++];
+        text += " and a medium float " + var_strs[var_ix++];
+        text += " and a weird float " + var_strs[var_ix++];
+        text += " and a string with numbers " + var_strs[var_ix++];
+        text += " and another string with numbers " + var_strs[var_ix++];
```
gibber9809: Since this sort of mimics the long test case from `test-EncodedVariableInterpreter.cpp` could we also add the case with a variable that contains a backslash? Figure we might as well include that edge case since we're doing all of these other cases.

components/core/src/clp/ffi/test/test_EncodedTextAst.cpp
LinZhihao-723: Sure, but I doubt we're gonna need this in the encode path: I prefer to preserve a list of string views and encode them directly, since the lifetime of the input message should outlive the encode method and we can avoid a memory copy.

gibber9809: Ok, leave it as is then.

components/core/src/clp/ffi/EncodedTextAstError.cpp
LinZhihao-723: I propose that we try to avoid adding periods unless it's a logging statement. Two benefits of doing so:

1. In a logging statement we usually do:
  ```
    LOG(fmt("{}: {}", err.category().name(), err.message()));
  ```
  While in other logging statement we do:
  ```
    LOG(fmt("{} is not valid.", addr))
  ```
  In the first statement we're assuming the error message will end with a period. But this becomes hard to manage as some error codes don't end with `.` while others do (e.g., std::errc doesn't end with `.`: https://en.cppreference.com/w/cpp/error/errc.html)

2. Not including the period in the error message allows us to format the message in the middle of a sentence if needed.

components/core/src/clp/ffi/EncodedTextAst.hpp
LinZhihao-723: I will switch back to use passing-by-value. However, do u think it makes more sense to update our code base to use rvalue ref for callbacks? This makes the callback object more flexible: the callback can be not just a lambda but also an object that holds stateful values. Using rvalue allows us to pass the object into the call and use it later.

LinZhihao-723:
> Did an initial review, mostly looks good. Besides the comments I left I'm also wondering about the `[]` usage in the decode method -- I'm personally fine with it since we're doing bounds checking ourselves before every access, but I'm wondering if we should add a comment since I think our coding standard prefers `.at()`.

Switched to use `.at`. Should do this refactoring when copying the code in the first place. My bad.

components/core/src/clp/ffi/test/test_EncodedTextAst.cpp
LinZhihao-723: Added. Also modernize the text string construction.

components/core/src/clp/ffi/EncodedTextAst.hpp
LinZhihao-723: @coderabbitai can u create an issue to track this codebase-level update?

components/core/src/clp/ffi/test/test_EncodedTextAst.cpp
LinZhihao-723: Added.

components/core/src/clp/ffi/EncodedTextAstError.cpp
gibber9809: I'm fine with that, but we should add an issue to track this change throughout the codebase since we're mostly using periods at the end of error messages at the moment.

PR #1569, opened 2025-11-07
feat(webui): Add hover messages for ANTLR parsing errors in Monaco single-line editor.

davemarco:
# Description

Add hover feature for ANTLR error. Also reworked highlight to use a different monaco API, since the marker api was not very flexible. 

Screenshot 2025-11-06 at 9 28 59 AM

# Checklist

  breaking change.

# Validation performed

Hover box appears

## Summary by CodeRabbit

* **New Features**
  * Real-time SQL validation with inline squiggle errors and hover messages for SELECT, WHERE and ORDER BY inputs
  * New validators for SELECT/WHERE/ORDER BY integrated into the UI and returned as structured validation results
  * "Run Query" editor action for quick execution
  * Improved editor hover support with safe escaping to ensure correct tooltip rendering

### DIFF
--- a/.github/workflows/clp-rust-checks.yaml
+++ b/.github/workflows/clp-rust-checks.yaml
@@ -25,13 +25,7 @@ concurrency:
 
 jobs:
   rust-checks:
-    strategy:
-      matrix:
-        os:
-          - "macos-15"
-          - "ubuntu-22.04"
-          - "ubuntu-24.04"
-    runs-on: "${{matrix.os}}"
+    runs-on: "ubuntu-22.04"
     steps:
       - uses: "actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683"
         with:
@@ -40,6 +34,9 @@ jobs:
       - name: "Install task"
         shell: "bash"
         run: "npm install -g @go-task/cli@3.44.0"
+
+      - name: "Install the latest version of uv"
+        uses: "astral-sh/setup-uv@85856786d1ce8acfbcc2f13a5f3fbd6b938f9f41"  # v7.1.2
 
       - name: "Validate lock files"
         shell: "bash"
--- a/components/clp-mcp-server/clp_mcp_server/clp_mcp_server.py
+++ b/components/clp-mcp-server/clp_mcp_server/clp_mcp_server.py
@@ -8,7 +8,7 @@ import sys
 from pathlib import Path
 
 import click
-from clp_py_utils.clp_config import CLPConfig, MCP_SERVER_COMPONENT_NAME
+from clp_py_utils.clp_config import ClpConfig, MCP_SERVER_COMPONENT_NAME
 from clp_py_utils.clp_logging import get_logger, get_logging_formatter, set_logging_level
 from clp_py_utils.core import read_yaml_config_file
 from pydantic import ValidationError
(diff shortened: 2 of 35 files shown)


components/webui/client/src/components/SqlInput/index.tsx
```diff
@@ -42,6 +62,51 @@ const SqlInput = (props: SqlEditorProps) => {
...
+
+            return;
+        }
+
+        const errors = validateFn(value);
+        const decorations: monaco.editor.IModelDeltaDecoration[] = errors.map((error) => ({
+            range: new monaco.Range(error.line, error.startColumn, error.line, error.endColumn),
+            options: {
+                className: "squiggly-error",
+                hoverMessage: {value: escapeHoverMarkdown(error.message), isTrusted: false},
```
hoophalab, line 89: how about
```suggestion
                hoverMessage: {value: renderToString(<>{error.message}</>), isTrusted: false},
```
and `import { renderToString } from 'react-dom/server';`

components/webui/client/src/sql-parser/index.ts
```diff
@@ -17,16 +18,36 @@ import {
...
     override syntaxError (
         _recognizer: Recognizer<TSymbol>,
-        _offendingSymbol: TSymbol,
+        offendingSymbol: TSymbol,
         line: number,
         column: number,
         msg: string,
     ) {
-        throw new SyntaxError(`line ${line}:${column}: ${msg}`);
+        const token = offendingSymbol as unknown as Token;
```
hoophalab, line 42: ANTLR's typing feels incomplete. Up to you: do you want to add a comment noting that [offendingSymbol is always a Token](https://github.com/antlr/antlr4/blob/cc82115a4e7f53d71d9d905caa2c2dfa4da58899/runtime/JavaScript/src/antlr4/Parser.js#L308) in ANTLR 4.13.2?

components/webui/client/src/components/SqlInput/index.tsx
hoophalab: I guess hoverMessage expects a HTML string, then we probably should give it a string rendered from an html element?
The old message doesn't escape & if there is a raw `&amp;`.

davemarco: I believe it is markdown not html? https://microsoft.github.io/monaco-editor/typedoc/interfaces/languages.Hover.html. May i can ampersand to the markdown escape function?

hoophalab: Oh yeah, I didn't realize it's actually markdown. I guess markdown special characters can still be easily triggered and we should escape them?

davemarco: i updated with more escapes

components/webui/client/src/components/SqlInput/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+    .replace(/#/g, "\\#")
+    .replace(/\//g, "\\/")
+    .replace(/\(/g, "\\(")
+    .replace(/\)/g, "\\)")
+    .replace(/\[/g, "\\[")
+    .replace(/\]/g, "\\]")
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/_/g, "\\_")
+    .replace(/`/g, "\\`");
```
hoophalab: how about
```suggestion
const escapeHoverMarkdown = (text: string) => {
    // eslint-disable-next-line no-useless-escape
    return text.replace(/([\\`*_\{\}\[\]<>\(\)#\+\-\.\!])/g, "\\$1");
};

```

PR #1586, opened 2025-11-11
feat(LocalStack): Add LocalStack support for AWS service emulation:

LinZhihao-723:
* Add scripts to start/stop LocalStack containers on the local host.
* Add script to provision S3 bucket with optional SQS queue listening to the bucket.
* Integrate LocalStack with the Rust testing task.

# Description

LocalStack is a cloud simulator that lets you test AWS services locally: https://github.com/localstack/localstack

This PR introduces three scripts to manage LocalStack containers:
- `start.py` to start a LocalStack container.
- `stop.py` to stop a running LocalStack container.
- `create-bucket.py` to create a bucket in the LocalStack, with an optional SQS queue listening to the bucket events.
- `get-free-port.py` to return an unused port on localhost.

These scripts are integrated into Rust's testing task to ensure they're compatible with the GitHub CI environment.

# Checklist

  breaking change.

# Validation performed

* [x] Ensure the scripts can pass all ruff checks with the config specified under `tools/yscope-dev-utils/exports/lint-configs/python/ruff.toml`.
* [x] Manually tested the script to ensure the following cases are covered:
  * The container name already exists.
  * The bucket/queue to create already exists.
  * The container to stop doesn't exist.

## Summary by CodeRabbit

* **Chores**
  * CI checks simplified to a single Ubuntu 22.04 runner and now install a UV tool during validation.
  * Added local cloud-emulation tooling and helper scripts to start/stop the emulator, pick a free port, and provision test resources; the test task now orchestrates the emulator lifecycle and exposes related variables.

* **Tests**
  * Tests are skipped on macOS runners due to the emulator dependency.

### DIFF
--- a/components/clp-mcp-server/clp_mcp_server/clp_mcp_server.py
+++ b/components/clp-mcp-server/clp_mcp_server/clp_mcp_server.py
@@ -8,7 +8,7 @@ import sys
 from pathlib import Path
 
 import click
-from clp_py_utils.clp_config import CLPConfig, MCP_SERVER_COMPONENT_NAME
+from clp_py_utils.clp_config import ClpConfig, MCP_SERVER_COMPONENT_NAME
 from clp_py_utils.clp_logging import get_logger, get_logging_formatter, set_logging_level
 from clp_py_utils.core import read_yaml_config_file
 from pydantic import ValidationError
@@ -72,7 +72,7 @@ def main(host: str, port: int, config_path: Path) -> int:
         exit_code = 1
 
     try:
-        clp_config = CLPConfig.model_validate(read_yaml_config_file(config_path))
+        clp_config = ClpConfig.model_validate(read_yaml_config_file(config_path))
     except ValidationError:
         logger.exception("Configuration validation failed.")
         exit_code = 1
--- a/components/clp-mcp-server/clp_mcp_server/server/server.py
+++ b/components/clp-mcp-server/clp_mcp_server/server/server.py
@@ -2,7 +2,7 @@
 
 from typing import Any
 
-from clp_py_utils.clp_config import CLPConfig
+from clp_py_utils.clp_config import ClpConfig
 from fastmcp import Context, FastMCP
 from starlette.requests import Request
 from starlette.responses import PlainTextResponse
@@ -14,7 +14,7 @@ from .session_manager import SessionManager
 from .utils import format_query_results, parse_timestamp_range, sort_by_timestamp
 
 
-def create_mcp_server(clp_config: CLPConfig) -> FastMCP:
+def create_mcp_server(clp_config: ClpConfig) -> FastMCP:
     """
     Creates and defines API tool calls for the CLP MCP server.
 
--- a/components/clp-package-utils/clp_package_utils/controller.py
+++ b/components/clp-package-utils/clp_package_utils/controller.py
@@ -13,7 +13,7 @@ from typing import Any, Optional
 from clp_py_utils.clp_config import (
     API_SERVER_COMPONENT_NAME,
     AwsAuthType,
-    CLPConfig,
+    ClpConfig,
     COMPRESSION_JOBS_TABLE_NAME,
     COMPRESSION_SCHEDULER_COMPONENT_NAME,
     COMPRESSION_WORKER_COMPONENT_NAME,
@@ -83,7 +83,7 @@ class BaseController(ABC):
     variables, directories, and configuration files for each component.
     """
 
-    def __init__(self, clp_config: CLPConfig) -> None:
+    def __init__(self, clp_config: ClpConfig) -> None:
         self._clp_config = clp_config
         self._clp_home = get_clp_home()
         self._conf_dir = self._clp_home / "etc"
@@ -450,7 +450,7 @@ class BaseController(ABC):
 
         return env_vars
 
-    def _set_up_env_for_webui(self, container_clp_config: CLPConfig) -> EnvVarsDict:
+    def _set_up_env_for_webui(self, container_clp_config: ClpConfig) -> EnvVarsDict:
         """
         Sets up environment variables and settings for the Web UI component.
 
@@ -690,7 +690,7 @@ class DockerComposeController(BaseController):
     Controller for orchestrating CLP components using Docker Compose.
     """
 
-    def __init__(self, clp_config: CLPConfig, instance_id: str) -> None:
+    def __init__(self, clp_config: ClpConfig, instance_id: str) -> None:
         self._project_name = f"clp-package-{instance_id}"
         super().__init__(clp_config)
 
@@ -857,7 +857,7 @@ class DockerComposeController(BaseController):
         return "docker-compose.yaml"
 
 
-def get_or_create_instance_id(clp_config: CLPConfig) -> str:
+def get_or_create_instance_id(clp_config: ClpConfig) -> str:
     """
     Gets or creates a unique instance ID for this CLP instance.
 
--- a/components/clp-package-utils/clp_package_utils/general.py
+++ b/components/clp-package-utils/clp_package_utils/general.py
@@ -15,7 +15,7 @@ import yaml
 from clp_py_utils.clp_config import (
     CLP_DEFAULT_CREDENTIALS_FILE_PATH,
     CLP_SHARED_CONFIG_FILENAME,
-    CLPConfig,
+    ClpConfig,
     CONTAINER_AWS_CONFIG_DIRECTORY,
     CONTAINER_CLP_HOME,
     CONTAINER_INPUT_LOGS_ROOT_DIR,
@@ -129,7 +129,7 @@ class DockerMount:
         return mount_str
 
 
-class CLPDockerMounts:
+class ClpDockerMounts:
     def __init__(self, clp_home: pathlib.Path, docker_clp_home: pathlib.Path):
         self.input_logs_dir: Optional[DockerMount] = None
         self.clp_home: Optional[DockerMount] = DockerMount(
@@ -231,8 +231,8 @@ def is_path_already_mounted(
 
 
 def generate_container_config(
-    clp_config: CLPConfig, clp_home: pathlib.Path
-) -> Tuple[CLPConfig, CLPDockerMounts]:
+    clp_config: ClpConfig, clp_home: pathlib.Path
+) -> Tuple[ClpConfig, ClpDockerMounts]:
     """
     Copies the given config and sets up mounts mapping the relevant host paths into the container
 
@@ -242,7 +242,7 @@ def generate_container_config(
     """
     container_clp_config = clp_config.model_copy(deep=True)
 
-    docker_mounts = CLPDockerMounts(clp_home, CONTAINER_CLP_HOME)
+    docker_mounts = ClpDockerMounts(clp_home, CONTAINER_CLP_HOME)
 
     if StorageType.FS == clp_config.logs_input.type:
         input_logs_dir = resolve_host_path(clp_config.logs_input.directory)
@@ -314,7 +314,7 @@ def generate_container_config(
     return container_clp_config, docker_mounts
 
 
-def generate_docker_compose_container_config(clp_config: CLPConfig) -> CLPConfig:
+def generate_docker_compose_container_config(clp_config: ClpConfig) -> ClpConfig:
     """
     Copies the given config and transforms mount paths and hosts for Docker Compose.
 
@@ -327,7 +327,7 @@ def generate_docker_compose_container_config(clp_config: CLPConfig) -> CLPConfig
     return container_clp_config
 
 
-def generate_worker_config(clp_config: CLPConfig) -> WorkerConfig:
+def generate_worker_config(clp_config: ClpConfig) -> WorkerConfig:
     worker_config = WorkerConfig()
     worker_config.package = clp_config.package.model_copy(deep=True)
     worker_config.archive_output = clp_config.archive_output.model_copy(deep=True)
@@ -344,7 +344,7 @@ def get_container_config_filename(container_name: str) -> str:
 
 
 def dump_container_config(
-    container_clp_config: CLPConfig, clp_config: CLPConfig, config_filename: str
+    container_clp_config: ClpConfig, clp_config: ClpConfig, config_filename: str
 ):
     """
     Writes the given container config to the logs directory, so that it's accessible in the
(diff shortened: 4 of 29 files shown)


.github/workflows/clp-rust-checks.yaml
```diff
@@ -50,5 +53,7 @@ jobs:
         run: "task lint:check-rust"
 
       - name: "Test"
+        # Skip tests on macOS since the test requires LocalStack
+        if: "matrix.os != 'macos-15'"
```
junhaoliao: in this workflow, we perform these actions:
- cargo lock validation
- cargo fmt / clippy
- cargo test

out of those actions, i feel only the tests really need to run on multiple platforms --- the lockfile, formatting, and linting should be platform-agnostic. if we decide to run tests only on Ubuntu, can we simplify the workflow by removing the matrix and just execute everything on a single platform (e.g., `ubuntu-24.04`)?

tools/scripts/localstack/create-bucket.py
```diff
@@ -0,0 +1,207 @@
...
+import sys
+from typing import Any
+
+import boto3
+
+_ACCESS_KEY_ID = "test"
+# To allow using hardcoded password
+# ruff: noqa: S105
+_SECRETE_ACCESS_KEY = "test"
+_REGION: str = "us-east-1"
```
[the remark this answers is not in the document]
junhaoliao: i agree with the rabbit

tools/scripts/localstack/create-bucket.py
```diff
@@ -0,0 +1,207 @@
...
+    try:
+        sqs_client.create_queue(QueueName=args.queue)
+        queue_url = sqs_client.get_queue_url(QueueName=args.queue)["QueueUrl"]
+        queue_arn = sqs_client.get_queue_attributes(
+            QueueUrl=queue_url, AttributeNames=["QueueArn"]
+        )["Attributes"]["QueueArn"]
+    except Exception as _:
+        logger.exception("Failed to create S3 bucket.")
+        return 1
+    logger.info("SQS queue '%s' created successfully with arn '%s'.", args.queue, queue_arn)
```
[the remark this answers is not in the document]
junhaoliao: i agree with the rabbit

tools/scripts/localstack/create-bucket.py
```diff
@@ -0,0 +1,207 @@
+#!/usr/bin/env python3
+# /// script
+# dependencies = [
+#   "boto3>=1.40.70",
+# ]
+# ///
+"""Script to create an S3 bucket with an optional SQS queue listen to the bucket."""
```
junhaoliao:
```suggestion
"""Script to create an S3 bucket with an optional SQS queue listening to the bucket."""
```

taskfiles/tests/main.yaml
```diff
@@ -6,8 +6,24 @@ includes:
 
 tasks:
   rust-all:
-    deps: ["toolchains:rust"]
+    vars:
+      LOG_INGESTOR_BUCKET: "clp-log-ingestor-test-bucket"
+      LOG_INGESTOR_QUEUE: "clp-log-ingestor-test-queue"
+      LOCALSTACK_CONTAINER_NAME_PREFIX: "clp-localstack-rust"
+      LOCALSTACK_CONTAINER_UUID:
+        sh: "uuidgen"
```
junhaoliao:
```suggestion
      LOCALSTACK_CONTAINER_NAME_PREFIX: "clp-localstack-rust"
      LOCALSTACK_CONTAINER_UUID:
        sh: "uuidgen"
      LOG_INGESTOR_BUCKET: "clp-log-ingestor-test-bucket"
      LOG_INGESTOR_QUEUE: "clp-log-ingestor-test-queue"
```

tools/scripts/localstack/start.py
```diff
@@ -0,0 +1,84 @@
...
+    level=logging.INFO,
+    format="%(asctime)s [%(levelname)s] %(message)s",
+    datefmt="%Y-%m-%d %H:%M:%S",
+)
+logger = logging.getLogger("localstack-start")
+
+
+def main() -> int:
+    """Main."""
+    # To silence Ruff S607
```
junhaoliao:
```suggestion
    # Silence Ruff S607: the absolute path of the Docker binary may vary depending on the installation method.
```

tools/scripts/localstack/start.py
```diff
@@ -0,0 +1,84 @@
...
+    docker_executable = "docker"
```
junhaoliao: possible to move this closer its usage?

tools/scripts/localstack/start.py
```diff
@@ -0,0 +1,84 @@
...
+
+    result = subprocess.run(
+        [docker_executable, "inspect", "-f", "{{.State.Running}}", args.name],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    if result.returncode == 0 and result.stdout.rstrip("\n") == "true":
+        logger.warning("Container '%s' already exists.", args.name)
+        return 1
```
junhaoliao, line 51: possible to extract this into a helper to avoid repetition in `stop.py`?

tools/scripts/localstack/stop.py
```diff
@@ -0,0 +1,59 @@
...
+        "stop",
+        args.name,
+    ]
+
+    result = subprocess.run(localstack_stop_cmd, capture_output=True, text=True, check=False)
+    if result.returncode != 0:
+        logger.error("Failed to stop LocalStack container:\n%s", result.stderr)
+        return result.returncode
+    logger.info("LocalStack container stopped successfully.")
+    return result.returncode
```
junhaoliao:
```suggestion
    return 0
```

tools/scripts/localstack/create-bucket.py
```diff
@@ -0,0 +1,207 @@
...
+
+# To allow using try-except-pass pattern to detect whether resources are already created.
+# ruff: noqa: TRY300, BLE001
+
+# Silence `too-many-return-statements` for `main`.
+# ruff: noqa: PLR0911
```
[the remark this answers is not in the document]
junhaoliao: can we confirm this?

tools/scripts/localstack/create-bucket.py
```diff
@@ -0,0 +1,207 @@
+#!/usr/bin/env python3
+# /// script
+# dependencies = [
+#   "boto3>=1.40.70",
```
junhaoliao: shall we pin the version with `==` to ensure reproducibility in CI / tests? if not, i think it's better to just write `"boto3"` to use the latest version

tools/scripts/localstack/create-bucket.py
```diff
@@ -0,0 +1,207 @@
...
+import json
+import logging
+import sys
+from typing import Any
+
+import boto3
+
+_ACCESS_KEY_ID = "test"
+# To allow using hardcoded password
+# ruff: noqa: S105
```
junhaoliao, line 25: does this disable S105 for the entire file?

tools/scripts/localstack/create-bucket.py
```diff
@@ -0,0 +1,207 @@
+#!/usr/bin/env python3
```
junhaoliao: if i'm not misreading [the manual](https://docs.astral.sh/uv/guides/scripts/#using-a-shebang-to-create-an-executable-file), the shebang should be written as
```
#!/usr/bin/env -S uv run --script
```

tools/scripts/localstack/create-bucket.py
```diff
@@ -0,0 +1,207 @@
...
+# ruff: noqa: S105
+_SECRETE_ACCESS_KEY = "test"
+_REGION: str = "us-east-1"
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s [%(levelname)s] %(message)s",
+    datefmt="%Y-%m-%d %H:%M:%S",
+)
+logger = logging.getLogger("create-log-ingestor-test-resources")
```
junhaoliao: why not
```suggestion
logger = logging.getLogger(__name__)
```

tools/scripts/localstack/create-bucket.py
```diff
@@ -0,0 +1,207 @@
...
+    queue_arn: str,
+) -> bool:
+    """
+    Sets SQS queue to listen to all S3 bucket events.
+
+    :param s3_client: Boto3 S3 client.
+    :param sqs_client: Boto3 SQS client.
+    :param bucket_name: The S3 bucket name.
+    :param queue_url: The SQS queue URL.
+    :param queue_arn: The SQS queue ARN.
```
junhaoliao, line 81: i feel the params descriptions can be removed if the typings are precise (non-`Any`)

tools/scripts/localstack/stop.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3
+# /// script
+# dependencies = [
+# ]
+# ///
```
junhaoliao, line 4: do we need this block if no external dependency is needed?

tools/scripts/localstack/start.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python3
+# /// script
+# dependencies = [
+# ]
+# ///
```
junhaoliao, line 4: ditto - is this block needed if there's no needed external dependency?

tools/scripts/localstack/create-bucket.py
junhaoliao: actually, i think the version range specification is fine to keep so long we generate a lock file: https://docs.astral.sh/uv/guides/scripts/#locking-dependencies

tools/scripts/localstack/create-bucket.py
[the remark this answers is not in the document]
LinZhihao-723: It is enabled when running with:
```bash
uv run ruff check --fix --config tools/yscope-dev-utils/exports/lint-configs/python/ruff.toml tools/scripts/localstack
```
@junhaoliao fyi

tools/scripts/localstack/start.py
LinZhihao-723: Perfer not to since now the script is self-contained.

tools/scripts/localstack/create-bucket.py
LinZhihao-723: Yes.

tools/scripts/localstack/create-bucket.py
LinZhihao-723: Prefer to use `==` then.
Don't wanna ship these scripts with a lock file before we have a clear plan for script management.

tools/scripts/localstack/stop.py
LinZhihao-723: Yes. Reference to the NOTE in this section: https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies

tools/scripts/localstack/create-bucket.py
LinZhihao-723: This is copied from `scripts/deps-download/download-dep.py`. After testing the shebang is not needed. We also don't need the shebang shown in this manual as we don't want to call this script without uv.

tools/scripts/localstack/create-bucket.py
LinZhihao-723: To get the type annotation correct we'd need the extra dependency... not sure if we want to add that as I hope this script can be more lightweight.

.github/workflows/clp-rust-checks.yaml
LinZhihao-723: Sure. How about only run on Ubuntu22.04 as it's where our package is running? I think we don't have to update this until we shift some of core implementation into Rust (which may not be in a near future, lol)

LinZhihao-723:
> for the title, how about simply:
> 
> ```
> feat(tools): Add LocalStack S3/SQS setup for Rust tests.
> ```

I prefer to focus on the fact that these scripts are for LocalStack orchestration. If you're ok with this idea I can try to come up with a more refined title.

.github/workflows/clp-rust-checks.yaml
junhaoliao: > only run on Ubuntu22.04 as it's where our package is running

sure

taskfiles/tests/main.yaml
```diff
@@ -6,8 +6,27 @@ includes:
 
 tasks:
   rust-all:
-    deps: ["toolchains:rust"]
+    vars:
+      LOG_INGESTOR_BUCKET: "clp-log-ingestor-test-bucket"
+      LOG_INGESTOR_QUEUE: "clp-log-ingestor-test-queue"
+      LOCALSTACK_CONTAINER_NAME:
+        sh: "uuidgen | tr '[:upper:]' '[:lower:]' | sed 's/^/clp-localstack-rust-/'"
```
junhaoliao, line 14:
```suggestion
        # Normalize UUID casing: macOS generates uppercase while Linux generates lowercase.
        sh: "uuidgen | tr '[:upper:]' '[:lower:]' | sed 's/^/clp-localstack-rust-/'"
```

tools/scripts/localstack/stop.py
```diff
@@ -0,0 +1,58 @@
+# /// script
+# dependencies = [
+# ]
```
junhaoliao:
```suggestion
# dependencies = []
```

tools/scripts/localstack/start.py
```diff
@@ -0,0 +1,84 @@
+# /// script
+# dependencies = [
+# ]
```
junhaoliao:
```suggestion
# dependencies = []
```

tools/scripts/localstack/get-free-port.py
```diff
@@ -0,0 +1,30 @@
+# /// script
+# dependencies = [
+# ]
```
junhaoliao:
```suggestion
# dependencies = []
```

junhaoliao:
> If you're ok with this idea I can try to come up with a more refined title.

sure

tools/scripts/localstack/create-bucket.py
junhaoliao: hmm... if we do specify a shebang `#!/usr/bin/env -S uv run --script` as instructed by the manual, the script can be run in one of these ways:
```
# if we have `chmod u+x` the script
./create-bucket.py

# or
uv run ./create-bucket.py
```
either way, `uv` has to be installed, or failures would be reported.

in the meanwhile, with or without the "uv" shebang, we can't really stop new developers from doing this
```
python3 ./create-bucket.py
```

i think the "uv" shebang actually self documents the script is a uv-managed script and therefore is worth keeping

tools/scripts/localstack/get-free-port.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def get_free_port() -> int:
+    """
+    Finds an unused port on localhost.
+
+    :return: An unused port number.
+    """
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind(("", 0))
```
junhaoliao, line 17: Note-to-self: this is equivalent to `s.bind(("0.0.0.0", 0))`

tools/scripts/localstack/create-bucket.py
LinZhihao-723: Done. Call the script directly without `uv` for simplicity.

PR #1590, opened 2025-11-12
feat(clp-s): Add support for outputting search results to a file.

hoophalab:
# Description

This PR adds support for outputting search results to a file:
`clp-s s <archives_dir> "*" file --path test.jsonl`

This is equivalent to `clp-s s <archives_dir> "*" stdout > test.jsonl`. We are adding this functionality due to the slight performance gains when calling it in Python.

Changes:
1. Add `FileOutputHandler`.
2. Add the corresponding command-line arguments.

# Checklist

  breaking change.

# Validation performed

Run several kql filters `clp-s s <archives_dir> "*" file --path test.jsonl` on PostgreSQL logs. The output file matches the stdout.

## Summary by CodeRabbit

* **New Features**
  * Added a "file" output option so search results can be written to a specified file.
  * File output can include optional timestamps and metadata in each written record.
  * Search now supports selecting file as an output destination.

* **Bug Fixes**
  * Command-line parsing validates a non-empty file path and reports an error if invalid.

* **Documentation**
  * Help/usage updated with examples and guidance for the file output option.

### DIFF
--- a/components/core/src/clp/make_dictionaries_readable/CMakeLists.txt
+++ b/components/core/src/clp/make_dictionaries_readable/CMakeLists.txt
@@ -2,8 +2,24 @@ set(
         MAKE_DICTIONARIES_READABLE_SOURCES
         ../dictionary_utils.cpp
         ../dictionary_utils.hpp
+        ../Defs.h
         ../DictionaryEntry.hpp
         ../DictionaryReader.hpp
+        ../ErrorCode.hpp
+        ../EncodedVariableInterpreter.cpp
+        ../EncodedVariableInterpreter.hpp
+        ../ffi/encoding_methods.cpp
+        ../ffi/encoding_methods.hpp
+        ../ffi/encoding_methods.inc
+        ../ffi/ir_stream/byteswap.hpp
+        ../ffi/ir_stream/decoding_methods.cpp
+        ../ffi/ir_stream/decoding_methods.hpp
+        ../ffi/ir_stream/decoding_methods.inc
+        ../ffi/ir_stream/protocol_constants.hpp
+        ../ffi/ir_stream/utils.cpp
+        ../ffi/ir_stream/utils.hpp
+        ../ffi/SchemaTree.cpp
+        ../ffi/SchemaTree.hpp
         ../FileDescriptor.cpp
         ../FileDescriptor.hpp
         ../FileReader.cpp
@@ -12,11 +28,17 @@ set(
         ../FileWriter.hpp
         ../ir/parsing.cpp
         ../ir/parsing.hpp
+        ../ir/parsing.inc
+        ../ir/types.hpp
         ../LogTypeDictionaryEntry.cpp
         ../LogTypeDictionaryEntry.hpp
+        ../LogTypeDictionaryEntryReq.hpp
         ../LogTypeDictionaryReader.hpp
+        ../Query.cpp
+        ../Query.hpp
         ../ParsedMessage.cpp
         ../ParsedMessage.hpp
+        ../Platform.hpp
         ../ReaderInterface.cpp
         ../ReaderInterface.hpp
         ../ReadOnlyMemoryMappedFile.cpp
@@ -27,11 +49,17 @@ set(
         ../streaming_compression/passthrough/Decompressor.hpp
         ../streaming_compression/zstd/Decompressor.cpp
         ../streaming_compression/zstd/Decompressor.hpp
+        ../type_utils.hpp
+        ../TraceableException.hpp
+        ../TimestampPattern.cpp
+        ../TimestampPattern.hpp
         ../Utils.cpp
         ../Utils.hpp
         ../VariableDictionaryEntry.cpp
         ../VariableDictionaryEntry.hpp
         ../VariableDictionaryReader.hpp
+        ../VariableDictionaryReaderReq.hpp
+        ../VariableDictionaryWriterReq.hpp
         ../WriterInterface.cpp
         ../WriterInterface.hpp
         CommandLineArguments.cpp
@@ -44,10 +72,12 @@ if(CLP_BUILD_EXECUTABLES)
         target_compile_features(make-dictionaries-readable PRIVATE cxx_std_20)
         target_link_libraries(make-dictionaries-readable
                 PRIVATE
-                Boost::filesystem Boost::program_options
-                log_surgeon::log_surgeon
-                spdlog::spdlog
                 clp::string_utils
+                Boost::filesystem Boost::program_options
+                date::date
+                log_surgeon::log_surgeon
+                nlohmann_json::nlohmann_json
+                spdlog::spdlog
                 ystdlib::containers
                 ${zstd_TARGET}
         )


components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -635,6 +643,8 @@ CommandLineArguments::parse_arguments(int argc, char const** argv) {
                           << " - Output to the results cache" << std::endl;
                 std::cerr << "  " << static_cast<char const*>(cReducerOutputHandlerName)
                           << " - Output to the reducer" << std::endl;
+                std::cerr << "  " << static_cast<char const*>(cFileOutputHandlerName)
+                          << " - Output to a file" << std::endl;
```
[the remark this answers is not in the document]
junhaoliao: i agree with the rabbit

components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -623,6 +630,7 @@ CommandLineArguments::parse_arguments(int argc, char const** argv) {
             constexpr char cReducerOutputHandlerName[] = "reducer";
             constexpr char cResultsCacheOutputHandlerName[] = "results-cache";
             constexpr char cStdoutCacheOutputHandlerName[] = "stdout";
+            constexpr char cFileOutputHandlerName[] = "file";
```
junhaoliao: alphabetize

components/core/src/clp_s/CommandLineArguments.hpp
```diff
@@ -34,6 +34,7 @@ class CommandLineArguments {
         Reducer,
         ResultsCache,
         Stdout,
+        File,
```
junhaoliao: alphabetize

junhaoliao: which might be an internally breaking change if we ever serialize / deserialize the value

components/core/src/clp_s/OutputHandlerImpl.hpp
```diff
@@ -170,6 +171,38 @@ class ResultsCacheOutputHandler : public ::clp_s::search::OutputHandler {
...
+    explicit FileOutputHandler(std::string const& path, bool should_output_metadata = false);
+
+    // Methods inherited from OutputHandler
+    void write(
+            std::string_view message,
+            epochtime_t timestamp,
+            std::string_view archive_id,
+            int64_t log_event_idx
+    ) override {
+        file << archive_id << ": " << log_event_idx << ": " << timestamp << " " << message;
```
junhaoliao: shall we append newlines as well? so that the output file is in standard JSONL format

components/core/src/clp_s/OutputHandlerImpl.hpp
```diff
@@ -170,6 +171,38 @@ class ResultsCacheOutputHandler : public ::clp_s::search::OutputHandler {
...
+    }
+
+    void write(std::string_view message) override { file << message; }
+
+private:
+    std::ofstream file;
```
junhaoliao:
```suggestion
    std::ofstream m_file;
```

components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -623,6 +630,7 @@ CommandLineArguments::parse_arguments(int argc, char const** argv) {
...
 
             if (parsed_command_line_options.count("help")) {
```
junhaoliao, line 635: maybe not in this PR - i feel it's worth documenting different output handler options in docs/src/user-docs/core-clp-s.md

components/core/src/clp_s/CommandLineArguments.hpp
hoophalab: I did a quick "find usage" and didn't find serialization of this enum.

components/core/src/clp_s/CommandLineArguments.cpp
hoophalab: @coderabbitai can you create an issue?

components/core/src/clp_s/OutputHandlerImpl.hpp
hoophalab: I guess `message` always contains a newline at the end. This is the format of standardoutputhandler.

junhaoliao: right. i overlooked and mistakenly thought we were outputting JSON. please disregard this comment

components/core/src/clp_s/OutputHandlerImpl.hpp
```diff
@@ -170,6 +171,38 @@ class ResultsCacheOutputHandler : public ::clp_s::search::OutputHandler {
...
+            std::string_view archive_id,
+            int64_t log_event_idx
+    ) override {
+        m_file << archive_id << ": " << log_event_idx << ": " << timestamp << " " << message;
+    }
+
+    void write(std::string_view message) override { m_file << message; }
+
+private:
+    std::ofstream m_file;
```
gibber9809: To be more consistent with the rest of the codebase we should be using a `FileWriter` or `BufferedFileWriter` here.

components/core/src/clp_s/OutputHandlerImpl.hpp
```diff
@@ -170,6 +171,38 @@ class ResultsCacheOutputHandler : public ::clp_s::search::OutputHandler {
...
+    void write(
+            std::string_view message,
+            epochtime_t timestamp,
+            std::string_view archive_id,
+            int64_t log_event_idx
+    ) override {
+        m_file << archive_id << ": " << log_event_idx << ": " << timestamp << " " << message;
+    }
+
+    void write(std::string_view message) override { m_file << message; }
```
gibber9809: Do we want to output a binary format here so that downstream systems don't need to parse a bunch of text data?

If not, I don't see much point in adding a FileOutputHandler (beyond making things a bit easier when we move more things out of python code, and even then, since this PR currently doesn't include an S3OutputHandler/S3 upload path, we aren't addressing the harder part of that switchover) -- we might as well just redirect output from `StandardOutputHandler`.

gibber9809 (changes requested):

hoophalab:
@gibber9809 I've modified the file output handler to output a binary format. I feel a bit weird outputting binary to stdout. `curl` even has a warning that prevents you from writing binary to stdout pipe unless you do that specifically

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -23,6 +23,21 @@ using std::string;
 using std::string_view;
 
 namespace clp_s {
+void FileOutputHandler::write(
+        string_view message,
+        epochtime_t timestamp,
+        string_view archive_id,
+        int64_t log_event_idx
+) {
+    static constexpr string_view cOrigFilePathPlaceholder{""};
```
gibber9809, line 32: We could actually start sending the file path here since the archive does store the mapping between `log_event_idx` and original file name -- we should probably create an issue to revisit this after we've updated the package metadata design, since after that change it should also be possible to retrieve the filename based on `archive_id` and `log_event_idx`, so it might also make sense to not send it here.

components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -645,6 +655,13 @@ CommandLineArguments::parse_arguments(int argc, char const** argv) {
                           << std::endl;
                 std::cerr << std::endl;
 
+                std::cerr << "  # Search archives in archives-dir for logs matching a KQL query"
+                             R"( "level: INFO" and output to a file)"
+                          << std::endl;
+                std::cerr << "  " << m_program_name << R"( s archives-dir "level: INFO")"
+                          << " " << cFileOutputHandlerName << " --path test.jsonl" << std::endl;
```
gibber9809:
```suggestion
                          << " " << cFileOutputHandlerName << " --path test.out" << std::endl;
```
Nit: since the output is binary, something like `.out` is probably more appropriate than `.jsonl`.

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -23,6 +23,21 @@ using std::string;
...
+    msgpack::type::tuple<epochtime_t, string, string, string, int64_t> const
+            src(timestamp, message, cOrigFilePathPlaceholder, archive_id, log_event_idx);
+    msgpack::sbuffer m;
+    msgpack::pack(m, src);
+
+    m_file_writer.write(m.data(), m.size());
```
gibber9809:
```suggestion
    msgpack::pack(m_file_writer, src);
```
I think this should work given that `msgpack::pack` is supposed to be templated to accept any type with a `write(char*,size_t)` function? Worth checking if it works/if there is any performance difference.

gibber9809 (changes requested):
A few minor comments, but otherwise looks good!

components/core/src/clp_s/OutputHandlerImpl.cpp
hoophalab: Sure, but I checked the code and couldn't find the metadata for the original file name in the archive reader. Do we already deserialize this metadata in the reader?

hoophalab: @coderabbitai can you create an issue on
> revisit this after we've updated the package metadata design, since after that change it should also be possible to retrieve the filename based on archive_id and log_event_idx

junhaoliao (approved):

PR #1591, opened 2025-11-12
feat(integration-tests): Add core binary tests for the `log-converter` and `clp-s` pipeline.

Bill-hbrhbr:
# Description

This PR adds integration tests for #1460 by testing examples mentioned in #1308.

# Checklist

  breaking change.

# Validation performed

* [x] `uv run pytest -k test_log_converter_clp_s_identity_transform` passes.

## Summary by CodeRabbit

* **Tests**
  * Added an identity-transformation integration test for the log-conversion pipeline.
  * Added multiple unstructured log test datasets (simple, escaped-sequence, Hive 24hr, OpenStack 24hr, Hadoop multiline).

* **Chores**
  * Added test infrastructure support for the log-converter binary and a canonical-output filename constant.
  * Improved test cleanup to remove prior outputs before re-running.

---
  - https://app.asana.com/0/0/1212888478575770


Bill-hbrhbr:
@coderabbitai
open an issue for
```
        assert is_json_file_structurally_equal(expected_out, actual_out), (
            f"Mismatch between {expected_out}(expected) and {actual_out}(actual)."
        )
```
where the pytest fail msg may be truncated due to path being too long. I don't know what's a good solution.

integration-tests/tests/test_identity_transformation.py
```diff
@@ -139,6 +144,58 @@ def test_clp_s_identity_transform(
...
+def test_log_converter_clp_s_identity_transform(
+    core_config: CoreConfig,
+    integration_test_config: IntegrationTestConfig,
+) -> None:
+    """
+    Validate the end-to-end functionality of the `log-converter` and `clp-s` pipeline.
+
+    This test ensures that:
+    1. `log-converter` correctly transforms unstructured logs into key-value IR format.
+    2. The kv-IR output can be compressed and decompressed by `clp-s` without data loss.
```
quinntaylormitchell, line 157:
```suggestion
    Validates the end-to-end functionality of the `log-converter` binary by compressing and
    decompressing the KV-IR it produces upon ingesting an unstructured log file.
```
See high-level comment.

integration-tests/tests/test_identity_transformation.py
```diff
@@ -139,6 +144,58 @@ def test_clp_s_identity_transform(
...
+
+    :param core_config:
+    :param integration_test_config:
+    """
+    log_converter_out_dir = integration_test_config.test_root_dir / "log-converter-outputs"
+    log_converter_out_dir.mkdir(parents=True, exist_ok=True)
+    log_converter_bin_path_str = str(core_config.log_converter_binary_path)
+
+    unstructured_logs_dir = Path(__file__).resolve().parent / "data" / "unstructured-logs"
+    for test_case_dir in unstructured_logs_dir.iterdir():
```
quinntaylormitchell, line 167: See my high-level comment about the dependence on directory structure to enumerate test cases.

integration-tests/tests/test_identity_transformation.py
```diff
@@ -139,6 +144,58 @@ def test_clp_s_identity_transform(
...
+    2. The kv-IR output can be compressed and decompressed by `clp-s` without data loss.
+
+    :param core_config:
+    :param integration_test_config:
+    """
+    log_converter_out_dir = integration_test_config.test_root_dir / "log-converter-outputs"
+    log_converter_out_dir.mkdir(parents=True, exist_ok=True)
+    log_converter_bin_path_str = str(core_config.log_converter_binary_path)
+
+    unstructured_logs_dir = Path(__file__).resolve().parent / "data" / "unstructured-logs"
```
quinntaylormitchell, line 166:
```suggestion
    unstructured_logs_dir = (
        Path(get_env_var("INTEGRATION_TESTS_DIR")).expanduser().resolve()
        / "data"
        / "unstructured-logs"
    )
```
How would you feel about having this be dependent on a new env variable in `.pytest.ini`? It clearly makes the code here longer, but I'm worried that having file-location-dependent path buried here in this code will be more difficult to notice/remember to change, if we ever need to.

integration-tests/tests/test_identity_transformation.py
```diff
@@ -139,6 +144,58 @@ def test_clp_s_identity_transform(
...
+    for test_case_dir in unstructured_logs_dir.iterdir():
+        if not test_case_dir.is_dir():
+            continue
+
+        test_name = test_case_dir.name
+        kv_ir_out = log_converter_out_dir / test_name
+        unlink(kv_ir_out)
```
quinntaylormitchell, line 173: It feels odd to have this `unlink` here; I realize it isn't wrong per se, but is there a reason you've put it here at the beginning of the test case rather than at the end with the `test_paths.clear_test_outputs()` line?

integration-tests/tests/test_identity_transformation.py
```diff
@@ -139,6 +144,58 @@ def test_clp_s_identity_transform(
...
+
+        test_paths = CompressionTestConfig(
+            test_name=f"log-converter-clp-s-{test_name}",
+            logs_source_dir=kv_ir_out,
+            integration_test_config=integration_test_config,
+        )
+        _clp_s_compress_and_decompress(core_config, test_paths)
+
+        expected_out = test_case_dir / "converted.json"
+        actual_out = test_paths.decompression_dir / CLP_S_CANONICAL_OUTPUT_FILENAME
```
quinntaylormitchell, line 191: I have a bit of trouble with the value of `CLP_S_CANONICAL_OUTPUT_FILENAME` being `"original"`, because it seems to me that the output of a decompression task is, by definition, not the original file (even though it should be structurally identical to the original). What do you think?

quinntaylormitchell (changes requested):
Looking pretty good; a few comments in the code, and some high-level comments as well:

1. Fundamentally, I think that `test_log_converter_clp_s_identity_transform` is testing the functionality of the `log-converter` binary, but in some ways it seems like it's testing the `clp-s` binary's ability to compress/decompress KV-IR as well. I think that if we're going to test the veracity of KV-IR compression/decompression, that should happen in a separate test, and we should make it clear that `test_log_converter_clp_s_identity_transform` is testing the veracity of the functionality of the `log-converter` binary _by way of_ compressing and decompressing the KV-IR that `log-converter` produces. To that end, I've suggested a new docstring; have a look.
2. The concept of having the test cases enumerated by the directory structure in `integration-tests/tests/data` doesn't really sit right with me, although I do see how it simplifies the code. Feel free to disregard, I just wanted to bring it up.
3. I think that overall, the tests should be logged better. The logging doesn't need to be verbose at all, but it would be good to have some little indications telling the dev what the test is doing while it's doing it.

integration-tests/tests/test_identity_transformation.py
Bill-hbrhbr: because there may be leftover files from a previous test that we'd like to remove

quinntaylormitchell: sure, but if the `kv_ir_out` directory is only ever used during this test, wouldn't it make sense to just unlink it when we're done, rather than unlinking it only when a dev runs the test again? kind of like a "clean up everything when you're done" policy?

PR #1595, opened 2025-11-12
build(rust): Install Rust toolchain under `build/` to avoid modifying the default per-user installation; Use `nextest` as the default test runner.

LinZhihao-723:
# Description

This PR switches to use a self-managed Rust toolchain installed under `build/`, instead of directly using/updating the default installation under the user's home directory.

The installation script has the following changes:

* We set `RUSTUP_HOME` and `CARGO_HOME` before calling the installation command to control where to install the toolchain.
* We use a self-made env file to set `RUSTUP_HOME` and `CARGO_HOME`, as well as sourcing `CARGO_HOME/env` to ensure the PATH env var is properly set up. This env file should be sourced whenever a task needs to call Rust toolchains. Ref:
  * https://rust-lang.github.io/rustup/installation/index.html#choosing-where-to-install
  * https://users.rust-lang.org/t/install-rust-toolchain-to-a-project-specific-location/136209/3
* We set up a checksum file to only rerun the toolchain installation if:
  * The task file is changed.
  * The env file is changed.
  * The downloaded binaries or toolchains are changed.
  
  NOTE: We don't include the entire CARGO_HOME or RUSTUP_HOME because these directories serve as an internal cache for Rust toolchains.
  
This PR also installs `nextest` as the default test runner.
  
As a result, this PR achieves:

* Managing Rust toolchain under the CLP project.
* Avoid re-installing/updating the Rust toolchain in every task call.

# Checklist

  breaking change.

# Validation performed

* [x] Ensure the installation task only runs once across multiple task executions.
* [x] Ensure the toolchain works in a fresh CLP env container where no default toolchain has been installed.

## Summary by CodeRabbit

* **Chores**
  * Centralized, templated Rust toolchain environment: tasks now source a project-specific env file and include toolchain management scaffolding for reproducible installs and environment setup.
  * Formatting and linting tasks now respect the configured toolchain environment before running checks.
* **Tests**
  * Test execution switched to a modern test runner (cargo-nextest) for improved performance and reporting.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
 * @y-scope/maintainers
-/components/core/src/clp_s/ @gibber9809 @wraymo @y-scope/maintainers
+/components/core/src/clp_s/ @gibber9809 @y-scope/maintainers


taskfiles/toolchains.yaml
```diff
@@ -1,13 +1,76 @@
...
+        export RUSTUP_HOME={{.G_RUSTUP_HOME}}
+        . {{.G_CARGO_HOME}}/env
+        EOF
+        chmod +x {{.G_RUST_TOOLCHAIN_ENV_FILE}}
+
+      - |-
+        . {{.G_RUST_TOOLCHAIN_ENV_FILE}}
         rustup toolchain install nightly --allow-downgrade --component clippy,rustfmt
+        rustup toolchain install stable
+        rustup default stable
```
junhaoliao, line 69: i haven't tried - if we install stable first, would it be the default toolchain?

```suggestion
        rustup toolchain install stable
        rustup toolchain install nightly --allow-downgrade --component clippy,rustfmt
```

junhaoliao (changes requested):

taskfiles/toolchains.yaml
LinZhihao-723: - We add `--default-toolchain none` in the rustup init command.
- I prefer to do this explicitly so we're for sure the default toolchain is in a stable version.

taskfiles/toolchains.yaml
```diff
@@ -1,13 +1,79 @@
...
+  G_RUST_TOOLCHAIN_ROOT: "{{.G_BUILD_DIR}}/toolchains/rust"
+
+  # Rust toolchain installation paths
+  G_CARGO_HOME: "{{.G_RUST_TOOLCHAIN_ROOT}}/.cargo"
+  G_RUSTUP_HOME: "{{.G_RUST_TOOLCHAIN_ROOT}}/.rustup"
+
+  # Rust toolchain checksum file
+  G_RUST_TOOLCHAIN_CHECKSUM_FILE: "{{.G_BUILD_DIR}}/toolchains-rust.md5"
+
+  G_RUST_TOOLCHAIN_ENV_FILE: "{{.G_RUST_TOOLCHAIN_ROOT}}/env"
```
junhaoliao: how about grouping this with `G_RUST_TOOLCHAIN_ROOT`

taskfiles/toolchains.yaml
```diff
@@ -1,13 +1,79 @@
 version: "3"
 
+includes:
+  utils: "../tools/yscope-dev-utils/exports/taskfiles/utils/utils.yaml"
+
+vars:
+  G_RUST_TOOLCHAIN_ROOT: "{{.G_BUILD_DIR}}/toolchains/rust"
```
junhaoliao: i don't have a strong preference but all our directory-related global variables end with `_DIR`. shall we rename it as `G_RUST_TOOLCHAIN_ROOT_DIR` for consistency?

junhaoliao: ditto for `G_CARGO_HOME` and `G_RUSTUP_HOME`

taskfiles/toolchains.yaml
```diff
@@ -1,13 +1,79 @@
...
+      CHECKSUM_FILE: "{{.G_RUST_TOOLCHAIN_CHECKSUM_FILE}}"
+      # `RUSTUP_HOME` and `CARGO_HOME` contains cache files and temporary files. Instead of
+      # including the entire directory, we only include the necessary files to verify whether the
+      # toolchain installation is up to date.
+      OUTPUT_PATTERNS:
+        - "{{.G_CARGO_HOME}}/bin"
+        - "{{.G_CARGO_HOME}}/env"
+        - "{{.G_RUSTUP_HOME}}/settings.toml"
+        - "{{.G_RUSTUP_HOME}}/toolchains"
+        - "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
```
junhaoliao:
```suggestion
        - "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
        - "{{.G_RUSTUP_HOME}}/settings.toml"
        - "{{.G_RUSTUP_HOME}}/toolchains"
```

taskfiles/toolchains.yaml
```diff
@@ -1,13 +1,79 @@
...
           | sh -s -- --default-toolchain none -y
-        . "$HOME/.cargo/env"
+
+      # Create an environment file to configure the custom Rust toolchain location.
+      # Source this file before using cargo/rustup commands to ensure they use the toolchain
+      # installed in `{{.G_RUST_TOOLCHAIN_ROOT}}`.
+      # For details, check:
+      # https://rust-lang.github.io/rustup/installation/index.html#choosing-where-to-install
+      - |-
+        cat > {{.G_RUST_TOOLCHAIN_ENV_FILE}} << 'EOF'
```
junhaoliao: ref: https://tldp.org/LDP/abs/html/here-docs.html

1. i believe we shouldn't need to quote the `'EOF'`? by the point this executes, the go template variables should have been expanded. (not to mention go templates' variables use a different syntax than env var, ) no env varaiable would be expanded even with the EOF being unquoted.
2. put the redirection immediately after the `<<EOF`.
3. remove the space after the `<<` so no extra token is counted. though technically with the space or not, the effect is the same. anyways let's follow the convention in the docs
4. quote file names
```suggestion
        cat <<EOF > "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
```

taskfiles/toolchains.yaml
```diff
@@ -1,13 +1,79 @@
...
+        #!/bin/sh
+        export CARGO_HOME={{.G_CARGO_HOME}}
+        export RUSTUP_HOME={{.G_RUSTUP_HOME}}
+        . {{.G_CARGO_HOME}}/env
+        EOF
+        chmod +x {{.G_RUST_TOOLCHAIN_ENV_FILE}}
```
junhaoliao: quote paths
```suggestion
        export CARGO_HOME="{{.G_CARGO_HOME}}"
        export RUSTUP_HOME="{{.G_RUSTUP_HOME}}"
        . "{{.G_CARGO_HOME}}/env"
        EOF
        chmod +x "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
```

taskfile.yaml
```diff
@@ -203,7 +203,7 @@ tasks:
     deps: ["toolchains:rust"]
     dir: "{{.ROOT_DIR}}"
     cmd: |-
-      . "$HOME/.cargo/env"
+      . {{.G_RUST_TOOLCHAIN_ENV_FILE}}
```
junhaoliao:
```suggestion
      . "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
```

taskfiles/deps/lock.yaml
```diff
@@ -47,7 +47,7 @@ tasks:
     dir: "{{.ROOT_DIR}}"
     deps: ["toolchains:rust"]
     cmd: |-
-      . "$HOME/.cargo/env"
+      . {{.G_RUST_TOOLCHAIN_ENV_FILE}}
```
junhaoliao:
```suggestion
      . "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
```

taskfiles/deps/lock.yaml
```diff
@@ -60,7 +60,7 @@ tasks:
...
```
junhaoliao:
```suggestion
      . "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
```

taskfiles/lint.yaml
```diff
@@ -839,7 +839,7 @@ tasks:
     dir: "{{.ROOT_DIR}}"
     deps: ["rust-lint-configs", "toolchains:rust"]
     cmd: |-
-      . "$HOME/.cargo/env"
+      . {{.G_RUST_TOOLCHAIN_ENV_FILE}}
```
junhaoliao:
```suggestion
      . "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
```

taskfiles/lint.yaml
```diff
@@ -852,7 +852,7 @@ tasks:
     dir: "{{.ROOT_DIR}}"
     deps: ["toolchains:rust"]
     cmd: |-
-      . "$HOME/.cargo/env"
+      . {{.G_RUST_TOOLCHAIN_ENV_FILE}}
```
junhaoliao:
```suggestion
      . "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
```

taskfiles/tests/main.yaml
```diff
@@ -29,5 +29,6 @@ tasks:
         --bucket "{{.LOG_INGESTOR_BUCKET}}" \
         --queue "{{.LOG_INGESTOR_QUEUE}}" \
         --port "{{.LOCALSTACK_PORT}}"
-      - ". $HOME/.cargo/env"
-      - "cargo test --all --all-features --release"
+      - |-
+        . {{.G_RUST_TOOLCHAIN_ENV_FILE}}
```
junhaoliao:
```suggestion
      . "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
```

taskfiles/toolchains.yaml
```diff
@@ -1,13 +1,79 @@
...
+
+      - |-
+        . {{.G_RUST_TOOLCHAIN_ENV_FILE}}
```
junhaoliao:
```suggestion
      . "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
```

taskfiles/toolchains.yaml
```diff
@@ -1,13 +1,79 @@
...
         rustup toolchain install nightly --allow-downgrade --component clippy,rustfmt
+        rustup toolchain install stable
+        rustup default stable
+        cargo install cargo-nextest
+        rustup set auto-self-update disable
```
junhaoliao, line 71: 1. for future developers, can we add a short comment to explain why we want to disable auto self update?
2. can we move this right before the nightly installation line, just so there's no chance rustup would auto self update as we install nightly / stable?

taskfiles/toolchains.yaml
```diff
@@ -1,13 +1,79 @@
...
+    sources:
+      - "{{.ROOT_DIR}}/taskfile.yaml"
+      - "{{.TASKFILE}}"
+    generates: ["{{.CHECKSUM_FILE}}"]
+    deps:
+      - task: "utils:checksum:validate"
+        vars:
+          CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
+          INCLUDE_PATTERNS:
+            ref: ".OUTPUT_PATTERNS"
```
junhaoliao: can we use YAML anchors instead of creating a task variable? e.g., https://github.com/y-scope/clp/blob/dfb4526000bf9d556ac3d1af360a45bf590cbbfc/taskfiles/lint.yaml#L58C15-L70

taskfiles/toolchains.yaml
```diff
@@ -1,13 +1,79 @@
...
+
+  # Rust toolchain checksum file
+  G_RUST_TOOLCHAIN_CHECKSUM_FILE: "{{.G_BUILD_DIR}}/toolchains-rust.md5"
+
+  G_RUST_TOOLCHAIN_ENV_FILE: "{{.G_RUST_TOOLCHAIN_ROOT}}/env"
+
 tasks:
   rust:
+    # We use a label to uniquely identify this task and its checksum under `.task/checksums/`.
+    label: "toolchains-rust"
```
junhaoliao, line 20: i saw the commit message in 0a64ab48606176d3827afa2fc74015d6909b969b but can't wrap my head around why there could be a bug without adding a label. could you explain what the bug is / why this mitigates the bug?

junhaoliao (changes requested):
for the title, how about
```
build(rust): Install Rust toolchain under `build/` instead of `$HOME`; Use `nextest` as the default test runner.
```

taskfiles/toolchains.yaml
LinZhihao-723: Strongly prefer `G_CARGO_HOME` and `G_RUSTUP_HOME`, as `CARGO_HOME` and `RUSTUP_HOME` are the required env variables for Rust installation and toolchains to work properly.

taskfile.yaml
LinZhihao-723: Sorry for this. I thought it's not a parameter so didn't quote it. But I just realized we also quote the env file in Python venv. Will fix them in a batch commit.

taskfiles/toolchains.yaml
LinZhihao-723: 1. It's to make the checksum file stable.
2. Perfer to lock rustup as the last command.

taskfiles/toolchains.yaml
LinZhihao-723: Without this label, task will generate checksum files (under `.task/checksums`) based on the full task path. Consider this toolchain task is called from both `task lint:check-rust` and `task tests:rust-all`. If no tag is specified, there will be two checksums files, one for `lint:toolchains:rust` one for `tests:toolchains:rust`. This means the toolchain installation will be triggered even though this task has already been called and our self-managed checksum never changes. By adding a label, the task will be uniquely identified by its label as the task-generated checksum will be named using the label. This ensures the toolchain task runs only once across different namespaces.

taskfiles/toolchains.yaml
LinZhihao-723: sg. Pushed in the latest commit.

taskfiles/toolchains.yaml
LinZhihao-723: `G_RUST_TOOLCHAIN_ROOT` -> `G_RUST_TOOLCHAIN_DIR`

taskfiles/toolchains.yaml
LinZhihao-723: Done.

taskfiles/toolchains.yaml
LinZhihao-723: Added a comment for why we need to lock it.

LinZhihao-723:
> for the title, how about
> 
> ```
> build(rust): Install Rust toolchain under `build/` instead of `$HOME`; Use `nextest` as the default test runner.
> ```

I think one of the issue with the current installation is it might modify the default installation. How about let's mention it explicitly as:
```
build(rust): Install Rust toolchain under `build/` to avoid modifying the default per-user installation; Use `nextest` as the default test runner.
```

PR #1518, opened 2025-10-28
fix(clp-package): Connect directly to the MongoDB replica (results cache) to avoid failed host resolution of in-container hostname (fixes #1519).

junhaoliao:
# Description

This change updates the results-cache MongoDB client initialization in
`components/clp-package-utils/clp_package_utils/scripts/native/search.py` to pass
`directConnection=True` when reading aggregation outputs. This is needed because the results cache is provisioned as a single-node replica with the in-container hostname; without `directConnection=True`, the Mongo server would serve as a proxy and try to resolve the given host in the initialization before making connections.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

```
task
cd build/clp-package
./sbin/start-clp.sh
./sbin/compress.sh ~/samples/hive-24hr
./sbin/search.sh --count 
# observed `tags: [] count: 42092` printed in in the console in roughly 2 seconds
```

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **Chores**
  * Optimised database connection configuration for improved reliability and performance during data operations.

---

**Note:** This release contains primarily internal infrastructure improvements with no direct user-facing changes.

PR #1598, opened 2025-11-13
fix(webui): Apply direct MongoDB replica connection as in #1518 to resolve dev-env startup issue.

davemarco:
# Description

Moving to docker compose introduced a bug that prevented connection to mongo database when running dev (no bug when actually running package). #1518 provided a fix, but not for webui. Adding fix for webui.

# Checklist

  breaking change.

# Validation performed

Can connect with dev launch command

## Summary by CodeRabbit

* **Chores**
  * Updated MongoDB connection configuration settings to ensure reliable database connectivity.


hoophalab (approved):

components/webui/server/src/plugins/external/mongo.ts
```diff
@@ -6,7 +6,7 @@ import settings from "../../../settings.json" with {type: "json"};
 export const autoConfig = () => {
     return {
         forceClose: true,
-        url: `mongodb://${settings.MongoDbHost}:${settings.MongoDbPort}/${settings.MongoDbName}`,
+        url: `mongodb://${settings.MongoDbHost}:${settings.MongoDbPort}/${settings.MongoDbName}?directConnection=true`,
```
junhaoliao, line 9: do we know if it affects oplog tailing (e.g., `Collection.watch()`)? 

if we can't easily gain confidence that there's no effect, would it be better if we enable it only in the development environment? we can expose a `settings.MongoDbDirectConnection` with default value as false. then we can set it as true in controller.py

alternatively, we can follow the suggestions at https://www.mongodb.com/docs/manual/reference/connection-string-options/#mongodb-urioption-urioption.directConnection about "configuring the cluster to make each MongoDB instance accessible outside of the Docker virtual network", although i haven't investigated what's needed for our Docker Compose setup. (maybe simply adding another hostname when calling rs.initReplicaSet: https://github.com/y-scope/clp/blob/98d56c85714843a3b765f899c6d2acb7853165b8/components/clp-py-utils/clp_py_utils/initialize-results-cache.py#L82 )

davemarco: > do we know if it affects oplog tailing (e.g., Collection.watch())?

Not sure but i guess in dev, it still works
Also this is the same fix you already made in #1518? Maybe we make an issue to expose mongo? Or another PR to expose mongo and revert #1518?

PR #1602, opened 2025-11-14
feat(log-ingestor): Add `SqsListener` to consume S3 notifications via SQS.

LinZhihao-723:
# Description

This PR introduces `SqsListener`, an SQS-based ingestion job implementation. This job listens to an SQS queue that subscribes to an S3 bucket. The S3 bucket will create event notifications, while `SqsListener` will consume these notifications by extracting the S3 object metadata and forwarding them into a downstream receiver (which should be a compression listener introduced in #1552).

Some key decisions made:

* In this version, we assume that the SQS passed into `SqsListener` is dedicated to the particular ingestion job. CLP has the permission to delete any received message after successfully processing it.
  * For now, we won't configure any DLQ service for unused messages.
* We design a client manager trait, `AwsClientManager`, to adapt the need for the coming credential manager service, where we need to refresh the AWS client with temporary credentials. In the current implementation, the client manager implementation should be just a simple wrapper of the raw AWS client.

This PR also adds test cases for emulated SQS services using LocalStack. By default, the tests can be executed with the task (same as the workflow). I also updated the component's README to include instructions for running these tests manually.

# Checklist

  breaking change.

# Validation performed

* [x] Add testing on top of LocalStack to test the basic functionality.

## Summary by CodeRabbit

* **New Features**
  * Configurable AWS endpoint for flexible environments.
  * New SQS listener with lifecycle API to stream relevant S3 object metadata to consumers.
  * AWS client management abstraction to simplify client usage.

* **Tests**
  * LocalStack-enabled integration tests and test helpers for end-to-end ingestion.
  * Test run config updated to execute ignored/integration tests.

* **Documentation**
  * README expanded with Testing quick-start and manual instructions.

* **Chores**
  * Internal size representation adjusted (may affect integrations).

### DIFF
--- a/components/log-ingestor/tests/test_ingestion_job.rs
+++ b/components/log-ingestor/tests/test_ingestion_job.rs
@@ -118,8 +118,11 @@ async fn test_sqs_listener() -> Result<()> {
     }
 
     // Spawn a task to PUT new S3 objects
-    let _ = tokio::spawn(create_s3_objects(s3_client.clone(), objects_to_create.clone()));
-    let _ = tokio::spawn(create_s3_objects(s3_client.clone(), irrelevant_objects));
+    let _creation_handle_0 = tokio::spawn(create_s3_objects(
+        s3_client.clone(),
+        objects_to_create.clone(),
+    ));
+    let _creation_handle_1 = tokio::spawn(create_s3_objects(s3_client.clone(), irrelevant_objects));
 
     // Spawn the SQS listener
     let sqs_listener_config = SqsListenerConfig {


components/log-ingestor/src/compression/listener.rs
LinZhihao-723, line 1: This file is updated so that it's consistent with `SqsListener`.

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -0,0 +1,175 @@
...
+                received_objects.push(object_metadata);
+            }
+            Ok(None) | Err(_) => {
+                break;
+            }
+        }
+    }
+
+    Ok(received_objects)
+}
```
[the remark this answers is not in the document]
LinZhihao-723: Good catch. Will fix it in the next commit.

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -0,0 +1,230 @@
...
+                    .await?
+                    .delete_message()
+                    .queue_url(self.config.queue_url.as_str())
+                    .receipt_handle(receipt_handle)
+                    .send()
+                    .await?;
+            }
+        }
+        Ok(ingested)
+    }
```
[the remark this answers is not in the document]
LinZhihao-723: As explained in the PR description, we will not use a DLQ for now.

components/log-ingestor/README.md
```diff
@@ -2,3 +2,49 @@
...
+
+# Define environment variables required for the tests
+export CLP_LOG_INGESTOR_S3_BUCKET="clp-log-ingestor-test-bucket"
+export CLP_LOG_INGESTOR_SQS_QUEUE="clp-log-ingestor-test-queue"
+
+# Run all tests
+cargo nextest run --all-features --run-ignored all
+
+# Stop and remove the LocalStack container after tests are done
+$REPO_ROOT/tools/scripts/localstack/stop.py --name clp-log-ingestor-tests
```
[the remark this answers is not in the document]
LinZhihao-723: Good catch.

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -0,0 +1,230 @@
...
+/// Configuration for a SQS listener job.
+#[derive(Debug, Clone)]
+pub struct SqsListenerConfig {
+    pub queue_url: String,
+    pub bucket_name: String,
+    pub prefix: String,
+    pub max_num_messages_to_fetch: i32,
+    pub init_polling_backoff_sec: i32,
+    pub max_polling_backoff_sec: i32,
+}
```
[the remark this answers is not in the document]
LinZhihao-723: We shall add validation in the future PR.

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -0,0 +1,230 @@
...
+/// Represents a SQS listener task that listens to SQS messages and extracts S3 object metadata.
+///
+/// # Type Parameters
+///
+/// * [`SqsClientManager`]: The type of the AWS SQS client manager.
+struct SqsListenerTask<SqsClientManager: AwsClientManagerType<Client>> {
+    sqs_client_manager: SqsClientManager,
+    config: SqsListenerConfig,
+    sender: mpsc::Sender<ObjectMetadata>,
+}
```
hoophalab: Why do we need to wrap `aws_sdk_sqs::Client` inside `SqsClientWrapper`/`AwsClientManagerType`? Furthermore, why do we need `AwsClientManagerType`?

I guess `SqsListenerTask` only works iff it holds `aws_sdk_sqs::Client` ?

```suggestion
struct SqsListenerTask {
    client: Client,
    config: SqsListenerConfig,
    sender: mpsc::Sender<ObjectMetadata>,
}
```

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -0,0 +1,230 @@
...
+        if response.messages.is_none() {
+            return Ok(false);
+        }
+        let mut ingested = false;
+        for msg in response.messages.unwrap() {
+            if msg.body.is_none() {
+                continue;
+            }
+
+            let event = match serde_json::from_str::<S3>(msg.body.as_ref().unwrap()) {
```
hoophalab:
```suggestion
    async fn process_sqs_response(&self, response: ReceiveMessageOutput) -> Result<bool> {
        let Some(messages) = response.messages else {
            return Ok(false);
        };
        let mut ingested = false;
        for msg in messages {
            let Some(body) = &msg.body else {
                continue;
            };

            let event = match serde_json::from_str::<S3>(body) {
```
how about more rusty

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -0,0 +1,230 @@
...
+                // Listen to SQS messages.
+                result = self.sqs_client_manager.get().await?
+                    .receive_message()
+                    .queue_url(self.config.queue_url.as_str())
+                    .max_number_of_messages(self.config.max_num_messages_to_fetch)
+                    .wait_time_seconds(polling_backoff_sec).send() => {
+                    polling_backoff_sec = if self.process_sqs_response(result?).await? {
+                        self.config.init_polling_backoff_sec
+                    } else { std::cmp::min(
+                        polling_backoff_sec * 2,
```
hoophalab:
```suggestion
                        polling_backoff_sec.saturating_mul(2),
```

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -0,0 +1,230 @@
...
+
+/// Configuration for a SQS listener job.
+#[derive(Debug, Clone)]
+pub struct SqsListenerConfig {
+    pub queue_url: String,
+    pub bucket_name: String,
+    pub prefix: String,
+    pub max_num_messages_to_fetch: i32,
+    pub init_polling_backoff_sec: i32,
+    pub max_polling_backoff_sec: i32,
```
hoophalab:
```suggestion
    pub max_num_messages_to_fetch: u32,
    pub init_polling_backoff_sec: u32,
    pub max_polling_backoff_sec: u32,
```

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -0,0 +1,230 @@
...
+                () = cancel_token.cancelled() => {
+                    return Ok(());
+                }
+
+                // Listen to SQS messages.
+                result = self.sqs_client_manager.get().await?
+                    .receive_message()
+                    .queue_url(self.config.queue_url.as_str())
+                    .max_number_of_messages(self.config.max_num_messages_to_fetch)
+                    .wait_time_seconds(polling_backoff_sec).send() => {
```
hoophalab, line 66: Do we really need to set `wait_time_seconds` here? Because even if we get an early return with an empty message list, we still send the next query immediately. It might be better to just let `send()` block the future.

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -0,0 +1,178 @@
...
+///
+/// A vector of received S3 object metadata on success.
+///
+/// # Errors
+///
+/// * Forwards [`mpsc::Receiver::recv`]'s return values on failure.
+async fn receive_object_metadata(
+    mut receiver: mpsc::Receiver<ObjectMetadata>,
+    max_num_objects: usize,
+    timeout_sec: u64,
```
hoophalab: Why do we need a timeout here? If we want to make sure the unit test doesn't take forever, then it might be easier to run `tokio::time::timeout(timeout_sec, receive_object_metadata)` outside the function, where `receive_object_metadata` is something like `async fn receive_object_metadata() { while len < max {}: objs.push(receiver.recv()) }`

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -0,0 +1,230 @@
...
+    pub init_polling_backoff_sec: i32,
+    pub max_polling_backoff_sec: i32,
+}
+
+/// Represents a SQS listener task that listens to SQS messages and extracts S3 object metadata.
+///
+/// # Type Parameters
+///
+/// * [`SqsClientManager`]: The type of the AWS SQS client manager.
+struct SqsListenerTask<SqsClientManager: AwsClientManagerType<Client>> {
```
hoophalab: Up to you: it might be easier to remove `Sqs` in names

`SqsListenerConfig -> Config`
`SqsListenerTask -> Task`
`SqsListener -> Listener`

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -0,0 +1,178 @@
...
+        objects_to_create.push(object_metadata);
+    }
+
+    // Spawn a task to PUT new S3 objects
+    let creation_handle = tokio::spawn(create_s3_objects(s3_client, objects_to_create.clone()));
+
+    // Spawn the SQS listener
+    let sqs_listener_config = SqsListenerConfig {
+        queue_url: format!(
+            "{}/{}/{}",
```
hoophalab, line 130: up to you: it might be better to use a url library. But this is just unit test.

components/log-ingestor/tests/aws_config.rs
```diff
@@ -0,0 +1,69 @@
...
+    ///   [`DEFAULT_AWS_SECRET_ACCESS_KEY`] if not set.
+    /// * `AWS_REGION`: The AWS region. Defaults to [`DEFAULT_AWS_REGION`] if not set.
+    /// * `AWS_ACCOUNT_ID`: The AWS account ID. Defaults to [`DEFAULT_AWS_ACCOUNT_ID`] if not set.
+    ///
+    /// # Returns
+    ///
+    /// * `Some(AwsConfig)` if all required environment variables are set.
+    /// * `None` if any required environment variable is missing.
+    #[must_use]
+    pub fn from_env() -> Option<Self> {
```
hoophalab: returning a Result might be more reasonable.

components/log-ingestor/src/ingestion_job/sqs_listener.rs
LinZhihao-723: As explained in the PR description, this is designed for S3 credential manager:
* Now, the client manager should always return the same client.
* In the future, the client may gets expired. The credential manager is responsible for freshing the credential and create a new client for it.

components/log-ingestor/src/ingestion_job/sqs_listener.rs
LinZhihao-723, line 21: Using i32 to be compatible with AWS SQS's API (which takes signed integers)

components/log-ingestor/src/ingestion_job/sqs_listener.rs
LinZhihao-723: We can't wait infinitely since the credential may get expired.

components/log-ingestor/src/ingestion_job/sqs_listener.rs
LinZhihao-723: Did for `SqsListenerTask -> Task` but keep the other two untouched. They will be outported from `ingestor_job` mod, and will be referenced as `ingestion_job::SqsListener` and `ingestion_job::SqsListenerConfig`.

components/log-ingestor/tests/test_ingestion_job.rs
LinZhihao-723: I think it's ok to use `fmt` for this PR.

PR #1606, opened 2025-11-15
feat(clp-package)!: Add Spider config interfaces; Integrate Spider compression orchestration into the Docker Compose project.

sitaowang1998:
# Description

This PR integrates spider compression orchestration into docker compose package by
* Adds Spider into `ClpConfig`. Minor cleanup for unnecessary `dump_to_primitive_type`.
* Adds setup scripts for Spider database.
* Adds native scripts to start multiple Spider workers in container.
* Adds docker compose files to start Spider scheduler and worker containers.

This PR also:
* Bumps `mariadb` python dependency to `>=1.1.14`.
* Bumps `spider-py` version to `0.2.0`.
* Bumps `Spider` to latest version for fix in starting python workers.

This PR is a breaking change because it updates the dependencies versions and and switches the CLP database name configuration in `clp-config.yaml` from `database.name` to `database.names.clp`.

The doc change will be put in #1647.

# Checklist

  breaking change.

# Validation performed

* [x] Spider compression workflow works end-to-end.
* [x] Search works with the compressed archives produces by previous step.
* [ ] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Spider orchestration for compression: scheduler, compression-worker, worker runner, and Spider DB initialization.
  * Optional queue/Redis support and Web UI now recognise Spider DB.

* **Chores**
  * Docker Compose and deployment packages updated with Spider services and shared volume.
  * Config/templates and credentials template augmented with Spider entries; dependency bumps and serialization adjustments.

### DIFF
--- a/components/clp-package-utils/clp_package_utils/controller.py
+++ b/components/clp-package-utils/clp_package_utils/controller.py
@@ -223,6 +223,8 @@ class BaseController(ABC):
             # Bundling
             return EnvVarsDict({"CLP_QUEUE_ENABLED": "0"})
 
+        logger.info("Setting up environment for bundling %s...", component_name)
+
         logs_dir = self._clp_config.logs_directory / component_name
         validate_queue_config(self._clp_config, logs_dir)
 
--- a/tools/deployment/package/docker-compose-all.yaml
+++ b/tools/deployment/package/docker-compose-all.yaml
@@ -337,9 +337,8 @@ services:
       db-table-creator:
         condition: "service_completed_successfully"
     command: [
-      "python3",
-      "-u",
-      "/opt/clp/lib/python3/site-packages/job_orchestration/executor/start-spider-worker.py",
+      "python3", "-u",
+      "-m", "job_orchestration.executor.start-spider-worker",
       "--host", "compression_worker",
       "--num-workers", "${CLP_COMPRESSION_WORKER_CONCURRENCY:-1}",
       "--storage-url", "jdbc:mariadb://database:3306/\


tools/deployment/package/docker-compose-all.yaml
```diff
@@ -323,7 +376,6 @@ services:
   garbage-collector:
     <<: *service_defaults
     hostname: "garbage_collector"
-    stop_grace_period: "10s"
```
junhaoliao, line 341: is this removal intentional?

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -501,6 +501,13 @@ def validate_and_load_redis_credentials_file(
     clp_config.redis.load_credentials_from_file(clp_config.credentials_file_path)
 
 
+def validate_and_load_spider_db_credentials_file(
+    clp_config: ClpConfig, clp_home: pathlib.Path, generate_default_file: bool
+):
+    validate_credentials_file_path(clp_config, clp_home, generate_default_file)
+    clp_config.spider_db.load_credentials_from_file(clp_config.credentials_file_path)
```
junhaoliao: if i'm not wrong, this will read the same credentials for the `database` rather than from a separate key in `credentials.yaml` right? if we support configuring the database to use a different host / port, shall we take the credentials from a different key?

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
```diff
@@ -0,0 +1,297 @@
...
+                logger.error(f"Invalid CLP database user name: {clp_user}")
+                return -1
+
+            db_cursor.execute(f"""CREATE DATABASE IF NOT EXISTS `{db_name}`""")
+            if db_password is not None:
+                db_cursor.execute(
+                    f"""CREATE USER IF NOT EXISTS '{db_user}'@'%' IDENTIFIED BY '{db_password}'"""
+                )
+            else:
+                db_cursor.execute(f"""CREATE USER IF NOT EXISTS '{db_user}'@'%' IDENTIFIED BY ''""")
```
junhaoliao: if we have `ensure_credentials_loaded`, the `password` field should always be available?
```suggestion
            db_cursor.execute(
                    f"""CREATE USER IF NOT EXISTS '{db_user}'@'%' IDENTIFIED BY '{db_password}'"""
                )
```

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
```diff
@@ -0,0 +1,297 @@
...
+
+    try:
+        sql_adapter = SqlAdapter(clp_config.database)
+        with closing(sql_adapter.create_root_mariadb_connection()) as db_conn, closing(
+            db_conn.cursor()
+        ) as db_cursor:
+            db_name = spider_db_config.name
+            db_user = spider_db_config.username
+            db_password = spider_db_config.password
+            clp_user = clp_config.database.username
```
junhaoliao: would renaming the variables make them more readable?

```suggestion
            clp_db_user = clp_config.database.username
            spider_db_name = spider_db_config.name
            spider_db_user = spider_db_config.username
            spider_db_password = spider_db_config.password
```

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -62,6 +63,8 @@ def main(argv):
         validate_and_load_db_credentials_file(clp_config, clp_home, True)
         validate_and_load_queue_credentials_file(clp_config, clp_home, True)
         validate_and_load_redis_credentials_file(clp_config, clp_home, True)
```
junhaoliao: i believe those calls should be conditional as well? like queue and redis can be null objects now

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -763,6 +808,9 @@ def set_up_env(self) -> None:
         env_vars |= self._set_up_env_for_database()
         env_vars |= self._set_up_env_for_queue()
         env_vars |= self._set_up_env_for_redis()
```
junhaoliao, line 940: shall those be called conditionally?

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -630,11 +658,14 @@ class ClpConfig(BaseModel):
 
     package: Package = Package()
     database: Database = Database()
```
junhaoliao: As discussed offline, we should allow users to specify in `clp-config.yaml` which 3rd-party services they want start-clp.sh to start. Otherwise, the behavior is inconsistent because:

* `_set_up_env_for_database` creates directories for database deployment and sets CLP DB config env vars.
* `_set_up_env_for_spider_db` only sets the Spider DB config env vars.

To address this, we're considering adding a `bundled: ["database", "queue", "redis", "results_cache"]` array to `clp-config.yaml` to explicitly define which 3rd-party services should be started. If we do this, we should also create dedicated setup functions for each 3rd-party service deployments so that any deployment-specific initialization logic is cleanly separated.

@coderabiitai create an issue for this

tools/deployment/package/docker-compose-spider-base.yaml
```diff
@@ -0,0 +1,44 @@
...
+  spider-scheduler:
+    extends:
+      file: "docker-compose-all.yaml"
+      service: "spider-scheduler"
+
+  compression-scheduler:
+    extends:
+      file: "docker-compose-all.yaml"
+      service: "compression-scheduler"
+    # No need to depend on Redis and RabbitMQ
```
junhaoliao:
```suggestion
    # Override: Spider does NOT require `queue` or `redis`.
```

tools/deployment/package/docker-compose-spider.yaml
```diff
@@ -0,0 +1,80 @@
...
+  spider-scheduler:
+    extends:
+      file: "docker-compose-all.yaml"
+      service: "spider-scheduler"
+
+  compression-scheduler:
+    extends:
+      file: "docker-compose-all.yaml"
+      service: "compression-scheduler"
+    # No need to depend on Redis and RabbitMQ
```
junhaoliao:
```suggestion
    # Override: Spider does NOT require `queue` or `redis`.
```

components/clp-package-utils/clp_package_utils/general.py
sitaowang1998: For now our plan is to use same mariadb instance, thus same host and address, but different database users (name and passwords) and databases (clp-db v.s. spider-db).

tools/deployment/package/docker-compose-all.yaml
junhaoliao, line 1: let's also update `user-docs/guides-multi-node.md`

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -251,9 +260,31 @@ def transform_for_container(self):
...
+            raise ValueError(f"Spider only support MariaDB storage.")
+        return value
+
+    def get_url(self):
+        self.ensure_credentials_loaded()
+        return f"jdbc:mariadb://{self.host}:{self.port}/{self.name}?user={self.username}&password={self.password}"
+
+
+class SpiderScheduler(BaseModel):
+    host: str = "localhost"
```
junhaoliao: does `SpiderScheduler` (act as a server) and need to be connected by other services? if so, i believe we should add a transform_for_container() method to set the host and port?

```suggestion
    host: DomainStr = "localhost"
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -251,9 +260,31 @@ def transform_for_container(self):
...
 
 
+class SpiderDb(Database):
+
+    name: str = "spider-db"
+
+    @field_validator("type")
+    @classmethod
+    def validate_type(cls, value):
+        if value != "mariadb":
```
junhaoliao: instead of the "mariadb" magic string, can we use the enum value

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -843,6 +871,30 @@ def validate_presto_config(self):
             )
         return self
 
+    @model_validator(mode="after")
+    def validate_spider_config(self):
+        orchestration_type = self.compression_scheduler.type
+        if orchestration_type != OrchestrationType.spider:
+            return self
+        if self.spider_db is None:
+            raise ValueError("SpiderDb config must be set when using spider orchestration.")
```
```suggestion
            raise ValueError("`spider_db` must be configured when using Spider orchestration.")
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -843,6 +871,30 @@ def validate_presto_config(self):
...
+        if self.spider_scheduler is None:
+            raise ValueError("SpiderScheduler config must be set when using spider orchestration.")
```
```suggestion
            raise ValueError("`spider_scheduler` must be configured when using Spider orchestration.")
```

components/clp-package-utils/clp_package_utils/general.py
junhaoliao: sure

> different database users (name and passwords)

so i believe we need to implement `SpiderDb.load_credentials_from_file()` and read from a different key?

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -251,9 +260,31 @@ def transform_for_container(self):
...
 
+class SpiderDb(Database):
+
+    name: str = "spider-db"
+
+    @field_validator("type")
+    @classmethod
+    def validate_type(cls, value):
+        if value != "mariadb":
+            raise ValueError(f"Spider only support MariaDB storage.")
```
```suggestion
            raise ValueError(f"Spider only supports MariaDB for the metadata database.")
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -843,6 +871,30 @@ def validate_presto_config(self):
...
+    def validate_spider_config(self):
+        orchestration_type = self.compression_scheduler.type
+        if orchestration_type != OrchestrationType.spider:
+            return self
+        if self.spider_db is None:
+            raise ValueError("SpiderDb config must be set when using spider orchestration.")
+        if self.spider_scheduler is None:
+            raise ValueError("SpiderScheduler config must be set when using spider orchestration.")
+        if self.database.type != "mariadb":
+            raise ValueError("Database type must be 'mariadb' when using spider orchestration.")
```
junhaoliao: for consistency with the other exception message?
```suggestion
            raise ValueError(f"Spider only supports MariaDB for the metadata database.")
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -843,6 +871,30 @@ def validate_presto_config(self):
...
+        return self
+
+    @model_validator(mode="after")
+    def validate_celery_config(self):
+        orchestration_type = self.compression_scheduler.type
+        if orchestration_type != OrchestrationType.celery:
+            return self
+        if self.queue is None:
+            raise ValueError("Queue config must be set when using celery orchestration.")
```
```suggestion
            raise ValueError("`queue` must be configured when using Celery orchestration.")
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -843,6 +871,30 @@ def validate_presto_config(self):
...
+        if self.redis is None:
+            raise ValueError("Redis config must be set when using celery orchestration.")
```
```suggestion
            raise ValueError("`redis` must be configured when using Celery orchestration.")
```

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
```diff
@@ -0,0 +1,302 @@
...
+
+from pydantic import ValidationError
+from sql_adapter import SqlAdapter
+
+from clp_py_utils.clp_config import ClpConfig
+from clp_py_utils.core import read_yaml_config_file
+
+# Setup logging
+# Create logger
+logger = logging.getLogger(__file__)
```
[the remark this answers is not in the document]
junhaoliao: cc @LinZhihao-723 

https://docs.astral.sh/ruff/rules/invalid-get-logger-argument/ actually guards against `__file__`. that said, i understand the script is to be called as a standalone script, so `__name__` is always `__main__` in the context. shall we supress LOG002 for the line, or use a hardcoded string like "initialize-spider-db", or simply just write `__name__` as the linter likes?

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,60 @@
...
+    parser.add_argument("--storage-url", type=str, required=True, help="Spider storage URL.")
+    parser.add_argument("--host", type=str, required=True, help="Worker host address.")
+    return parser.parse_args()
+
+
+def main() -> None:
+    """Main function to start multiple spider workers."""
+    # Parse arguments
+    args = parse_args()
+    concurrency = args.concurrency
```
junhaoliao: shall we check if this is greater than 1? (as in, would there be issues if 0 is specified?)

components/job-orchestration/job_orchestration/executor/start-spider-worker.py
junhaoliao, line 1: https://docs.astral.sh/ruff/rules/print/#print-t201

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -429,9 +460,6 @@ class S3IngestionConfig(BaseModel):
     type: Literal[StorageType.S3.value] = StorageType.S3.value
     aws_authentication: AwsAuthentication
 
-    def dump_to_primitive_dict(self):
```
junhaoliao, line 514: (just to help me understand) is this unrelated to the spider integration? 

even so, i appreciate we're cleaning it up. can we add this change to the PR description?

components/clp-py-utils/clp_py_utils/create-db-tables.py
```diff
@@ -51,6 +51,14 @@ def main(argv):
     # fmt: on
     subprocess.run(cmd, check=True)
 
+    # fmt: off
+    cmd = [
```
junhaoliao, line 63: maybe launching the spider db init script is a no-op when spider isn't configured, but would it be more clear if we conditionally run the script?

components/clp-py-utils/clp_py_utils/sql_adapter.py
```diff
@@ -104,6 +108,16 @@ def create_connection(self, disable_localhost_socket_connection: bool = False):
         else:
             raise NotImplementedError
 
+    def create_root_mariadb_connection(self, disable_localhost_socket_connection: bool = False):
```
junhaoliao: would it be cleaner to just create another SqlAdapter for the root connection?

junhaoliao: if we have to add this method
```suggestion
    def create_root_mariadb_connection(self, disable_localhost_socket_connection: bool = False) -> mariadb.Connection:
```

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -274,6 +293,40 @@ services:
...
+    depends_on:
+      db-table-creator:
+        condition: "service_completed_successfully"
+    command: [
+      "python3",
+      "-u",
+      "/opt/clp/lib/python3/site-packages/clp_py_utils/start-spider-worker.py",
+      "--concurrency", "${CLP_COMPRESSION_WORKER_CONCURRENCY:-1}",
+      "--storage-url", "${SPIDER_DB_URL}",
+      # NOTE: Leave host to spider scheduler's host. This only affects task placement.
```
junhaoliao: sorry i don't understand this. mind clarifying?

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -274,6 +293,40 @@ services:
...
+        target: "/var/tmp"
+    depends_on:
+      db-table-creator:
+        condition: "service_completed_successfully"
+    command: [
+      "python3",
+      "-u",
+      "/opt/clp/lib/python3/site-packages/clp_py_utils/start-spider-worker.py",
+      "--concurrency", "${CLP_COMPRESSION_WORKER_CONCURRENCY:-1}",
+      "--storage-url", "${SPIDER_DB_URL}",
```
junhaoliao: possible to use the internal network addresses?
e.g. `jdbc:mariadb://database:3306/{self.name}...`

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -274,6 +293,40 @@ services:
...
+      # NOTE: Leave host to spider scheduler's host. This only affects task placement.
+      "--host", "${SPIDER_SCHEDULER_HOST}",
```
junhaoliao: possible to use the internal network addresses?
```suggestion
      "--host", "spider-scheduler",
```

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
```diff
@@ -0,0 +1,297 @@
...
+    CONSTRAINT `output_task_task_id` FOREIGN KEY (`task_id`) REFERENCES `tasks` (`id`) ON UPDATE NO ACTION ON DELETE CASCADE,
+    INDEX (`job_id`, `position`),
+    PRIMARY KEY (`task_id`)
+);
+""",
+    """
+CREATE TABLE IF NOT EXISTS `data`
+(
+    `id`            BINARY(16)     NOT NULL,
+    `value`         BLOB(60000) NOT NULL,
```
junhaoliao, line 105: just curious, why 60000?

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
```diff
@@ -0,0 +1,297 @@
+#!/usr/bin/env python3
+import argparse
+import logging
+import pathlib
+import re
+import sys
+from contextlib import closing
+
+from pydantic import ValidationError
+from sql_adapter import SqlAdapter
```
junhaoliao: should it be
```suggestion
from .sql_adapter import SqlAdapter
```

junhaoliao (changes requested):

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao, line 791: As discussed offline, we should allow users to specify in `clp-config.yaml` which 3rd-party services they want start-clp.sh to start. Otherwise, the behavior is inconsistent because:

* `_set_up_env_for_database` creates directories for database deployment and sets CLP DB config env vars.
* `_set_up_env_for_spider_db` only sets the Spider DB config env vars.

To address this, we're considering adding a `bundled: ["database", "queue", "redis", "results_cache"]` array to `clp-config.yaml` to explicitly define which 3rd-party services should be started. If we do this, we should also create dedicated setup functions for each 3rd-party service deployments so that any deployment-specific initialization logic is cleanly separated.

@coderabbitai create an issue for this

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
sitaowang1998: For sake of sanity, I'll print an error and exit with error code instead.

components/clp-py-utils/clp_py_utils/sql_adapter.py
sitaowang1998: Then we need to create a new `Database` object. I think the current way is cleaner.

tools/deployment/package/docker-compose-all.yaml
sitaowang1998: Each worker gets its address as argument. The address is put into the database and used by scheduler if the task input has locality labels. Since we are not using any locality in current scheduling, we can put anything.

tools/deployment/package/docker-compose-all.yaml
sitaowang1998: I update the `get_url` in `clp_config.py` instead of docker compose file.

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
sitaowang1998: Large enough.

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
sitaowang1998: The other create database script does not have the dot. If we want to change this, maybe in another PR and change every imports.

junhaoliao: sure, do you mind submitting a quick PR for this?

tools/deployment/package/docker-compose-all.yaml
sitaowang1998: Exactly what part should I update?

tools/deployment/package/docker-compose-all.yaml
junhaoliao: i get that generating the JDBC URL only in one place is less error prone. however,

when writing the docker compose files, one of the goals is to keep the interfaces minimal, as in we ask for as least of env var configurations as possible from users. that way, people can still find it manageable to launch the docker-compose-orchestrated package without `start-clp.py`. if all variables in `SPIDER_DB_URL` is repeated from other env vars, i feel it's better we just construct the string in docker-compose.yaml rather than asking users to provide one

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -274,6 +295,40 @@ services:
...
+      PYTHONPATH: "/opt/clp/lib/python3/site-packages"
+      SPIDER_LOG_DIR: "/var/log/compression_worker"
+    volumes:
+      - *volume_clp_config_readonly
+      - *volume_clp_logs
+      - "${CLP_ARCHIVE_OUTPUT_DIR_HOST:-empty}:/var/data/archives"
+      - "${CLP_AWS_CONFIG_DIR_HOST:-empty}:/opt/clp/.aws:ro"
+      - "${CLP_LOGS_INPUT_DIR_HOST:-empty}:${CLP_LOGS_INPUT_DIR_CONTAINER:-/mnt/logs}"
+      - "${CLP_STAGED_ARCHIVE_OUTPUT_DIR_HOST:-empty}:/var/data/staged-archives"
+      - type: "bind"
```
junhaoliao: let's create a `volume_clp_tmp` anchor and avoid repeating this definition in both `spider-compression-worker` and `compression-worker`

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -627,11 +686,14 @@ class ClpConfig(BaseModel):
...
-    queue: Queue = Queue()
-    redis: Redis = Redis()
+    # Default to use celery backend
+    queue: Queue | None = Queue()
+    redis: Redis | None = Redis()
     reducer: Reducer = Reducer()
     results_cache: ResultsCache = ResultsCache()
     compression_scheduler: CompressionScheduler = CompressionScheduler()
+    spider_db: SpiderDb | None = None
+    spider_scheduler: SpiderScheduler | None = None
```
junhaoliao: sorry for missing this in the last review - let's update `components/package-template/src/etc/clp-config.yaml`

tools/deployment/package/docker-compose-all.yaml
junhaoliao: note to self: discuss when we sync up offline

tools/deployment/package/docker-compose-all.yaml
junhaoliao: let's at least add the service descriptions into the tables.

cc @kirkrodrigues 
i'm not sure if we want to create a separate Mermaid chart for the dependency diagram. maybe we can just use subgraphs to highlight the optional components?

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
sitaowang1998: #1640 fixes all other relative imports. We've updated the relative imports for new files in this PR.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -518,7 +520,16 @@ def main(argv):
     logger.info(f"Starting {COMPRESSION_SCHEDULER_COMPONENT_NAME}")
     sql_adapter = SqlAdapter(clp_config.database)
 
-    task_manager = CeleryTaskManager()
+    if clp_config.compression_scheduler.type == OrchestrationType.celery:
```
LinZhihao-723:
```suggestion
    task_manager: CeleryTaskManager | SpiderTaskManager
    if clp_config.compression_scheduler.type == OrchestrationType.celery:
```
Even in Python what u wrote should work, I'd still suggest to forward declare the variable so that the scope is more clear.

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def parse_args() -> argparse.Namespace:
+    """
+    Parses command line arguments.
+    :return: The parsed arguments.
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--concurrency", type=int, default=1, help="Number of concurrent spider workers."
```
LinZhihao-723: How about `num-workers`?

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+        exit(1)
+
+    failed = False
+    for process in processes:
+        exit_code = process.wait()
+        if exit_code != 0:
+            logger.error(f"Spider worker exited with code {exit_code}")
+            failed = True
+    if failed:
+        exit(1)
```
LinZhihao-723:
```suggestion
    exit_code = 0
    for process in processes:
        worker_proc_exit_code = process.wait()
        if worker_proc_exit_code != 0:
            logger.error(f"Spider worker exited with code {worker_proc_exit_code}")
            exit_code = 1

    exit(exit_code)
```

components/clp-py-utils/clp_py_utils/start-spider-worker.py
LinZhihao-723:
```suggestion
        "--concurrency", type=int, default=1, help="Number of concurrent Spider workers."
```
Please check all docstring/comments to ensure we use `Spider` not `spider`.

LinZhihao-723 (changes requested):
Overall lgtm.

components/package-template/src/etc/clp-config.template.json.yaml
junhaoliao, line 1: cc @kirkrodrigues since this is interface related

tools/deployment/package/docker-compose-all.yaml
kirkrodrigues: Yeah, I guess we can discuss it in the review of that PR.

components/package-template/src/etc/clp-config.template.json.yaml
kirkrodrigues: Lgtm

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -274,6 +323,38 @@ def get_clp_connection_params_and_type(
 
         return d
 
+    def get_url(self, user_type: ClpDbUserType = ClpDbUserType.CLP) -> str:
```
junhaoliao: this method doesn't seem used

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -188,13 +217,33 @@ class DbUserCredentials(BaseModel):
...
+
+
 class Database(BaseModel):
     DEFAULT_PORT: ClassVar[int] = 3306
 
     type: DatabaseEngineStr = DatabaseEngine.MARIADB
     host: DomainStr = "localhost"
     port: Port = DEFAULT_PORT
-    name: NonEmptyStr = "clp-db"
+    name: dict[ClpDbNameType, NonEmptyStr] = {
```
junhaoliao: Not introduced by this PR but we should fix in a future PR: we should forbid strings with pure spaces in type NonEmptyStr (or maybe introduce a new type)

otherwise, the name might not be a valid mysql database name

@coderabbitai submit an issue to track

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -288,6 +314,39 @@ services:
...
+      - "${CLP_AWS_CONFIG_DIR_HOST:-empty}:/opt/clp/.aws:ro"
+      - "${CLP_LOGS_INPUT_DIR_HOST:-empty}:${CLP_LOGS_INPUT_DIR_CONTAINER:-/mnt/logs}"
+      - "${CLP_STAGED_ARCHIVE_OUTPUT_DIR_HOST:-empty}:/var/data/staged-archives"
+    depends_on:
+      db-table-creator:
+        condition: "service_completed_successfully"
+    command: [
+      "python3",
+      "-u",
+      "/opt/clp/lib/python3/site-packages/clp_py_utils/start-spider-worker.py",
```
junhaoliao: did we intentionally leave out the `--host` argument and why?

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+    args = parse_args()
+    num_workers = args.num_workers
+    if num_workers < 1:
+        logger.error("Number of concurrent workers must be at least 1.")
+        exit(1)
+    storage_url = args.storage_url
+    host = args.host
+    if host is None:
+        try:
+            host = socket.gethostbyname(socket.gethostname())
```
junhaoliao:
```suggestion
            hostname = socket.gethostname()
            host = socket.gethostbyname(hostname)
```

components/job-orchestration/job_orchestration/executor/start-spider-worker.py
junhaoliao, line 1: would it be better to place this under `components/job-orchestration/job_orchestration/executor`?

components/job-orchestration/job_orchestration/executor/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
+import argparse
```
junhaoliao, line 5:
```suggestion
#!/usr/bin/env python3

"""Script to start multiple Spider workers."""

import argparse
```

components/job-orchestration/job_orchestration/executor/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+import logging
+import os
+import pathlib
+import socket
+import subprocess
```
junhaoliao, line 10:
```suggestion
import subprocess
import sys
```

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def main() -> None:
+    """Main function to start multiple Spider workers."""
+    # Parse arguments
+    args = parse_args()
+    num_workers = args.num_workers
+    if num_workers < 1:
+        logger.error("Number of concurrent workers must be at least 1.")
+        exit(1)
```
junhaoliao:
```suggestion
        sys.exit(1)
```

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+    storage_url = args.storage_url
+    host = args.host
+    if host is None:
+        try:
+            host = socket.gethostbyname(socket.gethostname())
+        except (socket.gaierror, socket.herror) as e:
+            logger.error(f"Failed to resolve hostname: {e}")
+            exit(1)
```
junhaoliao:
```suggestion
            sys.exit(1)
```

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+
+    clp_home = os.getenv("CLP_HOME", "/opt/clp")
+    spider_worker_path = pathlib.Path(clp_home) / "bin" / "spider_worker"
+    if not spider_worker_path.exists():
+        logger.error(f"spider_worker not found at {spider_worker_path}")
+        exit(1)
```
junhaoliao:
```suggestion
        sys.exit(1)
```

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+        for _ in range(num_workers):
+            process = subprocess.Popen(
+                [spider_worker_path, "--storage_url", storage_url, "--host", host]
+            )
+            processes.append(process)
+    except OSError as e:
+        logger.error(f"Failed to start Spider worker: {e}")
+        for process in processes:
+            process.terminate()
+        exit(1)
```
junhaoliao:
```suggestion
        sys.exit(1)
```

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+
+    exit_code = 0
+    for process in processes:
+        worker_proc_exit_code = process.wait()
+        if worker_proc_exit_code != 0:
+            logger.error(f"Spider worker exited with code {worker_proc_exit_code}")
+            exit_code = 1
+
+    exit(exit_code)
```
junhaoliao:
```suggestion
    sys.exit(exit_code)
```

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+    if num_workers < 1:
+        logger.error("Number of concurrent workers must be at least 1.")
+        exit(1)
+    storage_url = args.storage_url
+    host = args.host
+    if host is None:
+        try:
+            host = socket.gethostbyname(socket.gethostname())
+        except (socket.gaierror, socket.herror) as e:
+            logger.error(f"Failed to resolve hostname: {e}")
```
junhaoliao:
```suggestion
            logger.exception("Failed to resolve hostname.")
```

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+            exit(1)
+
+    clp_home = os.getenv("CLP_HOME", "/opt/clp")
+    spider_worker_path = pathlib.Path(clp_home) / "bin" / "spider_worker"
+    if not spider_worker_path.exists():
+        logger.error(f"spider_worker not found at {spider_worker_path}")
```
junhaoliao:
```suggestion
        logger.error("spider_worker not found at %s", spider_worker_path)
```

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+    num_workers = args.num_workers
+    if num_workers < 1:
+        logger.error("Number of concurrent workers must be at least 1.")
+        exit(1)
+    storage_url = args.storage_url
+    host = args.host
+    if host is None:
+        try:
+            host = socket.gethostbyname(socket.gethostname())
+        except (socket.gaierror, socket.herror) as e:
```
junhaoliao: although not clearly documented, it does appear that `socket.gethostbyname()` can throw an `OSError`: https://github.com/python/cpython/blob/b3bf2128989e550a7a02acbaa47389023b2c6bc9/Lib/test/test_socket.py#L1098-L1099

```suggestion
        except (OSError, socket.gaierror, socket.herror):
```

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+    # Start multiple spider workers
+    processes = []
+    try:
+        for _ in range(num_workers):
+            process = subprocess.Popen(
+                [spider_worker_path, "--storage_url", storage_url, "--host", host]
+            )
+            processes.append(process)
+    except OSError as e:
+        logger.error(f"Failed to start Spider worker: {e}")
```
junhaoliao:
```suggestion
    except OSError:
        logger.exception("Failed to start Spider worker.")
```

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+        for process in processes:
+            process.terminate()
+        exit(1)
+
+    exit_code = 0
+    for process in processes:
+        worker_proc_exit_code = process.wait()
+        if worker_proc_exit_code != 0:
+            logger.error(f"Spider worker exited with code {worker_proc_exit_code}")
```
junhaoliao:
```suggestion
            logger.error("Spider worker exited with code %d.", worker_proc_exit_code)
```

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+        exit(1)
+
+    # Start multiple spider workers
+    processes = []
+    try:
+        for _ in range(num_workers):
+            process = subprocess.Popen(
+                [spider_worker_path, "--storage_url", storage_url, "--host", host]
+            )
+            processes.append(process)
```
junhaoliao:
```suggestion
            processes.append(process)
            logger.info(f"Started Spider worker {i+1}/{num_workers} (PID: {process.pid})")
```

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+    spider_worker_path = pathlib.Path(clp_home) / "bin" / "spider_worker"
+    if not spider_worker_path.exists():
+        logger.error(f"spider_worker not found at {spider_worker_path}")
+        exit(1)
+
+    # Start multiple spider workers
+    processes = []
+    try:
+        for _ in range(num_workers):
+            process = subprocess.Popen(
```
junhaoliao: how about renaming this as `proc` to match the names in `sbin` scripts

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,81 @@
...
+                [spider_worker_path, "--storage_url", storage_url, "--host", host]
+            )
+            processes.append(process)
+    except OSError as e:
+        logger.error(f"Failed to start Spider worker: {e}")
+        for process in processes:
+            process.terminate()
```
junhaoliao: Let's adjust higher / lower to allow this run completely: ttps://github.com/y-scope/spider/blob/32711f9d0f18ff4db57407d7b8ccfdc8c0ed866a/src/spider/worker/worker.cpp#L70
```
SPIDER_WORKER_TERM_TIMEOUT_SECONDS = 5
```


```suggestion
            logger.info(f"Terminating worker process {process.pid}")
            process.terminate()

    # Wait for termination with timeout
    for process in processes:
        try:
            process.wait(timeout=SPIDER_WORKER_TERM_TIMEOUT_SECONDS)
        except subprocess.TimeoutExpired:
            logger.warning(f"Worker process {process.pid} did not terminate, sending SIGKILL")
            process.kill()
            process.wait()
```

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
```diff
@@ -0,0 +1,301 @@
...
+def main(argv):
+    args_parser = argparse.ArgumentParser(description="Sets up Spider database.")
+    args_parser.add_argument("--config", "-c", required=True, help="CLP configuration file.")
+    parsed_args = args_parser.parse_args(argv[1:])
+
+    config_path = pathlib.Path(parsed_args.config)
+    try:
+        clp_config = ClpConfig.model_validate(read_yaml_config_file(config_path))
+        clp_config.database.load_credentials_from_env()
+        clp_config.database.load_credentials_from_env(user_type=ClpDbUserType.CLP)
```
junhaoliao: since we load for CLP already
```suggestion
        clp_config.database.load_credentials_from_env(user_type=ClpDbUserType.CLP)
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -178,17 +186,20 @@ def _set_up_env_for_database(self) -> EnvVarsDict:
         # Connection config
         env_vars |= {
             "CLP_DB_HOST": _get_ip_from_hostname(self._clp_config.database.host),
-            "CLP_DB_NAME": self._clp_config.database.name,
+            "CLP_DB_NAME": self._clp_config.database.name[ClpDbNameType.CLP],
             "CLP_DB_PORT": str(self._clp_config.database.port),
+            "SPIDER_DB_NAME": self._clp_config.database.name[ClpDbNameType.SPIDER],
         }
```
junhaoliao:
```suggestion
        }
        if self._clp_config.compression_scheduler.type == OrchestrationType.SPIDER:
            env_vars["SPIDER_DB_NAME"] = self._clp_config.database.name[ClpDbNameType.SPIDER]
```

components/job-orchestration/job_orchestration/scheduler/compress/task_manager/spider_task_manager.py
```diff
@@ -21,8 +20,10 @@ def get_result(self, timeout: float = 0.1) -> list[CompressionTaskResult] | None
             job_results = self._spider_job.get_results()
```
junhaoliao, line 20: not directly related to the PR - `timeout` is unused

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -960,9 +994,17 @@ def _get_docker_file_name(self) -> str:
...
+            case DeploymentType.BASE:
+                return "docker-compose-base.yaml"
+            case DeploymentType.FULL:
+                return "docker-compose.yaml"
+            case DeploymentType.SPIDER_BASE:
+                return "docker-compose-spider-base.yaml"
+            case DeploymentType.SPIDER_FULL:
+                return "docker-compose-spider.yaml"
+            case _:
+                raise ValueError("Unsupported deployment type: %s", deployment_type)
```
junhaoliao: typo: this is not a logging statement
```suggestion
                raise ValueError(f"Unsupported deployment type: {deployment_type}")
```

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
```diff
@@ -0,0 +1,301 @@
...
+import sys
+from contextlib import closing
+
+from pydantic import ValidationError
+
+from clp_py_utils.clp_config import ClpDbNameType
+
+from .clp_config import ClpConfig, ClpDbUserType
+from .core import read_yaml_config_file
+from .sql_adapter import SqlAdapter
```
junhaoliao:
```suggestion
from clp_py_utils.clp_config import ClpConfig, ClpDbNameType, ClpDbUserType
from clp_py_utils.core import read_yaml_config_file
from clp_py_utils.sql_adapter import SqlAdapter
```

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
```diff
@@ -0,0 +1,301 @@
+#!/usr/bin/env python3
```
junhaoliao, line 1:
```suggestion
#!/usr/bin/env python3

"""Script to initialize Spider database."""

```

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
```diff
@@ -0,0 +1,301 @@
...
+    `kv_key`  VARCHAR(64)    NOT NULL,
+    `value`   BLOB(60000) NOT NULL,
+    `task_id` BINARY(16)     NOT NULL,
+    PRIMARY KEY (`task_id`, `kv_key`),
+    CONSTRAINT `kv_data_task_id` FOREIGN KEY (`task_id`) REFERENCES `tasks` (`id`) ON UPDATE NO ACTION ON DELETE CASCADE
+);
+""",
+]
+
+
```
junhaoliao, line 225: for this file:
1. use logging.exception() instead of logging.error wherever appropriate
2. use format strings than f-strings in logging statements

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
```diff
@@ -0,0 +1,301 @@
...
+def main(argv):
```
junhaoliao:
```suggestion
def main(argv: list[str]) -> int:
```

junhaoliao: can we split this into helpers to improve readability?

something like:
```
def _load_clp_config(config_path: pathlib.Path):
    pass

def _validate_database_credentials(clp_config: ClpConfig) -> bool:
    pass

def _setup_spider_database(clp_config: ClpConfig) -> bool:
    pass
```

it's true we're repeating certain logics in multiple places than this scripts - we can also leave it as-is in this PR, submit an issue and fix all together a future PR

components/clp-py-utils/clp_py_utils/pydantic_serialization_utils.py
```diff
@@ -0,0 +1,8 @@
+import pathlib
```
junhaoliao, line 3:
```suggestion
"""Pydantic serialization utilities."""

import pathlib
```

components/package-template/src/etc/credentials.template.yaml
```diff
@@ -13,3 +15,4 @@
 ## Redis credentials
 #redis:
 #  password: "pass"
+#
```
junhaoliao: not sure if it's a github display issue but let's make sure the file only ends with an empty line
```suggestion
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -188,13 +217,33 @@ class DbUserCredentials(BaseModel):
     password: NonEmptyStr
 
 
+def _get_db_name_for_user_type(user_type: ClpDbUserType) -> ClpDbNameType:
```
junhaoliao: instead of a helper method, can we just create a mapping?

```py
_DB_USER_TYPE_TO_DB_NAME_TYPE: MappingProxyType[ClpDbUserType, ClpDbNameType] = MappingProxyType({
    ClpDbUserType.CLP: ClpDbNameType.CLP,
    ClpDbUserType.ROOT: ClpDbNameType.CLP,
    ClpDbUserType.SPIDER: ClpDbNameType.SPIDER,
})
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -960,9 +994,17 @@ def _get_docker_file_name(self) -> str:
         :return: The Docker Compose file name to use based on the config.
         """
         deployment_type = self._clp_config.get_deployment_type()
```
junhaoliao: similar to the other comment, would it be cleaner to create a mapping instead?
```py
_DEPLOYMENT_TYPE_TO_COMPOSE_FILE: MappingProxyType[DeploymentType, str] = MappingProxyType({
    DeploymentType.BASE: "docker-compose-base.yaml",
    DeploymentType.FULL: "docker-compose.yaml",
    DeploymentType.SPIDER_BASE: "docker-compose-spider-base.yaml",
    DeploymentType.SPIDER_FULL: "docker-compose-spider.yaml",
})
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -871,8 +901,12 @@ def set_up_env(self) -> None:
...
         env_vars |= self._set_up_env_for_results_cache_bundling()
         env_vars |= self._set_up_env_for_database()
-        env_vars |= self._set_up_env_for_queue()
-        env_vars |= self._set_up_env_for_redis()
+        if self._clp_config.queue is not None:
+            env_vars |= self._set_up_env_for_queue()
+        if self._clp_config.redis is not None:
+            env_vars |= self._set_up_env_for_redis()
+        if self._clp_config.compression_scheduler.type == OrchestrationType.SPIDER:
+            env_vars |= self._set_up_env_for_spider_scheduler()
```
junhaoliao: Let's always call these helpers and log whether the components will be launched in the helpers. i.e.,
```suggestion
        env_vars |= self._set_up_env_for_queue()
        env_vars |= self._set_up_env_for_redis()
        env_vars |= self._set_up_env_for_spider_scheduler()
```
then
```py
    def _set_up_env_for_queue_bundling(self) -> EnvVarsDict:
        """
        Sets up environment variables and directories for bundling the queue component.

        :return: Dictionary of environment variables necessary to bundle the component.
        """
        component_name = QUEUE_COMPONENT_NAME

        if self._clp_config.queue is None or BundledService.QUEUE not in self._clp_config.bundled:
            logger.info(
                "%s is not configured or part of the 'bundled' configuration, skipping "
                "service bundling...",
                component_name,
            )
            # Bundling
            return EnvVarsDict({"CLP_QUEUE_ENABLED": "0"})
        ...
    def _set_up_env_for_queue(self) -> EnvVarsDict:
        """
        Sets up environment variables for the message queue component.
        :return: Dictionary of environment variables necessary to launch the component.
        """
        component_name = QUEUE_COMPONENT_NAME
        if self._clp_config.queue is None:
            logger.info(
                "%s is not configured, skipping environment setup...",
                component_name,
            )
            return EnvVarsDict()
        ...
```

components/job-orchestration/job_orchestration/executor/start-spider-worker.py
sitaowang1998: Seem all the scripts are placed inside `clp_py_utils`.

components/job-orchestration/job_orchestration/scheduler/compress/task_manager/spider_task_manager.py
sitaowang1998: This is documented in the `TaskManager` abstract method.

components/clp-py-utils/clp_py_utils/initialize-spider-db.py
sitaowang1998: Let's leave it to another PR. @coderabbit Please open an issue to track this.

components/job-orchestration/job_orchestration/executor/start-spider-worker.py
junhaoliao: not necessarily. the reducer's entry point is at:
https://github.com/y-scope/clp/blob/d6c2122737fdd8167063971fd77dd50dedf6e58a/components/job-orchestration/job_orchestration/reducer/reducer.py#L18

and though a bit implicit, the celery work's entry point is at:
https://github.com/y-scope/clp/blob/d6c2122737fdd8167063971fd77dd50dedf6e58a/components/job-orchestration/job_orchestration/executor/compress/celery.py#L8-L9

components/clp-py-utils/clp_py_utils/start-spider-worker.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    Parses command line arguments.
+    :return: The parsed arguments.
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--num-workers", type=int, default=1, help="Number of concurrent Spider workers."
+    )
+    parser.add_argument("--storage-url", type=str, required=True, help="Spider storage URL.")
+    parser.add_argument("--host", type=str, required=False, default=None, help="The worker host.")
```
junhaoliao: hmm... when i posted https://github.com/y-scope/clp/pull/1606#discussion_r2590752564 , i was just curious why we defined a --host arg parser but didn't specify a value. if we always call start-spider-worker.py with --host, i think we can mark this arg as require=True? Then we don't have to handle the default=None as well

components/clp-package-utils/clp_package_utils/controller.py
junhaoliao: This doesn't seem applied yet. Maybe it wasn't pushed?

components/clp-package-utils/clp_package_utils/controller.py
junhaoliao: Sorry i realized i made a typo in the above comment which might have caused confusion. I updated the comment to hopefully clarify the intent - 

Let's also update the "bundling" function to skip the bundling if the queue is not configured as well. Note i also suggested a change in the log message

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -194,7 +233,10 @@ class Database(BaseModel):
     type: DatabaseEngineStr = DatabaseEngine.MARIADB
     host: DomainStr = "localhost"
     port: Port = DEFAULT_PORT
-    name: NonEmptyStr = "clp-db"
+    name: dict[ClpDbNameType, NonEmptyStr] = {
```
junhaoliao: sorry for missing this in the last batch of comments. i believe it should be named "names" now that we take different names?

junhaoliao (changes requested):
following up on the opened discussions. also this one: https://github.com/y-scope/clp/pull/1606#discussion_r2592304789

the rest lgtm

Let's update the PR description a bit to explain what the BREAKING configuration change is:
> and change the configuration file format.

-> and switches the CLP database name configuration in `clp-config.yaml` from `database.name` to `database.names.clp`.

---

for the title, how about:
```
feat(clp-package)!: Add Spider config interfaces; Integrate Spider compression orchestration into the Docker Compose project.
```

components/clp-package-utils/clp_package_utils/controller.py
sitaowang1998: Not a part of this PR, but my PyCharm is flagging a lot of duplicated code fragements, especially in `controller.py`. Maybe we can refactor this in another PR?

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -201,16 +214,15 @@ def _set_up_env_for_queue_bundling(self) -> EnvVarsDict:
...
-        if BundledService.QUEUE not in self._clp_config.bundled:
+        if self._clp_config.queue is None or BundledService.QUEUE not in self._clp_config.bundled:
             logger.info(
-                "%s is not included in the 'bundled' configuration, skipping service bundling...",
+                "%s is not configured or part of the 'bundled' configuration, skipping "
+                "service bundling...",
                 component_name,
             )
             # Bundling
             return EnvVarsDict({"CLP_QUEUE_ENABLED": "0"})
```
junhaoliao, line 225: was the removal unintentional?

```suggestion

        logger.info("Setting up environment for bundling %s...", component_name)

```

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -288,6 +314,40 @@ services:
...
+      - "${CLP_AWS_CONFIG_DIR_HOST:-empty}:/opt/clp/.aws:ro"
+      - "${CLP_LOGS_INPUT_DIR_HOST:-empty}:${CLP_LOGS_INPUT_DIR_CONTAINER:-/mnt/logs}"
+      - "${CLP_STAGED_ARCHIVE_OUTPUT_DIR_HOST:-empty}:/var/data/staged-archives"
+    depends_on:
+      db-table-creator:
+        condition: "service_completed_successfully"
+    command: [
+      "python3",
+      "-u",
+      "/opt/clp/lib/python3/site-packages/job_orchestration/executor/start-spider-worker.py",
```
junhaoliao:
```suggestion
      "python3", "-u",
      "-m", "job_orchestration.executor.start-spider-worker",
```

junhaoliao (approved):

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -194,7 +233,10 @@ class Database(BaseModel):
     type: DatabaseEngineStr = DatabaseEngine.MARIADB
     host: DomainStr = "localhost"
     port: Port = DEFAULT_PORT
-    name: NonEmptyStr = "clp-db"
+    names: dict[ClpDbNameType, NonEmptyStr] = {
+        ClpDbNameType.CLP: "clp-db",
```
junhaoliao, line 237: note this interface change also needs to be ported to the K8s Helm chart at  `tools/deployment/package-helm`. @coderabbitai submit an issue

junhaoliao: a similar issue is found in the MCP server, which is more severe as a bug that prevents the MCP server from starting:
```bash
junhao@ASUS-X870E:~/workspace/clp$ docker logs clp-package-de74-mcp-server-1
2025-12-19 05:20:56,882 mcp_server [ERROR] Failed to start MCP server.
  File "/opt/clp/lib/python3/site-packages/clp_mcp_server/clp_mcp_server.py", line 84, in main
    mcp = create_mcp_server(clp_config)
  File "/opt/clp/lib/python3/site-packages/clp_mcp_server/server/server.py", line 33, in create_mcp_server
    connector = ClpConnector(clp_config)
  File "/opt/clp/lib/python3/site-packages/clp_mcp_server/clp_connector.py", line 35, in __init__
    "db": clp_config.database.name,
  File "/opt/clp/lib/python3/site-packages/pydantic/main.py", line 1026, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'Database' object has no attribute 'name'. Did you mean: 'names'?
```

@coderabbitai submit another issue

PR #1607, opened 2025-11-16
perf(kv-ir): Use string-blob-based encoded text AST implementation in kv-ir deserialization to avoid small string allocation overhead (resolves #1541).

### LINKED ISSUE #1541: Improve kv-ir deserialization speed

LinZhihao-723:
# Description

This PR fixes #1541 by using string-blob-based encoded text AST (aka. `ffi::EncodedTextAst`) in kv-ir deserialization. The PR has the following core changes:

* Add deserialization methods to deserialize and construct `ffi::EncodedTextAst` from an IR stream.
* Replace the use of `ir::EncodedTextAst` by `ffi::EncodedTextAst` in kv-ir and clp-s implementation.
* Add encoding method to directly ingest `ffi::EncodedTextAst` into an archive with decoding and re-parsing.
* Add `parse_and_encode_from` method to `ffi::EncodedTextAst` to avoid duplicating general helpers across different uses (clp-s and test cases).

Some parts of code introduced in this PR duplicate existing code base which deals with `ir::EncodedTextAst`. The reasons to not delete `ir::EncodedTextAst` and fully replace it with `ffi::EncodedTextAst` are:

* Some of the legacy decoding methods (dealing with the old version of encoded text AST) are directly used in our ffi libraries.
* These pieces of code are also used in clp-text for ingesting IRv1 to clp-text archive.

Prefer to defer the legacy code deprecation to a future PR to:

* Keep the scope of this PR smaller (as it's already a bit large)
* Keep clp-text untouched before we have a clearer plan for adding IRv1 deserialization support in kv-ir.

## Performance Benchmark

This PR is tested with a sample IR stream from a private encoded-text-AST-heavy dataset (raw: 8.8GB; zstd-compressed kv-ir: 633MB). All tests are measured in real time.

* Search performance
  * Before: 28.26s
  * After: 17.89s
  * Speedup: 1.58x

* Compression: vs. CLP v0.6.0 (with encoded text AST direct ingestion)
  * Before: 71.43s
  * After:  49.79s
  * Speedup: 1.43x

* Compression: vs. CLP v0.5.0 (without encoded test AST direct ingestion)
  * Before: 88.38s
  * After:  49.79s
  * Speedup: 1.78x

* Compression: vs. JSONL
  * kv-ir: 49.79s
  * JSONL: 48.8s

# Checklist

  breaking change.

# Validation performed

* [x] Check all touched sections pass clang-tidy.
* [x] vs. CLP v0.6.0:
  * Compress the input and decompress the archive to ensure the outputs match.
  * Directly compare the dictionaries and tables inside the archive to ensure they match.

## Summary by CodeRabbit

* **Refactor**
  * Migrated encoded-text handling onto a public FFI surface and updated consumers to use it.
  * Switched decoding paths to explicit result/outcome-based error handling so failures surface deterministically.

* **New Features**
  * Added a parse-and-encode factory and value equality for encoded-text objects.
  * Introduced a compact string-blob type for dictionary data.

* **Tests**
  * Updated tests and utilities to use the new encoding API.

* **Breaking Changes**
  * Removed a decoding error enum member (may affect error-code consumers).

### DIFF
--- a/components/clp-mcp-server/clp_mcp_server/clp_connector.py
+++ b/components/clp-mcp-server/clp_mcp_server/clp_connector.py
@@ -140,8 +140,7 @@ class ClpConnector:
                 break
             if status in error_states:
                 err_msg = (
-                    f"Query job with ID {query_id} ended in "
-                    f"status {QueryJobStatus(status).name}."
+                    f"Query job with ID {query_id} ended in status {QueryJobStatus(status).name}."
                 )
                 raise RuntimeError(err_msg)
             if status not in waiting_states:
@@ -167,9 +166,9 @@ class ClpConnector:
         async for doc in collection.find({}, limit=SEARCH_MAX_NUM_RESULTS):
             doc["link"] = (
                 f"{self._webui_addr}/streamFile?type=json"
-                f'&streamId={doc["archive_id"]}'
+                f"&streamId={doc['archive_id']}"
                 f"&dataset={CLP_DEFAULT_DATASET_NAME}"
-                f'&logEventIdx={doc["log_event_ix"]}'
+                f"&logEventIdx={doc['log_event_ix']}"
             )
             doc["_id"] = None
             results.append(doc)
--- a/components/clp-mcp-server/pyproject.toml
+++ b/components/clp-mcp-server/pyproject.toml
@@ -47,42 +47,7 @@ show_error_context = true
 show_error_end = true
 
 [tool.ruff]
-line-length = 100
-
-[tool.ruff.lint]
-select = ["ALL"]
-ignore = [
-    "ANN401",  # Allow using `Any` type for function signatures
-    "COM812",  # Redundant and conflicts with ruff format
-    "D203",  # No blank line before docstrings (D211)
-    "D205",  # Breaks if summary is larger than one line due to wrapping or if no summary exists
-    "D212",  # Enforce docstring summary line on the next line after quotes (D213)
-    "D400",  # First line of docstrings may not end in period
-    "D401",  # Docstrings should be written in present tense (not imperative)
-    "D415",  # First line of docstrings may not end in a period, question mark, or exclamation point
-    "FBT",  # Allow bool positional parameters since other value positions are allowed
-    "FIX002",  # Allow todo statements
-    "PERF401",  # Allow for loops when creating lists
-    "PERF403",  # Allow for loops when creating dicts
-    "S311",  # Allow usage of `random` package
-    "S603",  # Automatically trust inputs of subprocess execution
-    "SIM102",  # Allow collapsible if statements for readability
-    "SIM300",  # Skip Yoda-condition format fixes
-    "TD002",  # Author unnecessary for todo statement
-    "TD003",  # Issue link unnecessary for todo statement
-    "UP015",  # Explicit open modes are helpful
-]
-isort.order-by-type = false
-
-[tool.ruff.lint.per-file-ignores]
-"tests/**" = [
-    "S101",  # Allow usage of pytest `assert`
-    "TC003",  # Ignore performance overhead of imports only used for type checking
-]
-
-[tool.ruff.format]
-docstring-code-format = true
-docstring-code-line-length = 100
+extend = "../../tools/yscope-dev-utils/exports/lint-configs/python/ruff.toml"
 
 [tool.uv]
 default-groups = ["clp", "dev"]
--- a/components/clp-mcp-server/tests/server/test_utils.py
+++ b/components/clp-mcp-server/tests/server/test_utils.py
@@ -6,7 +6,7 @@ from clp_mcp_server.server.utils import (
     convert_date_string_to_epoch,
     format_query_results,
     parse_timestamp_range,
-    sort_by_timestamp
+    sort_by_timestamp,
 )
 
 
@@ -19,30 +19,25 @@ class TestUtils:
     INVALID_DATE_STRING_ERROR = "Invalid date string"
     INVALID_DATE_STRING_FORMAT_ERROR = "Timestamp must end with 'Z' to indicate UTC."
     INVALID_DATE_STRING_VALUE_ERROR = "is earlier than `formatted_begin_timestamp`"
-    
 
     # Test case: invalid timestamp types.
     INVALID_TYPE_ENTRIES = [
-        {
-            "timestamp": None,
-            "message": '{"message":"Log with None timestamp"}\n',
-            "link": LINK
-        },
+        {"timestamp": None, "message": '{"message":"Log with None timestamp"}\n', "link": LINK},
         {
             "timestamp": "1729267200000",  # str instead of int
             "message": '{"message":"Log with str timestamp"}\n',
-            "link": LINK
+            "link": LINK,
         },
         {
             "timestamp": 1729267200000.0,  # float instead of int
             "message": '{"message":"Log with float timestamp"}\n',
-            "link": LINK
+            "link": LINK,
         },
     ]
     EXPECTED_INVALID_TYPE = [
         f'timestamp: N/A, message: {{"message":"Log with None timestamp"}}\n, link: {LINK}',
         f'timestamp: N/A, message: {{"message":"Log with str timestamp"}}\n, link: {LINK}',
-        f'timestamp: N/A, message: {{"message":"Log with float timestamp"}}\n, link: {LINK}'
+        f'timestamp: N/A, message: {{"message":"Log with float timestamp"}}\n, link: {LINK}',
     ]
 
     # Test case: invalid timestamp values.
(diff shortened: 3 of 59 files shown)


components/core/src/clp/ffi/ir_stream/search/test/test_QueryHandlerImpl.cpp
```diff
@@ -249,10 +251,14 @@ auto get_matchable_values(SchemaTree::Node::Type node_type) -> std::vector<Value
             matchable_values.emplace_back(fmt::format("ThisIs{}", cRefTestStr));
             auto const long_str{fmt::format("This is {}", cRefTestStr)};
             matchable_values.emplace_back(
-                    get_encoded_text_ast<ir::four_byte_encoded_variable_t>(long_str)
+                    EncodedTextAst<ir::four_byte_encoded_variable_t>::parse_and_encode_from(
```
gibber9809:
```suggestion
                    FourByteEncodedTextAst::parse_and_encode_from(
```
Personal taste, but I think its easier to read without templates. Leaving suggestions for the other instances in this file if you agree.

components/core/src/clp/ffi/ir_stream/search/test/test_QueryHandlerImpl.cpp
```diff
@@ -249,10 +251,14 @@ auto get_matchable_values(SchemaTree::Node::Type node_type) -> std::vector<Value
...
+                            long_str
+                    )
             );
             matchable_values.emplace_back(
-                    get_encoded_text_ast<ir::eight_byte_encoded_variable_t>(long_str)
+                    EncodedTextAst<ir::eight_byte_encoded_variable_t>::parse_and_encode_from(
```
gibber9809:
```suggestion
                    EightByteEncodedTextAst::parse_and_encode_from(
```

components/core/src/clp/ffi/ir_stream/search/test/test_QueryHandlerImpl.cpp
```diff
@@ -280,10 +286,14 @@ auto get_unmatchable_values(SchemaTree::Node::Type node_type) -> std::vector<Val
             constexpr std::string_view cUnmatchableLongStr{"This is a static message: ID=0"};
             REQUIRE((cUnmatchableLongStr.find(cRefTestStr) == std::string::npos));
             unmatchable_values.emplace_back(
-                    get_encoded_text_ast<ir::four_byte_encoded_variable_t>(cUnmatchableLongStr)
+                    EncodedTextAst<ir::four_byte_encoded_variable_t>::parse_and_encode_from(
```
gibber9809:
```suggestion
                    FourByteEncodedTextAst::parse_and_encode_from(
```

components/core/src/clp/ffi/ir_stream/search/test/test_QueryHandlerImpl.cpp
```diff
@@ -280,10 +286,14 @@ auto get_unmatchable_values(SchemaTree::Node::Type node_type) -> std::vector<Val
...
+                            cUnmatchableLongStr
+                    )
             );
             unmatchable_values.emplace_back(
-                    get_encoded_text_ast<ir::eight_byte_encoded_variable_t>(cUnmatchableLongStr)
+                    EncodedTextAst<ir::eight_byte_encoded_variable_t>::parse_and_encode_from(
```
gibber9809:
```suggestion
                    EightByteEncodedTextAst::parse_and_encode_from(
```

components/core/src/clp/ffi/ir_stream/search/test/test_QueryHandlerImpl.cpp
```diff
@@ -904,9 +914,9 @@ TEST_CASE("query_handler_evaluation_kv_pair_log_event", "[ffi][ir_stream][search
                 schema_tree,
                 {},
                 {{cArrayNodeId,
-                  Value{
-                          get_encoded_text_ast<ir::four_byte_encoded_variable_t>(unstructured_array)
-                  }}},
+                  Value{EncodedTextAst<ir::four_byte_encoded_variable_t>::parse_and_encode_from(
```
gibber9809:
```suggestion
                  Value{FourByteEncodedTextAst::parse_and_encode_from(
```

components/core/src/clp/ffi/ir_stream/decoding_methods.hpp
```diff
@@ -111,6 +114,22 @@ auto deserialize_encoded_text_ast(
...
+ * @return A result containing the deserialized encoded text AST on success, or an error code
+ * indicating the failure:
+ * - IRErrorCode_Corrupted_IR: if IR stream is invalid.
+ * - IRErrorCode_Incomplete_IR: if IR stream is incomplete.
+ * - Forwards `deserialize_and_append_logtype`'s return values on failure.
+ * - Forwards `deserialize_and_append_dict_var`'s return values on failure.
+ */
+template <ir::EncodedVariableTypeReq encoded_variable_t>
+[[nodiscard]] auto deserialize_encoded_text_ast(ReaderInterface& reader, encoded_tag_t encoded_tag)
+        -> boost::outcome_v2::std_checked<EncodedTextAst<encoded_variable_t>, IRErrorCode>;
```
gibber9809, line 131: Why not `ystdlib::error_handling::Result`? I'm probably missing some context.

components/core/src/clp/ffi/ir_stream/decoding_methods.cpp
```diff
@@ -19,6 +23,124 @@ using std::string;
...
+) -> IRErrorCode;
+
+/**
+ * Deserializes a dictionary variable from the given reader and appends it to the given string blob.
+ * @param reader
+ * @param encoded_tag
+ * @param string_blob The string blob to append the deserialized logtype to.
+ * @return IRErrorCode_Success on success.
+ * @return IRErrorCode_Corrupted_IR if the encoded tag is invalid.
+ * @return IRErrorCode_Incomplete_IR if reader doesn't contain enough data to deserialize.
```
gibber9809:
```suggestion
 * @return IRErrorCode_Incomplete_IR if the reader doesn't contain enough data to deserialize.
```

components/core/src/clp/ffi/ir_stream/decoding_methods.cpp
```diff
@@ -19,6 +23,124 @@ using std::string;
...
 namespace clp::ffi::ir_stream {
+namespace {
+/**
+ * Deserializes a logtype from the given reader and appends it to the given string blob.
+ * @param reader
+ * @param encoded_tag
+ * @param string_blob The string blob to append the deserialized logtype to.
+ * @return IRErrorCode_Success on success.
+ * @return IRErrorCode_Corrupted_IR if the encoded tag is invalid.
+ * @return IRErrorCode_Incomplete_IR if reader doesn't contain enough data to deserialize.
```
gibber9809:
```suggestion
 * @return IRErrorCode_Incomplete_IR if the reader doesn't contain enough data to deserialize.
```

components/core/src/clp/ffi/ir_stream/decoding_methods.cpp
```diff
@@ -19,6 +23,124 @@ using std::string;
...
+        case cProtocol::Payload::VarStrLenUShort: {
+            uint16_t length{};
+            if (false == deserialize_int(reader, length)) {
+                return IRErrorCode_Incomplete_IR;
+            }
+            dict_var_length = length;
+            break;
+        }
+        case cProtocol::Payload::VarStrLenInt: {
+            // NOTE: Using `int32_t` to be match `DictionaryVariableHandler`.
```
gibber9809:
```suggestion
            // NOTE: Using `int32_t` to match `DictionaryVariableHandler`.
```

components/core/src/clp/EncodedVariableInterpreter.hpp
```diff
@@ -168,6 +170,37 @@ class EncodedVariableInterpreter {
...
+     * @param logtype_dict_entry The logtype dictionary entry to populate.
+     * @param var_dict The variable dictionary to add any dictionary variables to.
+     * @param encoded_vars A container to store the encoded variables in.
+     * @param var_ids A container to store the dictionary IDs for dictionary variables.
+     * @return A result containing the estimated number of bytes that this EncodedTextAST would
+     * occupy if it was not encoded in CLP's IR, or an error code indicating the failure:
+     * - Forwards `ffi::EncodedTextAst::decode`'s return values on failure.
+     */
+    template <
+            ir::EncodedVariableTypeReq encoded_variable_t,
```
gibber9809, line 192: Can we also use this requirement for the other `encode_and_add_to_dictionary` method?

gibber9809 (changes requested):
Looks good! Just a few minor comments.

components/core/src/clp/EncodedVariableInterpreter.hpp
LinZhihao-723: Sure. Do u want me to include in this PR?

components/core/src/clp/ffi/ir_stream/decoding_methods.hpp
LinZhihao-723: This has sth to do with the `no_value_policy`, which defaults to sth that checks whether the error type can be converted into `std::errc`. In our case it's not, so we can't use `std_result`.
Ref: https://www.boost.org/doc/libs/latest/libs/outcome/doc/html/reference/types/basic_result.html

components/core/src/clp/EncodedVariableInterpreter.hpp
gibber9809: Yep

LinZhihao-723:
@coderabbitai Create an issue to track that we need to deprecate `clp::ir::EncodedTextAst`.

Issue #1309, opened 2025-09-16
Add support for deploying CLP to Kubernetes clusters using Helm charts.


junhaoliao:
### Request

Currently, CLP package components can only be orchestrated by invoking `docker` subprocesses in `components/clp-package-utils/clp_package_utils/scripts/start_clp.py`.

A request to migrate the orchestration to Docker Compose is currently in progress: #1177 -> #1178 

This request aims to extend the deployment capabilities to support Kubernetes, providing users with a more scalable and production-ready deployment option.

The goal is to provide Helm charts and a Kubernetes "deployment controller" that can provision and deploy all CLP components to a Kubernetes cluster, similar to how the Docker Compose implementation works.

### Possible implementation

1. Helm charts for all components: Implement comprehensive Helm charts for all CLP components with proper configuration
    - Component charts:
      - Infrastructure services:
        - database (MySQL)
        - queue (RabbitMQ)
        - redis
        - results_cache (MongoDB)
      - Application services:
        - compression_scheduler
        - query_scheduler
        - compression_worker
        - query_worker
        - reducer
        - webui
        - garbage_collector
    - Chart features:
      - Persistent storage with PersistentVolumeClaims for data persistence
      - Configuration management using ConfigMaps for package configurations
      - Health checks with liveness and readiness probes for components
      - Resource management with CPU and memory limits/requests
      - Scaling support with configurable replica counts for worker components
      - Service discovery with proper service definitions for inter-component communication

2. Kubernetes "deployment controller" implementation: Implement a dedicated Kubernetes controller to handle deployment to Kubernetes clusters
    - Responsibilities:
      - Configuration translation to convert CLP configuration files to Kubernetes-compatible formats (Helm values)
      - Helm chart deployment to deploy CLP components using Helm charts
      - Namespace management to create and manage Kubernetes namespaces for CLP deployments
      - Platform abstraction to provide the same interface as DockerComposeController for consistent usage
    - Implementation:
      - The Kubernetes controller will inherit from the existing BaseController class, similar to DockerComposeController
      - It will be conditionally imported and used based on a `--platform` flag in the start/stop scripts
      - The controller will support a dry-run mode that generates Helm configuration values without actually starting the deployment, allowing users to tailor the configuration to their specific needs before deploying

3. Updated deployment scripts: Enhance the start/stop scripts to support Kubernetes deployment
    - start_clp.py enhancements:
      - Add `--platform` option with choices `docker` (default) and `k8s`
[...]

PR #1614, opened 2025-11-17
feat(deployment): Set up initial Helm chart for CLP package:

junhaoliao:
- Add database StatefulSet and db-table-creator Job.
- Introduce `lint:helm` task to lint the chart using chart-testing (ct).
- Update CI workflow to include Helm chart linting.

# Description

> [!NOTE]  

This PR adds a Helm chart for deploying the CLP package. It also introduces Helm chart linting and adds it to the CI pipeline.

Changes:

* Helm Chart: Added a new Helm chart (`tools/deployment/package-helm/`) for CLP package deployment, including templates for ConfigMaps, Secrets, Services, StatefulSets, Jobs, and PersistentVolumes.
* CI/CD: Added Helm linting (`ct lint`) to the `clp-lint.yaml` workflow and the `lint.yaml` taskfile.
    * Toolchain: Added a task in `toolchains.yaml` to install the `ct` (chart testing) tool.
	* Configuration Schema: Added `chart_schema.yaml` to define the structure for Helm `Chart.yaml` files and `ct.yaml` to configure the chart testing tool (`ct`).

# Checklist

  breaking change.

# Validation performed

1. Installed kind: https://kind.sigs.k8s.io/docs/user/quick-start/#installing-with-go-install
2. Ran `./test.sh`.
3. `kubectl get pods -w` and observed all pods were started successfully. the db-table-creator-job ran after the `database-statefulset` was ready in around 1 minute.
4. Connected to the database with the configured credentials and observed the tables were created:
   image

## Release Notes

* **New Features**
  * Added Helm chart support for Kubernetes deployment of the CLP package, enabling containerized installations with database configuration, persistent storage, and security management.

* **Chores**
  * Updated linting workflows to validate Helm chart configurations and dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -131,9 +131,12 @@ dependencies = [
  "thiserror 2.0.17",
  "tokio",
  "tokio-stream",
+ "tower-http",
  "tracing",
  "tracing-appender",
  "tracing-subscriber",
+ "utoipa",
+ "utoipa-axum",
 ]
 
 [[package]]
@@ -3985,6 +3988,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "tower-http"
+version = "0.6.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9cf146f99d442e8e68e585f5d798ccd3cad9a7835b917e09728880a862706456"
+dependencies = [
+ "bitflags 2.10.0",
+ "bytes",
+ "http 1.3.1",
+ "pin-project-lite",
+ "tower-layer",
+ "tower-service",
+]
+
+[[package]]
 name = "tower-layer"
 version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4183,6 +4200,43 @@ name = "utf8parse"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "06abde3611657adf66d383f00b093d7faecc7fa57071cce2578660c9f1010821"
+
+[[package]]
+name = "utoipa"
+version = "5.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2fcc29c80c21c31608227e0912b2d7fddba57ad76b606890627ba8ee7964e993"
+dependencies = [
+ "indexmap 2.12.0",
+ "serde",
+ "serde_json",
+ "utoipa-gen",
+]
+
+[[package]]
+name = "utoipa-axum"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7c25bae5bccc842449ec0c5ddc5cbb6a3a1eaeac4503895dc105a1138f8234a0"
+dependencies = [
+ "axum",
+ "paste",
+ "tower-layer",
+ "tower-service",
+ "utoipa",
+]
+
+[[package]]
+name = "utoipa-gen"
+version = "5.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6d79d08d92ab8af4c5e8a6da20c47ae3f61a0f1dabc1997cdf2d082b757ca08b"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "regex",
+ "syn",
+]
 
 [[package]]
 name = "uuid"
--- a/components/api-server/Cargo.toml
+++ b/components/api-server/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2024"
 
 [lib]
-name = "clp_client"
+name = "api_server"
 path = "src/lib.rs"
 
 [[bin]]
@@ -29,6 +29,9 @@ sqlx = { version = "0.8.6", features = ["runtime-tokio", "mysql"] }
 thiserror = "2.0.17"
 tokio = { version = "1.48.0", features = ["full"] }
 tokio-stream = { version = "0.1.17", features = ["fs"] }
+tower-http = { version = "0.6.7", features = ["cors"] }
 tracing = "0.1.41"
 tracing-appender = "0.2.3"
 tracing-subscriber = { version = "0.3.20", features = ["json", "env-filter", "fmt", "std"] }
+utoipa = { version = "5.4.0", features = ["axum_extras"] }
+utoipa-axum = "0.2.0"
--- a/components/api-server/src/bin/api_server.rs
+++ b/components/api-server/src/bin/api_server.rs
@@ -1,20 +1,6 @@
 use anyhow::Context;
-use axum::{
-    Json,
-    Router,
-    extract::{Path, State},
-    http::StatusCode,
-    response::{
-        IntoResponse,
-        Sse,
-        sse::{Event, KeepAlive},
-    },
-    routing::{get, post},
-};
 use clap::Parser;
 use clp_rust_utils::{clp_config::package, serde::yaml};
-use futures::{Stream, StreamExt};
-use thiserror::Error;
 use tracing_appender::rolling::{RollingFileAppender, Rotation};
 use tracing_subscriber::{self, fmt::writer::MakeWriterExt};
 
@@ -95,75 +81,15 @@ async fn main() -> anyhow::Result<()> {
         .await
         .context(format!("Cannot listen to {addr}"))?;
 
-    let client = clp_client::Client::connect(&config, &credentials)
+    let client = api_server::client::Client::connect(&config, &credentials)
         .await
         .context("Cannot connect to CLP")?;
 
-    let app = Router::new()
-        .route("/", get(health))
-        .route("/health", get(health))
-        .route("/query", post(query))
-        .route("/query_results/{search_job_id}", get(query_results))
-        .with_state(client);
+    let router = api_server::routes::from_client(client)?;
 
     tracing::info!("Server started at {addr}");
-    axum::serve(listener, app)
+    axum::serve(listener, router)
         .with_graceful_shutdown(shutdown_signal())
         .await?;
     Ok(())
-}
-
-async fn health() -> String {
-    "API server is running".to_owned()
-}
-
-async fn query(
-    State(client): State<clp_client::Client>,
-    Json(query_config): Json<clp_client::QueryConfig>,
-) -> Result<Json<serde_json::Value>, HandlerError> {
-    let search_job_id = client.submit_query(query_config).await?;
-    let uri = format!("/query_results/{search_job_id}");
-    Ok(Json(serde_json::json!({"query_results_uri": uri})))
-}
-
-async fn query_results(
-    State(client): State<clp_client::Client>,
-    Path(search_job_id): Path<u64>,
-) -> Result<Sse<impl Stream<Item = Result<Event, HandlerError>>>, HandlerError> {
-    let results_stream = client.fetch_results(search_job_id).await?;
-    let event_stream = results_stream.map(|res| {
-        let message = res?;
-        let trimmed_message = message.trim();
-        if trimmed_message.lines().count() != 1 {
-            return Err(HandlerError::InternalServer);
-        }
-        Ok(Event::default().data(trimmed_message))
-    });
-    Ok(Sse::new(event_stream).keep_alive(KeepAlive::default()))
-}
-
-/// Generic errors for request handlers.
-#[derive(Error, Debug)]
-enum HandlerError {
-    #[error("Internal server error")]
-    InternalServer,
-}
-
-trait IntoHandlerError {}
-
-impl IntoHandlerError for axum::Error {}
-
-impl IntoHandlerError for clp_client::ClientError {}
-
-impl<T: IntoHandlerError> From<T> for HandlerError {
-    fn from(_: T) -> Self {
-        Self::InternalServer
-    }
-}
-
-/// Converts [`HandlerError`] into an HTTP response.
-impl IntoResponse for HandlerError {
-    fn into_response(self) -> axum::response::Response {
-        StatusCode::INTERNAL_SERVER_ERROR.into_response()
-    }
 }
--- a/components/api-server/src/client.rs
+++ b/components/api-server/src/client.rs
@@ -13,24 +13,39 @@ use futures::{Stream, StreamExt};
 use pin_project_lite::pin_project;
 use serde::{Deserialize, Serialize};
 use sqlx::Row;
+use utoipa::ToSchema;
 
 pub use crate::error::ClientError;
 
 /// Defines the request configuration for submitting a search query.
-#[derive(Clone, Serialize, Deserialize)]
+#[derive(Clone, Serialize, Deserialize, ToSchema)]
 #[serde(deny_unknown_fields)]
 pub struct QueryConfig {
+    /// The search query as a KQL string.
     pub query_string: String,
+
+    /// The dataset to search within. If not provided, only `default` dataset will be searched.
     #[serde(default)]
     pub dataset: Option<String>,
+
+    /// The maximum number of results to return. Set to `0` for no limit.
     #[serde(default)]
     pub max_num_results: u32,
+
+    /// The beginning timestamp (in epoch milliseconds) for the search range (inclusive).
     #[serde(default)]
     pub begin_timestamp: Option<i64>,
+
+    /// The ending timestamp (in epoch milliseconds) for the search range (inclusive).
     #[serde(default)]
     pub end_timestamp: Option<i64>,
+
+    /// Whether the string match should be case-insensitive.
     #[serde(default)]
     pub ignore_case: bool,
+
+    /// Whether to write the search results to files. If `false`, results will be stored in
+    /// `MongoDB`.
     #[serde(default)]
     pub write_to_file: bool,
 }
--- a/components/api-server/src/lib.rs
+++ b/components/api-server/src/lib.rs
@@ -1,3 +1,3 @@
-mod client;
+pub mod client;
 mod error;
-pub use client::*;
+pub mod routes;
--- a/components/api-server/src/routes.rs
+++ b/components/api-server/src/routes.rs
@@ -1,0 +1,172 @@
+use axum::{
+    Json,
+    extract::{Path, State},
+    http::StatusCode,
+    response::{
+        IntoResponse,
+        Sse,
+        sse::{Event, KeepAlive},
+    },
+    routing::get,
+};
+use futures::{Stream, StreamExt};
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+use tower_http::cors::{Any, CorsLayer};
+use utoipa::{OpenApi, ToSchema};
+use utoipa_axum::{router::OpenApiRouter, routes};
+
+use crate::client::{Client, ClientError, QueryConfig};
+
+/// Factory method to create an Axum router configured with all API routes.
+///
+/// # Returns
+///
+/// A newly created [`axum::Router`] instance configured with all application routes.
+///
+/// # Errors
+///
+/// Returns an error if:
+///
+/// * Forwards [`OpenApi::to_json`]'s return values on failure.
+pub fn from_client(client: Client) -> Result<axum::Router, serde_json::Error> {
+    let (router, api) = OpenApiRouter::with_openapi(ApiDoc::openapi())
+        .route("/", get(health))
+        .routes(routes!(health))
+        .routes(routes!(query))
+        .routes(routes!(query_results))
+        .with_state(client)
+        .split_for_parts();
+    let api_json = api.to_json()?;
+    let router = router
+        .route(
+            "/openapi.json",
+            get(|| async { (StatusCode::OK, api_json) }),
+        )
+        .layer(CorsLayer::new().allow_origin(Any));
+    Ok(router)
+}
+
+// `utoipa::OpenApi` triggers `clippy::needless_for_each`
+#[allow(clippy::needless_for_each)]
+mod api_doc {
+    #[derive(utoipa::OpenApi)]
+    #[openapi(info(
+        title = "API Server",
+        description = "API Server for CLP",
+        contact(name = "YScope")
+    ))]
+    pub struct ApiDoc;
+}
+use api_doc::ApiDoc;
+
+#[utoipa::path(
+    get,
+    path = "/health",
+    responses((status = OK, body = String))
+)]
+async fn health() -> String {
+    "API server is running".to_owned()
+}
+
+#[utoipa::path(
+    post,
+    path = "/query",
+    description = "Submits a new query job.",
+    request_body(
+        content= QueryConfig,
+        example = json!({
+            "query_string": "*",
+            "dataset": "default",
+            "begin_timestamp": 0,
+            "end_timestamp": 17_356_896,
+            "ignore_case": true,
+            "max_num_results": 0,
+            "write_to_file": false
+        })),
+    responses(
+        (
+            status = OK,
+            body = QueryResultsUri,
+            description = "The URI to fetch the results of the submitted query.",
+            example = json!({"query_results_uri":"/query_results/1"})
+        ),
+        (status = INTERNAL_SERVER_ERROR)
+    )
+)]
+async fn query(
+    State(client): State<Client>,
+    Json(query_config): Json<QueryConfig>,
+) -> Result<Json<QueryResultsUri>, HandlerError> {
+    let search_job_id = client.submit_query(query_config).await?;
+    let uri = format!("/query_results/{search_job_id}");
+    Ok(Json(QueryResultsUri {
+        query_results_uri: uri,
+    }))
+}
+
+#[derive(Clone, Serialize, Deserialize, ToSchema)]
+#[serde(deny_unknown_fields)]
+struct QueryResultsUri {
+    /// The uri to get the query results.
+    query_results_uri: String,
+}
+
+#[utoipa::path(
+    get,
+    path = "/query_results/{search_job_id}",
+    description = "Streams the results of a previously submitted query as Server-Sent Events \
+        (SSE).",
+    responses(
+        (
+            status = OK,
+            body = String,
+            content_type = "text/event-stream",
+            description = "Server-Sent Events stream of query results. Each event contains a \
+                single line of the query result in JSON format.",
+            example = r#"data: {"timestamp": 1633036800, "message": "Example log message"}\n"#
+        ),
+        (status = INTERNAL_SERVER_ERROR)
+    )
+)]
+async fn query_results(
+    State(client): State<Client>,
+    Path(search_job_id): Path<u64>,
+) -> Result<Sse<impl Stream<Item = Result<Event, HandlerError>>>, HandlerError> {
+    let results_stream = client.fetch_results(search_job_id).await?;
+    let event_stream = results_stream.map(|res| {
+        let message = res?;
+        let trimmed_message = message.trim();
+        if trimmed_message.lines().count() != 1 {
+            return Err(HandlerError::InternalServer);
+        }
+        Ok(Event::default().data(trimmed_message))
+    });
+    Ok(Sse::new(event_stream).keep_alive(KeepAlive::default()))
+}
+
+/// Generic errors for request handlers.
+#[derive(Error, Debug)]
+enum HandlerError {
+    #[error("Internal server error")]
+    InternalServer,
+}
+
+trait IntoHandlerError {}
+
+impl IntoHandlerError for axum::Error {}
+
+impl IntoHandlerError for ClientError {}
+
+impl<T: IntoHandlerError> From<T> for HandlerError {
+    fn from(_: T) -> Self {
+        Self::InternalServer
+    }
+}
+
+/// Converts [`HandlerError`] into an HTTP response.
+impl IntoResponse for HandlerError {
+    fn into_response(self) -> axum::response::Response {
+        StatusCode::INTERNAL_SERVER_ERROR.into_response()
+    }
+}
--- a/components/clp-mcp-server/clp_mcp_server/clp_connector.py
+++ b/components/clp-mcp-server/clp_mcp_server/clp_connector.py
@@ -8,16 +8,13 @@ import msgpack
 from clp_py_utils.clp_config import CLP_DEFAULT_DATASET_NAME
 from pymongo import AsyncMongoClient
 
-from .constants import (
+from clp_mcp_server.constants import (
     POLLING_INTERVAL_SECONDS,
     QueryJobStatus,
     QueryJobType,
     SEARCH_MAX_NUM_RESULTS,
 )
-from .settings import (
-    CLP_DB_PASS,
-    CLP_DB_USER,
-)
+from clp_mcp_server.settings import CLP_DB_PASS, CLP_DB_USER
 
 
 class ClpConnector:
--- a/components/clp-mcp-server/clp_mcp_server/clp_mcp_server.py
+++ b/components/clp-mcp-server/clp_mcp_server/clp_mcp_server.py
@@ -13,7 +13,7 @@ from clp_py_utils.clp_logging import get_logger, get_logging_formatter, set_logg
 from clp_py_utils.core import read_yaml_config_file
 from pydantic import ValidationError
 
-from .server import create_mcp_server
+from clp_mcp_server.server import create_mcp_server
 
 logger = get_logger(MCP_SERVER_COMPONENT_NAME)
 
--- a/components/clp-mcp-server/clp_mcp_server/server/__init__.py
+++ b/components/clp-mcp-server/clp_mcp_server/server/__init__.py
@@ -1,5 +1,5 @@
 """Server module."""
 
-from .server import create_mcp_server
+from clp_mcp_server.server.server import create_mcp_server
 
 __all__ = ["create_mcp_server"]
--- a/components/clp-mcp-server/clp_mcp_server/server/server.py
+++ b/components/clp-mcp-server/clp_mcp_server/server/server.py
@@ -8,10 +8,13 @@ from starlette.requests import Request
 from starlette.responses import PlainTextResponse
 
 from clp_mcp_server.clp_connector import ClpConnector
-
-from . import constants
-from .session_manager import SessionManager
-from .utils import format_query_results, parse_timestamp_range, sort_by_timestamp
+from clp_mcp_server.server import constants
+from clp_mcp_server.server.session_manager import SessionManager
+from clp_mcp_server.server.utils import (
+    format_query_results,
+    parse_timestamp_range,
+    sort_by_timestamp,
+)
 
 
 def create_mcp_server(clp_config: ClpConfig) -> FastMCP:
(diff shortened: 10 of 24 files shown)


chart_schema.yaml
```diff
@@ -0,0 +1,38 @@
+# yamllint disable rule:document-start
```
junhaoliao: this is the only modified line

chart_schema.yaml
junhaoliao, line 1: the file was extracted from [chart-testing_3.14.0_linux_amd64.tar.gz](https://github.com/helm/chart-testing/releases/download/v3.14.0/chart-testing_3.14.0_linux_amd64.tar.gz)

lint-requirements.txt
```diff
@@ -1,6 +1,12 @@
...
 # Lock to v19.x until we can upgrade our code to fix new v20 issues.
 clang-tidy~=19.1
+
+# Python
+black>=24.4.2
 ruff>=0.4.4
+
+# YAML
 yamllint>=1.35.1
+yamale>=6.0.0
```
junhaoliao: chart-testing depends on not only yamllint but also yamale

taskfiles/lint.yaml
```diff
@@ -54,6 +56,19 @@ tasks:
...
+  helm:
+    aliases:
+      - "check-helm"
+      - "fix-helm"
+    deps:
+      - "toolchains:helm"
+      - "venv"
+    cmds:
+      - |-
+        . "{{.G_LINT_VENV_DIR}}/bin/activate"
```
junhaoliao, line 68: to access yamllint and yamale

taskfiles/lint.yaml
```diff
@@ -98,16 +113,19 @@ tasks:
...
-          .github \
-          .yamllint.yml \
-          components/core/.clang-format \
-          components/core/config \
-          components/core/tools/ \
-          components/package-template/src/etc \
-          docs \
-          taskfile.yaml \
-          taskfiles \
-          tools/deployment
```
junhaoliao, line 110: to exclude tools/deployment/package-helm

taskfiles/toolchains.yaml
```diff
@@ -15,6 +18,41 @@ vars:
...
+          FILE_SHA256: >-
+            {{- if and (eq OS "linux") (eq ARCH "amd64") -}}
+              d16f0583616885423826241164ce1f6589c6fe5332fa74f374ebd2bd3cb3fe1f
+            {{- else if and (eq OS "darwin") (eq ARCH "arm64") -}}
+              db10dbbb42b110c7a5da5a3202908f32ad2ca6ad600d423426dc7886d09aad07
+            {{- else -}}
+              unsupported
+            {{- end }}
+          INCLUDE_PATTERNS: ["ct"]
+          NUM_COMPONENTS_TO_STRIP: "0"
```
junhaoliao, line 43: can't simply write `0`. see https://github.com/y-scope/yscope-dev-utils/issues/96

taskfiles/toolchains.yaml
```diff
@@ -15,6 +18,41 @@ vars:
...
+      - |-
+        export PATH="{{.G_HELM_TOOLCHAIN_DIR}}:$PATH"
```
junhaoliao, line 45: the get-helm-4 script, after installing `helm`, verifies whether it can run `helm`, so we need to prepend the G_HELM_TOOLCHAIN_DIR to the PATH.

taskfiles/toolchains.yaml
```diff
@@ -15,6 +18,41 @@ vars:
...
+        export HELM_INSTALL_DIR="{{.G_HELM_TOOLCHAIN_DIR}}"
+        curl --proto "=https" --tlsv1.2 --silent --show-error --fail \
+        https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-4 | bash -s -- --no-sudo
+
+      # Create an environment file to configure the custom Helm toolchain location.
+      - |-
+        cat <<"EOF" > "{{.G_HELM_TOOLCHAIN_ENV_FILE}}"
```
junhaoliao, line 52: the `"EOF"` needs to be quoted to avoid expanding the `$PATH` in the task run environment

tools/deployment/package-helm/.helmignore
junhaoliao, line 1: this was created by `helm create`

tools/deployment/package-helm/Chart.yaml
junhaoliao, line 1: this was created by `helm create` and modified to include CLP specific information

tools/deployment/package-helm/templates/_helpers.tpl
junhaoliao, line 1: this was created by `helm create` and modified to include CLP specific helpers

tools/deployment/package-helm/templates/configmap.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: "v1"
+kind: "ConfigMap"
+metadata:
+  name: {{ include "clp.fullname" . }}-config
+  labels:
+    {{- include "clp.labels" . | nindent 4 }}
+data:
+  clp-config.yml: |
```
junhaoliao: this was partially copied from `var/log/.clp-config.yml` in an initialized clp-package, with modifications to use template variables

tools/deployment/package-helm/templates/configmap.yaml
```diff
@@ -0,0 +1,33 @@
...
+      name: {{ .Values.clpConfig.database.name | quote }}
+      port: 3306
+      ssl_cert: null
+      type: {{ .Values.clpConfig.database.type | quote }}
+    logs_directory: "/var/log"
+    package:
+      query_engine: {{ .Values.clpConfig.package.query_engine | quote }}
+      storage_engine: {{ .Values.clpConfig.package.storage_engine | quote }}
+
+  mysql-logging.cnf: |
```
junhaoliao, line 23: this was copied from https://github.com/y-scope/clp/blob/main/components/package-template/src/etc/mysql/conf.d/logging.cnf without any modification

tools/deployment/package-helm/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+    type: "mariadb"  # "mariadb" or "mysql"
+    host: "localhost"
+    port: 30306
+    name: "clp-db"
+
+  # Location where other data (besides archives) are stored. It will be created if
+  # it doesn't exist.
+  # NOTE: This directory must not overlap with any path used in CLP's execution container. An error
+  # will be raised if so.
+  data_directory: "/tmp/clp/var/data"
```
junhaoliao, line 36: note only absolute paths are accepted now

.github/workflows/clp-lint.yaml
```diff
@@ -78,3 +79,7 @@ jobs:
       - name: "Lint .yaml files"
         shell: "bash"
         run: "task lint:check-yaml"
+
+      - name: "Lint Helm charts"
```
junhaoliao, line 86: we could just use https://github.com/helm/chart-testing-action but it's not doing anything fancier than installing yamllint, yamale (and also `kind` which we don't need in any GH workflow). if needed in the future, we can also set up a `kind` cluster in our taskfiles.

.github/workflows/clp-lint.yaml
```diff
@@ -25,6 +25,7 @@ jobs:
     steps:
       - uses: "actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683"
         with:
+          fetch-depth: 0
```
junhaoliao: otherwise `helm lint` would complain about not finding the `origin/main` branch. see https://github.com/helm/chart-testing-action/issues/133

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -0,0 +1,141 @@
...
+
+{{/*
+Creates timings for liveness probes.
+*/}}
+{{- define "clp.livenessProbeTimings" -}}
+initialDelaySeconds: 30
+periodSeconds: 10
+timeoutSeconds: 5
+failureThreshold: 3
+{{- end }}
```
junhaoliao, line 95: note the numbers are different from the docker-compose healthcheck configurations. feel free to disagree

tools/deployment/package-helm/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: "v2"
+name: "clp"
+description: "A Helm chart for CLP (Compressed Log Processor) package deployment"
+type: "application"
+version: "0.1.0"
+appVersion: "0.6.1-dev"
+home: "https://github.com/y-scope/clp"
+icon: "https://github.com/y-scope/clp/raw/main/docs/src/clp-logo.png"
+sources: ["https://github.com/y-scope/clp"]
+maintainers: [{name: "y-scope", email: "dev@yscope.com"}]
```
junhaoliao: the name needs to be a valid github org / user name

lint-requirements.txt
kirkrodrigues: Alphabetize

.github/workflows/clp-lint.yaml
kirkrodrigues: Can we add a comment?

tools/deployment/package-helm/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: "v2"
+name: "clp"
+description: "A Helm chart for CLP (Compressed Log Processor) package deployment"
+type: "application"
+version: "0.1.0"
```
kirkrodrigues, line 3: Minor but can we move version below name so that all the required fields are grouped together?

tools/deployment/package-helm/Chart.yaml
kirkrodrigues:
```suggestion
maintainers: [{name: "y-scope", email: "dev@yscope.com", url: "https://yscope.com"}]
```

tools/deployment/package-helm/templates/database-statefulset.yaml
```diff
@@ -0,0 +1,106 @@
...
+              mountPath: "/etc/mysql/conf.d/logging.cnf"
+              subPath: "mysql-logging.cnf"
+              readOnly: true
+          livenessProbe:
+            {{- include "clp.livenessProbeTimings" . | nindent 12 }}
+            exec: &database-health-check
+              command: [
+                "mysqladmin", "ping",
+                "--silent",
+                "-h", "127.0.0.1"
```
kirkrodrigues: Do we need to specify the username and password like we did for Docker Compose?

tools/deployment/package-helm/templates/db-table-creator-job.yaml
```diff
@@ -0,0 +1,59 @@
...
+      securityContext:
+        runAsUser: {{ .Values.securityContext.firstParty.uid }}
+        runAsGroup: {{ .Values.securityContext.firstParty.gid }}
+        fsGroup: {{ .Values.securityContext.firstParty.gid }}
+      initContainers:
+        - name: "wait-for-database"
+          image: "busybox:latest"
+          command: [
+            "sh", "-c",
+            "until nc -z {{ include "clp.fullname" . }}-database 3306; do echo waiting for database; sleep 2; done"
```
kirkrodrigues: Break line?

tools/deployment/package-helm/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+image:
+  clpPackage:
+    repository: "ghcr.io/y-scope/clp/clp-package"
+    pullPolicy: "IfNotPresent"
+    tag: "nightly"
+
+clpConfig:
+  package:
+    storage_engine: "clp-s"
+    query_engine: "clp-s"
```
kirkrodrigues, line 24: How would the choice between clp and clp-s work?

tools/deployment/package-helm/templates/_helpers.tpl
kirkrodrigues: Why do we want to change them? Is Kubernetes faster or something?

tools/deployment/package-helm/templates/configmap.yaml
```diff
@@ -0,0 +1,33 @@
...
+kind: "ConfigMap"
+metadata:
+  name: {{ include "clp.fullname" . }}-config
+  labels:
+    {{- include "clp.labels" . | nindent 4 }}
+data:
+  clp-config.yml: |
+    data_directory: "/var/data"
+    database:
+      auto_commit: false
```
kirkrodrigues, line 11: Why do we need to set parameters like auto_commit that we don't usually set in clp-config.yaml?

tools/deployment/package-helm/values.yaml
junhaoliao: sorry, not sure if i get this. mind clarifying the question?

kirkrodrigues: How are we going to have clp-json and clp-text packages in the Kubernetes flow?

junhaoliao: ah, users can just configure these two as both "clp-s" or both "clp"

junhaoliao: as discussed offline, there are mainly two ways to install the Helm chart:
1. Users may pull our git repo directly, make modifications to `tools/deployment/package-helm/values.yaml` and run `helm install .` to install the updated chart.
2. Users may also `helm install <release-name> <release-url> --set clpConfig.storage_engine="clp" --set clpConfig.query_engine="clp"` or  `helm install <release-name> <release-url> -f value-override.yaml`, where the `<release-url>` can be a release tar or a HTTP static server (e.g., public s3, GitHub Pages).

tools/deployment/package-helm/templates/database-statefulset.yaml
junhaoliao: we should. previously i removed the arguments and found services were up without restarting and thought that was fine, but actually the pod was reporting bad health. i updated the command to include the credentials now

tools/deployment/package-helm/templates/configmap.yaml
junhaoliao: This was copied from the generated `var/.clp-config.yaml`. there's probably no difference but would it be better to keep it for explicitness?  (easier diff as well

tools/deployment/package-helm/templates/database-logs-pv.yaml
```diff
@@ -0,0 +1,8 @@
+{{- include "clp.createLocalPv" (dict
+  "root" .
+  "name" (printf "%s-database-logs" (include "clp.fullname" .))
+  "component" "database"
+  "capacity" "5Gi"
+  "accessModes" (list "ReadWriteOnce")
+  "hostPath" (printf "%s/database" .Values.clpConfig.logs_directory)
+) }}
```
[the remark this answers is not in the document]
junhaoliao: I can be wrong but the rancher/local-path-provisioner that i have previously tried can't be specified in any Helm chart and therefore cannot be automated by `helm install` right? Then it should be more convenient to keep using `local-path` for development?

tools/deployment/package-helm/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+  # Location where logs are stored. It will be created if it doesn't exist.
+  # NOTE: This directory must not overlap with any path used in CLP's execution container. An error
+  # will be raised if so.
+  logs_directory: "/tmp/clp/var/log"
+
+credentials:
+  # Database credentials
+  database:
+    username: "clp-user"
+    password: "pass"
```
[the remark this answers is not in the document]
junhaoliao: is it possible to check whether a non-default / non-empty value has been provided during `helm install`?

tools/deployment/package-helm/templates/_helpers.tpl
junhaoliao: in fact, init delay seems to be longer in k8s. i've tuned the values a bit

(i think we should also extend the timeout period in docker-compose to 4s, for avoiding false negatives on hosts with higher loads. if you agree, we can submit an issue and address separately

the k8s healthcheck mechanism is easier to be read so (as a writer) i feel inline comments are unnecessary. let me know if anything needs to be clarified and put into inline comments.

taskfiles/toolchains.yaml
```diff
@@ -15,6 +18,41 @@ vars:
...
+        #!/bin/sh
+        export PATH="{{.G_HELM_TOOLCHAIN_DIR}}:$PATH"
+        EOF
+        chmod +x "{{.G_HELM_TOOLCHAIN_ENV_FILE}}"
```
kirkrodrigues:
```suggestion
```

Since we only source the file, I don't think we need to make it executable, right?

chart_schema.yaml
```diff
@@ -0,0 +1,38 @@
...
+name: "str()"
```
kirkrodrigues: Hm, when I download the file, these lines are also modified to add quotes. Not the end of the world, but I think if we don't actually need to modify this file, maybe we should download it as part of the toolchain and then copy it here? We can also add it to gitignore and don't add it to the set of YAML files we lint.

ct.yaml
```diff
@@ -0,0 +1,8 @@
+chart-dirs:
+  - "tools/deployment"
+chart-yaml-schema: "chart_schema.yaml"
+check-version-increment: true
```
kirkrodrigues: This defaults to `true`, no? Our general guideline (added docs [here](https://www.notion.so/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?source=copy_link#2b604e4d9e6b801fb427cfe81c45b0d7)) is to avoid specifying values that are equivalent to the default.

ct.yaml
```diff
@@ -0,0 +1,8 @@
...
+lint-conf: ".yamllint.yaml"
+validate-chart-schema: true
+validate-maintainers: true
+validate-yaml: true
```
kirkrodrigues: These all default to `true`, no?

lint-requirements.txt
```diff
@@ -1,6 +1,12 @@
...
 clang-format>=21.1.2
 # Lock to v19.x until we can upgrade our code to fix new v20 issues.
 clang-tidy~=19.1
+
+# Python
+black>=24.4.2
 ruff>=0.4.4
+
+# YAML
+yamale>=6.0.0
```
kirkrodrigues:
```suggestion
yamale>=6.1.0
```

kirkrodrigues (commented):
Partial review.

ct.yaml
junhaoliao: Right. I mistakenly thought they were by default false when I was debugging some false positive issue. those options should be removed then

ct.yaml
junhaoliao: Right, i mistakenly thought they were false. corrected now

taskfiles/toolchains.yaml
junhaoliao: Right

chart_schema.yaml
junhaoliao: sure

tools/deployment/package-helm/templates/database-service.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: "v1"
+kind: "Service"
+metadata:
+  name: {{ include "clp.fullname" . }}-database
+  labels:
+    {{- include "clp.labels" . | nindent 4 }}
```
kirkrodrigues, line 6: Would `indent` work? If not, why do we need the prepended newline? Ditto elsewhere in the PR.

tools/deployment/package-helm/templates/database-statefulset.yaml
```diff
@@ -0,0 +1,110 @@
+apiVersion: "apps/v1"
+kind: "StatefulSet"
+metadata:
+  name: {{ include "clp.fullname" . }}-database
+  labels:
+    {{- include "clp.labels" . | nindent 4 }}
+    app.kubernetes.io/component: "database"
+spec:
+  serviceName: {{ include "clp.fullname" . }}-database
+  replicas: 1
```
kirkrodrigues, line 10: I think this defaults to 1, but I gu8ess you want to be explicit for clarity?

tools/deployment/package-helm/templates/database-statefulset.yaml
```diff
@@ -0,0 +1,110 @@
...
+            {{- include "clp.livenessProbeTimings" . | nindent 12 }}
+            exec:
+              <<: *database-health-check
+      volumes:
+        - name: "config"
+          configMap:
+            name: {{ include "clp.fullname" . }}-config
+  persistentVolumeClaimRetentionPolicy:
+    whenDeleted: "Delete"
+  volumeClaimTemplates:
```
kirkrodrigues, line 86: Forgive my ignorance, but why do the objects in this not have/need `labels` in `metadata`?

tools/deployment/package-helm/templates/database-data-pv.yaml
```diff
@@ -0,0 +1,8 @@
+{{- include "clp.createLocalPv" (dict
+  "root" .
```
kirkrodrigues, line 2: I asked AI about this and got a reasonable answer, but do you know where I can find the docs for what `.` does in this context?

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -0,0 +1,141 @@
...
+  persistentVolumeReclaimPolicy: "Retain"
+  storageClassName: "local-storage"
+  local:
+    path: {{ .hostPath }}
+  nodeAffinity:
+    required:
+      nodeSelectorTerms:
+        - matchExpressions:
+            - key: "kubernetes.io/hostname"
+              operator: "Exists"
```
kirkrodrigues, line 135: Can you explain this to me?

tools/deployment/package-helm/templates/db-table-creator-job.yaml
```diff
@@ -0,0 +1,63 @@
...
+        fsGroup: {{ .Values.securityContext.firstParty.gid }}
+      initContainers:
+        - name: "wait-for-database"
+          image: "busybox:latest"
+          command:
+            - "sh"
+            - "-c"
+            - >-
+              until nc -z {{ include "clp.fullname" . }}-database 3306;
+                do echo waiting for database;
```
kirkrodrigues:
```suggestion
                do echo Waiting for database to be ready...;
```

tools/deployment/package-helm/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: "v2"
+name: "clp"
+version: "0.1.0"
+description: "A Helm chart for CLP (Compressed Log Processor) package deployment"
```
kirkrodrigues:
```suggestion
description: "A Helm chart for CLP's (Compressed Log Processor) package deployment"
```

tools/deployment/package-helm/values.yaml
```diff
@@ -0,0 +1,47 @@
+nameOverride: ""
+fullnameOverride: ""
+
+# Whether to allow sbin script accesses to services on the node host.
```
kirkrodrigues:
```suggestion
# Whether to allow scripts in `sbin` access to services on the host.
```

tools/deployment/package-helm/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+allowSbinScripts: true
```
kirkrodrigues: This isn't a great name. How about `allowHostAccessForSbinScripts`?

tools/deployment/package-helm/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+    uid: 1000
+    gid: 1000
+  thirdParty:
+    uid: 1000
+    gid: 1000
+
+image:
+  clpPackage:
+    repository: "ghcr.io/y-scope/clp/clp-package"
+    pullPolicy: "IfNotPresent"
```
kirkrodrigues, line 18: Do we need this? I don't think we're using it currently, right?

tools/deployment/package-helm/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+clpConfig:
+  package:
+    storage_engine: "clp-s"
+    query_engine: "clp-s"
+
+  database:
+    type: "mariadb"  # "mariadb" or "mysql"
+    host: "localhost"
+    port: 30306
```
kirkrodrigues, line 29: Why 30306?

tools/deployment/package-helm/test.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# TODO: to be deleted / migrated to integration tests
```
kirkrodrigues:
```suggestion
#!/usr/bin/env bash

# TODO: Migrate into integration test

set -o errexit
set -o nounset
set -o pipefail
```

tools/deployment/package-helm/templates/database-data-pv.yaml
junhaoliao: this might be the best: https://pkg.go.dev/text/template#:~:text=Execution%20of%20the%20template%20walks%20the%20structure%20and%20sets%20the%20cursor%2C%20represented%20by%20a%20period%20%27.%27%20and%20called%20%22dot%22%2C%20to%20the%20value%20at%20the%20current%20location%20in%20the%20structure%20as%20execution%20proceeds.

the `.` represents the context of the template expansion

tools/deployment/package-helm/templates/database-service.yaml
junhaoliao: the `{{-` + `nident` pattern allows us to place the template anywhere in the line.

e.g., these are also valid
```yaml
  labels:
    app.kubernetes.io/component: "database"
```

```yaml
  labels:
    app.kubernetes.io/component: "database"
```

---

in comparison, if we use `{{` + `indent`, we can't really indent the template text. e.g., we have to write this:
```yaml
  labels:
{{ include "clp.labels" . | indent 4 }}
    app.kubernetes.io/component: "database"
```
the missing indent may look fine if the object isn't deeply nested. but when so, readability can be quite bad

tools/deployment/package-helm/templates/database-statefulset.yaml
junhaoliao: yes. people usually set up `StatefulSet`s / `Deployments` with multiple replicas

tools/deployment/package-helm/templates/database-statefulset.yaml
junhaoliao: there are different purposes of setting labels in K8s:

* StatefulSet metadata labels: Labels on the StatefulSet resource itself
  - Used by Helm for lifecycle management (upgrades, rollbacks)
  - Used for querying: `kubectl get statefulsets -l app.kubernetes.io/name=clp`

* Pod template labels: Labels on the Pods that will be created
  - REQUIRED - must match the StatefulSet's `selector.matchLabels`
  - Without these, the StatefulSet can't manage its pods

it wasn't necessary to add metadata labels to `VolumeClaimTemplate` metadata since:
1. PVCs don't need labels for functional reasons (unlike pods)
2. they're tightly coupled to the `StatefulSet` already

that said, adding labels would be help consistency/observability so we should add those consistently

tools/deployment/package-helm/templates/database-service.yaml
kirkrodrigues: Gotcha, gotcha.

tools/deployment/package-helm/templates/database-statefulset.yaml
junhaoliao: fwiw, labels from the `set.Spec.Selector.MatchLabels` do get copied into the volumeClaimTemplates created PVC:
https://github.com/kubernetes/kubernetes/blob/4bed36e03e7bd699b089d33da6f7d7c9ef9eb661/pkg/controller/statefulset/stateful_set_utils.go#L402

There was actually a bug in 2019 where custom labels get overridden by `set.Spec.Selector.MatchLabels`.

As explain by https://github.com/kubernetes/kubernetes/pull/74941/files#r263894283

> This code is intentional to allow the user of the statefulset to find the PVCs it creates.

I think it's fine to not specify any custom label for the volumeClaimTemplates metadata then.

---

there is no doc regarding the behaviour as the k8s maintainers admit:

> On review history, there was not comment related to this code. And the above explanation of this method(line 141-143) just mentions the name of claim, not the label.
I'm still not sure this code is intentional or not.. I hope the specialists can help this.

tools/deployment/package-helm/templates/_helpers.tpl
junhaoliao: https://kubernetes.io/docs/concepts/storage/volumes/#local:~:text=You%20must%20set%20a%20PersistentVolume%20nodeAffinity%20when%20using%20local%20volumes.

this was to get over the restriction but i agree this is too permissive. i think it's more proper to use https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-control-plane instead. e.g.,
```yaml
  nodeAffinity:
    required:
      nodeSelectorTerms:
        - matchExpressions:
            - key: "node-role.kubernetes.io/control-plane"
              operator: "Exists"
```

let me refactor this template function so other node-role types can be specified as well

tools/deployment/package-helm/values.yaml
junhaoliao: we should use it in db-table-creator-job.yaml , and the default should be updated as "Always".

when users want to debug a local image, they can import the image into the cluster, `--set` the `repository` as the project name, `tag` as the image tag, and `pullPolicy` as `IfNotPresent`

tools/deployment/package-helm/values.yaml
junhaoliao: https://kubernetes.io/docs/reference/networking/ports-and-protocols/#node

only ports in range 30000-32767	is allowed to be mapped. i removed the thousand-bit from the well-known 3306 port and add that to 30000

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -0,0 +1,165 @@
...
+  accessModes: {{ .accessModes }}
+  storageClassName: "local-storage"
+  selector:
+    matchLabels:
+      {{- include "clp.selectorLabels" .root | nindent 6 }}
+      app.kubernetes.io/component: {{ .component | quote }}
+  resources:
+    requests:
+      storage: {{ .capacity }}
+{{- end }}
```
kirkrodrigues: I guess this helper is for a future PR? Could we introduce it in that PR instead?

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -0,0 +1,165 @@
...
+    path: {{ .hostPath }}
+  nodeAffinity:
+    required:
+      nodeSelectorTerms:
+        - matchExpressions:
+            - key: {{ printf "node-role.kubernetes.io/%s" .nodeRole | quote }}
+              operator: "Exists"
+        - matchExpressions:
+            - key: "node-role.kubernetes.io/control-plane"
+              operator: "Exists"
```
kirkrodrigues, line 135: Why do the volume have to be on the control-plane node(s)? Or am I misunderstanding something?

ct.yaml
```diff
@@ -0,0 +1,4 @@
+chart-dirs:
+  - "tools/deployment"
+chart-yaml-schema: "build/toolchains/helm/etc/chart_schema.yaml"
```
kirkrodrigues: As discussed offline, can we pass this as a command-line argument that uses the taskfile variable `G_BUILD_DIR`?

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -0,0 +1,165 @@
...
+@return {string} The chart name (truncated to 63 characters)
+*/}}
+{{- define "clp.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Creates a default fully qualified app name. We truncate at 63 chars because
+some Kubernetes name fields are limited to this (by the DNS naming spec). If
+release name contains chart name it will be used as a full name.
```
kirkrodrigues:
```suggestion
Creates a default fully qualified app name. We truncate at 63 chars because some Kubernetes name
fields are limited to this (by the DNS naming spec). If release name contains chart name it will be
used as a full name.
```

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -0,0 +1,165 @@
...
+
+@return {string} YAML-formatted selector labels
+*/}}
+{{- define "clp.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "clp.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+Creates image reference for CLP Package.
```
kirkrodrigues:
```suggestion
Creates image reference for the CLP Package.
```

tools/deployment/package-helm/templates/_helpers.tpl
junhaoliao: Right. it should be used in Deployments (TBA) rather than StatefulSets. it should be removed then

tools/deployment/package-helm/templates/_helpers.tpl
junhaoliao: Here's the flow of how a Pod (StatefulSet / Job / Deployment) is scheduled on a node:

1. A PV is created with nodeAffinity that restricts which nodes can use it.
2. A PVC can then bind to the PV.
3. The Pod that uses the PVC is scheduled onto the same node.

This is to help users who can only set up a single-node cluster (when not using `kind`). They may only be able to set up a control-plane node, then:

0. Say we are calling this helper with "nodeRole=worker".
1. A PV specifies nodeAffinity for nodes with role "worker", but since no such node exists, it falls back to the only node with the control-plane role.
2. A PVC can then bind to the PV on this node.
3. The Pod that uses the PVC is scheduled onto the one-and-only control-plane node.

ct.yaml
junhaoliao: right, sorry for missing this in the last round

junhaoliao: when i created the PR, i initially tried to set up ct without the ct.yaml but this didn't work
```
ct lint \
  --chart-dirs="{{.ROOT_DIR}}/tools/deployment" \
  --chart-yaml-schema="{{.G_HELM_TOOLCHAIN_DIR}}/etc/chart_schema.yaml" \
  --lint-conf="{{.ROOT_DIR}}/.yamllint.yaml"
```
would yield
```
Linting charts...                                                                                                                                                                                                                  
Directory "tools/deployment/package-helm" is not a valid chart directory. Skipping...
Directory "tools/deployment/package-helm" is not a valid chart directory. Skipping...
```

---

with the belief that their arg parser should not process the config args differently than their yaml config parser, i finally figured out it's because they don't support absolute paths in in `--chart-dirs`. not sure if it's a bug...

so this works
```
ct lint \
  --chart-dirs="tools/deployment" \
  --chart-yaml-schema="{{.G_HELM_TOOLCHAIN_DIR}}/etc/chart_schema.yaml" \
  --lint-conf=".yamllint.yaml"
```

we can remove the `ct.yaml` altogether then

tools/deployment/package-helm/templates/_helpers.tpl
junhaoliao: sorry. was looking at the wrong ruler in my IDE

PR #1629, opened 2025-11-19
feat(log-ingestor): Add `S3Scanner` to scan a S3 bucket periodically to find and ingest newly created objects with a given prefix.

LinZhihao-723:
# Description

This PR  adds `S3Scanner` to periodically scan an S3 bucket to find and ingest newly created objects with a given prefix. As the implementation is using `list_object_v2` API with `start_after` marker, it requires the newly ingested objects' keys to preserve an increasing lexical order.

This PR also updates:

* `test_ingestion_job` to share code logic among SQS listener and S3 scanner.
* `s3::ObjectMetadata` to use `u64` to represent the object's size, as S3 APIs return this value as an `i64`. In practice it's possible that a file is larger than 4GB. We should use a 64-bit type to be future-proof.

# Checklist

  breaking change.

# Validation performed

* [x] Ensure new test cases for S3 scanner passed.

## Summary by CodeRabbit

* **New Features**
  * Added a periodic S3 prefix scanner with configurable interval, background execution and graceful shutdown.

* **Refactor**
  * Switched object size and buffer thresholds to a wider integer type (u64) to better handle large objects.
  * Introduced a public S3 client wrapper and relaxed client lifetime bounds to improve client management.

* **Tests**
  * Expanded tests for S3 scanning and ingestion flows, including noise-object scenarios and new test helpers.


components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -140,25 +178,110 @@ async fn test_sqs_listener() -> Result<()> {
...
+        NUM_TEST_OBJECTS,
+    ));
+
+    noise_upload_handle
+        .await
+        .expect("Error while awaiting noise upload");
+    let (mut created_objects, mut received_objects) = upload_and_receive_handle
+        .await
+        .expect("Error while awaiting upload and receive");
+    s3_scanner.shutdown_and_join().await?;
```
hoophalab, line 280: Sorry I didn't notice this last time. I guess even in test function, we should either use `anyhow::Context` or `Result::expect`. It makes little sense to mix unwrap/expect with Result in test. (We need to update those upload functions if we choose the Result route)
either 

```suggestion
    noise_upload_handle
        .await
        .context("Error while awaiting noise upload")?;
    let (mut created_objects, mut received_objects) = upload_and_receive_handle
        .await
        .context("Error while awaiting upload and receive")?;
    s3_scanner.shutdown_and_join().await?;
```

or
```suggestion
    noise_upload_handle
        .await
        .expect("Error while awaiting noise upload");
    let (mut created_objects, mut received_objects) = upload_and_receive_handle
        .await
        .expect("Error while awaiting upload and receive");
    s3_scanner.shutdown_and_join().await.unwrap();
```

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -67,27 +72,89 @@ async fn receive_object_metadata(
...
+    ));
+
+    let objects_received = tokio::time::timeout(
+        Duration::from_secs(RECEIVER_TIMEOUT_SEC),
+        receive_object_metadata(receiver, objects_to_create.len()),
+    )
+    .await
+    .expect("Timed out while receiving object metadata");
+
+    let () = creation_handler
```
hoophalab: Do I miss something?
```suggestion
    creation_handler
```

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -67,27 +72,89 @@ async fn receive_object_metadata(
...
+) {
+    let objects_to_create: Vec<_> = (0..num_objects_to_create)
+        .map(|_| ObjectMetadata {
+            bucket: bucket.clone(),
+            key: format!("{}.log", Uuid::new_v4()),
+            size: 16,
+        })
+        .collect();
+
+    let () = create_s3_objects(s3_client.clone(), objects_to_create)
```
hoophalab: Do I miss something?
```suggestion
    create_s3_objects(s3_client.clone(), objects_to_create)
```

components/log-ingestor/src/ingestion_job/s3_scanner.rs
```diff
@@ -0,0 +1,199 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`Self::scan`]'s return values on failure.
+    pub async fn run(mut self, cancel_token: CancellationToken) -> Result<()> {
+        select! {
+            // Cancellation requested.
+            () = cancel_token.cancelled() => {
```
hoophalab: critical: Coderabbit's comment is reasonable. `cancel` is only checked once. We need to do a `loop{ select! {...}}` rather than select-loop

components/log-ingestor/src/ingestion_job/s3_scanner.rs
```diff
@@ -0,0 +1,199 @@
...
+    ///   [`aws_sdk_s3::operation::list_objects_v2::builders::ListObjectsV2FluentBuilder::send`]'s
+    ///   return values on failure.
+    pub async fn scan(&mut self) -> Result<()> {
+        loop {
+            let client = self.s3_client_manager.get().await?;
+            let response = client
+                .list_objects_v2()
+                .bucket(self.config.bucket_name.as_str())
+                .prefix(self.config.prefix.as_str())
+                .set_start_after(self.config.start_after.clone())
```
hoophalab: critical: Am I misunderstanding something here? `start_after` means start after the alphabetical order, rather than the time of put.

If the user uploads `a.jsonl` after `z.jsonl` is ingested, log ingestor won't check `a.jsonl` again.

Apply the following diff manifests the bug in the unit test
```diff
diff --git a/components/log-ingestor/tests/test_ingestion_job.rs b/components/log-ingestor/tests/test_ingestion_job.rs
index 8add4dcd..78e38c37 100644
--- a/components/log-ingestor/tests/test_ingestion_job.rs
+++ b/components/log-ingestor/tests/test_ingestion_job.rs
@@ -95,6 +95,7 @@ async fn upload_and_receive(
             key: format!("{prefix}/{idx:05}.log"),
             size: 16,
         })
+        .rev()
         .collect();
```

hoophalab (changes requested):

components/log-ingestor/src/ingestion_job/s3_scanner.rs
LinZhihao-723: Yes, that's the expected behavior. The scanner now requires the input to be lexically increasing.

components/log-ingestor/src/ingestion_job/s3_scanner.rs
```diff
@@ -0,0 +1,199 @@
...
+/// # Type Parameters
+///
+/// * [`S3Client`]: The type of the AWS S3 client manager.
+struct Task<S3Client: AwsClientManagerType<Client>> {
+    s3_client_manager: S3Client,
+    config: S3ScannerConfig,
+    sender: mpsc::Sender<ObjectMetadata>,
+}
+
+impl<S3Client: AwsClientManagerType<Client> + 'static> Task<S3Client> {
```
hoophalab: It might be best to include 'static in the trait it self. As it is essentially an enum for type.

```diff
diff --git a/components/log-ingestor/src/aws_client_manager.rs b/components/log-ingestor/src/aws_client_manager.rs
index f962055c..da310b11 100644
--- a/components/log-ingestor/src/aws_client_manager.rs
+++ b/components/log-ingestor/src/aws_client_manager.rs
@@ -4,7 +4,7 @@ use aws_sdk_s3::Client as S3Client;
 use aws_sdk_sqs::Client as SqsClient;
 
 /// A marker trait for AWS client types.
-pub trait AwsClientType {}
+pub trait AwsClientType: Clone {}
 
 impl AwsClientType for SqsClient {}
 
@@ -17,7 +17,7 @@ impl AwsClientType for S3Client {}
 ///
 /// * [`Client`]: The AWS SKD client type. Must implement the [`AwsClientType`].
 #[async_trait]
-pub trait AwsClientManagerType<Client: AwsClientType + Clone>: Send + Sync {
+pub trait AwsClientManagerType<Client: AwsClientType>: Send + Sync + 'static {
     /// Retrieves an AWS client instance. The specific behavior depends on the implementation.
     ///
     /// # Returns:
diff --git a/components/log-ingestor/src/ingestion_job/sqs_listener.rs b/components/log-ingestor/src/ingestion_job/sqs_listener.rs
index e82ed112..84db7c28 100644
--- a/components/log-ingestor/src/ingestion_job/sqs_listener.rs
+++ b/components/log-ingestor/src/ingestion_job/sqs_listener.rs
@@ -32,7 +32,7 @@ struct Task<SqsClientManager: AwsClientManagerType<Client>> {
     sender: mpsc::Sender<ObjectMetadata>,
 }
 
-impl<SqsClientManager: AwsClientManagerType<Client> + 'static> Task<SqsClientManager> {
+impl<SqsClientManager: AwsClientManagerType<Client>> Task<SqsClientManager> {
     /// Runs the SQS listener task to listen to SQS messages and extract S3 object metadata. The
     /// extracted metadata is sent to the provided channel sender.
     ///
```

components/clp-rust-utils/src/s3.rs
```diff
@@ -6,5 +6,5 @@ pub use client::create_new_client;
 pub struct ObjectMetadata {
     pub bucket: String,
     pub key: String,
-    pub size: usize,
+    pub size: u64,
```
hoophalab, line 9: Maybeb we should keep using usize?  The size from aws sdk's object metadata is usize.

Did you change it to u64 because it has a fixed serialization size across the compilation target? I feel that should be an error raised by the deserializer, rather than by us using a bunch of `try_from` in the code.

```diff
diff --git a/components/clp-rust-utils/src/s3.rs b/components/clp-rust-utils/src/s3.rs
index 7d70d5e5..5d030c81 100644
--- a/components/clp-rust-utils/src/s3.rs
+++ b/components/clp-rust-utils/src/s3.rs
@@ -6,5 +6,5 @@ pub use client::create_new_client;
 pub struct ObjectMetadata {
     pub bucket: String,
     pub key: String,
-    pub size: u64,
+    pub size: usize,
 }
diff --git a/components/log-ingestor/src/compression/buffer.rs b/components/log-ingestor/src/compression/buffer.rs
index 45eca78d..1c4eef96 100644
--- a/components/log-ingestor/src/compression/buffer.rs
+++ b/components/log-ingestor/src/compression/buffer.rs
@@ -26,8 +26,8 @@ pub trait BufferSubmitter {
 pub struct Buffer<Submitter: BufferSubmitter> {
     submitter: Submitter,
     buf: Vec<ObjectMetadata>,
-    total_size: u64,
-    size_threshold: u64,
+    total_size: usize,
+    size_threshold: usize,
 }
 
 impl<Submitter: BufferSubmitter> Buffer<Submitter> {
@@ -36,7 +36,7 @@ impl<Submitter: BufferSubmitter> Buffer<Submitter> {
     /// # Returns
     ///
     /// A newly created [`Buffer`] with the given submitter of type `T` and size threshold.
-    pub const fn new(submitter: Submitter, size_threshold: u64) -> Self {
+    pub const fn new(submitter: Submitter, size_threshold: usize) -> Self {
         Self {
             submitter,
             buf: Vec::new(),
diff --git a/components/log-ingestor/src/ingestion_job/sqs_listener.rs b/components/log-ingestor/src/ingestion_job/sqs_listener.rs
index 0afdcce3..e82ed112 100644
--- a/components/log-ingestor/src/ingestion_job/sqs_listener.rs
+++ b/components/log-ingestor/src/ingestion_job/sqs_listener.rs
@@ -151,7 +151,7 @@ impl<SqsClientManager: AwsClientManagerType<Client> + 'static> Task<SqsClientMan
         Some(ObjectMetadata {
             bucket: record.s3.bucket.name,
             key: record.s3.object.key,
-            size: record.s3.object.size as u64,
+            size: record.s3.object.size,
         })
     }
 
diff --git a/components/log-ingestor/tests/test_compression_listener.rs b/components/log-ingestor/tests/test_compression_listener.rs
index 6337e66b..cc377701 100644
--- a/components/log-ingestor/tests/test_compression_listener.rs
+++ b/components/log-ingestor/tests/test_compression_listener.rs
@@ -6,7 +6,7 @@ use clp_rust_utils::s3::ObjectMetadata;
 use log_ingestor::compression::{Buffer, BufferSubmitter, DEFAULT_LISTENER_CAPACITY, Listener};
 use tokio::sync::{Mutex, mpsc};
 
-const TEST_OBJECT_SIZE: u64 = 1024;
+const TEST_OBJECT_SIZE: usize = 1024;
 
 /// A test submitter that stores submitted buffers in memory for inspection.
 struct TestBufferSubmitter {
diff --git a/components/log-ingestor/tests/test_ingestion_job.rs b/components/log-ingestor/tests/test_ingestion_job.rs
index 09013ee0..8add4dcd 100644
--- a/components/log-ingestor/tests/test_ingestion_job.rs
+++ b/components/log-ingestor/tests/test_ingestion_job.rs
@@ -34,7 +34,7 @@ async fn create_s3_objects(
     objects: Vec<ObjectMetadata>,
 ) -> Result<()> {
     for object in objects {
-        let body = vec![b'0'; usize::try_from(object.size).expect("size overflow")];
+        let body = vec![b'0'; object.size];
 
         s3_client
             .put_object()
```

LinZhihao-723: `usize` is platform-dependent: it can be a 4-byte integer. However, it is possible that the input type is larger than 4GB. Beyond that, the buffer will also need to accumulate all the sizes. I think it's worth to explicitly use a 64-bit unsigned int across all components.

hoophalab: but the size from s3 is usize. You cannot do anything (not even an error message) if the file is more than 4g and log ingestor is compiled to a 32bit target.

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -151,7 +151,7 @@ impl<SqsClientManager: AwsClientManagerType<Client> + 'static> Task<SqsClientMan
         Some(ObjectMetadata {
             bucket: record.s3.bucket.name,
             key: record.s3.object.key,
-            size: record.s3.object.size,
+            size: record.s3.object.size as u64,
```
hoophalab: If we've decided u64, we probably should replace the type of `record.s3.object.size` from usize to u64 in `Record`.

components/log-ingestor/tests/test_ingestion_job.rs
LinZhihao-723: Yeah make sense. My bad. I think we should forward the result then. I will use `context`.

components/log-ingestor/src/ingestion_job/s3_scanner.rs
LinZhihao-723: good catch. Sorry for missing Rabitt's comment on this.

components/log-ingestor/tests/test_ingestion_job.rs
LinZhihao-723: Yeah u're good. I first wrote it with only one `except` and the compiler alerts me to add sth to receive the return value. But later I realize the returned value is also a result and there should be a second expect.

PR #1610, opened 2025-11-16
refactor(clp-package)!: Update `user` in credentials config file for consistency (resolves #1609):

junhaoliao:
- Rename `user` to `username`.
- Update default user to `clp-user` in template credentials.

# Description

This PR introduces a `BREAKING` change: `database` and `redis` usernames must now be configured under the `username` fields in `etc/credentials.yml` within the CLP package. e.g.,

`etc/credentials.yml`:
```yaml
# Database credentials
database:
  # OLD
  # user: "user"
  # NEW
  username: "clp-user"
  ...

# Queue credentials
queue:
  # OLD
  # user: "user"
  # NEW
  username: "clp-user"
  ...

...
```

---

The PR fixes inconsistency where credential key names and default values differed between the template file, Pydantic config objects, and auto-generated credentials. All components now consistently use `username` as the key name with `clp-user` as the default value.

Changes:
  - Updated `credentials.template.yml` to show `username: "clp-user"` (instead of `user: "user"`)
  - Modified `clp_config.py` to read `database.username` and `queue.username` from YAML
  - Updated `general.py` to generate credentials with `username` keys and `clp-user` as default
  - Fixed Presto deployment script to read `database.username` and renamed variable to `database_username`

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

- Verified all credential loading code now reads `username` keys
- Confirmed template default values (`clp-user`) match auto-generated credentials
- Checked documentation (already showed correct `username: "clp-user"` format)
- Searched codebase for remaining references to old `user` key format

```
task
cd build/clp-package
./sbin/start-clp.sh
./sbin/compress.sh ~/samples/postgresql.jsonl
# observed no error
```

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Release Notes

* **Bug Fixes**
  * Updated credential key naming from "user" to "username" for database and queue components to ensure consistent credential handling across the system. Configuration templates and credential loading have been updated accordingly.

PR #1633, opened 2025-11-19
feat(clp-package): Add `clp_aws_credentials` table to CLP metadata database and initialize privileged user.

Eden-D-Zhang:
# Description

This PR adds initialization for the `clp_aws_credentials` table and the privileged user that interacts with the new database table.

Changes
- Added credential table schema in CLP metadata DB creation.
- Added `initialize-privileged-db-user.py` script to create new DB user with privileged access and revoke credential table access from default user.

# Checklist

  breaking change.

# Validation performed

- Started and used the CLP package

---
  - https://app.asana.com/0/0/1212888487228208

### DIFF
--- a/components/clp-package-utils/clp_package_utils/controller.py
+++ b/components/clp-package-utils/clp_package_utils/controller.py
@@ -13,9 +13,9 @@ from typing import Any
 from clp_py_utils.clp_config import (
     API_SERVER_COMPONENT_NAME,
     AwsAuthType,
+    BundledService,
     CLP_DB_ADMIN_PASS_ENV_VAR_NAME,
     CLP_DB_ADMIN_USER_ENV_VAR_NAME,
-    BundledService,
     CLP_DB_PASS_ENV_VAR_NAME,
     CLP_DB_ROOT_PASS_ENV_VAR_NAME,
     CLP_DB_ROOT_USER_ENV_VAR_NAME,
--- a/components/clp-py-utils/clp_py_utils/initialize-admin-db-user.py
+++ b/components/clp-py-utils/clp_py_utils/initialize-admin-db-user.py
@@ -58,9 +58,10 @@ def main(argv):
         escaped_admin_password = admin_password.replace("'", "''")
 
         credentials_table = get_aws_credentials_table_name(table_prefix)
-        with closing(sql_adapter.create_connection(True, ClpDbUserType.ROOT)) as db_conn, closing(
-            db_conn.cursor(dictionary=True)
-        ) as cursor:
+        with (
+            closing(sql_adapter.create_connection(True, ClpDbUserType.ROOT)) as db_conn,
+            closing(db_conn.cursor(dictionary=True)) as cursor,
+        ):
             logger.info(f"Creating admin database user '{admin_username}'...")
 
             cursor.execute(


tools/deployment/package/docker-compose-all.yaml
```diff
@@ -82,6 +82,8 @@ services:
     hostname: "db_table_creator"
     environment:
       CLP_DB_PASS: "${CLP_DB_PASS:?Please set a value.}"
+      CLP_DB_PRIVILEGED_PASS: "${CLP_DB_PRIVILEGED_PASS:-}"
```
junhaoliao: shall we use the CLP_DB_PRIVILEGED_PASS for MYSQL_ROOT_PASSWORD?

components/package-template/src/etc/credentials.template.yml
```diff
@@ -2,6 +2,8 @@
 #database:
 #  username: "clp-user"
 #  password: "pass"
+#  privileged_user: "admin-user"
```
junhaoliao: how about `clp-admin`?

Eden-D-Zhang: That works, but I was also wondering if we want to keep the `privileged_` prefix, or use `credential_manager_` instead given that's what's in the doc. Most of the changes here are from before the new design, so there might be a few small differences.

components/clp-py-utils/clp_py_utils/clp_metadata_db_utils.py
```diff
@@ -110,6 +112,41 @@ def _create_column_metadata_table(db_cursor, table_prefix: str, dataset: str) ->
...
+            `last_used_at` DATETIME,
+            PRIMARY KEY (`id`) USING BTREE,
+            INDEX `idx_name` (`name`),
+            INDEX `idx_credential_type` (`credential_type`),
+            INDEX `idx_transient_created` (`transient`, `created_at`),
+            INDEX `idx_expiration` (`expiration`),
+            INDEX `idx_source_credential_id` (`source_credential_id`),
+            FOREIGN KEY (`source_credential_id`)
+                REFERENCES `{aws_credentials_table_name}`(`id`)
+                ON DELETE CASCADE
```
sitaowang1998, line 140: I am a little confused by this foreign key. Correct me if I'm wrong, but it is referencing the `id` from the same table. Is this intended behaviour?

Eden-D-Zhang: It references the `id` of a different entry in the same table. Is there a better way to do this?

sitaowang1998: I think the best way is to separate into two tables. One table specifically for storing user info, e.g. id, name. The other table stores the aws credentials. the `source_credential_id` in aws credential table should reference user id in user table. @junhaoliao What's your opinion?
Apart from separate table, it is a bad idea to use auto increment key to serve as a unique id. Auto increment key could be reused after deletion and database restart. In general, uuid is a better choice. MariaDB has native support for uuid data type and functions https://mariadb.com/docs/server/reference/sql-functions/secondary-functions/miscellaneous-functions/uuid_v7.

Eden-D-Zhang: I don't think that would really solve anything since generated entries would still need to reference the long-term keys they were created from; would there be multiple credentials referencing the same user id in the other table?

We considered separating temporary credentials into a separate 'cache' table, but I think Junhao decided against it. Can you explain why we should avoid self-referencing?

sitaowang1998: The primary concern is readability. Even after your first explanation, I still don't get it. If we want to use self-referencing foreign key, we should add comments on it.

junhaoliao: if i'm not wrong, MySQL does not recycle AUTO_INCREMENT IDs after deletion:

> The current maximum auto-increment counter value is written to the redo log each time it changes and saved to the data dictionary on each checkpoint; this makes the current maximum auto-increment counter value persistent across server restarts.

...
> The current maximum auto-increment value is persisted, preventing the reuse of previously allocated values.

https://dev.mysql.com/doc/refman/8.4/en/innodb-auto-increment-handling.html

junhaoliao: i think a single table with a self-referencing FK is easier to maintain if the behaviours of all credential types are similar. as in:

- all credential types share a common set of fields (name, type, expiration, created/updated timestamps, description, etc.), so keeping them in one table makes the system simpler and easier to maintain
- a single table also avoids the overhead of maintaining separate schemas, duplicate indexes, or multiple JOINs when querying credentials. Having them in one place reduces code complexity and ensures consistent handling

that adding the foreign key is just to register a trigger to delete generated temporary credentials when "source" long-term credential is removed. the foreign key isn't and shouldn't be used for anything else (@Eden-D-Zhang correct me if i'm wrong)

i think the current design is fine, unless there's an easy way to "extend" a table, which voids all above supporting points of maintaining a single table?

Eden-D-Zhang: Actually I believe the foreign key would be used to retrieve a temporary credential given a long-term credential.

i.e:
- User uses cli/webui to add long-term keys to DB
- User selects said credentials to ingest from S3
- Credential manager queries DB by the foreign key (`source_credential_id`) in case valid temporary credentials have previously been generated

Essentially the foreign key also serves to avoid making STS calls when a valid temporary credential already exists.

components/package-template/src/etc/credentials.template.yml
junhaoliao: since this is interface related, @kirkrodrigues can we get your insights?

junhaoliao: we tried renaming `user` to `username` last week to resolve some inconsistencies in the previous code. wherever applicable, let's rename all `user` references to `username` in this PR for consistency with the rest of the code

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
```diff
@@ -0,0 +1,101 @@
...
+
+    if not clp_config.database.has_privileged_credentials():
+        logger.error(
+            "Privileged database credentials not configured."
+            " Please set CLP_DB_PRIVILEGED_USER and CLP_DB_PRIVILEGED_PASS."
+        )
+        return -1
+
+    admin_db_config = clp_config.database.model_copy()
+    admin_db_config.username = "root"
```
junhaoliao: (related: https://github.com/y-scope/clp/pull/1633/files#r2546691294 )

@kirkrodrigues i think we previously configured the root user's password to be the same user as the clp-user for convenience. Is it time to explicitly expect a root user password in credentials.yaml?

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -11,6 +12,7 @@
     model_validator,
     PlainSerializer,
     PrivateAttr,
+    SecretStr,
```
junhaoliao: this seems unused? did we attempt using this type for passwords? (and why do we end up not using it?)

components/clp-py-utils/clp_py_utils/initialize-clp-metadata-db.py
```diff
@@ -61,6 +62,8 @@ def main(argv):
         with closing(sql_adapter.create_connection(True)) as metadata_db, closing(
             metadata_db.cursor(dictionary=True)
         ) as metadata_db_cursor:
+            create_aws_credentials_table(metadata_db_cursor, table_prefix)
+
             if StorageEngine.CLP_S == storage_engine:
                 create_datasets_table(metadata_db_cursor, table_prefix)
```
junhaoliao: since clp-text (`StorageEngine.CLP`) is missing s3 support - 

let's call this only when the storage engine is clp-s

```suggestion
            if StorageEngine.CLP_S == storage_engine:
                create_aws_credentials_table(metadata_db_cursor, table_prefix)
                create_datasets_table(metadata_db_cursor, table_prefix)
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -453,7 +453,12 @@ def load_config_file(
 
 def generate_credentials_file(credentials_file_path: pathlib.Path):
     credentials = {
-        DB_COMPONENT_NAME: {"username": "clp-user", "password": secrets.token_urlsafe(8)},
+        DB_COMPONENT_NAME: {
+            "username": "clp-user",
+            "password": secrets.token_urlsafe(8),
+            "privileged_username": "clp-credential-admin",
+            "privileged_password": secrets.token_urlsafe(16),
```
junhaoliao: is the inconsistency intentional? i.e., 8 vs 16

Eden-D-Zhang: I was thinking having it longer would be more secure, happy to change it.

components/clp-py-utils/clp_py_utils/clp_config.py
Eden-D-Zhang: Oh this was leftover from when everything was supposed to be in Python. It was meant for the aws credentials.

components/clp-package-utils/clp_package_utils/general.py
junhaoliao: i see your points. for now, let's use 8 for consistency in the database passwords. as in, we shouldn't hint the clp-user is less secure than the admin password.

(@gibber9809 do you remember why we picked 16 for the Redis password generation? I checked [the Redis requirements](https://redis.io/docs/latest/operate/rs/security/access-control/manage-passwords/password-complexity-rules/) and it seems only minimum 8 characters are required

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -143,6 +143,17 @@ def _set_up_env_for_database(self) -> EnvVarsDict:
...
+        if self._clp_config.database.has_root_password():
+            env_vars |= {
+                "CLP_DB_ROOT_PASS": self._clp_config.database.root_password,
+            }
+
+        if self._clp_config.database.has_privileged_credentials():
+            env_vars |= {
+                "CLP_DB_PRIVILEGED_PASS": self._clp_config.database.privileged_password,
+                "CLP_DB_PRIVILEGED_USER": self._clp_config.database.privileged_username,
+            }
```
junhaoliao: I believe we can expect all of them to be available? like, even if the user didn't set any of those, we would have generated the credentials for them at this point?

```
        env_vars |= {
            "CLP_DB_ROOT_PASS": self._clp_config.database.root_password,

            "CLP_DB_ADMIN_PASS": self._clp_config.database.admin_password,
            "CLP_DB_ADMIN_USER": self._clp_config.database.admin_username,

            "CLP_DB_PASS": self._clp_config.database.password,
            "CLP_DB_USER": self._clp_config.database.username,
        }
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -453,7 +453,13 @@ def load_config_file(
...
```
junhaoliao: note the length is also updated as 8 as suggested at https://github.com/y-scope/clp/pull/1633/files#r2558412419
```suggestion
            "admin_username": "clp-admin",
            "admin_password": secrets.token_urlsafe(8),
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -453,7 +453,13 @@ def load_config_file(
...
+            "root_password": secrets.token_urlsafe(16),
+        },
```
junhaoliao, line 461: let's alphabetize the items

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: i see. if it's unused, let's remove it then. i think `task lint:fix-py` could have helped fixing this too

```suggestion
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -1,5 +1,6 @@
 import os
 import pathlib
+from datetime import datetime, timedelta, timezone
```
junhaoliao: this seems unused too. `task lint:fix-py` might help catching / fixing this

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -71,6 +73,9 @@
 # Environment variable names
 CLP_DB_USER_ENV_VAR_NAME = "CLP_DB_USER"
 CLP_DB_PASS_ENV_VAR_NAME = "CLP_DB_PASS"
+CLP_DB_PRIVILEGED_USER_ENV_VAR_NAME = "CLP_DB_PRIVILEGED_USER"
+CLP_DB_PRIVILEGED_PASS_ENV_VAR_NAME = "CLP_DB_PRIVILEGED_PASS"
+CLP_DB_ROOT_PASS_ENV_VAR_NAME = "CLP_DB_ROOT_PASS"
```
```suggestion
CLP_DB_ROOT_PASS_ENV_VAR_NAME = "CLP_DB_ROOT_PASS"
CLP_DB_ADMIN_USER_ENV_VAR_NAME = "CLP_DB_ADMIN_USER"
CLP_DB_ADMIN_PASS_ENV_VAR_NAME = "CLP_DB_ADMIN_PASS"
CLP_DB_USER_ENV_VAR_NAME = "CLP_DB_USER"
CLP_DB_PASS_ENV_VAR_NAME = "CLP_DB_PASS"
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -224,9 +234,46 @@ def get_clp_connection_params_and_type(self, disable_localhost_socket_connection
         return connection_params_and_type
 
     def dump_to_primitive_dict(self):
-        d = self.model_dump(exclude={"username", "password"})
+        d = self.model_dump(
+            exclude={"username", "password", "privileged_username", "privileged_password"}
```
junhaoliao: note the `root_password` should be excluded too
```suggestion
            exclude={"root_password", "admin_username", "admin_password", "username", "password"}
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -224,9 +234,46 @@ def get_clp_connection_params_and_type(self, disable_localhost_socket_connection
...
+        """
+        return self.root_password is not None
+
+    def has_privileged_credentials(self) -> bool:
+        """
+        Checks if privileged credentials are configured.
+
+        :return: True if both privileged username and password are set.
+        """
+        return self.privileged_username is not None and self.privileged_password is not None
```
junhaoliao: i can be wrong but this may not be valid code paths due to password generations
```suggestion
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -224,9 +234,46 @@ def get_clp_connection_params_and_type(self, disable_localhost_socket_connection
...
+
+    def get_privileged_connection_params(self, disable_localhost_socket_connection: bool = False):
```
```suggestion
    def get_admin_connection_params(self, disable_localhost_socket_connection: bool = False):
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -224,9 +234,46 @@ def get_clp_connection_params_and_type(self, disable_localhost_socket_connection
...
+
+        :param disable_localhost_socket_connection:
+        :return:
+        :raises ValueError: If privileged credentials are not configured.
+        """
+        if not self.has_privileged_credentials():
+            raise ValueError(
+                "Privileged database credentials not configured."
+                " Set privileged_username and privileged_password in database configuration."
+            )
```
junhaoliao: can we extract a `ensure_admin_credentials_loaded` method for this? (should be similar to similar to `ensure_credentials_loaded`

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -224,9 +234,46 @@ def get_clp_connection_params_and_type(self, disable_localhost_socket_connection
...
+        """
+        Checks if privileged credentials are configured.
+
+        :return: True if both privileged username and password are set.
+        """
+        return self.privileged_username is not None and self.privileged_password is not None
+
+    def get_privileged_connection_params(self, disable_localhost_socket_connection: bool = False):
+        """
+        Gets MySQL connection parameters using privileged credentials.
```
```suggestion
        Gets MySQL connection parameters for the CLP admin user.
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -224,9 +234,46 @@ def get_clp_connection_params_and_type(self, disable_localhost_socket_connection
...
+        :raises ValueError: If privileged credentials are not configured.
+        """
+        if not self.has_privileged_credentials():
+            raise ValueError(
+                "Privileged database credentials not configured."
+                " Set privileged_username and privileged_password in database configuration."
+            )
+
+        connection_params = self.get_mysql_connection_params(disable_localhost_socket_connection)
+        connection_params["user"] = self.privileged_username
```
junhaoliao: this should also be renamed as `username` but was missed in #1610

no action is required in the current PR

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -239,13 +286,41 @@ def load_credentials_from_file(self, credentials_file_path: pathlib.Path):
...
+        try:
+            self.root_password = _get_env_var(CLP_DB_ROOT_PASS_ENV_VAR_NAME)
+        except ValueError:
+            pass
+
+        try:
+            self.privileged_username = _get_env_var(CLP_DB_PRIVILEGED_USER_ENV_VAR_NAME)
+            self.privileged_password = _get_env_var(CLP_DB_PRIVILEGED_PASS_ENV_VAR_NAME)
+        except ValueError:
+            pass
```
junhaoliao: instead of this, can we write:
```
    def load_credentials_from_env(self):
        """
        Loads database credentials for the CLP user from environment variables.
        :raise ValueError: If any expected environment variable is not set.
        """
        self.username = _get_env_var(CLP_DB_USER_ENV_VAR_NAME)
        self.password = _get_env_var(CLP_DB_PASS_ENV_VAR_NAME)

    def load_admin_credentials_from_env(self):
        """
        Loads database credentials for the CLP admin user from environment variables.

        :raise ValueError: If any expected environment variable is not set.
        """
        self.admin_username = _get_env_var(CLP_DB_ADMIN_USER_ENV_VAR_NAME)
        self.admin_password = _get_env_var(CLP_DB_ADMIN_PASS_ENV_VAR_NAME)
```

components/clp-py-utils/clp_py_utils/clp_metadata_db_utils.py
```diff
@@ -21,6 +22,7 @@
 TABLE_SUFFIX_MAX_LEN = max(
     len(ARCHIVE_TAGS_TABLE_SUFFIX),
     len(ARCHIVES_TABLE_SUFFIX),
+    len(CREDENTIAL_MANAGER_TABLE_SUFFIX),
     len(COLUMN_METADATA_TABLE_SUFFIX),
```
junhaoliao:
```suggestion
    len(COLUMN_METADATA_TABLE_SUFFIX),
    len(CREDENTIAL_MANAGER_TABLE_SUFFIX),
```

components/clp-py-utils/clp_py_utils/create-db-tables.py
```diff
@@ -51,6 +51,14 @@ def main(argv):
     # fmt: on
     subprocess.run(cmd, check=True)
 
+    # fmt: off
+    cmd = [
+        "python3", str(script_dir / "initialize-privileged-db-user.py"),
```
junhaoliao:
```suggestion
        "python3", str(script_dir / "initialize-admin-db-user.py"),
```

components/clp-py-utils/clp_py_utils/initialize-admin-db-user.py
junhaoliao, line 1: how about renaming this as `initialize-admin-db-user.py`

components/package-template/src/etc/credentials.template.yml
```diff
@@ -2,6 +2,9 @@
 #database:
 #  username: "clp-user"
 #  password: "pass"
+#  privileged_username: "clp-credentials-user"
+#  privileged_password: "admin-pass"
+#  root_password: "root-pass"
```
junhaoliao, line 8:
```suggestion
#  root_password: "root-pass"
#  admin_username: "admin-user"
#  admin_password: "admin-pass"
#  username: "clp-user"
#  password: "pass"
```

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
```diff
@@ -0,0 +1,102 @@
...
+    args_parser = argparse.ArgumentParser(
+        description="Creates privileged database user for credential management."
+    )
+    args_parser.add_argument("--config", "-c", required=True, help="CLP configuration file.")
+    parsed_args = args_parser.parse_args(argv[1:])
+
+    config_path = pathlib.Path(parsed_args.config)
+    try:
+        clp_config = ClpConfig.model_validate(read_yaml_config_file(config_path))
+        clp_config.database.load_credentials_from_env()
```
junhaoliao: instead, we can write write this as
```suggestion
        clp_config.database.load_credentials_from_env()
        clp_config.database.load_admin_credentials_from_env()
```

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
```diff
@@ -0,0 +1,102 @@
...
+    except Exception:
+        logger.exception("Failed to load CLP configuration.")
+        return -1
+
+    if not clp_config.database.has_privileged_credentials():
+        logger.error(
+            "Privileged database credentials not configured."
+            " Please set CLP_DB_PRIVILEGED_USER and CLP_DB_PRIVILEGED_PASS."
+        )
+        return -1
```
junhaoliao: then we don't have to check this any more
```suggestion
```

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python3
+import argparse
+import logging
+import pathlib
+import sys
+from contextlib import closing
+
+from pydantic import ValidationError
+from sql_adapter import SqlAdapter
```
junhaoliao: 1. avoid relative imports (if you're interested, see [PEP8 - Imports](https://peps.python.org/pep-0008/#imports) for the rationale)
2. even when relative imports are permitted, avoid missing `.` before the relative module name. e.g., we could write `from .sql_adapter import SqlAdapter` if we ever allow relative imports. without the `.`, readers might be even more confused, thinking the import is a 3rd party module. (i understand this code might be copied from some other files. the issues in those files should be fixed by #1640 soon)

```suggestion
from clp_py_utils.sql_adapter import SqlAdapter
```

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao, line 1: while we don't want to over engineer things, for cleaner integration with SqlAdapter, i think it's better to change the interfaces as follows to support multiple users in the DB config:
```py
class ClpDbUserType(KebabCaseStrEnum):
    ADMIN = auto()
    CLP = auto()
    ROOT = auto()

class DbUserCredentials(BaseModel):
    username: NonEmptyStr = None
    password: NonEmptyStr = None

class Database(BaseModel):
    # ...
    users: Dict[str, DbUserCredentials]
    # ...

    def ensure_credentials_loaded(self, user_type: ClpDbUserType):
        pass

    def get_mysql_connection_params(self, user_type: ClpDbUserType, disable_localhost_socket_connection: bool = False):
        pass

    def get_clp_connection_params_and_type(self, user_type: ClpDbUserType, disable_localhost_socket_connection: bool = False):
        """
        NOTE: i really don't like how the code in this method is repeated from `get_mysql_connection_params()`. i think it's possible to write something like:
        return {
            **self.get_mysql_connection_params(user_type, disable_localhost_socket_connection),

            NOTE: clp-core does not distinguish between mysql and mariadb
            "type": DatabaseEngine.MYSQL.value,
            "table_prefix": CLP_METADATA_TABLE_PREFIX,
        }
        """
        pass

        d = self.model_dump(exclude={"users"})
        return d

    def load_credentials_from_file(self, credentials_file_path: pathlib.Path):
        """
        NOTE: In this method we should load for all credential types. Throw if any is missing.
        """
        pass

    def load_credentials_from_env(self, user_type: ClpDbUserType):
        """
        NOTE: no matter what user type is requested in this method, we should always throw, so that callers of load_credentials_from_file() don't need to check again with methods like `has_root_credentials()` and throw.
        """

    def transform_for_container(self):
        self.host = DB_COMPONENT_NAME
        self.port = self.DEFAULT_PORT


class SqlAdapter:
    ...

    def create_connection(self, user_type: ClpDbUserType, disable_localhost_socket_connection: bool = False):
        if "mysql" == self.database_config.type:
            return self._create_mysql_connection(user_type, disable_localhost_socket_connection)
        if "mariadb" == self.database_config.type:
            return self._create_mariadb_connection(user_type, disable_localhost_socket_connection)
        raise NotImplementedError

    def create_connection_pool(
        self,
        logger: logging.Logger,
        user_type: ClpDbUserType,
        pool_size: int,
        disable_localhost_socket_connection: bool = False,
    ):
        def create_connection():
            return self.create_connection(user_type, disable_localhost_socket_connection)

        ...

    def _create_mariadb_connection(
        self, user_type: ClpDbUserType, disable_localhost_socket_connection: bool = False
    ) -> mariadb.connection:
        """
        NOTE: This is changed as as private method now, so we are also moving this method towards the end of the class.
        """
        try:
            connection = mariadb.connect(
                **self.database_config.get_mysql_connection_params(
                    user_type,
                    disable_localhost_socket_connection
                )
            )
        except mariadb.Error as err:
            ...
        else:
            return connection

    def _create_mysql_connection(
        self, user_type: ClpDbUserType, disable_localhost_socket_connection: bool = False
    ) -> mysql.connector.MySQLConnection:
        """
        NOTE: This is changed as as private method now, so we are also moving this method towards the end of the class.
        """
        try:
            connection = mysql.connector.connect(
                **self.database_config.get_mysql_connection_params(
                    user_type, 
                    disable_localhost_socket_connection
                )
            )
        except mysql.connector.Error as err:
            ...
        else:
            return connection
```

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -82,6 +82,9 @@ services:
...
+      CLP_DB_PRIVILEGED_USER: "${CLP_DB_PRIVILEGED_USER:-}"
+      CLP_DB_ROOT_PASS: "${CLP_DB_ROOT_PASS:?Please set a value.}"
       CLP_DB_USER: "${CLP_DB_USER:?Please set a value.}"
```
junhaoliao, line 92:
```suggestion
      CLP_DB_ADMIN_PASS: "${CLP_DB_ADMIN_PASS:?Please set a value.}"
      CLP_DB_ADMIN_USER: "${CLP_DB_ADMIN_USER:?Please set a value.}"
      CLP_DB_PASS: "${CLP_DB_PASS:?Please set a value.}"
      CLP_DB_ROOT_PASS: "${CLP_DB_ROOT_PASS:?Please set a value.}"
      CLP_DB_USER: "${CLP_DB_USER:?Please set a value.}"
```

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
```diff
@@ -0,0 +1,102 @@
...
+    admin_db_config.password = admin_db_config.root_password
+
+    try:
+        sql_adapter = SqlAdapter(admin_db_config)
+        clp_db_connection_params = clp_config.database.get_clp_connection_params_and_type(True)
+        database_name = clp_db_connection_params["name"]
+        table_prefix = clp_db_connection_params.get("table_prefix", CLP_METADATA_TABLE_PREFIX)
+
+        privileged_username = clp_config.database.privileged_username
+        privileged_password = clp_config.database.privileged_password
```
junhaoliao:
```suggestion
        admin_username = clp_config.database.users[ClpDbUserType.ADMIN].username
        admin_password = clp_config.database.users[ClpDbUserType.ADMIN].password
```

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
```diff
@@ -0,0 +1,102 @@
...
+        regular_username = clp_config.database.username
```
junhaoliao:
```suggestion
        regular_username = clp_config.database.users[ClpDbUserType.CLP].username
```

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
junhaoliao: with the above refactoring proposal
```suggestion
        clp_config.database.load_credentials_from_env(ClpDbUserType.ROOT)
        clp_config.database.load_credentials_from_env(ClpDbUserType.ADMIN)
        clp_config.database.load_credentials_from_env(ClpDbUserType.CLP)
```

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
junhaoliao:
```suggestion
```
if we throw when loading from env fails, we don't need this

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
```diff
@@ -0,0 +1,102 @@
...
+    if not clp_config.database.has_privileged_credentials():
+        logger.error(
+            "Privileged database credentials not configured."
+            " Please set CLP_DB_PRIVILEGED_USER and CLP_DB_PRIVILEGED_PASS."
+        )
+        return -1
+
+    admin_db_config = clp_config.database.model_copy()
+    admin_db_config.username = "root"
+    admin_db_config.password = admin_db_config.root_password
```
junhaoliao: if we implement the above refactoring proposal, we don't need this
```suggestion
```

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: if you see anyone is calling `.create_mysql_connection()` in the current code, it's a mistake and should be corrected as `create_connection()` instead

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
```diff
@@ -0,0 +1,102 @@
...
+        database_name = clp_db_connection_params["name"]
+        table_prefix = clp_db_connection_params.get("table_prefix", CLP_METADATA_TABLE_PREFIX)
+
+        privileged_username = clp_config.database.privileged_username
+        privileged_password = clp_config.database.privileged_password
+        regular_username = clp_config.database.username
+
+        if privileged_username is None or privileged_password is None or regular_username is None:
+            logger.error("Database credentials unexpectedly missing during privileged user setup.")
+            return -1
```
junhaoliao:
```suggestion
```

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
```diff
@@ -0,0 +1,102 @@
...
+
+        escaped_privileged_username = privileged_username.replace("'", "''")
+        escaped_privileged_password = privileged_password.replace("'", "''")
```
junhaoliao:
```suggestion
        escaped_admin_username = privileged_username.replace("'", "''")
        escaped_admin_password = privileged_password.replace("'", "''")
```

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
```diff
@@ -0,0 +1,102 @@
...
+
+        credentials_table = get_aws_credentials_table_name(table_prefix)
+        with closing(sql_adapter.create_connection(True)) as db_conn, closing(
+            db_conn.cursor(dictionary=True)
+        ) as cursor:
+            logger.info(f"Creating privileged database user '{privileged_username}'...")
```
junhaoliao:
```suggestion
            logger.info(f"Creating admin database user '{admin_username}'...")
```

components/clp-py-utils/clp_py_utils/initialize-admin-db-user.py
```diff
@@ -0,0 +1,102 @@
...
+        credentials_table = get_aws_credentials_table_name(table_prefix)
+        with closing(sql_adapter.create_connection(True)) as db_conn, closing(
+            db_conn.cursor(dictionary=True)
+        ) as cursor:
+            logger.info(f"Creating privileged database user '{privileged_username}'...")
+
+            cursor.execute(
+                f"CREATE USER IF NOT EXISTS '{escaped_privileged_username}'@'%'"
+                f" IDENTIFIED BY '{escaped_privileged_password}'"
+            )
```
junhaoliao, line 70:
```suggestion
            cursor.execute(
                f"CREATE USER IF NOT EXISTS '{escaped_admin_username}'@'%'"
                f" IDENTIFIED BY '{escaped_admin_password}'"
            )
```

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
```diff
@@ -0,0 +1,102 @@
...
+
+            cursor.execute(
+                f"CREATE USER IF NOT EXISTS '{escaped_privileged_username}'@'%'"
+                f" IDENTIFIED BY '{escaped_privileged_password}'"
+            )
+
+            logger.info(f"Granting privileges on credential tables to '{privileged_username}'...")
```
junhaoliao:
```suggestion
            logger.info(
                f"Granting privileges on credentials tables to admin user '{admin_username}'."
            )
```

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
```diff
@@ -0,0 +1,102 @@
...
+
+            cursor.execute(
+                f"GRANT SELECT, INSERT, UPDATE, DELETE ON `{database_name}`.`{credentials_table}`"
+                f" TO '{escaped_privileged_username}'@'%'"
+            )
+            cursor.execute("FLUSH PRIVILEGES")
+            db_conn.commit()
+
+            logger.info("Privileged database user setup completed successfully.")
```
junhaoliao:
```suggestion
            logger.info("Admin database user '%s' created and permissions applied successfully.", admin_username)
```

components/clp-py-utils/clp_py_utils/initialize-privileged-db-user.py
```diff
@@ -0,0 +1,102 @@
...
+
+    except Exception:
+        logger.exception("Failed to create privileged database user.")
```
junhaoliao:
```suggestion
        logger.exception("Failed to create admin database user.")
```

junhaoliao (changes requested):

components/clp-package-utils/clp_package_utils/general.py
gibber9809: iirc we found some resources indicating that 8 is a bit low since Redis allows you to brute force many auth requests very quickly. It seems like 16 is also the minimum that [aws](https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/auth.html#auth-overview) uses (though I guess this page says they now don't want you using password auth at all).

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -295,6 +299,12 @@ def load_credentials_from_file(self, credentials_file_path: pathlib.Path):
         if config is None:
             raise ValueError(f"Credentials file '{credentials_file_path}' is empty.")
         try:
+            self.credentials[ClpDbUserType.ADMIN].username = get_config_value(
+                config, f"{DB_COMPONENT_NAME}.admin_username"
+            )
+            self.credentials[ClpDbUserType.ADMIN].password = get_config_value(
+                config, f"{DB_COMPONENT_NAME}.admin_password"
+            )
```
junhaoliao, line 307: it appears assignments on an already created Pydantic Model objects does not trigger (the non-empty-str) validations, unless we explicitly state so. e.g.,
```py
class DbUserCredentials(BaseModel, validate_assignment=True):
```

---

instead of specifying the `validate_assignment=True` in `DbUserCredentials` and expect validations to happen during attribute assignments, i think a better way is to construct `DbUserCredentials` objects directly with user values. e.g.
```py
    def load_credentials_from_file(self, credentials_file_path: pathlib.Path):
    ...
            self.credentials[ClpDbUserType.ADMIN] = DbUserCredentials(
                username=get_config_value(config, f"{DB_COMPONENT_NAME}.admin_username"),
                password=get_config_value(config, f"{DB_COMPONENT_NAME}.admin_password"),
            )
            self.credentials[ClpDbUserType.CLP] = DbUserCredentials(
                username=get_config_value(config, f"{DB_COMPONENT_NAME}.username"),
                password=get_config_value(config, f"{DB_COMPONENT_NAME}.password"),
            )
            self.credentials[ClpDbUserType.ROOT] = DbUserCredentials(
                username=get_config_value(config, f"{DB_COMPONENT_NAME}.root_username"),
                password=get_config_value(config, f"{DB_COMPONENT_NAME}.root_password"),
            )
    ...
```
and don't forget
```py
    def load_credentials_from_env(self, user_type: ClpDbUserType = ClpDbUserType.CLP):
    ...
        self.credentials[user_type] = DbUserCredentials(
            username=_get_env_var(user_env_var),
            password=_get_env_var(pass_env_var)
        )
```

then, we can also set an empty dict for the initial value of Database.credentials:
```py
    credentials: dict[ClpDbUserType, DbUserCredentials] = {}
```

now, since we can always expect the username and password to be non-None in `DbUserCredentials` models, we can update the model as:
```py
class DbUserCredentials(BaseModel):
    """Credentials for a database user."""

    username: NonEmptyStr
    password: NonEmptyStr
```

then we can update the check in 
```
    def ensure_credentials_loaded(self, user_type: ClpDbUserType) -> None:
        """
        Ensures that credentials for the given `user_type` are loaded.

        :param user_type:
        :raise ValueError: If credentials for the given `user_type` are not loaded.
        """
        if user_type not in self.credentials:
            err_msg = f"Credentials for user type '{user_type}' are not loaded."
            raise ValueError(err_msg)
```

components/clp-py-utils/clp_py_utils/initialize-admin-db-user.py
```diff
@@ -0,0 +1,101 @@
...
+        admin_username = clp_config.database.credentials[ClpDbUserType.ADMIN].username
+        admin_password = clp_config.database.credentials[ClpDbUserType.ADMIN].password
+        regular_username = clp_config.database.credentials[ClpDbUserType.CLP].username
+
+        if admin_username is None or admin_password is None or regular_username is None:
+            logger.error("Database credentials unexpectedly missing during privileged user setup.")
+            return -1
+
+        escaped_admin_username = admin_username.replace("'", "''")
+        escaped_admin_password = admin_password.replace("'", "''")
```
junhaoliao, line 58: the check `if admin_username is None or admin_password is None or regular_username is None` is a bit redundant now, given `load_credentials_from_env()` should have ensured those credentials exist. also, following the above suggestions about removing the `None` type specifications from the `DbUserCredentials.username` and `DbUserCredentials.password`, i think we can simply this block as:
```suggestion
        admin_username = clp_config.database.credentials[ClpDbUserType.ADMIN].username
        admin_password = clp_config.database.credentials[ClpDbUserType.ADMIN].password
        escaped_admin_username = admin_username.replace("'", "''")
        escaped_admin_password = admin_password.replace("'", "''")
```

(note `regular_username` is remove too, cuz it's unused below)

components/clp-py-utils/clp_py_utils/initialize-admin-db-user.py
```diff
@@ -0,0 +1,101 @@
...
+        credentials_table = get_aws_credentials_table_name(table_prefix)
+        with closing(sql_adapter.create_connection(True, ClpDbUserType.ROOT)) as db_conn, closing(
+            db_conn.cursor(dictionary=True)
+        ) as cursor:
+            logger.info(f"Creating admin database user '{admin_username}'...")
+
+            cursor.execute(
+                f"CREATE USER IF NOT EXISTS '{escaped_admin_username}'@'%'"
+                f" IDENTIFIED BY '{escaped_admin_password}'"
+            )
```
junhaoliao, line 70: let's also handle the case where Package users may ever change the clp-admin password in between Package restarts (for whatever reason...)
```suggestion
            cursor.execute(
                f"CREATE USER IF NOT EXISTS '{escaped_admin_username}'@'%'"
                f" IDENTIFIED BY '{escaped_admin_password}'"
            )
            cursor.execute(
                f"ALTER USER '{escaped_admin_username}'@'%' IDENTIFIED BY '{escaped_admin_password}'"
            )
```

junhaoliao:
there's also a lint failure in the new metadata db init script. running `task lint:fix-py` should fix it

components/clp-py-utils/clp_py_utils/clp_config.py
Eden-D-Zhang: Should this change be made on the main branch?

Eden-D-Zhang: @coderabbitai make an issue to track this change

PR #1636, opened 2025-11-20
feat(api-server): Add support for streaming query results from files.

hoophalab:
# Description

1. If the job config sets write_to_disk as true, then fetch_result in client will stream results from the files.

# Checklist

* [x] The PR satisfies the [contribution guidelines][y1. Send cope-contrib-guidelines].
  breaking change.

# Validation performed

1. Compress `postgresql` compressed to dataset `postgresql`.
2. Post `{"query_string": "line_num: 4", "dataset": "postgresql", "write_to_file": true}` to `http://localhost:3001/query`
Response: `{"query_results_uri":"/query_results/10"}`
2. Get `http://localhost:3001/query_results/10`. The result is correct.

## Summary by CodeRabbit

* **New Features**
  * Configurable result streaming with per-job selectable storage backends (filesystem or database).

* **Improvements**
  * Per-job routing of result streams to the chosen backend for consistent delivery.
  * Broader malformed-data detection for more resilient error handling.
  * New configuration options and sensible defaults for stream output storage.

* **Chores**
  * Deployment now mounts host directories for staged and live stream output.


components/api-server/src/client.rs
```diff
@@ -167,25 +170,48 @@ impl Client {
...
+                while let Some(entry) = entries.next_entry().await? {
+                    let file_path = entry.path();
+                    let reader = std::fs::File::open(file_path)?;
+                    let mut deserializer = rmp_serde::Deserializer::new(reader);
+                    while let Ok(e) = Deserialize::deserialize(&mut deserializer) {
+                        let e: (i64, String, String, String, i64) = e;
+                        yield Ok(e.1);
+                    }
+                }
             };
```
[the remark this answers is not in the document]
hoophalab: `rmp_serde::Deserializer::new` requires a blocking reader. We yield on each message. This shouldn't block the thread too long.

components/clp-rust-utils/src/clp_config/package/config.rs
```diff
@@ -147,3 +149,33 @@ impl Default for ResultsCache {
         }
     }
 }
+
+/// Mirror of `clp_py_utils.clp_config.StreamOutput`.
+///
+/// # NOTE
+///
+///   deserialization.
```
LinZhihao-723:
```suggestion
```

components/api-server/src/client.rs
```diff
@@ -167,25 +170,48 @@ impl Client {
...
-            .database(&self.config.results_cache.db_name);
-        let collection: mongodb::Collection<mongodb::bson::Document> =
-            database.collection(&search_job_id.to_string());
-        let cursor = collection.find(mongodb::bson::doc! {}).await?;
-
-        let mapped = cursor.map(|res| {
-            let doc = res?;
-            let Some(msg) = doc.get("message") else {
-                return Err(ClientError::MalformedData);
+        let job_config = self.get_job_config(search_job_id).await?;
```
LinZhihao-723, line 173: How about rewriting in the following way?
```rust
        let job_config = self.get_job_config(search_job_id).await?;
        if !job_config.write_to_file {
            let database = self
                .mongodb_client
                .database(&self.config.results_cache.db_name);
            let collection: mongodb::Collection<mongodb::bson::Document> =
                database.collection(&search_job_id.to_string());
            let cursor = collection.find(mongodb::bson::doc! {}).await?;

            let mapped = cursor.map(|res| {
                let doc = res?;
                let Some(msg) = doc.get("message") else {
                    return Err(ClientError::MalformedData);
                };
                let mongodb::bson::Bson::String(message) = msg else {
                    return Err(ClientError::MalformedData);
                };
                Ok(message.clone())
            });
            return Ok(ResultsStream::new(mapped));
        }

        let StreamOutputStorage::Fs {
            directory: stream_output_dir,
        } = &self.config.stream_output.storage
        else {
            todo!("Outputting query results to S3 is not supported for now.");
        };
        let search_job_output_dir = std::path::Path::new("/")
            .join(stream_output_dir)
            .join(search_job_id.to_string());
        let stream = stream! {
                let mut entries = tokio::fs::read_dir(search_job_output_dir).await?;
                while let Some(entry) = entries.next_entry().await? {
                    let search_result_path = entry.path();
                    let reader = std::fs::File::open(search_result_path)?;
                    let mut deserializer = rmp_serde::Deserializer::new(reader);
                    while let Ok(e) = Deserialize::deserialize(&mut deserializer) {
                        let e: (i64, String, String, String, i64) = e;
                        yield Ok(e.1);
                    }
                }
            };
        Ok(ResultsStream::new(stream))
```
* We should really avoid having return in both if and else branch.
* Rename some variables for clarity.
* Ideally, we should serialize the search result into some named struct. Using positional access is hard to read and maintain. Can u try if we can deserialize the msgpack array into a struct and we explicitly name the fields inside?
```rust
                while let Ok(e) = Deserialize::deserialize(&mut deserializer) {
                    let e: (i64, String, String, String, i64) = e;
                    yield Ok(e.1);
                }
```

components/api-server/src/client.rs
```diff
@@ -211,4 +237,43 @@ impl Client {
...
+        .bind(search_job_id)
+        .fetch_one(&self.sql_pool)
+        .await?;
+        let msgpack: &[u8] = row.try_get("job_config")?;
+        Ok(rmp_serde::from_slice(msgpack)?)
+    }
+}
+
+type PinnedStream = Pin<Box<dyn Stream<Item = Result<String, ClientError>> + Send>>;
+struct ResultsStream(PinnedStream);
```
LinZhihao-723: * How about `SearchResultStream`?
* I'm not sure I fully get this. Can u explain why we need a wrapper of the stream which requires `unsafe` to access it? That said, I think it's worth having a docstring for this type briefly explaining why we need to do this.
* If we want to keep this wrapper, we should define the pinned stream as a named field. We shouldn't allow positional fields in general.

LinZhihao-723 (changes requested):

components/api-server/src/client.rs
hoophalab: `Sse` requires a `Stream` that implements `Send`. This wasn't an issue before because we only had mongo stream. Now, we have two concrete stream types: file system stream and mongo stream, both of which are `Send`. 
We need to return a concrete type `ResultsStream` in `fetch_results` that implements `Send`.

Box<dyn ...> is to dynamically dispatch two concrete stream types.
Pin is to satisfy the Unpin condition in `Stream::poll_next`
Polling from a stream requires a mut Stream, we could have used a `Mutex`, but that would be unnecessarily costly as suggested by coderabbit. Admittedly, this part is a bit nasty. I'm happy to change the code if we can come up with a better solution.

- This type is internal. I guess both no docstring and tuple types are common, at least for internal types, but I'm open to add them.

LinZhihao-723: * Haven't thought this very deeply, but can we use an enum?
* We should avoid using tuple types to follow our standard, no matter whether it's internal or external. This was also suggested by Adrian when I asked his opinion a year ago.
* In this case I think the docstring is necessary if u decide to go with this implementation. Otherwise future developers could ask the same questions as I did here, lol. I think in general we should always add docstring for types.

hoophalab: enum could work, but it's just an alternative to dyn, and we would still need unsafe

LinZhihao-723: So if I understand correctly, the reason for using unsafe in an enum is also to avoid using mutex? But since `poll_next` takes a mutable pinned ref of the enum, is it still a problem for it?

hoophalab: Unless I miss something, enum probably also requires this unsafe projection

LinZhihao-723: As discussed offline, we will go with `enum` and use `pin_project` to not use unsafe explicitly.

components/api-server/src/client.rs
```diff
@@ -167,6 +177,48 @@ impl Client {
...
+            let mut entries = tokio::fs::read_dir(search_job_output_dir).await?;
+            while let Some(entry) = entries.next_entry().await? {
+                let search_result_path = entry.path();
+                let reader = std::fs::File::open(search_result_path)?;
+                let mut deserializer = rmp_serde::Deserializer::new(reader);
+                while let Ok(event) = Deserialize::deserialize(&mut deserializer) {
+                    let event: (i64, String, String, String, i64) = event;
+                    let message = event.1;
+                    yield Ok(message);
+                }
```
hoophalab, line 259: Tuple is indeed hard to read, but clp-s serializes the message in a tuple to save the bandwidth of transferring key names. The format is inherited from NetworkAddress output's format. We probably cannot do more here unless there is an automatic way to serialize a tuple into a struct.

I did something here. Does this read better?

components/api-server/src/client.rs
```diff
@@ -167,6 +177,48 @@ impl Client {
...
+        let job_config = self.get_job_config(search_job_id).await?;
+        if job_config.write_to_file {
+            Ok(ResultsStream::File {
+                inner: self.fetch_results_from_file(search_job_id),
+            })
+        } else {
+            self.fetch_results_from_mongo(search_job_id)
+                .await
+                .map(|s| ResultsStream::Mongo { inner: s })
+        }
```
hoophalab, line 178: How about this? Or we should still put everything in one function.

components/api-server/src/client.rs
```diff
@@ -167,6 +177,48 @@ impl Client {
...
+                .await
+                .map(|s| SearchResultStream::Mongo { inner: s })
+        }
+    }
+
+    fn fetch_results_from_file(
+        &self,
+        search_job_id: u64,
+    ) -> impl Stream<Item = Result<String, ClientError>> + use<> {
+        let StreamOutputStorage::Fs { directory } = &self.config.stream_output.storage else {
```
LinZhihao-723, line 244: Nit: `directory` -> `stream_storage_dir`?
`directory` seems too general.

components/api-server/src/client.rs
LinZhihao-723: I think this is what I can get for readability.
I think using the tuple should also be fine. The real issue is we don't have tests to enforce the format on clp-s side and the tuple here will match.
```
#[derive(Deserialize)]
struct EventTuple(i64, String, String, String, i64);

#[derive(Deserialize, From)]
#[serde(from = "EventTuple")]
struct Event {
    x: i64,
    message: String,
    y: String,
    z: String,
    n: i64,
}
```

components/api-server/src/client.rs
```diff
@@ -167,6 +177,48 @@ impl Client {
...
+                inner: self.fetch_results_from_file(search_job_id),
+            })
+        } else {
+            self.fetch_results_from_mongo(search_job_id)
+                .await
+                .map(|s| SearchResultStream::Mongo { inner: s })
+        }
+    }
+
+    fn fetch_results_from_file(
```
LinZhihao-723, line 240: Let's group private methods after public methods.

components/api-server/src/client.rs
```diff
@@ -167,6 +177,48 @@ impl Client {
...
+                    let event: (i64, String, String, String, i64) = event;
+                    let message = event.1;
+                    yield Ok(message);
+                }
+            }
+        };
+        stream
+    }
+
+    async fn fetch_results_from_mongo(
```
LinZhihao-723, line 289:
```suggestion
    /// Asynchronously fetches results of a completed search job from MongoDB.
    ///
    /// # Returns
    ///
    /// A stream of the job's results on success. Each item in the stream is a [`Result`] that:
    ///
    /// ## Returns
    ///
    /// A parsed JSON value representing a search result on success.
    ///
    /// ## Errors
    ///
    /// Returns an error if:
    ///
    /// * [`ClientError::MalformedData`] if a retrieved document does not contain a "message" field,
    ///   or if the "message" field is not a BSON string.
    /// * Forwards [`mongodb::error::Error`] produced by the `MongoDB` cursor item access.
    /// * Forwards [`serde_json::from_str`]'s return values on failure.
    ///
    /// # Errors
    ///
    /// Returns an error if:
    ///
    /// * Forwards [`mongodb::Collection::find`]'s return values on failure.
    async fn fetch_results_from_mongo(
```
Docstring for MongoDB.

components/api-server/src/client.rs
LinZhihao-723, line 240:
```suggestion
    /// Asynchronously fetches results of a completed search job from files.
    ///
    /// # Returns
    ///
    /// A stream of the job's results on success. Each item in the stream is a [`Result`] that:
    ///
    /// ## Returns
    ///
    /// The log message in string representation on success.
    ///
    /// ## Errors
    ///
    /// Returns an error if:
    ///
    /// * Forwards [`std::fs::File::open`]'s return values on failure.
    /// * Forwards [`tokio::fs::read_dir`]'s return values on failure.
    /// * Forwards [`tokio::fs::ReadDir::next_entry`]'s return values on failure.
    ///
    /// # Panics
    ///
    /// Panics if the stream output storage is not file system.
    fn fetch_results_from_file(
```
Docstring for file system.

Just to confirm: the message we return is the entire JSON message serailized as JSON string, right?

components/api-server/src/client.rs
```diff
@@ -149,7 +153,13 @@ impl Client {
     pub async fn fetch_results(
```
LinZhihao-723, line 146: Docstring update:
```
    /// Asynchronously fetches the results of a completed search job.
    ///
    /// # Returns
    ///
    /// A stream of the job's results on success. The stream variant indicates where the results
    /// are stored:
    ///
    /// * [`SearchResultStream::File`] if the job's results are stored in files.
    /// * [`SearchResultStream::Mongo`] if the job's results are stored in MongoDB.
    ///
    /// # Errors
    ///
    /// Returns an error if:
    ///
    /// * Forwards [`Client::get_status`]'s return values on failure.
    /// * Forwards [`Client::get_job_config`]'s return values on failure.
    /// * Forwards [`Client::fetch_results_from_mongo`]'s return values on failure.
```

components/api-server/src/client.rs
```diff
@@ -167,6 +177,48 @@ impl Client {
...
+        let job_config = self.get_job_config(search_job_id).await?;
+        if job_config.write_to_file {
+            Ok(SearchResultStream::File {
+                inner: self.fetch_results_from_file(search_job_id),
+            })
+        } else {
+            self.fetch_results_from_mongo(search_job_id)
+                .await
+                .map(|s| SearchResultStream::Mongo { inner: s })
+        }
```
LinZhihao-723, line 178:
```suggestion
        if job_config.write_to_file {
            return Ok(SearchResultStream::File {
                inner: self.fetch_results_from_file(search_job_id),
            });
        }
        self.fetch_results_from_mongo(search_job_id)
            .await
            .map(|s| SearchResultStream::Mongo { inner: s })
```
Nit: I think in Rust, this if-else is actually treated as a single return statement. However, for the purpose of reducing indentations, we may want to write it this way, lol

components/api-server/src/client.rs
```diff
@@ -211,4 +262,49 @@ impl Client {
...
+    ) -> std::task::Poll<Option<Self::Item>> {
+        let poll = match self.project() {
+            SearchResultStreamProj::File { inner } => inner.poll_next(cx),
+            SearchResultStreamProj::Mongo { inner } => inner.poll_next(cx),
+        };
+        if let std::task::Poll::Ready(Some(Err(err))) = &poll {
+            tracing::error!("An error occurred when streaming results: {}", err);
+        }
+        poll
+    }
```
LinZhihao-723, line 351:
```suggestion
pin_project! {
    /// Enum for search result streams from different storage backends.
    ///
    /// # Type Parameters
    ///
    /// * `FileStream`: Streaming from file system storage.
    /// * `MongoStream`: Streaming from MongoDB storage.
    #[project = SearchResultStreamProj]
    pub enum SearchResultStream<FileStream, MongoStream>
    where
        FileStream: Stream<Item = Result<String, ClientError>>,
        MongoStream: Stream<Item = Result<String, ClientError>>
    {
        File{#[pin] inner: FileStream},
        Mongo{#[pin] inner: MongoStream},
    }
}

impl<FileStream, MongoStream> Stream for SearchResultStream<FileStream, MongoStream>
where
    FileStream: Stream<Item = Result<String, ClientError>>,
    MongoStream: Stream<Item = Result<String, ClientError>>,
{
    type Item = Result<String, ClientError>;

    fn poll_next(
        self: Pin<&mut Self>,
        cx: &mut std::task::Context<'_>,
    ) -> std::task::Poll<Option<Self::Item>> {
        let poll = match self.project() {
            SearchResultStreamProj::File { inner } => inner.poll_next(cx),
            SearchResultStreamProj::Mongo { inner } => inner.poll_next(cx),
        };
        if let std::task::Poll::Ready(Some(Err(err))) = &poll {
            tracing::error!("An error occurred when streaming results: {}", err);
        }
        poll
    }
```
* Adding docstring.
* Rename `S1` to `FileStream`, `S2` to `MongoStream` for clarity. In general let's try to make the type parameter names explicit to improve readabilitiy.

LinZhihao-723 (changes requested):
The enum implementation is clear now, lgtm.
 The comments are mostly about docstrings and naming conventions.

components/api-server/src/client.rs
LinZhihao-723: Yeah I think this is clear. I added docstrings for both methods. Splitting them makes it easier to document tbh

components/api-server/src/client.rs
hoophalab: yes

hoophalab:
> Nit: directory -> stream_storage_dir?
> directory seems too general.

"directory" is the field name in the yaml file. It might not be ideal to rename it.

LinZhihao-723 (approved):
For the PR title, how about:
```
feat(api-server): Add support for streaming query results from files.
```

LinZhihao-723 (approved):

PR #1638, opened 2025-11-20
feat(webui): Enforce 2-column layout for SQL input grid in guided Presto UI.

davemarco:
# Description

Previous grid in presto UI used auto-fit to dynamically position sql fields. Therefore on a larger screen, the layout could potentially be on one row. We determined this is undesirable, and now it is always on one row.

# Checklist

  breaking change.

# Validation performed

always on 2 rows

## Summary by CodeRabbit

* **Style**
  * Updated search filter layout to a fixed two-column grid with smaller column minima.
  * Converted previously full-width controls to single-column grid items for consistent alignment.
  * Removed multi-column spanning for several controls to tighten layout.
  * Optimized element sizing for a more compact, uniform search interface presentation.

---
  - https://app.asana.com/0/0/1212018015778917


components/webui/client/src/pages/SearchPage/SearchControls/Presto/GuidedControls/index.module.css
```diff
@@ -1,17 +1,17 @@
 .gridContainer {
     display: grid;
-    grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));
+    grid-template-columns: repeat(2, minmax(100px, 1fr));
     grid-auto-rows: 32px;
     gap: 5px;
 }
 
 .select {
```
junhaoliao: kinda unrelated to the changes in this PR - would it be cleaner to just write
```css
.select .from .where .order {
    grid-column: span 1;
    display: flex;
}
```
to avoid repeating the style definitions? if in the future they will have different styles, we can always split the block?

junhaoliao (approved):
i use the browser debugger tab to simulate various screen resolutions whose widths range from 1000px - 2560px. the layout is consistently 2-column which matches the purpose of the change. 

for the title, how about:
```
feat(webui): Enforce a fixed 2-column layout for guided Presto search controls.
```

---

note the time range selector may overflow the window when the window width is less than 1000px, though i think that's unrelated to the current PR. we should submit an issue to track the fix if the problem is confirmed.

image

components/webui/client/src/pages/SearchPage/SearchControls/Presto/GuidedControls/index.module.css
davemarco: I changed to one class

davemarco:
@junhaoliao the overflow issue will be better after we merge the PR where the timekey selector is moved. There will be more space on the top row.

davemarco:
@junhaoliao can you reapprove to due to merge conflicts

junhaoliao (approved):

PR #1640, opened 2025-11-20
style(clp-package): Use absolute imports for all Python sources; Bump `yscope-dev-utils` to y-scope/yscope-dev-utils@307e711.

sitaowang1998:
# Description

This PR changes all Python imports to absolute, conforming to the requirement of [`ruff` TID252](https://docs.astral.sh/ruff/rules/relative-imports/) rule, and our own guideline to ban all relative imports.

This PR also changes Python script call to use `python -m <project>.<file_name>` install of `python <file_path.py>`.

# Checklist

  breaking change.

# Validation performed

* [x] `task lint:check-py` no longer complains about TID252.
* [x] End-to-end compression and search using docker compose works.

## Summary by CodeRabbit

* **Refactor**
  * Standardized internal module import structure across components for improved code organization.

* **Chores**
  * Updated development dependencies and tooling versions.

### DIFF
--- a/components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
+++ b/components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
@@ -3,9 +3,8 @@ import json
 from clp_py_utils.clp_logging import get_logger
 from spider_py import Int8, Int64, TaskContext
 
-from job_orchestration.utils.spider_utils import int8_list_to_utf8_str, utf8_str_to_int8_list
-
 from job_orchestration.executor.compress.compression_task import compression_entry_point
+from job_orchestration.utils.spider_utils import int8_list_to_utf8_str, utf8_str_to_int8_list
 
 # Setup logging
 logger = get_logger("spider_compression")
--- a/components/clp-mcp-server/clp_mcp_server/server/server.py
+++ b/components/clp-mcp-server/clp_mcp_server/server/server.py
@@ -8,7 +8,6 @@ from starlette.requests import Request
 from starlette.responses import PlainTextResponse
 
 from clp_mcp_server.clp_connector import ClpConnector
-
 from clp_mcp_server.server import constants
 from clp_mcp_server.server.session_manager import SessionManager
 from clp_mcp_server.server.utils import (
--- a/components/clp-package-utils/clp_package_utils/scripts/native/archive_manager.py
+++ b/components/clp-package-utils/clp_package_utils/scripts/native/archive_manager.py
@@ -28,7 +28,6 @@ from clp_package_utils.scripts.archive_manager import (
     END_TS_ARG,
     FIND_COMMAND,
 )
-
 from clp_package_utils.scripts.native.utils import validate_dataset_exists
 
 logger: logging.Logger = logging.getLogger(__file__)
--- a/components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
+++ b/components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
@@ -31,7 +31,6 @@ from clp_package_utils.general import (
     get_clp_home,
     load_config_file,
 )
-
 from clp_package_utils.scripts.native.utils import (
     run_function_in_process,
     submit_query_job,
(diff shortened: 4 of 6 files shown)


junhaoliao:
I think there are more files that could be modified to use relative imports from its direct parent level:
- `components\job-orchestration\job_orchestration\scheduler\compress\task_manager\celery_task_manager.py`
- `components\job-orchestration\job_orchestration\scheduler\compress\compression_scheduler.py`
- `components\job-orchestration\job_orchestration\executor\query\extract_stream_task.py`
- `components\job-orchestration\job_orchestration\executor\query\fs_search_task.py`
- `components\job-orchestration\job_orchestration\scheduler\query\query_scheduler.py`
- `components\job-orchestration\job_orchestration\scheduler\scheduler_data.py`
- `components\job-orchestration\job_orchestration\scheduler\compress\task_manager\spider_task_manager.py`
- `components\job-orchestration\job_orchestration\scheduler\utils.py`

---

Please allow me to take a step back. 

The reason why we initiated this PR was because we were importing relative modules without `.`. e.g., `from sql_adapter import xxx` should be written as `from .sql_adapter import xxx` or `from path.to.sql_adapter import xxx`. That we should avoid importing relative modules without `.` is definitely not questionable.

We later agree that we could change any direct relative imports because such imports do not violate ruff's TID252 (by default, because the [lint.flake8-tidy-imports](https://docs.astral.sh/ruff/settings/#lintflake8-tidy-imports) 's [ban-relative-imports](https://docs.astral.sh/ruff/settings/#lint_flake8-tidy-imports_ban-relative-imports) is by default configured as `"parents"` rather than `"all"`). However, if we revisit [the PEP8 style guide's Import section](https://peps.python.org/pep-0008/#imports):

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages)

...

> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose

We can see that "explicit relative imports" are allowed as a compromise when the layout is too complex, which isn't a case in our projects (i.e., we don't really have too many deeply nested modules).

For code explicitness and maintainability concerns, i think it's better to ban all relative imports by changing `ban-relative-imports = "all"`, then:
1. IDEs / humans can easily catch & modify any import paths that should be updated when we move files around;
2. Linters can easily catch any inconsistent ways of writing the same code.

There's been an issue y-scope/yscope-dev-utils#99 to track the linter rule change request

For this PR, we can manually update tools/yscope-dev-utils/exports/lint-configs/python/ruff.toml to set
```
[lint.flake8-tidy-imports]
# Disallow all relative imports.
ban-relative-imports = "all"
```
Then rerun `uv run --directory "xxx" ruff check --select TID252 --fix --unsafe-fixes` for the Python projects

junhaoliao:
https://github.com/y-scope/yscope-dev-utils/pull/100 is now merged, let's update the yscopde-dev-utils submodule in this PR then this PR will be good to be merged. thanks

components/job-orchestration/job_orchestration/executor/compress/spider_compress.py
```diff
@@ -3,9 +3,10 @@
 from clp_py_utils.clp_logging import get_logger
 from spider_py import Int8, Int64, TaskContext
 
-from job_orchestration.executor.compress.compression_task import compression_entry_point
 from job_orchestration.utils.spider_utils import int8_list_to_utf8_str, utf8_str_to_int8_list
 
+from job_orchestration.executor.compress.compression_task import compression_entry_point
+
```
junhaoliao, line 8: let's run `task lint:fix-py` before committing then the PR is good to go

```suggestion
from job_orchestration.utils.spider_utils import int8_list_to_utf8_str, utf8_str_to_int8_list
from job_orchestration.executor.compress.compression_task import compression_entry_point

```

junhaoliao (changes requested):
for the title, how about:
```
style(clp-package): Use absolute imports for all Python sources; Bump `yscope-dev-utils` to y-scope/yscope-dev-utils@307e711.
```

junhaoliao (approved):

PR #1653, opened 2025-11-24
feat(clps): Add experimental support for storing and searching semantic information from unstructured text fields in JSON logs.

davidlion:
# Description

This PR adds a few experimental features while refactoring some shared code:

* Adds experimental semantic support using log-surgeon.
    * When enabled, JSON fields that are deemed to be unstructured text are parsed with log-surgeon and stored inside the CLP-S columns/schema using the variable (or capture group) name as the key.
    * Supports searching as if the variables found in the unstructured text were JSON fields.

* Adds experimental statistic that can be output through hard-coded query strings.
    * `clp_s::CommandLineArguments::ExperimentalQueries` defines:
        * `stats.logtypes`: For each logtype outputs a JSON object containing: the logtype ID, the occurrence count, and the logtype itself.
            * e.g. `{"id":0,"count":1,"log_type":"... the logtype with placeholders ..."}`
        * `stats.variables`: For each dictionary variable outputs a JSON object containing: the dictionary variable ID, the type (variable or capture group name in the log-surgeon schema), the occurrence count, and the variable itself.
             * e.g. `{"id":2298,"count":1,"type":"url_link","variable":"/YouShouldSeeThis.txt"}`
     * Stored to `/stats` in the archive.

* Adds a `Array` class made to be compressed/decompressed into the archive.
    * Essentially a simplified `DictionaryEntry` without the ability to look-up an entry.
    * Used to implement/store the experimental statistics.

* Adds a command line option to enable experimental features.
    * Adds `ArchiveReader::Options` to generalize propagating options.

# Checklist

  breaking change.

# Validation performed


components/core/src/clp/ir/types.hpp
```diff
@@ -18,6 +18,7 @@ enum class VariablePlaceholder : char {
     Integer = 0x11,
     Dictionary = 0x12,
     Float = 0x13,
+    Schema = 0x14,
```
gibber9809: A new variable placeholder is added here, but it seems like `is_variable_placeholder` in `clp/ir/parsing.hpp` hasn't been updated to include the placeholder. This means that the escaping code won't end up properly escaping `0x14` when it appears in logtext, which is a bug.

Granted, if we do properly escape it then this constitutes a major breaking format change that isn't guarded by the experimental flag. Given that, it might actually be better to intentionally leave this bug for now to avoid the breaking format change. The only alternative I can think of is guarding everything related to `VariablePlacehodler::Schema` by a bunch of compile time macros, but then we would need to ship a separate build for this experimental change.

components/core/src/clp_s/ColumnWriter.hpp
```diff
@@ -43,6 +46,12 @@ class BaseColumnWriter {
      */
     virtual size_t get_total_header_size() const { return 0; }
 
+    /**
+     *
+     * @return the total size of header data that will be written to the compressor in bytes
+     */
+    virtual auto get_stats() const -> size_t { return 0; }
+
```
gibber9809:
```suggestion
```
Unused?

components/core/src/clp_s/ColumnWriter.hpp
```diff
@@ -219,6 +223,65 @@ class ClpStringColumnWriter : public BaseColumnWriter {
...
+    }
+
+private:
+    static constexpr int cOffsetBitPosition = 24;
+    static constexpr uint64_t cLogDictIdMask = (1ULL << cOffsetBitPosition) - 1;
+    static constexpr uint64_t cOffsetMask = ~cLogDictIdMask;
+
+    std::shared_ptr<LogTypeDictionaryWriter> m_log_dict;
+
+    std::vector<encoded_log_dict_id_t> m_logtypes;
```
gibber9809: Looks like we end up just storing `clp::logtype_dictioary_id_t` with a static cast to `encoded_log_dict_id_t`. Not strictly necessary for this prototype, but I'd probably just change this to directly storing `clp::logtype_dictionary_id_t` -- no real reason to use `encoded_log_dict_id_t here`.

components/core/src/clp_s/SchemaTree.hpp
```diff
@@ -44,6 +62,10 @@ enum class NodeType : uint8_t {
     DeltaInteger,
     FormattedFloat,
     DictionaryFloat,
+    LogMessage,
+    LogType,
+    TypedVar,
```
gibber9809:
```suggestion
```
Since the actual storage format for `TypedVar` is identical to `VarString`, I think it's probably best to just get rid of `TypedVar`. Reasoning generally being that it gets harder to manage the storage format over time if we have multiple column types that are basically the same thing, and also that we have a limited number of `NodeType`s (since they're 8 bits), so we generally want to avoid creating new ones when possible.

You'd need to move the variable dictionary encoding + the stat tracking into `JsonParser` (or maybe some utility in `ArchiveWriter` that you call from `JsonParser`) & change `ParsedMessage`/`VarStringColumnWriter` to allow directly passing a `clp::variable_dictionary_id_t`, but I think it's worth it.

components/core/src/clp_s/SchemaTree.hpp
```diff
@@ -44,6 +62,10 @@ enum class NodeType : uint8_t {
     DeltaInteger,
     FormattedFloat,
     DictionaryFloat,
+    LogMessage,
```
gibber9809:
```suggestion
    LogMessage = 100,
```
I think this is a worthwhile precaution just to make sure that the IDs for these unfinalized node types that'll only be produced under the experimental flag don't overlap with IDs for node types in non-experimental releases for the foreseeable future.

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -1337,4 +1471,229 @@ bool JsonParser::check_and_log_curl_error(
...
+// TODO clpsls: capture group repetition
+// Similarly, to variable repetition storing the capture group variable node as an array in cases
+// with repetition creates a node type of T|Array[T].
+//
+// TODO clpsls:
+// Storing both the full match and capture groups creates potential duplication of the capture
+// group's substring.
+// One option is to re-write the full match similar to a logtype where the capture groups are
+// replaced with encoded placeholders. The full match would need to be rebuilt using its capture
+// groups to display.
```
gibber9809: Also maybe possible to just store the parts of the full match that aren't the capture groups as part of the static logtext? Not sure if that easily fits into how log-surgeon works though.

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -1337,4 +1471,229 @@ bool JsonParser::check_and_log_curl_error(
...
+// Additionally wildcard search must know to avoid searching both the full match and capture groups.
+// Using placeholders allows for rebuilding the variable (and log message) through node
+// position/ordering as the placeholders in the full match enable you to know how many subsequent
+// nodes are captures. Otherwise we either need to specially type the node (e.g. VarString and
+// CaptureString) or do some other indexing/checking.
+//
+// ... Storing as a separate type seems the most sensible.
+// There is a possible trade-off where de-duplicating the capture group string using placeholders in
+// the full match node improves compression ration, but slows down search in certain cases as you
+// maybe need to rebuild the full match's value.
```
gibber9809: Definitely all questions we should have answers for before this makes it into a proper non-experimental release, but seems ok for this prototype. I am a bit worried the full match/capture groups duplication will make it a bit harder to get a baseline for how these changes affect compression ratio though.

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -1337,4 +1471,229 @@ bool JsonParser::check_and_log_curl_error(
...
+                };
+                auto capture_start{m_current_schema.start_unordered_object(NodeType::CaptureVar)};
+
+                logtype_dict_entry.add_schema_var();
+                m_current_parsed_message.add_unordered_value(
+                        ParsedMessage::TypedVar{std::string{cFullMatch}, token_view.to_string()}
+                );
+                m_current_schema.insert_unordered(
+                        m_archive_writer->add_node(capture_node_id, NodeType::TypedVar, cFullMatch)
+                );
```
gibber9809: I think this works, but its a little bit strange from the perspective of the current implementation.

Issue here is that the schema is supposed to be a set of leaf nodes that dominate some subtree and imply the full structure of the record. Here we're storing the full match and all of the captures as siblings, but arguably the full match is the parent of the captures. That said, we definitely don't support having a schema tree node and its children be part of the same schema, so the way you're storing this is probably fine. I guess the "conventional" thing to do would be to just store the captures, but that requires some thought.

Should be fine for now since this is a prototype, but its something we need to think through.

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -1337,4 +1471,229 @@ bool JsonParser::check_and_log_curl_error(
...
+                    // TODO clpsls: change to store as an array?
+                    auto capture_name{lexer.m_id_symbol.at(capture_id)};
+                    if (false == start_positions.empty() && -1 < start_positions[0]
+                        && false == end_positions.empty() && -1 < end_positions[0])
+                    {
+                        auto capture_view{token_view};
+                        capture_view.set_start_pos(start_positions[0]);
+                        capture_view.set_end_pos(end_positions[0]);
+
+                        logtype_dict_entry.add_schema_var();
```
gibber9809: What does the logtype end up looking like? It seems like we add a schema var for the full match _and_ all of the captures, which seems weird.

gibber9809 (changes requested):
Left an initial set of comments mostly focused on format-related stuff.

The most major thing we need a resolution on is escaping for the variable placeholder since that's breaking, but it would also be nice to get rid of `TypedVar`. We should definitely thing through stuff related to the other comments, but doesn't necessarily require any action for the upcoming release since its guarded by the experimental flag.

Will also try to do a pass through the rest of the code to see if I notice anything else, but the storage format stuff is definitely the most important.

davidlion:
We decided to not merge this version of the experimental features upstream in favour of waiting for a more mature version.

PR #1658, opened 2025-11-25
feat(integration-tests): Record required components in `PackageConfig` based on the CLP package config.

quinntaylormitchell:
# Description

This PR introduces the `component_list` data member to `PackageConfig` data class. The data member is used for tracking which containerized components a given CLP package configuration needs to run.

### PR dependencies

This PR depends on the following being merged:
* #1437 

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

Ran the following command from `clp/integration-tests`:

```bash
uv run pytest -m 'package'
```

Both the `clp-json` and `clp-text` tests passed.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **Chores**
  * Component selection now adapts dynamically to deployment mode and retention/archive settings for more efficient setups.
* **New Features**
  * Package metadata includes an explicit list of required components.
  * Start-up logs now show the package mode and its required component list for easier verification and debugging.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1437, opened 2025-10-17
feat(integration-tests): Add fixtures and tests for starting and stopping CLP packages.

quinntaylormitchell:
# Description

This PR builds on the infrastructure introduced in [PR #1100](https://github.com/y-scope/clp/pull/1100) and [PR #1549](https://github.com/y-scope/clp/pull/1549) by introducing a parametrized test called `test_clp_package` that ensures two key configurations of the clp package (`clp-text` and `clp-json`) spin up and tear down successfully. `test_clp_package` uses the contents of a dictionary called `CLP_MODE_CONFIGS` to spin up each mode in its proper configuration.

### Dependent PRs

This is a foundational infra PR, and so there are other PRs which depend on this one. They are listed below:
* #1658
* #1659 
* #1660 
* #1662 
* #1664 

Compression PRs
* #1651 

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

Ran the following command from `clp/integration-tests`:

```bash
uv run pytest -m 'package'
```

Both the `clp-json` and `clp-text` tests passed.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **Tests**
  * Added integration tests to validate package startup across multiple modes, parameterized via fixtures.
  * New fixtures manage package configuration, lifecycle and cleanup; setup/teardown starts/stops the package and removes temporary config files.
  * Added mode utilities and helpers; improved command execution logging, timeout handling and file-existence validation.

* **Chores**
  * Pytest configured to report all outcomes and a new "package" test marker added.
  * Added YAML-related dev dependencies for test config handling.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1659, opened 2025-11-25
feat(integration-tests): Validate that the components of the CLP package are running (fixes #1644).

### LINKED ISSUE #1644: Switch to `docker compose ps` for container validation in integration tests
## Description

Currently, the integration test utilities in `integration-tests/tests/utils/docker_utils.py` use `docker ps` with name prefix matching to validate running containers (see `list_running_containers_with_prefix()`).

## Proposed Enhancement

Switch to using `docker compose ps` with the project name specified. This approach would provide:

1. **Exact container list**: Query the precise set of containers managed by the docker compose project, eliminating false positives from containers with similar name patterns
2. **Exit status validation**: Check for containers that should have exited at certain stages of the package run, enabling more comprehensive lifecycle validation
3. **Better integration**: Align with docker compose's native container management and project scoping

## Context

Backlinks:
- PR: https://github.com/y-scope/clp/pull/1437
- Comment: https://github.com/y-scope/clp/pull/1437#discussion_r2547812265
- Requested by: @Bill-hbrhbr

quinntaylormitchell:
# Description

This PR introduces the `validate_package_running` function to check if the containerized components of the given package are all running after spin-up.

`validate_package_running` uses `list_running_containers_in_compose_project` to get the list of running components, and processes this list in such a way that a 1:1 comparison can be performed against the `required_components` list for that package.

### PR dependencies

This PR depends on the following being merged:
* #1437 
* #1658 

# Checklist

  breaking change.

# Validation performed

```bash
uv run pytest -m 'package'
```

Both the `clp-json` and `clp-text` tests passed.

## Summary by CodeRabbit

* **Tests**
  * Simplified integration test package verification by using a single package-running validator to check active components.

* **Chores**
  * Added utilities to track package instance IDs and manage test log directories.
  * Added helpers to locate required system binaries and to list running Docker Compose services for more reliable test checks.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2142,6 +2142,7 @@ dependencies = [
  "clp-rust-utils",
  "const_format",
  "secrecy",
+ "serde",
  "serde_json",
  "serial_test",
  "sqlx",
--- a/components/clp-mcp-server/pyproject.toml
+++ b/components/clp-mcp-server/pyproject.toml
@@ -47,7 +47,7 @@ show_error_context = true
 show_error_end = true
 
 [tool.ruff]
-extend = "../../tools/yscope-dev-utils/exports/lint-configs/python/ruff.toml"
+extend = "../../ruff.toml"
 
 [tool.uv]
 default-groups = ["clp", "dev"]
--- a/components/clp-package-utils/pyproject.toml
+++ b/components/clp-package-utils/pyproject.toml
@@ -38,7 +38,7 @@ show_error_context = true
 show_error_end = true
 
 [tool.ruff]
-extend = "../../tools/yscope-dev-utils/exports/lint-configs/python/ruff.toml"
+extend = "../../ruff.toml"
 
 [tool.uv]
 default-groups = ["clp", "dev"]
--- a/components/clp-py-utils/pyproject.toml
+++ b/components/clp-py-utils/pyproject.toml
@@ -41,7 +41,7 @@ show_error_context = true
 show_error_end = true
 
 [tool.ruff]
-extend = "../../tools/yscope-dev-utils/exports/lint-configs/python/ruff.toml"
+extend = "../../ruff.toml"
 
 [tool.uv]
 default-groups = ["clp", "dev"]
--- a/components/clp-rust-utils/src/clp_config/package/config.rs
+++ b/components/clp-rust-utils/src/clp_config/package/config.rs
@@ -197,6 +197,7 @@ impl Default for StreamOutputStorage {
 pub struct LogIngestor {
     pub host: String,
     pub port: u16,
+    #[serde(rename = "buffer_timeout")]
     pub buffer_timeout_sec: u64,
     pub buffer_size_threshold: u64,
     pub channel_capacity: usize,
@@ -206,7 +207,7 @@ impl Default for LogIngestor {
     fn default() -> Self {
         Self {
             host: "localhost".to_owned(),
-            port: 3333,
+            port: 3270,
             buffer_timeout_sec: 300,
             buffer_size_threshold: 50 * 1024 * 1024,
             channel_capacity: 10,
--- a/components/clp-rust-utils/src/job_config.rs
+++ b/components/clp-rust-utils/src/job_config.rs
@@ -2,9 +2,8 @@ mod clp_io_config;
 pub use clp_io_config::*;
 
 mod compression;
-mod ingestion;
+pub mod ingestion;
 mod search;
 
 pub use compression::*;
-pub use ingestion::*;
 pub use search::*;
--- a/components/clp-rust-utils/src/job_config/ingestion.rs
+++ b/components/clp-rust-utils/src/job_config/ingestion.rs
@@ -1,24 +1,67 @@
-/// Base configuration for ingesting logs from S3.
-#[derive(Clone, Debug)]
-pub struct S3IngestionBaseConfig {
-    /// AWS service region.
-    pub region: String,
+pub mod s3 {
+    use serde::{Deserialize, Serialize};
 
-    /// The S3 bucket to ingest from.
-    pub bucket_name: String,
+    /// Base configuration for ingesting logs from S3.
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    pub struct BaseConfig {
+        /// AWS service region.
+        pub region: String,
 
-    /// The S3 key prefix to ingest from.
-    pub key_prefix: String,
+        /// The S3 bucket to ingest from.
+        pub bucket_name: String,
 
-    /// The dataset to ingest into. Using `default` if not specified.
-    pub dataset: Option<String>,
+        /// The S3 key prefix to ingest from.
+        pub key_prefix: String,
 
-    /// The optional key for extracting timestamps from object metadata.
-    pub timestamp_key: Option<String>,
+        /// The dataset to ingest into. Defaults to `None` (which uses the default dataset).
+        #[serde(default)]
+        pub dataset: Option<String>,
 
-    /// Whether to treat the ingested objects as unstructured logs.
-    pub unstructured: bool,
+        /// The optional key for extracting timestamps from object metadata. Defaults to `None`.
+        #[serde(default)]
+        pub timestamp_key: Option<String>,
 
-    /// Tags to apply on the compressed archives.
-    pub tags: Option<Vec<String>>,
+        /// Whether to treat the ingested objects as unstructured logs. Defaults to `false`.
+        #[serde(default = "default_unstructured")]
+        pub unstructured: bool,
+
+        /// Tags to apply on the compressed archives. Defaults to `None`.
+        #[serde(default)]
+        pub tags: Option<Vec<String>>,
+    }
+
+    /// Configuration for a SQS listener job.
+    #[derive(Debug, Clone, Serialize, Deserialize)]
+    pub struct SqsListenerConfig {
+        #[serde(flatten)]
+        pub base: BaseConfig,
+
+        /// The SQS queue URL to poll for S3 event notifications. The given queue must be dedicated
+        /// to this ingestion job.
+        pub queue_url: String,
+    }
+
+    /// Configuration for a S3 scanner job.
+    #[derive(Debug, Clone, Serialize, Deserialize)]
+    pub struct S3ScannerConfig {
+        #[serde(flatten)]
+        pub base: BaseConfig,
+
+        /// The scan interval in seconds. Defaults to 30 seconds.
+        #[serde(default = "default_scanning_interval_sec")]
+        pub scanning_interval_sec: u32,
+
+        /// The key to ingest after. If specified, only objects with keys lexicographically greater
+        /// than this value will be ingested. Defaults to `None`.
+        #[serde(default)]
+        pub start_after: Option<String>,
+    }
+
+    const fn default_unstructured() -> bool {
+        false
+    }
+
+    const fn default_scanning_interval_sec() -> u32 {
+        30
+    }
 }
--- a/components/core/tools/scripts/utils/build-and-run-unit-tests.py
+++ b/components/core/tools/scripts/utils/build-and-run-unit-tests.py
@@ -1,9 +1,13 @@
+#!/usr/bin/env python3
+"""Builds the CLP-core's binaries and runs its unit tests."""
+
+from __future__ import annotations
+
 import argparse
 import logging
 import subprocess
 import sys
 from pathlib import Path
-from typing import List, Optional
 
 # Set up console logging
 logging_console_handler = logging.StreamHandler()
@@ -21,7 +25,7 @@ root_logger.addHandler(logging_console_handler)
 logger = logging.getLogger(__name__)
 
 
-def _config_cmake_project(src_dir: Path, build_dir: Path, use_shared_libs: bool):
+def _config_cmake_project(src_dir: Path, build_dir: Path, use_shared_libs: bool) -> None:
     cmd = [
         "cmake",
         "-S",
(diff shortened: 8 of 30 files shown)


integration-tests/tests/utils/docker_utils.py
```diff
@@ -0,0 +1,48 @@
...
+    Finds the absolute path to the Docker client in the current PATH.
+
+    :return: Absolute path to the Docker binary.
+    :raise RuntimeError: docker is not found on PATH.
+    """
+    docker_bin = shutil.which("docker")
+    if docker_bin is None:
+        err_msg = "docker not found in PATH"
+        raise RuntimeError(err_msg)
+    return docker_bin
```
Bill-hbrhbr: Let's make this a generic `shutil.which` wrapper that can be run on any binary.

integration-tests/tests/utils/docker_utils.py
```diff
@@ -0,0 +1,48 @@
...
+
+    compose_ps_cmd = [
+        docker_bin,
+        "compose",
+        "--project-name",
+        project_name,
+        "ps",
+        "--format",
+        "{{.Name}}",
+    ]
```
Bill-hbrhbr, line 24:
```suggestion
    # fmt: off
    compose_ps_cmd = [
        docker_bin,
        "compose",
        "--project-name", project_name,
        "ps",
        "--format", "{{.Service}}",
    ]
    # fmt: on
```
Let's use service here so that we don't rely on the custom container names.

integration-tests/tests/utils/docker_utils.py
```diff
@@ -0,0 +1,48 @@
...
+        raise RuntimeError(err_msg)
+    return docker_bin
+
+
+def list_running_containers_in_compose_project(project_name: str) -> list[str]:
+    """
+    Lists running Docker containers that belong to the given Docker Compose project.
+
+    :param project_name:
+    :return: List of the names of the running containers that belong to the compose project.
```
Bill-hbrhbr:
```suggestion
    :param project_name:
    :return: List of the running services that belong to the specified Docker Compose project.
```

integration-tests/tests/utils/docker_utils.py
```diff
@@ -0,0 +1,48 @@
...
+
+    compose_ps_proc = subprocess.run(compose_ps_cmd, stdout=subprocess.PIPE, text=True, check=True)
+
+    container_names: list[str] = []
+    for line in (compose_ps_proc.stdout or "").splitlines():
+        container_name_candidate = line.strip()
+        if container_name_candidate:
+            container_names.append(container_name_candidate)
+
+    return container_names
```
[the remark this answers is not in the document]
Bill-hbrhbr: we want pytest to handle the stderr directly. We already have check=True

integration-tests/tests/utils/docker_utils.py
```diff
@@ -0,0 +1,48 @@
...
+    :raise RuntimeError: docker is not found on PATH.
+    """
+    docker_bin = shutil.which("docker")
+    if docker_bin is None:
+        err_msg = "docker not found in PATH"
+        raise RuntimeError(err_msg)
+    return docker_bin
+
+
+def list_running_containers_in_compose_project(project_name: str) -> list[str]:
```
Bill-hbrhbr:
```suggestion
def list_compost_project_running_services(project_name: str) -> list[str]:
```
Basically, let's change all concepts of container names to service. I won't post suggested fix everywhere so as not to clutter the review space with all the variable replacement stuff.

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -28,3 +32,43 @@ def run_and_assert(cmd: list[str], **kwargs: Any) -> subprocess.CompletedProcess
...
+    expected set of component containers.
+
+    :param package_instance:
+    """
+    # Get list of containers currently running in the Compose project.
+    instance_id = package_instance.clp_instance_id
+    project_name = f"clp-package-{instance_id}"
+    running_containers = list_running_containers_in_compose_project(project_name)
+
+    # Construct list of unique running component names and sort alphabetically.
```
Bill-hbrhbr: The regex operations can be avoided if we use `{{.Service}}` instead of `{{.Name}}`. We would just need to map service names to component names, which I believe we already have a helper for in the codebase.

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -28,3 +32,43 @@ def run_and_assert(cmd: list[str], **kwargs: Any) -> subprocess.CompletedProcess
...
+        component_name = strip_prefix(container_name, prefix)
+        component_name = strip_regex_suffix(component_name, regex_suffix)
+        if component_name not in running_components:
+            running_components.append(component_name)
+    running_components = sorted(running_components)
+
+    # Get list of required components and sort alphabetically.
+    required_components = sorted(package_instance.package_config.component_list)
+
+    if required_components != running_components:
```
Bill-hbrhbr: Let's do an equality comparison and exit early if there's no error message to report (for less indentations overall).

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -28,3 +32,43 @@ def run_and_assert(cmd: list[str], **kwargs: Any) -> subprocess.CompletedProcess
...
+        fail_msg = "Component mismatch."
+
+        missing_components = set(required_components) - set(running_components)
```
Bill-hbrhbr: Since we are already using set for more informative error logs, we should also perform the comparisons using sets instead of trying to sort them.

integration-tests/tests/utils/asserting_utils.py
quinntaylormitchell: very good point, that will simplify things.

integration-tests/tests/utils/asserting_utils.py
quinntaylormitchell: agreed.

integration-tests/tests/utils/asserting_utils.py
quinntaylormitchell: agreed.

integration-tests/tests/utils/docker_utils.py
quinntaylormitchell: agreed

integration-tests/tests/utils/docker_utils.py
quinntaylormitchell: agreed.

integration-tests/tests/utils/docker_utils.py
quinntaylormitchell: agreed

integration-tests/tests/utils/docker_utils.py
quinntaylormitchell: agreed.

integration-tests/tests/test_package_start.py
```diff
@@ -20,13 +23,5 @@ def test_clp_package(fixt_package_instance: PackageInstance) -> None:
 
     :param fixt_package_instance:
     """
-    # TODO: write code that properly validates that the package is running. This is a placeholder.
-    mode_name = fixt_package_instance.package_config.mode_name
-    logger.info("The '%s' package has been spun up successfully.", mode_name)
-
-    component_list = fixt_package_instance.package_config.component_list
-    logger.info(
```
Bill-hbrhbr, line 28: remove this redundant variable (outside line diff range).

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -28,3 +31,33 @@ def run_and_assert(cmd: list[str], **kwargs: Any) -> subprocess.CompletedProcess
...
         pytest.fail(f"Command timed out: {' '.join(cmd)}: {e}")
     return proc
+
+
+def validate_package_running(package_instance: PackageInstance) -> None:
+    """
+    Validate that the given package instance is running by checking that the set of services running
+    in the Compose project exactly matches the list of required components.
+
+    :param package_instance:
```
Bill-hbrhbr, line 41:
```suggestion
    :param package_instance:
    :raise pytest.fail: if the package instance’s running services differs from the list of required components.
```

integration-tests/tests/utils/utils.py
```diff
@@ -8,6 +8,20 @@
 from typing import IO
 
 
+def get_binary_path(name: str) -> str:
+    """
+    Finds the absolute path to the program binary called `name` in the current PATH.
+
+    :return: Absolute path to the program binary.
+    :raise RuntimeError: `name` is not found on PATH.
+    """
```
Bill-hbrhbr, line 16:
```suggestion
    """
    :param name: Name of the program binary to locate.
    :return: Absolute path to the program binary.
    :raise RuntimeError: if the program binary is not found on $PATH.
    """
```

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -28,3 +31,33 @@ def run_and_assert(cmd: list[str], **kwargs: Any) -> subprocess.CompletedProcess
...
+
+    fail_msg = "Component mismatch."
+
+    missing_components = set(required_components) - set(running_services)
+    if missing_components:
+        fail_msg += f" Missing components: {missing_components}."
+
+    unexpected_components = set(running_services) - set(required_components)
+    if unexpected_components:
+        fail_msg += f" Unexpected services: {unexpected_components}."
```
```suggestion
    running_services = set(list_running_services_in_compose_project(project_name))

    # Compare with list of required components.
    required_components = set(package_instance.package_config.component_list)
    if required_components == running_services:
        return

    fail_msg = "Component mismatch."

    missing_components = required_components - running_services
    if missing_components:
        fail_msg += f"\nMissing components: {missing_components}."

    unexpected_components = running_services - required_components
    if unexpected_components:
        fail_msg += f"\nUnexpected services: {unexpected_components}."
```
let's avoid recomputing sets and also add newlines for the potentially long fail message.

integration-tests/tests/utils/config.py
```diff
@@ -2,21 +2,22 @@
...
+    ClpConfig,
+)
 
 from tests.utils.utils import (
     unlink,
     validate_dir_exists,
     validate_file_exists,
 )
 
-if TYPE_CHECKING:
```
Bill-hbrhbr, line 17: I was talking about this `TYPE_CHECKING`. Potentially we need to restore it for linting?

Bill-hbrhbr: On 2nd thought, maybe it's due to the way you restructured the code.

integration-tests/tests/utils/asserting_utils.py
quinntaylormitchell: agreed.

integration-tests/tests/utils/asserting_utils.py
quinntaylormitchell: agreed.

integration-tests/tests/utils/config.py
quinntaylormitchell: oh right, this one. yeah when there's more than one thing imported from `clp_py_utils.clp_config`, then the linting warning goes away for some reason.

integration-tests/tests/utils/utils.py
quinntaylormitchell: agreed

integration-tests/tests/test_package_start.py
quinntaylormitchell: which variable?

Bill-hbrhbr: `logger`

integration-tests/tests/utils/docker_utils.py
```diff
@@ -0,0 +1,35 @@
...
+    Lists running Docker services that belong to the given Docker Compose project.
+
+    :param project_name:
+    :return: List of the running services that belong to the specified Docker Compose project.
+    """
+    docker_bin = get_binary_path("docker")
+
+    # fmt: off
+    compose_ps_cmd = [
+        docker_bin,
```
junhaoliao, line 19: can we simply write
```suggestion
    # fmt: off
    compose_ps_cmd = [
        "/usr/bin/env",
        "docker"
```
if the docker_bin absolute path isn't consumed (e.g. logged) anywhere?

quinntaylormitchell: it doesn't really seem like there's a clear benefit to this change; plus, using `get_binary_path` gives a clear error message when Docker is not on PATH, so eliminating it would remove that. @junhaoliao could you elaborate?

junhaoliao: ah that was more of a question, cuz i don't know the background of adding `get_binary_path()`

i don't really have a preference. practically, i feel they do the same thing - look up the environment to find an executable in `PATH` and run it.

if you ask me to list some benefits for the approaches -

For `env`:
- `env` resolves the binary at exec time, not before.
- With `env`, we get an `/usr/bin/env: ‘docker’: No such file or directory` OS error and don't need to manually check the return value of `shutil.which()`. in a sense, we don't reinvent the wheels

For `get_binary_path()`:
- we can have the error wrapped in a Python exception and customize the error message

are there things i missed? to be clear, i'm fine keeping `get_binary_path` if we want an explicit Python RuntimeException

quinntaylormitchell: resolved offline

junhaoliao: a summary for record keeping:

* `get_binary_path` was added in an effort to "suppress" ruff S607. it is considered a suppression because we're still getting the binary from the exec environment.
* Using binaries from the environment is fine in the test code (, until the day we're targeted by some supply chain attacks)
* The `get_binary_path` was decided to be kept for this specific reason: we can have the error wrapped in a Python exception and customize the error message

PR #1660, opened 2025-11-25
feat(integration-tests): Validate generated package service configuration integrity.

quinntaylormitchell:
# Description

This PR introduces the `validate_running_mode_correct` function to read the shared config file (`clp-package/var/log/.clp-config.yaml`, created when the package spins up) and ensure that the package is running in the correct mode. 

`load_yaml_to_dict` loads the `yaml` in the shared config file into a dictionary, which is then converted to a `ClpConfig` object with `ClpConfig.model_validate`. The "signature" of this `ClpConfig` object is compared to the signature of the `ClpConfig` object stored in the `PackageConfig` object for the test run with `compare_mode_signatures`.

### PR dependencies

This PR depends on the following being merged:
* #1437 
* #1658 

# Checklist

  breaking change.

# Validation performed

```bash
uv run pytest -m 'package'
```

Both the `clp-json` and `clp-text` tests passed.

## Summary by CodeRabbit

* **Tests**
  * Strengthened integration tests with a reusable validator that loads and validates the shared YAML, compares running vs intended mode signatures, and fails on mismatches.
  * Added YAML-loading support for test configurations with robust parsing and top-level structure validation.

* **Chores**
  * Test setup now discovers and validates the shared configuration file path used during test runs.

### DIFF
--- a/components/clp-package-utils/clp_package_utils/scripts/native/compress.py
+++ b/components/clp-package-utils/clp_package_utils/scripts/native/compress.py
@@ -172,9 +172,12 @@ def _generate_clp_io_config(
     urls = logs_to_compress[1:]
 
     if s3_compress_subcommand == S3_OBJECT_COMPRESSION:
-        region_code, bucket, key_prefix, keys = _parse_and_validate_s3_object_urls(urls)
+        endpoint_url, region_code, bucket, key_prefix, keys = _parse_and_validate_s3_object_urls(
+            urls
+        )
         return S3InputConfig(
             dataset=parsed_args.dataset,
+            endpoint_url=endpoint_url,
             region_code=region_code,
             bucket=bucket,
             key_prefix=key_prefix,
@@ -188,9 +191,10 @@ def _generate_clp_io_config(
             raise ValueError(
                 f"`{S3_KEY_PREFIX_COMPRESSION}` requires exactly one URL, got {len(urls)}"
             )
-        region_code, bucket, key_prefix = parse_s3_url(urls[0])
+        endpoint_url, region_code, bucket, key_prefix = parse_s3_url(urls[0])
         return S3InputConfig(
             dataset=parsed_args.dataset,
+            endpoint_url=endpoint_url,
             region_code=region_code,
             bucket=bucket,
             key_prefix=key_prefix,
@@ -223,7 +227,7 @@ def _get_logs_to_compress(logs_list_path: pathlib.Path) -> list[str]:
 
 def _parse_and_validate_s3_object_urls(
     urls: list[str],
-) -> tuple[str, str, str, list[str]]:
+) -> tuple[str | None, str | None, str, str, list[str]]:
     """
     Parses and validates S3 object URLs.
 
@@ -234,6 +238,7 @@ def _parse_and_validate_s3_object_urls(
 
     :param urls:
     :return: A tuple containing:
+        - The endpoint url.
         - The region code.
         - The bucket.
         - The common key prefix.
@@ -243,24 +248,25 @@ def _parse_and_validate_s3_object_urls(
     if len(urls) == 0:
         raise ValueError("No URLs provided.")
 
-    region_code: str | None = None
-    bucket_name: str | None = None
-    keys = set()
+    endpoint_url, region_code, bucket_name, key = parse_s3_url(urls[0])
+    keys = {key}
 
-    for url in urls:
-        parsed_region_code, parsed_bucket_name, key = parse_s3_url(url)
+    for url in urls[1:]:
+        parsed_endpoint_url, parsed_region_code, parsed_bucket_name, key = parse_s3_url(url)
 
-        if region_code is None:
-            region_code = parsed_region_code
-        elif region_code != parsed_region_code:
+        if endpoint_url != parsed_endpoint_url:
+            raise ValueError(
+                "All S3 URLs must have the same endpoint."
+                f" Found {endpoint_url} and {parsed_endpoint_url}."
+            )
+
+        if region_code != parsed_region_code:
             raise ValueError(
                 "All S3 URLs must be in the same region."
                 f" Found {region_code} and {parsed_region_code}."
             )
 
-        if bucket_name is None:
-            bucket_name = parsed_bucket_name
-        elif bucket_name != parsed_bucket_name:
+        if bucket_name != parsed_bucket_name:
             raise ValueError(
                 "All S3 URLs must be in the same bucket."
                 f" Found {bucket_name} and {parsed_bucket_name}."
@@ -276,7 +282,7 @@ def _parse_and_validate_s3_object_urls(
     if len(key_prefix) == 0:
         raise ValueError("The given S3 URLs have no common prefix.")
 
-    return region_code, bucket_name, key_prefix, key_list
+    return endpoint_url, region_code, bucket_name, key_prefix, key_list
 
 
 def _get_aws_authentication_from_config(clp_config: ClpConfig) -> AwsAuthentication:
--- a/components/clp-package-utils/clp_package_utils/scripts/native/dataset_manager.py
+++ b/components/clp-package-utils/clp_package_utils/scripts/native/dataset_manager.py
@@ -141,6 +141,7 @@ def _try_deleting_archives_from_s3(s3_config: S3Config, archive_storage_key_pref
         archive_storage_key_prefix += "/"
 
     s3_delete_by_key_prefix(
+        s3_config.endpoint_url,
         s3_config.region_code,
         s3_config.bucket,
         archive_storage_key_prefix,
--- a/components/clp-py-utils/clp_py_utils/clp_config.py
+++ b/components/clp-py-utils/clp_py_utils/clp_config.py
@@ -588,7 +588,8 @@ class AwsAuthentication(BaseModel):
 
 
 class S3Config(BaseModel):
-    region_code: NonEmptyStr
+    endpoint_url: str | None = None
+    region_code: NonEmptyStr | None = None
     bucket: NonEmptyStr
     key_prefix: str
     aws_authentication: AwsAuthentication
--- a/components/clp-py-utils/clp_py_utils/s3_utils.py
+++ b/components/clp-py-utils/clp_py_utils/s3_utils.py
@@ -36,6 +36,13 @@ AWS_ENV_VAR_SESSION_TOKEN: Final[str] = "AWS_SESSION_TOKEN"
 
 S3_OBJECT_DELETION_BATCH_SIZE_MAX: Final[int] = 1000
 
+SCHEME_REGEXP = r"(?P<scheme>(http|https))"
+S3_PREFIX_REGEXP = r"(?P<s3>s3)"
+ENDPOINT_REGEXP = r"(?P<endpoint>[a-z0-9.-]+(\:[0-9]+)?)"
+REGION_CODE_REGEXP = r"(?P<region_code>[a-z0-9\-]+)"
+BUCKET_NAME_REGEXP = r"(?P<bucket_name>[a-z0-9.-]+)"
+KEY_PREFIX_REGEXP = r"(?P<key_prefix>[^?]+)"
+
 
 def _get_session_credentials(aws_profile: str | None = None) -> S3Credentials | None:
     """
@@ -180,7 +187,10 @@ def generate_container_auth_options(
 
 
 def _create_s3_client(
-    region_code: str, s3_auth: AwsAuthentication, boto3_config: Config | None = None
+    endpoint_url: str | None,
+    region_code: str | None,
+    s3_auth: AwsAuthentication,
+    boto3_config: Config | None = None,
 ) -> boto3.client:
     aws_session: boto3.Session | None
     if AwsAuthType.profile == s3_auth.type:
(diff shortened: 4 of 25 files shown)


integration-tests/tests/utils/asserting_utils.py
```diff
@@ -28,3 +36,32 @@ def run_and_assert(cmd: list[str], **kwargs: Any) -> subprocess.CompletedProcess
...
+
+def validate_running_mode_correct(package_instance: PackageInstance) -> None:
+    """
+    Validate that the mode described in the shared config of the instance matches the intended mode
+    defined by the instance configuration. Calls pytest.fail if the shared config fails validation
+    or if the running mode does not match the intended mode.
+
+    :param package_instance:
+    :raise: Propagates `load_yaml_to_dict`'s errors.
+    :raise ValidationError: if the ClpConfig object cannot be validated.
```
```suggestion
    :raise pytest.fail: if the ClpConfig object cannot be validated.
```

integration-tests/tests/utils/clp_mode_utils.py
```diff
@@ -74,6 +75,26 @@ def _to_docker_compose_service_name(name: str) -> str:
 CLP_MCP_SERVER_COMPONENT = _to_docker_compose_service_name(MCP_SERVER_COMPONENT_NAME)
 
 
+def compute_mode_signature(config: ClpConfig) -> tuple[Any, ...]:
+    """
+    Constructs a signature that captures the mode-defining aspects of a ClpConfig object.
```
Bill-hbrhbr: Maybe add a secondary function description saying what's not included in the **mode-defining aspects**. I can think of port numbers.

quinntaylormitchell: I think it would be a much shorter list to say what _is_ included in the check, cause there are a lot of things about the `ClpConfig` object that we aren't checking for.

That being said, the check in this function might be pretty flimsy to maintain in the future cause it relies on what are essentially "magic" qualities, i.e. there won't be any warning (in the code) when the "mode-defining aspects" are redefined. My reason for making the function is that there's no direct comparison possible for `ClpConfig` objects. Thoughts?

quinntaylormitchell:
@Bill-hbrhbr conflict merge complete

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -6,9 +6,15 @@
 from typing import Any
 
 import pytest
+from clp_py_utils.clp_config import ClpConfig
+from pydantic import ValidationError
 
+from tests.utils.clp_mode_utils import (
+    compute_mode_signature,
+)
```
```suggestion
from tests.utils.clp_mode_utils import compute_mode_signature
```

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -62,3 +68,32 @@ def validate_package_running(package_instance: PackageInstance) -> None:
...
+
+def validate_running_mode_correct(package_instance: PackageInstance) -> None:
+    """
+    Validate that the mode described in the shared config of the instance matches the intended mode
+    defined by the instance configuration. Calls pytest.fail if the shared config fails validation
+    or if the running mode does not match the intended mode.
+
+    :param package_instance:
+    :raise: Propagates `load_yaml_to_dict`'s errors.
+    :raise pytest.fail: if the ClpConfig object cannot be validated.
```
Bill-hbrhbr, line 79:
```suggestion
    :raise pytest.fail: if the ClpConfig object cannot be validated.
    :raise pytest.fail: if the running ClpConfig does not match the intended ClpConfig.
```
sorry, missed one failure scenario.

integration-tests/tests/utils/clp_mode_utils.py
Bill-hbrhbr: I don't have a strong opinion. Probably should ask for someone else's opinion.

quinntaylormitchell: @junhaoliao, when you review this, could you share your thoughts on this?

junhaoliao: as discussed offline, we should avoid writing the signature generator in a way that any modification of the `ClpConfig` class definition would require modification of the generator.

Instead, can we do equality matches on the config dictionaries? If there are certain patterns like the port numbers we want to exclude, can we try something like this?
```py
def check_dicts_equal(
    left: dict,
    right: dict,
    exclude_key_patterns: list[str]
) -> bool:
    """
    Check if two dictionaries are equal, excluding keys matching given patterns.
    
    :param left: First dictionary to compare
    :param right: Second dictionary to compare
    :param exclude_key_patterns: Substrings to match against keys (case-insensitive).
                                 Keys containing any of these patterns will be excluded
                                 from comparison.
    :return: True if dictionaries are equal after removing matching keys
    """
    return _filter_keys_recursive(left, exclude_key_patterns) == _filter_keys_recursive(right, exclude_key_patterns)


def _filter_keys_recursive(obj: dict | list | any, exclude_key_patterns: list[str]) -> dict | list | any:
    """
    Recursively remove dictionary keys matching any of the given patterns.
    
    :param obj:
    :param exclude_key_patterns: Substrings to match against keys (case-insensitive)
    :return: Filtered copy of the object with matching keys removed
    """
    def _should_exclude_key(key: str) -> bool:
        """Check if a key matches any exclusion pattern."""
        key_lower = key.lower()
        return any(pattern.lower() in key_lower for pattern in exclude_key_patterns)

    if isinstance(obj, dict):
        return {
            key: _filter_keys_recursive(value, exclude_key_patterns)
            for key, value in obj.items()
            if not _should_exclude_key(key)
        }
    elif isinstance(obj, list):
        return [_filter_keys_recursive(item, exclude_key_patterns) for item in obj]
    else:
        return obj
```

integration-tests/tests/utils/clp_mode_utils.py
```diff
@@ -93,25 +103,35 @@ def get_required_component_list(config: ClpConfig) -> list[str]:
...
+def _match_objects_by_explicit_fields(intended_obj: Any, running_obj: Any) -> bool:
+    """
+    Compares `intended_obj` and `running_obj` using Pydantic's `model_fields_set` to recursively
+    match only the fields that were explicitly set when `intended_obj` was initialized.
+
+    When both objects are Pydantic models, the function only inspects fields that were explicitly
+    set on the `intended_obj`. For other data types, the function uses standard equality.
+    """
+    if isinstance(intended_obj, BaseModel):
+        for field_name in intended_obj.model_fields_set:
```
junhaoliao, line 135: otherwise the `getattr(running_obj, field_name)` below may thrown an AttributeError
```suggestion
        if not isinstance(running_obj, BaseModel):
            return False
        for field_name in intended_obj.model_fields_set:
```

integration-tests/tests/utils/utils.py
```diff
@@ -64,6 +66,33 @@ def is_json_file_structurally_equal(json_fp1: Path, json_fp2: Path) -> bool:
         return is_dir_tree_content_equal(Path(temp_file_1.name), Path(temp_file_2.name))
 
 
+def load_yaml_to_dict(path: Path) -> dict[str, Any]:
+    """
+    Parses a UTF-8 YAML file into a dictionary.
```
junhaoliao:
```suggestion
    Parses a YAML file into a dictionary.
```

i feel the encoding is less important given that's implementation-specific and the de-facto standard. also, Python 3.15 will change the default encoding to "utf-8" as well: https://docs.python.org/3.15/whatsnew/3.15.html#whatsnew315-utf8-default

integration-tests/tests/utils/config.py
```diff
@@ -169,8 +170,13 @@ class PackageInstance:
     #: The instance ID of the running package.
     clp_instance_id: str = field(init=False, repr=True)
 
+    #: The path to the .clp-config.yaml file constructed by the package during spin up.
+    shared_config_file_path: Path = field(init=False, repr=True)
+
     def __post_init__(self) -> None:
         """Validates init values and initializes attributes."""
```
junhaoliao, line 177: seems duplicated byL184

integration-tests/tests/utils/clp_mode_utils.py
```diff
@@ -93,25 +103,35 @@ def get_required_component_list(config: ClpConfig) -> list[str]:
...
 
-    return component_list
+def _match_objects_by_explicit_fields(intended_obj: Any, running_obj: Any) -> bool:
+    """
+    Compares `intended_obj` and `running_obj` using Pydantic's `model_fields_set` to recursively
+    match only the fields that were explicitly set when `intended_obj` was initialized.
+
+    When both objects are Pydantic models, the function only inspects fields that were explicitly
+    set on the `intended_obj`. For other data types, the function uses standard equality.
+    """
```
junhaoliao, line 130:
```suggestion
    """
    Compares `intended_obj` and `running_obj` using Pydantic's `model_fields_set` to recursively
    match only the fields that were explicitly set when `intended_obj` was initialized.

    When both objects are Pydantic models, the function only inspects fields that were explicitly
    set on the `intended_obj`. For other data types, the function uses standard equality.

    :param intended_obj:
    :param running_obj:
    :return: True if all explicitly set fields in intended_obj match running_obj, False otherwise.
    """
```

integration-tests/tests/utils/clp_mode_utils.py
```diff
@@ -62,16 +64,24 @@ def _to_docker_compose_service_name(name: str) -> str:
...
-CLP_MCP_SERVER_COMPONENT = _to_docker_compose_service_name(MCP_SERVER_COMPONENT_NAME)
+
+
+def compare_mode_signatures(intended_config: ClpConfig, running_config: ClpConfig) -> bool:
+    """
+    Compares the running signatures of `intended_config` and `running_config` with
+    `_match_objects_by_explicit_fields`.
+
+    :param intended_config:
+    :param running_config:
```
junhaoliao:
```suggestion
    :param intended_config: The reference config object.
    :param running_config: The config object to compare against.
```

integration-tests/tests/utils/utils.py
```diff
@@ -64,6 +66,33 @@ def is_json_file_structurally_equal(json_fp1: Path, json_fp2: Path) -> bool:
...
+            target_dict = yaml.safe_load(file)
+    except yaml.YAMLError as err:
+        err_msg = f"Invalid YAML in target file '{path}'"
+        raise ValueError(err_msg) from err
+    except OSError as err:
+        err_msg = f"Cannot read target file '{path}'"
+        raise ValueError(err_msg) from err
+
+    if not isinstance(target_dict, dict):
+        err_msg = f"Target file {path} must have a top-level mapping."
```
junhaoliao:
```suggestion
        err_msg = f"Target file '{path}' must have a top-level mapping."
```

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -62,3 +66,26 @@ def validate_package_running(package_instance: PackageInstance) -> None:
...
+    shared_config_dict = load_yaml_to_dict(package_instance.shared_config_file_path)
+    try:
+        running_config = ClpConfig.model_validate(shared_config_dict)
+    except ValidationError as err:
+        pytest.fail(f"Shared config failed validation: {err}")
+
+    intended_config = package_instance.package_config.clp_config
+
+    if not compare_mode_signatures(intended_config, running_config):
+        pytest.fail("Mode mismatch: running configuration does not match intended configuration.\n")
```
junhaoliao:
```suggestion
        pytest.fail("Mode mismatch: running configuration does not match intended configuration.")
```

unless there are display issues without it?

quinntaylormitchell: agreed

integration-tests/tests/utils/clp_mode_utils.py
quinntaylormitchell: agreed

integration-tests/tests/utils/clp_mode_utils.py
quinntaylormitchell: agreed

integration-tests/tests/utils/clp_mode_utils.py
quinntaylormitchell: agreed

integration-tests/tests/utils/config.py
quinntaylormitchell: good catch thanks!

integration-tests/tests/utils/utils.py
quinntaylormitchell: agreed

integration-tests/tests/utils/utils.py
quinntaylormitchell: agreed

Bill-hbrhbr:
@quinntaylormitchell @junhaoliao 
Not sure how I would polish the title, my way of phrasing it would be:
```
feat(integration-tests): Validate the CLP configuration integrity at runtime.
```
or something in the lines of:
```
feat(integration-tests): Validate the CLP package is deployed with the intended configuration.
```

junhaoliao:
> @quinntaylormitchell @junhaoliao Not sure how I would polish the title, my way of phrasing it would be:
> 
> ```
> feat(integration-tests): Validate the CLP configuration integrity at runtime.
> ```
> 
> or something in the lines of:
> 
> ```
> feat(integration-tests): Validate the CLP package is deployed with the intended configuration.
> ```

how about
```
feat(integration-tests): Validate generated package service configuration integrity.
```

Issue #1657, opened 2025-11-25
feat: `clp-s` ordered compression and decompression identity transformation integration tests.


Bill-hbrhbr:
### Request

This issue tracks the changes needed to support a complete `clp-s` identity transformation testing workflow under both ordered and unordered modes.

### Motivation

The primary motivation is to introduce stronger and more explicit assertions for `clp-s` identity transformation tests. `clp-s` compression [does not preserve the input directory structure](https://docs.yscope.com/clp/v0.6.0/user-docs/core-clp-s.html#current-limitations), but it is able to preserve the log ordering.

### End result

Testing compression and decompression without data loss in the following two ways:

* Run `clp-s c` with defaults and `clp-s x` with `--ordered`
* Run `clp-s c` with `--disable-log-order` and `clp-s x` with defaults

### Implementation Plan

1. Remove the potentially redundant clp-s compression and decompression merge step

For comparison purposes, both the input and output of `clp-s` workflow must be a single JSON file. However, some downloaded datasets contain split JSON logs.

The current approach merges them by running an additional cycle of `clp-s` compression and decompression.
- This is inefficient and should be replaced by a simple merge operation that follows the same log ingestion order used internally by `clp-s`.
- This merge step should be skipped entirely when the downloaded logs are already a single JSON file.

2. Strip the top level directory when extracting downloaded dataset tarballs.

For example, `build/integration-tests/postgresql.tar.gz` currently extracts into `build/integration-tests/postgresql/postgresql/postgresql.log`, introducing an unnecessary extra directory layer. Stripping this top level layer enables the tests to accurately detect when a dataset contains a single log file, which is required for the merge skip logic described in step 1.

3. Add a dedicated assertion for `clp-s` decompression producing a single file.

Since `clp-s` does not preserve the input directory structure, and the default flag value `--target-ordered-chunk-size=0` guarantees a single output file, the tests should include an explicit assertion verifying this condition. 

The output file may also have a uniquely generated UUID name that differs across test runs (see step 4), so the assertion must focus on cardinality rather than filename.

4. Locate the output filename dynamically for comparison.

The current implementation assumes that the single output file described in step 3 will always be named `original`, which only holds when `clp-s x` is run with default flags. When `--ordered` is provided during decompression, the output filename becomes a UUID. The comparison logic must therefore be updated to locate the single output file dynamically rather than relying on a fixed name.

5. Extend the JSON sorting workflow to support ordered comparisons.

[...]

PR #1661, opened 2025-11-25
feat(integration-tests): Extract tarballs with `tar` to strip top-level directory; Fix path cleanup utilities.

Bill-hbrhbr:
# Description

This PR belongs to the integration-tests feature plan described in #1657.

This change replaces the previous archive extraction workflow, which used `shutil.unpack_archive`, with an explicit `tar` based implementation. The new approach correctly handles stripping a leading directory component when the tarball includes a top level folder. For example, the old workflow extracted `build/integration-tests/postgresql.tar.gz` into:
```
build/integration-tests/postgresql/postgresql/postgresql.log
``` 
which adds an unnecessary extra level of nesting due to the preserved top level directory inside the archive. Since we wish to control the extraction directory name, we explicitly remove that leading component. The extraction path becomes:
```
build/integration-tests/postgresql/postgresql.log
```

Because `shutil.unpack_archive` does not support stripping directory components, switching to `tar` provides the required extraction results. The chosen extraction flags remain compatible with both macOS and Linux.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **Chores**
  * Enhanced internal testing infrastructure with improved reliability and robustness. Modernized dataset extraction workflows with better error handling mechanisms. Implemented flexible options for directory component management during extraction. Updated permission handling to ensure consistent access across extraction operations.

---
  - https://app.asana.com/0/0/1212888482202074

### DIFF
--- a/components/clp-rust-utils/src/database/mysql.rs
+++ b/components/clp-rust-utils/src/database/mysql.rs
@@ -6,6 +6,28 @@ use crate::clp_config::package::{
     credentials::Database as DatabaseCredentials,
 };
 
+/// Implements [`sqlx::Type<sqlx::MySql>`] for `$ty` by delegating to `$delegate`.
+///
+/// # Examples
+///
+/// ```rust
+/// impl_sqlx_type!(IngestedS3ObjectMetadataStatus => str);
+/// ```
+#[macro_export]
+macro_rules! impl_sqlx_type {
+    ($ty:ty => $delegate:ty $(,)?) => {
+        impl ::sqlx::Type<::sqlx::MySql> for $ty {
+            fn type_info() -> <::sqlx::MySql as ::sqlx::Database>::TypeInfo {
+                <$delegate as ::sqlx::Type<::sqlx::MySql>>::type_info()
+            }
+
+            fn compatible(ty: &<::sqlx::MySql as ::sqlx::Database>::TypeInfo) -> bool {
+                <$delegate as ::sqlx::Type<::sqlx::MySql>>::compatible(ty)
+            }
+        }
+    };
+}
+
 /// Trait for formatting Rust enums as SQL `ENUM(...)` declarations.
 pub trait MySqlEnumFormat: IntoEnumIterator + Sized + ToString
 where
--- a/components/clp-rust-utils/src/job_config/compression.rs
+++ b/components/clp-rust-utils/src/job_config/compression.rs
@@ -1,6 +1,8 @@
 use num_enum::{IntoPrimitive, TryFromPrimitive};
 use serde::{Deserialize, Serialize};
 
+pub type CompressionJobId = i32;
+
 /// Mirror of `job_orchestration.scheduler.constants.CompressionJobStatus`. Must be kept in sync.
 #[derive(Clone, Debug, Deserialize, Eq, IntoPrimitive, PartialEq, Serialize, TryFromPrimitive)]
 #[repr(i32)]
--- a/components/clp-rust-utils/src/s3.rs
+++ b/components/clp-rust-utils/src/s3.rs
@@ -3,9 +3,13 @@ mod client;
 pub use client::create_new_client;
 use non_empty_string::NonEmptyString;
 
+/// Represents the unique identifier for an S3 object metadata entry in CLP DB.
+pub type S3ObjectMetadataId = u64;
+
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct ObjectMetadata {
     pub bucket: NonEmptyString,
     pub key: NonEmptyString,
     pub size: u64,
+    pub id: Option<S3ObjectMetadataId>,
 }
--- a/components/core/CMakeLists.txt
+++ b/components/core/CMakeLists.txt
@@ -506,6 +506,8 @@ set(SOURCE_FILES_unitTest
         src/clp/ffi/ir_stream/encoding_methods.hpp
         src/clp/ffi/ir_stream/IrErrorCode.cpp
         src/clp/ffi/ir_stream/IrErrorCode.hpp
+        src/clp/ffi/ir_stream/IrSerializationError.cpp
+        src/clp/ffi/ir_stream/IrSerializationError.hpp
         src/clp/ffi/ir_stream/IrUnitHandlerReq.hpp
         src/clp/ffi/ir_stream/IrUnitType.hpp
         src/clp/ffi/ir_stream/ir_unit_deserialization_methods.cpp
--- a/components/core/src/clp/clg/CMakeLists.txt
+++ b/components/core/src/clp/clg/CMakeLists.txt
@@ -21,6 +21,8 @@ set(
         ../ffi/ir_stream/decoding_methods.cpp
         ../ffi/ir_stream/decoding_methods.hpp
         ../ffi/ir_stream/decoding_methods.inc
+        ../ffi/ir_stream/IrSerializationError.cpp
+        ../ffi/ir_stream/IrSerializationError.hpp
         ../ffi/StringBlob.hpp
         ../FileDescriptor.cpp
         ../FileDescriptor.hpp
--- a/components/core/src/clp/clo/CMakeLists.txt
+++ b/components/core/src/clp/clo/CMakeLists.txt
@@ -27,6 +27,8 @@ set(
         ../ffi/ir_stream/decoding_methods.inc
         ../ffi/ir_stream/encoding_methods.cpp
         ../ffi/ir_stream/encoding_methods.hpp
+        ../ffi/ir_stream/IrSerializationError.cpp
+        ../ffi/ir_stream/IrSerializationError.hpp
         ../ffi/ir_stream/utils.cpp
         ../ffi/ir_stream/utils.hpp
         ../ffi/StringBlob.hpp
--- a/components/core/src/clp/clp/CMakeLists.txt
+++ b/components/core/src/clp/clp/CMakeLists.txt
@@ -28,6 +28,10 @@ set(
         ../ffi/ir_stream/decoding_methods.inc
         ../ffi/ir_stream/encoding_methods.cpp
         ../ffi/ir_stream/encoding_methods.hpp
+        ../ffi/ir_stream/IrErrorCode.cpp
+        ../ffi/ir_stream/IrErrorCode.hpp
+        ../ffi/ir_stream/IrSerializationError.cpp
+        ../ffi/ir_stream/IrSerializationError.hpp
         ../ffi/ir_stream/utils.cpp
         ../ffi/ir_stream/utils.hpp
         ../ffi/StringBlob.hpp
--- a/components/core/src/clp/ffi/ir_stream/IrErrorCode.cpp
+++ b/components/core/src/clp/ffi/ir_stream/IrErrorCode.cpp
@@ -16,13 +16,13 @@ template <>
 auto IrErrorCategory::message(IrErrorCodeEnum error_enum) const -> std::string {
     switch (error_enum) {
         case IrErrorCodeEnum::DecodingMethodFailure:
-            return "The decoding method failed.";
+            return "the decoding method failed";
         case IrErrorCodeEnum::EndOfStream:
-            return "The end-of-stream IR unit has already been consumed.";
+            return "the end-of-stream IR unit has already been consumed";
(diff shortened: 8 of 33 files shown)


integration-tests/tests/fixtures/integration_test_logs.py
```diff
@@ -48,7 +47,18 @@ def _download_and_extract_dataset(
...
+    """
+    :param request:
+    :param integration_test_path_config:
+    "param name:
+    :param tarball_url:
+    :param has_leading_directory_component: Whether all files inside the tarball are stored under a
+    single top level directory. Defaults to True.
+    :return: The IntegrationTestPathConfig object with its associated logs properly downloaded,
+    extracted, and permission changed to be overritable.
+    """
```
quinntaylormitchell:
```suggestion
    :param name:
    :param tarball_url:
    :param has_leading_directory_component: Whether all files inside the tarball are stored under a
    single top-level directory.
    :return: The IntegrationTestLogs object describing the donloaded and extracted logs.
    """
```
clarity and typo fixes.

integration-tests/tests/fixtures/integration_test_logs.py
```diff
@@ -58,38 +68,56 @@ def _download_and_extract_dataset(
         logger.info("Test logs `%s` are up-to-date. Skipping download.", name)
         return integration_test_logs
 
+    integration_test_logs.tarball_path.unlink(missing_ok=True)
+    shutil.rmtree(integration_test_logs.extraction_dir)
+    integration_test_logs.extraction_dir.mkdir(parents=True, exist_ok=False)
```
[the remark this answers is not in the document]
quinntaylormitchell: This is correct, I did get an error when I ran this directly after a fresh build. The rabbit's committable suggestion works.

quinntaylormitchell: A couple things about these three lines (in addition to the rabbit comment, which is correct):
1. I remember that in `tests.utils.utils`, you define [a function called `unlink`](https://github.com/y-scope/clp/blob/55f2c9c6ae8b62bf5e8adf16a28d0e82dcb80550/integration-tests/tests/utils/utils.py#L65); you should use it. That being said, I've realized that function doesn't actually do what it says in the docstring. It says `Remove a file or directory at `path`.`, but it only uses `shutil.rmtree`, which doesn't remove files (afaik). I think that we should do either one of two things: A. we should update that function to something like `def remove_file_or_dir`, make it do what the docstring says, and then use it everywhere in our integration tests, or B. we should remove that function entirely and use only a mixture of `shutil.rmtree` for directories and `os.unlink` for files.
2. Why don't we make `extraction_dir` in [the `IntegrationTestLogs` `__post_init__`](https://github.com/y-scope/clp/blob/55f2c9c6ae8b62bf5e8adf16a28d0e82dcb80550/integration-tests/tests/utils/config.py#L128)?

integration-tests/tests/fixtures/integration_test_logs.py
```diff
@@ -48,7 +47,18 @@ def _download_and_extract_dataset(
     integration_test_path_config: IntegrationTestPathConfig,
     name: str,
     tarball_url: str,
+    has_leading_directory_component: bool = True,
```
quinntaylormitchell: This name is a bit awkward; how about
```suggestion
    strip_leading_dir: bool = True,
```
or
```suggestion
    strip_top_dir: bool = True,
```

integration-tests/tests/fixtures/integration_test_logs.py
Bill-hbrhbr: 1. discussed in other posts.
2. I think `_download_and_extract_dataset` is a giant post init setup for `IntegrationTestLogs`, so we don't really need to put any set up into the actual `__post_init__`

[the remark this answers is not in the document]
Bill-hbrhbr: resolved by using the new helper.

integration-tests/tests/fixtures/integration_test_logs.py
Bill-hbrhbr: revamped the docstring for clarity.

integration-tests/tests/utils/utils.py
```diff
@@ -93,6 +93,23 @@ def load_yaml_to_dict(path: Path) -> dict[str, Any]:
     return target_dict
 
 
+def remove_path(path_to_remove: Path) -> None:
+    """
+    Remove a file, directory, or symlink at `path_to_remove` if it exists.
+
+    :param path_to_remove:
+    :raise: Propagates `pathlib.Path.unlink`'s exceptions.
+    :raise: Propagates `shutil.rmtree`'s exceptions.
```
quinntaylormitchell, line 102: should these `:raise:` statements go in alphabetical order (as they currently are) or in the order that they appear in the function?

integration-tests/tests/utils/utils.py
```diff
@@ -93,6 +93,23 @@ def load_yaml_to_dict(path: Path) -> dict[str, Any]:
...
+    """
+    if not path_to_remove.exists():
+        return
```
quinntaylormitchell: coderabbit left a comment on my implementation of this in #1801, not sure if you want to address it or not:

"Broken symlinks will not be removed.

Path.exists() follows symlinks and returns False for broken symlinks, causing the function to return early without removing them. This could leave stale symlinks in test directories."

integration-tests/tests/fixtures/integration_test_logs.py
```diff
@@ -58,38 +74,42 @@ def _download_and_extract_dataset(
...
+    ]
+    # fmt: on
+    if strip_leading_dir:
+        extract_cmd.extend(["--strip-components", "1"])
+    subprocess.run(extract_cmd, check=True)
+
+    # Allow the downloaded and extracted contents to be deletable or overwritable
+    chmod_bin = get_binary_path("chmod")
+    subprocess.run([chmod_bin, "-R", "gu+w", tarball_path_str], check=True)
+    subprocess.run([chmod_bin, "-R", "gu+w", extract_path_str], check=True)
```
quinntaylormitchell:
```suggestion
    subprocess.run([chmod_bin, "gu+w", tarball_path_str], check=True)
    subprocess.run([chmod_bin, "--recursive", "gu+w", extract_path_str], check=True)
```
a couple nits:
1. I don't think we need the `-R` flag in the first one because the tarball is a single file (no recursion needed)
2. for the second line, change `-R` to `--recursive` for readability

Bill-hbrhbr: Agree with the first one. For the second one, I think MacOS doesn't support a lot of long options.
https://www.unix.com/man_page/osx/1/chmod/

integration-tests/tests/utils/utils.py
Bill-hbrhbr: Let's leave it for maintainer review round. @junhaoliao

integration-tests/tests/utils/utils.py
Bill-hbrhbr: Updated my implementation.

integration-tests/tests/fixtures/integration_test_logs.py
quinntaylormitchell: ohhhh ok good to know

integration-tests/tests/utils/utils.py
junhaoliao: https://docs.yscope.com/dev-guide/contrib-guides-general.html#:~:text=Alphabetically%20%28unless%20a%20different%20ordering%20is%20required%20for%20functionality%20or%20enhanced%20clarity%29%2E

i agree the "enhanced clarity" wording leaves a lot of flexibility (/ambiguity).

since a method may throw multiple exceptions, and alphabetical ordering helps symbol look-ups when there are many items, i prefer consistently ordering alphabetically than by when they show up.

integration-tests/tests/utils/utils.py
```diff
@@ -93,6 +93,24 @@ def load_yaml_to_dict(path: Path) -> dict[str, Any]:
     return target_dict
 
 
+def remove_path(path_to_remove: Path) -> None:
```
junhaoliao, line 96: regarding `unlink()` below - Having two similarly-purposed functions with different
semantics is a maintenance hazard. It seems like `remove_path` supports more features than `unlink()` except the ability to fallback to `sudo`. Is it possible to merge the two methods and let callers consistently call only one?

otherwise, we should clearly document the distinction (e.g. cross-reference) in both docstrings so future developers know which to use

integration-tests/tests/fixtures/integration_test_logs.py
```diff
@@ -48,7 +50,21 @@ def _download_and_extract_dataset(
     integration_test_path_config: IntegrationTestPathConfig,
     name: str,
     tarball_url: str,
+    strip_leading_dir: bool = True,
 ) -> IntegrationTestLogs:
+    """
+    Download and extract a dataset tarball for setting up the `IntegrationTestLogs` fixture, and
+    adjust its file permissions for test use.
```
junhaoliao:
```suggestion
    Download and extract a dataset gzip-compressed tarball for setting up the `IntegrationTestLogs`
    fixture, and adjust its file permissions for test use.
```

integration-tests/tests/fixtures/integration_test_logs.py
```diff
@@ -58,38 +74,42 @@ def _download_and_extract_dataset(
...
+        tarball_url,
+    ]
+    # fmt: on
+    subprocess.run(curl_cmd, check=True)
+
+    # fmt: off
+    extract_cmd = [
+        get_binary_path("tar"),
+        "--extract",
+        "--gzip",
```
junhaoliao, line 100: it's kinda unfortunate that GNU tar doesn't support compression type detection (the BSD one does). see above docstring sugguestion about explicitly saying this only supports gz tarballs

integration-tests/tests/fixtures/integration_test_logs.py
```diff
@@ -58,38 +74,42 @@ def _download_and_extract_dataset(
...
+    ]
+    # fmt: on
+    if strip_leading_dir:
+        extract_cmd.extend(["--strip-components", "1"])
+    subprocess.run(extract_cmd, check=True)
+
+    # Allow the downloaded and extracted contents to be deletable or overwritable
+    chmod_bin = get_binary_path("chmod")
+    subprocess.run([chmod_bin, "gu+w", tarball_path_str], check=True)
+    subprocess.run([chmod_bin, "-R", "gu+w", extract_path_str], check=True)
```
junhaoliao, line 113: why `gu+w`? maybe it's worth documenting the reason in a docstring?

integration-tests/tests/fixtures/integration_test_logs.py
```diff
@@ -58,38 +74,42 @@ def _download_and_extract_dataset(
...
+    # fmt: on
+    subprocess.run(curl_cmd, check=True)
+
+    # fmt: off
+    extract_cmd = [
+        get_binary_path("tar"),
+        "--extract",
+        "--gzip",
+        "--file", tarball_path_str,
+        "-C", extract_path_str,
```
junhaoliao: long-form
```suggestion
        "--directory", extract_path_str,
```

integration-tests/tests/fixtures/integration_test_logs.py
```diff
@@ -48,7 +50,21 @@ def _download_and_extract_dataset(
     integration_test_path_config: IntegrationTestPathConfig,
     name: str,
     tarball_url: str,
+    strip_leading_dir: bool = True,
```
junhaoliao: it's usually better to let optional **Boolean** arguments fallback to falseful values. how about flipping the behaviour with `keep_leading_dir: bool = False`.

integration-tests/tests/utils/utils.py
Bill-hbrhbr: I'm planning on replace `unlink()` with `remove_path()` for the next PR. Injecting `sudo` commands in the codebase doesn't seem like the best idea after all.
Fyi, we run `chmod` on downloaded objects so that we don't need `sudo` when trying to change/delete them.

Bill-hbrhbr: Actually, it's not a chore to remove `unlink()`. Let me update together in this PR.

PR #1663, opened 2025-11-25
feat(api-server): Document APIs using OpenAPI and serve on `/openapi.json` endpoint.

hoophalab:
# Description

This PR adds an endpoint to server open api docs.
Move routes to lib.

# Checklist

  breaking change.

# Validation performed

Put http://localhost:3001/openapi.json on https://petstore.swagger.io/. The endpoints are shown correctly.

## Summary by CodeRabbit

* **New Features**
  * Added OpenAPI documentation at /openapi.json with improved API docs.
  * Implemented CORS support for cross-origin requests.
  * Added a health check endpoint.
  * Expanded query configuration with dataset selection, result limits, timestamp filtering, case sensitivity, and file-output options.
  * Added Server-Sent Events streaming for query results.

### DIFF
--- a/components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
+++ b/components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
@@ -998,7 +998,7 @@ async def handle_finished_stream_extraction_job(
     else:
         task_result = QueryTaskResult.model_validate(task_results[0])
         task_id = task_result.task_id
-        if not QueryJobStatus.SUCCEEDED == task_result.status:
+        if not QueryTaskStatus.SUCCEEDED == task_result.status:
             logger.error(
                 f"Extraction task job-{job_id}-task-{task_id} failed. "
                 f"Check {task_result.error_log_path} for details."


components/api-server/src/bin/api_server.rs
LinZhihao-723, line 1: This file gets a bit too overloaded now. I propose:
- Move all API calls into a dedicated submod, along with their documentation.
- Move error definition to a destinated file (since it's needed for both).

components/api-server/src/bin/api_server.rs
```diff
@@ -99,24 +101,73 @@ async fn main() -> anyhow::Result<()> {
...
+  "begin_timestamp": 0,
+  "end_timestamp": 1735689600000000,
+  "ignore_case": true,
+  "max_num_results": 0,
+  "write_to_file": false
+}"#),
+    responses(
+        (
+            status = OK,
+            body = serde_json::Value,
```
LinZhihao-723: This won't render the schema properly. Can we make a struct like `struct xxx { query_results_uri: String }` and derive `ToSchema`?

components/api-server/README.md
junhaoliao, line 1: is this / can this be auto-generated?

components/api-server/src/bin/api_server.rs
```diff
@@ -99,24 +101,73 @@ async fn main() -> anyhow::Result<()> {
...
+        content= clp_client::QueryConfig,
+        example = r#"{
+  "query_string": "*",
+  "dataset": "default",
+  "begin_timestamp": 0,
+  "end_timestamp": 1735689600000000,
+  "ignore_case": true,
+  "max_num_results": 0,
+  "write_to_file": false
+}"#),
```
```suggestion
        example = json!({
            "query_string": "*",
            "dataset": "default",
            "begin_timestamp": 0,
            "end_timestamp": 17356896,
            "ignore_case": true,
            "max_num_results": 0,
            "write_to_file": false
        }),
```
Why not doing it this way?
The embedded string won't get rendered properly:

Image

components/api-server/src/bin/api_server.rs
```diff
@@ -126,6 +177,20 @@ async fn query(
...
+#[utoipa::path(
+    get,
+    path = "/query_results/{search_job_id}",
+    responses(
+        (
+            status = OK,
+            body = String,
+            content_type = "text/event-stream",
+            example = "data: {\"timestamp\":\"2023-03-27 00:26:37.134\",\"pid\":7821,\
+\"line_num\":7,\"message\":\"CommitTransaction\"}\n"
```
```suggestion
            description = "Server-Sent Events stream of query results. Each event contains a \
                single line of the query result in JSON format.",
```
I think the example here probably won't be helpful.
- A description might be more clear, as it explicitly defines what's inside each event.
- I think the format of `data: xxx` is indicated by SSE.

components/api-server/src/bin/api_server.rs
```diff
@@ -99,24 +101,73 @@ async fn main() -> anyhow::Result<()> {
...
+    path = "/health",
+    responses((status = OK, body = String, example = "API server is running"))
+)]
 async fn health() -> String {
     "API server is running".to_owned()
 }
 
+#[utoipa::path(
+    post,
+    path = "/query",
```
```suggestion
    path = "/query",
    description = "Submits a new query job.",
```
I think it's worth to add an API level description.

components/api-server/src/bin/api_server.rs
```diff
@@ -126,6 +177,20 @@ async fn query(
     Ok(Json(serde_json::json!({"query_results_uri": uri})))
 }
 
+#[utoipa::path(
+    get,
+    path = "/query_results/{search_job_id}",
```
```suggestion
    path = "/query_results/{search_job_id}",
    description = "Streams the results of a previously submitted query as Server-Sent Events \
        (SSE).",
```
Same

components/api-server/src/client.rs
LinZhihao-723, line 51:
```suggestion
#[serde(deny_unknown_fields)]
pub struct QueryConfig {
    /// The search query as a KQL string.
    pub query_string: String,

    /// The dataset to search within. If not provided, defaults to all.
    #[serde(default)]
    pub dataset: Option<String>,

    /// The maximum number of results to return. Set to `0` for no limit.
    #[serde(default)]
    pub max_num_results: u32,

    /// The beginning timestamp (in epoch milliseconds) for the search range (inclusive).
    #[serde(default)]
    pub begin_timestamp: Option<i64>,

    /// The ending timestamp (in epoch milliseconds) for the search range (inclusive).
    #[serde(default)]
    pub end_timestamp: Option<i64>,

    /// Whether the string match should be case-insensitive.
    #[serde(default)]
    pub ignore_case: bool,

    /// Whether to write the search results to files. If `false`, results will be stored in MongoDB.
    #[serde(default)]
    pub write_to_file: bool,
}
```
I think it's worth to add descriptions for each parameter (an example is not sufficient to explain all the details).
utoipa supports direct translation from the field comment: https://github.com/juhaku/utoipa/issues/900

components/api-server/src/bin/api_server.rs
```diff
@@ -99,24 +101,73 @@ async fn main() -> anyhow::Result<()> {
...
+        contact(name = "YScope")
+    ))]
+    pub struct ApiDoc;
+}
+use api_doc::ApiDoc;
+
+#[utoipa::path(
+    get,
+    path = "/health",
+    responses((status = OK, body = String, example = "API server is running"))
```
```suggestion
    responses((status = OK, body = String))
```

LinZhihao-723 (changes requested):
Partially reviewed. Will get back to u tmr.

components/api-server/src/bin/api_server.rs
LinZhihao-723: Note to myself: An example might still be useful in Swagger UI. Will come back to this tmr.

components/api-server/src/bin/api_server.rs
hoophalab: utopia + json!  stores json numbers as i32, but we need nanoseconds to specify timestamp after Devin's timestamp PR. Maybe the plan has changed since my last sync up with kirk?

LinZhihao-723: Yeah, that's why I stripped the trailing 0s. I think as a demo example it should be fine. In the doc side we probably need to update it to nanosec once Devin's PR is in.

components/api-server/src/client.rs
```diff
@@ -13,24 +13,33 @@ use futures::{Stream, StreamExt};
...
     #[serde(default)]
     pub end_timestamp: Option<i64>,
+    /// Whether the string match should be case-insensitive.
     #[serde(default)]
     pub ignore_case: bool,
+    /// Whether to write the search results to files. If `false`, results will be stored in
+    /// `MongoDB`.
     #[serde(default)]
     pub write_to_file: bool,
 }
```
LinZhihao-723, line 51:
```suggestion
pub struct QueryConfig {
    /// The search query as a KQL string.
    pub query_string: String,

    /// The dataset to search within. If not provided, only `default` dataset will be searched.
    #[serde(default)]
    pub dataset: Option<String>,

    /// The maximum number of results to return. Set to `0` for no limit.
    #[serde(default)]
    pub max_num_results: u32,

    /// The beginning timestamp (in epoch milliseconds) for the search range (inclusive).
    #[serde(default)]
    pub begin_timestamp: Option<i64>,

    /// The ending timestamp (in epoch milliseconds) for the search range (inclusive).
    #[serde(default)]
    pub end_timestamp: Option<i64>,

    /// Whether the string match should be case-insensitive.
    #[serde(default)]
    pub ignore_case: bool,

    /// Whether to write the search results to files. If `false`, results will be stored in
    /// `MongoDB`.
    #[serde(default)]
    pub write_to_file: bool,
}
```
- I think adding an empty line here would be necessary for clarity.
- Update the content as only `default` will be searched when not specified.

components/api-server/src/routes.rs
```diff
@@ -0,0 +1,170 @@
...
+///
+/// # Returns
+///
+/// A newly created [`Router`] instance configured with all application routes.
+///
+/// # Errors
+///
+/// Returns an error if:
+///
+/// * Forwards [`serde_json::Error`] if serializing the `OpenAPI` documentation fails.
```
```suggestion
/// Factory method to create an Axum router configured with all API routes.
///
/// # Returns
///
/// A newly created [`axum::Router`] instance configured with all application routes.
///
/// # Errors
///
/// Returns an error if:
///
/// * Forwards [`OpenApi::to_json`]'s return values on failure.
```
To match our standard, we should use `Forwards [`FUNCTION`]'s return values on failures.`

components/api-server/src/routes.rs
```diff
@@ -0,0 +1,170 @@
...
+pub fn from_client(client: Client) -> Result<axum::Router, serde_json::Error> {
```
LinZhihao-723: We should move this function to the top: it's the only `pub` symbol.

components/api-server/src/routes.rs
```diff
@@ -0,0 +1,170 @@
...
+        .routes(routes!(query_results))
+        .with_state(client)
+        .split_for_parts();
+    let api_json = api.to_json()?;
+    let router = router
+        .route(
+            "/openapi.json",
+            get(|| async { (StatusCode::OK, api_json) }),
+        )
+        .layer(CorsLayer::new().allow_origin(Any));
```
LinZhihao-723: Just curious: Why do we need this?

components/api-server/src/routes.rs
```diff
@@ -0,0 +1,170 @@
...
+    path = "/query_results/{search_job_id}",
+    description = "Streams the results of a previously submitted query as Server-Sent Events \
+        (SSE).",
+    responses(
+        (
+            status = OK,
+            body = String,
+            content_type = "text/event-stream",
+            description = "Server-Sent Events stream of query results. Each event contains a \
+                single line of the query result in JSON format.",
```
LinZhihao-723, line 126:
```suggestion
                single line of the query result in JSON format.",
            example = r#"data: {"timestamp": 1633036800, "message": "Example log message"}\n"#
```
Add a short example to address my own notes last night.

components/api-server/src/routes.rs
```diff
@@ -0,0 +1,170 @@
...
+            "begin_timestamp": 0,
+            "end_timestamp": 17_356_896,
+            "ignore_case": true,
+            "max_num_results": 0,
+            "write_to_file": false
+        })),
+    responses(
+        (
+            status = OK,
+            body = QueryResultsUri,
```
LinZhihao-723, line 90:
```suggestion
            body = QueryResultsUri,
            description = "The URI to fetch the results of the submitted query.",
```

components/api-server/src/bin/api_server.rs
```diff
@@ -95,75 +81,15 @@ async fn main() -> anyhow::Result<()> {
         .await
         .context(format!("Cannot listen to {addr}"))?;
 
-    let client = clp_client::Client::connect(&config, &credentials)
+    let client = api_server::client::Client::connect(&config, &credentials)
```
LinZhihao-723, line 84: since we name the library crate to `api_server`, shall we rename `client` module to `clp_client` for clarity?

hoophalab: hmm I still feel we shouldn't mention clp in any place of the code. Let's rename this after we have another client

components/api-server/src/routes.rs
hoophalab: Without this, the user cannot put openapi.json url to swagger ui in the browser, because the json url is from a different host.

PR #1668, opened 2025-11-26
feat(clp-credential-manager): Create `clp-credential-manager` Rust service skeleton.

Eden-D-Zhang:
# Description

This PR sets the foundation of the `clp-credential-manager` service. 

Added:
- New component: `clp-credential-manager` Rust Crate.
- Dependencies, including those needed in the future (AWS, JWT, SQLx, etc.)
- Service configuration structs
- Error enumeration considering planned functionality
- Basic `health` route for service API
- Barebones service initialization

Next up:
- Implement credential DB CRUD operations
- Add DB operation auditing

# Checklist

  breaking change.

# Validation performed

Cargo fmt, check, clippy


components/clp-credential-manager/src/config.rs
```diff
@@ -0,0 +1,113 @@
...
+    ///
+    /// # Returns:
+    ///
+    /// A parsed [`AppConfig`] instance when the file can be read and deserialized successfully.
+    ///
+    /// # Errors:
+    ///
+    /// * Returns [`ServiceError::Io`] if the file cannot be read.
+    /// * Returns [`ServiceError::Yaml`] if parsing fails.
+    pub fn from_file(path: &Path) -> ServiceResult<Self> {
```
hoophalab: We can use this function:
https://github.com/y-scope/clp/blob/main/components/clp-rust-utils/src/serde/yaml.rs#L18

components/clp-credential-manager/src/config.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+/// Network settings for the Axum server.
+#[derive(Debug, Clone, Deserialize)]
+pub struct ServerConfig {
+    #[serde(default = "default_bind_address")]
+    pub bind_address: String,
+    #[serde(default = "default_bind_port")]
+    pub port: u16,
+}
```
hoophalab, line 51:
```suggestion
#[serde(default)]
pub struct ServerConfig {
    pub bind_address: String,
    pub port: u16,
}
```
We actually don't need standalone functions such as `default_bind_address` and `default_bind_port` if we have a `impl Default for ServerConfig`. However, feel free to keep the code as is. It's totally fine.

components/clp-credential-manager/src/config.rs
```diff
@@ -0,0 +1,113 @@
...
+    ///
+    /// Returns [`ServiceError::Config`] if the address string cannot be parsed.
+    pub fn socket_addr(&self) -> ServiceResult<SocketAddr> {
+        let addr = format!("{}:{}", self.bind_address, self.port);
+        addr.parse().map_err(|err| {
+            ServiceError::Config(format!(
+                "invalid bind address `{}`:{} ({err})",
+                self.bind_address, self.port
+            ))
+        })
```
hoophalab: Not sure whether we need this function. We can simply pass a (hostname, port) tuple to `TcpListener::bind` because of this impl
https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html#impl-ToSocketAddrs-for-(String,+u16)

components/clp-credential-manager/src/error.rs
```diff
@@ -0,0 +1,161 @@
...
+
+impl From<sqlx::Error> for ServiceError {
+    /// Maps raw [`sqlx::Error`] values into domain-specific variants so callers can react
+    /// precisely.
+    ///
+    /// # Parameters:
+    ///
+    /// * `err`: The database error surfaced by `sqlx`.
+    fn from(err: sqlx::Error) -> Self {
+        if matches!(err, sqlx::Error::RowNotFound) {
```
hoophalab, line 52:
```suggestion
        if let sqlx::Error::RowNotFound = err {
```
This might be easier to read

components/clp-credential-manager/src/service.rs
```diff
@@ -0,0 +1,56 @@
...
+use crate::{config::AppConfig, error::ServiceResult};
+
+/// Reference-counted handle that Axum stores as application state.
+pub type SharedService = Arc<CredentialManagerService>;
+
+/// High-level facade that wires together persistence, auditing, and JWT handling.
+#[allow(dead_code)]
+pub struct CredentialManagerService {
+    db_pool: MySqlPool,
+}
```
hoophalab: I could be missing some context here. Feel free to push back.

MySqlPool is actually already [wrapped by an Arc](https://docs.rs/sqlx-core/0.8.6/src/sqlx_core/pool/mod.rs.html#260). So we can actually simply clone db_pool to share the pool in multiple modules/threads.

So, based on the current code (and the code in #1650), I would suggest simply derive clone.
```suggestion
/// High-level facade that wires together persistence, auditing, and JWT handling.
#[derive(Clone)]
pub struct CredentialManagerService {
    db_pool: MySqlPool,
}
```

However, I don't have the full picture, and there might be more stuffs we want to add to `CredentialManagerService`, which need to be protected by an Arc.

components/clp-credential-manager/src/main.rs
```diff
@@ -0,0 +1,68 @@
...
+/// defaults to the `info` level so local development remains verbose enough.
+fn init_tracing() {
+    let env_filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info"));
+
+    let _ = tracing_subscriber::fmt()
+        .with_env_filter(env_filter)
+        .with_target(false)
+        .json()
+        .try_init();
+}
```
hoophalab: 1. The tracing_subscriber::fmt()...try_init() call returns a Result which is being dropped, meaning any initialization failure will be ignored, leading to silent bugs.

Possible implementations: we could use anyhow::Result<()>, or define a crate-wide `Error` type and use Result<(), Error>
```suggestion
fn init_tracing() -> Result<(), Error>{
    let env_filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info"));

    tracing_subscriber::fmt()
        .with_env_filter(env_filter)
        .with_target(false)
        .json()
        .try_init()?;
}
```
2. Zhihao will probably suggest you implementing log file rotation  in code.
https://github.com/y-scope/clp/blob/main/components/api-server/src/bin/api_server.rs#L45C1-L66C2

components/clp-credential-manager/src/routes/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+/// Composes the service's HTTP surface, layering tracing instrumentation globally.
+///
+/// # Returns:
+///
+/// An [`axum::Router`] wired with all route trees and middleware layers.
+pub fn build_router() -> Router<AppState> {
+    Router::new()
+        .merge(health::router())
+        .layer(TraceLayer::new_for_http())
+}
```
hoophalab:
```suggestion
pub fn build_router() -> Router<AppState> {
    Router::new()
        .route("/health", get(health::health_check))
        .layer(TraceLayer::new_for_http())
}
```
Can we remove `health::router()` and reference the function here directly

components/clp-credential-manager/src/error.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+}
+
+/// HTTP-friendly representation of service failures.
+#[derive(Debug)]
+pub struct ApiError {
+    status: StatusCode,
+    message: String,
+}
+
```
hoophalab: I've checked #1650. I guess `ApiError` is here so that we can return `Result<..., ApiError>` in handlers.

However, it might be easier if we 1) remove `ApiError`, 2) `impl IntoResponse for ServiceError`, and 3) return `Result<..., ServiceError>` directly in handlers.

components/clp-credential-manager/src/service.rs
Eden-D-Zhang: I believe the plan is also to add the AWS client to `CredentialManagerService` (as well as the JWT stuff). I'd rather leave it for now and consider whether or not it's needed once the implementation is more fleshed-out.

components/clp-credential-manager/src/error.rs
Eden-D-Zhang: This syntax is forced by our linter, lol.

hoophalab: lol fine xD

components/clp-credential-manager/src/config.rs
```diff
@@ -111,3 +88,12 @@ const fn default_mysql_port() -> u16 {
...
 }
+
+/// Maps shared YAML helper errors into service-specific error variants.
+fn map_yaml_error(err: UtilsError) -> ServiceError {
+    match err {
+        UtilsError::Io(io_err) => ServiceError::Io(io_err),
+        UtilsError::SerdeYaml(yaml_err) => ServiceError::Yaml(yaml_err),
+        other => ServiceError::Config(other.to_string()),
+    }
+}
```
hoophalab:
```suggestion
impl From<UtilsError> for ServiceError{
fn from(err: UtilsError) -> ServiceError {
    match err {
        UtilsError::Io(io_err) => ServiceError::Io(io_err),
        UtilsError::SerdeYaml(yaml_err) => ServiceError::Yaml(yaml_err),
        other => ServiceError::Config(other.to_string()),
    }
}
}
```
related to another comment
```
        .try_init()?
```

components/clp-credential-manager/src/config.rs
```diff
@@ -33,42 +34,18 @@ impl AppConfig {
     /// * Returns [`ServiceError::Io`] if the file cannot be read.
     /// * Returns [`ServiceError::Yaml`] if parsing fails.
     pub fn from_file(path: &Path) -> ServiceResult<Self> {
-        let contents = fs::read_to_string(path)?;
-        let config: Self = serde_yaml::from_str(&contents)?;
-        Ok(config)
+        yaml::from_path(path).map_err(map_yaml_error)
```
hoophalab:
```suggestion
        yaml::from_path(path)?
```
Also, it might be easier to let main call ` yaml::from_path`
For the modularity, `config.rs` only defines the data format and shouldn't care about whether the data is deserialized from yaml or json.

components/clp-credential-manager/src/main.rs
```diff
@@ -57,12 +58,13 @@ async fn main() -> anyhow::Result<()> {
...
     let env_filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info"));
 
-    let _ = tracing_subscriber::fmt()
+    tracing_subscriber::fmt()
         .with_env_filter(env_filter)
         .with_target(false)
         .json()
-        .try_init();
+        .try_init()
+        .map_err(|err| anyhow::Error::msg(err.to_string()))
```
hoophalab:
```suggestion
        .try_init()?
```

hoophalab (approved):
A few comments to make the code more rusty, otherwise all good. @LinZhihao-723 you  can start having another review.

components/clp-credential-manager/src/config.rs
hoophalab: but feel free to keep the code as-is. Different people would have different opinions on this.

components/clp-credential-manager/src/config.rs
hoophalab: sorry related to yaml::from_path()?

components/clp-credential-manager/src/main.rs
Eden-D-Zhang: This change isn't working for me:

"`?` couldn't convert the error: `dyn StdError + Send + Sync: Sized` is not satisfied"

components/clp-credential-manager/Cargo.toml
LinZhihao-723, line 1: Can u check whether the following dependencies are actually in use:
- chrono
- serde_json
- tower
- uuid
- aws-config
- aws-sdk-sts
- reqwest

Also, `jsonwebtoken`'s latest version is 10.2.0. Can u update?

components/clp-credential-manager/src/routes/mod.rs
LinZhihao-723, line 1: We decided to follow the latest guideline of not using `mod` (like other mod created in this project isn't using `mod.rs` explicitly either). Please check `api-server` to see how route was set up there.

components/clp-credential-manager/src/routes/mod.rs
```diff
@@ -0,0 +1,20 @@
+mod health;
+
+use axum::{Router, routing::get};
+use tower_http::trace::TraceLayer;
+
+use crate::service::SharedService;
+
+/// Shared Axum state for every route module.
+pub type AppState = SharedService;
```
LinZhihao-723: There is no need for this alias. Using the actual type is more direct and clear.

components/clp-credential-manager/src/routes/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+/// Composes the service's HTTP surface, layering tracing instrumentation globally.
+///
+/// # Returns:
+///
+/// An [`axum::Router`] wired with all route trees and middleware layers.
+pub fn build_router() -> Router<AppState> {
+    Router::new()
+        .route("/health", get(health::health_check))
+        .layer(TraceLayer::new_for_http())
+}
```
LinZhihao-723: Can u pass the state into this method to be consistent with the API server?
https://github.com/y-scope/clp/blob/2534e391e833904689abc0387e5de2a2c3e1efbe/components/api-server/src/routes.rs#L38

components/clp-credential-manager/src/routes/health.rs
```diff
@@ -0,0 +1,10 @@
+use axum::http::StatusCode;
+
+/// Responds with `200 OK` to signal that the service is alive.
+///
+/// # Returns:
+///
+/// [`StatusCode::OK`] regardless of request parameters.
+pub async fn health_check() -> StatusCode {
+    StatusCode::OK
```
LinZhihao-723: We should be consistent with other components:
- API server: https://github.com/y-scope/clp/blob/2534e391e833904689abc0387e5de2a2c3e1efbe/components/api-server/src/routes.rs#L70
- Log ingestor: https://github.com/y-scope/clp/blob/2534e391e833904689abc0387e5de2a2c3e1efbe/components/log-ingestor/src/main.rs#L101

components/clp-credential-manager/src/service.rs
```diff
@@ -0,0 +1,56 @@
+use std::sync::Arc;
+
+use secrecy::ExposeSecret;
+use sqlx::{MySqlPool, mysql::MySqlPoolOptions};
+
+use crate::{config::AppConfig, error::ServiceResult};
+
+/// Reference-counted handle that Axum stores as application state.
+pub type SharedService = Arc<CredentialManagerService>;
```
LinZhihao-723: I don't think we need `Arc`. When passing the service as a state to axum, it automatically wraps it with an arc. Check the source code of `Route<S>`: https://github.com/tokio-rs/axum/blob/5c4c1658a7a5151986c3a22bb8fdf71d5e749679/axum/src/routing/mod.rs#L72
@hoophalab please also check if I'm wrong.

components/clp-credential-manager/src/service.rs
```diff
@@ -0,0 +1,56 @@
...
+
+/// High-level facade that wires together persistence, auditing, and JWT handling.
```
LinZhihao-723, line 11: I'm not sure if I understand this docstring, lol. The wording is too general without any concrete explanation of what this class is doing.

components/clp-credential-manager/src/service.rs
```diff
@@ -0,0 +1,56 @@
...
+            .host(&db_config.host)
+            .port(db_config.port)
+            .database(&db_config.name)
+            .username(&db_config.user)
+            .password(db_config.password.expose_secret());
+
+        let pool = MySqlPoolOptions::new()
+            .max_connections(db_config.max_connections)
+            .connect_with(options)
+            .await?;
```
LinZhihao-723, line 48: This is duplicating https://github.com/y-scope/clp/blob/2534e391e833904689abc0387e5de2a2c3e1efbe/components/clp-rust-utils/src/database/mysql.rs#L20. Please use the existing util.

components/clp-credential-manager/src/service.rs
```diff
@@ -0,0 +1,56 @@
...
+    db_pool: MySqlPool,
+}
+
+impl CredentialManagerService {
+    /// Establishes the database connection pool used by all route handlers.
+    ///
+    /// # Parameters:
+    ///
+    /// * `config`: Fully parsed application configuration that contains database settings.
+    ///
```
LinZhihao-723:
```suggestion
```
In Rust we don't document the parameters, to follow the spirit of standard Rust coding guideline that the parameter itself should be self explanatory.

components/clp-credential-manager/src/service.rs
```diff
@@ -0,0 +1,56 @@
...
+    /// # Returns:
+    ///
+    /// A fully initialized [`CredentialManagerService`] ready to be shared with Axum routes.
+    ///
+    /// # Errors:
+    ///
+    /// * Propagates errors from [`MySqlPoolOptions::connect_with`] when the pool cannot be created.
```
LinZhihao-723: The format of `Errors` section should follow: https://github.com/y-scope/clp/blob/2534e391e833904689abc0387e5de2a2c3e1efbe/components/log-ingestor/src/ingestion_job/sqs_listener.rs#L43
* Add `Returns an error if:`
* When propagating errors, it should be `Forwards [xxx]'s return values on failure`.

components/clp-credential-manager/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+    let listener = TcpListener::bind((bind_address.as_str(), bind_port)).await?;
+    axum::serve(listener, router).await?;
+
+    Ok(())
+}
+
+/// Sets up JSON-formatted tracing using environment filters.
+///
+/// The subscriber honors the `RUST_LOG` environment variable when present and otherwise
+/// defaults to the `info` level so local development remains verbose enough.
```
LinZhihao-723, line 77: Note: this comment isn't follow the standard as it's missing `#Returns` and `#Errors` sections.

components/clp-credential-manager/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+fn init_tracing() -> anyhow::Result<()> {
```
LinZhihao-723: Let's try to be consistent: https://github.com/y-scope/clp/blob/2534e391e833904689abc0387e5de2a2c3e1efbe/components/log-ingestor/src/main.rs#L43
Ideally we should address this by #1714. But before that, let's copy the code to here.

components/clp-credential-manager/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+    )]
+    config: PathBuf,
+}
+
+/// Binary entry point that configures logging, loads config, and starts Axum.
+///
+/// # Errors:
+///
+/// Returns [`anyhow::Error`] when configuration parsing, database connections, TCP binding,
+/// or Axum startup fail.
```
LinZhihao-723: This error section doesn't match the guideline. Errors are all propagated from other functions. The correct way to do it is:
```suggestion
/// # Errors:
///
/// Returns an error if:
///
/// * Forwards [`AppConfig::from_file`]'s return values on failure.
/// * Forwards [`CredentialManagerService::new`]'s return values on failure.
/// * Forwards [`TcpListener::bind`]'s return values on failure.
/// * Forwards [`axum::serve`]'s return values on failure.
```
Notice that if you rename some of the classes, you might need to change my suggestion accordingly.

components/clp-credential-manager/src/error.rs
```diff
@@ -0,0 +1,99 @@
...
+    #[error("jwt error: {0}")]
+    Jwt(#[source] JwtError),
+
+    #[error("i/o error: {0}")]
+    Io(#[from] std::io::Error),
+
+    #[error("serialization error: {0}")]
+    Yaml(#[from] serde_yaml::Error),
+}
```
LinZhihao-723, line 38: We should avoid pre-define errors that are not used yet. Can you strip non-used errors from this PR to sharp the scope of the review (like `Jwt`)? Without the actual context of use it's hard to determine whether the name and error message makes sense from a reviewer's perspective.

components/clp-credential-manager/src/error.rs
```diff
@@ -0,0 +1,99 @@
...
+
+impl From<sqlx::Error> for ServiceError {
+    /// Maps raw [`sqlx::Error`] values into domain-specific variants so callers can react
+    /// precisely.
+    ///
+    /// # Parameters:
+    ///
+    /// * `err`: The database error surfaced by `sqlx`.
```
LinZhihao-723, line 46:
```suggestion
```
I would say you can even remove the entire docstring for this method.

components/clp-credential-manager/src/error.rs
```diff
@@ -0,0 +1,99 @@
...
+    NotFound(String),
+
+    #[error("jwt error: {0}")]
+    Jwt(#[source] JwtError),
+
+    #[error("i/o error: {0}")]
+    Io(#[from] std::io::Error),
+
+    #[error("serialization error: {0}")]
+    Yaml(#[from] serde_yaml::Error),
```
LinZhihao-723: When forwarding errors from another type, you should do: https://github.com/y-scope/clp/blob/2534e391e833904689abc0387e5de2a2c3e1efbe/components/api-server/src/error.rs#L15
Do it for other errors.

components/clp-credential-manager/src/error.rs
```diff
@@ -0,0 +1,99 @@
...
+    Json,
+    http::StatusCode,
+    response::{IntoResponse, Response},
+};
+use jsonwebtoken::errors::Error as JwtError;
+use serde::Serialize;
+use thiserror::Error;
+
+/// Convenience alias for functions that return a [`ServiceError`].
+pub type ServiceResult<T> = Result<T, ServiceError>;
```
LinZhihao-723: I think aliasing non "Result" name if Rust is weird (@hoophalab correct me if I'm wrong). We should use sth like:
```rust
pub type Result<T> = std::result::Result<T, ServiceError>;
```
And refer to this type as `crate::error::Result`.

components/clp-credential-manager/src/config.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/// Supplies the default bind port of `8080` that matches other CLP services.
+const fn default_bind_port() -> u16 {
+    DEFAULT_SERVER_PORT
+}
+
+/// Connection options for the `MySQL` backend that stores credentials.
+#[derive(Debug, Clone, Deserialize)]
+pub struct DatabaseConfig {
```
LinZhihao-723, line 79: In another comment I mentioned that we should use the existing utils. But that assumes it's using the same config from `clp-rust-utils`. Can we share this object with CLP's DB config? I think other than `max_connections` it's identical with CLP's DB config. @junhaoliao

components/clp-credential-manager/src/config.rs
```diff
@@ -0,0 +1,100 @@
...
+
+/// Root configuration deserialized from `credential-manager-config.yml`.
+#[derive(Debug, Clone, Deserialize)]
+pub struct AppConfig {
+    #[serde(default)]
+    pub server: ServerConfig,
+    pub database: DatabaseConfig,
+}
+
+impl AppConfig {
```
LinZhihao-723: We should avoid using type names like `AppConfig`. "App" doesn't mean anything here. We should use a meaningful word instead, like `CredentialManagerConfig`.

components/clp-credential-manager/src/config.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// * `path`: Filesystem location of the YAML configuration file.
+    ///
+    /// # Returns:
+    ///
+    /// A parsed [`AppConfig`] instance when the file can be read and deserialized successfully.
+    ///
+    /// # Errors:
+    ///
+    /// * Returns [`ServiceError::Io`] if the file cannot be read.
+    /// * Returns [`ServiceError::Yaml`] if parsing fails.
```
LinZhihao-723: Rewrite the error sections as suggested in other comments.
This comes to the confusing part: `yaml::from_path` returns a `clp_rust_utils::Error`. Does it make more sense to just map this error type in `ServiceError`? You should dig into the details of that error to distinguish `Io` and `Yaml` like in your comment, as that implementation may change.

components/clp-credential-manager/src/config.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// Provides sensible defaults that bind the HTTP server to all interfaces on port 8080.
+    fn default() -> Self {
+        Self {
+            bind_address: default_bind_address(),
+            port: default_bind_port(),
+        }
+    }
+}
+
+/// Supplies the default bind address of `0.0.0.0` so the service listens on every interface.
```
LinZhihao-723:
```suggestion
/// Returns the default bind address of `0.0.0.0` so the service listens on every interface.
```
Prefer `Returns`. Do it for other ones.

components/clp-credential-manager/src/config.rs
```diff
@@ -0,0 +1,100 @@
...
+
+/// Supplies the default `MySQL` server port (`3306`).
+const fn default_mysql_port() -> u16 {
+    3306
+}
+
+/// Supplies the default maximum connection count for the `MySQL` pool.
+const fn default_max_connections() -> u32 {
+    DEFAULT_MAX_CONNECTIONS
+}
```
LinZhihao-723: We should group all "default" wrappers together.

components/clp-credential-manager/src/config.rs
LinZhihao-723, line 79: `pub` should comes before private.

components/clp-credential-manager/src/config.rs
```diff
@@ -0,0 +1,100 @@
...
+impl From<UtilsError> for ServiceError {
```
LinZhihao-723:
```suggestion

impl From<UtilsError> for ServiceError {
```

components/clp-credential-manager/src/config.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn from(err: UtilsError) -> Self {
+        match err {
+            UtilsError::Io(io_err) => Self::Io(io_err),
+            UtilsError::SerdeYaml(yaml_err) => Self::Yaml(yaml_err),
+            other => Self::Config(other.to_string()),
+        }
+    }
```
LinZhihao-723: Back to the previous comment: I would say this won't make things clear. When you have an error returned from `yaml::from_path`, the top-level caller logs that error and return. You don't need to distinguish whether it's an IO error or Yaml error or anything else (which shoudn't happen in this case, which makes `Self::Config` return a bit more confusing). I would suggest just map `UtilsError` in `ServiceError` and directly return that error.

components/clp-credential-manager/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    let args = Args::parse();
+
+    init_tracing()?;
+
+    let config = AppConfig::from_file(&args.config)?;
```
LinZhihao-723: Back to the previous comment: if you want to make the errors traceable, you should log the error before return. Otherwise when you start this thing inside docker compose I'm not sure how that error will be reflected.

LinZhihao-723 (changes requested):
In general a great job, but still some things to fix. @hoophalab please also check my comments so that you can help catch more later.
I think there are two lessons learned in this PR:
* As your first PR to Rust this is still too large. Personally, I would say this PR can be split into three PRs: one to set up the error handling system, one to set up the server skeleton like #1715, and last one for the rest. In that way, you get more chance to learn step by step and save more time on the later review process.
* When contributing to sth you're not familiar with, you can check what other components are doing. We had similar set up for the API server and clp-rust-utils a month ago, which can be references to you to pick up.

components/clp-credential-manager/src/error.rs
hoophalab: Personally, I would avoid aliasing and use ` Result<T, ServiceError>` in the code.
This is because an aliased name which seem reasonable to me might not be reasonable to the other people. And there is always a cognitive pressure to picking up aliases. ServiceResult also doesn't save much typing than `Result<T, ServiceError>`

components/clp-credential-manager/src/config.rs
hoophalab: I agree

components/clp-credential-manager/src/config.rs
hoophalab: Not sure but I guess Eden might don't want to provide default value for every field? So he cannot use `impl Default for ...`

components/clp-credential-manager/src/service.rs
hoophalab: We've had a similar discussion here https://github.com/y-scope/clp/pull/1668#discussion_r2577938315
Eden, do you want to reconsider the suggestion?

LinZhihao-723: * I wasn't suggesting things inside `CredentialManagerService`; I'm saying the state itself is wrapped by an Arc when creating an router.
* **However**, I realize I was wrong after reading more carefully on their implementation. The router implementation itself was wrapped by an Arc, but not the state. So I think an Arc of the state would still be needed.
* But still, I don't like the way of adding an alias of the service as it's exposing the Arc to the actual user of the service manager. I would suggest we do sth similar to axum:
  * The shared state is not an Arc, but it's a wrapper of an Arc. It should be sth like `struct CredentialManagerServiceState { inner: Arc<CredentialManagerService>; }`
  * Implement `CredentialManagerServiceState` to wrap the use of Arc.

This design should also provide flexibility for performance improvements. For example, both mysql and aws client are wrappers of Arc already. You probably don't need to put them into another layer of Arc. You can protect the rest of shared states using `Arc`, but it will be wrapped by `CredentialManagerServiceState` while `CredentialManagerServiceState` is always safe to copy and share.

components/clp-credential-manager/src/error.rs
LinZhihao-723: Let's go with `Result<T, ServiceEreror>` then.

components/clp-credential-manager/src/config.rs
LinZhihao-723: Let's wait for @junhaoliao to check whether we should share the DB credential config schema with CLP DB config. If they're shareable then we can get rid of this entirely.

components/clp-credential-manager/src/service.rs
hoophalab: I agree with this approach: "The shared state is not an Arc, but it's a wrapper of an Arc. It should be sth like `struct CredentialManagerServiceState { inner: Arc<CredentialManagerService>; }`

Or, we could even do  `struct CredentialManagerService { inner: Arc<State>; pool: MySqlPool}` to minimize the critical region, if there aren't races between pool and inner

components/clp-credential-manager/src/config.rs
hoophalab: Yes, we can.

Eden-D-Zhang: This might be irrelevant pending the option of using the `clp_rust_utils` config instead, right?

components/clp-credential-manager/src/routes/mod.rs
Eden-D-Zhang: I had trouble getting the above suggestion to work, plus it seems the current implementation is actually recommended.

https://docs.rs/axum/0.8.7/axum/routing/struct.Router.html#returning-routers-with-states-from-functions

components/clp-credential-manager/src/config.rs
LinZhihao-723: Yes. But it's a heads up for the proper ordering.

components/clp-credential-manager/src/routes/mod.rs
LinZhihao-723: kk thx for referring to the doc. Let me check.

components/clp-credential-manager/src/main.rs
Eden-D-Zhang: I'm not sure I understand what you mean here; can you explain what I need to add here?

components/clp-credential-manager/src/config.rs
junhaoliao: i'm fine if we want to share this. so long max_connections is optionally configurable rather than mandatory

Eden-D-Zhang: Does `max_connections` need to be configurable? I noticed the api server hardcodes the value as 10.

PR #1646, opened 2025-11-20
feat(clp-s::timestamp_parser): Add support for marshalling timestamps.

gibber9809:
# Description

This PR follows up on #1626, #1599, #1564, and #1385 by implementing marshalling for numeric and date-time timestamp patterns. The marshalling implementation currently uses `fmt` to format some parts of a timestamp, which results in some extra copying of small strings; if this ends up being a performance problem we can switch to a custom implementation that elides these small copies.

This PR may be followed up by one more PR to the `clp-s::timestamp_parser` library to improve support for parsing leap-seconds before a PR to fully integrate `clp-s::timestamp_parser` with clp-s.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* Added assertions to verify that all timestamps that we parse in the tests can be marshalled correctly.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **New Features**
  * Added timestamp marshaling/formatting to produce date‑time and numeric epoch outputs (12/24h, AM/PM, day/month/year, day‑of‑week, timezone, subsecond precision).

* **Bug Fixes**
  * Improved validation for zero‑padding and clearer errors for invalid/incompatible timestamp patterns.

* **Tests**
  * Added unit tests that parse timestamps and verify round‑trip formatting for several patterns and edge cases.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1626, opened 2025-11-19
feat(clp-s::timestamp_parser): Add support for parsing CAT sequences; Add utility functions to provide a default set of timestamp patterns.

gibber9809:
# Description

This PR follows up on #1599, #1564, and #1385 by implementing support for parsing CAT sequences.

In particular this includes:

`\Z` - parse an arbitrary ISO8601 timezone and convert it into static text and a `\z{...}` format specifier.
`\P` - parse an unknown precision epoch timestamp and convert it into `\E`, `\L`, `\C`, or `\N` based on a heuristic.
`\?` - parse the sub-second fractional part of a timestamp and convert it into `\3`, `\6`, `\9`, or `\T` depending on the number of digits.
`\O{...}` - match one of the characters between `{}` and convert it into the matched literal character.

We use these CAT sequences, in combination with the previously implemented format specifiers, to create several utility functions that provide a default set of timestamp patterns. This default set of patterns is a superset of the default set of patterns specified in the old implementation.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* Added tests to validate that default sets of patterns are initialized without error.
* Added assertions to validate that all timestamps in the tests have their formats accurately inferred using the default set of timestamp patterns.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **New Features**
  * Enhanced timestamp parsing: better timezone, literal and fractional‑second support, dynamic precision estimation; new APIs to retrieve and search default timestamp patterns (including quoted variants).

* **Bug Fixes**
  * Removed a formerly special‑cased format‑specifier error, normalizing reported message for that case.

* **Chores**
  * Build configuration updated to add a formatting library as a private dependency.

* **Tests**
  * Added tests for CAT‑sequence transformations, default pattern validity and parsing/search parity.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1599, opened 2025-11-13
feat(clp-s::timestamp_parser): Add support for parsing date-time timestamps with timezones; Add `TimestampPattern` class to pre-validate timestamp patterns.

gibber9809:
# Description

This PR follows up on #1564 and #1385 by adding support for parsing date-time timestamps with timezones. To simplify the code in the `parse_timestamp` function and avoid some kinds of duplicated parsing work each time we parse a timestamp with a given pattern, this PR also introduces a `TimestampPattern` class which pre-validates timestamp patterns, so that `parse_timestamp` can skip many kinds of validation.

Note: the `\z{}` format specifier is only designed to capture [ISO 8601 timezone offsets](https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC). The intention is that ISO 8601 timezone information that isn't this offset should be captured as static text (e.g., `UTC\z{...}`). The `\Z` CAT sequence will be responsible for trying to find such static text once it is implemented.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* Added tests for parsing timestamps with timezones
* Added tests to ensure that \z sequences accept valid content
* Added assertions that timestamp patterns are correct in tests

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **New Features**
  * Support for parsing and validating timezone offset formats (e.g., ±HH, ±HHMM, ±HH:MM) in timestamps
  * Added a pattern-based timestamp object for validating and classifying formats (including 12‑hour/AM‑PM)

* **Bug Fixes**
  * Standardized and refined error message text for consistency
  * New error case surfaced for invalid timezone offsets

* **Tests**
  * Expanded tests to cover timezone offset variants and validate patterns before parsing

PR #1564, opened 2025-11-06
feat(clp-s::timestamp_parser): Add support for parsing date-time and epoch timestamps without timezones.

gibber9809:
# Description

This PR follows up on #1385 by implementing the remainder of the initial set of format specifier excepting timezones, and CAT sequences.

Since the set of supported format specifiers is nearly complete, we also add tests for parsing all of the unique timestamps from `tests/test-TimestampPattern.cpp`, with some additional testcases focused on numeric timestamps.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* Added tests to ensure every newly implemented format specifier accepts the expected kind of content.
* Added tests to ensure that timestamps are parsed accurately for known cases.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **New Features**
  * Enhanced timestamp parsing: 12/24‑hour support with AM/PM, minutes, seconds, and variable‑length subseconds (up to nanoseconds); accepts direct epoch inputs.

* **Bug Fixes**
  * Stricter validation and error handling: range checks, consistent 12‑hour/AM‑PM rules, rejection of mixed or trailing formats, and correct epoch/subsecond composition.

* **Tests**
  * Expanded test coverage for padding variants, AM/PM and 12‑hour scenarios, and wide‑range epoch representations.

PR #1673, opened 2025-11-26
feat(clp-s::timestamp_parser): Add support for preserving leap seconds in date-time timestamps.

gibber9809:
# Description

This is the final (for now) PR for implementing the new clp_s::timestamp_parser utility. This PR follows up on #1646, #1626, #1599, #1564, and #1385.

This PR implements support for preserving leap-seconds in date-time timestamps. This is accomplished by:

1. Modifying the `\S` format specifier to only accept seconds in the range `00-59`
2. Adding the `\J` format specifier which accepts the second value `60`, always marshalls the seconds value as `60`, and parses the seconds value as `59` (since we store timestamps in epoch nanosecond time, which doesn't count leap-seconds).
3. Adding the `\s` CAT sequence which transforms into `\S` or `\J` depending on the seconds value.
4. Modifying all of the default format specifiers to use `\s` instead of `\S`.

Note that while this can make timestamps during a leap second and prior to a leap second unordered in epoch time (which is what we store), it is possible to accurately transform each timestamp into UTC time, in which the timestamps are properly ordered.

This solution is explained in more detail in the timestamp parser design doc.

# Checklist

  breaking change.

# Validation performed

* Added tests validating that timestamps directly before, during, and after a leap second are transformed into epoch time as expected, and are round-tripped correctly.

## Summary by CodeRabbit

* **New Features**
  * Added leap-second support across parsing and marshalling, enabling 60-second timestamps and a dedicated leap-second specifier.
  * Introduced a generic zero-padded second specifier to improve pattern recognition and dynamic substitution between regular and leap seconds.

* **Bug Fixes**
  * Adjusted standard second validation to 00–59 and harmonized second-handling behavior.

* **Tests**
  * Expanded tests to cover leap-second and end-of-year timestamp scenarios.

### DIFF
--- a/components/webui/client/src/components/InputLabel/index.tsx
+++ b/components/webui/client/src/components/InputLabel/index.tsx
@@ -13,9 +13,10 @@ const {Text} = Typography;
  *
  * @param props
  * @param props.children The label text to display.
+ * @param props.width Optional width for the label.
  * @return
  */
-const InputLabel = ({children}: {children: React.ReactNode}) => {
+const InputLabel = ({children, width}: {children: React.ReactNode; width?: string | number}) => {
     const {token} = theme.useToken();
 
     return (
@@ -26,6 +27,7 @@ const InputLabel = ({children}: {children: React.ReactNode}) => {
                 borderBottomLeftRadius: `${token.borderRadius}px`,
                 borderTopLeftRadius: `${token.borderRadius}px`,
                 fontSize: token.fontSize,
+                minWidth: width,
             }}
         >
             {children}
--- a/components/webui/client/src/pages/SearchPage/SearchControls/Presto/GuidedControls/From.tsx
+++ b/components/webui/client/src/pages/SearchPage/SearchControls/Presto/GuidedControls/From.tsx
@@ -1,6 +1,7 @@
 import InputLabel from "../../../../../components/InputLabel";
 import DatasetSelect from "../../Dataset/DatasetSelect";
 import guidedGrid from "./index.module.css";
+import {LABEL_WIDTH} from "./typings";
 
 
 /**
@@ -11,7 +12,7 @@ import guidedGrid from "./index.module.css";
 const From = () => {
     return (
         <div className={guidedGrid["gridItem"]}>
-            <InputLabel>FROM</InputLabel>
+            <InputLabel width={LABEL_WIDTH}>FROM</InputLabel>
             <DatasetSelect
                 className={
                     `${guidedGrid["noLeftBorderRadiusSelect"]} ${
--- a/components/webui/client/src/pages/SearchPage/SearchControls/Presto/GuidedControls/OrderBy.tsx
+++ b/components/webui/client/src/pages/SearchPage/SearchControls/Presto/GuidedControls/OrderBy.tsx
@@ -5,6 +5,7 @@ import useSearchStore from "../../../SearchState/index";
 import usePrestoSearchState from "../../../SearchState/Presto";
 import {SEARCH_UI_STATE} from "../../../SearchState/typings";
 import guidedGrid from "./index.module.css";
+import {LABEL_WIDTH} from "./typings";
 
 
 /**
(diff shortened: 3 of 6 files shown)


components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -41,7 +41,8 @@ constexpr int cMaxParsedHour12HourClock{12};
 constexpr int cMinParsedMinute{0};
 constexpr int cMaxParsedMinute{59};
 constexpr int cMinParsedSecond{0};
-constexpr int cMaxParsedSecond{60};
+constexpr int cMaxParsedSecond{59};
+constexpr int cParsedLeapSecond{60};
```
LinZhihao-723: How about `cLeapSecond`?

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -1207,6 +1216,26 @@ auto parse_timestamp(
...
+                }
+
+                auto const parsed_leap_second{
+                        YSTDLIB_ERROR_HANDLING_TRYX(convert_padded_string_to_number(
+                                timestamp.substr(timestamp_idx, cFieldLength),
+                                '0'
+                        ))
+                };
+                if (cParsedLeapSecond != parsed_leap_second) {
+                    return ErrorCode{ErrorCodeEnum::IncompatibleTimestampPattern};
```
```suggestion
                if (cParsedLeapSecond
                    != YSTDLIB_ERROR_HANDLING_TRYX(convert_padded_string_to_number(
                            timestamp.substr(timestamp_idx, cFieldLength),
                            '0'
                    )))
                {
                    return ErrorCode{ErrorCodeEnum::IncompatibleTimestampPattern};
```
Nit: I feel like it's more clear if we don't have both `parsed_leap_second` and `parsed_second`.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -1472,6 +1501,35 @@ auto parse_timestamp(
...
+                }
+
+                if (cParsedLeapSecond == parsed_generic_second) {
+                    parsed_second = cMaxParsedSecond;
+                    cat_sequence_replacements.emplace_back(pattern_idx, 1ULL, "J");
+                } else {
+                    parsed_second = parsed_generic_second;
+                    cat_sequence_replacements.emplace_back(pattern_idx, 1ULL, "S");
+                }
```
```suggestion
                parsed_second = YSTDLIB_ERROR_HANDLING_TRYX(convert_padded_string_to_number(
                        timestamp.substr(timestamp_idx, cFieldLength),
                        '0'
                ));
                if (parsed_second < cMinParsedSecond || parsed_second > cParsedLeapSecond) {
                    return ErrorCode{ErrorCodeEnum::IncompatibleTimestampPattern};
                }

                if (cParsedLeapSecond == parsed_second) {
                    parsed_second = cMaxParsedSecond;
                    cat_sequence_replacements.emplace_back(pattern_idx, 1ULL, "J");
                } else {
                    cat_sequence_replacements.emplace_back(pattern_idx, 1ULL, "S");
                }
```
Similarly, how about using `parsed_second` directly to avoid having both `parsed_second` and `parsed_generic_second`?

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
```diff
@@ -106,7 +106,8 @@ class TimestampPattern {
  * - \I 12-hour clock, zero-padded hour (01-12).
  * - \l 12-hour clock, space-padded hour ( 1-12).
  * - \M Zero-padded minute (00-59).
- * - \S Zero-padded second (00-60) (60 for leap seconds).
+ * - \S Zero-padded second (00-59).
```
LinZhihao-723: How about:

```suggestion
 * - \S Zero-padded non-leap second (00-59).
```
or:

```suggestion
 * - \S Zero-padded standard second (00-59).
```

gibber9809: I like "non-leap". Standard is ambiguous, I think.

LinZhihao-723: Sure, sgtm.
We might also wanna update the inline comments in the implementation file.

PR #1684, opened 2025-11-27
docs(api-server): Add guide for how to use the API server.

hoophalab:
# Description

1. Add a new sphinx page describing the api server
2. Add a new docs task that generates openapi.json 

Cannot disable try it out feature by the url parameters. The `supportedSubmitMethods` url has no effect on their demo. Also checked redoc ui. It doesn't look as good as swagger...

# Checklist

  breaking change.

# Validation performed

1. `task docs:serve`
2. The user guide page looks correct
3. `http://localhost:8080/_static/openapi.json` is the openapi spec. Cannot verify whether https://petstore.swagger.io/?url=https://docs.yscope.com/clp/main/_static/openapi.json shows correctly yet.

## Summary by CodeRabbit

* **New Features**
  * Added OpenAPI spec generation for the API server and an automated build step to produce the JSON file.
  * Deployment can toggle the API server via an environment-controlled replica setting.

* **Behaviour**
  * API server is automatically disabled for installations using the CLP storage engine; a startup warning is emitted when disabled.

* **Documentation**
  * New user guide explaining API server usage, OpenAPI/Swagger access, examples and streaming query workflows; navigation updated.

### DIFF
--- a/components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
+++ b/components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
@@ -41,7 +41,8 @@ constexpr int cMaxParsedHour12HourClock{12};
 constexpr int cMinParsedMinute{0};
 constexpr int cMaxParsedMinute{59};
 constexpr int cMinParsedSecond{0};
-constexpr int cMaxParsedSecond{60};
+constexpr int cMaxParsedSecond{59};
+constexpr int cLeapSecond{60};
 constexpr int cMinParsedSubsecondNanoseconds{0};
 constexpr int cMinTimezoneOffsetHour{0};
 constexpr int cMaxTimezoneOffsetHour{23};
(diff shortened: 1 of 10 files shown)


components/api-server/src/routes.rs
```diff
@@ -50,15 +50,22 @@ pub fn from_client(client: Client) -> Result<axum::Router, serde_json::Error> {
 // `utoipa::OpenApi` triggers `clippy::needless_for_each`
 #[allow(clippy::needless_for_each)]
 mod api_doc {
+    // Referencing handlers with `super::health` adds an undesired "super" tag to the
+    // `openapi.json`. Import them here to avoid that.
```
```suggestion
    // Using `super::...` can cause `super` to appear as a tag in the generated OpenAPI
    // documentation. Importing the paths directly prevents this issue.
```

taskfiles/docs.yaml
LinZhihao-723, line 1: I think we should explicitly name the task and the output as `api-server-openapi-json`. The other compoenents may also generate openapi standard docs, for example, the MCP server.

docs/src/user-docs/guides-api-server.md
```diff
@@ -0,0 +1,11 @@
...
+
+The API Server exposes a simple, REST-style interface on top of the CLP data-backend. It lets you:
+
+* Submit search queries against CLP data.
+
+* Stream query results via `text/event-stream` format.
+
+The API is defined using the OpenAPI Specification. You can check the API via [Swagger UI][swagger].
+
+[swagger]: https://petstore.swagger.io/?url=https://docs.yscope.com/clp/main/_static/openapi.json
```
LinZhihao-723: I hope there can be a better way to manage this... This hardcodes the path to `main`, while in a release this should be the version number.
I'm not too familiar with the doc structure. Will leave this to @kirkrodrigues

docs/src/user-docs/guides-using-the-api-server.md
LinZhihao-723, line 1: Rewrite this doc to:

* Add a section to explain how to start the server
* Using the existing APIs to make a short example (please help me check whether this example looks good to you). I think enumerating what the API lets users do in this doc won't scale as we add more APIs. Adding this example section might be sufficient. We may or may not need to add more example sections in the future release.

Notice that when I'm trying this example, I realize if I query results to a non-existing query like `curl -N http://localhost:3001/query_results/1234567` it won't fail and instead exit on success without error message. Is there a way we can exit on failures if the query is invalid (not necessarily in this release)? (or maybe check whether this is my setup issue...

[guides-api-server.md](https://github.com/user-attachments/files/23813840/guides-api-server.md)

LinZhihao-723 (changes requested):
Correct me if I'm wrong: I realize API server is probably not specific to clp-json. If this is true, we need to update the API document as my comments in the previous PR for doc rewriting were clp-json specific.

hoophalab:
> Notice that when I'm trying this example, I realize if I query results to a non-existing query like curl -N http://localhost:3001/query_results/1234567 it won't fail and instead exit on success without error message. Is there a way we can exit on failures if the query is invalid (not necessarily in this release)? (or maybe check whether this is my setup issue...

I guess we return a 500 internal server error?
`curl -N  -w "%{http_code}" http://localhost:3001/query_results/1234567` shows 500
Maybe we should also output "500 Internal Server Error" in http body when we return the status code

docs/src/user-docs/guides-using-the-api-server.md
hoophalab: Thanks for polishing the docs! :pray:

LinZhihao-723: Pushed two commits to:
- Add a note to notify users that only clp-json is supported at this moment.
- Move the API reference section before the example section.

docs/src/user-docs/guides-api-server.md
kirkrodrigues: I have a way to make it a little less error-prone. I'll do it in a subsequent PR while modifying a few related things.

docs/src/user-docs/guides-overview.md
LinZhihao-723, line 1: Note to myself: this page needs an update.

taskfiles/docs.yaml
```diff
@@ -111,3 +112,16 @@ tasks:
         vars:
           CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
           INCLUDE_PATTERNS: ["{{.OUTPUT_DIR}}"]
+
+  api-server-openapi-json:
+    vars:
+      OUTPUT_FILE: "{{.G_DOCS_BUILD_DIR}}/_static/openapi.json"
```
LinZhihao-723: We should also rename the output file name to `api-server-openapi.json`.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -810,6 +810,15 @@ def validate_aws_config_dir(self, use_host_mount: bool = False):
                 "aws_config_directory should not be set when profile authentication is not used"
             )
 
+    def validate_api_server(self):
+        is_default = self.api_server == ApiServer()
```
LinZhihao-723: So in clp-text, you can actually uncomment the config to add API server spec with default values, which won't fail the validation and no API server will be started.
Is there a way we can:
* Make the API server config nullable
* For clp-text, we default the config to null; For clp-json, we default the config to what we currently set.

docs/src/user-docs/guides-overview.md
LinZhihao-723: Fixed in my latest commit.

components/clp-package-utils/clp_package_utils/scripts/start_clp.py
```diff
@@ -70,6 +70,10 @@ def main(argv):
         clp_config.validate_data_dir(True)
         clp_config.validate_logs_dir(True)
         clp_config.validate_tmp_dir(True)
+
+        if clp_config.package.storage_engine == StorageEngine.CLP:
+            logger.warning(f"API Server is not supported for storage engine `{StorageEngine.CLP}`")
```
LinZhihao-723:
```suggestion
            logger.warning(f"API Server is not supported for storage engine `{StorageEngine.CLP}`.")
```
To be consistent with other log messages.

LinZhihao-723 (approved):

PR #1704, opened 2025-11-30
feat(log-ingestor): Add partial implementation of `IngestionJobManager` for creating and deleting ingestion jobs.

LinZhihao-723:
# Description

As the title suggests. This PR adds a partial implementation of `IngestionJobManager` for creating and deleting S3 ingestion jobs (using `SqsListener` or `S3Scanner`). This PR includes:

* The method to create new jobs, while the manager will ensure the newly created job is mutually prefix free.
* The method to delete a running job.

A future PR will be required to implement `CompressionJobSubmitter` to actually submit jobs to CLP.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **New Features**
  * Added an ingestion job manager to create and manage S3 scanner and SQS listener jobs with automatic prefix-conflict detection and graceful shutdown.
  * Standardized S3 ingestion configuration across components for simpler job setup.

* **Chores**
  * Updated dependencies: added thiserror, adjusted secrecy entry, removed a dev dependency.
  * Simplified AWS credential handling by removing secret-wrapper usage in client creation.


components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -0,0 +1,228 @@
...
+
+/// Represents an entry in the ingestion job table.
+struct IngestionJobTableEntry {
+    ingestion_job: IngestionJob,
+    listener: Listener,
+    prefix: String,
+}
+
+/// Submitter implementation for creating CLP compression jobs.
+struct CompressionJobSubmitter {}
```
LinZhihao-723, line 228: Not implemented yet.

components/clp-rust-utils/src/job_config/ingestion.rs
LinZhihao-723, line 1: Added in `clp-rust-utils` as we may need this to define API server parameters to create ingestion jobs.

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -0,0 +1,228 @@
...
+    async fn create_s3_client_manager(&self, region: &str) -> S3ClientWrapper {
+        let s3_endpoint = format!("https://s3.{region}.amazonaws.com");
+        let s3_client = create_s3_client(
+            s3_endpoint.as_str(),
+            region,
+            self.aws_access_key_id.as_str(),
+            &self.aws_secret_access_key,
+        )
+        .await;
+        S3ClientWrapper::from(s3_client)
```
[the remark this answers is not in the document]
LinZhihao-723: CLP doesn't support non-default S3 endpoints.

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -0,0 +1,259 @@
...
+        let sqs_endpoint = format!("https://sqs.{region}.amazonaws.com");
+        let sqs_client = create_sqs_client(
+            sqs_endpoint.as_str(),
+            region,
+            self.aws_access_key_id.as_str(),
+            &self.aws_secret_access_key,
+        )
+        .await;
+        SqsClientWrapper::from(sqs_client)
+    }
```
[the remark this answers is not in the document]
LinZhihao-723: Similarly, we don't support non-default endpoint now.

components/clp-rust-utils/src/clp_config/s3_config.rs
```diff
@@ -18,8 +19,61 @@ pub enum AwsAuthentication {
...
+    pub secret_access_key: SecretString,
+}
+
+/// Implement [`PartialEq`] for [`AwsCredentials`] to bypass the constraints of [`SecretString`].
+impl PartialEq for AwsCredentials {
+    fn eq(&self, other: &Self) -> bool {
+        self.access_key_id == other.access_key_id
+            && self.secret_access_key.expose_secret() == other.secret_access_key.expose_secret()
+    }
+}
```
hoophalab: I'm thinking how about not using SecretString at all... Considering that we even pass keys in env variables, avoid copying keys in memory by SecretString doesn't provide any real protection and only causes hassle...

If we don't use SecretString, we don't need to modify job_config at all

components/clp-rust-utils/src/job_config/ingestion.rs
```diff
@@ -0,0 +1,24 @@
+/// Base configuration for ingesting logs from S3.
+#[derive(Clone, Debug)]
+pub struct S3IngestionBaseConfig {
```
hoophalab, line 3: I guess this is going to be serialized into clp-db? Can we reuse/extend S3InputConfig which inherits S3Config? The fields look same except `tags`. If we use an existing job config, we don't need to change any Python code in compression_scheduler/compression_worker

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -0,0 +1,272 @@
...
+            IngestionJobTableEntry {
+                ingestion_job: create_ingestion_job(job_id, sender),
+                listener: job_listener,
+                bucket_name,
+                region,
+                key_prefix,
+                dataset,
+            },
+        );
+        drop(job_table);
```
hoophalab, line 203: probably don't need to drop explicitly,guards are automatically dropped at the end of the scope.

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -0,0 +1,272 @@
...
+            {
+                continue;
+            }
+            return Err(Error::PrefixConflict(format!(
+                "Cannot create ingestion job with prefix '{}' as it conflicts with existing job \
+                 with prefix '{}', which ingests from the same region and bucket into the same \
+                 dataset.",
+                ingestion_job_config.key_prefix, table_entry.key_prefix
+            )));
+        }
```
hoophalab, line 177: 1. A simpler logic might be: if there is a match then return err. Something like:
```
for entry in ... {
    if region == region and bucket == bucket and ...{
        return Err(...)
    }
}
```
2. Do we really need to do key_prefix detection? Otherwise, we should be able to store metadata in a reverse hashmap whose keys are (region, bucket_name, dataset) and values are uuid. This comparison is much faster.

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -0,0 +1,272 @@
...
+/// An async-safe manager for creating and managing ingestion jobs.
+pub struct IngestionJobManager {
+    job_table: Arc<Mutex<HashMap<Uuid, IngestionJobTableEntry>>>,
+    buffer_timeout: Duration,
+    buffer_size_threshold: u64,
+    channel_capacity: usize,
+    aws_credentials: AwsCredentials,
+}
+
+impl IngestionJobManager {
```
hoophalab, line 40: One high-level question: Is it easier to only create one Listener for each job manager and both `s3_scanner` and `sqs_listener` write to that buffer/listener. After all, all data are ingested to a single location: clp-db

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -0,0 +1,272 @@
...
+        let sqs_endpoint = format!("https://sqs.{region}.amazonaws.com");
+        let sqs_client = create_sqs_client(
+            sqs_endpoint.as_str(),
+            region,
+            self.aws_credentials.access_key_id.as_str(),
+            &self.aws_credentials.secret_access_key,
+        )
+        .await;
+        SqsClientWrapper::from(sqs_client)
+    }
```
hoophalab: Can we move these two functions to `src/aws_client_manger.rs`? Something like `pub fn from_singleton(endpoint, region, access_key_id, secret_access_key)`? These two functions technically doesn't belong to ingestion manager

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -0,0 +1,272 @@
...
+            job_id,
+            IngestionJobTableEntry {
+                ingestion_job: create_ingestion_job(job_id, sender),
+                listener: job_listener,
+                bucket_name,
+                region,
+                key_prefix,
+                dataset,
+            },
+        );
```
hoophalab: Does removing local vars look cleaner?
```suggestion
        // At this point, we use one listener per ingestion job. However, the listener itself is
        // designed to be shared among multiple ingestion jobs in the future.
        let job_listener = self.create_listener(ingestion_job_config);
        let sender = job_listener.get_new_sender();
        job_table.insert(
            job_id,
            IngestionJobTableEntry {
                ingestion_job: create_ingestion_job(job_id, sender),
                listener: job_listener,
                bucket_name: ingestion_job_config.bucket_name.clone(),
                region: ingestion_job_config.region.clone(),
                key_prefix: ingestion_job_config.key_prefix.clone(),
                dataset: ingestion_job_config.dataset.clone(),
            },
        );
```

components/log-ingestor/src/ingestion_job.rs
```diff
@@ -1,5 +1,56 @@
 mod s3_scanner;
 mod sqs_listener;
 
+use anyhow::Result;
 pub use s3_scanner::*;
 pub use sqs_listener::*;
+
+/// Enum for different types of ingestion jobs.
+pub enum IngestionJob {
```
hoophalab, line 9: I feel it might be better to
1. make `IngestionJob` a trait,
2. let both s3scanner and sqslistener impl that trait
3. use `Box<dyn IngestionJob>` in the ingestion manager

This saves us a lot of code -- we don't need to implement an additional manual dispatch function `fn XXX(&self) -> {match self {...}}` for each trait functions

components/clp-rust-utils/src/job_config/ingestion.rs
LinZhihao-723: * This won't get into clp-db for job submission. It's only interfacing users to submit ingestion jobs to the server.
* Prefer to have a dedicated struct: it makes API document more clear.

components/log-ingestor/src/ingestion_job_manager.rs
LinZhihao-723: This is to address the clippy linter: it's suggesting me to shortern the lifetime of the guard to reduce lock contention.

components/log-ingestor/src/ingestion_job_manager.rs
LinZhihao-723: 1. I wrote it this way on purpose to reduce the number of indentations required. I can change to your suggestion if you want.
2. I think it's a part of plan to avoid ingesting duplicate logs. Check "Duplication Handling" in this doc: https://docs.google.com/document/d/10jimR0KTzMwiCSUnra1ghIyf5qBckO3GUSR5edXFqPs/edit?tab=t.0

components/log-ingestor/src/ingestion_job_manager.rs
LinZhihao-723: As discussed offline: When submitting jobs to clp-db, it requires the given keys to:
- Belong to the same bucket
- Belong to the same region
- Belong to the same dataset and share the same timestamp key for ingestion
- Share a non-empty common prefix

Inputs from different ingestion jobs will break the assumption above.
The long term plan is to rewrite CLP compression job to be able to handle different input sources and only partition data based on the dataset (and potentially the timestamp key). But for now we don't have better choices but split them manually by sending them from different producers to dedicated consumer listeners...

components/log-ingestor/src/ingestion_job_manager.rs
LinZhihao-723: Ingestion job is moved into `create_listener` so I have to do copies before that.

components/log-ingestor/src/ingestion_job.rs
LinZhihao-723: Had a conversation with Adrian a year ago and I thought this is what he suggested if all supported types are known at the compile time, and only use trait + box if the types are actually dynamic. I agree that using trait + box can reduce the number of code, but personally I'd prefer using enum to:
* Make the underlying memory view more compact.
* Enforce explicity as now the operations on each ingestion job needs to be explicitly handled for all support ingestion job types.
* Reduce virtual function overhead. Technically the super method implemented to the enum is also an extra layer of function call, but at least it's not a virtual function.

It's essentially the same spirit we advocate for C++: use `std::variant` if possible, not inheritence.

components/clp-rust-utils/src/clp_config/s3_config.rs
LinZhihao-723: I agree. Let me drop these and update client creation methods to take a str ref instead.

components/log-ingestor/src/ingestion_job_manager.rs
hoophalab: Not sure if we can take a reference in `create_listener`?

LinZhihao-723: Sure. I will pass a ref instead.

components/log-ingestor/src/aws_client_manager.rs
```diff
@@ -47,6 +51,18 @@ impl SqsClientWrapper {
     pub const fn from(client: SqsClient) -> Self {
         Self { client }
     }
+
+    pub async fn create(region: &str, access_key_id: &str, secret_access_key: &str) -> Self {
+        let sqs_endpoint = format!("https://sqs.{region}.amazonaws.com");
```
hoophalab, line 52: We might need to revisit this later if we want to test ingestion manager on localstack. Not sure if we can accept an endpoint in parameter and in S3IngestionBaseConfig and pass the endpoint here.

But let's merge this as-is for now.

components/log-ingestor/src/aws_client_manager.rs
```diff
@@ -47,6 +51,18 @@ impl SqsClientWrapper {
...
+        let sqs_client = create_sqs_client(
```
hoophalab:
```suggestion
        let sqs_client = sqs::create_new_client(
```

How about not renaming function at the top? Same for s3

hoophalab (dismissed):
lgtm. one nitpick

components/log-ingestor/src/aws_client_manager.rs
LinZhihao-723: Yeah we can figure out how to pass endpoint later. Should be fine.

PR #1708, opened 2025-12-01
feat(clp-s::timestamp_parser): Add support for specifying locale-dependent weekday and month names.

gibber9809:
# Description

This PR removes the `\a` and `\B` patterns, replacing them with `\A{...}` and `\B{...}`, in order to offer support for locale-dependent weekday and month names. `\A{...}` accepts a list of seven comma-separated weekday names starting with Sunday and ending with Saturday. `\B{...}` accepts a list of twelve comma-separated month names starting with January and ending with December.

E.g.,
- `\B{January,February,March,April,May,June,July,August,September,October,November,December}` 
- `\B{Jan,Feb,Mar,Apr,May,Jun,Jul,Aug,Sep,Oct,Nov,Dec}`
- `\A{Sun,Mon,Tue,Wed,Thu,Fri,Sat}`
- `\A{Sunday,Monday,Tuesday,Wednesday,Thursday,Friday,Saturday}`

This PR follows up on some discussion while wrapping up #1673 about how we should make the new timestamp pattern format locale-independent before including it in the archive format. We will also make one more PR that prevents users from including control characters or JSON escape sequences in their timestamp patterns, in order to remove several edge cases.

# Checklist

  breaking change.

# Validation performed

* Added tests verifying that `\A` and `\B` sequences accept valid content
* Modified end-to-end parsing and marshalling tests to use `\A` and `\B` patterns.

## Summary by CodeRabbit

* **New Features**
  * Bracket-style month/day specifiers (e.g., B{...}, A{...}) for customizable month and weekday names, improving date-time formatting and parsing flexibility.

* **Refactor**
  * Replaced fixed name tables with dynamic, pattern-driven name resolution and enriched pattern metadata to drive parsing and marshalling.

* **Bug Fixes**
  * Improved validation and clearer error reporting for invalid bracket patterns and mismatches.

* **Tests**
  * Expanded tests to cover string-based specifiers and bracket-pattern parsing/formatting.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -4,4 +4,3 @@ rustflags = [
     "-Dclippy::nursery",
     "-Dclippy::pedantic",
 ]
-target-dir = "build/rust-targets"
--- a/taskfiles/toolchains.yaml
+++ b/taskfiles/toolchains.yaml
@@ -94,9 +94,12 @@ tasks:
       - |-
         cat <<EOF > "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
         #!/bin/sh
+        # Set up environment for Rust toolchain
         export CARGO_HOME="{{.G_CARGO_HOME}}"
         export RUSTUP_HOME="{{.G_RUSTUP_HOME}}"
         . "{{.G_CARGO_HOME}}/env"
+        # Set up environment variable for Rust build directory
+        export CARGO_TARGET_DIR="{{.G_RUST_BUILD_DIR}}"
         EOF
         chmod +x "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
 


components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
```diff
@@ -94,12 +125,11 @@ class TimestampPattern {
...
  * - \y Zero-padded year in century (69-99 -> 1969-1999, 00-68 -> 2000-2068).
  * - \Y Zero-padded year (0000-9999).
- * - \B Full month name (e.g., January).
- * - \b Abbreviated month name (e.g., Jan).
+ * - \B{January,...,December} Month name, one of twelve in the list between {}.
  * - \m Zero-padded month (01-12).
  * - \d Zero-padded day in month (01-31).
  * - \e Space-padded day in month( 1-31).
- * - \a Abbreviated day in week (e.g., Mon).
+ * - \A{Sunday,...,Saturday} Day in week, one of seven in the list between {}.
```
```suggestion
 * - \A{Sunday,...,Saturday} Day in a week. The names inside `{}` must be provided in order starting
 *   from Sunday (indexed at 0) and include exactly seven entries (Sunday through Saturday)
 *   splitting by `,`.
```
Adding more clarification. Some people may treat Monday as the first day in a week (e.g., people from China), lol

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
```diff
@@ -94,12 +125,11 @@ class TimestampPattern {
  *
  * - \y Zero-padded year in century (69-99 -> 1969-1999, 00-68 -> 2000-2068).
  * - \Y Zero-padded year (0000-9999).
- * - \B Full month name (e.g., January).
- * - \b Abbreviated month name (e.g., Jan).
+ * - \B{January,...,December} Month name, one of twelve in the list between {}.
```
```suggestion
 * - \B{January,...,December} Month in a year. The names inside `{}` must be provided in order
 *   starting from January (indexed at 0) and include exactly twelve entries (January through
 *   December) splitting by `,`.
```
Just to match the clarification we add for week days.

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
```diff
@@ -52,22 +53,52 @@ class TimestampPattern {
...
               m_uses_date_type_representation{uses_date_type_representation},
               m_uses_twelve_hour_clock{uses_twelve_hour_clock} {}
 
     // Variables
     std::string m_pattern;
     std::optional<std::pair<size_t, int>> m_optional_timezone_size_and_offset;
+    std::vector<std::pair<uint16_t, uint16_t>> m_month_name_offsets_and_lengths;
+    std::vector<std::pair<uint16_t, uint16_t>> m_weekday_name_offsets_and_lengths;
+    uint16_t m_month_name_bracket_pattern_length{};
+    uint16_t m_weekday_name_bracket_pattern_length{};
```
LinZhihao-723, line 137: Some design level comment: I might write a class of `DateTypeRepresentation` to wrap these, and make this class optional and drop `m_uses_date_type_representation` flag as we can use the optional value to determine whether it's a daty-type representation. This ensures that `m_uses_data_type_representation` can't be false when the date-related members are provided.
In terms of parsing: I may move `find_first_matching_prefix` as a member function of `TimestampPattern` or `DateTypeRepresentation` as sth like `find_matching_month` or `find_matching_week_day`, which wraps the underlying offset list inside the class itself (working on the offset in the parsing function is a bit hard to read imo)
In terms of marshaling: we can add a member method to return a result of month/day string's view given by the offset, which also avoids directly operating on the offset list.
Open to discussion if u think the current way is preferred.

gibber9809: I like the idea of adding a `find_first_matching_month`/`find_first_matching_weekday`/`get_month`/`get_weekday` function to `TimestampPattern` to make the parsing/marshalling code more readable.

I'm not as sure about the encapsulation inside an optional class -- main reasons being that:
1. The `m_uses_date_type_representation` flag is independent from those other date-time timestamp related properties, in the sense that date time timestamps don't necessarily need to have a timezone, month name, etc.
2.  I think it will make the create function/members of `TimestampPattern` more complicated

Tentatively going to go ahead with just adding the helpers to simplify the parsing/marshalling code, but open to discussing the encapsulation approach more.

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
```diff
@@ -52,22 +53,88 @@ class TimestampPattern {
...
 
+    /**
+     * Finds the first matching month as a prefix of `timestamp`, and advances `pattern_idx` to the
+     * closing bracket of the month name format specifier if found.
+     * @param pattern_idx An index into the raw pattern starting at the "B" in a `\B{}` format
+     * specifier.
+     * @return A result containing the index and length of the first matching month, or an error
+     * code indicating the failure:
+     * - ErrorCodeEnum::IncompatibleTimestampPattern if no matching month is found.
+     */
```
LinZhihao-723, line 64: How about:
```suggestion
    /**
     * Finds the first matching month as a prefix of `timestamp`.
     * @param timestamp
     * @param pattern_idx An index into the raw pattern starting at the "B" in a `\B{}` format
     * specifier. On success, it returns the index of the closing bracket in the raw pattern.
     * @return A result containing the index and length of the first matching month, or an error
     * code indicating the failure:
     * - Forwards `find_first_matching_prefix`'s return values on failure.
     */
```
* `@param timestamp` is missing.
  * Notice that we should order `pattern_idx` after `timestamp`, as it's an output parameter. I think this is a doxygen convention that we follow.
* I move the `advaces ...` part to `pattern_idx`.
* The error documentation should go with `Forwards xxx`. It should still be fine even the function is private.

U can apply it to other methods if it looks good.

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
```diff
@@ -52,22 +53,88 @@ class TimestampPattern {
...
+              m_month_name_bracket_pattern_length{month_name_bracket_pattern_length},
+              m_weekday_name_bracket_pattern_length{weekday_name_bracket_pattern_length},
               m_uses_date_type_representation{uses_date_type_representation},
               m_uses_twelve_hour_clock{uses_twelve_hour_clock} {}
 
     // Variables
     std::string m_pattern;
     std::optional<std::pair<size_t, int>> m_optional_timezone_size_and_offset;
+    std::vector<std::pair<uint16_t, uint16_t>> m_month_name_offsets_and_lengths;
+    std::vector<std::pair<uint16_t, uint16_t>> m_weekday_name_offsets_and_lengths;
```
LinZhihao-723, line 135: With the helpers, it might be worth to store these as a vector of string views for simplicity. It also means we don't need to maintain two versions of `find_first_matching_prefix`.

LinZhihao-723 (changes requested):

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
gibber9809: The problem is that a vector of string views breaks whenever the `TimestampPattern` is moved/copied since the address of the underlying string changes (can change even with move because of the small string optimization). Simpler to just store the offsets and derive the string views when needed.

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
gibber9809: Might do a little bit of rewording, but looks good to me. I'll update the docstrings.

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
LinZhihao-723: oh true I forgot that. My bad.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -550,14 +604,12 @@ auto marshal_date_time_timestamp(
...
-                break;
-            }
-            case 'b': {  // Abbreviated month name.
+            case 'B': {  // Month name.
                 auto const month_idx{year_month_day.month().operator unsigned int() - 1};
-                buffer.append(cAbbreviatedMonthNames.at(month_idx));
+                auto const month_str{YSTDLIB_ERROR_HANDLING_TRYX(
+                        pattern.get_month_and_advance_pattern_idx(month_idx, pattern_idx)
+                )};
+                buffer.append(month_str);
```
LinZhihao-723: Nit: with the try macro we can actually do this without creating a temp variable:
```suggestion
                buffer.append(YSTDLIB_ERROR_HANDLING_TRYX(
                        pattern.get_month_and_advance_pattern_idx(month_idx, pattern_idx)
                ));
```

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -575,13 +627,16 @@ auto marshal_date_time_timestamp(
...
+                auto const weekday_idx{
                         (date::year_month_weekday(timestamp_date).weekday_indexed().weekday()
                          - date::Sunday)
                                 .count()
                 };
-                buffer.append(cAbbreviatedDaysOfWeek.at(day_in_week_idx));
+                auto const weekday_str{YSTDLIB_ERROR_HANDLING_TRYX(
+                        pattern.get_weekday_and_advance_pattern_idx(weekday_idx, pattern_idx)
+                )};
+                buffer.append(weekday_str);
```
```suggestion
                buffer.append(YSTDLIB_ERROR_HANDLING_TRYX(
                        pattern.get_weekday_and_advance_pattern_idx(weekday_idx, pattern_idx)
                ));
```
Same above

LinZhihao-723 (dismissed):
Two nit comment otherwise lgtm.
The PR title also lgtm.

LinZhihao-723 (approved):

PR #1710, opened 2025-12-02
feat(clp-s): Add `CLP_BUILD_CLP_S_ENABLE_CURL` build option to exclude libcurl/libcrypto paths for WebAssembly compatibility; Improve Curl error handling utilities.

Bill-hbrhbr:
# Description

This PR adds a compile-time option, `CLP_S_DISABLE_CURL`, to cleanly build clp-s without curl-dependent code paths.

The primary goal is Emscripten compatibility. Since `USE_CURL` is no longer supported in modern emsdk versions and building curl for wasm is unnecessary for our current browser workflow, we decouple curl from clp-s so wasm builds can succeed without network stack dependencies.

Even if network-fetch functionality is needed later, it's generally recommended to implement it at the clp-ffi-js/Emscripten boundary (JS + wasm integration layer), rather than using curl-based networking inside C++ code paths. 

Curl libraries have also historically been a recurring source of cross-toolchain build and link issues, so this change is useful beyond wasm and can be reused whenever upstream or downstream projects need a curl-free clp-s build.

With `CLP_S_DISABLE_CURL` enabled, all code that depends on `CURL_LIBRARIES` is conditionally excluded. This disables network-based input paths in clp-s, while preserving core archive functionality needed by downstream clp-ffi-js.

Result:
  - No curl requirement for wasm builds
  - Reduced dependency surface for browser-targeted builds
  - Core clp-s archive/read/search functionality remains available for downstream clp-ffi-js and clp-ffi-py

# Checklist

  breaking change.

# Validation performed

* [X] Compile `clp-s` in manylinux and verifies the exe works on the following platforms:
  * Debian 10+
  * Ubuntu 18.10+
  * Fedora 29+
  * CentOS/RHEL 8+

## Summary by CodeRabbit

* **New Features**
  * Option to build a simplified, statically linked clp-s executable with toggles to exclude network input, exclude MongoDB-backed output, and enable a static C++ runtime.
  * Added ZSTD dependency support for compression compatibility.

* **Changes**
  * Network and DB integrations can be omitted at build time; excluded features produce clear errors if invoked.
  * New static-core build workflow and build-directory variable to produce static artifacts.
  * Dependency install prefix is now overridable for flexible install paths.

---
  - https://app.asana.com/0/0/1212888440703770

### DIFF
--- a/components/clp-package-utils/clp_package_utils/controller.py
+++ b/components/clp-package-utils/clp_package_utils/controller.py
@@ -196,14 +196,17 @@ class BaseController(ABC):
 
         if BundledService.DATABASE not in self._clp_config.bundled:
             env_vars |= {
-                "CLP_DB_PORT": str(self._clp_config.database.port),
+                "CLP_DB_CONNECT_PORT": str(self._clp_config.database.port),
                 "CLP_EXTRA_HOST_DATABASE_NAME": DB_COMPONENT_NAME,
                 "CLP_EXTRA_HOST_DATABASE_ADDR": _resolve_external_host(
                     self._clp_config.database.host
                 ),
             }
         else:
-            env_vars["CLP_DB_HOST"] = _get_ip_from_hostname(self._clp_config.database.host)
+            env_vars |= {
+                "CLP_DB_HOST": _get_ip_from_hostname(self._clp_config.database.host),
+                "CLP_DB_PORT": str(self._clp_config.database.port),
+            }
 
         # Credentials
         credentials = self._clp_config.database.credentials
@@ -273,12 +276,15 @@ class BaseController(ABC):
         # Connection config
         if BundledService.QUEUE not in self._clp_config.bundled:
             env_vars |= {
-                "CLP_QUEUE_PORT": str(self._clp_config.queue.port),
+                "CLP_QUEUE_CONNECT_PORT": str(self._clp_config.queue.port),
                 "CLP_EXTRA_HOST_QUEUE_NAME": QUEUE_COMPONENT_NAME,
                 "CLP_EXTRA_HOST_QUEUE_ADDR": _resolve_external_host(self._clp_config.queue.host),
             }
         else:
-            env_vars["CLP_QUEUE_HOST"] = _get_ip_from_hostname(self._clp_config.queue.host)
+            env_vars |= {
+                "CLP_QUEUE_HOST": _get_ip_from_hostname(self._clp_config.queue.host),
+                "CLP_QUEUE_PORT": str(self._clp_config.queue.port),
+            }
 
         # Credentials
         env_vars |= {
@@ -359,12 +365,15 @@ class BaseController(ABC):
         # Connection config
         if BundledService.REDIS not in self._clp_config.bundled:
             env_vars |= {
-                "CLP_REDIS_PORT": str(self._clp_config.redis.port),
+                "CLP_REDIS_CONNECT_PORT": str(self._clp_config.redis.port),
                 "CLP_EXTRA_HOST_REDIS_NAME": REDIS_COMPONENT_NAME,
                 "CLP_EXTRA_HOST_REDIS_ADDR": _resolve_external_host(self._clp_config.redis.host),
             }
         else:
-            env_vars["CLP_REDIS_HOST"] = _get_ip_from_hostname(self._clp_config.redis.host)
+            env_vars |= {
+                "CLP_REDIS_HOST": _get_ip_from_hostname(self._clp_config.redis.host),
+                "CLP_REDIS_PORT": str(self._clp_config.redis.port),
+            }
 
         # Credentials
         env_vars |= {
@@ -463,16 +472,19 @@ class BaseController(ABC):
         }
         if BundledService.RESULTS_CACHE not in self._clp_config.bundled:
             env_vars |= {
-                "CLP_RESULTS_CACHE_PORT": str(self._clp_config.results_cache.port),
+                "CLP_RESULTS_CACHE_CONNECT_PORT": str(self._clp_config.results_cache.port),
                 "CLP_EXTRA_HOST_RESULTS_CACHE_NAME": RESULTS_CACHE_COMPONENT_NAME,
                 "CLP_EXTRA_HOST_RESULTS_CACHE_ADDR": _resolve_external_host(
                     self._clp_config.results_cache.host
                 ),
             }
         else:
-            env_vars["CLP_RESULTS_CACHE_HOST"] = _get_ip_from_hostname(
-                self._clp_config.results_cache.host
-            )
+            env_vars |= {
+                "CLP_RESULTS_CACHE_HOST": _get_ip_from_hostname(
+                    self._clp_config.results_cache.host
+                ),
+                "CLP_RESULTS_CACHE_PORT": str(self._clp_config.results_cache.port),
+            }
 
         return env_vars
 
--- a/tools/deployment/package/docker-compose-all.yaml
+++ b/tools/deployment/package/docker-compose-all.yaml
@@ -233,7 +233,7 @@ services:
       "-u",
       "-m", "clp_py_utils.initialize-results-cache",
       "--uri", "mongodb://${CLP_EXTRA_HOST_RESULTS_CACHE_ADDR:-results_cache}\
-        :${CLP_RESULTS_CACHE_PORT:-27017}/${CLP_RESULTS_CACHE_DB_NAME:-clp-query-results}",
+        :${CLP_RESULTS_CACHE_CONNECT_PORT:-27017}/${CLP_RESULTS_CACHE_DB_NAME:-clp-query-results}",
       "--stream-collection", "${CLP_RESULTS_CACHE_STREAM_COLLECTION_NAME:-stream-files}",
     ]
 
@@ -253,7 +253,7 @@ services:
       "/opt/clp/bin/spider_scheduler",
       "--host", "spider_scheduler",
       "--port", "6000",
-      "--storage_url", "jdbc:mariadb://database:${CLP_DB_PORT:-3306}/\
+      "--storage_url", "jdbc:mariadb://database:${CLP_DB_CONNECT_PORT:-3306}/\
         ${SPIDER_DB_NAME:-spider-db}?\
         user=${SPIDER_DB_USER:-spider-user}\
         &password=${SPIDER_DB_PASS:?Please set a value.}",
@@ -265,14 +265,14 @@ services:
     stop_grace_period: "300s"
     environment:
       BROKER_URL: "amqp://${CLP_QUEUE_USER:-clp-user}:${CLP_QUEUE_PASS:?Please set a value.}\
-        @queue:${CLP_QUEUE_PORT:-5672}"
+        @queue:${CLP_QUEUE_CONNECT_PORT:-5672}"
       CLP_DB_PASS: "${CLP_DB_PASS:?Please set a value.}"
       CLP_DB_USER: "${CLP_DB_USER:-clp-user}"
       CLP_LOGGING_LEVEL: "${CLP_COMPRESSION_SCHEDULER_LOGGING_LEVEL:-INFO}"
       CLP_LOGS_DIR: "/var/log/compression_scheduler"
       PYTHONPATH: "/opt/clp/lib/python3/site-packages"
       RESULT_BACKEND: "redis://default:${CLP_REDIS_PASS:?Please set a value.}\
-        @redis:${CLP_REDIS_PORT:-6379}\
+        @redis:${CLP_REDIS_CONNECT_PORT:-6379}\
         /${CLP_REDIS_BACKEND_DB_COMPRESSION:-1}"
     volumes:
       - *volume_clp_config_readonly
@@ -298,7 +298,7 @@ services:
     hostname: "compression_worker"
     environment:
       BROKER_URL: "amqp://${CLP_QUEUE_USER:-clp-user}:${CLP_QUEUE_PASS:?Please set a value.}\
-        @queue:${CLP_QUEUE_PORT:-5672}"
+        @queue:${CLP_QUEUE_CONNECT_PORT:-5672}"
       CLP_CONFIG_PATH: "/etc/clp-config.yaml"
       CLP_HOME: "/opt/clp"
       CLP_LOGGING_LEVEL: "${CLP_COMPRESSION_WORKER_LOGGING_LEVEL:-INFO}"
@@ -306,7 +306,7 @@ services:
       CLP_WORKER_LOG_PATH: "/var/log/compression_worker/worker.log"
       PYTHONPATH: "/opt/clp/lib/python3/site-packages"
       RESULT_BACKEND: "redis://default:${CLP_REDIS_PASS:?Please set a value.}\
-        @redis:${CLP_REDIS_PORT:-6379}\
+        @redis:${CLP_REDIS_CONNECT_PORT:-6379}\
         /${CLP_REDIS_BACKEND_DB_COMPRESSION:-1}"
     volumes:
       - *volume_clp_config_readonly
@@ -356,7 +356,7 @@ services:
       "-m", "job_orchestration.executor.start-spider-worker",
       "--host", "compression_worker",
       "--num-workers", "${CLP_COMPRESSION_WORKER_CONCURRENCY:-1}",
-      "--storage-url", "jdbc:mariadb://database:${CLP_DB_PORT:-3306}/\
+      "--storage-url", "jdbc:mariadb://database:${CLP_DB_CONNECT_PORT:-3306}/\
         ${SPIDER_DB_NAME:-spider-db}?\
         user=${SPIDER_DB_USER:-spider-user}\
         &password=${SPIDER_DB_PASS:?Please set a value.}",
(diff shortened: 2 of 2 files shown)


components/core/src/clp_s/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+option(CLP_S_STATIC_EXE "Enable clp-s static exe" OFF)
```
Bill-hbrhbr:
```suggestion
option(CLP_S_STATIC_EXE "Build clp-s as a statically linked executable." OFF)
```

gibber9809: If `CLP_USE_STATIC_LIBS` is `OFF` doesn't this flag not actually result in static linking? Seems like what this flag does specifically is exclude, libcurl, mongocxx and related libraries, and statically link clp-s against libgcc and libstdc++.

Since this flag is doing three different things that aren't all directly related to making the clp-s executable statically linked, I think they should be three separate flags.

E.g.,
CLP_S_EXCLUDE_LIBCURL
CLP_S_EXCLUDE_MONGOCXX
CLP_S_STATIC_LIBSTD

or something similar.

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -75,6 +75,9 @@ ResultsCacheOutputHandler::ResultsCacheOutputHandler(
         : ::clp_s::search::OutputHandler(should_output_timestamp, true),
           m_batch_size(batch_size),
           m_max_num_results(max_num_results) {
+#if CLP_S_STATIC_EXE
```
LinZhihao-723: Before we move forward: we should use the macro to disable the entire function if this whole function should be enabled/disabled. For example:
```C++
#if CLP_S_STATIC_EXE
    auto foo() -> void { throw ... }
#else
    auto foo() -> void { /*actual implementation*/ }
#endif
```

Also, I think it's worth to derive a dedicated exception (ideally from ystdlib) for disabled features.

LinZhihao-723:
> Besides the comment about the flags I think we also need to hide all of the options that don't work from `CommandLineArguments.cpp`, since it's pretty confusing for the binary to display a bunch of options that don't work in the help text.
> 
> That means hiding:
> 
> * All of the `auth` options
> * All of the mongo db related options/the entire sections they are in
> * All of the examples/help text referencing these options
> 
> As well, I'm wondering why we have to exclude libcurl? If it's just too much work to get the linking working in the short term that's fine I guess, but it is a pretty significant limitation, so we should try to fix it if possible.
> 
> I'm also not totally sure about the style for how we're excluding code using macros. Will give it some more thought for the next round of review.

@gibber9809  In terms of excluding `libcurl`:

Yeah it's the linking issue. `libcurl` has too many dependencies and it's almost impossible to build it as a static library in a maintainable way. As discussed with Kirk offline, we should probably switch to a more maintainable curl substitution that can be statically linked, and use that thing to reimplement the network reader.

components/core/src/clp_s/OutputHandlerImpl.cpp
Bill-hbrhbr: Is the dedicated exception gonna be based on `TraceableException`?

components/core/src/clp_s/clp-s.cpp
```diff
@@ -254,13 +255,17 @@ bool search_archive(
...
+                throw std::runtime_error("Simplified static clp-s executable does not support mongocxx.");
+#else
                 output_handler = std::make_unique<clp_s::ResultsCacheOutputHandler>(
                         command_line_arguments.get_mongodb_uri(),
                         command_line_arguments.get_mongodb_collection(),
                         command_line_arguments.get_batch_size(),
                         command_line_arguments.get_max_num_results()
                 );
                 break;
+#endif
```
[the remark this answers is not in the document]
Bill-hbrhbr: We plan to add a dedicated exception type later in `ystdlib-cpp`

taskfile.yaml
```diff
@@ -206,6 +207,37 @@ tasks:
...
+      - "deps:core"
+    cmds:
+      - task: "utils:cmake:generate"
+        vars:
+          BUILD_DIR: "{{.G_CORE_STATIC_COMPONENT_BUILD_DIR}}"
+          EXTRA_ARGS:
+            - "-DCLP_S_BUILD_SIMPLIFIED_STATIC_EXECUTABLE=ON"
+            - "-DCLP_USE_STATIC_LIBS=ON"
+          SOURCE_DIR: "{{.G_CORE_COMPONENT_DIR}}"
+
```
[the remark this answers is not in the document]
Bill-hbrhbr: @gibber9809 Do we need to generate parsers for clp-s static build?

Bill-hbrhbr:
@coderabbitai review

components/core/src/clp_s/CMakeLists.txt
```diff
@@ -3,23 +3,20 @@ add_subdirectory(log_converter)
 add_subdirectory(search)
 add_subdirectory(timestamp_parser)
 
+option(
+        CLP_S_EXCLUDE_LIBCURL
+        "Exclude libcurl support from clp-s."
+        ON
+)
+
```
gibber9809: Could we move this to `components/core/cmake/Options/options.cmake` and make it `OFF` by default?

I think we want to keep the default behaviour as including curl support.

components/core/src/clp_s/InputConfig.cpp
```diff
@@ -155,6 +159,15 @@ auto try_create_file_reader(std::string_view const file_path)
     }
 }
 
+#if CLP_S_EXCLUDE_LIBCURL
+auto try_create_network_reader(std::string_view const url, NetworkAuthOption const& auth)
+        -> std::shared_ptr<clp::ReaderInterface> {
+    std::ignore = url;
+    std::ignore = auth;
+    SPDLOG_ERROR("This build of clp-s does not support network reading (libcurl excluded).");
```
gibber9809:
```suggestion
    SPDLOG_ERROR("This build of clp-s does not support network inputs (libcurl excluded).");
```

components/core/src/clp_s/Utils.hpp
```diff
@@ -47,6 +49,19 @@ class FileUtils {
...
 };
 
+class NetworkUtils {
+public:
+    /**
+     * Checks if a reader is a clp::NetworkReader that has encountered a CURL error and logs
+     * relevant CURL error information if a CURL error has occurred.
+     * @param path The path that the reader has opened.
+     * @param reader The open reader which may have experienced a CURL error.
+     * @return true if a CURL error has occurred on the reader, false otherwise.
```
gibber9809:
```suggestion
     * Checks if a reader is a clp::NetworkReader that has encountered a CURL error, and logs
     * relevant CURL error information if a CURL error has occurred.
     * @param path The path that the reader has opened.
     * @param reader The open reader which may have experienced a CURL error.
     * @return Whether a CURL error has occurred on the reader.
     ```
     I know that this is copied from the previous docstring, but we may as well improve it since it's part of the diff.

File: components/core/src/clp_s/Utils.cpp [hunk 8]
```diff
@@ -149,6 +152,37 @@ bool FileUtils::get_last_non_empty_path_component(std::string_view const path, s
...
+        -> bool {
+    std::ignore = path;
+    std::ignore = reader;
+    return false;
+}
+#else
+auto
+NetworkUtils::check_and_log_curl_error(std::string_view path, clp::ReaderInterface const* reader)
+        -> bool {
+    auto const* network_reader = dynamic_cast<clp::NetworkReader const*>(reader);
```
gibber9809:
```suggestion
    auto const* network_reader{dynamic_cast<clp::NetworkReader const*>(reader)};
```

components/core/src/clp_s/Utils.cpp
```diff
@@ -11,6 +11,9 @@
 #include <spdlog/spdlog.h>
 #include <string_utils/string_utils.hpp>
 
+#if !CLP_S_EXCLUDE_LIBCURL
+    #include "../clp/NetworkReader.hpp"
```
gibber9809, line 15:
```suggestion
    #include <clp/NetworkReader.hpp>
```

components/core/src/clp_s/Utils.hpp
```diff
@@ -10,6 +10,8 @@
 #include <string_view>
 #include <vector>
 
+#include "../clp/ReaderInterface.hpp"
```
gibber9809, line 13:
```suggestion
#include <clp/ReaderInterface.hpp>
```

components/core/src/clp/NetworkReader.hpp
```diff
@@ -66,6 +66,14 @@ class NetworkReader : public ReaderInterface {
         Finished
     };
 
+    /**
+     * CURL error details set by the underlying CURL handler.
+     */
+    struct CurlErrorInfo {
+        CURLcode code;
+        std::string_view message;
+    };
```
LinZhihao-723: Prefer to use a struct since that's the only way to make the underlying members immutable in C++.

components/core/src/clp/NetworkReader.hpp
```diff
@@ -224,16 +232,18 @@ class NetworkReader : public ReaderInterface {
...
+    [[nodiscard]] auto get_curl_error_info() const -> std::optional<CurlErrorInfo> {
+        auto const ret_code = get_curl_ret_code();
+        if (false == ret_code.has_value() || CURLcode::CURLE_OK == ret_code.value()) {
             return std::nullopt;
         }
-        return std::string_view{m_curl_error_msg_buf->data()};
+        return CurlErrorInfo{
+                .code = ret_code.value(),
+                .message = std::string_view{m_curl_error_msg_buf->data()}
+        };
```
LinZhihao-723: I would highly recommend to not using aggregate initialization in C++ (even if it's added to standard in C++20) when u have a chance to define a proper constructor.
Aggregate initialization allows fields to be partially initialized (as shown in the [doc](https://en.cppreference.com/w/cpp/language/aggregate_initialization.html)), while the remaining fields will be left uninitialized.

components/core/src/clp_s/Utils.cpp
```diff
@@ -149,6 +152,37 @@ bool FileUtils::get_last_non_empty_path_component(std::string_view const path, s
     return false;
 }
 
+#if CLP_S_EXCLUDE_LIBCURL
+auto
+NetworkUtils::check_and_log_curl_error(std::string_view path, clp::ReaderInterface const* reader)
+        -> bool {
+    std::ignore = path;
+    std::ignore = reader;
```
LinZhihao-723:
```suggestion
auto NetworkUtils::check_and_log_curl_error(
        [[maybe_unused]] std::string_view path,
        [[maybe_unused]] clp::ReaderInterface const* reader
) -> bool {
```
I think `[[maybe_unused]]` should be the correct way to get off the clang-tidy errors.

components/core/src/clp_s/InputConfig.cpp
```diff
@@ -14,11 +14,15 @@
 #include <simdjson.h>
 #include <spdlog/spdlog.h>
 
-#include "../clp/aws/AwsAuthenticationSigner.hpp"
+#if !CLP_S_EXCLUDE_LIBCURL
+    #include "../clp/aws/AwsAuthenticationSigner.hpp"
+#endif
```
LinZhihao-723: Is it possible to group all CURL-related headers? I think that would be more readable tbh.

LinZhihao-723: Looks like this would pass clang-format:
```C++
#include "../clp/BufferedReader.hpp"
#include "../clp/ffi/ir_stream/protocol_constants.hpp"
#include "../clp/FileReader.hpp"
#include "../clp/ReaderInterface.hpp"
#include "../clp/spdlog_with_specializations.hpp"
#include "../clp/streaming_compression/Decompressor.hpp"
#include "../clp/streaming_compression/zstd/Decompressor.hpp"
#include "../clp/utf8_utils.hpp"
#include "Utils.hpp"

#if !CLP_S_EXCLUDE_LIBCURL
    #include "../clp/aws/AwsAuthenticationSigner.hpp"
    #include "../clp/NetworkReader.hpp"
#endif
```

components/core/src/clp_s/InputConfig.cpp
```diff
@@ -155,6 +159,15 @@ auto try_create_file_reader(std::string_view const file_path)
     }
 }
 
+#if CLP_S_EXCLUDE_LIBCURL
+auto try_create_network_reader(std::string_view const url, NetworkAuthOption const& auth)
```
LinZhihao-723: Similarly, we can use `[[maybe_unused]]`

components/core/src/clp_s/Utils.cpp
LinZhihao-723: We have a few more uses of `"../clp/NetworkReader.hpp"`. I think they should all be replaced by `<clp/NetworkReader.hpp>`?

components/core/src/clp_s/Utils.cpp
Bill-hbrhbr: I think auto = is the preferred way? we use {} for explicitly typed vars

LinZhihao-723: `{}` is always preferred. But for `auto` it's not mandatory.

components/core/cmake/Options/options.cmake
```diff
@@ -95,6 +95,12 @@ option(
     ON
 )
 
+option(
+    CLP_BUILD_CLP_S_ENABLE_CURL
```
gibber9809, line 99: Sorry I didn't mean that you had to change the name of the flag/treat it as a target somehow, I just meant that you should just move it here so that it is with our other options.

Though I guess you still need to conditionally change whether curl is required for certain clp-s targets based on this flag.

Do you think that approach (using `CLP_S_DISABLE_CURL` flag + conditionally changing the needs dependencies on the rest of the targets) or your approach here makes more sense?

components/core/src/clp_s/CMakeLists.txt
```diff
@@ -140,14 +147,20 @@ if(CLP_BUILD_CLP_S_CLP_DEPENDENCIES)
...
                 nlohmann_json::nlohmann_json
-                OpenSSL::Crypto
                 spdlog::spdlog
                 ystdlib::error_handling
         )
+        if(CLP_BUILD_CLP_S_ENABLE_CURL)
+                target_link_libraries(
+                        clp_s_clp_dependencies
+                        PRIVATE
+                        ${CURL_LIBRARIES}
```
gibber9809, line 160: I think `${CURL_LIBRARIES}` needs to be public since `NetworkReader.hpp` directly includes `<curl/curl.h>`

components/core/cmake/Options/options.cmake
Bill-hbrhbr: 1. I think using `CLP_S_DISABLE_CURL` is fundamentally the same as `CLP_S_ENABLE_CURL`.
2. 

> Though I guess you still need to conditionally change whether curl is required for certain clp-s targets based on this flag.

If you’re referring to removing the hard requirement for `curl` and `openssl` that previously applied to the clp-s binary targets, I think keeping those dependency logic would just be duplicating
https://github.com/Bill-hbrhbr/clp/blob/09d2cf5bd83387abe0bb586381075ca860b4f862/components/core/cmake/Options/options.cmake#L482

components/core/src/clp/NetworkReader.hpp
LinZhihao-723: Sorry this is my mistake. What I really meant was to use a class. I don't know why clang-tidy didn't complain but this should violate [this](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html) rule.

components/core/src/clp_s/Utils.hpp
```diff
@@ -47,6 +49,19 @@ class FileUtils {
...
+class NetworkUtils {
+public:
+    /**
+     * Checks if a reader is a clp::NetworkReader that has encountered a CURL error, and logs
+     * relevant CURL error information if a CURL error has occurred.
+     * @param path The path that the reader has opened.
+     * @param reader The open reader which may have experienced a CURL error.
+     * @return Whether a CURL error has occurred on the reader.
+     */
+    [[nodiscard]] static auto
```
LinZhihao-723: Sorry for missing this in the last round. I think you can mark it `maybe_unused` as well, then u don't need `std::ignore = check_and_log_curl_error` anymore. I believe this is the suggested way to make the return value optional to use.
```suggestion
    [[maybe_unused]] static auto
```

components/core/src/clp_s/Utils.hpp
```diff
@@ -47,6 +49,19 @@ class FileUtils {
     static bool get_last_non_empty_path_component(std::string_view const path, std::string& name);
 };
 
+class NetworkUtils {
+public:
+    /**
+     * Checks if a reader is a clp::NetworkReader that has encountered a CURL error, and logs
```
LinZhihao-723:
```suggestion
     * Checks if a reader is a `clp::NetworkReader` that has encountered a CURL error, and logs
```

components/core/src/clp_s/log_converter/log_converter.cpp
```diff
@@ -86,7 +56,8 @@ auto convert_files(CommandLineArguments const& command_line_arguments) -> bool {
             case clp_s::FileType::Zstd:
             case clp_s::FileType::Unknown:
             default: {
-                std::ignore = check_and_log_curl_error(path, reader.get());
+                std::ignore
+                        = clp_s::NetworkUtils::check_and_log_curl_error(path.path, reader.get());
```
Bill-hbrhbr:
```suggestion
                clp_s::NetworkUtils::check_and_log_curl_error(path.path, reader.get());
```

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -682,15 +681,17 @@ bool JsonParser::ingest() {
             case FileType::Zstd:
             case FileType::Unknown:
             default: {
-                std::ignore = check_and_log_curl_error(path, reader);
+                std::ignore = NetworkUtils::check_and_log_curl_error(path.path, reader.get());
```
Bill-hbrhbr:
```suggestion
                NetworkUtils::check_and_log_curl_error(path.path, reader.get());
```

LinZhihao-723 (approved):

PR #1713, opened 2025-12-02
build(rust): Configure Rust build target directory by setting `CARGO_TARGET_DIR` in Rust toolchain's env file.

LinZhihao-723:
# Description

Before this PR, we configure Rust's target directory using a config option in `.cargo/config.toml`. This has a problem when developers want to change where to store the generated artifacts: we use `G_BUILD_DIR` in the root taskfile to specify the build directory for the entire CLP repo. Modifying this value won't reflect Rust's build directory as `.cargo/config.toml` is not synchronized with the taskfile. To solve this problem, we remove the config from `.cargo/config.toml`, and instead we specify the build target directory by setting `CARGO_TARGET_DIR` inside the task-generated Rust toolchain file. This environment variable ensures that the build behavior remains the same as using `.cargo/config.toml`, while it can be properly synchronized with the root taskfile. Since sourcing the env file will be required for all Rust toolchain calls (including build, linting, and testing), this environment variable will be applied to all subsequent cargo execution automatically.

# Checklist

  breaking change.

# Validation performed

* [x] Ensure the default build behavior remains unchanged before/after this PR.
* [x] Override `G_BUILD_DIR` to somewhere else and ensure Rust targets are successfully built.

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration and development toolchain environment setup for improved consistency and code quality standards.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2241,6 +2241,8 @@ dependencies = [
  "async-trait",
  "aws-sdk-s3",
  "aws-sdk-sqs",
+ "axum",
+ "clap",
  "clp-rust-utils",
  "secrecy",
  "serde_json",
@@ -2248,6 +2250,9 @@ dependencies = [
  "stdext",
  "tokio",
  "tokio-util",
+ "tracing",
+ "tracing-appender",
+ "tracing-subscriber",
  "uuid",
 ]
 
--- a/components/log-ingestor/Cargo.toml
+++ b/components/log-ingestor/Cargo.toml
@@ -5,14 +5,19 @@ edition = "2024"
 
 [dependencies]
 anyhow = "1.0.100"
+axum = { version = "0.8.6", features = ["json"] }
 async-trait = "0.1.89"
 aws-sdk-s3 = "1.110.0"
 aws-sdk-sqs = "1.89.0"
+clap = { version = "4.5.51", features = ["derive"] }
 clp-rust-utils = { path = "../clp-rust-utils" }
 secrecy = { version = "0.10.3", features = ["serde"] }
 serde_json = "1.0.145"
-tokio = { version = "1.48.0", features = ["macros", "rt-multi-thread", "time"] }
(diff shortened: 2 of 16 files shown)


taskfiles/toolchains.yaml
```diff
@@ -94,9 +94,12 @@ tasks:
...
         cat <<EOF > "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
         #!/bin/sh
+        # Set up environment for Rust toolchain
         export CARGO_HOME="{{.G_CARGO_HOME}}"
         export RUSTUP_HOME="{{.G_RUSTUP_HOME}}"
         . "{{.G_CARGO_HOME}}/env"
+        # Set up environment variable for Rust build directory
+        export CARGO_TARGET_DIR="{{.G_RUST_BUILD_DIR}}"
         EOF
         chmod +x "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
```
junhaoliao, line 104: unrelated to the PR - do we have to make this file executable if it is only to be `source`d?

LinZhihao-723:
> for the title, how about:
> 
> ```
> build(rust): Remove hardcoded `target-dir` cargo config and use Taskfile parameter to control Rust build output directory.
> ```

I prefer the current title as it's more precise. Problem with your suggestion:
* `target-dir` isn't hardcoded. It's actually what's suggested by cargo.
* It doesn't explain how to use taskfile parameters to control the build behavior which sounds a bit general.

taskfiles/toolchains.yaml
LinZhihao-723: You are right. Shall I remove that permission in this PR?

junhaoliao: maybe file an issue and fix separately? (i.e., i can't think of a shorter PR title if we also change it in the same PR lol

@coderabbitai come to help

junhaoliao: @coderabbitai come to help filing an issue i mean, please

PR #1723, opened 2025-12-03
feat(clp-s::timestamp_parser): Make control characters and lone double quotes illegal in timestamp patterns.

gibber9809:
# Description

This PR is the last in the series of PRs readying `clp_s::timestamp_parser` for use in clp-s (#1708, #1673, #1646, #1626, #1599, #1564, and #1385).

This PR makes several characters and escape sequences illegal in timestamp patterns, as well as fixes parsing and marshalling behaviour for the `\\` escape sequence, in order to put timestamp patterns in a state where:

1. We can use timestamp patterns to parse and marshal JSON literals and raw UTF-8, using a flag to indicate the content type
2. We can add support for JSON escape sequences in a backwards compatible way over time

The banned characters and escape sequences are:

1. Control characters (`\x00` - `\x1F`)
2. Double quote `"` (unless it is both the first and last character in a timestamp pattern, then the start and end quote are allowed)
3. All JSON escape sequences except `\\` (`\uXXXX`, `\b`, `\f`, `\n`, `\r`, `\t`)

Section 2.4 of the timestamp format doc has been updated to reflect these changes.

# Checklist

  breaking change.

# Validation performed

* Added test case to validate that illegal characters and escape sequences are not allowed in timestamp patterns
* Added test case to validate that timestamp patterns containing JSON escape sequences can be used to parse and marshal JSON literals
* Added test case to validate that JSON escape sequences work as expected for parsing both JSON literals and raw UTF-8

## Summary by CodeRabbit

* **New Features**
  * JSON-style escaped backslashes supported when parsing and marshaling timestamps.
  * Dynamic generation of timestamp patterns for variable components.
  * New parsing-mode flag to toggle JSON-literal vs raw handling and small public helpers to validate escapes/characters.

* **Bug Fixes**
  * Stronger validation rejecting unsupported escape sequences and illegal characters with distinct error classifications.
  * Safer string handling in pattern processing.

* **Documentation**
  * Clarifies JSON-literal behaviour and new error cases.

* **Tests**
  * New and expanded tests for JSON escapes and invalid-pattern cases.

### DIFF
--- a/components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
+++ b/components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
@@ -206,8 +206,8 @@ private:
  * Any escape sequence not listed above is invalid. In addition, patterns must not contain:
  *
  * - ASCII control characters (U+0000 through U+001F).
- * - The double-quote character `"` (except for the opening an closing quote if the whole pattern is
- *   written like a JSON string literal e.g. `"<pattern>"`).
+ * - The double-quote character `"` (except for the surrounding quotes if the pattern is provided as
+ *   a JSON string literal, e.g., `"<pattern>"`).
  *
  * @param timestamp
  * @param pattern A timestamp pattern made up of literals, format specifiers, and potentially CAT


components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -861,18 +911,33 @@ auto TimestampPattern::create(std::string_view pattern)
...
+            pattern.size() >= 2 && '"' == pattern.front() && '"' == pattern.back()
+    };
+    size_t const start_idx{quoted_pattern ? 1ULL : 0ULL};
+    size_t const end_idx{quoted_pattern ? pattern.size() - 1ULL : pattern.size()};
     bool escaped{false};
-    for (size_t pattern_idx{0ULL}; pattern_idx < pattern.size(); ++pattern_idx) {
+    for (size_t pattern_idx{start_idx}; pattern_idx < end_idx; ++pattern_idx) {
         auto const cur_format_specifier{pattern.at(pattern_idx)};
+        if (is_unsupported_character(cur_format_specifier)) {
+            return ErrorCode{ErrorCodeEnum::InvalidTimestampPattern};
```
LinZhihao-723: How about we create two special error code for unsupported control characters and unsupported escape sequence?

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -1676,10 +1746,19 @@ auto parse_timestamp(
...
+
+                if (false == is_json_literal && '\\' == timestamp.at(timestamp_idx)) {
                     ++timestamp_idx;
                     continue;
                 }
+                if (timestamp.substr(timestamp_idx).starts_with(cJsonEscapedBackslash)) {
+                    timestamp_idx += cJsonEscapedBackslash.size();
+                    continue;
+                }
                 return ErrorCode{ErrorCodeEnum::IncompatibleTimestampPattern};
```
LinZhihao-723: How about rewrite in this way to make it cleaner to understand:
```c++
                auto const expected_backslash{is_json_literal ? cJsonEscapedBackslash : "\\"};
                if (false == timestamp.substr(timestamp_idx).starts_with(expected_backslash)) {
                    return ErrorCode{ErrorCodeEnum::IncompatibleTimestampPattern};
                }
                timestamp_idx += expected_backslash.size();
                break;
```
`start_with` should handle the case where the string view is shorter than `expected_backslash`.

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
```diff
@@ -218,11 +219,12 @@ class TimestampPattern {
 [[nodiscard]] auto parse_timestamp(
         std::string_view timestamp,
         TimestampPattern const& pattern,
-        std::string& generated_pattern
+        std::string& generated_pattern,
+        bool is_json_literal
```
LinZhihao-723: We should swap them since `generated_pattern` is an output param

components/core/src/clp_s/timestamp_parser/test/test_TimestampParser.cpp
```diff
@@ -139,6 +142,47 @@ auto generate_padded_number_subset(size_t num_digits) -> std::vector<std::string
...
+                R"(\b)",
+                R"(\f)",
+                R"(\n)",
+                R"(\r)",
+                R"(\t)",
+                R"(\u)"
+        };
+        for (auto const& illegal_timestamp_pattern_template : illegal_timestamp_pattern_templates) {
+            auto const result{TimestampPattern::create(illegal_timestamp_pattern_template)};
+            REQUIRE(result.has_error());
```
LinZhihao-723, line 176: We should assert the returned error code if we plan to distinguish invalid/unsupported chars from a general error.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -289,6 +291,24 @@ find_first_matching_prefix(std::string_view str, std::span<std::string_view cons
         std::string& buffer
 ) -> ystdlib::error_handling::Result<void>;
 
+/**
+ * @param c
+ * @return Whether `c` indicates an unsupported JSON escape sequence when preceded by a backslash.
```
```suggestion
 * @return Whether `c` indicates an unsupported JSON escape sequence when treated as an escaped
 * character.
```
Trying to make it more precise as a backslash could be technically escaped, lol

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -289,6 +291,24 @@ find_first_matching_prefix(std::string_view str, std::span<std::string_view cons
...
+ */
+[[nodiscard]] auto is_unsupported_json_escape_sequence(char c) -> bool;
+
+/**
+ * @param c
+ * @return Whether `c` is an unsupported character.
+ */
+[[nodiscard]] auto is_unsupported_character(char c) -> bool;
```
LinZhihao-723, line 303: Nit: Shall we be explicit to jsut call it "control characters"?

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
LinZhihao-723, line 201: Shall we document our limitations on those unsupported chars in this docstring?

LinZhihao-723 (changes requested):
Overall lgtm.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
gibber9809: The thing is that it is control characters + `"`, so I wanted a more generic name.

LinZhihao-723: Sure sure, make sense

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
```diff
@@ -197,9 +199,20 @@ class TimestampPattern {
...
  *
+ * As well as the following escape sequences:
+ *
+ * - \\ Literal backslash.
+ *
+ * Any escape sequences not in these lists is illegal. As well, all of the following characters are
+ * illegal:
+ *
+ * - Control characters (characters in the range `\x00` to `\x1F`).
+ * - Double quotes `"`, excepting quotes that surround a timestamp pattern like `"<pattern>"`.
```
```suggestion
 * Any escape sequence not listed above is invalid. In addition, patterns must not contain:
 *
 * - ASCII control characters (U+0000 through U+001F).
 * - The double-quote character `"` (unless the whole pattern is supplied as a JSON string literal,
 *   e.g. `"<pattern>"`).
```

gibber9809:
> 
> ```
> feat(clp-s::timestamp_parser): Invalidate unsupported characters and escape sequences in timestamp patterns.
> ```

How about:
```
feat(clp-s::timestamp_parser): Make control characters and lone double quotes illegal in timestamp patterns.
```
Trying to be a bit more precise since technically the unsupported escape sequences were already illegal, we're just being a bit more defensive about them.

LinZhihao-723 (approved):

PR #1722, opened 2025-12-03
feat(clp-package): Add support for uploading query result files to S3.

hoophalab:
# Description

Currently, the query workers can write query results to a local filesystem by setting  `"write_to_file": true` in  the API Server, which supports higher max_num_results than mongo db.

However, in a multi node setup without seaweed fs, the API server cannot read query result files on the local filesystem of the query worker.

This PR adds support for uploading those query result files to S3 in the query worker.

There will be an upcoming PR today which reads results from S3 in the API server.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

1. Change `stream_output` storage to `s3`
```
stream_output:
  storage:
    type: "s3"
    staging_directory: "var/data/staged-streams"  # Or a path of your choosing
    s3_config:
      region_code: "us-east-1"
      bucket: "test-bucket"
      key_prefix: "streams/"
      aws_authentication:
        type: "credentials"
        credentials:
          access_key_id: "..."
          secret_access_key: "..."
```
2. Send a request to the API Server with `"write_to_file": true`
```
curl -X POST http://localhost:3001/query \
 -H "Content-Type: application/json" \
 -d '{
   "query_string": "*log*",
   "dataset": "default",
   "ignore_case": false,
   "max_num_results": 100,
   "write_to_file": true
 }'
```

3. The query result files are under `s3://test-bucket/streams/{search_job_id}`
(tested on localstack)

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **New Features**
  * Query-generated files are automatically uploaded to S3 when S3 storage and file output are enabled.

* **Bug Fixes / Reliability**
  * Uploads now include start/finish logging.
  * Upload failures are caught, recorded (error path set), cause the task to be marked as failed, and local source files are removed after upload.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1728, opened 2025-12-04
feat(api-server): Add support for steaming query result files from S3.

hoophalab:
# Description

This is a successor to #1722, which adds support for uploading query result files to S3.
In this PR, we add support for streaming query results in api server from s3

# Checklist

  breaking change.

# Validation performed

1. Change `stream_output` storage to `s3`
```
stream_output:
  storage:
    type: "s3"
    staging_directory: "var/data/staged-streams"  # Or a path of your choosing
    s3_config:
      region_code: "us-east-1"
      bucket: "test-bucket"
      key_prefix: "streams/"
      aws_authentication:
        type: "credentials"
        credentials:
          access_key_id: "..."
          secret_access_key: "..."
```
2. Send a request to the API Server with `"write_to_file": true`
```
curl -X POST http://localhost:3001/query \
 -d '{
   "query_string": "*log*",
   "dataset": "default",
   "ignore_case": false,
   "max_num_results": 100,
   "write_to_file": true
 }'
```
3. curl http://localhost:3001/query_results/1 fetches the results correctly.
(tested on localstack)

## Summary by CodeRabbit

* **New Features**
  * Stream search query results directly from or to Amazon S3.

* **Chores**
  * Integrated AWS S3 SDK and added AWS-related configuration support (including S3 settings and credentials deserialization).
  * Made AWS service endpoints optional and improved client configuration flexibility.
  * Added AWS-aware error reporting to surface S3-related errors.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -46,15 +46,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
-name = "android_system_properties"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
-dependencies = [
- "libc",
-]
-
-[[package]]
 name = "anstream"
 version = "0.6.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -86,22 +77,22 @@ dependencies = [
 
 [[package]]
 name = "anstyle-query"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9e231f6134f61b71076a3eab506c379d4f36122f2af15a9ff04415ea4c3339e2"
+checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
 name = "anstyle-wincon"
-version = "3.0.10"
+version = "3.0.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e0633414522a32ffaac8ac6cc8f748e090c5717661fddeea04219e2344f5f2a"
+checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -129,7 +120,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sqlx",
- "thiserror 2.0.17",
+ "thiserror",
  "tokio",
  "tokio-stream",
  "tower-http",
@@ -196,9 +187,9 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "aws-config"
-version = "1.8.10"
+version = "1.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1856b1b48b65f71a4dd940b1c0931f9a7b646d4a924b9828ffefc1454714668a"
+checksum = "a0149602eeaf915158e14029ba0c78dedb8c08d554b024d54c8f239aab46511d"
 dependencies = [
  "aws-credential-types",
  "aws-runtime",
@@ -215,7 +206,7 @@ dependencies = [
  "bytes",
  "fastrand",
  "hex",
- "http 1.3.1",
+ "http 1.4.0",
  "ring",
  "time",
  "tokio",
@@ -238,9 +229,9 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.14.1"
+version = "1.15.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "879b6c89592deb404ba4dc0ae6b58ffd1795c78991cbb5b8bc441c48a070440d"
+checksum = "6b5ce75405893cd713f9ab8e297d8e438f624dde7d706108285f7e17a25a180f"
 dependencies = [
  "aws-lc-sys",
  "zeroize",
@@ -248,11 +239,10 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-sys"
-version = "0.32.3"
+version = "0.34.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "107a4e9d9cab9963e04e84bb8dee0e25f2a987f9a8bad5ed054abd439caa8f8c"
+checksum = "179c3777a8b5e70e90ea426114ffc565b2c1a9f82f6c4a0c5a34aa6ef5e781b6"
 dependencies = [
- "bindgen",
  "cc",
  "cmake",
  "dunce",
@@ -308,7 +298,7 @@ dependencies = [
  "hex",
  "hmac",
  "http 0.2.12",
- "http 1.3.1",
+ "http 1.4.0",
  "http-body 0.4.6",
  "lru",
  "percent-encoding",
@@ -320,9 +310,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-sqs"
-version = "1.89.0"
+version = "1.90.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d20475f442a6b4ebf4dc3c8383fcb6c98346fbdac815b643790e1b765054dcfb"
+checksum = "201073d85c1852c22672565b9ddd8286ec4768ad680a261337e395b4d4699d44"
 dependencies = [
  "aws-credential-types",
  "aws-runtime",
@@ -342,9 +332,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-sso"
-version = "1.89.0"
+version = "1.90.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a9c1b1af02288f729e95b72bd17988c009aa72e26dcb59b3200f86d7aea726c9"
+checksum = "4f18e53542c522459e757f81e274783a78f8c81acdfc8d1522ee8a18b5fb1c66"
 dependencies = [
  "aws-credential-types",
  "aws-runtime",
@@ -364,9 +354,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-ssooidc"
-version = "1.91.0"
+version = "1.92.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4e8122301558dc7c6c68e878af918880b82ff41897a60c8c4e18e4dc4d93e9f1"
+checksum = "532f4d866012ffa724a4385c82e8dd0e59f0ca0e600f3f22d4c03b6824b34e4a"
 dependencies = [
  "aws-credential-types",
  "aws-runtime",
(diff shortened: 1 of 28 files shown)


components/job-orchestration/job_orchestration/executor/query/fs_search_task.py
```diff
@@ -246,4 +251,19 @@ def search(
...
+
+        logger.info(f"Uploading query results {dest_path} to S3...")
+        try:
+            s3_put(s3_config, src_file, dest_path)
+            logger.info(f"Finished uploading query results {dest_path} to S3.")
+        except Exception as err:
+            logger.error(f"Failed to upload query results {dest_path}: {err}")
+            task_results.status = QueryTaskStatus.FAILED
+            task_results.error_log_path = str(os.getenv("CLP_WORKER_LOG_PATH"))
+
```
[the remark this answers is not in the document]
hoophalab: will be done in clp#1722

components/api-server/src/client.rs
```diff
@@ -277,6 +296,89 @@ impl Client {
...
+                        .bucket(s3_config.bucket.as_str())
+                        .key(key)
+                        .send()
+                        .await?;
+                    let bytes = obj.body.collect().await?.into_bytes();
+                    let mut deserializer = rmp_serde::Deserializer::from_read_ref(bytes.as_ref());
+                    while let Ok(event) = Deserialize::deserialize(&mut deserializer) {
+                        let event: (i64, String, String, String, i64) = event;
+                        yield Ok(event.1);
+                    }
```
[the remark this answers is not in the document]
hoophalab: @LinZhihao-723 coderabbit is valid, but `rmp_serde::Deserializer` requires a sync `Read` while `obj.body` is a bytestream and can only provide async read. Any idea on how to fix this? Or we just collect the file to memory.

components/api-server/src/client.rs
```diff
@@ -186,11 +191,25 @@ impl Client {
...
+                }
+                StreamOutputStorage::S3 {
+                    staging_directory: _,
+                    s3_config: _,
+                } => {
+                    return Ok(SearchResultStream::S3 {
+                        inner: self.fetch_results_from_s3(search_job_id).await,
+                    });
+                }
+            }
```
hoophalab: I guess we shouldn't return in both branches, any idea on how to make this part reads better?

components/clp-rust-utils/src/s3/client.rs
```diff
@@ -17,7 +17,7 @@ use secrecy::{ExposeSecret, SecretString};
 /// A newly created S3 client.
 #[must_use]
 pub async fn create_new_client(
-    endpoint: &str,
+    optional_endpoint: Option<&str>,
```
hoophalab: I changed the argument to optional so that s3 client can automatically resolve endpoint for aws (we don't need to provide a `https://{}.amazonaws.com`). Any issue with this change?

components/api-server/src/client.rs
LinZhihao-723: How about:

```suggestion
            return Ok(match &self.config.stream_output.storage {
                StreamOutputStorage::Fs { .. } => SearchResultStream::File {
                    inner: self.fetch_results_from_file(search_job_id),
                },
                StreamOutputStorage::S3 { .. } => SearchResultStream::S3 {
                    inner: self.fetch_results_from_s3(search_job_id).await,
                },
            });
```

components/api-server/src/client.rs
```diff
@@ -277,6 +296,88 @@ impl Client {
...
+    ///
+    /// Panics if the stream output storage is not S3.
+    async fn fetch_results_from_s3(
+        &self,
+        search_job_id: u64,
+    ) -> impl Stream<Item = Result<String, ClientError>> + use<> {
+        tracing::info!("Streaming results from S3");
+        let StreamOutputStorage::S3 {
+            staging_directory: _,
+            s3_config,
```
LinZhihao-723: How about:

```suggestion
            s3_config,
            ..
```

components/clp-rust-utils/src/s3/client.rs
LinZhihao-723: Sure. Probably need to merge log-ingestor changes and drop the AWS endpoint generation: 
- https://github.com/y-scope/clp/blob/bfc474f4938c533ea9447ac2e3ffc120459c5662/components/log-ingestor/src/aws_client_manager.rs#L52
- https://github.com/y-scope/clp/blob/bfc474f4938c533ea9447ac2e3ffc120459c5662/components/log-ingestor/src/aws_client_manager.rs#L83

components/api-server/src/client.rs
```diff
@@ -277,6 +296,88 @@ impl Client {
...
+        let key_prefix = format!("{}{}/", s3_config.key_prefix, search_job_id);
+        tracing::info!("Streaming results from S3 prefix: {}", key_prefix);
+        let mut object_pages = s3_client
+            .list_objects_v2()
+            .bucket(s3_config.bucket.as_str())
+            .prefix(key_prefix)
+            .into_paginator()
+            .send();
+
+        let stream = stream! {
```
LinZhihao-723: Can we return the `stream!` directly?

components/api-server/src/client.rs
```diff
@@ -277,6 +296,88 @@ impl Client {
...
+            while let Some(object_page) = object_pages.next().await {
+                tracing::info!("Received S3 object page: {:?}", object_page);
```
LinZhihao-723: How about using `debug` here? We first print the page and then print each key in the inner loop, which seems a bit inefficient.

components/api-server/src/client.rs
```diff
@@ -277,6 +296,88 @@ impl Client {
...
+                for object in object_page?.contents() {
+                    let Some(key) = object.key() else {
+                        tracing::info!("S3 object {:?} doesn't have a key", object);
+                        continue;
+                    };
```
LinZhihao-723, line 355:
```suggestion
                    };
                    if key.ends_with('/') {
                        tracing::info!("Skipping S3 object with key {} as it is a directory", key);
                        continue;
                    }

```
We should probably skip `/` (in this case it probably never exist).

components/api-server/src/client.rs
```diff
@@ -277,6 +296,88 @@ impl Client {
...
+                    tracing::info!("Streaming results from S3 object with key: {}", key);
+                    let obj = s3_client
+                        .get_object()
+                        .bucket(s3_config.bucket.as_str())
+                        .key(key)
+                        .send()
+                        .await?;
+                    let bytes = obj.body.collect().await?.into_bytes();
```
LinZhihao-723, line 367: To reply [this](https://github.com/y-scope/clp/pull/1728/files#r2587282252)
I also realized this is an issue... iiuc, it seems like it needs to download the entire file to memory. Then in this way, technically it's not "streaming" results from S3 any more...
I think for now, let's document this limitation explicitly and leave it as it is.
Long term speaking, we could:
* Implement our own msgpack deserializer, or
* Switch to a more async-friendly format for serializing search results.

If this sounds good to u, let's ask rabbit to create an issue.

components/api-server/src/error.rs
```diff
@@ -36,6 +44,30 @@ impl<T: IsMalformedData> From<T> for ClientError {
...
+    }
+}
+
+impl From<SdkError<ListObjectsV2Error>> for ClientError {
+    fn from(value: SdkError<ListObjectsV2Error>) -> Self {
+        Self::Aws {
+            description: value.to_string(),
+        }
+    }
+}
```
LinZhihao-723: How about:

```suggestion
impl<AwsSdkErrorType> From<SdkError<AwsSdkErrorType>> for ClientError {
    fn from(value: SdkError<AwsSdkErrorType>) -> Self {
        Self::Aws {
            description: value.to_string(),
        }
    }
}
```

components/clp-rust-utils/src/s3/client.rs
LinZhihao-723: I feel like in this case, we should probably:
- Name it `endpoint`; in Rust optional is already indicated by its type so it's probably not necessary to have this prefix.
- Move it to the last parameter.

components/api-server/src/client.rs
```diff
@@ -277,6 +296,88 @@ impl Client {
...
+    /// ## Returns
+    ///
+    /// The log message in string representation on success.
+    ///
+    /// ## Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`aws_sdk_s3::error::SdkError`]'s return values on failure.
+    /// * Forwards [`aws_sdk_s3::error::primitives::ByteStreamError`]'s return values on failure.
```
LinZhihao-723: Hmmm these are wrong. We should document which method's error it forwards, not the error type. How about:
```suggestion
    /// * Forwards the pagination stream's error returned from S3 list-object-v2 operation's
    ///   paginator (i.e., from
    ///   [`aws_smithy_async::future::pagination_stream::PaginationStream::next`]).
    /// * Forwards S3 get-object operation's return values on failure (i.e., from
    ///   [`aws_sdk_s3::operation::get_object::builders::GetObjectFluentBuilder::send`]).
    /// * Forwards [`aws_smithy_types::byte_stream::ByteStream::collect`]'s return values on
    ///   failure.
```
I kind of realized documenting the source function of the error makes little sense here, so I annotated which operation it is related to. Do you think it makes sense to you? We can update our guideline to follow this if you think it makes sense.

components/api-server/src/client.rs
```diff
@@ -334,20 +435,24 @@ pin_project! {
     /// * `FileStream`: Streaming from file system storage.
     /// * `MongoStream`: Streaming from MongoDB storage.
```
LinZhihao-723, line 433: We need to update the docstring here:

```suggestion
    /// * `MongoStream`: Streaming from MongoDB storage.
    /// * `S3Stream`: Streaming from S3 storage.
```

components/api-server/src/client.rs
```diff
@@ -277,6 +296,88 @@ impl Client {
...
+
+        let s3_client = clp_rust_utils::s3::create_new_client(
+            None,
+            s3_config.region_code.as_str(),
+            credentials.access_key_id.as_str(),
+            &SecretString::from(credentials.secret_access_key.as_str()),
+        )
+        .await;
+
+        let key_prefix = format!("{}{}/", s3_config.key_prefix, search_job_id);
```
LinZhihao-723, line 339: Just to confirm: The key_prefix here contains an ending `/`, right?

LinZhihao-723 (changes requested):

components/api-server/src/client.rs
hoophalab: There could be multiple files. We don't have to download all files to start replying. So it's still a "stream".
However, we would always have to download the whole file either to memory or to disk from S3.

components/api-server/src/client.rs
hoophalab: yes, it should contain `/`

components/api-server/src/client.rs
LinZhihao-723: > However, we would always have to download the whole file either to memory or to disk from S3.

This might be a limitation that can potentially cause issues. We should document this + potential solutions.

components/api-server/src/client.rs
```diff
@@ -186,11 +190,22 @@ impl Client {
         }
 
         let job_config = self.get_job_config(search_job_id).await?;
+
         if job_config.write_to_file {
-            return Ok(SearchResultStream::File {
-                inner: self.fetch_results_from_file(search_job_id),
-            });
+            let stream = match &self.config.stream_output.storage {
+                StreamOutputStorage::Fs { directory: _ } => SearchResultStream::File {
```
```suggestion
                StreamOutputStorage::Fs { .. } => SearchResultStream::File {
```

components/api-server/src/client.rs
```diff
@@ -186,11 +190,22 @@ impl Client {
...
+                    inner: self.fetch_results_from_file(search_job_id),
+                },
+                StreamOutputStorage::S3 {
+                    staging_directory: _,
+                    s3_config: _,
+                } => SearchResultStream::S3 {
```
```suggestion
                StreamOutputStorage::S3 { .. } => SearchResultStream::S3 {
```

components/api-server/src/client.rs
hoophalab: > To reply [this](https://github.com/y-scope/clp/pull/1728/files#r2587282252) I also realized this is an issue... iiuc, it seems like it needs to download the entire file to memory. Then in this way, technically it's not "streaming" results from S3 any more... I think for now, let's document this limitation explicitly and leave it as it is. Long term speaking, we could:
> 
> * Implement our own msgpack deserializer, or
> * Switch to a more async-friendly format for serializing search results.
> 
> If this sounds good to u, let's ask rabbit to create an issue.

@coderabbit can you create an issue on this? Link this code

PR #1732, opened 2025-12-05
feat(yscope-clp-core): Add `yscope-clp-core` to package CLP core as a Python distribution:

LinZhihao-723:
* Package the `clp-s` binary as part of the wheel.
* Add a task to build the wheel.
* Update the workflow to build the wheel targeting `manylinux_2_28_x86_64`.

# Description

This PR adds support for building CLP core (clp-s only for now) wheels. The structure of this wheel project is learned from https://github.com/ssciwr/clang-tidy-wheel.

The wheel should be built on `manylinux_2_28_x86_64` for maximum Linux compatibility.

To build the wheel:

* [Optional] Build and start CLP's `manylinux_2_28_x86_64` container: https://github.com/y-scope/clp/tree/main/components/core/tools/docker-images/clp-env-base-manylinux_2_28-x86_64
* Inside the container:
  * Build and clone the CLP repo from [PR source](https://github.com/LinZhihao-723/clp/tree/python-wheels). 
  * Run the following commands:
```
git checkout python-wheels
./tools/scripts/deps-download/init.sh
task python-wheel-clp-core
```
  * The wheel file can be found inside the directory: `build/python-wheels/clp-core`

Q&A:

* Why not use `cibuildwheel`?
  * Not all dependencies are installable via `task`. The build still needs to run inside a CLP environment.
* Why not let CMake (scikit-build) initiate wheel building instead of `task`?
  * Invoking `task` from the wheel project’s CMake currently fails due to an unresolved `CMAKE_ROOT` missing error.

# Checklist

  breaking change.

# Validation performed

* [X] Wheel building succeeds and virtual env installed with the wheel can use `clp-s` properly.

## Summary by CodeRabbit

* **New Features**
  * Added yscope-clp-core Python package exposing a clp-s console command and programmatic entry; automated wheel build, repair and CI packaging.

* **Documentation**
  * Added project README and Apache 2.0 license with installation and compatibility guidance.

* **Chores**
  * Packaging/build tooling and Python version pinning; lint/project path updates; ignore rule for bundled binaries; CI workflows extended to build and upload wheel artifacts.

### DIFF
--- a/python-wheels/yscope-clp-core/README.md
+++ b/python-wheels/yscope-clp-core/README.md
@@ -27,5 +27,4 @@ For detailed usage of CLP core features and design documentation, see the offici
 
 [clp-core-docs]: https://docs.yscope.com/clp/main/user-docs/index.html#core
 [clp-s-docs]: https://docs.yscope.com/clp/main/user-docs/core-clp-s
-[manylinux]: https://github.com/pypa/manylinux
 [pypi-release]: https://pypi.org/project/yscope-clp-core


.github/workflows/clp-artifact-build.yaml
```diff
@@ -40,6 +41,7 @@ jobs:
       musllinux_1_2_x86_64_image_changed: "${{steps.filter.outputs.musllinux_1_2_x86_64_image}}"
       ubuntu_jammy_image_changed: "${{steps.filter.outputs.ubuntu_jammy_image}}"
       clp_changed: "${{steps.filter.outputs.clp}}"
+      yscope_clp_core_changed: "${{steps.filter.outputs.yscope-clp-core}}"
```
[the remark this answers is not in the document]
LinZhihao-723: Prefer to do it in a separate PR. This is to make it consistent with the existing code.

.github/workflows/clp-artifact-build.yaml
```diff
@@ -283,6 +292,60 @@ jobs:
...
+          echo "output_dir=${output_dir}" >> "$GITHUB_OUTPUT"
+          mkdir -p "${output_dir}"
+
+          find "$GITHUB_WORKSPACE/build/python-wheels" \
+            -type f \
+            -name '*.whl' \
+            -exec cp {} "${output_dir}/" \;
+
+          tar cfvv "${output_dir}/python-wheels.tar" -C "${output_dir}" .
+        shell: "bash"
```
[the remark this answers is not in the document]
LinZhihao-723: Fixed.

.github/workflows/clp-artifact-build.yaml
LinZhihao-723, line 1: @junhaoliao Can you help me review the changes in this file?
Technically we should reuse artifacts from the binary build job. However, I prefer to leave it to a future PR as I'd like to keep the scope of this PR self-contained (don't touch other tasks/jobs).

python-wheels/yscope-clp-core/yscope_clp_core/_bin.py
```diff
@@ -0,0 +1,50 @@
...
+    Locates a bundled CLP core executable installed with the Python distribution.
+
+    Executables are installed into the package at build time by CMake under `yscope_clp_core/bin`.
+    This helper resolves the on disk path at runtime and caches the result to avoid repeated
+    filesystem lookups.
+
+    :param name: Name of the executable to locate.
+    :return: Path to the executable.
+    """
+    exe = Path(str(files("yscope_clp_core") / f"bin/{name}"))
```
junhaoliao, line 24: (haven't confirmed if it would work, but it seems like files() returns [a `Traversable` object which implements certain `pathlib.Path` methods?](https://docs.python.org/3/library/importlib.resources.abc.html#importlib.resources.abc.Traversable)
```suggestion
    exe = files("yscope_clp_core") / "bin" / name)
```

python-wheels/yscope-clp-core/yscope_clp_core/_bin.py
```diff
@@ -0,0 +1,50 @@
...
+    if exe.exists():
+        return exe
+    err_msg = f"No executable found for {name} at {exe}."
+    raise FileNotFoundError(err_msg)
+
+
+def _run_clp_exe(name: str, *args: Any) -> int:
```
junhaoliao: > *args: Any

not sure if this is type safe. shall we change it to `*args: str` to avoid us passing in some argument that might be type converted in unexpected ways?

junhaoliao: or `*args: str | os.PathLike[str]` since the binaries do work a lot with files

python-wheels/yscope-clp-core/yscope_clp_core/_bin.py
```diff
@@ -0,0 +1,50 @@
...
+    if args:
+        command += list(args)
+    else:
+        command += sys.argv[1:]
+    return subprocess.call(command)
+
+
+def clp_s() -> None:
+    """Entry point that dispatches to the clp_s executable."""
+    raise SystemExit(_run_clp_exe("clp-s"))
```
junhaoliao: i understand it's perfectly conventionally to raise an SystemExit exception here; however:
1. technically we are still throwing an exception so this is worth documenting in the docstring?
2. should the return type of `clp_s()` be `NoReturn`?

python-wheels/yscope-clp-core/pyproject.toml
```diff
@@ -0,0 +1,55 @@
+[project]
+name = "yscope-clp-core"
+version = "0.7.1-beta.1"
+description = "YScope CLP core Python Distribution"
```
junhaoliao: would this be better for SEO? (also the "C" is capitalized
```suggestion
description = "YScope CLP (Compressed Log Processor) Core Python Distribution"
```

python-wheels/yscope-clp-core/pyproject.toml
```diff
@@ -0,0 +1,55 @@
...
+readme = "README.md"
+authors = [
+    { name = "YScope Inc.", email = "dev@yscope.com" }
+]
+license = { text = "Apache License 2.0" }
```
junhaoliao: can we simply write
```suggestion
license = "Apache-2.0"
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -281,6 +290,56 @@ jobs:
...
+      - id: "copy_wheels"
+        run: |-
+          output_dir="/tmp/${{env.PYTHON_WHEELS_ARTIFACT_NAME_PREFIX}}${{env.OS_NAME}}"
+          echo "output_dir=${output_dir}" >> "$GITHUB_OUTPUT"
+          mkdir -p "${output_dir}"
+          tar cfvv "${output_dir}/python-wheels.tar" \
+            --directory "$GITHUB_WORKSPACE/build/python-wheels" .
+        shell: "bash"
+
+      - uses: "actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02"
```
junhaoliao: 1. use newer version?
2. add version docstring
```suggestion
      - uses: "actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f"  # v6.0.0
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -281,6 +290,56 @@ jobs:
...
+      - "manylinux_2_28-x86_64-deps-image"
+      - "filter-relevant-changes"
+    env:
+      OS_NAME: "manylinux_2_28"
+    name: >-
+      manylinux_2_28-x86_64-python-wheels
+    continue-on-error: true
+    runs-on: *runner
+    steps:
+      - uses: "actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683"
```
junhaoliao:
```suggestion
      - uses: "actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8"  # v6.0.1
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -281,6 +290,56 @@ jobs:
...
+  manylinux_2_28-x86_64-python-wheels:
+    # Run if the ancestor jobs succeeded OR they were skipped and clp was changed.
+    if: >-
+      success() || (!cancelled() && !failure() && (
+        needs.filter-relevant-changes.outputs.clp_changed == 'true'
+        || needs.filter-relevant-changes.outputs.yscope_clp_core_changed == 'true'
+      ))
+    needs:
+      - "manylinux_2_28-x86_64-deps-image"
+      - "filter-relevant-changes"
```
junhaoliao: if the order doesn't matter
```suggestion
      - "filter-relevant-changes"
      - "manylinux_2_28-x86_64-deps-image"
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -281,6 +290,56 @@ jobs:
...
+    env:
+      OS_NAME: "manylinux_2_28"
+    name: >-
+      manylinux_2_28-x86_64-python-wheels
```
junhaoliao: since it can fit on one line
```suggestion
    name: "manylinux_2_28-x86_64-python-wheels"
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -281,6 +290,56 @@ jobs:
...
+            CLP_CPP_MAX_PARALLELISM_PER_BUILD_TASK=$(getconf _NPROCESSORS_ONLN)
+            task python-wheels-yscope-clp-core
+
+      - id: "copy_wheels"
+        run: |-
+          output_dir="/tmp/${{env.PYTHON_WHEELS_ARTIFACT_NAME_PREFIX}}${{env.OS_NAME}}"
+          echo "output_dir=${output_dir}" >> "$GITHUB_OUTPUT"
+          mkdir -p "${output_dir}"
+          tar cfvv "${output_dir}/python-wheels.tar" \
+            --directory "$GITHUB_WORKSPACE/build/python-wheels" .
```
junhaoliao:
```suggestion
          tar \
            --create \
            --verbose \
            --verbose \
            --file="${output_dir}/python-wheels.tar" \
            --directory="$GITHUB_WORKSPACE/build/python-wheels" \
            .
```

python-wheels/yscope-clp-core/README.md
```diff
@@ -0,0 +1,32 @@
+# CLP core Python Distribution
+
+This project packages the CLP core as a Python distribution, providing the CLP core command line
```
junhaoliao: ... YScope CLP (Compressed Log Processor) core ...

python-wheels/yscope-clp-core/README.md
```diff
@@ -0,0 +1,32 @@
...
+```shell
+pip install yscope-clp-core
+```
+
+See the PyPI project page for [release][pypi-release] details.
+
+## Scope and Compatibility
+
+> [!IMPORTANT]
+> The current package only includes the `clp-s` binary. Support for additional CLP core tools will
```
junhaoliao, line 19: shall we mention x86_64 only? i believe we want to add arm binaries in the future

junhaoliao (changes requested):

python-wheels/yscope-clp-core/README.md
LinZhihao-723: I kind of missed this. We should probably not mention any platform specific info/notes. These are already indicated in the wheels we provide.

python-wheels/yscope-clp-core/yscope_clp_core/_bin.py
```diff
@@ -0,0 +1,50 @@
+"""Wrapper for locating and invoking the bundled yscope_clp_core command line binaries."""
+
+import functools
+import subprocess
+import sys
+from importlib.resources import files
+from pathlib import Path
+from typing import Any
```
Bill-hbrhbr:
```suggestion
from typing import Any, NoReturn
```

python-wheels/yscope-clp-core/yscope_clp_core/_bin.py
Bill-hbrhbr:
```suggestion
def clp_s() -> NoReturn:
    """
    Entry point that dispatches to the clp_s executable.
    
    :raise SystemExit: Terminates the process with the executable's exit code.
    """
    raise SystemExit(_run_clp_exe("clp-s"))
```

.github/workflows/clp-artifact-build.yaml
Bill-hbrhbr: double verbose?

python-wheels/yscope-clp-core/yscope_clp_core/_bin.py
LinZhihao-723: Sorry again for missing this change (cuz it's indeed different from my initial version).
I'm going to make this function takes a list of strings to keep its behavior simple. The current behavior over complicates what it needs to. You could leave it to the caller to decide what to pass in as the arg, instead of do this magic check. Gonna push it soon unless @Bill-hbrhbr strongly disagrees.

python-wheels/yscope-clp-core/yscope_clp_core/_bin.py
LinZhihao-723: As discussed offline, let's use `sys.exit` for simplicit because:
* Using `SystemExit` to terminate a process is too Pythonic and not a traditional programming practice.
* Adding a raise section in the docstring is even more confusing where this is not to throw an error. You can't even catch `SystemExit` with `except Exception`, lol.

I think for later Python code, we should also try to follow the common standard as the Python world is just too crazy to be a part of.

.github/workflows/clp-artifact-build.yaml
LinZhihao-723: Even though double `verbose` here mean verbose level 2, I'd prefer to leave it with only one to avoid confusing. We probably don't need extra verbosed metadata anyway since this tar is simple.

python-wheels/yscope-clp-core/yscope_clp_core/_bin.py
LinZhihao-723: * Tried out that it indeed implements `is_file` method.
* However, conventionally, returning a `Path` object from this function might be more obvious, so I would just keep it simple by converting things into a `Path` object and then do the check.

fyi this is actually copied from clang-tidy's setup: https://github.com/ssciwr/clang-tidy-wheel/blob/edaff845de8d03de419741b32d3c88499bec4c5c/clang_tidy/__init__.py#L19

python-wheels/yscope-clp-core/pyproject.toml
```diff
@@ -0,0 +1,55 @@
+[project]
+name = "yscope-clp-core"
+version = "0.7.1-beta.1"
```
junhaoliao, line 3:
```suggestion
version = "0.8.1-dev"
```

This PR was merged after #1876 but the version string wasn't updated properly. this is fine because the `yscope-clp-core` project currently on the main branch is only some boilerplate. though, we should remember to bump this version string when we bump version strings for the next development cycle


(in fact, for the next development version, it should be)
```suggestion
version = "0.9.1-dev"
```

PR #1734, opened 2025-12-05
feat(webui): Move timestamp presets into AntD date picker. 

davemarco:
# Description

Moves presets into the Ant Design time selector.

AntD’s built-in presets feature couldn’t be used properly because it automatically triggers the onChange callback and isn’t asynchronous. Triggering onChange is problematic here — we use that callback to update the picker value when switching to a custom timestamp, but we don’t want it to fire when selecting a preset, since in that case we just want to display the preset label instead of the raw date/time values.

Implementation Details

To fix this, I rendered a fully custom preset component next to the AntD date picker. This prevents unwanted onChange triggers and supports asynchronous behavior.

When the user opens the picker, it shows the current timestamp value. I tried keeping the preset text visible inside the picker, but it made editing timestamps awkward and complicated the code. If the user doesn’t make any changes and simply closes the picker, the preset text reappears.

Screenshot 2025-12-12 at 10 24 35 AM

# Checklist

  breaking change.

# Validation performed

Presets change

## Summary by CodeRabbit

* **New Features**
  * Integrated range date picker with a dedicated preset panel for faster selection.
  * New preset panel presents common ranges alongside a date picker for quick picks.

* **Improvements**
  * Input shows contextual text when picker is open and reflects current selection.
  * Editing a range auto-switches to a custom range; calendar selections normalized to UTC.
  * Improved focus/blur/change handling and refined panel styling for clearer presentation.


components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/TimeRangePanel/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+                            }}
+                            onClick={() => {
+                                handlePresetClick(option);
+                            }}
+                            onMouseEnter={(e) => {
+                                e.currentTarget.style.backgroundColor = token.controlItemBgHover;
+                            }}
+                            onMouseLeave={(e) => {
+                                e.currentTarget.style.backgroundColor = "transparent";
+                            }}
```
hoophalab: This part isn't reactive. I guess you've tried having a backgroundColor state and use that state in style={{...}}? There is a short delay?

components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/TimeDateInput.tsx
```diff
@@ -19,20 +20,47 @@ interface TimeDateInputProps {
...
+    };
+
     return (
         <input
-            {...props}
-            readOnly={true}
-            value={displayText}/>
+            {...restProps}
+            value={displayText}
+            onBlur={handleBlur}
```
hoophalab: Can we pass onblur and onfocus directly to input through restprops? This way we don't need the two handle.* delegates

components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/TimeRangePanel/index.tsx
davemarco: this is doing the hover effect. and u cant put on hover in style {}. I would put in css file, but then u cant use antD token

davemarco: i may not understand what u mean by your comment

hoophalab: Sorry, I didn't make myself clear. How about this:
```diff
@@ -9,6 +9,7 @@ import {
     TIME_RANGE_OPTION_NAMES,
 } from "../utils";
 import styles from "./index.module.css";
+import {Nullable} from "@webui/common/utility-types";
 
 
 interface TimeRangePanelProps {
@@ -42,6 +43,8 @@ const TimeRangePanel = ({panelNode, onClose}: TimeRangePanelProps) => {
             });
     };
 
+    const [activeOption, setActiveOption] = React.useState<Nullable<string>>(null);
+
     return (
         <div className={styles["panelContainer"]}>
             <ul
@@ -61,15 +64,13 @@ const TimeRangePanel = ({panelNode, onClose}: TimeRangePanelProps) => {
                                 borderRadius: token.borderRadiusSM,
                                 paddingInline: token.paddingXS,
                                 paddingBlock: token.paddingXXS,
+                                backgroundColor: option === activeOption ? token.controlItemBgHover : "transparent"
                             }}
                             onClick={() => {
                                 handlePresetClick(option);
                             }}
                             onMouseEnter={(e) => {
-                                e.currentTarget.style.backgroundColor = token.controlItemBgHover;
-                            }}
-                            onMouseLeave={(e) => {
-                                e.currentTarget.style.backgroundColor = "transparent";
+                                setActiveOption(option);
                             }}
                         >
                             {option}

```

hoophalab: might need to set option as null when mouse leaves

davemarco:
@hoophalab i realized i could use antd menu instead of custom list. From their docs looks like menu is for navigation, but turns out u can use anywhere. So it handles hover. I also changed it to keep highlight when selected preset, so it is a bit fancier

junhaoliao (approved):

PR #1740, opened 2025-12-07
feat(log-ingestor): Add ingestion job orchestration:

LinZhihao-723:
* Add APIs to start a new S3-scanning or SQS-listener ingestion job.
* Add an API to peacefully stop a running ingestion job.
* Refine ingestion job config schemas for submission.
* Improve internal logging to enhance end-to-end visibility and traceability during job execution.

# Description

**NOTE**: This PR depends on #1736.

This PR adds ingestion job orchestration service as described above. With this PR, users will be able to submit ingestion jobs to log-ingestor, which automatically runs to detect new files on S3 and create CLP compression jobs.

This PR is an internal breaking change becuase it contains changes to the ingestion job schema:

* For S3 scanner jobs: We make the scanning interval as an option exposed to users, which defaults to 30 sec. We also expose `start_from` as an option.
* For SQS listener jobs: We no longer make the number of messages per request and polling backoffs configurable to users. This is because these values may have upper and lower bounds that requires extra checks. On the other hand, making them configurable would not make a visible usability difference to users. We just use a hardcoded value to set both the waiting time and the number of messages per polling to the maximum.

This PR also improves how sleep logic is handled inside S3 scanner job: it now allows the job to be cancelled during the sleep so that job deletion API doesn't need to wait for the sleep duration to pass.

# Checklist

  breaking change.

# Validation performed

* [x] Ensure all workflow passed.
* [x] Service testing:
  * [x] Start 3 ingestion jobs running in-parallel (1 SQS listener job, 2 S3 scanner jobs). Upload logs to different prefix using [this](https://github.com/LinZhihao-723/LogIngestorServer/blob/main/scripts/generate_and_upload.py) script. Ensure the generated files are captured by log-ingestor and eventually submitted to a CLP compression job.
  * [x] Ensure running jobs can be successfully deleted.
  * [x] Ensure deleting a non-existing job will fail explicitly with an understandable error message.
  * [x] Ensure deleting a job with an invalid ID format will fail explicitly with an understandable error message.
  * [x] When creating a job that contains prefix conflict (`test-1765140288` vs. `test-`), ensure the job creation will fail.

## Summary by CodeRabbit

* **New Features**
  * HTTP API and CLI endpoints to create/stop ingestion jobs and check health; router integrated into server.

* **Improvements**
  * Ingestion configuration reorganized into composable S3 configs with explicit defaults and simpler fields.
  * Public API unified behind a manager state for job lifecycle operations.
  * Improved logging, error handling and cancellation-aware background tasks.
  * Default log‑ingestor port updated.

* **Chores**
  * Added serialization, SQL and formatting support; tests adapted to new config shapes.

### DIFF
--- a/components/log-ingestor/src/ingestion_job/sqs_listener.rs
+++ b/components/log-ingestor/src/ingestion_job/sqs_listener.rs
@@ -56,7 +56,7 @@ impl<SqsClientManager: AwsClientManagerType<Client>> Task<SqsClientManager> {
                     .queue_url(self.config.queue_url.as_str())
                     .max_number_of_messages(MAX_NUM_MESSAGES_TO_FETCH)
                     .wait_time_seconds(MAX_WAIT_TIME_SEC).send() => {
-                    let _ = self.process_sqs_response(result?).await?;
+                    self.process_sqs_response(result?).await?;
                 }
             }
         }


components/log-ingestor/src/compression/compression_job_submitter.rs
```diff
@@ -0,0 +1,187 @@
...
+                    }
+                    _ => {}
+                }
+            }
+            Err(e) => {
+                tracing::error!("Failed to fetch compression job status: {}", e);
+                return;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
LinZhihao-723: Out side of the scope of this PR.

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -28,16 +34,61 @@ pub enum Error {
...
+                     only supports S3 input."
+                ));
+            }
+        };
+        let mysql_pool = clp_rust_utils::database::mysql::create_mysql_pool(
+            &clp_config.database,
+            &clp_credentials.database,
+            10,
+        )
+        .await?;
```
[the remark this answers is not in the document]
LinZhihao-723: Will address this in a future PR.

components/log-ingestor/src/routes.rs
```diff
@@ -0,0 +1,138 @@
...
+        }
+        Err(err) => {
+            tracing::error!(err = ? err, "Failed to create SQS listener ingestion job.");
+            return Err(Error::IngestionJobManagerError(err));
+        }
+    };
+    Ok(Json(CreationResponse {
+        id: job_id.to_string(),
+    }))
+}
```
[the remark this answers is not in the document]
LinZhihao-723: No need for now.

components/log-ingestor/src/routes.rs
```diff
@@ -0,0 +1,139 @@
...
+    {
+        Ok(job_id) => {
+            tracing::info!(job_id = ? job_id, "Created S3 scanner ingestion job.");
+            job_id
+        }
+        Err(err) => {
+            tracing::error!(err = ? err, "Failed to create S3 scanner ingestion job.");
+            return Err(Error::IngestionJobManagerError(err));
+        }
+    };
```
hoophalab: Does using inspect,map_err + `?` look nicer than match? Same comment applies to other `match`es in this PR.
```suggestion
    let job_id = match ingestion_job_manager_state
        .create_s3_scanner_job(config)
        .await
        .map_err(|err| => {
            tracing::error!(err = ? err, "Failed to create S3 scanner ingestion job.");
            return Error::IngestionJobManagerError(err);
        })?;
    tracing::info!(job_id = ? job_id, "Created S3 scanner ingestion job.");
```

components/log-ingestor/src/routes.rs
```diff
@@ -0,0 +1,139 @@
...
+                    axum::http::StatusCode::INTERNAL_SERVER_ERROR
+                }
+                IngestionJobManagerError::JobNotFound(_) => axum::http::StatusCode::NOT_FOUND,
+                IngestionJobManagerError::PrefixConflict(_) => axum::http::StatusCode::CONFLICT,
+            },
+            Self::InvalidJobId(_) => axum::http::StatusCode::BAD_REQUEST,
+        };
+        let body = serde_json::json!({
+            "error": self.to_string()
+        });
```
hoophalab, line 57: Returning serialized error could accidentally expose critical internal information. Might be best to return "Internal Server Error" here.

components/log-ingestor/src/routes.rs
```diff
@@ -0,0 +1,139 @@
...
+/// # Returns
+///
+/// A newly created router instance configured without setting the state.
+pub fn create_router() -> Router<IngestionJobManagerState> {
+    Router::new()
+        .route("/", get(health))
+        .route("/health", get(health))
+        .route("/create_s3_scanner_job", post(create_s3_scanner_job))
+        .route("/create_sqs_listener_job", post(create_sqs_listener_job))
+        .route("/stop_and_delete_job/{job_id}", delete(stop_and_delete_job))
```
hoophalab: At your discretion: do you want to make the endpoint sound more restful/less rpc?

The following are copied from ai

Original Endpoint | HTTP Method | Recommended Endpoint | Rationale
-- | -- | -- | --
/health | GET | /health or /status | Standard practice for service checks.
/create_s3_scanner_job | POST | /s3 | Creates a new resource (s3-scanner) within the collection (jobs). The action is implied by POST.
/create_sqs_listener_job | POST | /sqs | Similar to the above, focusing on the resource being created.
/stop_and_delete_job/{job_id} | DELETE | /job/{job_id} | The action is implied by the DELETE method. The URL targets the specific resource (job_id) to be removed.

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -34,18 +34,13 @@ pub enum Error {
...
-    archive_output_config: ArchiveOutput,
-    mysql_pool: sqlx::MySqlPool,
+/// An async-safe state for creating and managing ingestion jobs.
+#[derive(Clone)]
+pub struct IngestionJobManagerState {
+    inner: Arc<IngestionJobManager>,
 }
 
-impl IngestionJobManager {
+impl IngestionJobManagerState {
```
hoophalab, line 43: By implementing functions on `IngestionJobManagerState`, the user cannot use `IngestionJobManager` directly. It might sounds more reasonable to call
`IngestionJobManagerState` as `IngestionJobManager` and call `IngestionJobManager` as `IngestionJobManagerInner`.

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -62,14 +54,9 @@ impl<SqsClientManager: AwsClientManagerType<Client>> Task<SqsClientManager> {
...
-                    .wait_time_seconds(polling_backoff_sec).send() => {
-                    polling_backoff_sec = if self.process_sqs_response(result?).await? {
-                        self.config.init_polling_backoff_sec
-                    } else { std::cmp::min(
-                        polling_backoff_sec.saturating_mul(2),
-                        self.config.max_polling_backoff_sec
-                    ) };
+                    .max_number_of_messages(MAX_NUM_MESSAGES_TO_FETCH)
+                    .wait_time_seconds(MAX_WAIT_TIME_SEC).send() => {
+                    let _ = self.process_sqs_response(result?).await?;
```
hoophalab: maybe log the error?

components/clp-rust-utils/src/clp_config/package/config.rs
```diff
@@ -197,6 +197,7 @@ impl Default for StreamOutputStorage {
 pub struct LogIngestor {
     pub host: String,
     pub port: u16,
+    #[serde(rename = "buffer_timeout")]
     pub buffer_timeout_sec: u64,
     pub buffer_size_threshold: u64,
```
hoophalab, line 202: Do you want to rename `buffer_size_threshold` to something like `buffer_size_bytes`?

components/log-ingestor/src/routes.rs
LinZhihao-723: Maybe:
- `s3` -> `s3_scanner`
- `sqs` -> `sqs_listener`

Technically they both ingest from S3.

components/log-ingestor/src/ingestion_job_manager.rs
LinZhihao-723: I kind of doubt whether we need to make `IngestionJobManager` useable as a standalone class. It's supposed to be a private struct right?

components/clp-rust-utils/src/clp_config/package/config.rs
LinZhihao-723: I prefer not to in this case: unlike a time variable where the unit can be a second, a millisecond, or else, the unit should always default to byte when we use a variable for size:
* In the config level, all size variables are not explicitly named with `_bytes`: https://github.com/y-scope/clp/blob/main/components/package-template/src/etc/clp-config.template.json.yaml
* Similarly in the code level, we use `xxx_size` directly in both the Python config and the C++ implementation:
  * https://github.com/y-scope/clp/blob/21fa311294da3796947f5f31b08e84dd9e4eb0be/components/core/src/clp_s/CommandLineArguments.hpp#L192
  * https://github.com/y-scope/clp/blob/21fa311294da3796947f5f31b08e84dd9e4eb0be/components/clp-py-utils/clp_py_utils/clp_config.py#L693

components/log-ingestor/src/routes.rs
hoophalab: sure, sounds good

components/log-ingestor/src/ingestion_job_manager.rs
hoophalab: sure

components/log-ingestor/src/ingestion_job/sqs_listener.rs
LinZhihao-723: Oh actually this error is logged at the top-level when we spawn the task.run

hoophalab:
```suggestion
                    self.process_sqs_response(result?).await?;
```
Oh I see. You are right. We probably don't need to ignore `()`?

hoophalab:
For the title, how about removing `service` in the title. `service` sounds like something specific to a service section in  `docker-compose.yaml`.

PR #1749, opened 2025-12-08
feat(helm): Add compression & query job orchestration and the webui to the chart.

junhaoliao:
> [!IMPORTANT]  
> This PR should be merged:
> 1. AFTER #1771
>    1. (AFTER #1784 as well because a bug in the test script was introduced by #1771)
> 2. AFTER #1772
>
> because changes done in #1771 & #1772 are incremental to reach the state in #1749

# Description

> [!NOTE]  

This PR enhances the CLP Package Helm chart to support compressions and queries in the WebUI. Changes include:
1. `PersistentVolumeClaim` helper
   Added clp.createPvc template for generating PVCs for components.
2. Component PVs and PVCs
   Added PV and PVC templates for:
   * archives
   * streams
   * compression-scheduler logs
   * compression-worker logs
   * query-scheduler logs
   * query-worker logs
   * reducer logs
3. Deployments
   * compression-scheduler
   * compression-worker
   * query-scheduler
   * query-worker
   * reducer

   Each deployment includes:
   * Init containers for readiness checks (database, queue, Redis, query-scheduler, results-cache)
   * Security context using first-party UID/GID (for the matching owner when writing files on the host)
   * Config and log volume mounts
   * Environment variables from Values and secrets
4. Services
   * query-scheduler
   * reducer
   * webui
5. ConfigMap updates
   * Expanded clp-config.yaml with configurations for all components:
	   * archive_output
	   * stream_output
	   * logs_input
	   * compression_scheduler
	   * compression_worker
	   * query_scheduler
	   * query_worker
	   * reducer
	   * webui
   * Added WebUI server and client settings
6. Worker concurrency are configurable via chart values.yaml.

These changes lay the groundwork for full CLP deployment on Kubernetes with all main services, logging, and persistent storage correctly configured.

# Checklist

  breaking change.

# Validation performed

## Verified that the clp-s (CLP-JSON) flavour is working
1. Ran `test.sh` to start the cluster in kind, which maps the host's `/home` and `/tmp` into the k8s host.
2. `kubectl get pods -w` and observed all pods were up.
3. Launched the webui at http://localhost:30400 and submit a job using the "Submit Compression Job" form:
   * `/tmp/clp/samples/postgresql/postgresql.log`
   * i had some test log file at so i also tried `/home/junhao/samples/postgresql.jsonl`
4. Waited for the compression job to complete successfully.
5. Went to the "Search" page in the webui and perform a search with string 1.
   * Observed the query completed successfully
   * The timeline showed up
   * Clicking on any "Original file" link on the results navigated to the log viewer with the log context shown

---

## Verified that the clp (CLP-Text) flavour is working

In `tools/deployment/package-helm/values.yaml`, modified `clpConfig.package.storage_engine` and `clpConfig.package.query_engine` both as "clp". Repeated Step 1-5 above.

## Summary by CodeRabbit

* **New Features**
  * Added deployments/services for schedulers, workers, reducer and web UI (configurable port) with init waits for dependencies.
  * Added PVs/PVCs for logs, user-logs, tmp, archives, streams and shared-data; web UI mounts client/server settings.
  * New chart helpers for consistent volume naming, PVC/volume creation, Celery env vars and enhanced wait/init logic.
  * Expanded runtime/config maps for scheduling, logging, storage, web UI and archive/stream settings.

* **Chores**
  * Bumped chart version; updated local test bootstrap to use configurable paths and sample data download.

### DIFF
--- a/tools/deployment/package-helm/templates/configmap.yaml
+++ b/tools/deployment/package-helm/templates/configmap.yaml
@@ -114,7 +114,7 @@ data:
       "ClpStorageEngine": {{ .Values.clpConfig.package.storage_engine | quote }},
       "ClpQueryEngine": {{ .Values.clpConfig.package.query_engine | quote }},
       "LogsInputType": {{ .Values.clpConfig.logs_input.type | quote }},
-      "LogsInputRootDir": "/",
+      "LogsInputRootDir": "/mnt/logs",
       "MongoDbSearchResultsMetadataCollectionName":
         {{ .Values.clpConfig.webui.results_metadata_collection_name | quote }},
       "SqlDbClpArchivesTableName": "clp_archives",
@@ -140,6 +140,7 @@ data:
         {{ .Values.clpConfig.results_cache.stream_collection_name | quote }},
       "ClientDir": "/opt/clp/var/www/webui/client",
       "LogViewerDir": "/opt/clp/var/www/webui/yscope-log-viewer",
+      "LogsInputRootDir": "/mnt/logs",
       "StreamFilesDir": "/var/data/streams",
       "StreamTargetUncompressedSize":
         {{ .Values.clpConfig.stream_output.target_uncompressed_size | int }},


tools/deployment/package-helm/templates/configmap.yaml
```diff
@@ -6,6 +6,25 @@ metadata:
     {{- include "clp.labels" . | nindent 4 }}
 data:
   clp-config.yaml: |
+    archive_output:
+      compression_level: {{ .Values.clpConfig.archive_output.compression_level }}
+      storage:
+        directory: "/var/data/archives"
+        type: "fs"
+      target_archive_size: {{ .Values.clpConfig.archive_output.target_archive_size | int }}
```
junhaoliao, line 14: Helm like to convert larger numbers into float type when building charts, (confirmed by `helm template .`), so for the larger numbers we have to add `int` to explicitly ask for unsigned numbers.

tools/deployment/package-helm/templates/archives-pv.yaml
```diff
@@ -0,0 +1,9 @@
+{{- include "clp.createLocalPv" (dict
+  "root" .
+  "name" (printf "%s-archives" (include "clp.fullname" .))
+  "component" "archives"
```
kirkrodrigues: Maybe we should have some kind of naming convention to differentiate runtime components from "data" components?

tools/deployment/package-helm/templates/compression-scheduler-deployment.yaml
kirkrodrigues, line 1: Could this have been a statefulset?

tools/deployment/package-helm/templates/compression-scheduler-deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+        {{- include "clp.labels" . | nindent 8 }}
+        app.kubernetes.io/component: "compression-scheduler"
+    spec:
+      terminationGracePeriodSeconds: 300
+      securityContext:
+        runAsUser: {{ .Values.securityContext.firstParty.uid }}
+        runAsGroup: {{ .Values.securityContext.firstParty.gid }}
+        fsGroup: {{ .Values.securityContext.firstParty.gid }}
+      initContainers:
+        - name: "wait-for-database"
```
kirkrodrigues: Don't we need to wait for the db-table-creator job? Is that not possible in K8s?

tools/deployment/package-helm/templates/compression-scheduler-deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+                do echo Waiting for redis to be ready...;
+                sleep 2;
+              done
+      containers:
+        - name: "compression-scheduler"
+          image: "{{ include "clp.image.ref" . }}"
+          imagePullPolicy: "{{ .Values.image.clpPackage.pullPolicy }}"
+          env:
+            - name: "BROKER_URL"
+              value: "amqp://{{ .Values.credentials.queue.username }}:{{ .Values.credentials.queue.password }}@{{ include "clp.fullname" . }}-queue:5672"
```
kirkrodrigues: Can we wrap these long lines?

tools/deployment/package-helm/templates/compression-scheduler-logs-pv.yaml
```diff
@@ -0,0 +1,9 @@
+{{- include "clp.createLocalPv" (dict
+  "root" .
+  "name" (printf "%s-compression-scheduler-logs" (include "clp.fullname" .))
+  "component" "compression-scheduler-logs"
```
kirkrodrigues: Should this just be `compression-scheduler` (similar to the `database` logs volume)?

tools/deployment/package-helm/templates/compression-scheduler-logs-pv.yaml
```diff
@@ -0,0 +1,9 @@
...
+  "nodeRole" "control-plane"
+  "capacity" "5Gi"
+  "accessModes" (list "ReadWriteOnce")
+  "hostPath" (printf "%s/compression-scheduler" .Values.clpConfig.logs_directory)
```
kirkrodrigues: Hm, we're using hyphen here but in the volume mount, we use underscore. Any reason we can't pick one for consistency?

tools/deployment/package-helm/templates/compression-scheduler-logs-pvc.yaml
```diff
@@ -0,0 +1,6 @@
+{{- include "clp.createPvc" (dict
+  "root" .
+  "component" "compression-scheduler-logs"
```
kirkrodrigues: Like in the PV, it's seem a bit odd to me to pass in `compression-scheduler-logs` rather than `compression-scheduler` for the component.

tools/deployment/package-helm/templates/compression-worker-deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+      volumes:
+        - name: "config"
+          configMap:
+            name: {{ include "clp.fullname" . }}-config
+        - name: "compression-worker-logs"
+          persistentVolumeClaim:
+            claimName: {{ include "clp.fullname" . }}-compression-worker-logs
+        - name: "archives"
+          persistentVolumeClaim:
+            claimName: {{ include "clp.fullname" . }}-archives
```
kirkrodrigues: Do you think it's useful to deduplicate things like these volumes since we're repeating them in different deployments? That would avoid cases where we have a typo in "archives" or something.

tools/deployment/package-helm/templates/compression-scheduler-deployment.yaml
junhaoliao: getting the status of other pods require certain k8s privileges, and there were some difficulties in getting that privilege (i.e. setting up service account) when i tried that previously. there were no correctness issues with the current setup (as in, if the table is not created, those services would exit with an exception and be restarted by k8s) so i just left it.

now i have figured out how to configure the service accounts. i think for all services we should use `kubectl wait` for waiting containers from now on.

tools/deployment/package-helm/templates/compression-worker-deployment.yaml
junhaoliao: sure. let me create some helpers for those

tools/deployment/package-helm/templates/compression-scheduler-logs-pv.yaml
junhaoliao: right. sorry for the typo

tools/deployment/package-helm/templates/compression-scheduler-logs-pv.yaml
junhaoliao: sorry it's a typo. we should use `compression_scheduler` consistently, until the day we change the service names to use `-` for both docker compose and k8s (when things are more stabled

tools/deployment/package-helm/templates/compression-scheduler-logs-pvc.yaml
junhaoliao: right, let's add a name field for this createPvc helper

tools/deployment/package-helm/templates/archives-pv.yaml
junhaoliao: sure. any ideas? i'm thinking about `"component" "clp-runtime"` now, but that's probably a bad idea?

tools/deployment/package-helm/templates/compression-scheduler-deployment.yaml
junhaoliao: mind explaining why?

for example, the reasons why we are using statefulset for the database are:
1. it can have multiple replicas with a stable network identity (though i know we have only one for DB for now)
   * the compression scheduler shouldn't need multiple replicas (at least for now?)
   * the (celery) compression scheduler shouldn't need network identity because communications happen over DB / queue / redis.
2. when multiple replicas are created, the pod creation / deletion are ordered
   * again, the compression scheduler is single replica
3. the database data (states) stays on a single host so we want to ensure each pod gets a persistent and predictable identify, so its own PVC stays bound to that identity across restarts/rescheduling
   * we are creating PVCs for the compression scheduler only for logs. the compression scheduler's state is maintained in the databse

tools/deployment/package-helm/templates/archives-pv.yaml
kirkrodrigues: Like as a prefix or?

junhaoliao: Sorry.. i definitely see their distinctions, but for some reason i was mistakenly thinking that you were referring to "archive" as a runtime component and "database" as a data component...

instead of thinking about "runtime", let's analyze their characteristics first:
* database-logs-pv, database-data-pv -> component="database" makes sense because the storage belongs to the "database" service
* archives-pv -> component="archives" can be misleading because "archives" isn't a service, it's shared storage used by compression-worker, query-worker, garbage-collector, etc.

we should still keep the "app.kubernetes.io/component:" label for both types of PV because it's a "well-known" label.

though, i think this is what we can do:
* Service-owned: Use service name: database, compression-scheduler
* Shared storage: Prefix with `shared-`: shared-archives, shared-streams, shared-staged-streams

kirkrodrigues: Sure, seems reasonable.

junhaoliao: actually, let's try this:
0. update `createLocalPv` to handle `name` differently:
```
{{/*
Creates a local PersistentVolume.

@param {object} root Root template context
@param {string} component Component category (e.g., "shared-data", "database")
@param {string} name Storage name (e.g., "archives", "logs", "data")
@param {string} nodeRole Node role for affinity. Targets nodes with label
  "node-role.kubernetes.io/<nodeRole>". Always falls back to
  "node-role.kubernetes.io/control-plane"
@param {string} capacity Storage capacity
@param {string[]} accessModes Access modes
@param {string} hostPath Absolute path on host
@return {string} YAML-formatted PersistentVolume resource
*/}}
{{- define "clp.createLocalPv" -}}
...
metadata:
  name: {{ include "clp.fullname" .root }}-{{ .component }}-{{ .name }}
  labels:
    {{- include "clp.labels" .root | nindent 4 }}
    app.kubernetes.io/component: {{ .component | quote }}
...
{{- end }}
```
for 
1. pass "shared-data" as the `component` (or `shared-storage`)
2. pass "archive" as the `name`

tools/deployment/package-helm/templates/compression-scheduler-deployment.yaml
junhaoliao: instead, i created a helper to generate this env var so we don't have to repeat this 4 times

tools/deployment/package-helm/templates/compression-scheduler-logs-pv.yaml
junhaoliao: i also updated the prototype of `createLocalPv` to match `createPvc`

tools/deployment/package-helm/templates/compression-scheduler-logs-pv.yaml
junhaoliao: do note that `staged-archives` will use hyphen, though it's not shown in the current PR (no s3 support here yet)

tools/deployment/package-helm/templates/archives-pv.yaml
kirkrodrigues: Sgtm. Can we rename component to category or component_category?

junhaoliao: component_category was picked. see #1771

tools/deployment/package-helm/test.sh
```diff
@@ -52,28 +54,46 @@ wait_for_pods() {
 }
 
 kind delete cluster --name clp-test
-rm -rf /tmp/clp
-mkdir -p /tmp/clp/var/{data,log}/{database,queue,redis,results_cache}
+rm -rf "$CLP_HOME"
+mkdir -p  "$CLP_HOME/var/"{data,log}/{database,queue,redis,results_cache}
+          "$CLP_HOME/var/data/"{archives,streams}
+          "$CLP_HOME/var/log/"{compression_scheduler,compression_worker,user}
+          "$CLP_HOME/var/log/"{query_scheduler,query_worker,reducer}
```
junhaoliao: have to split those to avoid line length violations

tools/deployment/package-helm/templates/compression-worker-deployment.yaml
```diff
@@ -62,22 +48,20 @@ spec:
               value: "/var/log/compression_worker/worker.log"
             - name: "PYTHONPATH"
               value: "/opt/clp/lib/python3/site-packages"
-            - name: "RESULT_BACKEND"
-              value: "redis://default:{{ .Values.credentials.redis.password }}@{{ include "clp.fullname" . }}-redis:6379/{{ .Values.clpConfig.redis.compression_backend_database }}"
           volumeMounts:
+            - name: "compression-worker-logs"
```
kirkrodrigues: Technically, we could create a method that just composes this name and use that both in the helpers and here, right?

tools/deployment/package-helm/templates/query-worker-deployment.yaml
```diff
@@ -0,0 +1,91 @@
...
+    spec:
+      serviceAccountName: {{ include "clp.fullname" . }}-job-watcher
+      terminationGracePeriodSeconds: 60
+      securityContext:
+        runAsUser: {{ .Values.securityContext.firstParty.uid }}
+        runAsGroup: {{ .Values.securityContext.firstParty.gid }}
+        fsGroup: {{ .Values.securityContext.firstParty.gid }}
+      initContainers:
+        - {{- include "clp.waitFor" (dict "root" . "type" "service" "name" "queue") | nindent 10 }}
+        - {{- include "clp.waitFor" (dict "root" . "type" "service" "name" "redis") | nindent 10 }}
```
kirkrodrigues: Should we have these same waiting conditions in docker compose?

tools/deployment/package-helm/templates/reducer-service.yaml
```diff
@@ -0,0 +1,18 @@
...
+spec:
+  clusterIP: "None"
+  selector:
+    {{- include "clp.selectorLabels" . | nindent 4 }}
+    app.kubernetes.io/component: "reducer"
+  ports:
+    {{- range $i := until (int .Values.workerConcurrency) }}
+    - name: "reducer-{{ $i }}"
+      port: {{ add 14009 $i }}
+      targetPort: {{ add 14009 $i }}
```
kirkrodrigues: `targetPort` is ignored when `clusterIP=None`, right?

tools/deployment/package-helm/templates/webui-deployment.yaml
```diff
@@ -0,0 +1,98 @@
...
+            - name: "CLP_DB_USER"
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "clp.fullname" . }}-database
+                  key: "username"
+            - name: "CLP_DB_PASS"
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "clp.fullname" . }}-database
+                  key: "password"
```
kirkrodrigues: Swap to alphabetize?

tools/deployment/package-helm/templates/webui-deployment.yaml
```diff
@@ -0,0 +1,98 @@
...
+            - name: "NODE_PATH"
+              value: "/opt/clp/var/www/webui/server/node_modules"
+            - name: "PORT"
+              value: "4000"
+            - name: "RATE_LIMIT"
+              value: {{ .Values.clpConfig.webui.rate_limit | quote }}
+          ports:
+            - name: "webui"
+              containerPort: 4000
+          volumeMounts:
```
kirkrodrigues, line 65: Are we missing the input logs mount?

tools/deployment/package-helm/test.sh
```diff
@@ -52,28 +54,46 @@ wait_for_pods() {
...
     extraPortMappings:
     - containerPort: 30306
       hostPort: 30306
       protocol: TCP
     - containerPort: 30017
       hostPort: 30017
       protocol: TCP
+    - containerPort: 30400
+      hostPort: 30400
+      protocol: TCP
```
kirkrodrigues: Is this for the webui? If so, why not `30000`?

tools/deployment/package-helm/templates/configmap.yaml
```diff
@@ -60,3 +108,53 @@ data:
       path: /var/log/mongodb/mongod.log
       logAppend: true
       timeStampFormat: iso8601-local
+
+  webui-client-settings.json: |
+    {
+      "ClpStorageEngine": {{ .Values.clpConfig.package.storage_engine | quote }},
+      "ClpQueryEngine": {{ .Values.clpConfig.package.query_engine | quote }},
+      "LogsInputType": {{ .Values.clpConfig.logs_input.type | quote }},
```
kirkrodrigues, line 116:
```suggestion
      "LogsInputType": {{ .Values.clpConfig.logs_input.type | quote }},
      "LogsInputRootDir": "/",
```

tools/deployment/package-helm/templates/configmap.yaml
```diff
@@ -60,3 +108,53 @@ data:
...
+      "MongoDbSearchResultsMetadataCollectionName":
+        {{ .Values.clpConfig.results_cache.stream_collection_name | quote }},
```
kirkrodrigues:
```suggestion
        {{ .Values.clpConfig.webui.results_metadata_collection_name | quote }},
```

tools/deployment/package-helm/templates/configmap.yaml
```diff
@@ -60,3 +108,53 @@ data:
...
+      "SqlDbHost": "{{ include "clp.fullname" . }}-database",
+      "SqlDbPort": 3306,
+      "SqlDbName": {{ .Values.clpConfig.database.name | quote }},
+      "SqlDbQueryJobsTableName": "query_jobs",
+      "SqlDbCompressionJobsTableName": "compression_jobs",
+      "MongoDbHost": "{{ include "clp.fullname" . }}-results-cache",
+      "MongoDbPort": 27017,
+      "MongoDbName": {{ .Values.clpConfig.results_cache.db_name | quote }},
+      "MongoDbSearchResultsMetadataCollectionName":
+        {{ .Values.clpConfig.results_cache.stream_collection_name | quote }},
```
kirkrodrigues, line 140:
```suggestion
        {{ .Values.clpConfig.webui.results_metadata_collection_name | quote }},
```

tools/deployment/package-helm/templates/configmap.yaml
```diff
@@ -60,3 +108,53 @@ data:
...
+      "MongoDbStreamFilesCollectionName": "stream-files",
```
kirkrodrigues:
```suggestion
      "MongoDbStreamFilesCollectionName":
        {{ .Values.clpConfig.results_cache.stream_collection_name | quote }},
```

tools/deployment/package-helm/templates/configmap.yaml
```diff
@@ -60,3 +108,53 @@ data:
...
+      "StreamFilesS3Profile": null,
+      "ArchiveOutputCompressionLevel": {{ .Values.clpConfig.archive_output.compression_level }},
+      "ArchiveOutputTargetArchiveSize":
+        {{ .Values.clpConfig.archive_output.target_archive_size | int }},
+      "ArchiveOutputTargetDictionariesSize":
+        {{ .Values.clpConfig.archive_output.target_dictionaries_size | int }},
+      "ArchiveOutputTargetEncodedFileSize":
+        {{ .Values.clpConfig.archive_output.target_encoded_file_size | int }},
+      "ArchiveOutputTargetSegmentSize":
+        {{ .Values.clpConfig.archive_output.target_segment_size | int }},
```
kirkrodrigues, line 158: Are we missing these in `DockerComposeController`?

tools/deployment/package-helm/templates/configmap.yaml
```diff
@@ -60,3 +108,53 @@ data:
...
+      "ClpQueryEngine": {{ .Values.clpConfig.package.query_engine | quote }},
+      "ClpStorageEngine": {{ .Values.clpConfig.package.storage_engine | quote }},
```
kirkrodrigues, line 160: Missing in DockerComposeController?

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -135,6 +135,78 @@ spec:
               operator: "Exists"
 {{- end }}
 
+{{/*
+Creates a PersistentVolumeClaim for the given component.
+
+@param {object} root Root template context
+@param {string} component_category (e.g., "shared-data", "database")
+@param {string} name (e.g., "archives", "logs", "data")
```
kirkrodrigues:
```suggestion
@param {string} component_category (e.g., "database", "shared-data")
@param {string} name (e.g., "archives", "data", "logs")
```

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -135,6 +135,78 @@ spec:
...
+    matchLabels:
+      {{- include "clp.selectorLabels" .root | nindent 6 }}
+      app.kubernetes.io/component: {{ .component_category | quote }}
+  resources:
+    requests:
+      storage: {{ .capacity }}
+{{- end }}
+
+{{/*
+Creates a volume definition that references a PersistentVolumeClaim.
```
kirkrodrigues, line 185:
```suggestion
Creates a volume definition `{{ .component_category }}-{{ .name }}` that references a
PersistentVolumeClaim.
```

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -135,6 +135,78 @@ spec:
...
+
+@param {object} root Root template context
+@param {string} component_category (e.g., "shared-data", "database")
+@param {string} name (e.g., "archives", "logs", "data")
```
kirkrodrigues:
```suggestion
@param {string} component_category (e.g., "database", "shared-data")
@param {string} name (e.g., "archives", "data", "logs")
```

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -135,6 +135,78 @@ spec:
...
+@return {string} YAML-formatted volume definition
+*/}}
+{{- define "clp.pvcVolume" -}}
+name: {{ printf "%s-%s" .component_category .name | quote }}
+persistentVolumeClaim:
+  claimName: {{ include "clp.fullname" .root }}-{{ .component_category }}-{{ .name }}
+{{- end }}
+
+{{/*
+Creates the BROKER_URL env var for Celery workers.
```
kirkrodrigues:
```suggestion
Gets the BROKER_URL env var for Celery workers.
```

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -135,6 +135,78 @@ spec:
...
+{{- define "clp.celeryBrokerUrlEnvVar" -}}
+{{- $user := .Values.credentials.queue.username -}}
+{{- $pass := .Values.credentials.queue.password -}}
+{{- $host := printf "%s-queue" (include "clp.fullname" .) -}}
+name: "BROKER_URL"
+value: {{ printf "amqp://%s:%s@%s:5672" $user $pass $host | quote }}
+{{- end }}
+
+{{/*
+Creates the RESULT_BACKEND env var for Celery workers.
```
kirkrodrigues:
```suggestion
Gets the RESULT_BACKEND env var for Celery workers.
```

tools/deployment/package-helm/templates/compression-worker-logs-pv.yaml
```diff
@@ -0,0 +1,9 @@
+{{- include "clp.createLocalPv" (dict
+  "root" .
+  "component_category" "compression-worker"
+  "name" "logs"
+  "nodeRole" "control-plane"
+  "capacity" "5Gi"
+  "accessModes" (list "ReadWriteOnce")
```
kirkrodrigues, line 7: ReadWriteMany?

tools/deployment/package-helm/templates/query-worker-logs-pv.yaml
```diff
@@ -0,0 +1,9 @@
+{{- include "clp.createLocalPv" (dict
+  "root" .
+  "component_category" "query-worker"
+  "name" "logs"
+  "nodeRole" "control-plane"
+  "capacity" "5Gi"
+  "accessModes" (list "ReadWriteOnce")
```
kirkrodrigues, line 7: ReadWriteMany?

tools/deployment/package-helm/templates/shared-data-archives-pv.yaml
```diff
@@ -0,0 +1,9 @@
+{{- include "clp.createLocalPv" (dict
+  "root" .
+  "component_category" "shared-data"
+  "name" "archives"
+  "nodeRole" "control-plane"
+  "capacity" "50Gi"
+  "accessModes" (list "ReadWriteMany")
+  "hostPath" (printf "%s/archives" .Values.clpConfig.data_directory)
```
kirkrodrigues:
```suggestion
  "hostPath" .Values.clpConfig.archive_output.storage.directory
```

Unless I'm missing something?

tools/deployment/package-helm/templates/shared-data-streams-pv.yaml
```diff
@@ -0,0 +1,9 @@
+{{- include "clp.createLocalPv" (dict
+  "root" .
+  "component_category" "shared-data"
+  "name" "streams"
+  "nodeRole" "control-plane"
+  "capacity" "20Gi"
+  "accessModes" (list "ReadWriteMany")
+  "hostPath" (printf "%s/streams" .Values.clpConfig.data_directory)
```
kirkrodrigues:
```suggestion
  "hostPath" .Values.clpConfig.stream_output.storage.directory
```

tools/deployment/package-helm/templates/query-worker-deployment.yaml
junhaoliao: actually these should not exist for the query worker. let me double check and remove them

tools/deployment/package-helm/templates/webui-deployment.yaml
junhaoliao: right, the latest code should be sync'ed

tools/deployment/package-helm/test.sh
junhaoliao: yes. the 30400 for webui was used to maintain some of consistency with the Compose flow. We can certainly change it to 30000

kirkrodrigues: > to maintain some of consistency with the Compose flow

Sorry, how do you mean?

tools/deployment/package-helm/templates/compression-worker-deployment.yaml
junhaoliao: not sure if i'm getting this. mind clarifying?

Do you mean that we can create a helper like
```
createVolumeMount(component_catagory, name)
```
which generates a `volumeMounts` element in YAML?

junhaoliao: never mind. i get what you're saying

tools/deployment/package-helm/test.sh
junhaoliao: We by default configure the port to be `30400` here in `clpConfig`: https://github.com/junhaoliao/clp/blob/67d35492775dc51de574daef1329cb545133bb9c/tools/deployment/package-helm/values.yaml#L73 , in an attempt to somewhat match the `4000` default configured here: https://github.com/y-scope/clp/blob/9650b8ce86b53b1c8f89299592721b1b4e564a11/components/package-template/src/etc/clp-config.template.json.yaml#L103

"somewhat": mask out the "thousand" digit in the port (database.port: 3306 -> 306) then add 30000 (database.port: 30306). in the case that the value becomes 0 after masking (webui.port: 4000 -> 000), right shift one digit instead of applying the mask (webui.port: 4000 -> 400), then add 30000 (webui.port: 30400).

i thought you mean we could change all such 30400 references to 30000?

tools/deployment/package-helm/templates/_helpers.tpl
junhaoliao: i tend not to mention this detail, after the addition of helper `volumeName` in cd49f3a56c55816388c25cc6ec418a1d9b903877

feel free to disagree and i can revert

tools/deployment/package-helm/test.sh
kirkrodrigues: I was following the mask and add rule, without the additional right shift. So yeah, I was suggesting replacing all 30400 references with 30000. We don't have to, it just feels more consistent to me.

tools/deployment/package-helm/templates/configmap.yaml
junhaoliao: right. submitted an issue #1799 . will address it later in the day

tools/deployment/package-helm/templates/configmap.yaml
junhaoliao: right. submitted an issue #1799 . will address it later in the day

tools/deployment/package-helm/templates/reducer-service.yaml
junhaoliao: not exactly ignored, but it's not used for routing in our uses either. i think we can remove

tools/deployment/package-helm/templates/shared-data-archives-pv.yaml
junhaoliao: right, we should use `.Values.clpConfig.archive_output.storage.directory`

tools/deployment/package-helm/templates/shared-data-streams-pv.yaml
junhaoliao: right

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -123,7 +139,7 @@ spec:
   persistentVolumeReclaimPolicy: "Retain"
   storageClassName: "local-storage"
   local:
-    path: {{ .hostPath }}
+    path: {{ .hostPath | quote }}
```
junhaoliao, line 142: not directly related to this PR but we should fix, for consistency. and just in case users configure paths with `#` or `:` which doesn't automatically be considered part of a string in the YAML syntax

tools/deployment/package-helm/templates/compression-worker-logs-pv.yaml
junhaoliao: i probably get what you're saying - when we scale these workers to a multi-node cluster, a ReadWriteMany (RWX) access mode is needed by a multi-node accessible PV. however, note that `local-storage` doesn't support RWX because they only work on local nodes. (there is no reason to launch multiple replicas of query/compression workers on a single node; even if we do, do note that volumes / claims are created per node rather than per pod so RWO is still appropriate)

for now, we still expect users to set up some distributed storage (e.g. SeaweedFS and mount them to a local path) on individual hosts. in the future, once we implement #1760 , we really shouldn't create mounts for logs any more.

kirkrodrigues: Hm, then why do we need `ReadWriteMany` for compression-worker-tmp-pv?

junhaoliao: I think that's a typo. sorry

speaking of that tmp directory, we don't really need it to be mapped to the host unless we want to debug what's passed from the Python code to the clp binaries. in the near future, we should probably revisit whether all such host-container mounts are necessary in both docker compose and k8s. you can ask the rabbit to create an issue if you agree

kirkrodrigues: Hey @coderabbitai, please create an issue about the above.

tools/deployment/package-helm/templates/query-worker-logs-pv.yaml
junhaoliao: discussed at https://github.com/y-scope/clp/pull/1749/files#r2630640491

kirkrodrigues (approved):

tools/deployment/package-helm/values.yaml
```diff
@@ -18,25 +18,101 @@ image:
...
     port: 30306
     name: "clp-db"
 
+  compression_scheduler:
+    jobs_poll_delay: 0.1  # seconds
+    logging_level: "INFO"
+    max_concurrent_tasks_per_job: 0  # A value of 0 disables the limit
+
+  query_scheduler:
+    port: 7000
```
junhaoliao, line 48: Note this value is unused. Since query_scheduler is not a user interfacing service, we should remove this configuration field to avoid confusion. @coderabbitai submit an issue

tools/deployment/package-helm/templates/query-scheduler-deployment.yaml
```diff
@@ -0,0 +1,96 @@
...
+            "root" .
+            "type" "service"
+            "name" "queue"
+          ) | nindent 10 }}
+        - {{- include "clp.waitFor" (dict
+            "root" .
+            "type" "service"
+            "name" "redis"
+          ) | nindent 10 }}
+      containers:
```
junhaoliao, line 44: readinessProbeTimings and livenessProbeTimings could have been added to enhance healthcheck and also match `docker-compose-all.yaml`. @coderabbitai submit an issue

Issue #764, opened 2025-03-12
feat(clang-tidy): Fix all warnings and errors for all C++ source files.


davidlion:
### Request

https://github.com/y-scope/clp/commit/36140bf46b8d4d93b4383c2b17ebe6dc5cfebe1e adds support for running clang-tidy through `task` and in the GH workflows.
Currently, the task `lint:check-cpp-static-full` succeeds thanks to an exclude list that skips all files with known clang-tidy issues (see https://github.com/y-scope/clp/blob/main/lint-tasks.yml for the task and exclude list).

The goal is to fix all the clang-tidy warnings and errors so that `lint:check-cpp-static-full` can succeed without the need for an exclude list.

### Possible implementation

When you create a PR fixing clang-tidy issues:
1. Link to this issue for tracking.
2. Remove the file from `lint:check-cpp-static-full`'s exclude list.
3. Comment on this issue, documenting any difficult fixes or discussion/guidelines that resulted from fixes.

LinZhihao-723:
Add line filter to bypass [this](https://github.com/msgpack/msgpack-c/issues/1098) msgpack issue in #890.

PR #1751, opened 2025-12-09
feat(core): Add runtime CA bundle path detection for `CurlDownloadHandler`; Resolve clang-tidy violations in `clp/CurlDownloadHandler.{cpp,hpp}`.

Bill-hbrhbr:
# Description

Part of #764 CPP lint fixes.

This change updates `CurlDownloadHandler` to support runtime CA bundle path detection across a variety of execution environments. The handler now resolves the CA bundle using a stable search order: 
* check CURL_CA_BUNDLE and SSL_CERT_FILE
* probe well known Linux CA bundle locations
If both are to no avail, we fall back to curl’s default certificate store when it's built.. 

This makes curl operations behave consistently on Linux systems even when the runtime environment differs from the
build environment, while macOS and Windows naturally continue to use their native CA bundle trust stores.

# Checklist

  breaking change.

# Validation performed

* [X] Tested cross-compilation works by building on manylinux_2_28 and running on ubuntu/debian.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SSL/TLS certificate handling for downloads: the app detects and uses a host CA bundle on Linux-like systems (respecting common environment variables and distro locations) and skips host bundle resolution on macOS, preserving system stores on Windows and macOS for more reliable secure download verification.

* **Chores**
  * Static analysis now includes the download-handling sources to ensure continued code quality and maintenance.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -150,7 +150,7 @@ checksum = "c7c24de15d275a1ecfd47a380fb4d5ec9bfe0933f309ed5e705b775596a3574d"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.111",
 ]
 
 [[package]]
@@ -161,7 +161,7 @@ checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.111",
 ]
 
 [[package]]
@@ -874,7 +874,7 @@ dependencies = [
  "heck",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.111",
 ]
 
 [[package]]
@@ -892,6 +892,7 @@ dependencies = [
  "aws-sdk-sqs",
  "brotli",
  "hex",
+ "non-empty-string",
  "num_enum",
  "rmp-serde",
  "secrecy",
@@ -1151,7 +1152,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "strsim",
- "syn",
+ "syn 2.0.111",
 ]
 
 [[package]]
@@ -1162,7 +1163,7 @@ checksum = "d38308df82d1080de0afee5d069fa14b0326a88c14f15c5ccda35b4a6c414c81"
 dependencies = [
  "darling_core",
  "quote",
- "syn",
+ "syn 2.0.111",
 ]
 
 [[package]]
@@ -1170,6 +1171,17 @@ name = "data-encoding"
 version = "2.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2a2330da5de22e8a3cb63252ce2abb30116bf5265e89c0e01bc17015ce30a476"
+
+[[package]]
+name = "delegate"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "082a24a9967533dc5d743c602157637116fc1b52806d694a5a45e6f32567fcdd"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
 
 [[package]]
 name = "der"
@@ -1209,7 +1221,7 @@ checksum = "d65d7ce8132b7c0e54497a4d9a55a1c2a0912a0d786cf894472ba818fba45762"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.111",
 ]
 
 [[package]]
@@ -1220,7 +1232,7 @@ checksum = "ef941ded77d15ca19b40374869ac6000af1c9f2a4c0f3d4c70926287e6364a8f"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.111",
 ]
 
 [[package]]
@@ -1242,7 +1254,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "rustc_version",
- "syn",
+ "syn 2.0.111",
  "unicode-xid",
 ]
 
@@ -1266,7 +1278,7 @@ checksum = "97369cbbc041bc366949bc74d34658d6cda5621039731c6310521892a3a20ae0"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.111",
 ]
 
 [[package]]
@@ -1331,7 +1343,7 @@ dependencies = [
  "heck",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.111",
 ]
 
 [[package]]
@@ -1495,7 +1507,7 @@ checksum = "162ee34ebcb7c64a8abebc059ce0fee27c2262618d7b60ed8faf72fef13c3650"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.111",
 ]
 
 [[package]]
@@ -2141,6 +2153,8 @@ dependencies = [
  "clap",
  "clp-rust-utils",
  "const_format",
+ "log",
+ "non-empty-string",
  "secrecy",
  "serde",
  "serde_json",
@@ -2173,7 +2187,7 @@ dependencies = [
  "macro_magic_core",
  "macro_magic_macros",
  "quote",
- "syn",
+ "syn 2.0.111",
 ]
 
 [[package]]
@@ -2187,7 +2201,7 @@ dependencies = [
  "macro_magic_core_macros",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.111",
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


components/core/src/clp/CurlDownloadHandler.cpp
```diff
@@ -115,4 +120,30 @@ CurlDownloadHandler::CurlDownloadHandler(
...
+        return std::string(cCentOsCaBundlePath);
+    }
+    throw CurlOperationFailed(
+            ErrorCode_Failure,
+            __FILE__,
+            __LINE__,
+            CURLE_SSL_CACERT_BADFILE,
+            "`CurlDownloadHandler` failed to find the CA bundle file path."
+    );
+}
```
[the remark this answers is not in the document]
Bill-hbrhbr: Let's add additional CA bundle path if their use cases actually occur.

components/core/src/clp/CurlDownloadHandler.hpp
```diff
@@ -91,6 +92,27 @@ class CurlDownloadHandler {
...
+     * Resolution order:
+     *   1. Environment variables CURL_CA_BUNDLE then SSL_CERT_FILE, if set.
+     *   2. Well-known CA bundle paths.
+     *   3. If nothing is found, return an empty string. The caller needs to check the output so as
+     *      not to override curl defaults and let it use the platform's own certificate store or
+     *      whatever it was built with.
+     *
+     * @return Absolute path to the CA bundle file.
+     */
+    static auto get_host_ca_bundle_path() -> std::string;
```
LinZhihao-723:
```suggestion
    [[nodiscard]] static auto get_host_ca_bundle_path() -> std::optional<std::string>;
```
Let's return `std::optional` instead.

components/core/src/clp/CurlDownloadHandler.hpp
```diff
@@ -91,6 +92,27 @@ class CurlDownloadHandler {
...
+
+    static constexpr std::string_view cDebianCaBundlePath = "/etc/ssl/certs/ca-certificates.crt";
+    static constexpr std::string_view cCentOsCaBundlePath = "/etc/pki/tls/certs/ca-bundle.crt";
```
LinZhihao-723:
```suggestion
    static constexpr std::string_view cDebianCaBundlePath{"/etc/ssl/certs/ca-certificates.crt"};
    static constexpr std::string_view cCentOsCaBundlePath{"/etc/pki/tls/certs/ca-bundle.crt"};
```
Move these into `get_host_ca_bundle_path`.
Just a heads up that static vars should come before methods in a class declaration.

components/core/src/clp/CurlDownloadHandler.hpp
```diff
@@ -91,6 +92,27 @@ class CurlDownloadHandler {
...
+     * native certificate stores and are not checked. See also: https://curl.se/docs/sslcerts.html
+     *
+     * Resolution order:
+     *   1. Environment variables CURL_CA_BUNDLE then SSL_CERT_FILE, if set.
+     *   2. Well-known CA bundle paths.
+     *   3. If nothing is found, return an empty string. The caller needs to check the output so as
+     *      not to override curl defaults and let it use the platform's own certificate store or
+     *      whatever it was built with.
+     *
+     * @return Absolute path to the CA bundle file.
```
LinZhihao-723:
```suggestion
     * Locate the certificate authority (CA) bundle file available on the current host.
     *
     * This method performs a runtime lookup to provide a predictable CA bundle path for libcurl on
     * Linux-based systems. It is intended to handle cases where the runtime environment differs
     * from the build environment. macOS typically use their native system certificate stores and
     * are not searched by this function.
     *
     * Resolution order (first match wins):
     *   1. Environment variables, checked in order: `CURL_CA_BUNDLE`, then `SSL_CERT_FILE`. If the
     *      variable is set and points to an existing, readable file, that path is returned.
     *   2. Known distribution-specific default bundle locations (checked in order):
     *      - Debian / Ubuntu: `/etc/ssl/certs/ca-certificates.crt`
     *      - CentOS / RHEL / Fedora: `/etc/pki/tls/certs/ca-bundle.crt`
     *
     * @return Absolute path to the discovered CA bundle file.
     * @return std::nullopt if no CA bundle file could be located. The caller should handle this
     * case by relying on libcurl's built-in defaults.
```

components/core/src/clp/CurlDownloadHandler.cpp
```diff
@@ -115,4 +123,29 @@ CurlDownloadHandler::CurlDownloadHandler(
...
+    }
+    if (std::filesystem::exists(cDebianCaBundlePath)) {
+        return std::string(cDebianCaBundlePath);
+    }
+    if (std::filesystem::exists(cCentOsCaBundlePath)) {
+        return std::string(cCentOsCaBundlePath);
+    }
+    return {};
+#endif
+}
```
LinZhihao-723:
```suggestion
auto CurlDownloadHandler::get_host_ca_bundle_path() -> std::optional<std::string> {
    if constexpr (Platform::MacOs == cCurrentPlatform) {
        return std::nullopt;
    }

    // Read-only operation. No multithreaded context.
    // NOLINTNEXTLINE(concurrency-mt-unsafe)
    if (auto const* path{std::getenv("CURL_CA_BUNDLE")}; nullptr != path) {
        return path;
    }
    // Read-only operation. No multithreaded context.
    // NOLINTNEXTLINE(concurrency-mt-unsafe)
    if (auto const* path{std::getenv("SSL_CERT_FILE")}; nullptr != path) {
        return path;
    }
    if (std::filesystem::exists(cDebianCaBundlePath)) {
        return std::string{cDebianCaBundlePath};
    }
    if (std::filesystem::exists(cCentOsCaBundlePath)) {
        return std::string{cCentOsCaBundlePath};
    }
    return std::nullopt;
}
```
- Use `Platform.hpp`. Don't use macro.
- Use `{}` for initialization.
- Use `auto const*`, not `auto* const`.
- Don't use `x = y` in an if condition.

components/core/src/clp/CurlDownloadHandler.cpp
```diff
@@ -42,7 +44,13 @@ CurlDownloadHandler::CurlDownloadHandler(
     }
 
     // Set up src url
-    m_easy_handle.set_option(CURLOPT_URL, src_url.data());
+    m_easy_handle.set_option(CURLOPT_URL, std::string(src_url).c_str());
```
LinZhihao-723:
```suggestion
    m_easy_handle.set_option(CURLOPT_URL, std::string{src_url}.c_str());
```

components/core/src/clp/CurlDownloadHandler.cpp
```diff
@@ -42,7 +44,13 @@ CurlDownloadHandler::CurlDownloadHandler(
...
+
+    // Set up CA bundle path
+    auto const ca_bundle_path = get_host_ca_bundle_path();
+    if (false == ca_bundle_path.empty()) {
+        m_easy_handle.set_option(CURLOPT_CAINFO, ca_bundle_path.c_str());
+    }
```
LinZhihao-723:
```suggestion
    if (auto const optional_ca_bundle_path{get_host_ca_bundle_path()};
        optional_ca_bundle_path.has_value())
    {
        m_easy_handle.set_option(CURLOPT_CAINFO, optional_ca_bundle_path->c_str());
        return;
    }
```

LinZhihao-723 (changes requested):

components/core/src/clp/CurlDownloadHandler.hpp
```diff
@@ -11,6 +11,7 @@
 #include <unordered_map>
 
 #include <curl/curl.h>
+#include <curl/system.h>
```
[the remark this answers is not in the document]
Bill-hbrhbr: it's needed for curl_off_t

components/core/src/clp/CurlDownloadHandler.cpp
```diff
@@ -115,4 +125,31 @@ CurlDownloadHandler::CurlDownloadHandler(
...
+        return path;
+    }
+    if (std::filesystem::exists(cDebianCaBundlePath)) {
+        return std::string{cDebianCaBundlePath};
+    }
+    if (std::filesystem::exists(cCentOsCaBundlePath)) {
+        return std::string{cCentOsCaBundlePath};
+    }
+    return std::nullopt;
+}
```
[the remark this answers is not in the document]
Bill-hbrhbr: @LinZhihao-723 Do we want to validate the env values? I personally don't think we need to. If someone set these values, they probably know what they are doing with the CA bundle files.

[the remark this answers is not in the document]
LinZhihao-723: Yeah I think it's ok to not check it.

components/core/src/clp/CurlDownloadHandler.hpp
```diff
@@ -91,6 +92,27 @@ class CurlDownloadHandler {
     [[nodiscard]] auto perform() -> CURLcode { return m_easy_handle.perform(); }
 
 private:
+    /**
+     * Locate the certificate authority (CA) bundle file available on the current host.
```
LinZhihao-723:
```suggestion
     * Locates the certificate authority (CA) bundle file available on the current host.
```

LinZhihao-723 (approved):

PR #1757, opened 2025-12-09
feat(clp-s)!: Change how timestamps are specified in the AST and KQL to take advantage of `clp_s::timestamp_parser`:

gibber9809:
- Refactor `clp_s::search::ast::TimestampLiteral` to rely on nanosecond precision timestamps; do not allow `TimestampLiteral` to be interpreted as a string; do not parse timestamps while constructing `TimestampLiteral`.
- Remove undocumented `date()` expression from KQL and replace it with `timestamp()` expression.
- Clean up KQL grammar and parse tree visitor.
-  Ensure that `TimestampLiteral` operands are converted to millisecond precision for comparison against the timestamp range index.
- Add `SetTimestampLiteralPrecision` transformation pass to set the precision of integer literals returned by `TimestampLiteral::as_int` to a set precision for all `TimestampLiteral`s in the AST.
- Use `SetTimestampLiteralPrecision` to treat timestamps as millisecond precision for archive and kv-ir search.
 

# Description

This PR makes significant changes to the KQL grammar and AST in order to integrate `clp_s::timestamp_parser`, and make sure that operations involving timestamps occur at well-defined precisions. These changes have been split out of a larger set of changes aimed at introducing a new timestamp column type that relies on `clp_s::timestamp_parser` while deprecating `clp_s::TimestampPattern` and the `DateString` column type.

Note that a large part of the diff in this PR comes from generated KQL code, which can be ignored by reviewers.

In the KQL grammar we remove the undocumented `date()` expression that relied on `clp_s::TimestampPattern`, and replace it with a new `timestamp()` expression with better defined semantics.

The timestamp expression has the following grammar `'timestamp(' QUOTED_STRING ( ',' QUOTED_STRING )? ')'`, with the semantics that the first quoted string is a timestamp, and the second optional quoted string is a timestamp pattern. When a pattern isn't provided the parser will attempt to parse the provided timestamp using the default set of timestamp patterns from `timestamp_parser::get_all_default_quoted_timestamp_patterns`. The optional pattern follows the same semantics as the patterns accepted by `clp_s::timestamp_parser`.

Documentation for the `timestamp()` expression will be added in a separate PR.

The `DateLiteral` in the AST has been significantly simplified, as summarized below:
- They can now only be constructed in terms of a nanosecond precision timestamp
- They can no longer match string columns
- They have configurable timestamp precision for the epoch timestamp returned by `as_int` (effectively, the actual value we compare the timestamp column against in search).

The `AddTimestampConditions` and `EvaluateTimestampIndex` passes have been updated to account for these new semantics.

The `SetDateLiteralPrecision` pass has been added so that we can set the timestamp precision we compare against at search time for the whole AST -- it is being used to ensure that timestamp comparisons for kv-ir streams and current archives happen in millisecond precision.

This overall change is breaking for two reasons:
- The query language has a breaking change caused by the removal of the `date()` expression.
- Since the `DateLiteral` is now correctly converted to millisecond precision for comparison against the timestamp range index certain situations involving undefined behaviour caused by the stored timestamps not actually being millisecond precision will now behave differently.

# Checklist

  breaking change.

# Validation performed

* Added tests for parsing timestamp expressions
* Manually validated that `--tge`, `--tle`, and comparisons against `timestamp()` expressions return results as expected for v0.4.1 archives when using the `clp-s` binary.

## Summary by CodeRabbit

* **New Features**
  * Added native timestamp(...) expressions in queries with configurable precisions (s, ms, µs, ns).
  * Query parser now accepts and validates timestamp operands; malformed timestamp expressions are rejected.

* **Refactor**
  * Unified and modernized timestamp literal handling across search and evaluation pipelines for consistent precision semantics.

* **Tests**
  * Added tests validating correct parsing and rejection of invalid timestamp expressions.


components/core/src/clp_s/search/ast/DateLiteral.hpp
```diff
@@ -46,17 +45,35 @@ class DateLiteral : public Integral {
...
+    // Methods
+    /**
+     * @param precision
+     * @return The literal timestamp as the given precision (rounded towards zero).
+     */
+    auto as_precision(Precision precision) -> epochtime_t;
+
+    /**
+     * Sets the precision of the timestamp returned by `as_int`. The returned value from `as_int` is
+     * the same as the value obtained returned by calling `as_precision(precision)`.
```
20001020ycx:
```suggestion
     * the same as the value returned from `as_precision(precision)`.
```

components/core/src/clp_s/search/ast/DateLiteral.cpp
```diff
@@ -2,77 +2,55 @@
...
+constexpr epochtime_t cNanosecondsInSecond{1000LL * cNanosecondsInMillisecond};
+}  // namespace
+
+DateLiteral::DateLiteral(epochtime_t v)
+        : m_timestamp{v},
+          m_default_precision_timestamp{v},
+          m_double_timestamp{static_cast<double>(v) / cNanosecondsInSecond} {}
+
+std::shared_ptr<Literal> DateLiteral::create(epochtime_t v) {
+    return std::shared_ptr<Literal>(static_cast<Literal*>(new DateLiteral(v)));
```
20001020ycx: Why do we need to statically cast to `Literal*` first? Shouldn't `std::shared_ptr<Literal>(new DateLiteral(v));` sufficient as `DateLiteral` inherits from `Literal`

components/core/src/clp_s/search/ast/SetDateLiteralPrecision.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include "DateLiteral.hpp"
+#include "Expression.hpp"
+#include "Transformation.hpp"
+
+namespace clp_s::search::ast {
+/**
+ * Transformation pass that sets the precision of all date literals in the expression AST to a
+ * given precision.
+ */
+class SetDateLiteralPrecision : public Transformation {
```
20001020ycx: Do we need comment in this class?

components/core/src/clp_s/search/ast/SetDateLiteralPrecision.cpp
```diff
@@ -0,0 +1,34 @@
+#include "SetDateLiteralPrecision.hpp"
+
+#include <memory>
+#include <vector>
+
+#include "DateLiteral.hpp"
+#include "Expression.hpp"
+
+namespace clp_s::search::ast {
+auto SetDateLiteralPrecision::run(std::shared_ptr<Expression>& expr)
```
20001020ycx: Missing docstring, I think what you have in the pr description is clear:
> Add SetDateLiteralPrecision transformation pass to set the precision of integer literals returned by DateLiteral::as_int to a set precision for all DateLiterals in the AST.

components/core/src/clp_s/search/ast/DateLiteral.cpp
```diff
@@ -2,77 +2,55 @@
...
-        return false;
-    }
+bool DateLiteral::as_float(double& ret, FilterOperation op) {
+    ret = m_double_timestamp;
+    return true;
+}
 
-    if (m_epoch_str.find(' ') != std::string::npos) {
-        return false;
+auto DateLiteral::as_precision(Precision precision) -> epochtime_t {
```
20001020ycx: Just to confirm my understanding, this method is used for comparison with different precision. After the optimization pass in AST (SetDateLiteralPercision), every leaf node with date literal is to be set in nanosecond, so this serves as the casting of the precision for comparison with other date literal with different precision?

Does this mean once we support nanosecond precision in both user queries and archives, this normalization step becomes unnecessary?

20001020ycx: Also, I didn't quite get what you said about this after reading the code:
> I guess the other thing to note is that for stage 1/2 that you've outlined above all of the precisions are assumed/technically we don't actually know them.

So looks like we somehow know the precision in here?

components/core/src/clp_s/search/ast/SetDateLiteralPrecision.hpp
20001020ycx: I mean the comment such as `// Members` & `// Constructors`. The comment for the class looks good for me

gibber9809: As far as I can tell those kinds of comments aren't actually part of our coding standard, they're just convention -- besides docstrings, it seems like the only class-level commenting guidelines are about documenting default/deleted constructors.

In this specific instance I think it's cleaner to not have the extra comments for `// Members` etc., so I'll get rid of them.

components/core/src/clp_s/search/ast/DateLiteral.cpp
gibber9809: I think the first thing to note is that `SetDateLiteralPrecision` will change the precision to whatever you configure it to, which is millisecond precision for all of the use-cases in this PR.

To give full context, we do know the precision of the timestamp stored in the `DateLiteral` itself, since it's always epoch nanoseconds by construction. The `as_precision` method allows you to convert that nanosecond timestamp into a different precision epoch timestamp, which is useful for comparing against other epoch timestamps that aren't in epoch nanosecond precision. 

For us, the need to do comparisons in precisions other than nanosecond comes up in the following three scenarios:

1. Comparisons against timestamps in the timestamp range index -- by definition the start and end timestamps in this range are in millisecond precision (note that this isn't actually enforced in current archives though)
2. Comparisons against timestamps stored in kv-ir -- these timestamps are almost always going to be in millisecond precision, though this technically isn't guaranteed by the spec
3. Comparisons against timestamps stored in current archives -- when stored as `DateTime` or `Integer` they will often be millisecond precision, though are precisions are allowed

Basically in the current version of the code we're mostly certain about the precisions for the first two cases, and fairly uncertain in the 3rd case, though for all of these the best guess is millisecond precision.

Once we update the archive format the 3rd case is no longer an issue, and the undefined behaviour for the 1st case goes away (note that the range will still be stored in milliseconds, we'll just be actually ensuring that it is stored in milliseconds).

components/core/src/clp_s/search/ast/SetDateLiteralPrecision.cpp
gibber9809: Sure -- the transformation passes typically don't have a docstring for the `run` method, and instead leave the explanation for the class comment, but that's probably bad practice. I'll add a docstring.

components/core/src/clp_s/search/ast/DateLiteral.cpp
gibber9809: Yeah, just didn't think about this enough when re-writing the old code -- changed it.

components/core/src/clp_s/search/ast/DateLiteral.hpp
```diff
@@ -46,17 +45,35 @@ class DateLiteral : public Integral {
...
+
+    bool as_float(double& ret, FilterOperation op) override;
 
-    bool as_var_string(std::string& ret, FilterOperation op) override;
+    // Methods
+    /**
+     * @param precision
+     * @return The literal timestamp as the given precision (rounded towards zero).
+     */
+    auto as_precision(Precision precision) -> epochtime_t;
```
LinZhihao-723: Shall we make this method const and add `[[nodiscard]]`?

```suggestion
    [[nodiscard]] auto as_precision(Precision precision) const -> epochtime_t;
```

components/core/src/clp_s/search/ast/DateLiteral.hpp
```diff
@@ -46,17 +45,35 @@ class DateLiteral : public Integral {
...
+     * Sets the precision of the timestamp returned by `as_int`. The returned value from `as_int` is
+     * the same as the value returned from `as_precision(precision)`.
+     * @param precision
+     */
+    void set_default_precision(Precision precision);
 
 private:
-    std::string m_epoch_str;
+    // Types
+    static constexpr literal_type_bitmask_t cDateLiteralTypes = EpochDateT;
```
LinZhihao-723:
```suggestion
    static constexpr literal_type_bitmask_t cDateLiteralTypes{EpochDateT};
```

components/core/src/clp_s/search/ast/DateLiteral.hpp
```diff
@@ -1,38 +1,37 @@
...
 /**
  * Class for Date literal in the search AST. Represents time
  * in epoch time.
  */
-class DateLiteral : public Integral {
+class DateLiteral : public Literal {
 public:
-    // Deleted copy
+    // Types
+    enum Precision : uint8_t {
```
LinZhihao-723: Shall we use enum class?

components/core/src/clp_s/search/ast/TimestampLiteral.hpp
```diff
@@ -1,38 +1,37 @@
...
-    static std::shared_ptr<Literal> create_from_float(double v);
-    static std::shared_ptr<Literal> create_from_int(epochtime_t v);
-
-    /**
-     * Attempt to create a Date literal from string. Tries to parse the string using
-     * TimestampPattern.
-     * @return A Date Literal or nullptr if the string can not be parsed as date.
-     */
-    static std::shared_ptr<Literal> create_from_string(std::string const& v);
+    static std::shared_ptr<Literal> create(epochtime_t v);
```
LinZhihao-723, line 31: Since this is the factory function, shall we move it before constructors and after types?

components/core/src/clp_s/search/kql/kql.cpp
```diff
@@ -76,6 +79,72 @@ class ParseTreeVisitor : public KqlBaseVisitor {
...
+        }
+
+        auto const timestamp_str{ctx.timestamp->getText()};
+        std::string generated_pattern;
+        if (ctx.pattern) {
+            auto const pattern_str{ctx.pattern->getText()};
+            auto const pattern_result{timestamp_parser::TimestampPattern::create(pattern_str)};
+            if (pattern_result.has_error()) {
+                SPDLOG_ERROR(
+                        "Invalid timestamp pattern {} - {}.",
```
LinZhihao-723:
```suggestion
                        "Invalid timestamp pattern {} - {}",
```
For consistency

components/core/src/clp_s/search/kql/kql.cpp
```diff
@@ -76,6 +79,72 @@ class ParseTreeVisitor : public KqlBaseVisitor {
...
+    static std::shared_ptr<Literal> create_timestamp_literal(
+            KqlParser::Timestamp_expressionContext const& ctx
+    ) {
+        if (nullptr == ctx.timestamp) {
+            return nullptr;
+        }
+
+        auto const timestamp_str{ctx.timestamp->getText()};
+        std::string generated_pattern;
+        if (ctx.pattern) {
```
LinZhihao-723:
```suggestion
        if (nullptr != ctx.pattern) {
```

components/core/src/clp_s/search/kql/kql.cpp
```diff
@@ -76,6 +79,72 @@ class ParseTreeVisitor : public KqlBaseVisitor {
...
+                generated_pattern
+        )};
+        if (false == optional_timestamp.has_value()) {
+            SPDLOG_ERROR(
+                    "Failed to parse timestamp {} using default timestamp patterns.",
+                    timestamp_str
+            );
+            return nullptr;
+        }
+        return DateLiteral::create(optional_timestamp.value().first);
```
LinZhihao-723: Very nit: we can dereference it:

```suggestion
        return DateLiteral::create(optional_timestamp->first);
```

components/core/src/clp_s/search/AddTimestampConditions.hpp
```diff
@@ -14,7 +14,8 @@
 namespace clp_s::search {
 /**
  * Transformation pass which takes optional begin and end filters on a timestamp column and augments
- * the AST with additional filters on the provided timestamp column.
+ * the AST with additional filters on the provided timestamp column. The begin and end timestamp are
```
LinZhihao-723:
```suggestion
 * the AST with additional filters on the provided timestamp column. The begin and end timestamps
 * are assumed to be in epoch milliseconds.
```

LinZhihao-723 (changes requested):
Overall lgtm. Ignored some legacy format issues and only commented on the newly added code.
One question for date literal: does it make more sense if we name it `TimetampLiteral`?
- We call date-related specification "timestamp expression".
- We use `timestamp(xxx, xxx)` to specify a date.

gibber9809:
> * related

Yeah I agree `TimestampLiteral` makes sense -- I'll change the naming throughout the code.

components/core/src/clp_s/search/ast/TimestampLiteral.hpp
```diff
@@ -0,0 +1,83 @@
+#ifndef CLP_S_SEARCH_DATELITERAL_HPP
+#define CLP_S_SEARCH_DATELITERAL_HPP
```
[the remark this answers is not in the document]
LinZhihao-723: I agree with the rabbit

components/core/src/clp_s/search/ast/TimestampLiteral.hpp
LinZhihao-723, line 1: @20001020ycx can u do a follow up PR to fix clang-tidy warnings in this file (and the implementation)?
I think it might be a good small task for you to get familiar with the C++ coding standard

components/core/src/clp_s/search/ast/SetTimestampLiteralPrecision.hpp
```diff
@@ -0,0 +1,32 @@
+#ifndef CLP_S_SEARCH_AST_SETDATELITERALPRECISION_HPP
+#define CLP_S_SEARCH_AST_SETDATELITERALPRECISION_HPP
```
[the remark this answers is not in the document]
LinZhihao-723: I agree with the rabbit

PR #1763, opened 2025-12-10
refactor(clp-s): Add utility functions to create and decompose the archive version; Add constructors for `ArchiveHeader`.

gibber9809:
# Description

This PR adds the constexpr utility functions `make_archive_version` and `decompose_archive_version` to make it easier to create a 32-bit archive version out of a major, minor, and patch version and vice versa. The main intention of this change is to avoid issues where some pieces of code create or interpret archive versions incorrectly.

Having these utilities will make it easier to compare archive versions, which will be useful in an upcoming PR that needs to introduce some special behaviour for older archives in order to maintain backwards compatibility.

e.g.,
```
if (archive_reader->get_header().version < make_archive_version(0, 5, 0) {
    // special case for older archives
}
```

# Checklist

  breaking change.

# Validation performed

* Added extra validation in the end to end clp-s tests to check that the archive header (including the archive version) is written and read correctly.

## Summary by CodeRabbit

* **Improvements**
  * Updated archive header format and standardized version encoding for consistent create/read flows.
  * Streamlined header initialization and magic-number handling for more robust archive files.

* **Utilities**
  * Added helpers to compose and decompose archive version components.

* **Tests**
  * End-to-end tests now validate archive headers, version components, sizes, padding and reserved fields after compression.

### DIFF
--- a/integration-tests/tests/test_package_start.py
+++ b/integration-tests/tests/test_package_start.py
@@ -4,6 +4,7 @@ import pytest
 
 from tests.utils.asserting_utils import (
     validate_package_running,
+    validate_running_mode_correct,
 )
 from tests.utils.clp_mode_utils import CLP_MODE_CONFIGS
 from tests.utils.config import PackageInstance
@@ -21,3 +22,6 @@ def test_clp_package(fixt_package_instance: PackageInstance) -> None:
     """
     # Ensure that all package components are running.
     validate_package_running(fixt_package_instance)
+
+    # Ensure that the package is running in the correct mode.
+    validate_running_mode_correct(fixt_package_instance)
--- a/integration-tests/tests/utils/asserting_utils.py
+++ b/integration-tests/tests/utils/asserting_utils.py
@@ -6,9 +6,13 @@ import subprocess
 from typing import Any
 
 import pytest
+from clp_py_utils.clp_config import ClpConfig
+from pydantic import ValidationError
 
+from tests.utils.clp_mode_utils import compare_mode_signatures
 from tests.utils.config import PackageInstance
 from tests.utils.docker_utils import list_running_services_in_compose_project
+from tests.utils.utils import load_yaml_to_dict
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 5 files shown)


components/core/src/clp_s/SingleFileArchiveDefs.hpp
```diff
@@ -3,15 +3,49 @@
...
+ */
+constexpr auto decompose_archive_version(uint32_t archive_version)
+        -> std::tuple<uint8_t, uint8_t, uint16_t> {
+    constexpr uint32_t cMajorVersionOffset{24U};
+    constexpr uint32_t cMinorVersionOffset{16U};
+    uint8_t const major_version{static_cast<uint8_t>(archive_version >> cMajorVersionOffset)};
+    uint8_t const minor_version{static_cast<uint8_t>(archive_version >> cMinorVersionOffset)};
+    uint16_t const patch_version{static_cast<uint16_t>(archive_version)};
+    return std::make_tuple(major_version, minor_version, patch_version);
+}
```
gibber9809, line 42: I didn't separate declaration and definition here since constexpr functions need their definition available before first use, i.e. I can't move the body to a `.cpp` file, and if I separated out the definitions they would still have to appear before line 47.

LinZhihao-723: That's fine.

components/core/src/clp_s/SingleFileArchiveDefs.hpp
```diff
@@ -3,15 +3,49 @@
...
+ * @param archive_version
+ * @return A tuple containing the major, minor, and patch version for an archive.
+ */
+constexpr auto decompose_archive_version(uint32_t archive_version)
+        -> std::tuple<uint8_t, uint8_t, uint16_t> {
+    constexpr uint32_t cMajorVersionOffset{24U};
+    constexpr uint32_t cMinorVersionOffset{16U};
+    uint8_t const major_version{static_cast<uint8_t>(archive_version >> cMajorVersionOffset)};
+    uint8_t const minor_version{static_cast<uint8_t>(archive_version >> cMinorVersionOffset)};
+    uint16_t const patch_version{static_cast<uint16_t>(archive_version)};
```
LinZhihao-723:
```suggestion
    auto const major_version{static_cast<uint8_t>(archive_version >> cMajorVersionOffset)};
    auto const minor_version{static_cast<uint8_t>(archive_version >> cMinorVersionOffset)};
    auto const patch_version{static_cast<uint16_t>(archive_version)};
```
Since the rvalue is `static_cast`

components/core/src/clp_s/SingleFileArchiveDefs.hpp
```diff
@@ -3,15 +3,49 @@
...
 
 namespace clp_s {
+/**
+ * @param major_version
+ * @param minor_version
+ * @param patch_version
+ * @return The archive version, composed of a major, minor, and patch version.
+ */
+constexpr auto
+make_archive_version(uint8_t major_version, uint8_t minor_version, uint16_t patch_version)
```
LinZhihao-723, line 21: Can we move these two methods after type declarations?
Or it won't get compiled because `cArchiveVersion` is using `make_archive_version`?

components/core/src/clp_s/ArchiveWriter.cpp
```diff
@@ -239,8 +240,7 @@ void ArchiveWriter::write_archive_files(
 void ArchiveWriter::write_archive_header(FileWriter& archive_writer, size_t metadata_section_size) {
     ArchiveHeader header{
```
LinZhihao-723, line 241: Can we add a constructor for this struct?

LinZhihao-723 (changes requested):
Overall lgtm.

components/core/src/clp_s/SingleFileArchiveDefs.hpp
gibber9809: Yeah, wouldn't compile when I tried it unfortunately.

LinZhihao-723: kk, leave it then

components/core/src/clp_s/ArchiveWriter.cpp
gibber9809: I also added a default constructor that'll just zero-initialize the header so that the read-path code can keep working the same way.

PR #1767, opened 2025-12-10
feat(clp-package): Add support for ingesting from custom S3 endpoints (resolves #1758).

### LINKED ISSUE #1758: Add Support for Ingesting Logs from Custom S3 Endpoints
### Request

One of our users has requested the ability to ingest logs from S3-compatible storage services. However, the current implementation of the CLP package assumes that all S3 interactions are directed to AWS's default S3 endpoint (amazonaws.com).

This limitation restricts integration with other S3-compatible storage providers, which may use custom or non-AWS endpoints. It would be nice to add support for ingesting from custom S3 endpoints.

### Possible implementation

Currently, when the user calls `compress-from-s3.sh`, the following steps occur:

1. The URL supplied by the user is parsed and validated. If the endpoint isn't amazonaws.com, an error is raised.

2. The URL is converted into an `S3InputConfig` object, which stores the region, bucket, and key of the input.

3. `S3InputConfig` is used to fetch all keys for a key prefix, if the user requested a key-prefix ingestion.

3. In the compression worker, `S3InputConfig` is used to generate a list of virtual-hosted style URLs, which is then passed to clp-s.

4. clp-s deserializes the virtual-hosted URLs using the `S3Url` class.

5. clp-s signs the url based on the fields in the S3Url class.

6. `S3Url` and the signature is serialized into another virtual-hosted URL.

**Required changes**

To support custom S3-compatible endpoints, we propose the following changes:

1. Add a new endpoint_url field to both the `S3InputConfig` and `S3Url` classes.

2. Most S3-compatible storage services is only accessible using path-style URLs. Therefore, clp-s would have to send the request using path-style URLs. As a solution, when the `endpoint` isn't `amazonaws.com`, both clp-s and the compression worker switch to path-style URLs.

These changes will enable the integration of custom S3-compatible endpoints while maintaining compatibility with AWS's default S3 endpoint.

[Path style urls](https://docs.localstack.cloud/aws/services/s3/#path-style-and-virtual-hosted-style-requests)

hoophalab:
# Description

1. Add a new endpoint_url field to both the `S3InputConfig` and `S3Url` classes.
2. When the `endpoint` isn't `amazonaws.com`, both clp-s and the compression worker switch to path-style URLs.
3. Update the regex to allow path-style urls. (tbh, we should switch to a true url parser in the future.)

# Checklist

  breaking change.

# Validation performed

1. run minio/localstack and setup a authentication key
2. setup `log_input` section in `clp-config.yaml`
```
logs_input:
  type: "s3"
  aws_authentication:
    type: "credentials"
    credentials:
      access_key_id: "id"
      secret_access_key: "key"
```
4. `compress-from-s3.sh --timestamp-key timestamp --dataset default s3-key-prefix http://<endpoint>/test-bucket/log-file-name.log`

## Summary by CodeRabbit

* **New Features**
  * Support for custom S3 endpoints propagated across S3 clients and operations.
  * S3 URL generation now emits virtual‑hosted or path‑style links.

* **Improvements**
  * S3 config accepts optional endpoint and optional region for greater flexibility.
  * S3 parsing/generation is endpoint‑aware, enforces consistent endpoint/region/bucket, computes common key prefixes, and detects duplicate keys.
  * S3 deletion and upload flows respect configured endpoint.

* **Tests**
  * Added unit tests covering both S3 URL styles.

### DIFF
--- a/tools/deployment/package-helm/Chart.yaml
+++ b/tools/deployment/package-helm/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: "v2"
 name: "clp"
-version: "0.1.2-dev.3"
+version: "0.1.2-dev.4"
 description: "A Helm chart for CLP's (Compressed Log Processor) package deployment"
 type: "application"
 appVersion: "0.7.1-dev"
--- a/tools/deployment/package-helm/templates/_helpers.tpl
+++ b/tools/deployment/package-helm/templates/_helpers.tpl
@@ -98,8 +98,8 @@ failureThreshold: 3
 Creates a local PersistentVolume.
 
 @param {object} root Root template context
-@param {string} name PV name
-@param {string} component Component label
+@param {string} component_category (e.g., "database", "shared-data")
+@param {string} name (e.g., "archives", "data", "logs")
 @param {string} nodeRole Node role for affinity. Targets nodes with label
   "node-role.kubernetes.io/<nodeRole>". Always falls back to
   "node-role.kubernetes.io/control-plane"
@@ -112,10 +112,10 @@ Creates a local PersistentVolume.
 apiVersion: "v1"
 kind: "PersistentVolume"
 metadata:
-  name: {{ .name }}
+  name: {{ include "clp.fullname" .root }}-{{ .component_category }}-{{ .name }}
   labels:
     {{- include "clp.labels" .root | nindent 4 }}
-    app.kubernetes.io/component: {{ .component | quote }}
+    app.kubernetes.io/component: {{ .component_category | quote }}
 spec:
   capacity:
     storage: {{ .capacity }}
--- a/tools/deployment/package-helm/templates/database-data-pv.yaml
+++ b/tools/deployment/package-helm/templates/database-data-pv.yaml
@@ -1,7 +1,7 @@
 {{- include "clp.createLocalPv" (dict
   "root" .
-  "name" (printf "%s-database-data" (include "clp.fullname" .))
-  "component" "database"
+  "component_category" "database"
+  "name" "data"
   "nodeRole" "control-plane"
   "capacity" "20Gi"
   "accessModes" (list "ReadWriteOnce")
--- a/tools/deployment/package-helm/templates/database-logs-pv.yaml
+++ b/tools/deployment/package-helm/templates/database-logs-pv.yaml
@@ -1,7 +1,7 @@
 {{- include "clp.createLocalPv" (dict
   "root" .
-  "name" (printf "%s-database-logs" (include "clp.fullname" .))
-  "component" "database"
+  "component_category" "database"
+  "name" "logs"
   "nodeRole" "control-plane"
   "capacity" "5Gi"
   "accessModes" (list "ReadWriteOnce")
--- a/tools/deployment/package-helm/templates/database-statefulset.yaml
+++ b/tools/deployment/package-helm/templates/database-statefulset.yaml
@@ -53,14 +53,14 @@ spec:
             - name: "database"
               containerPort: 3306
           volumeMounts:
-            - name: "database-data"
-              mountPath: "/var/lib/mysql"
-            - name: "database-logs"
-              mountPath: "/var/log/mysql"
             - name: "config"
               mountPath: "/etc/mysql/conf.d/logging.cnf"
               subPath: "mysql-logging.cnf"
               readOnly: true
+            - name: "database-data"
+              mountPath: "/var/lib/mysql"
+            - name: "database-logs"
+              mountPath: "/var/log/mysql"
           readinessProbe:
             {{- include "clp.readinessProbeTimings" . | nindent 12 }}
             exec: &database-health-check
--- a/tools/deployment/package-helm/templates/queue-logs-pv.yaml
+++ b/tools/deployment/package-helm/templates/queue-logs-pv.yaml
@@ -1,7 +1,7 @@
 {{- include "clp.createLocalPv" (dict
   "root" .
-  "name" (printf "%s-queue-logs" (include "clp.fullname" .))
-  "component" "queue"
+  "component_category" "queue"
+  "name" "logs"
   "nodeRole" "control-plane"
   "capacity" "5Gi"
   "accessModes" (list "ReadWriteOnce")
--- a/tools/deployment/package-helm/templates/redis-data-pv.yaml
+++ b/tools/deployment/package-helm/templates/redis-data-pv.yaml
@@ -1,7 +1,7 @@
 {{- include "clp.createLocalPv" (dict
   "root" .
-  "name" (printf "%s-redis-data" (include "clp.fullname" .))
-  "component" "redis"
+  "component_category" "redis"
+  "name" "data"
   "nodeRole" "control-plane"
   "capacity" "20Gi"
   "accessModes" (list "ReadWriteOnce")
--- a/tools/deployment/package-helm/templates/redis-logs-pv.yaml
+++ b/tools/deployment/package-helm/templates/redis-logs-pv.yaml
@@ -1,7 +1,7 @@
 {{- include "clp.createLocalPv" (dict
   "root" .
-  "name" (printf "%s-redis-logs" (include "clp.fullname" .))
-  "component" "redis"
+  "component_category" "redis"
+  "name" "logs"
   "nodeRole" "control-plane"
   "capacity" "5Gi"
   "accessModes" (list "ReadWriteOnce")
--- a/tools/deployment/package-helm/templates/redis-statefulset.yaml
+++ b/tools/deployment/package-helm/templates/redis-statefulset.yaml
@@ -41,14 +41,14 @@ spec:
             - name: "redis"
               containerPort: 6379
           volumeMounts:
-            - name: "redis-data"
-              mountPath: "/data"
-            - name: "redis-logs"
-              mountPath: "/var/log/redis"
             - name: "config"
               mountPath: "/etc/redis/redis.conf"
               subPath: "redis.conf"
               readOnly: true
+            - name: "redis-data"
+              mountPath: "/data"
+            - name: "redis-logs"
+              mountPath: "/var/log/redis"
           readinessProbe:
             {{- include "clp.readinessProbeTimings" . | nindent 12 }}
             exec: &redis-health-check
--- a/tools/deployment/package-helm/templates/results-cache-data-pv.yaml
+++ b/tools/deployment/package-helm/templates/results-cache-data-pv.yaml
@@ -1,9 +1,9 @@
 {{- include "clp.createLocalPv" (dict
   "root" .
-  "name" (printf "%s-results-cache-data" (include "clp.fullname" .))
-  "component" "results-cache"
+  "component_category" "results-cache"
+  "name" "data"
   "nodeRole" "control-plane"
   "capacity" "20Gi"
   "accessModes" (list "ReadWriteOnce")
-  "hostPath" (printf "%s/results-cache" .Values.clpConfig.data_directory)
+  "hostPath" (printf "%s/results_cache" .Values.clpConfig.data_directory)
 ) }}
--- a/tools/deployment/package-helm/templates/results-cache-logs-pv.yaml
+++ b/tools/deployment/package-helm/templates/results-cache-logs-pv.yaml
@@ -1,9 +1,9 @@
 {{- include "clp.createLocalPv" (dict
   "root" .
-  "name" (printf "%s-results-cache-logs" (include "clp.fullname" .))
-  "component" "results-cache"
+  "component_category" "results-cache"
+  "name" "logs"
   "nodeRole" "control-plane"
   "capacity" "5Gi"
   "accessModes" (list "ReadWriteOnce")
-  "hostPath" (printf "%s/results-cache" .Values.clpConfig.logs_directory)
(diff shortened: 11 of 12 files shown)


components/core/src/clp/aws/AwsAuthenticationSigner.cpp
```diff
@@ -234,14 +245,31 @@ AwsAuthenticationSigner::generate_presigned_url(S3Url const& s3_url, string& pre
...
+                    s3_url.get_schema(),
+                    s3_url.get_host(),
+                    s3_url.get_bucket(),
+                    s3_url.get_key(),
+                    canonical_query_string,
+                    cXAmzSignature,
+                    signature_str
+            );
+            break;
+    }
```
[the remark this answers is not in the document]
hoophalab: this enum only has two values

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -578,7 +578,8 @@ def validate_authentication(cls, data):
 
 
 class S3Config(BaseModel):
-    region_code: NonEmptyStr
+    endpoint_url: str | None = None
+    region_code: NonEmptyStr | None = None
     bucket: NonEmptyStr
     key_prefix: str
     aws_authentication: AwsAuthentication
```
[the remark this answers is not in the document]
hoophalab: Path style urls don't necessarily have a region code. We need to support that.

components/core/src/clp/aws/AwsAuthenticationSigner.hpp
```diff
@@ -51,12 +59,16 @@ class S3Url {
 
     [[nodiscard]] auto get_host() const -> std::string_view { return m_host; }
 
+    [[nodiscard]] auto get_schema() const -> std::string_view { return m_schema; }
+
 private:
+    Style m_style;
```
gibber9809:
```suggestion
    Style m_style{Style::VirtualHost};
```
Should be default brace-initialized, per our style guide.

components/core/src/clp/aws/AwsAuthenticationSigner.hpp
```diff
@@ -51,12 +59,16 @@ class S3Url {
 
     [[nodiscard]] auto get_host() const -> std::string_view { return m_host; }
 
+    [[nodiscard]] auto get_schema() const -> std::string_view { return m_schema; }
```
gibber9809: As far as I can tell it's usually called the `scheme`. Could we change the naming here to `get_scheme()` and `m_scheme`, and update any other names like the scheme regex name throughout the code?

components/core/src/clp/aws/AwsAuthenticationSigner.cpp
```diff
@@ -163,23 +172,29 @@ auto get_canonical_request(S3Url const& url, string_view query_string) -> string
 
 S3Url::S3Url(string const& url) {
```
gibber9809, line 173: While we're changing the parsing here could we add some basic tests for `S3Url` to ensure that this code works correctly for the kinds of virtual-hosted and path-style URLs we expect to see?

A few basic cases where we can confirm that the various components of the URLs are extracted correctly + canonical requests match what we expect should be good.

Also, not sure if you're familiar with our new-ish standard for c++ tests, but in case you're not, they should go in a `tests` subdirectory of this `aws` directory.

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -202,55 +211,81 @@ def _create_s3_client(
...
-    endpoint = match.group("endpoint")
-    key_prefix = match.group("key_prefix")
 
-    if AWS_ENDPOINT != endpoint:
-        raise ValueError(f"Unsupported endpoint: {endpoint}")
+    endpoint_url = f"{schema}://{endpoint}" if endpoint != AWS_ENDPOINT else None
+    key_prefix = match.group("key_prefix")
 
-    return region_code, bucket_name, key_prefix
+    return endpoint_url, region_code, bucket_name, key_prefix
```
gibber9809, line 255: Rabbit seems to have a point here, though I haven't read the python side of this PR carefully enough to be sure.

components/core/src/clp/aws/AwsAuthenticationSigner.hpp
hoophalab: sorry xD

hoophalab:
> Rabbit seems to have a point here, though I haven't read the python side of this PR carefully enough to be sure.

I confirm rabbit is right. Let me make some changes.

components/core/src/clp/aws/constants.hpp
```diff
@@ -7,6 +7,13 @@ namespace clp::aws {
 // Endpoint
 constexpr std::string_view cAwsEndpoint{"amazonaws.com"};
 
+// Regex
+constexpr std::string_view cSchemeRegex{R"((?<scheme>(http|https)))"};
+constexpr std::string_view cEndpointRegex{R"((?<endpoint>[a-z0-9.-]+(\:[0-9]+)?))"};
+constexpr std::string_view cRegionRegex{R"((?<region>[a-z]+-[a-z]+-[0-9]))"};
+constexpr std::string_view cBucketRegex{R"((?<bucket>[a-z0-9.-]+))"};
+constexpr std::string_view cKeyRegex{R"((?<key>[^?]+))"};
+
```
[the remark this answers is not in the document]
hoophalab: as far as I know, current region code only contains 1 digit. This code is copied from the original regex

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -202,55 +212,72 @@ def _create_s3_client(
...
+    :return: A tuple of (endpoint_url, region_code, bucket, key_prefix).
     :raise: ValueError if `s3_url` is not a valid host-style URL or path-style URL.
     """
     host_style_url_regex = re.compile(
-        r"https://(?P<bucket_name>[a-z0-9.-]+)\.s3(\.(?P<region_code>[a-z]+-[a-z]+-[0-9]))"
-        r"\.(?P<endpoint>[a-z0-9.-]+)/(?P<key_prefix>[^?]+).*"
+        rf"{SCHEME_REGEXP}://{BUCKET_NAME_REGEXP}\."
+        rf"{S3_PREFIX_REGEXP}\.({REGION_CODE_REGEXP}\.)?"
+        rf"{ENDPOINT_REGEXP}/{KEY_PREFIX_REGEXP}.*"
     )
```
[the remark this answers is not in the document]
hoophalab: this is to keep consistent with other parts of the code.

components/core/src/clp/aws/test/test_AwsAuthenticationSigner.cpp
```diff
@@ -0,0 +1,34 @@
+#include <catch2/catch_message.hpp>
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/generators/catch_generators.hpp>
+
+#include "aws/AwsAuthenticationSigner.hpp"
+#include "aws/constants.hpp"
+
+namespace clp::aws::test {
+TEST_CASE("s3_url_parsing", "[aws][AwsAuthenticationSigner]") {
```
hoophalab, line 9: `generate_canonical_request` is a private method. We probably don't unittest private methods?

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -36,6 +36,13 @@
 
 S3_OBJECT_DELETION_BATCH_SIZE_MAX: Final[int] = 1000
 
+SCHEME_REGEXP = r"(?P<scheme>(http|https))"
+S3_PREFIX_REGEXP = r"(?P<s3>s3)"
+ENDPOINT_REGEXP = r"(?P<endpoint>[a-z0-9.-]+(\:[0-9]+)?)"
+REGION_CODE_REGEXP = r"(?P<region_code>[a-z]+-[a-z]+-[0-9])"
```
gibber9809: It seems like the region rabbit is complaining about is real, so maybe we could change this to something like `?P<region_code>[a-z0-9\-]+`?

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -202,55 +212,75 @@ def _create_s3_client(
...
 
-def parse_s3_url(s3_url: str) -> tuple[str, str, str]:
+def parse_s3_url(s3_url: str) -> tuple[str | None, str | None, str, str]:
     """
-    Parses the region_code, bucket, and key_prefix from the given S3 URL.
+    Parses the endpoint_url, region_code, bucket, and key_prefix from the given S3 URL.
     :param s3_url: A host-style URL or path-style URL.
-    :return: A tuple of (region_code, bucket, key_prefix).
+    :return: A tuple of (endpoint_url, region_code, bucket, key_prefix).
     :raise: ValueError if `s3_url` is not a valid host-style URL or path-style URL.
```
gibber9809, line 225: I feel like it's a bit weird that we parse the endpoint url for aws as `None`. I can see that we're consistently using a `None` endpoint to indicate aws everywhere, but maybe it would be simpler to have a small EndpointUrl class that stores the endpoint string + a flag that indicates whether the endpoint is S3 or not?

It's also potentially fine to stick with the `str` | `None` approach if you prefer, I just think we need to update these docstrings to explain that `None` means AWS in that case .

components/core/src/clp/aws/constants.hpp
```diff
@@ -7,6 +7,13 @@ namespace clp::aws {
 // Endpoint
 constexpr std::string_view cAwsEndpoint{"amazonaws.com"};
 
+// Regex
+constexpr std::string_view cSchemeRegex{R"((?<scheme>(http|https)))"};
+constexpr std::string_view cEndpointRegex{R"((?<endpoint>[a-z0-9.-]+(\:[0-9]+)?))"};
+constexpr std::string_view cRegionRegex{R"((?<region>[a-z]+-[a-z]+-[0-9]))"};
```
gibber9809: If we change the other region regex we should change this one too.

components/core/src/clp/aws/test/test_AwsAuthenticationSigner.cpp
gibber9809: Yeah, should be fine.

gibber9809 (changes requested):
A few nits, but otherwise looks good to me.

components/clp-py-utils/clp_py_utils/s3_utils.py
hoophalab: I've updated the documentation. The behavior feels analogous to when we don't specify an `AWS_ENDPOINT_URL` env in the AWS CLI, or when we pass `endpoint_url=None` to a `boto3.client(...)` call.

Honestly, I also feel returning a None for endpoint_url is very weird. Maybe the logic should be applied in `_create_s3_client`? Open to discussion.

gibber9809: I think it's probably fine to stick with `None` for now since it's at least consistent everywhere. If we find it ends up being too confusing we should definitely come back to this after the release though.

components/clp-py-utils/clp_py_utils/s3_utils.py
```diff
@@ -36,6 +36,13 @@
 
 S3_OBJECT_DELETION_BATCH_SIZE_MAX: Final[int] = 1000
 
+SCHEME_REGEXP = r"(?P<scheme>(http|https))"
+S3_PREFIX_REGEXP = r"(?P<s3>s3)"
+ENDPOINT_REGEXP = r"(?P<endpoint>[a-z0-9.-]+(\:[0-9]+)?)"
+REGION_CODE_REGEXP = r"(?P<region_code>[a-z0-9\-]+)"
+BUCKET_NAME_REGEXP = r"(?P<bucket_name>[a-z0-9.-]+)"
```
junhaoliao, line 43: it seems those patterns prohibit UPPER CASE letters, which are techically allowed as per RFC952? ditto for the ones in C++ code

https://datatracker.ietf.org/doc/html/rfc952
ASSUMPTIONS -> 1.
> A "name" (Net, Host, Gateway, or Domain name) is a text string up
>    to 24 characters drawn from the alphabet (A-Z), digits (0-9), minus
>    sign (-), and period (.).  Note that periods are only allowed when
>    they serve to delimit components of "domain style names". (See
>    [RFC-921](https://datatracker.ietf.org/doc/html/rfc921), "Domain Name System Implementation Schedule", for
>    background).  No blank or space characters are permitted as part of a
>    name. No distinction is made between upper and lower case.  The first
>    character must be an alpha character.  The last character must not be
>    a minus sign or period.  A host which serves as a GATEWAY should have
>    "-GATEWAY" or "-GW" as part of its name.  Hosts which do not serve as
>    Internet gateways should not use "-GATEWAY" and "-GW" as part of
>    their names. A host which is a TAC should have "-TAC" as the last
>    part of its host name, if it is a DoD host.  Single character names
>    or nicknames are not allowed.

junhaoliao (approved):
the code lgtm. posted a question

components/clp-py-utils/clp_py_utils/s3_utils.py
hoophalab: @coderabbitai Can you create an issue on this? 1. We need to verify if hostname is case insensitive and path is case sensitive. 2. `endpoint_url = f"{scheme}://{s3_prefix}{endpoint}" if endpoint != AWS_ENDPOINT else None` the comparison needs to be case insensitive.

PR #1769, opened 2025-12-11
feat(clp-rust-utils): Use `NonEmptyString` in S3-related configs to ensure validation at the type level.

LinZhihao-723:
# Description

Before this PR, we use raw strings in S3-related configs to represent:
- S3 bucket name
- S3 object key-prefix
- S3 object key
- Dataset name
- Timestamp key
- SQS URL to listen to for S3 ingestion
- Archive tags

These fields are supposed to be non-empty.

This PR migrates to [non_empty_string](https://docs.rs/non-empty-string/latest/non_empty_string/) to ensure validation at the type level:
- All fields mentioned above are guaranteed to be non-empty
- Empty fields (such as an empty dataset name) will be rejected when creating an ingestion job with the following error:
```
☁  clp [use-non-empty-string] ⚡  curl -X POST http://localhost:3002/s3_scanner \
-d '{"bucket_name": "lzh-test", "region": "us-east-2",  "dataset": ""}'
Failed to deserialize the JSON body into the target type: invalid value: string "", expected a string with a length of more than 0 at line 1 column 66
```

Notice that field validation is beyond just checking whether the string is empty. For example, the dataset name has the following requirements: https://github.com/y-scope/clp/blob/e6d7a58e1daea136476ef2f63f68caa821bbaeb6/components/clp-package-utils/clp_package_utils/general.py#L642-L644

These validations are beyond type checks. We should implement per-field validations in future PRs.

# Checklist

  breaking change.

# Validation performed

* [x] Ensure errors can be properly propagate to users with `NonEmptyString` (as shown in the description).

## Summary by CodeRabbit

* **New Features**
  * Added compression configuration options: target segment size and compression level.

* **Refactor**
  * Tightened public config types to require non-empty values for S3 buckets, keys, datasets and related fields, affecting configuration and serialization usage.
  * Restructured default dataset handling to use a lazily-initialized non-empty value.
  * Exposed a new public types module to support non-empty string helpers; tests updated accordingly.

### DIFF
--- a/components/clp-package-utils/clp_package_utils/general.py
+++ b/components/clp-package-utils/clp_package_utils/general.py
@@ -12,8 +12,8 @@ from enum import auto
 
 import yaml
 from clp_py_utils.clp_config import (
-    CLP_DEFAULT_CREDENTIALS_FILE_PATH,
     CLP_DEFAULT_CONFIG_FILE_RELATIVE_PATH,
+    CLP_DEFAULT_CREDENTIALS_FILE_PATH,
     CLP_SHARED_CONFIG_FILENAME,
     ClpConfig,
     CONTAINER_AWS_CONFIG_DIRECTORY,
--- a/components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
+++ b/components/clp-package-utils/clp_package_utils/scripts/native/decompress.py
@@ -13,8 +13,8 @@ from clp_py_utils.clp_config import (
     CLP_DB_USER_ENV_VAR_NAME,
     CLP_DEFAULT_CONFIG_FILE_RELATIVE_PATH,
     ClpConfig,
-    ClpDbUserType,
     ClpDbNameType,
+    ClpDbUserType,
     Database,
 )
 from clp_py_utils.clp_metadata_db_utils import get_files_table_name
--- a/components/clp-package-utils/clp_package_utils/scripts/start_clp.py
+++ b/components/clp-package-utils/clp_package_utils/scripts/start_clp.py
@@ -1,8 +1,11 @@
-import argparse
+#!/usr/bin/env python3
+"""Script to start the CLP Package."""
+
 import logging
 import pathlib
 import sys
 
+import click
 from clp_py_utils.clp_config import CLP_DEFAULT_CONFIG_FILE_RELATIVE_PATH
 from clp_py_utils.core import resolve_host_path_in_container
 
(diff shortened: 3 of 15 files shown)


components/clp-rust-utils/src/clp_config/package.rs
```diff
@@ -1,9 +1,25 @@
...
+///
+/// A [`NonEmptyString`] representing the default dataset name.
+///
+/// # Panics
+///
+/// Panics if the default dataset name (configured at compile time) is empty.
+#[must_use]
+pub fn default_dataset() -> NonEmptyString {
+    NonEmptyString::new(DEFAULT_DATASET_NAME.to_string())
+        .unwrap_or_else(|_| panic!("Default dataset name should be a valid NonEmptyString"))
```
hoophalab:
```suggestion
        .expect("Default dataset name constant is non-empty")
```

components/clp-rust-utils/src/clp_config/package.rs
```diff
@@ -1,9 +1,25 @@
...
+///
+/// # Returns
+///
+/// A [`NonEmptyString`] representing the default dataset name.
+///
+/// # Panics
+///
+/// Panics if the default dataset name (configured at compile time) is empty.
+#[must_use]
+pub fn default_dataset() -> NonEmptyString {
```
hoophalab: Not sure whether this is better/possible. This is probably the standard way to initialize something whose `new` is not a const expr.

```rust
static DEFAULT_DATASET: LazyLock<NonEmptyString> = LazyLock::new(|| {
    NonEmptyString::new("default".to_owned()).expect("message")
});

// on use
(*DEFAULT_DATASET).clone();
```

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -143,8 +144,8 @@ async fn upload_noise_objects(
 /// # Returns
 ///
 /// A unique testing prefix for S3 object keys. The prefix is formatted as `test-{job_id}/`.
-fn get_testing_prefix(job_id: &Uuid) -> String {
-    format!("test-{job_id}")
+fn get_testing_prefix(job_id: &Uuid) -> NonEmptyString {
+    NonEmptyString::new(format!("test-{job_id}")).unwrap()
 }
```
hoophalab:
```suggestion
fn get_testing_prefix(job_id: &Uuid) -> String {
    format!("test-{job_id}"))
}
```

Probably better to keep using `String` in all function prototypes, and restrict NonEmptyString in serde annotated structs. Otherwise, we will create two separate worlds in code (similar to sync vs async)

components/clp-rust-utils/tests/clp_config_test.rs
```diff
@@ -3,14 +3,15 @@ use clp_rust_utils::{
...
     serde::BrotliMsgpack,
 };
+use non_empty_string::NonEmptyString;
 use serde_json::Value;
 
 #[test]
 fn test_clp_io_config_serialization() {
     let s3_config = S3Config {
-        bucket: "yscope".into(),
+        bucket: NonEmptyString::new("yscope".into()).unwrap(),
```
hoophalab: Adding many `unwrap`'s in the code looks scary, but rust has a poor const support.
As a workaround, how about adding the following function to rust-utils:

```rust

trait FromStatic {
    fn from_static_unchecked(str: &'static str) -> NonEmptyString {
        NonEmptyString::new(str.to_owned()).expect("const string shouldn't be empty")
    }
}

impl FromStatic for NonEmptyString {}

fn main() {
    // let str = NonEmptyString::from_static_unchecked(""); // panics
    let str = NonEmptyString::from_static_unchecked("clp");
}
```

hoophalab (commented):
some comments

components/log-ingestor/tests/test_ingestion_job.rs
LinZhihao-723: I understand your concern. Here're my thoughts:
- In the actual code level, we shouldn't worry too much about whether the underlying type is a String or a NonEmptyString. Usually the code is operating on the struct level (e.g., the config, or the S3 object metadata). The underlying type won't explicitly exposed to helper functions in most cases. For example, in `log_ingestor_manager.rs`: The NonEmptyString is wrapped by `IngestionJobTableEntry`. When a function needs to use the actual string value, it's probably gonna take `&str` anyways like when checking prefix conflicts in `is_mutually_prefix_free`. So I think it should be fine if we don't actively using `NonEmptyString` outside of a struct.
- In this particular case: this function is in a test case, and it's been called multiple times to create testing prefix. It would be more verbose if we return a String and do NonEmptyString conversion outside. So how about rename this function to `get_testing_prefix_as_non_empty_string` to be explicit?

components/log-ingestor/Cargo.toml
```diff
@@ -23,6 +24,7 @@ tracing-appender = "0.2.2"
 tracing-subscriber = { version = "0.3.20", features = ["json", "env-filter", "fmt", "std"] }
 uuid = { version = "1.18.1", features = ["v4"] }
 thiserror = "2.0.17"
+log = "0.4.29"
```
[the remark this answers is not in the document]
LinZhihao-723: Can you create an issue to keep track of it?
It seems like this file also has ordering issue.

components/log-ingestor/src/ingestion_job/s3_scanner.rs
```diff
@@ -3,6 +3,7 @@ use std::time::Duration;
 use anyhow::Result;
 use aws_sdk_s3::Client;
 use clp_rust_utils::{job_config::ingestion::s3::S3ScannerConfig, s3::ObjectMetadata};
+use non_empty_string::NonEmptyString;
```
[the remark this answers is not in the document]
LinZhihao-723: The original error is a String. What I'm doing is correct.

PR #1787, opened 2025-12-16
feat(webui): Use DirectoryTree for file path selection (closes #1779).

### LINKED ISSUE #1779: webui: Full-width title click for tree node toggle in the file lister UI
### Request

> [!NOTE]
> This feature builds on top of the file listing UI added in #1292

Currently, users can only expand/collapse tree nodes by clicking the small +/- icon. The node title text itself is not clickable for toggling; instead, clicking toggles the node's selection state. This creates a smaller click target and requires more precision from users.

**Why it's important:**
- Increases clickable area for a more comfortable UX
- Reduces user frustration with small click targets
- Follows common file browser patterns (e.g., VS Code, macOS Finder)
- Improves accessibility for users with motor control difficulties

### Possible implementation

Add a custom `treeTitleRender` prop that wraps the title in a clickable span with full width:

```typescript
const handleTitleClick = useCallback((ev: React.MouseEvent, node: TreeNode) => {
    ev.stopPropagation();

    const isExpanded = expandedKeys.includes(node.value as string);
    if (isExpanded) {
        setExpandedKeys((prev) => prev.filter((k) => k !== node.value));
    } else {
        setExpandedKeys((prev) => [...prev, node.value as string]);
        // Trigger load if needed
        if (!loadedPaths.current.has(node.value as string) && !node.isLeaf) {
            handleLoadData(node).catch(console.error);
        }
    }
}, [expandedKeys, handleLoadData]);

const treeTitleRender = useCallback((node: TreeNode) => (
    <span
        style={{display: "flex"}}
        onClick={(e) => handleTitleClick(e, node)}
    >
        {node.title as string}
    </span>
), [handleTitleClick]);

// Add to TreeSelect props:
treeTitleRender={treeTitleRender}
```

#### Considerations
- Need to prevent event bubbling to avoid conflicts with TreeSelect's native behavior
- CSS styling required for proper full-width clickable area
- Should maintain consistency with existing expand/collapse behavior

junhaoliao:
# Description

Currently, tree node expansion/collapse in the file listing UI requires clicking the small +/- switcher icon. 

This PR replaces the TreeSelect's default tree rendering with Ant Design's `DirectoryTree` component
via `popupRender`, providing a more intuitive file browser experience with folder/file icons.

**Changes:**
- Add `DirectoryTreeSelect` component that wraps `TreeSelect` with a `DirectoryTree` popup
- Display folder icons (open/closed) for directories and file icons for files
- Extract `flatToHierarchy` utility to convert flat tree data to hierarchical format
- Extract `useListHeight` hook for responsive dropdown sizing
- Remove `SwitcherIcon` component (DirectoryTree has built-in icons)

# Checklist

  breaking change.

# Validation performed

Confirmed the above clicking on leaf / clicking on non-leaf node behaviours matched expectations.

## Summary by CodeRabbit

* **New Features**
  * Replaced legacy tree selector with a directory-style path selector supporting expandable folders, multi-select, clear/deselect and hierarchical display.

* **Behavior & Reliability**
  * Directory contents load on demand; loading failures now show user-facing error messages.
  * Simplified popup behavior by removing dynamic resizing for a more consistent fixed-height experience.

* **Chores**
  * Minor padding adjustment to the directory popup; legacy switcher icon removed.

---
  - https://app.asana.com/0/0/1212888465344304


junhaoliao:
fixing the function length violation

davemarco:
I think its okay for a small improvement. I think long term we should try to copy https://ant.design/components/tree#tree-demo-directory. the open/closed folders and files, I think make more intuitive.

components/webui/client/src/pages/IngestPage/Compress/PathsSelectFormItem/DirectoryTreeSelect.tsx
```diff
@@ -0,0 +1,152 @@
...
+interface DirectoryTreeSelectProps {
+    expandedKeys: string[];
+    treeData: TreeNode[];
+    onChange: (values: string[]) => void;
+    onDataLoad: (path: string) => Promise<void>;
+    onExpand: (keys: string[]) => void;
+}
+
+/**
+ * TreeSelect component that uses DirectoryTree for the popup.
```
davemarco: flow comment on one line.

davemarco (commented):
I initially expected this change to build on your previous update by adding icons, but using DirectoryTree directly works well too. I wonder if we could simplify things by using a regular Select as the base component. That way, the tree and its related logic could live in a dedicated component (e.g., DirectoryTreePopup, rendered as a popup), which might help keep the select and form components simpler and reduce some function duplication.

junhaoliao:
> I wonder if we could simplify things by using a regular Select as the base component.

makes a lot of sense. i'm making the modifications and testing now

davemarco (commented):
Does it make sense handleLoadData and similar functions directly into DirectoryTreeSelect instead of keeping them inside PathsSelectFormItem, which could make the tree component more self-contained and less passing functions. Also, I’m not entirely sure, could it work to get rid of simple mode layout entirely and storing the tree data directly in a hierarchical structure? This im less sure about

junhaoliao:
> Does it make sense handleLoadData and similar functions directly into DirectoryTreeSelect instead of keeping them inside PathsSelectFormItem, which could make the tree component more self-contained and less passing functions. 

The current structure is:
- PathsSelectFormItem: owns all state (treeData, expandedKeys, checkedKeys), API calls, deduplication (loadedPathsRef), form integration
- DirectoryTreeSelect: receives state/callbacks as props, converts flat→hierarchical, handles display logic
- DirectoryTreePopup: pure rendering

Moving it would mean DirectoryTreeSelect owns data loading, state, and deduplication?

While moving would encourage:
  - Less prop drilling (6 props currently passed)
  - More self-contained component

, we need to look at:
1. Form integration coupling - handleCheck calls form.setFieldValue("paths", keys). If DirectoryTreeSelect owned state, it would need the form instance passed in, or expose the checked state via a callback anyway
2. API dependency - loadPath calls listFiles(addServerPrefix(path)). Moving this into DirectoryTreeSelect ties a generic tree component to a specific API, reducing reusability
3. Deduplication ref - The loadedPathsRef needs to persist across renders. This is currently scoped to the form item lifecycle, which makes sense.
4. Current design follows "lifting state up" - The form item is the natural owner since it needs to sync with AntD Form

I believe the current separation is reasonable. The form item is the stateful controller, and DirectoryTreeSelect is a controlled presentation component.

---

> Also, I’m not entirely sure, could it work to get rid of simple mode layout entirely and storing the tree data directly in a hierarchical structure? This im less sure about

Storing directly in a hierarchical structure would work, but the conversion from flat -> hierarchical is cheap (linear scan, runs only when treeData changes). The real complexity is in state updates, and flat makes those trivial. Keeping flat storage is a reasonable choice for the lazy-loaded tree. The flatToHierarchy conversion is also memoized and only runs when data changes. The hierarchical structure would add complexity to addNodes and any future operations that modify the tree

davemarco:
for 1. I think you can do something like this maybe? https://ant.design/components/form#form-demo-customized-form-controls  -Like just match what the form expects, and you shouldnt need to set the value? 
2. we can just move the function to utils. 
3. The deduplication ref can also just be moved up one level. 
4. i dont disagree with that it does fit with "lift-up-state". but i think its better to only "lift-up-state" when really neccesary because its makes code harder to read moving across all the files. Can potentially also trigger more rerenders. 

I will respond to the second point soon

davemarco:
for second point 
> Storing directly in a hierarchical structure would work, but the conversion from flat -> hierarchical is cheap (linear scan, runs only when treeData changes). The real complexity is in state updates, and flat makes those trivial. Keeping flat storage is a reasonable choice for the lazy-loaded tree. The flatToHierarchy conversion is also memoized and only runs when data changes. The hierarchical structure would add complexity to addNodes and any future operations that modify the tree

Im noticing these new functions filterToParents/removeWithDescendants look complicated, and are already managed by treeSelect. Does it make sense to just go back to treeSelect where it has treeDataSimple, and  checkedStrategy which manages all this stuff for us. Then we can just icons to "fake" directory tree. Maybe eventually they will even add support. Anyways let me know ur thoughts.

PR #1788, opened 2025-12-16
feat(clp-s)!: Add `Timestamp` column type to replace`DateString` column type; Bump the archive version to `0.5.0`.

gibber9809:
# Description

This PR integrates the new `Timestamp` column type into the archive format, deprecating `DateString` columns. The new `Timestamp` column uses `clp_s::timestamp_parser` to represent string, float, and integer timestamp literals in the same column, storing timestamps as delta-encoded nanosecond-precision epoch timestamps.

This PR deprecates the following encoding features on the write-side, but retains support for reading older archives with this content:
1. `DateString` columns
2. `DoubleEpoch` encoded timestamp ranges

We remove most code related to the write-side for these encodings, including most of the functionality for parsing timestamps of unknown format using `clp_s::TimestampPattern`.

Most of the rest of the diff in this PR comprises:
1. Substantially rewriting `TimestampDictionaryReader/Writer` to work with `clp_s::timestamp_parser` and the new `Timestamp` column
2. Updating `JsonParser` to handle parsing timestamps as the new `Timestamp` type
3. Adding/updating tests for the new `Timestamp` type

This change is marked breaking for the following reasons:
1. It removes undefined behaviour in the timestamp dictionary that could previously occur when timestamps in an archive were not millisecond precision
2. It temporarily prevents us from ingesting some datasets that we could handle before when those datasets contain timezones that require the tzdb for correct interpretation (the open source postgresql dataset, for example). Support for parsing timezones using the tzdb is planned for `clp_s::timestamp_parser`, which will remove this limitation.
3. We remove the special-case indexing behaviour related to timestamps in the `indexer`, since the timestamp column in new archives should always be of type `Timestamp`

Benchmarking results on compatible open-source datasets:

| dataset | % improvement compression ratio | % improvement compression speed | % improvement decompression speed |
| --- | --- | --- | --- |
| cockroachdb | 0.29% | -9.58% | 0.15%|
| mongodb | 16.03% | 11.96% | -21.42% |
| elasticsearch | 4.64% | -1.03% | -16.16% |
| spark | -0.29% | 0.53%| 29.57% |
| geomean| 4.97% | 0.77% | -3.84% |

Generally, compression ratio improves (sometimes significantly), compression speed is largely unchanged, and decompression speed seems to have a strong dependence on dataset. I will investigate the datasets with slow decompression speed, but if performance work is required it will likely be part of a follow-up PR.

Note: compression speed results were very high variance, so I might re-gather them.

# Checklist

  breaking change.

# Validation performed

* Updated search tests to use `timestamp()` expression when querying timestamp column
* Added tests to validate end-to-end compression/decompression for `Timestamp` columns. 
* Manually validated that timeline renders correctly in webui, and timestamps associated with each record are correct.

## Summary by CodeRabbit

* **New Features**
  * Dedicated Timestamp type with DeprecatedDateString fallback; archives expose deprecated-format awareness.

* **Behaviour**
  * Unified timestamp parsing/ingestion, storage, search and JSON emission (timestamps as literal fields); search queries accept explicit timestamp literals.

* **Tests**
  * New end-to-end timestamp test, updated search tests to use timestamp literals, and added diverse timestamp test data.

* **Chores**
  * Archive format bumped to 0.5.0; build now requires the timestamp parsing component.

---
  - https://app.asana.com/0/0/1212888427978286

### DIFF
--- a/.github/workflows/clp-docs-generated-code-checks.yaml
+++ b/.github/workflows/clp-docs-generated-code-checks.yaml
@@ -21,7 +21,7 @@ jobs:
 
       - name: "Install task"
         shell: "bash"
-        run: "npm install -g @go-task/cli@3.44.0"
+        run: "npm install -g @go-task/cli@3.48.0"
 
       - name: "Generate OpenAPI docs"
         shell: "bash"
--- a/.github/workflows/clp-docs.yaml
+++ b/.github/workflows/clp-docs.yaml
@@ -34,7 +34,7 @@ jobs:
 
       - name: "Install task"
         shell: "bash"
-        run: "npm install -g @go-task/cli@3.44.0"
+        run: "npm install -g @go-task/cli@3.48.0"
 
       - if: "matrix.os == 'macos-15'"
         name: "Install coreutils (for md5sum)"
--- a/.github/workflows/clp-lint.yaml
+++ b/.github/workflows/clp-lint.yaml
@@ -33,7 +33,7 @@ jobs:
 
       - name: "Install task"
         shell: "bash"
-        run: "npm install -g @go-task/cli@3.44.0"
+        run: "npm install -g @go-task/cli@3.48.0"
 
       - name: "Install uv"
         shell: "bash"
--- a/.github/workflows/clp-package-helm.yaml
+++ b/.github/workflows/clp-package-helm.yaml
@@ -35,7 +35,7 @@ jobs:
 
       - name: "Install task"
         shell: "bash"
-        run: "npm install -g @go-task/cli@3.44.0"
+        run: "npm install -g @go-task/cli@3.48.0"
 
       - name: "Lint Helm charts"
         shell: "bash"
@@ -58,7 +58,7 @@ jobs:
 
       - name: "Install task"
         shell: "bash"
-        run: "npm install -g @go-task/cli@3.44.0"
+        run: "npm install -g @go-task/cli@3.48.0"
 
       - name: "Package Helm chart"
         shell: "bash"
--- a/.github/workflows/clp-rust-checks.yaml
+++ b/.github/workflows/clp-rust-checks.yaml
@@ -35,7 +35,7 @@ jobs:
 
       - name: "Install task"
         shell: "bash"
-        run: "npm install -g @go-task/cli@3.44.0"
+        run: "npm install -g @go-task/cli@3.48.0"
 
       - name: "Install the latest version of uv"
         uses: "astral-sh/setup-uv@85856786d1ce8acfbcc2f13a5f3fbd6b938f9f41"  # v7.1.2
--- a/.github/workflows/clp-s-generated-code-checks.yaml
+++ b/.github/workflows/clp-s-generated-code-checks.yaml
@@ -32,7 +32,7 @@ jobs:
 
       - name: "Install task"
         shell: "bash"
-        run: "npm install -g @go-task/cli@3.44.0"
+        run: "npm install -g @go-task/cli@3.48.0"
 
       - if: "matrix.os == 'macos-15'"
         name: "Install coreutils (for md5sum)"
--- a/.github/workflows/clp-uv-checks.yaml
+++ b/.github/workflows/clp-uv-checks.yaml
@@ -44,7 +44,7 @@ jobs:
 
       - name: "Install task"
         shell: "bash"
-        run: "npm install -g @go-task/cli@3.44.0"
+        run: "npm install -g @go-task/cli@3.48.0"
 
       - name: "Install the latest version of uv"
         uses: "astral-sh/setup-uv@85856786d1ce8acfbcc2f13a5f3fbd6b938f9f41"  # v7.1.2
--- a/components/core/tools/scripts/lib_install/pipx-packages/install-go-task.sh
+++ b/components/core/tools/scripts/lib_install/pipx-packages/install-go-task.sh
@@ -9,21 +9,27 @@ if ! command -v pipx >/dev/null 2>&1; then
     exit 1
 fi
 
-# We lock to version 3.44.0 to avoid https://github.com/y-scope/clp/issues/1352
-readonly required_version="3.44.0"
+readonly required_version_major_min=3
+readonly required_version_minor_min=48
+readonly required_version_min="${required_version_major_min}.${required_version_minor_min}"
 
 package_preinstalled=0
 if ! command -v task >/dev/null 2>&1; then
     package_preinstalled=1
-    pipx install --force "go-task-bin==${required_version}"
+    pipx install --force "go-task-bin>=${required_version_min}"
     pipx ensurepath
 fi
 
 script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
-task_version=$(task --silent --taskfile "${script_dir}/print-go-task-version.yaml")
-if [[ "${task_version}" != "${required_version}" ]]; then
-    echo "Error: Task version ${task_version} is currently unsupported (must be" \
-        "${required_version})."
+installed_version=$(task --silent --taskfile "${script_dir}/print-go-task-version.yaml")
+IFS=. read -r installed_version_major installed_version_minor _ <<<"${installed_version}"
+
+is_major_older=$(( installed_version_major < required_version_major_min ))
+is_minor_older=$(( installed_version_major == required_version_major_min
+    && installed_version_minor < required_version_minor_min ))
+if (( is_major_older || is_minor_older )); then
+    echo "Error: Task version ${installed_version} is unsupported (requires version" \
+        ">= ${required_version_min})."
 
     if ((0 == "${package_preinstalled}")); then
         echo "Please uninstall Task and then re-run the install script."
--- a/components/webui/package-lock.json
+++ b/components/webui/package-lock.json
@@ -2831,27 +2831,6 @@
         "url": "https://github.com/sponsors/nzakas"
       }
     },
-    "node_modules/@isaacs/balanced-match": {
-      "version": "4.0.1",
-      "resolved": "https://registry.npmjs.org/@isaacs/balanced-match/-/balanced-match-4.0.1.tgz",
-      "integrity": "sha512-yzMTt9lEb8Gv7zRioUilSglI0c0smZ9k5D65677DLWLtWJaXIS3CqcGyUFByYKlnUj6TkjLVs54fBl6+TiGQDQ==",
-      "license": "MIT",
-      "engines": {
-        "node": "20 || >=22"
-      }
-    },
-    "node_modules/@isaacs/brace-expansion": {
-      "version": "5.0.0",
-      "resolved": "https://registry.npmjs.org/@isaacs/brace-expansion/-/brace-expansion-5.0.0.tgz",
-      "integrity": "sha512-ZT55BDLV0yv0RBm2czMiZ+SqCGO7AvmOM3G/w2xhVPH+te0aKgFjmBvGlL1dH+ql2tgGO3MVrbb3jCKyvpgnxA==",
-      "license": "MIT",
-      "dependencies": {
-        "@isaacs/balanced-match": "^4.0.1"
-      },
-      "engines": {
-        "node": "20 || >=22"
-      }
-    },
     "node_modules/@isaacs/cliui": {
       "version": "8.0.2",
       "resolved": "https://registry.npmjs.org/@isaacs/cliui/-/cliui-8.0.2.tgz",
@@ -3362,9 +3341,9 @@
       }
     },
     "node_modules/@npmcli/agent/node_modules/lru-cache": {
-      "version": "11.2.4",
-      "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-11.2.4.tgz",
-      "integrity": "sha512-B5Y16Jr9LB9dHVkh6ZevG+vAbOsNOYCX+sXvFWFu7B3Iz5mijW3zdbMyhsh8ANd2mSWBYdJgnqi+mL7/LrOPYg==",
+      "version": "11.2.6",
+      "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-11.2.6.tgz",
+      "integrity": "sha512-ESL2CrkS/2wTPfuend7Zhkzo2u0daGJ/A2VucJOgQ/C48S/zB8MMeMHSGKYpXhIjbPxfuezITkaBH1wqv00DDQ==",
       "dev": true,
       "license": "BlueOak-1.0.0",
       "engines": {
@@ -3385,9 +3364,9 @@
       }
     },
     "node_modules/@npmcli/fs/node_modules/semver": {
-      "version": "7.7.3",
-      "resolved": "https://registry.npmjs.org/semver/-/semver-7.7.3.tgz",
-      "integrity": "sha512-SdsKMrI9TdgjdweUSR9MweHA4EJ8YxHn8DFaDisvhVlUOe4BF1tLD7GAj0lIqWVl+dPb/rExr0Btby5loQm20Q==",
+      "version": "7.7.4",
+      "resolved": "https://registry.npmjs.org/semver/-/semver-7.7.4.tgz",
+      "integrity": "sha512-vFKC2IEtQnVhpT78h1Yp8wzwrf8CM+MzKMHGJZfBtzhZNycRFnXsHk6E5TxIkkMsgNS7mdX3AGB7x2QM2di4lA==",
       "dev": true,
       "license": "ISC",
       "bin": {
@@ -3418,19 +3397,19 @@
       }
     },
     "node_modules/@npmcli/git/node_modules/isexe": {
-      "version": "3.1.1",
-      "resolved": "https://registry.npmjs.org/isexe/-/isexe-3.1.1.tgz",
-      "integrity": "sha512-LpB/54B+/2J5hqQ7imZHfdU31OlgQqx7ZicVlkm9kzg9/w8GKLEcFfJl/t7DCEDueOyBAD6zCCwTO6Fzs0NoEQ==",
+      "version": "4.0.0",
+      "resolved": "https://registry.npmjs.org/isexe/-/isexe-4.0.0.tgz",
+      "integrity": "sha512-FFUtZMpoZ8RqHS3XeXEmHWLA4thH+ZxCv2lOiPIn1Xc7CxrqhWzNSDzD+/chS/zbYezmiwWLdQC09JdQKmthOw==",
       "dev": true,
-      "license": "ISC",
+      "license": "BlueOak-1.0.0",
       "engines": {
-        "node": ">=16"
+        "node": ">=20"
       }
     },
     "node_modules/@npmcli/git/node_modules/lru-cache": {
-      "version": "11.2.4",
-      "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-11.2.4.tgz",
-      "integrity": "sha512-B5Y16Jr9LB9dHVkh6ZevG+vAbOsNOYCX+sXvFWFu7B3Iz5mijW3zdbMyhsh8ANd2mSWBYdJgnqi+mL7/LrOPYg==",
+      "version": "11.2.6",
+      "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-11.2.6.tgz",
+      "integrity": "sha512-ESL2CrkS/2wTPfuend7Zhkzo2u0daGJ/A2VucJOgQ/C48S/zB8MMeMHSGKYpXhIjbPxfuezITkaBH1wqv00DDQ==",
       "dev": true,
       "license": "BlueOak-1.0.0",
       "engines": {
@@ -3438,9 +3417,9 @@
       }
     },
     "node_modules/@npmcli/git/node_modules/semver": {
-      "version": "7.7.3",
-      "resolved": "https://registry.npmjs.org/semver/-/semver-7.7.3.tgz",
-      "integrity": "sha512-SdsKMrI9TdgjdweUSR9MweHA4EJ8YxHn8DFaDisvhVlUOe4BF1tLD7GAj0lIqWVl+dPb/rExr0Btby5loQm20Q==",
+      "version": "7.7.4",
+      "resolved": "https://registry.npmjs.org/semver/-/semver-7.7.4.tgz",
+      "integrity": "sha512-vFKC2IEtQnVhpT78h1Yp8wzwrf8CM+MzKMHGJZfBtzhZNycRFnXsHk6E5TxIkkMsgNS7mdX3AGB7x2QM2di4lA==",
       "dev": true,
       "license": "ISC",
       "bin": {
@@ -3451,13 +3430,13 @@
       }
     },
     "node_modules/@npmcli/git/node_modules/which": {
-      "version": "6.0.0",
-      "resolved": "https://registry.npmjs.org/which/-/which-6.0.0.tgz",
-      "integrity": "sha512-f+gEpIKMR9faW/JgAgPK1D7mekkFoqbmiwvNzuhsHetni20QSgzg9Vhn0g2JSJkkfehQnqdUAx7/e15qS1lPxg==",
+      "version": "6.0.1",
+      "resolved": "https://registry.npmjs.org/which/-/which-6.0.1.tgz",
+      "integrity": "sha512-oGLe46MIrCRqX7ytPUf66EAYvdeMIZYn3WaocqqKZAxrBpkqHfL/qvTyJ/bTk5+AqHCjXmrv3CEWgy368zhRUg==",
       "dev": true,
       "license": "ISC",
       "dependencies": {
-        "isexe": "^3.1.1"
+        "isexe": "^4.0.0"
       },
       "bin": {
         "node-which": "bin/which.js"
@@ -3523,9 +3502,9 @@
       }
     },
     "node_modules/@npmcli/package-json/node_modules/semver": {
-      "version": "7.7.3",
-      "resolved": "https://registry.npmjs.org/semver/-/semver-7.7.3.tgz",
-      "integrity": "sha512-SdsKMrI9TdgjdweUSR9MweHA4EJ8YxHn8DFaDisvhVlUOe4BF1tLD7GAj0lIqWVl+dPb/rExr0Btby5loQm20Q==",
+      "version": "7.7.4",
+      "resolved": "https://registry.npmjs.org/semver/-/semver-7.7.4.tgz",
+      "integrity": "sha512-vFKC2IEtQnVhpT78h1Yp8wzwrf8CM+MzKMHGJZfBtzhZNycRFnXsHk6E5TxIkkMsgNS7mdX3AGB7x2QM2di4lA==",
       "dev": true,
       "license": "ISC",
       "bin": {
@@ -3549,23 +3528,23 @@
       }
     },
     "node_modules/@npmcli/promise-spawn/node_modules/isexe": {
-      "version": "3.1.1",
-      "resolved": "https://registry.npmjs.org/isexe/-/isexe-3.1.1.tgz",
-      "integrity": "sha512-LpB/54B+/2J5hqQ7imZHfdU31OlgQqx7ZicVlkm9kzg9/w8GKLEcFfJl/t7DCEDueOyBAD6zCCwTO6Fzs0NoEQ==",
+      "version": "4.0.0",
+      "resolved": "https://registry.npmjs.org/isexe/-/isexe-4.0.0.tgz",
+      "integrity": "sha512-FFUtZMpoZ8RqHS3XeXEmHWLA4thH+ZxCv2lOiPIn1Xc7CxrqhWzNSDzD+/chS/zbYezmiwWLdQC09JdQKmthOw==",
       "dev": true,
-      "license": "ISC",
+      "license": "BlueOak-1.0.0",
       "engines": {
-        "node": ">=16"
+        "node": ">=20"
       }
     },
     "node_modules/@npmcli/promise-spawn/node_modules/which": {
-      "version": "6.0.0",
-      "resolved": "https://registry.npmjs.org/which/-/which-6.0.0.tgz",
-      "integrity": "sha512-f+gEpIKMR9faW/JgAgPK1D7mekkFoqbmiwvNzuhsHetni20QSgzg9Vhn0g2JSJkkfehQnqdUAx7/e15qS1lPxg==",
+      "version": "6.0.1",
+      "resolved": "https://registry.npmjs.org/which/-/which-6.0.1.tgz",
+      "integrity": "sha512-oGLe46MIrCRqX7ytPUf66EAYvdeMIZYn3WaocqqKZAxrBpkqHfL/qvTyJ/bTk5+AqHCjXmrv3CEWgy368zhRUg==",
       "dev": true,
       "license": "ISC",
       "dependencies": {
-        "isexe": "^3.1.1"
+        "isexe": "^4.0.0"
       },
       "bin": {
         "node-which": "bin/which.js"
@@ -3603,23 +3582,23 @@
       }
     },
     "node_modules/@npmcli/run-script/node_modules/isexe": {
-      "version": "3.1.1",
-      "resolved": "https://registry.npmjs.org/isexe/-/isexe-3.1.1.tgz",
-      "integrity": "sha512-LpB/54B+/2J5hqQ7imZHfdU31OlgQqx7ZicVlkm9kzg9/w8GKLEcFfJl/t7DCEDueOyBAD6zCCwTO6Fzs0NoEQ==",
+      "version": "4.0.0",
+      "resolved": "https://registry.npmjs.org/isexe/-/isexe-4.0.0.tgz",
+      "integrity": "sha512-FFUtZMpoZ8RqHS3XeXEmHWLA4thH+ZxCv2lOiPIn1Xc7CxrqhWzNSDzD+/chS/zbYezmiwWLdQC09JdQKmthOw==",
       "dev": true,
-      "license": "ISC",
+      "license": "BlueOak-1.0.0",
       "engines": {
-        "node": ">=16"
+        "node": ">=20"
       }
     },
     "node_modules/@npmcli/run-script/node_modules/which": {
-      "version": "6.0.0",
-      "resolved": "https://registry.npmjs.org/which/-/which-6.0.0.tgz",
-      "integrity": "sha512-f+gEpIKMR9faW/JgAgPK1D7mekkFoqbmiwvNzuhsHetni20QSgzg9Vhn0g2JSJkkfehQnqdUAx7/e15qS1lPxg==",
+      "version": "6.0.1",
+      "resolved": "https://registry.npmjs.org/which/-/which-6.0.1.tgz",
+      "integrity": "sha512-oGLe46MIrCRqX7ytPUf66EAYvdeMIZYn3WaocqqKZAxrBpkqHfL/qvTyJ/bTk5+AqHCjXmrv3CEWgy368zhRUg==",
       "dev": true,
       "license": "ISC",
       "dependencies": {
-        "isexe": "^3.1.1"
+        "isexe": "^4.0.0"
       },
       "bin": {
         "node-which": "bin/which.js"
@@ -5477,19 +5456,19 @@
       }
     },
     "node_modules/@tapjs/run/node_modules/isexe": {
-      "version": "3.1.1",
-      "resolved": "https://registry.npmjs.org/isexe/-/isexe-3.1.1.tgz",
-      "integrity": "sha512-LpB/54B+/2J5hqQ7imZHfdU31OlgQqx7ZicVlkm9kzg9/w8GKLEcFfJl/t7DCEDueOyBAD6zCCwTO6Fzs0NoEQ==",
+      "version": "3.1.5",
+      "resolved": "https://registry.npmjs.org/isexe/-/isexe-3.1.5.tgz",
+      "integrity": "sha512-6B3tLtFqtQS4ekarvLVMZ+X+VlvQekbe4taUkf/rhVO3d/h0M2rfARm/pXLcPEsjjMsFgrFgSrhQIxcSVrBz8w==",
       "dev": true,
-      "license": "ISC",
+      "license": "BlueOak-1.0.0",
       "engines": {
-        "node": ">=16"
+        "node": ">=18"
       }
     },
     "node_modules/@tapjs/run/node_modules/semver": {
-      "version": "7.7.3",
-      "resolved": "https://registry.npmjs.org/semver/-/semver-7.7.3.tgz",
-      "integrity": "sha512-SdsKMrI9TdgjdweUSR9MweHA4EJ8YxHn8DFaDisvhVlUOe4BF1tLD7GAj0lIqWVl+dPb/rExr0Btby5loQm20Q==",
+      "version": "7.7.4",
+      "resolved": "https://registry.npmjs.org/semver/-/semver-7.7.4.tgz",
+      "integrity": "sha512-vFKC2IEtQnVhpT78h1Yp8wzwrf8CM+MzKMHGJZfBtzhZNycRFnXsHk6E5TxIkkMsgNS7mdX3AGB7x2QM2di4lA==",
       "dev": true,
       "license": "ISC",
       "bin": {
@@ -5702,14 +5681,40 @@
         "node": "^20.17.0 || >=22.9.0"
       }
     },
+    "node_modules/@tufjs/models/node_modules/balanced-match": {
+      "version": "4.0.2",
+      "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-4.0.2.tgz",
+      "integrity": "sha512-x0K50QvKQ97fdEz2kPehIerj+YTeptKF9hyYkKf6egnwmMWAkADiO0QCzSp0R5xN8FTZgYaBfSaue46Ej62nMg==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "jackspeak": "^4.2.3"
+      },
+      "engines": {
+        "node": "20 || >=22"
+      }
+    },
+    "node_modules/@tufjs/models/node_modules/brace-expansion": {
+      "version": "5.0.2",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.2.tgz",
+      "integrity": "sha512-Pdk8c9poy+YhOgVWw1JNN22/HcivgKWwpxKq04M/jTmHyCZn12WPJebZxdjSa5TmBqISrUSgNYU3eRORljfCCw==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "balanced-match": "^4.0.2"
+      },
+      "engines": {
+        "node": "20 || >=22"
+      }
+    },
     "node_modules/@tufjs/models/node_modules/minimatch": {
-      "version": "10.1.1",
-      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-10.1.1.tgz",
-      "integrity": "sha512-enIvLvRAFZYXJzkCYG5RKmPfrFArdLv+R+lbQ53BmIMLIry74bjKzX6iHAm8WYamJkhSSEabrWN5D97XnKObjQ==",
+      "version": "10.2.0",
+      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-10.2.0.tgz",
+      "integrity": "sha512-ugkC31VaVg9cF0DFVoADH12k6061zNZkZON+aX8AWsR9GhPcErkcMBceb6znR8wLERM2AkkOxy2nWRLpT9Jq5w==",
       "dev": true,
       "license": "BlueOak-1.0.0",
       "dependencies": {
-        "@isaacs/brace-expansion": "^5.0.0"
+        "brace-expansion": "^5.0.2"
       },
       "engines": {
         "node": "20 || >=22"
@@ -5894,9 +5899,9 @@
       }
     },
     "node_modules/@types/unist": {
-      "version": "3.0.3",
-      "resolved": "https://registry.npmjs.org/@types/unist/-/unist-3.0.3.tgz",
-      "integrity": "sha512-ko/gIFJRv177XgZsZcBwnqJN5x/Gien8qNOn0D5bQU/zAzVf9Zt3BlcUiLqhV9y4ARk0GbT3tnUiPNgnTXzc/Q==",
+      "version": "2.0.11",
+      "resolved": "https://registry.npmjs.org/@types/unist/-/unist-2.0.11.tgz",
+      "integrity": "sha512-CmBKiL6NNo/OqgmMn95Fk9Whlp2mtvIv+KNpQKN2F4SjvrEesubTRWGYSg+BnWZOnlCaSTU1sMpsBOzgbYhnsA==",
       "license": "MIT"
     },
     "node_modules/@types/webidl-conversions": {
@@ -6140,9 +6145,9 @@
       }
     },
     "node_modules/@typescript-eslint/typescript-estree/node_modules/semver": {
-      "version": "7.7.3",
-      "resolved": "https://registry.npmjs.org/semver/-/semver-7.7.3.tgz",
-      "integrity": "sha512-SdsKMrI9TdgjdweUSR9MweHA4EJ8YxHn8DFaDisvhVlUOe4BF1tLD7GAj0lIqWVl+dPb/rExr0Btby5loQm20Q==",
+      "version": "7.7.4",
+      "resolved": "https://registry.npmjs.org/semver/-/semver-7.7.4.tgz",
+      "integrity": "sha512-vFKC2IEtQnVhpT78h1Yp8wzwrf8CM+MzKMHGJZfBtzhZNycRFnXsHk6E5TxIkkMsgNS7mdX3AGB7x2QM2di4lA==",
       "dev": true,
       "license": "ISC",
       "peer": true,
@@ -6609,9 +6614,9 @@
       }
     },
     "node_modules/ajv": {
-      "version": "8.17.1",
-      "resolved": "https://registry.npmjs.org/ajv/-/ajv-8.17.1.tgz",
-      "integrity": "sha512-B/gBuNg5SiMTrPkC+A2+cW0RszwxYmn6VYxB/inlBStS5nx6xHIt/ehKRhIMhqusl7a8LjQoZnjCs5vhwxOQ1g==",
+      "version": "8.18.0",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-8.18.0.tgz",
+      "integrity": "sha512-PlXPeEWMXMZ7sPYOHqmDyCJzcfNrUr3fGNKtezX14ykXOEIvyK81d+qydx89KY5O71FKMPaQ2vBfBFI5NHR63A==",
       "license": "MIT",
       "dependencies": {
         "fast-deep-equal": "^3.1.3",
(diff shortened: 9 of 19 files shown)


gibber9809:
Besides the one presto webui change that rabbit pointed out, we likely also need to change how the webui interprets timestamps in search results -- I'll double check and change that if necessary.

gibber9809:
> Besides the one presto webui change that rabbit pointed out, we likely also need to change how the webui interprets timestamps in search results -- I'll double check and change that if necessary.

Confirmed that this is the case, and also that the timeline reduction runs into a similar issue since it currently doesn't account for the fact that bucket sizes are in milliseconds, but timestamps are in nanoseconds.

Since these are both related to the fact that the timestamp we associate with each search result is now nanosecond precision instead of millisecond precision we can either change that back to milliseconds, or change all of the other code to expect milliseconds.

components/core/src/clp_s/TimestampDictionaryWriter.hpp
```diff
@@ -25,7 +27,15 @@ class TimestampDictionaryWriter {
...
-    TimestampDictionaryWriter() {}
+    explicit TimestampDictionaryWriter() {
+        auto quoted_patterns_result{timestamp_parser::get_all_default_quoted_timestamp_patterns()};
+        auto numeric_patterns_result{timestamp_parser::get_default_numeric_timestamp_patterns()};
+        if (quoted_patterns_result.has_error() || numeric_patterns_result.has_error()) {
+            throw OperationFailed(ErrorCode::ErrorCodeFailure, __FILENAME__, __LINE__);
+        }
+        m_quoted_timestamp_patterns = std::move(quoted_patterns_result.value());
+        m_numeric_timestamp_patterns = std::move(numeric_patterns_result.value());
+    }
```
LinZhihao-723, line 38: It's not the emphasis of this PR, but wanna bring it out since we touch this code: I think we should highly prefer to use a factory function in this case for the following two reasons:
* It's easier for the error propagation of the result: the caller is always free to throw it as an exception, but the initialization itself shouldn't throw.
* Technically, the lifetime of the members starts before the constructor body's execution. It's preferred to initialize them in the initialization list.

components/core/src/clp_s/TimestampDictionaryWriter.hpp
```diff
@@ -87,29 +110,18 @@ class TimestampDictionaryWriter {
...
-     * @param ranges
-     * @param compressor
-     */
-    static void write_timestamp_entries(
-            std::map<std::string, TimestampEntry> const& ranges,
-            std::stringstream& stream
-    );
-
-    using pattern_to_id_t = std::unordered_map<TimestampPattern const*, uint64_t>;
+    using pattern_to_id_t = absl::flat_hash_map<std::string, uint64_t>;
```
LinZhihao-723: - According to the standard, I think we should rename it to `PatternToIdMap`.
- In fact, seems like it's no used

components/core/src/clp_s/TimestampDictionaryWriter.hpp
```diff
@@ -87,29 +110,18 @@ class TimestampDictionaryWriter {
...
 
     // Variables
-    pattern_to_id_t m_pattern_to_id;
+    std::vector<std::pair<timestamp_parser::TimestampPattern, uint64_t>> m_string_patterns_and_ids;
```
LinZhihao-723: Nit: `m_string_patterns_and_ids` -> `m_string_pattern_and_id_pairs`?

components/core/src/clp_s/TimestampEntry.cpp
```diff
@@ -9,54 +9,26 @@
...
+    };
+    auto const millisecond_timestamp_lower_bound{
+            whole_milliseconds_in_timestamp - (remainder_nanoseconds_in_timestamp < 0 ? 1 : 0)
+    };
+    if (millisecond_timestamp_lower_bound < m_epoch_start) {
+        m_epoch_start = millisecond_timestamp_lower_bound;
+    }
+    if (millisecond_timestamp_upper_bound > m_epoch_end) {
+        m_epoch_end = millisecond_timestamp_upper_bound;
     }
```
LinZhihao-723:
```suggestion
    m_epoch_start = std::min(m_epoch_start, millisecond_timestamp_lower_bound);
    m_epoch_end = std::max(m_epoch_end, millisecond_timestamp_upper_bound);
```

components/core/src/clp_s/TimestampEntry.cpp
```diff
@@ -9,54 +9,26 @@
...
+void TimestampEntry::ingest_timestamp(epochtime_t timestamp) {
+    m_encoding = Epoch;
+    auto const whole_milliseconds_in_timestamp{timestamp / cNanosecondsInMillisecond};
+    auto const remainder_nanoseconds_in_timestamp{timestamp % cNanosecondsInMillisecond};
+    auto const millisecond_timestamp_upper_bound{
+            whole_milliseconds_in_timestamp + (remainder_nanoseconds_in_timestamp > 0 ? 1 : 0)
+    };
+    auto const millisecond_timestamp_lower_bound{
+            whole_milliseconds_in_timestamp - (remainder_nanoseconds_in_timestamp < 0 ? 1 : 0)
+    };
```
LinZhihao-723, line 26: Just to make sure my understanding is correct: This is to handle a negative epoch, right?

components/core/src/clp_s/TimestampEntry.cpp
```diff
@@ -9,54 +9,26 @@
...
-    if (timestamp < m_epoch_start) {
-        m_epoch_start = timestamp;
-    }
-    if (timestamp > m_epoch_end) {
-        m_epoch_end = timestamp;
-    }
-}
+namespace {
+// Constants
+constexpr epochtime_t cNanosecondsInMillisecond{1'000'000};
```
LinZhihao-723, line 14: Not necessarily for this PR: but the current way of mixing nanosec vs. millisec with the same type `epochtime_t` is confusing. Are we planning to make different types for these two precisions?

components/core/src/clp_s/TimestampDictionaryWriter.cpp
```diff
@@ -2,133 +2,188 @@
...
-) {
-    epochtime_t ret;
-    size_t timestamp_begin_pos = 0, timestamp_end_pos = 0;
-    TimestampPattern const* pattern{nullptr};
+        bool is_json_literal
+) -> std::pair<epochtime_t, uint64_t> {
+    auto timestamp_entry_it = m_column_id_to_range.find(node_id);
+    if (m_column_id_to_range.end() == timestamp_entry_it) {
+        timestamp_entry_it = m_column_id_to_range.emplace(node_id, key).first;
+    }
```
LinZhihao-723: We can simplify it using `try_emplace`:
```suggestion
    auto const timestamp_entry_it = m_column_id_to_range.try_emplace(node_id, key).first;
```
Or, even a bit more readable
```suggestion
    auto& [_, timestamp_entry]= *m_column_id_to_range.try_emplace(node_id, key).first;
```

components/core/src/clp_s/TimestampDictionaryWriter.cpp
```diff
@@ -2,133 +2,188 @@
...
 
     // Try parsing the timestamp as one of the previously seen timestamp patterns
-    for (auto it : m_pattern_to_id) {
-        if (it.first->parse_timestamp(timestamp, ret, timestamp_begin_pos, timestamp_end_pos)) {
-            pattern = it.first;
-            pattern_id = it.second;
-            break;
+    std::string generated_pattern;
```
LinZhihao-723: Shall we consider to make it a class member?

components/core/src/clp_s/TimestampDictionaryWriter.cpp
```diff
@@ -2,133 +2,188 @@
...
+) -> std::pair<epochtime_t, uint64_t> {
+    auto timestamp_entry_it = m_column_id_to_range.find(node_id);
+    if (m_column_id_to_range.end() == timestamp_entry_it) {
+        timestamp_entry_it = m_column_id_to_range.emplace(node_id, key).first;
     }
 
-    return ret;
-}
+    std::string generated_pattern;
+    for (auto const& [raw_pattern, pattern_and_id] : m_numeric_pattern_to_id) {
```
LinZhihao-723, line 105:
```suggestion
    for (auto const& [raw_pattern, pattern_and_id] : m_numeric_pattern_to_id) {
            auto const& [pattern, id] = pattern_and_id;
```
Nit: I think another layer of struct binding is preferred for better readability.

components/core/src/clp_s/TimestampDictionaryWriter.cpp
```diff
@@ -2,133 +2,188 @@
...
+    }
+
+    auto const parsing_result{timestamp_parser::search_known_timestamp_patterns(
+            timestamp,
+            m_numeric_timestamp_patterns,
+            true,
+            generated_pattern
+    )};
+    if (false == parsing_result.has_value()) {
+        SPDLOG_ERROR("Failed to parse timestamp {} against known timestamp patterns.", timestamp);
```
LinZhihao-723:
```suggestion
        auto const& error{quoted_pattern_result.error()};
        SPDLOG_ERROR(
                "Failed to parse timestamp `{}`: {} - {}",
                timestamp,
                error.category().name(),
                error.message()
        );
```
Might be worth to also log the error details.

components/core/src/clp_s/TimestampDictionaryWriter.cpp
```diff
@@ -2,133 +2,188 @@
...
+                          )
+                          .first;
     }
+    return {epoch_timestamp, pattern_it->second.second};
 }
 
 epochtime_t TimestampDictionaryWriter::get_begin_timestamp() const {
-    auto it = m_column_key_to_range.begin();
-    if (m_column_key_to_range.end() == it) {
+    auto it = m_column_id_to_range.begin();
```
LinZhihao-723, line 190: Might be unrelated to this PR, but why we only check the first one? Is it because clp-s only accepts a single timestamp key and it is expected that this map only contains one pair?

components/core/src/clp_s/TimestampDictionaryWriter.cpp
```diff
@@ -2,133 +2,188 @@
...
+#include <clp_s/timestamp_parser/TimestampParser.hpp>
 
+namespace clp_s {
 void TimestampDictionaryWriter::write(std::stringstream& stream) {
-    merge_range();
-    write_timestamp_entries(m_column_key_to_range, stream);
+    write_numeric_value<uint64_t>(stream, m_column_id_to_range.size());
+    for (auto const& range : m_column_id_to_range) {
+        range.second.write_to_stream(stream);
+    }
```
LinZhihao-723:
```suggestion
    for (auto const& [id, range] : m_column_id_to_range) {
        range.write_to_stream(stream);
    }
```

LinZhihao-723 (changes requested):
Will break the review into multiple parts (as discussed offline).
Will focus on the compression end first. This batch contains a completed review of  `TimestampDictionaryWriter`, `TimestampEntry`, and `TimestampPattern`.

components/core/src/clp_s/TimestampEntry.cpp
gibber9809: Yes, that is correct.

components/core/src/clp_s/TimestampDictionaryWriter.hpp
gibber9809: Yeah I think I just forgot to get rid of it -- will remove.

components/core/src/clp_s/TimestampDictionaryWriter.hpp
gibber9809: Yeah, that name is a bit better.

components/core/src/clp_s/TimestampDictionaryWriter.cpp
gibber9809: Yeah... this is a bit of a historical holdover. Technically the format allows multiple timestamps in this range index, and the first timestamp/range is the "authoritative" one. In practice though, every open source version of clp-s has only allowed users to specify one timestamp column, so we can rely on this map having at most one element. The research version of clp-s allowed multiple timestamp columns.

Honestly, if we start allowing users to specify multiple timestamp columns + track the ranges of all of these columns I would probably avoid pushing those ranges into this timestamp dictionary anyway, and instead build a separate more generic range index that can be used for any numeric column.

components/core/src/clp_s/TimestampEntry.cpp
gibber9809: That's probably a good idea, I could create an issue to track?

components/core/src/clp_s/TimestampDictionaryWriter.cpp
LinZhihao-723: Sounds good. Let's keep this untouched for now to avoid regression (even though it looks like we should be able to use a single range object, lol)

components/core/src/clp_s/TimestampDictionaryWriter.hpp
gibber9809: Yeah I agree -- re-requesting review now, but will get around to moving this stuff into a factory method later today.

components/core/src/clp_s/TimestampEntry.cpp
LinZhihao-723: Yeah sure, u can ask code rabbit to create one (but it might not be able to explain the context clearly, lol)

components/core/src/clp_s/TimestampDictionaryWriter.cpp
```diff
@@ -2,133 +2,178 @@
...
-        TimestampEntry new_entry(key);
-        new_entry.ingest_timestamp(ret);
-        m_column_id_to_range.emplace(node_id, std::move(new_entry));
-    } else {
-        entry->second.ingest_timestamp(ret);
+    auto const [epoch_timestamp, pattern] = parsing_result.value();
+    timestamp_entry.ingest_timestamp(epoch_timestamp);
+    auto const quoted_pattern_result{timestamp_parser::TimestampPattern::create(pattern)};
+    if (quoted_pattern_result.has_error()) {
+        SPDLOG_ERROR("Failed to create timestamp pattern.");
```
LinZhihao-723:
```suggestion
        auto const error{quoted_pattern_result.error()};
        SPDLOG_ERROR(
                "Failed to create timestamp pattern: {} - {}.",
                error.category().name(),
                error.message()
        );
```
Sorry, I just realized I commented in the wrong place. Let's add context of the error in the log.

components/core/src/clp_s/TimestampDictionaryWriter.cpp
```diff
@@ -2,133 +2,178 @@
...
+    if (false == optional_parsed_timestamp.has_value()) {
+        SPDLOG_ERROR("Failed to parse timestamp `{}` against known timestamp patterns.", timestamp);
+        throw OperationFailed(ErrorCodeFailure, __FILE__, __LINE__);
     }
+
+    auto const [epoch_timestamp, pattern] = optional_parsed_timestamp.value();
+    timestamp_entry.ingest_timestamp(epoch_timestamp);
+    auto const pattern_result{timestamp_parser::TimestampPattern::create(pattern)};
+    if (pattern_result.has_error()) {
+        SPDLOG_ERROR("Failed to create timestamp pattern.");
```
LinZhihao-723:
```suggestion
        auto const error{pattern_result.error()};
        SPDLOG_ERROR(
                "Failed to create timestamp pattern: {} - {}.",
                error.category().name(),
                error.message()
        );
```

components/core/src/clp_s/TimestampDictionaryWriter.cpp
```diff
@@ -2,133 +2,178 @@
...
+    auto const [factor, precision] = timestamp_parser::estimate_timestamp_precision(timestamp);
+    auto const epoch_timestamp{timestamp * factor};
+    timestamp_entry.ingest_timestamp(epoch_timestamp);
+    auto pattern{fmt::format("\\{}", precision)};
+
+    auto pattern_it{m_numeric_pattern_to_id.find(pattern)};
+    if (m_numeric_pattern_to_id.end() == pattern_it) {
+        auto pattern_result{timestamp_parser::TimestampPattern::create(pattern)};
+        if (pattern_result.has_error()) {
+            SPDLOG_ERROR("Failed to create timestamp pattern.");
```
LinZhihao-723:
```suggestion
            auto const error{pattern_result.error()};
            SPDLOG_ERROR(
                    "Failed to create timestamp pattern: {} - {}.",
                    error.category().name(),
                    error.message()
            );
```

components/core/src/clp_s/TimestampDictionaryWriter.hpp
LinZhihao-723: Comment as a marker for myself in case I forgot

components/core/src/clp_s/ArchiveWriter.hpp
```diff
@@ -192,34 +192,54 @@ class ArchiveWriter {
...
+     * Ingests a timestamp entry from a string.
      * @param key
      * @param node_id
      * @param timestamp
-     * @param pattern_id
-     * @return the epoch time corresponding to the string timestamp
+     * @param is_json_literal
+     * @return A pair containing:
+     * - The timestamp in epoch nanoseconds.
+     * - The pattern ID corresponding to the timestamp format.
```
LinZhihao-723: How about:
```suggestion
     * @return Forwards `TimestampDictionaryWriter::ingest_string_timestamp`'s return values.
```
Same for the other two mirrored methods.

components/core/src/clp_s/ArchiveWriter.cpp
```diff
@@ -330,12 +330,15 @@ void ArchiveWriter::initialize_schema_writer(SchemaWriter* writer, Schema const&
             case NodeType::UnstructuredArray:
                 writer->append_column(new ClpStringColumnWriter(id, m_var_dict, m_array_dict));
                 break;
-            case NodeType::DateString:
-                writer->append_column(new DateStringColumnWriter(id));
+            case NodeType::DeprecatedDateString:
+                throw OperationFailed(ErrorCodeUnsupported, __FILENAME__, __LINE__);
                 break;
```
LinZhihao-723, line 335: Drop this line?

components/core/src/clp_s/ArchiveWriter.cpp
```diff
@@ -330,12 +330,15 @@ void ArchiveWriter::initialize_schema_writer(SchemaWriter* writer, Schema const&
...
             case NodeType::DeltaInteger:
                 writer->append_column(new DeltaEncodedInt64ColumnWriter(id));
                 break;
+            case NodeType::Timestamp:
+                writer->append_column(new TimestampColumnWriter(id));
```
LinZhihao-723: gg, didin't realize we have this raw use of `new` in our codebase. We should probably refactor it (in a separate PR) before we move on and forget...

components/core/src/clp_s/TimestampEntry.cpp
gibber9809: @coderabbitai create an issue to track using a different 64 bit integer type for millisecond and nanosecond precision timestamps in clp-s to increase code clarity.

components/core/src/clp_s/ArchiveWriter.cpp
gibber9809: Sure, I was trying to be defensive of some future situation where we let users access ArchiveWriter programmatically, but not really needed for now + probably not the best way to prevent users from trying to create deprecated column types.

components/core/src/clp_s/ArchiveWriter.cpp
LinZhihao-723: @coderabbitai Can u create an issue to track this?

components/core/src/clp_s/TimestampDictionaryReader.hpp
```diff
@@ -30,11 +36,13 @@ class TimestampDictionaryReader {
...
 
     /**
-     * Gets the string encoding for a given epoch and format ID
+     * Gets the string encoding for a given epoch and format ID by interpreting the pattern
+     * identified by `format_id` as a `clp_s::TimestampPattern`.
      * @param epoch
      * @param format_id
      */
-    std::string get_string_encoding(epochtime_t epoch, uint64_t format_id) const;
+    auto get_deprecated_timestamp_string_encoding(epochtime_t epoch, uint64_t format_id) const
```
LinZhihao-723:
```suggestion
    [[nodiscard]] auto
    get_deprecated_timestamp_string_encoding(epochtime_t epoch, uint64_t format_id) const
```

components/core/src/clp_s/TimestampDictionaryReader.hpp
```diff
@@ -30,11 +36,13 @@ class TimestampDictionaryReader {
     ErrorCode read(ZstdDecompressor& decompressor);
 
     /**
-     * Gets the string encoding for a given epoch and format ID
+     * Gets the string encoding for a given epoch and format ID by interpreting the pattern
+     * identified by `format_id` as a `clp_s::TimestampPattern`.
      * @param epoch
      * @param format_id
```
LinZhihao-723, line 44:
```suggestion
     * @param format_id
     * @return The string encoding for the given epoch and format ID.
```

components/core/src/clp_s/TimestampDictionaryReader.cpp
```diff
@@ -64,16 +71,42 @@ ErrorCode TimestampDictionaryReader::read(ZstdDecompressor& decompressor) {
...
 
     return ret;
 }
+
+void TimestampDictionaryReader::append_timestamp_to_buffer(
+        epochtime_t timestamp,
+        uint64_t format_id,
+        std::string& buffer
+) const {
+    auto const& pattern{m_timestamp_patterns.at(format_id)};
```
LinZhihao-723, line 117: `pattern` -> `optional_pattern`?

components/core/src/clp_s/TimestampDictionaryReader.cpp
```diff
@@ -64,16 +71,42 @@ ErrorCode TimestampDictionaryReader::read(ZstdDecompressor& decompressor) {
...
+    if (false == pattern.has_value()) {
+        throw OperationFailed(ErrorCodeCorrupt, __FILENAME__, __LINE__);
```
LinZhihao-723: Shall we add log?

components/core/src/clp_s/TimestampDictionaryReader.cpp
```diff
@@ -64,16 +71,42 @@ ErrorCode TimestampDictionaryReader::read(ZstdDecompressor& decompressor) {
...
             return error;
         }
-        m_patterns[id] = TimestampPattern(0, pattern);
+        m_deprecated_patterns.emplace(id, TimestampPattern(0, pattern));
+        auto const timestamp_pattern_result{timestamp_parser::TimestampPattern::create(pattern)};
+        if (timestamp_pattern_result.has_error()) {
+            m_timestamp_patterns.emplace(id, std::nullopt);
+        } else {
+            m_timestamp_patterns.emplace(id, std::move(timestamp_pattern_result.value()));
+        }
```
LinZhihao-723: iiuc, this is to provide backward compatibility. We load both `m_timestamp_pattern` and `m_deprecated_patterns` to be ready to serve both old archives and new archives.

components/core/src/clp_s/TimestampDictionaryReader.cpp
```diff
@@ -64,16 +71,42 @@ ErrorCode TimestampDictionaryReader::read(ZstdDecompressor& decompressor) {
         if (ErrorCodeSuccess != error) {
             return error;
         }
-        m_patterns[id] = TimestampPattern(0, pattern);
+        m_deprecated_patterns.emplace(id, TimestampPattern(0, pattern));
+        auto const timestamp_pattern_result{timestamp_parser::TimestampPattern::create(pattern)};
```
LinZhihao-723:
```suggestion
        auto timestamp_pattern_result{timestamp_parser::TimestampPattern::create(pattern)};
```
Need to drop `const` to make the coming `std::move` meaningful

components/core/src/clp_s/TimestampDictionaryReader.cpp
LinZhihao-723: Is it possible to pass the archive version into this open method?
In that way, we can make `m_timestamp_patterns` and `m_deprecated_patterns` into a union. It's a bit confusing when I saw the latest timestamp pattern is legally nullable...

gibber9809: I think that should be possible -- I'll probably keep the `m_deprecated_patterns` and `m_timestamp_patterns` separate, but make them not nullable/populate the relevant one based on archive version.

components/core/src/clp_s/TimestampDictionaryReader.cpp
```diff
@@ -64,16 +75,46 @@ ErrorCode TimestampDictionaryReader::read(ZstdDecompressor& decompressor) {
...
+                SPDLOG_ERROR(
+                        "Error loading timestamp pattern `{}` - {} - {}",
+                        pattern,
+                        timestamp_error.category().name(),
+                        timestamp_error.message()
+                );
+                return ErrorCodeCorrupt;
+            }
+            m_timestamp_patterns.emplace(id, std::move(timestamp_pattern_result.value()));
+        }
```
LinZhihao-723: Nit refactoring
```suggestion
        if (archive_version < cNewTimestampFormatVersion) {
            m_deprecated_patterns.emplace(id, TimestampPattern(0, pattern));
            continue;
        }

        auto timestamp_pattern_result{timestamp_parser::TimestampPattern::create(pattern)};
        if (timestamp_pattern_result.has_error()) {
            auto const& timestamp_error{timestamp_pattern_result.error()};
            SPDLOG_ERROR(
                    "Error loading timestamp pattern `{}` - {} - {}",
                    pattern,
                    timestamp_error.category().name(),
                    timestamp_error.message()
            );
            return ErrorCodeCorrupt;
        }
        m_timestamp_patterns.emplace(id, std::move(timestamp_pattern_result.value()));
```

components/core/src/clp_s/clp-s.cpp
```diff
@@ -183,10 +183,12 @@ bool search_archive(
         return true;
     }
 
-    ast::SetTimestampLiteralPrecision date_precision_pass{
-            ast::TimestampLiteral::Precision::Milliseconds
-    };
-    expr = date_precision_pass.run(expr);
+    if (archive_reader->get_header().version < clp_s::cNewTimestampFormatVersion) {
```
LinZhihao-723: We have two places using this version comparison verbosely. How about:
- Instead of calling `cNewTimestampFormatVersion`, we use `cDeprecatedTimestampFormatVersionMarker`, since "new" format won't always be "new" as the time goes on, lol
- We wrap this raw version comparison with a helper like `is_deprecated_timestamp_format`.

Optionally, we can be more verbose on what's deprecated, like `DeprecatedMillisecondTimestampFormatXXX`

components/core/src/clp_s/indexer/IndexManager.cpp
LinZhihao-723, line 1: I'm not too familiar with this component. Can u explain why we don't need the special handling here anymore? Is it mainly because the timestamp node type is set throughout the KQL AST already (with the updated syntax)?

components/core/src/clp_s/search/QueryRunner.cpp
```diff
@@ -73,16 +74,24 @@ void QueryRunner::initialize_reader(int32_t column_id, BaseColumnReader* column_
              & node_to_literal_type(m_schema_tree->get_node(column_id).get_type())))
         || m_match->schema_searches_against_column(m_schema, column_id))
     {
-        auto* clp_reader = dynamic_cast<ClpStringColumnReader*>(column_reader);
-        auto* var_reader = dynamic_cast<VariableStringColumnReader*>(column_reader);
-        auto* date_reader = dynamic_cast<DateStringColumnReader*>(column_reader);
-        if (nullptr != clp_reader && clp_reader->get_type() == NodeType::ClpString) {
+        if (auto const clp_reader = dynamic_cast<ClpStringColumnReader*>(column_reader);
```
LinZhihao-723:
```suggestion
        if (auto* const clp_reader = dynamic_cast<ClpStringColumnReader*>(column_reader);
```

components/core/src/clp_s/search/QueryRunner.cpp
```diff
@@ -73,16 +74,24 @@ void QueryRunner::initialize_reader(int32_t column_id, BaseColumnReader* column_
...
+            nullptr != clp_reader && NodeType::ClpString == clp_reader->get_type())
+        {
             m_clp_string_readers[column_id].push_back(clp_reader);
-        } else if (nullptr != var_reader && var_reader->get_type() == NodeType::VarString) {
+        } else if (auto const var_reader = dynamic_cast<VariableStringColumnReader*>(column_reader);
```
LinZhihao-723:
```suggestion
        } else if (auto* const var_reader = dynamic_cast<VariableStringColumnReader*>(column_reader);
```

components/core/src/clp_s/search/QueryRunner.cpp
```diff
@@ -73,16 +74,24 @@ void QueryRunner::initialize_reader(int32_t column_id, BaseColumnReader* column_
...
+                   nullptr != var_reader)
+        {
             m_var_string_readers[column_id].push_back(var_reader);
-        } else if (nullptr != date_reader) {
-            // Datestring readers with a given column ID are guaranteed not to repeat
-            m_datestring_readers.emplace(column_id, date_reader);
+        } else if (auto const timestamp_reader
```
LinZhihao-723:
```suggestion
        } else if (auto* const timestamp_reader
```

components/core/src/clp_s/search/QueryRunner.cpp
```diff
@@ -73,16 +74,24 @@ void QueryRunner::initialize_reader(int32_t column_id, BaseColumnReader* column_
...
+                   = dynamic_cast<TimestampColumnReader*>(column_reader);
+                   nullptr != timestamp_reader)
+        {
+            m_timestamp_readers.emplace(column_id, timestamp_reader);
+        } else if (auto* deprecated_date_reader
```
LinZhihao-723:
```suggestion
        } else if (auto* const deprecated_date_reader
```

components/core/src/clp_s/search/QueryRunner.hpp
```diff
@@ -133,7 +133,8 @@ class QueryRunner : public FilterClass {
     std::unordered_map<ast::Expression*, std::unordered_set<int64_t>*> m_expr_var_match_map;
     std::unordered_map<int32_t, std::vector<ClpStringColumnReader*>> m_clp_string_readers;
     std::unordered_map<int32_t, std::vector<VariableStringColumnReader*>> m_var_string_readers;
-    std::unordered_map<int32_t, DateStringColumnReader*> m_datestring_readers;
+    std::unordered_map<int32_t, TimestampColumnReader*> m_timestamp_readers;
+    DeprecatedDateStringColumnReader* m_deprecated_datestring_reader{nullptr};
```
LinZhihao-723, line 137: A high-level question before I can proceed:
Would this be mutually exclusive?
Like the runner can only be using the deprecated datestring reader, or it's using the latest?

LinZhihao-723 (changes requested):
Partially reviewed `QueryRunner`. Will get into more details later.

components/core/src/clp_s/indexer/IndexManager.cpp
gibber9809: This was basically a hack to make things easier for our presto integration -- since we would always index the authoritative timestamp column as DateString, presto would be able to understand that it was a Timestamp column, and treat it as a presto Timestamp type (even when the underlying column was int or float in the archive). Now, the timestamp column in clp-s will always have Timestamp type (and even if we add more timestamp types in the future we won't go back to allowing the timestamp column to be stored as into or float), so this hack is no longer required.

components/core/src/clp_s/search/QueryRunner.hpp
gibber9809: Yes, they're mutually exclusive -- now that we're introducing `Timestamp` we won't allow use of `DateString` in new archives. Also, for `DateString`, I'm getting rid of the code for handling the cases where we have multiple `DateString` columns in a schema, since we ended up never releasing a version of the compressor that supported compressing multiple columns as `DateString`.

gibber9809:
Did some quick profiling/investigation of the cases where we have slower decompression speeds. Seems like it mostly comes from our usage of `fmt::format` (particularly for padded integer fields), which we suspected might happen. Replacing these usages with our own function to marshal these padded fields seems to eliminate the slowdown for the datasets where we perform poorly right now -- I'll put up a follow-up PR that improves performance for these cases once this PR is merged.

LinZhihao-723:
> Did some quick profiling/investigation of the cases where we have slower decompression speeds. Seems like it mostly comes from our usage of `fmt::format` (particularly for padded integer fields), which we suspected might happen. Replacing these usages with our own function to marshal these padded fields seems to eliminate the slowdown for the datasets where we perform poorly right now -- I'll put up a follow-up PR that improves performance for these cases once this PR is merged.

Can u create an issue to track this?

components/core/src/clp_s/search/QueryRunner.cpp
```diff
@@ -228,11 +238,16 @@ bool QueryRunner::evaluate_wildcard_filter(FilterExpr* expr, int32_t schema) {
     }
 
     if (column->matches_type(LiteralType::TimestampT)) {
-        for (auto entry : m_datestring_readers) {
+        if (nullptr != m_deprecated_datestring_reader) {
+            if (evaluate_epoch_date_filter(op, m_deprecated_datestring_reader, literal)) {
+                return true;
+            }
+        }
```
LinZhihao-723: Since `m_deprecated_datestring_reader` and `m_timestamp_readers` are mutually exclusive, shall we add `continue` instead of falling back to `m_timestamp_readers` check?
In general, I think there might be a better way to do this (for example, using a variant). But let's defer this to a future PR...

components/core/src/clp_s/search/QueryRunner.cpp
```diff
@@ -307,12 +322,22 @@ bool QueryRunner::evaluate_filter(FilterExpr* expr, int32_t schema) {
...
+                return evaluate_epoch_date_filter(
+                        expr->get_operation(),
+                        m_deprecated_datestring_reader,
+                        literal
+                );
+            }
+            return evaluate_timestamp_filter(
                     expr->get_operation(),
-                    m_datestring_readers[column_id],
+                    m_timestamp_readers.at(column_id),
```
LinZhihao-723, line 338: I guess the upstream caller will ensure the `column_id` here must be a valid key in the reader map?

gibber9809: Yep, it happens in the code to prepare a given ERT to be searched.

components/core/src/clp_s/search/QueryRunner.cpp
gibber9809: Sorry, could you explain what you mean by that? I could put the iteration over the timestamp readers into the else case, but in general we can't short circuit this function because we need to go through all of the columns for the other possibly matching types.

E.g., for a query like `* > 10` against the record `{"ts": "1970-01-01", "int": 11}`, where `ts` is stored as a `DateString` completely shortcircuiting this function after the `DateString` case would lead us to falsely reject this record since we wouldn't end up falling into the code block below where we consider all of the primitive types.

LinZhihao-723: Uh right, sorry for the confusion. I think the right thing to do is probably put `m_timestamp_reader` in an else case.

gibber9809:
@junhaoliao could you take a look at the presto webui change?

Presto/velox won't work with the v0.5.0 archives introduced by this PR until I update them to understand the new Timestamp column type, but I figured I'd try to update the presto webui code related to finding the timestamp column to work with v0.5.0 archives while I was changing other references the DateString column. I can revert the presto webui change and put it in a separate small PR after the presto/velox changes are merged if you prefer though.

I'm fairly certain this change won't affect the rest of the webui, since I made sure the timestamp we associate with each search result as extra metadata is still millisecond precision after this change, and when I run the webui the timeline & timestamps appear correct.

junhaoliao:
> @junhaoliao could you take a look at the presto webui change?
> 
> 
> 
> Presto/velox won't work with the v0.5.0 archives introduced by this PR until I update them to understand the new Timestamp column type, but I figured I'd try to update the presto webui code related to finding the timestamp column to work with v0.5.0 archives while I was changing other references the DateString column. I can revert the presto webui change and put it in a separate small PR after the presto/velox changes are merged if you prefer though.
> 
> 
> 
> I'm fairly certain this change won't affect the rest of the webui, since I made sure the timestamp we associate with each search result as extra metadata is still millisecond precision after this change, and when I run the webui the timeline & timestamps appear correct.
> 
> 

will do. Thanks for also updating the webui part from your side!

components/webui/client/src/pages/SearchPage/SearchState/Presto/useTimestampKeyInit/sql.ts
junhaoliao, line 1: lgtm. risk of merging is low as well - the change is self-contained and doesn't affect the rest of the WebUI.

LinZhihao-723 (approved):
For the PR title, how about:
```
feat(clp-s)!: Add `Timestamp` column type to deprecate `DateString` column type; Bump the archive version to `0.5.0`.
```

junhaoliao:
After this PR was merged, timestamp strings that are not ISO 8601–compliant (e.g., RFC 2822 / RFC 822 formats) are now rejected instead of being parsed on a best-effort basis. Attempting to compress logs with such timestamps will produce errors like:

```
2026-02-19T12:49:09.225+00:00 [error] Failed to parse timestamp `"2023-03-27 00:26:35.719 EDT"` against known timestamp patterns.
2026-02-19T12:49:09.226+00:00 [error] FileWriter not closed before being destroyed - may cause data loss
2026-02-19T12:49:09.226+00:00 [error] FileWriter not closed before being destroyed - may cause data loss
2026-02-19T12:49:09.226+00:00 [error] FileWriter not closed before being destroyed - may cause data loss
2026-02-19T12:49:09.226+00:00 [error] Encountered error during compression - /home/junhao/workspace/2-clp/components/core/src/clp_s/TimestampDictionaryWriter.cpp:74  Error code: 16
```

Since none of the existing CLP users rely on non–ISO8601–compliant timestamps, and our goal is to ensure strict and accurate timestamp parsing, we will retain the current behavior of rejecting unsupported timestamp formats.

@coderabbitai please create: 

1. An issue to temporarily remove non-compliant datasets (e.g., the “postgresql” dataset) from the resources documentation:
   https://github.com/y-scope/clp/blob/a456dc914f26e85344e5ffe46471a45760fe804b/docs/src/user-docs/resources-datasets.md

2. A separate feature request to add support for parsing RFC 2822 / RFC 822 timestamp strings, after which the affected datasets can be restored to the resources documentation.

junhaoliao:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> `@junhaoliao`, I'll create both issues for you now.
> 
> 🏁 Script executed:
> 
> ```
> #!/bin/bash
> 
> # Issue 1: Temporarily remove non-compliant datasets from documentation
> gh issue create \
>   --title "docs: Temporarily remove non-ISO8601 datasets from resources documentation" \
>   --body "## Background
> 
> After PR `#1788` was merged, CLP-S now uses stricter timestamp parsing that rejects non–ISO8601–compliant timestamp formats (e.g., RFC 2822 / RFC 822 formats). Attempting to compress logs with such timestamps now produces errors:
> 
> \`\`\`
> 2026-02-19T12:49:09.225+00:00 [error] Failed to parse timestamp \\"2023-03-27 00:26:35.719 EDT\\" against known timestamp patterns.
> 2026-02-19T12:49:09.226+00:00 [error] Encountered error during compression
> \`\`\`
> 
> ## Task
> 
> Temporarily remove datasets with non–ISO8601–compliant timestamps (e.g., the \"postgresql\" dataset) from the resources documentation at:
> https://github.com/y-scope/clp/blob/a456dc914f26e85344e5ffe46471a45760fe804b/docs/src/user-docs/resources-datasets.md
> 
> These datasets can be restored once RFC 2822 / RFC 822 timestamp parsing support is added (see related feature request).
> 
> ## References
> 
> - PR: https://github.com/y-scope/clp/pull/1788
> - Comment: https://github.com/y-scope/clp/pull/1788#discussion_r1914400000
> - Requested by: `@junhaoliao`" \
>   --assignee junhaoliao \
>   --label documentation
> 
> # Issue 2: Feature request for RFC 2822/RFC 822 timestamp support
> gh issue create \
>   --title "feat: Add support for parsing RFC 2822 / RFC 822 timestamp strings" \
>   --body "## Feature Request
> 
> Add support for parsing RFC 2822 and RFC 822 timestamp formats in CLP-S timestamp parser.
> 
> ## Background
> 
> Currently, CLP-S only accepts ISO8601-compliant timestamp formats. After PR `#1788` introduced stricter timestamp parsing, non-compliant formats like RFC 2822 / RFC 822 are rejected, causing compression errors:
> 
> \`\`\`
> 2026-02-19T12:49:09.225+00:00 [error] Failed to parse timestamp \\"2023-03-27 00:26:35.719 EDT\\" against known timestamp patterns.
> \`\`\`
> 
> ## Use Case
> 
> Some datasets (e.g., \"postgresql\" logs) use RFC 2822 / RFC 822 timestamp formats. Supporting these formats would allow CLP to compress a wider range of log datasets.
> 
> ## Acceptance Criteria
> 
> - CLP-S timestamp parser can parse RFC 2822 and RFC 822 timestamp strings
> - Affected datasets can be restored to the resources documentation
> 
> ## References
> 
> - PR: https://github.com/y-scope/clp/pull/1788
> - Comment: https://github.com/y-scope/clp/pull/1788#discussion_r1914400000
> - Requested by: `@junhaoliao`" \
>   --assignee junhaoliao \
>   --label enhancement
> ```
> 
> Length of output: 3173
> 
> Let me fix the script to properly handle the multi-line issue bodies:
> 
> 2026-02-19T12:49:09.225+00:00 [error] Failed to parse timestamp "2023-03-27 00:26:35.719 EDT" against known timestamp patterns. 2026-02-19T12:49:09.226+00:00 [error] Encountered error during compression
> 
> ```
> 
> ## Task
> 
> Temporarily remove datasets with non–ISO8601–compliant timestamps (e.g., the "postgresql" dataset) from the resources documentation at:
> https://github.com/y-scope/clp/blob/a456dc914f26e85344e5ffe46471a45760fe804b/docs/src/user-docs/resources-datasets.md
> 
> These datasets can be restored once RFC 2822 / RFC 822 timestamp parsing support is added (see related feature request).
> 
> ## References
> 
> - PR: https://github.com/y-scope/clp/pull/1788
> - Comment: https://github.com/y-scope/clp/pull/1788#issuecomment
> - Requested by: `@junhaoliao`
> EOF
> 
> # Write Issue 2 body
> cat > "$ISSUE2_BODY" << 'EOF'
> ## Feature Request
> 
> Add support for parsing RFC 2822 and RFC 822 timestamp formats in CLP-S timestamp parser.
> 
> ## Background
> 
> Currently, CLP-S only accepts ISO8601-compliant timestamp formats. After PR `#1788` introduced stricter timestamp parsing, non-compliant formats like RFC 2822 / RFC 822 are rejected, causing compression errors:
> ```
> 
> 2026-02-19T12:49:09.225+00:00 [error] Failed to parse timestamp "2023-03-27 00:26:35.719 EDT" against known timestamp patterns.
> 
> ```
> 
> ## Use Case
> 
> Some datasets (e.g., "postgresql" logs) use RFC 2822 / RFC 822 timestamp formats. Supporting these formats would allow CLP to compress a wider range of log datasets.
> 
> ## Acceptance Criteria
> 
> - CLP-S timestamp parser can parse RFC 2822 and RFC 822 timestamp strings
> - Affected datasets can be restored to the resources documentation
> 
> ## References
> 
> - PR: https://github.com/y-scope/clp/pull/1788
> - Comment: https://github.com/y-scope/clp/pull/1788#issuecomment
> - Requested by: `@junhaoliao`
> EOF
> 
> # Create Issue 1
> echo "Creating Issue 1..."
> gh issue create \
>   --title "docs: Temporarily remove non-ISO8601 datasets from resources documentation" \
>   --body-file "$ISSUE1_BODY" \
>   --assignee junhaoliao \
>   --label documentation
> 
> echo ""
> echo "Creating Issue 2..."
> # Create Issue 2
> gh issue create \
>   --title "feat: Add support for parsing RFC 2822 / RFC 822 timestamp strings" \
>   --body-file "$ISSUE2_BODY" \
>   --assignee junhaoliao \
>   --label enhancement
> 
> # Clean up temporary files
> rm -f "$ISSUE1_BODY" "$ISSUE2_BODY"
> 
> echo ""
> echo "Both issues created successfully!"
> ```
> 
> 🏁 Script executed:
> 
> ```
> #!/bin/bash
> 
> # Create temporary files for issue bodies
> ISSUE1_BODY=$(mktemp)
> ISSUE2_BODY=$(mktemp)
> 
> # Write Issue 1 body
> cat > "$ISSUE1_BODY" << 'EOF'
> ## Background
> 
> After PR `#1788` was merged, CLP-S now uses stricter timestamp parsing that rejects non–ISO8601–compliant timestamp formats (e.g., RFC 2822 / RFC 822 formats). Attempting to compress logs with such timestamps now produces errors:
> ```
> 
> Length of output: 92
> 
> Let me use Python to avoid shell escaping issues:
> 
> 2026-02-19T12:49:09.225+00:00 [error] Failed to parse timestamp "2023-03-27 00:26:35.719 EDT" against known timestamp patterns. 2026-02-19T12:49:09.226+00:00 [error] Encountered error during compression
> 
> ```
> 
> ## Task
> 
> Temporarily remove datasets with non–ISO8601–compliant timestamps (e.g., the "postgresql" dataset) from the resources documentation at:
> https://github.com/y-scope/clp/blob/a456dc914f26e85344e5ffe46471a45760fe804b/docs/src/user-docs/resources-datasets.md
> 
> These datasets can be restored once RFC 2822 / RFC 822 timestamp parsing support is added (see related feature request).
> 
> ## References
> 
> - PR: https://github.com/y-scope/clp/pull/1788
> - Requested by: `@junhaoliao`"""
> 
> # Issue 2 body
> issue2_body = """## Feature Request
> 
> Add support for parsing RFC 2822 and RFC 822 timestamp formats in CLP-S timestamp parser.
> 
> ## Background
> 
> Currently, CLP-S only accepts ISO8601-compliant timestamp formats. After PR `#1788` introduced stricter timestamp parsing, non-compliant formats like RFC 2822 / RFC 822 are rejected, causing compression errors:
> ```
> 
> 2026-02-19T12:49:09.225+00:00 [error] Failed to parse timestamp "2023-03-27 00:26:35.719 EDT" against known timestamp patterns.
> 
> ```
> 
> ## Use Case
> 
> Some datasets (e.g., "postgresql" logs) use RFC 2822 / RFC 822 timestamp formats. Supporting these formats would allow CLP to compress a wider range of log datasets.
> 
> ## Acceptance Criteria
> 
> - CLP-S timestamp parser can parse RFC 2822 and RFC 822 timestamp strings
> - Affected datasets can be restored to the resources documentation
> 
> ## References
> 
> - PR: https://github.com/y-scope/clp/pull/1788
> - Requested by: `@junhaoliao`"""
> 
> try:
>     # Create Issue 1
>     print("Creating Issue 1...")
>     result1 = subprocess.run(
>         [
>             "gh", "issue", "create",
>             "--title", "docs: Temporarily remove non-ISO8601 datasets from resources documentation",
>             "--body", issue1_body,
>             "--assignee", "junhaoliao",
>             "--label", "documentation"
>         ],
>         capture_output=True,
>         text=True,
>         check=True
>     )
>     print(result1.stdout)
>     
>     # Create Issue 2
>     print("\nCreating Issue 2...")
>     result2 = subprocess.run(
>         [
>             "gh", "issue", "create",
>             "--title", "feat: Add support for parsing RFC 2822 / RFC 822 timestamp strings",
>             "--body", issue2_body,
>             "--assignee", "junhaoliao",
>             "--label", "enhancement"
>         ],
>         capture_output=True,
>         text=True,
>         check=True
>     )
>     print(result2.stdout)
>     
>     print("\nBoth issues created successfully!")
>     
> except subprocess.CalledProcessError as e:
>     print(f"Error: {e}", file=sys.stderr)
>     print(f"stdout: {e.stdout}", file=sys.stderr)
>     print(f"stderr: {e.stderr}", file=sys.stderr)
>     sys.exit(1)
> 
> PYTHON_SCRIPT
> ```
> 
> 🏁 Script executed:
> 
> ```
> #!/bin/bash
> 
> python3 << 'PYTHON_SCRIPT'
> import subprocess
> import sys
> 
> # Issue 1 body
> issue1_body = """## Background
> 
> After PR `#1788` was merged, CLP-S now uses stricter timestamp parsing that rejects non–ISO8601–compliant timestamp formats (e.g., RFC 2822 / RFC 822 formats). Attempting to compress logs with such timestamps now produces errors:
> ```
> 
> Length of output: 258
> 
> Let me create the issues one at a time with simpler scripts:
> 
> 🏁 Script executed:
> 
> ```
> #!/bin/bash
> # Create Issue 1 only
> 
> cat > /tmp/issue1.txt << 'END'
> ## Background
> 
> After PR `#1788` was merged, CLP-S now uses stricter timestamp parsing that rejects non-ISO8601-compliant timestamp formats (e.g., RFC 2822 / RFC 822 formats). Attempting to compress logs with such timestamps now produces errors.
> 
> ## Task
> 
> Temporarily remove datasets with non-ISO8601-compliant timestamps (e.g., the "postgresql" dataset) from the resources documentation at:
> https://github.com/y-scope/clp/blob/a456dc914f26e85344e5ffe46471a45760fe804b/docs/src/user-docs/resources-datasets.md
> 
> These datasets can be restored once RFC 2822 / RFC 822 timestamp parsing support is added.
> 
> ## References
> 
> - PR: https://github.com/y-scope/clp/pull/1788
> - Requested by: `@junhaoliao`
> END
> 
> gh issue create \
>   --title "docs: Temporarily remove non-ISO8601 datasets from resources documentation" \
>   --body-file /tmp/issue1.txt \
>   --assignee junhaoliao \
>   --label documentation
> 
> rm /tmp/issue1.txt
> ```
> 
> Length of output: 75
> 
> `@junhaoliao`, I apologize—I'm unable to create GitHub issues directly because the sandbox environment doesn't have GitHub authentication configured. However, I've prepared the complete issue templates below for you to create manually:
> 
> ## Issue 1: Documentation Update
> **Title:** `docs: Temporarily remove non-ISO8601 datasets from resources documentation`
> 
> **Labels:** `documentation`
> 
> **Assignee:** `junhaoliao`
> 
> **Body:**
> 
> ```
> ## Background
> 
> After PR `#1788` was merged, CLP-S now uses stricter timestamp parsing that rejects non–ISO8601–compliant timestamp formats (e.g., RFC 2822 / RFC 822 formats). Attempting to compress logs with such timestamps now produces errors:
> ```
> 
> 2026-02-19T12:49:09.225+00:00 [error] Failed to parse timestamp "2023-03-27 00:26:35.719 EDT" against known timestamp patterns. 2026-02-19T12:49:09.226+00:00 [error] Encountered error during compression
> 
> ```
> 
> ## Task
> 
> Temporarily remove datasets with non–ISO8601–compliant timestamps (e.g., the "postgresql" dataset) from the resources documentation at:
> https://github.com/y-scope/clp/blob/a456dc914f26e85344e5ffe46471a45760fe804b/docs/src/user-docs/resources-datasets.md
> 
> These datasets can be restored once RFC 2822 / RFC 822 timestamp parsing support is added (see related feature request).
> 
> ## References
> 
> - PR: https://github.com/y-scope/clp/pull/1788
> - Requested in: https://github.com/y-scope/clp/pull/1788#discussion_comment
> ```
> 
> ## Issue 2: Feature Request
> **Title:** `feat: Add support for parsing RFC 2822 / RFC 822 timestamp strings`
> 
> **Labels:** `enhancement`
> 
> **Assignee:** `junhaoliao`
> 
> **Body:**
> 
> ```
> ## Feature Request
> 
> Add support for parsing RFC 2822 and RFC 822 timestamp formats in CLP-S timestamp parser.
> 
> ## Background
> 
> Currently, CLP-S only accepts ISO8601-compliant timestamp formats. After PR `#1788` introduced stricter timestamp parsing, non-compliant formats like RFC 2822 / RFC 822 are rejected, causing compression errors:
> ```
> 
> 2026-02-19T12:49:09.225+00:00 [error] Failed to parse timestamp "2023-03-27 00:26:35.719 EDT" against known timestamp patterns.
> 
> ```
> 
> ## Use Case
> 
> Some datasets (e.g., "postgresql" logs) use RFC 2822 / RFC 822 timestamp formats. Supporting these formats would allow CLP to compress a wider range of log datasets.
> 
> ## Acceptance Criteria
> 
> - CLP-S timestamp parser can parse RFC 2822 and RFC 822 timestamp strings
> - Affected datasets can be restored to the resources documentation
> 
> ## References
> 
> - PR: https://github.com/y-scope/clp/pull/1788
> - Requested in: https://github.com/y-scope/clp/pull/1788#discussion_comment
> ```
> 
> You can create these issues by:
> 
> 1. Going to [y-scope/clp/issues/new](https://github.com/y-scope/clp/issues/new)
> 2. Copy-pasting the title and body for each issue
> 3. Adding the appropriate labels and assignee
> 
> Or use the GitHub CLI from your authenticated terminal:
> 
> ```shell
> gh issue create --title "..." --body-file <file> --assignee junhaoliao --label <label>
> ```
> 
> 🧠 Learnings used

@coderabbitai i'm sure you can create the issues. try again

PR #1798, opened 2025-12-18
feat(webui): Add dataset and paths to compression job table.

davemarco:
# Description

PR adds support for showing dataset and paths in webui ingest table. 

Unfortunately these fields cannot be queried since they are compressed blobs in the database. 

At first i tried to decompress client side, but code was a mess. So i moved the entire query to the server, and decompressed on the server. I think longer term this is probably a better approach for all yscope (non-user) webui queries. Initially we thought it was better for user dashboard capability in the future to send queries from client, but for querying our own clp metadata server handling may be better. One advantage is fastify will type the response.

Since fastify is typing response. I cleaned up the compress types junhao added for his compress job form, since they are reused here. Unfortunately, some fields were missing, so i added them, and then allowed partial types so it wont give errors for anyone who compressed with v0.6.0 in UI. 

Did not add support for s3 here. paths are jsut hidden. The PR is already large.

# Checklist

  breaking change.

# Validation performed

Tested with clp-s
Testing with clp and S3 still underway. Even though there is no s3 support, the fastify validation may still cause errors so need to check with command line compression.

## Summary by CodeRabbit

* **New Features**
  * Jobs list now fetches recent compression metadata from a new backend endpoint.
  * Added Dataset and Paths columns to the compression jobs table; Paths support copyable text and tooltip ellipses.

* **Improvements**
  * Table layout supports horizontal scrolling and fixed layout for better readability.
  * Job entries now include inferred dataset and paths extracted from decoded job configuration.

### DIFF
--- a/components/core/src/clp/ffi/ir_stream/search/test/test_QueryHandlerImpl.cpp
+++ b/components/core/src/clp/ffi/ir_stream/search/test/test_QueryHandlerImpl.cpp
@@ -127,6 +127,9 @@ auto generate_matchable_kql_expressions(
             matchable_kql_expressions.emplace_back(
                     fmt::format("{}: {}", column_query_with_namespace, cRefTestInt)
             );
+            matchable_kql_expressions.emplace_back(
+                    fmt::format("{}: timestamp(\"{}\")", column_query_with_namespace, cRefTestInt)
+            );
             auto [it, inserted] = expected_column_resolutions.try_emplace(
                     column_query_with_namespace,
                     std::set<SchemaTree::Node::id_t>{}
@@ -144,6 +147,13 @@ auto generate_matchable_kql_expressions(
         {
             matchable_kql_expressions.emplace_back(
                     fmt::format("{}: {:.2f}", column_query_with_namespace, cRefTestFloat)
+            );
+            matchable_kql_expressions.emplace_back(
+                    fmt::format(
+                            "{}: timestamp(\"{:.3f}\")",
+                            column_query_with_namespace,
+                            cRefTestFloat
+                    )
             );
             auto [it, inserted] = expected_column_resolutions.try_emplace(
                     column_query_with_namespace,
--- a/components/core/src/clp/ffi/ir_stream/search/test/test_utils.cpp
+++ b/components/core/src/clp/ffi/ir_stream/search/test/test_utils.cpp
@@ -22,6 +22,7 @@
 #include "../../../../../clp_s/search/ast/Integral.hpp"
 #include "../../../../../clp_s/search/ast/Literal.hpp"
 #include "../../../../../clp_s/search/ast/StringLiteral.hpp"
+#include "../../../../../clp_s/search/ast/TimestampLiteral.hpp"
 #include "../../../../ir/types.hpp"
 #include "../../../EncodedTextAst.hpp"
 #include "../../../Value.hpp"
@@ -44,6 +45,7 @@ using clp_s::search::ast::Integral;
 using clp_s::search::ast::Literal;
 using clp_s::search::ast::LiteralType;
 using clp_s::search::ast::StringLiteral;
+using clp_s::search::ast::TimestampLiteral;
 
 using ValueToMatchedFilterOpsPair
         = std::pair<std::optional<Value>, std::unordered_set<FilterOperation>>;
@@ -393,11 +395,13 @@ TEST_CASE("ffi_ir_stream_search_filter_evaluation", "[ffi][ir_stream][search]") 
     auto ref_bool_literal{BooleanLiteral::create_from_bool(cRefBool)};
     auto ref_int_literal{Integral::create_from_int(cRefInt)};
     auto ref_float_literal{Integral::create_from_float(cRefFloat)};
+    auto ref_timestamp_literal{TimestampLiteral::create(cRefInt)};
 
     REQUIRE_FALSE((nullptr == ref_str_literal));
     REQUIRE_FALSE((nullptr == ref_bool_literal));
     REQUIRE_FALSE((nullptr == ref_int_literal));
     REQUIRE_FALSE((nullptr == ref_float_literal));
+    REQUIRE_FALSE((nullptr == ref_timestamp_literal));
 
     constexpr std::string_view cColumnName{"test_column"};
     auto column_descriptor{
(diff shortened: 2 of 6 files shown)


components/webui/server/src/routes/api/compress-metadata/index.ts
```diff
@@ -0,0 +1,44 @@
...
+                response: {
+                    [constants.HTTP_STATUS_OK]: Type.Array(
+                        CompressionMetadataDecodedSchema
+                    ),
+                },
+                tags: ["Compression Metadata"],
+            },
+        },
+        async () => {
+            console.log("Fetching compression metadata...");
```
hoophalab: Use fastify logger as coderabbit suggested?

components/webui/common/src/schemas/stream-file-metadata.ts
hoophalab, line 1: This file isn't meant to be included in this PR?

components/webui/server/src/routes/api/compress-metadata/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+    jobConfig: unknown
+): {clp_config: ClpIoConfig} => {
+    if (!(jobConfig instanceof Buffer)) {
+        throw new Error("Missing clp_config buffer for compression metadata");
+    }
+
+    try {
+        const decodedClpConfig = decode(
+            brotliDecompressSync(jobConfig)
+        ) as ClpIoConfig;
```
hoophalab: can we use typebox's `Value.Parse` here instead of `as`?

components/webui/common/src/schemas/compression.ts
```diff
@@ -72,14 +72,97 @@ const CompressionJobSchema = Type.Object({
...
+    target_segment_size: Type.Number(),
+});
+
+/**
+ * Matching `ClpIoConfig` in `job_orchestration.scheduler.job_config`.
+ */
+const ClpIoConfigSchema =
+    Type.Object({
+        input: Type.Union([ClpIoFsInputConfigSchema,
+            (ClpIoS3InputConfigSchema)]),
```
hoophalab: Can we remove the parentheses around ClpIoS3InputConfigSchema?

components/webui/common/src/schemas/compression.ts
hoophalab, line 1: How about renaming this file as job_config to keep consistent with python code?

components/webui/common/src/schemas/compress-metadata.ts
```diff
@@ -0,0 +1,59 @@
...
+
+
+export {
+    CompressionMetadataDecodedSchema,
+    CompressionMetadataSchema,
+};
+export type {
+    CompressionMetadata,
+    CompressionMetadataDecoded,
+};
```
hoophalab, line 59: not sure if we still need to export CompressionMetadata since only CompressionMetadataDecoded is needed.

components/webui/client/src/pages/IngestPage/Jobs/utils.ts
```diff
@@ -1,33 +1,104 @@
...
+        }
+
+        return input.paths_to_compress;
+    }
+
+    // eslint-disable-next-line no-warning-comments
+    // TODO: Add support to parse S3 paths; may need a bucket prefix from the CLP config.
+
+    return [];
+};
```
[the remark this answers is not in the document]
hoophalab: Coderabbit is probably correct? `extract.*` doesn't sound that it will modify the argument.

components/webui/client/src/pages/IngestPage/Jobs/utils.ts
hoophalab, line 1: Can we rename variable clpConfig in this file as clpIoConfig as well? clpConfig sounds like clp-config.yaml

components/webui/server/src/routes/api/compress-metadata/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+
+
+/**
+ * Decodes a compressed job config stored in the database.
+ *
+ * @param jobConfig
+ * @return
+ * @throws {Error} When the buffer is missing or cannot be decoded.
+ */
+export const decodeJobConfig = (
```
hoophalab: IIRC, junhao said we should only export functions at the end of the file?

hoophalab (commented):
Some comments.

One high level question: Why would decoding the ClpIoConfig in the client be too messy? We should be able to just call decodeJobConfig on the client side. This would avoid adding a new schema to common and a new route to the server. Does the Brotli library not work in the browser?

Validation: the dataset and path are displayed correctly for files compressed using `compress.sh`

davemarco:
for high level comment. 

I tried decoding on the client, but Brotli depends on Node zlib, so in the browser it requires JS/WASM libs  — similar to how our client-side decompression is in the single-file archive code which looks a bit weird ([xz.ts](https://github.com/davemarco/yscope-log-viewer-fork/blob/sfa/src/utils/xz.ts).

Moving decompression to the server avoids that complexity, and gives us a typed Fastify response for the UI.

junhaoliao:
> I tried decoding on the client

kinda unrelated. and not to say we should do this today -

interestingly, browser implementers have finally been considering adding more decompression formats to the `DecompressionStream` API, which include "brotli" and "zstd":

https://developer.mozilla.org/en-US/docs/Web/API/DecompressionStream/DecompressionStream#brotli

the diff: https://github.com/mdn/content/pull/42312

the proposals make sense cuz modern browsers can already decode web page contents compressed in such modern compression formats - they just need to make such decoders available to the JS APIs. maybe in the coming year(s) we will not have to install external libraries to decompress data streams in formats the browsers can already decode.

components/webui/common/src/schemas/compression.ts
davemarco: i feel we have more stuff that the original job_config file. Also i think we are using - instead of _ in webui. In you want i can make a new file scheduler-job-config and put some of the types in there.

components/webui/common/src/schemas/compress-metadata.ts
davemarco: i change the type of the sql reply to use CompressionMetadata so now its used. I think its better to use it, then just not export it since not used. I think this what i originally planned for, but then didnt use it since there was already a type

components/webui/server/src/routes/api/compress-metadata/index.ts
davemarco: i just removed the log not neccesary.

davemarco:
@hoophalab the Value.Parse we added for the IO config was rejecting configs from earlier release. I am now using the partial type, then casting it.

davemarco:
based on ur log. it looks like your missing the type field. Maybe this is an issue with the command line not adding the FS type? I did now change it to be less picky

components/webui/server/src/routes/api/compress-metadata/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Decodes a compressed job config stored in the database.
+ *
+ * @param jobConfig
+ * @return
+ * @throws {Error} When the buffer is missing or cannot be decoded.
+ */
+const decodeJobConfig = (
+    jobConfig: unknown
+): {clp_config: ClpIoConfig} => {
+    if (!(jobConfig instanceof Buffer)) {
```
hoophalab:
```suggestion
    if (false === (jobConfig instanceof Buffer)) {
```

hoophalab (dismissed):
You are right. I was probably using an outdated python script. One more nitpick, otherwise all good.

Validation:
1. paths and datasets are displayed correctly in the table
2. paths of logs compressed from s3 are hidden

components/webui/client/src/api/compress-metadata/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+
+/**
+ * Retrieves recent compression jobs (last 30 days).
+ *
+ * @return Recent compression jobs metadata.
+ */
+const fetchCompressionJobs = async (): Promise<CompressionMetadataDecoded[]> => {
+    const {data} = await axios.get<CompressionMetadataDecoded[]>(
+        "/api/compress-metadata"
```
junhaoliao, line 12: just wondering, if users can `POST /api/compression` to submit a compression job, would it be more intuitive to `GET /api/compression` to get the submitted compression jobs' details?

junhaoliao (approved):

components/webui/client/src/api/compress-metadata/index.ts
davemarco: @coderabbitai can you make an issue for this I agree

davemarco:
Okay i have done final testing with s3 so i will merge. supposedly quinn hasnt tested package yet for s3.

PR #1801, opened 2025-12-18
feat(integration-tests): Test package compression.

quinntaylormitchell:
# Description

This PR implements integration tests for package compression. The following new features are added:

* New tests `test_clp_json_compression_json_multifile` and `test_clp_text_compression_text_multifile`.
* Functions `run_package_compression_script` and `verify_compression`.
* New class `PackageCompressionJob`.
* Two sample datasets: `json-multifile` and `text-multifile`.

Future development of compression package testing can be accomplished by adding new tests to the `test_clp_*` file using the same format as the `test_clp_*_compression_*` tests.

**NOTE:** `verify_compression` does not check for the correctness of compression when testing `clp-json`; to accomplish this check, we need #1299 merged.

# Checklist

  breaking change.

# Validation performed

```bash
uv run pytest -m 'package'
```

All `clp-json` and `clp-text` tests passed.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end compression tests for JSON and text multi-file datasets with setup, execution, verification and cleanup.

* **Documentation**
  * Added README metadata for JSON and text test datasets describing file counts, event volumes and timestamp ranges.

* **Chores**
  * Enhanced test tooling and utilities: compression job execution, verification helpers, directory cleanup, and an environment variable to locate test assets.

---
  - https://app.asana.com/0/0/1212888467142252

### DIFF
--- a/components/core/CMakeLists.txt
+++ b/components/core/CMakeLists.txt
@@ -402,6 +402,7 @@ set(SOURCE_FILES_clp_s_unitTest
     src/clp_s/search/QueryRunner.hpp
     src/clp_s/search/SchemaMatch.cpp
     src/clp_s/search/SchemaMatch.hpp
+    src/clp_s/test/test_InputConfig.cpp
     src/clp_s/TimestampDictionaryReader.cpp
     src/clp_s/TimestampDictionaryReader.hpp
     src/clp_s/TimestampDictionaryWriter.cpp
--- a/components/core/src/clp_s/CommandLineArguments.cpp
+++ b/components/core/src/clp_s/CommandLineArguments.cpp
@@ -206,6 +206,9 @@ CommandLineArguments::parse_arguments(int argc, char const** argv) {
 
             po::options_description compression_options("Compression options");
             std::string input_path_list_file_path;
+            bool normalize_file_paths{false};
+            std::string path_prefix_to_remove;
+            bool remove_leading_slash{false};
             std::string auth{cNoAuth};
             // clang-format off
             compression_options.add_options()(
@@ -248,6 +251,20 @@ CommandLineArguments::parse_arguments(int argc, char const** argv) {
                     "no-retain-float-format",
                     po::bool_switch(&m_no_retain_float_format),
                     "Do not store extra information to losslessly decompress floats."
+            )(
+                    "normalize-paths",
+                    po::bool_switch(&normalize_file_paths),
+                    "Make all paths for ingested files absolute relative to the root directory."
+            )(
+                    "remove-path-prefix",
+                    po::value<std::string>(&path_prefix_to_remove)
+                            ->value_name("DIR")
+                            ->default_value(path_prefix_to_remove),
+                    "Remove the given path prefix from each compressed file path."
+            )(
+                    "remove-leading-slash",
+                    po::bool_switch(&remove_leading_slash),
+                    "Remove the leading `/` from each compressed file path."
             )(
                     "single-file-archive",
                     po::bool_switch(&m_single_file_archive),
@@ -318,14 +335,94 @@ CommandLineArguments::parse_arguments(int argc, char const** argv) {
                 }
             }
 
+            std::optional<std::string> path_prefix_to_remove_option{};
+            if (false == path_prefix_to_remove.empty()) {
+                if (false == std::filesystem::exists(path_prefix_to_remove)) {
+                    throw std::invalid_argument("Specified prefix to remove does not exist.");
+                }
+                if (false == std::filesystem::is_directory(path_prefix_to_remove)) {
+                    throw std::invalid_argument("Specified prefix to remove is not a directory.");
+                }
+                if (normalize_file_paths) {
+                    std::error_code ec;
+                    auto const normalized_path{
+                            std::filesystem::canonical(path_prefix_to_remove, ec)
+                    };
+                    if (ec) {
+                        throw std::invalid_argument(
+                                fmt::format(
+                                        "Failed to normalize prefix {} - {}",
+                                        path_prefix_to_remove,
+                                        ec.message()
+                                )
+                        );
+                    }
+                    path_prefix_to_remove = normalized_path.string();
+                }
+                path_prefix_to_remove_option.emplace(path_prefix_to_remove);
+            }
+
             for (auto const& path : input_paths) {
-                if (false == get_input_files_for_raw_path(path, m_input_paths)) {
+                auto path_object{get_path_object_for_raw_path(path)};
+                if (normalize_file_paths && InputSource::Filesystem == path_object.source) {
+                    std::error_code ec;
+                    auto const normalized_path{std::filesystem::canonical(path_object.path, ec)};
+                    if (ec) {
+                        throw std::invalid_argument(
+                                fmt::format(
+                                        "Failed to normalize path {} - {}",
+                                        path_object.path,
+                                        ec.message()
+                                )
+                        );
+                    }
+                    path_object.path = normalized_path.string();
+                }
+                if (false == get_input_files_for_path(path_object, m_input_paths)) {
                     throw std::invalid_argument(fmt::format("Invalid input path \"{}\".", path));
                 }
             }
 
             if (m_input_paths.empty()) {
                 throw std::invalid_argument("No input paths specified.");
+            }
+
+            for (auto const& input_path : m_input_paths) {
+                if (false == path_prefix_to_remove_option.has_value()
+                    || InputSource::Filesystem != input_path.source)
+                {
+                    m_input_paths_and_canonical_filenames.emplace_back(input_path, input_path.path);
+                    continue;
+                }
+
+                auto const result_option{
+                        remove_path_prefix(input_path.path, path_prefix_to_remove_option.value())
+                };
+                if (false == result_option.has_value()) {
+                    throw std::invalid_argument(
+                            fmt::format(
+                                    "Failed to remove prefix \"{}\" from path \"{}\".",
+                                    path_prefix_to_remove_option.value(),
+                                    input_path.path
+                            )
+                    );
+                }
+                m_input_paths_and_canonical_filenames.emplace_back(
+                        input_path,
+                        result_option.value()
+                );
+            }
+
+            if (remove_leading_slash) {
+                for (auto& [input_path, canonical_file_name] :
+                     m_input_paths_and_canonical_filenames)
+                {
+                    if (InputSource::Filesystem == input_path.source
+                        && false == canonical_file_name.empty() && '/' == canonical_file_name.at(0))
+                    {
+                        canonical_file_name = canonical_file_name.substr(1);
+                    }
+                }
             }
 
             validate_network_auth(auth, m_network_auth);
--- a/components/core/src/clp_s/CommandLineArguments.hpp
+++ b/components/core/src/clp_s/CommandLineArguments.hpp
@@ -3,6 +3,7 @@
 
 #include <optional>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include <boost/program_options/option.hpp>
@@ -48,6 +49,11 @@ public:
     Command get_command() const { return m_command; }
 
     std::vector<Path> const& get_input_paths() const { return m_input_paths; }
+
+    [[nodiscard]] auto get_input_paths_and_canonical_filenames() const
+            -> std::vector<std::pair<Path, std::string>> const& {
+        return m_input_paths_and_canonical_filenames;
+    }
 
     NetworkAuthOption const& get_network_auth() const { return m_network_auth; }
 
@@ -193,6 +199,7 @@ private:
 
     // Compression and decompression variables
     std::vector<Path> m_input_paths;
+    std::vector<std::pair<Path, std::string>> m_input_paths_and_canonical_filenames;
     NetworkAuthOption m_network_auth{};
     std::string m_archives_dir;
     std::string m_output_dir;
--- a/components/core/src/clp_s/InputConfig.cpp
+++ b/components/core/src/clp_s/InputConfig.cpp
@@ -39,6 +39,37 @@ auto get_path_object_for_raw_path(std::string_view const path) -> Path {
     return Path{.source = get_source_for_path(path), .path = std::string{path}};
 }
 
+auto remove_path_prefix(std::string_view path, std::string_view prefix)
+        -> std::optional<std::string> {
+    try {
+        std::filesystem::path input_path{path};
+        std::filesystem::path prefix_path{prefix};
+
+        auto path_it = input_path.begin();
+        auto prefix_end_idx = prefix_path.end();
+        if (false == prefix_path.empty() && (--prefix_path.end())->empty()) {
+            --prefix_end_idx;
+        }
+        for (auto prefix_it = prefix_path.begin(); prefix_end_idx != prefix_it; ++prefix_it) {
+            if (input_path.end() == path_it) {
+                return std::nullopt;
+            }
+            if (*prefix_it != *path_it) {
+                return std::nullopt;
+            }
+            ++path_it;
+        }
+
+        std::filesystem::path path_without_prefix{"/"};
+        for (; input_path.end() != path_it; ++path_it) {
+            path_without_prefix.append(path_it->string());
+        }
+        return path_without_prefix.string();
+    } catch (std::exception const&) {
+        return std::nullopt;
+    }
+}
+
 auto get_input_files_for_raw_path(std::string_view const path, std::vector<Path>& files) -> bool {
     return get_input_files_for_path(get_path_object_for_raw_path(path), files);
 }
--- a/components/core/src/clp_s/InputConfig.hpp
+++ b/components/core/src/clp_s/InputConfig.hpp
@@ -3,6 +3,7 @@
 
 #include <cstdint>
 #include <memory>
+#include <optional>
 #include <string>
 #include <string_view>
 #include <utility>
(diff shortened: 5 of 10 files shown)


integration-tests/tests/package_tests/clp_json/data/json-multifile/README.md
```diff
@@ -0,0 +1,16 @@
+# Description of logs
```
Bill-hbrhbr, line 1: I think this is a nice touch that makes the test dataset more descriptive, but I’m a bit concerned that this
file will need frequent updates as the dataset evolves, for example if new entries fall outside the
specified timestamp range or additional test files are added.

Also, it might be more useful to record higher level metadata such as where the logs were extracted from,
the field name used for timestamps, and the schema.

integration-tests/tests/conftest.py
```diff
@@ -23,3 +23,14 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+        "--job-name-contains",
+        dest="JOB_NAME_CONTAINS",
+        help="Filter CLP jobs by their name (case-insensitive substring match).",
+    )
+    parser.addoption(
+        "--no-jobs",
+        action="store_true",
+        dest="NO_JOBS",
+        help="Only validate CLP package start and stop. Do not create or run any test jobs.",
+    )
```
Bill-hbrhbr: These additional parser options aren’t very complicated on their own, but I think we can achieve the same
filtering using pytest’s built in functionality. My concern is that if we keep adding custom options, the
testing workflow will gradually become more complex and start requiring test authors to learn and follow a
long list of project specific flags.

integration-tests/tests/data/jobs/compression/compression_jobs.py
```diff
@@ -0,0 +1,19 @@
+"""Compression job definitions for CLP package tests."""
```
Bill-hbrhbr: This file probably shouldn’t live under `data/` semantically because it's python code, not data.
Also, I think I previously mentioned keeping test data closer to the pytest implementation files that use them, rather
than accumulating everything under a single `data/` directory. Testing data we actually commit to the codebase shouldn't be very huge in size, so a bit of repetition across test folders is acceptable if we decide to reuse anything for different tests.

----------------------

Right now we keep all pytest implementations under `integration-tests/tests` (at the same level as `conftest.py`). As the integration test suite grows, I think it would be helpful to introduce more levels of categorization under `tests/` to better organize test files and testing data.
I can see that more tests will likely start needing accompanying test data (whether that’s raw logs or query+result pairs.)

integration-tests/tests/fixtures/package_instance.py
```diff
@@ -15,14 +15,24 @@
 
 
 @pytest.fixture
-def fixt_package_instance(fixt_package_config: PackageConfig) -> Iterator[PackageInstance]:
+def fixt_package_instance(
```
Bill-hbrhbr: What do you think of adding a `scope` to this fixture? I know your design is to put everything under
```
def test_clp_package(fixt_package_instance: PackageInstance) -> None:
```
but if you add `scope=module` for example, you would be able to write multiple
```
def test_*(fixt_package_instance: PackageInstance) -> None:
```
in a python file, and they would use share the same CLP package instance. Then, you can use pytest name filtering instead of implementing `--job-name-contains` yourself. It just feels like reinventing the wheel.

integration-tests/tests/utils/config.py
```diff
@@ -139,6 +170,9 @@ class PackageConfig:
     #: The base port from which all ports for the components are derived.
     base_port: int
 
+    #: The list of jobs that this package will run during the test.
+    package_job_list: PackageJobList | None
+
```
Bill-hbrhbr: I’d prefer splitting this out of `PackageConfig`, since it isn’t related to spinning up or tearing down the
package instance.

Bill-hbrhbr (commented):
These are some ideas we’ve touched on before but never really circled back to or explored in depth. I think I have a good understanding of your overall design, and it makes sense to me. That said, I wonder if we could lean toward a more modular approach rather than bundling all package related tests together.

In particular, I’m a bit hesitant about adding more pytest parser options. They work, but I worry they might make the testing setup harder to reason about over time. I’d be interested in discussing whether we can achieve the same goals using pytest’s existing mechanisms while keeping things simpler and more modular.

integration-tests/tests/package_tests/clp_json/data/json-multifile/README.md
quinntaylormitchell: The dataset won't be evolving, it's just meant to be a static thing. I don't see any reason why we would be adding new entries. We will need to add other separate datasets (e.g., `text-singlefile`, `json-singlefile`, etc.). When would you anticipate us updating the dataset?

I agree with adding the field for the authoritative timestamp, that makes sense. As for the other stuff, I just generated the logs with an LLM, and I don't see how the schema would be valuable. Could you explain your thought process on those latter two?

integration-tests/tests/utils/config.py
quinntaylormitchell: When you say "split out", what do you mean? Do you think it should go in a different class, or not be attached to any class?

integration-tests/tests/data/jobs/compression/compression_jobs.py
quinntaylormitchell: I agree re. more levels of categorization under `tests/`; I think that's a good solution. Then I'll put the job descriptors under that.

integration-tests/tests/fixtures/package_instance.py
quinntaylormitchell: I think that this would be good in the case of a more modular design; see my offline message.

integration-tests/tests/conftest.py
quinntaylormitchell: To be honest, it feels like a bit of catastrophization to talk about "keep adding custom options"; these are two concrete additions. But with a more modular overall design approach, they may not be necessary; let's check back in with this after I attempt the redesign I talked about in my offline message.

integration-tests/tests/utils/utils.py
```diff
@@ -10,6 +10,22 @@
...
+    :param directory:
+    """
+    if not directory.exists():
+        return
+
+    for item in directory.iterdir():
+        if item.is_file() or item.is_symlink():
+            item.unlink()
+        elif item.is_dir():
+            shutil.rmtree(item)
```
Bill-hbrhbr: Let's make this section into a standalone helper because I think it'd be useful in general (and replace what I currently have in [`unlink()`](https://github.com/y-scope/clp/blob/de9fc089fa1ea8204f528a3706f251cf5648614f/integration-tests/tests/utils/utils.py#L107)). I think we can name this helper something like `remove_path`.
Also i think to be exhaustive (`is_file` doesn't include special files like pipes/sockets), you can use the following logic:
1. check `is_symlink`
2. check `is_dir`
3. everything else is a regular file or special file

Bill-hbrhbr: And then we have this `clear_directory` call the helper inside `iterdir()`

integration-tests/tests/utils/utils.py
```diff
@@ -10,6 +10,22 @@
 import yaml
 
 
+def clean_directory(directory: Path) -> None:
```
Bill-hbrhbr:
```suggestion
def clear_directory(directory: Path) -> None:
```
`clear` fits the intention better than `clean`

integration-tests/tests/package_tests/clp_json/test_clp_json.py
```diff
@@ -54,20 +57,47 @@ def test_clp_json_startup(fixt_package_instance: PackageInstance) -> None:
...
+    # Clear archives before compressing.
+    package_test_config = fixt_package_instance.package_test_config
+    package_path_config = package_test_config.path_config
+    package_path_config.clear_package_archives()
+
+    # Compress a dataset.
+    compression_job = PackageCompressionJob(
+        path_to_dataset=(
+            resolve_path_env_var("INTEGRATION_TESTS_DIR")
+            / "tests"
```
Bill-hbrhbr: Let's make this path a part of [`IntegrationTestPathConfig`](https://github.com/y-scope/clp/blob/de9fc089fa1ea8204f528a3706f251cf5648614f/integration-tests/tests/fixtures/path_configs.py#L20). We can name it `test_scripts_dir` or `test_source_dir`. In hindsight, should've named the other attribute `test_build_dir`. We can fix it in a later PR.

integration-tests/tests/package_tests/clp_json/test_clp_json.py
```diff
@@ -54,20 +57,47 @@ def test_clp_json_startup(fixt_package_instance: PackageInstance) -> None:
...
+            / "package_tests"
+            / "clp_json"
+            / "sample_datasets"
```
Bill-hbrhbr: Instead of `sample_datasets`, wondering if we can simplify the directory names to `data`.

integration-tests/.pytest.ini
```diff
@@ -11,6 +11,7 @@ env =
     D:CLP_BUILD_DIR=../build
     D:CLP_CORE_BINS_DIR=../build/core
     D:CLP_PACKAGE_DIR=../build/clp-package
+    D:INTEGRATION_TESTS_DIR=./
```
Bill-hbrhbr:
```suggestion
    D:INTEGRATION_TESTS_PROJECT_ROOT=./
```
The old name feels ambiguous

integration-tests/tests/utils/package_utils.py
```diff
@@ -44,3 +47,38 @@ def stop_clp_package(package_test_config: PackageTestConfig) -> None:
     ]
     # fmt: on
     run_and_assert(stop_cmd, timeout=DEFAULT_CMD_TIMEOUT_SECONDS)
+
+
+def run_package_compression_script(
```
Bill-hbrhbr, line 52: This function doesn't feel like it's unique to running the compression job script. This function can be generalized to running any binary/script that takes option flags first and then positional arguments in the end.
Of course there are also cases where these two scenarios are interleaved. E.g.
```
clp-s x --ordered archive decomp
```
where the binary is first followed by pos arg, then by an option, and eventually by pos args again. Anyways let's not account for this situation now.
What I suggest is moving the core of this function into a new helper called `build_command`, and call `run_and_assert` externally. That means you should also generalize `PackageCompressionJob` to `CommandArgumentBuilder` or something that fits, and `compression_job.path_to_dataset` actually becomes the one and only positional argument for `CommandArgumentBuilder`.
`CommandArgumentBuilder` shouldn’t be a dataclass since it will hold construction logic, and it can have an API called `build()` (instead of the `build_command()` as I suggested before).

integration-tests/tests/utils/config.py
```diff
@@ -119,6 +127,35 @@ def stop_script_path(self) -> Path:
...
+        return self.clp_package_dir / "sbin" / "decompress.sh"
+
+    def clear_package_archives(self) -> None:
+        """Removes the contents of `clp-package/var/data/archives`."""
+        archives_dir = self.clp_package_dir / "var" / "data" / "archives"
+        clean_directory(archives_dir)
+
+
+@dataclass(frozen=True)
+class PackageCompressionJob:
```
Bill-hbrhbr, line 162: See discussion about `CommandArgumentBuilder`

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -104,3 +104,48 @@ def _validate_running_mode_correct(package_instance: PackageInstance) -> None:
...
+
+        # Verify content equality.
+        path_to_dataset = compression_job.path_to_dataset
+        output_path = decompression_dir / path_to_dataset.relative_to(path_to_dataset.anchor)
+
+        if not is_dir_tree_content_equal(path_to_dataset, output_path):
+            err_msg = f"Mismatch between clp input {path_to_dataset} and output {output_path}."
+            pytest.fail(err_msg)
+
+        unlink(decompression_dir)
```
Bill-hbrhbr: Use the new `remove_path` helper I suggested and also remember to remove before the test run to ensure no leftover data from previous runs (in case of test failures).

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -104,3 +104,48 @@ def _validate_running_mode_correct(package_instance: PackageInstance) -> None:
...
+        # Decompress the contents of `clp-package/var/data/archives`.
+        path_config = package_test_config.path_config
+        decompress_script_path = path_config.decompress_script_path
+        decompression_dir = path_config.package_decompression_dir
+        temp_config_file_path = package_test_config.temp_config_file_path
+        decompress_cmd = [
+            str(decompress_script_path),
+            "--config",
+            str(temp_config_file_path),
+            "x",
```
Bill-hbrhbr: I know I haven't done this myself (partially because we just introduced editable installed packages recently), but let's replace this with [`EXTRACT_FILE_CMD`](https://github.com/y-scope/clp/blob/de9fc089fa1ea8204f528a3706f251cf5648614f/components/clp-package-utils/clp_package_utils/general.py#L48)

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -104,3 +104,48 @@ def _validate_running_mode_correct(package_instance: PackageInstance) -> None:
 
     if not compare_mode_signatures(intended_config, running_config):
         pytest.fail("Mode mismatch: running configuration does not match intended configuration.")
+
+
+def verify_compression(
```
```suggestion
def verify_package_compression(
```
Also this seems more like part of the testing code instead of code that can be reused. I don't know how I feel about putting this in `asserting_utils` instead of directly in `test_clp_text` and `test_clp_json`

integration-tests/.pytest.ini
quinntaylormitchell: agreed

integration-tests/tests/utils/utils.py
quinntaylormitchell: agreed

integration-tests/tests/utils/package_utils.py
quinntaylormitchell: In concept I agree with you. I don't see that there is any real reason to generalize it within the context of this PR though; how about I leave it as-is for now, and then when I put up the search PRs, I can generalize it to `build_command` then?

integration-tests/tests/utils/config.py
quinntaylormitchell: replied

integration-tests/tests/utils/asserting_utils.py
quinntaylormitchell: agreed

integration-tests/tests/utils/asserting_utils.py
quinntaylormitchell: agree with name change.

I do see the value in having a version of this function that is specific to each mode to avoid the multiplexing that's going on here, but that means that each future mode module (e.g., `test_clp_presto`, `test_clp_json_s3`, etc.) will need one too. I think that keeping it here is ultimately the better call, especially as the number of modes will grow faster than the number of ways we verify compression (e.g., the future `test_clp_presto` and `test_clp_json_s3` will both use the same verification workflow as `clp-json`).

integration-tests/tests/package_tests/clp_json/test_clp_json.py
quinntaylormitchell: Considering the fact that the package test code doesn't use `IntegrationTestPathConfig` anywhere at the moment, it seems a bit weird to introduce that class throughout the package test code just to store one path (cause then we'd have to have `IntegrationTestPathConfig` as a parameter in every package test function). My next thought is "What if we stored it in `PackagePathConfig` instead", but that doesn't really seem right either.

I think ultimately what we need is an ironed-out (standardized) way to store paths that are relevant to all integration tests (maybe that's what should get stored in `pytest.ini`?), and then a different way to store paths for integration tests of different types (maybe `PackagePathConfig` for package, `BinaryPathConfig` for binaries, something like that).

Let's talk about this offline; it doesn't seem like it's a blocker for this PR in particular though.

integration-tests/tests/package_tests/clp_json/test_clp_json.py
quinntaylormitchell: sure

integration-tests/tests/utils/utils.py
quinntaylormitchell: agreed

integration-tests/tests/utils/asserting_utils.py
quinntaylormitchell: agreed

integration-tests/tests/package_tests/clp_json/test_clp_json.py
Bill-hbrhbr: My approach to path definitions in the integration tests project is as follows:

1. `pytest.ini` defines environment variable to path mappings with sensible defaults, while still allowing users to override locations as needed. For example, `CLP_CORE_BINS_DIR` typically points to `clp/build/core`, but users may choose a different build location such as `clp/components/core/build`. New additions to `pytest.ini` should not contain any paths that can be fully derived from existing ones (i.e. it's not possible for them to have custom locations).

2. All paths defined in `pytest.ini` are resolved centrally in `path_configs.py` and exposed through session-scoped fixtures.

3. Derived paths, such as the path to the `clp-s` binary or to the package compression script `compress.py`, are provided as `@property` attributes on these session-scoped fixture classes.

junhaoliao:
@quinntaylormitchell so sorry for getting back to this late. the code changes lgtm. once we merge main into this PR and resolve the conflicts, i believe the PR will be good to go

PR #1802, opened 2025-12-18
feat(integration-tests): Clean up logging and error reporting in integration tests.

quinntaylormitchell:
# Description

This PR modifies integration test run output. It provides a minimal amount of information via the command line, and stores more verbose information in a test log directory (unique; one per test run). The output of each subprocess is stored within its own file in said directory.

# Checklist

  breaking change.

# Validation performed

Ran `uv run pytest -m 'package'`; all tests pass and information is reported accurately.

---
  - https://app.asana.com/0/0/1212861692182928

## Summary by CodeRabbit

* **Tests**
  * Per-run test log files with millisecond timestamps, richer file output, and header showing log locations.
  * Colourised test name display and labeled test parameter ids for clearer reporting.
  * Tests now write command output to log files and use logging-backed validation for clearer failure context.

* **Chores**
  * Refined test runner output options for concise tracebacks and suppressed console capture.
  * Standardised startup/teardown logging and error messages across test utilities.

### DIFF
--- a/components/core/src/clp_s/SchemaReader.cpp
+++ b/components/core/src/clp_s/SchemaReader.cpp
@@ -5,6 +5,7 @@
 
 #include <clp_s/archive_constants.hpp>
 #include <clp_s/BufferViewReader.hpp>
+#include <clp_s/ErrorCode.hpp>
 #include <clp_s/Schema.hpp>
 
 namespace clp_s {
@@ -224,76 +225,101 @@ bool SchemaReader::get_next_message(std::string& message) {
         message += '\n';
     }
 
-    m_cur_message++;
+    ++m_cur_message;
     return true;
 }
 
-bool SchemaReader::get_next_message(std::string& message, FilterClass* filter) {
-    while (m_cur_message < m_num_messages) {
-        if (false == filter->filter(m_cur_message)) {
-            m_cur_message++;
-            continue;
-        }
-
-        if (m_should_marshal_records) {
-            if (false == m_serializer_initialized) {
-                initialize_serializer();
-            }
-            message = generate_json_string(m_cur_message);
-
-            if (message.back() != '\n') {
-                message += '\n';
-            }
-        }
-
-        m_cur_message++;
-        return true;
+bool SchemaReader::get_next_message(std::string& message, FilterClass& filter) {
+    while (m_cur_message < m_num_messages && false == filter.filter(m_cur_message)) {
+        ++m_cur_message;
     }
 
-    return false;
+    if (m_cur_message >= m_num_messages) {
+        return false;
+    }
+
+    if (m_should_marshal_records) {
+        if (false == m_serializer_initialized) {
+            initialize_serializer();
+        }
+        message = generate_json_string(m_cur_message);
+
+        if (message.back() != '\n') {
+            message += '\n';
+        }
+    }
+
+    ++m_cur_message;
+    return true;
+}
+
+bool SchemaReader::get_next_message_with_metadata(
+        std::string& message,
+        epochtime_t& timestamp,
+        int64_t& log_event_idx
+) {
+    if (m_cur_message >= m_num_messages) {
+        return false;
+    }
+
+    if (m_should_marshal_records) {
+        if (false == m_serializer_initialized) {
+            initialize_serializer();
+        }
+        message = generate_json_string(m_cur_message);
+
+        if (message.back() != '\n') {
+            message += '\n';
+        }
+    }
+
+    timestamp = m_get_timestamp();
+    log_event_idx = get_next_log_event_idx();
+
+    ++m_cur_message;
+    return true;
 }
 
 bool SchemaReader::get_next_message_with_metadata(
         std::string& message,
         epochtime_t& timestamp,
         int64_t& log_event_idx,
-        FilterClass* filter
+        FilterClass& filter
 ) {
     // TODO: If we already get max_num_results messages, we can skip messages
     // with the timestamp less than the smallest timestamp in the priority queue
-    while (m_cur_message < m_num_messages) {
-        if (false == filter->filter(m_cur_message)) {
-            m_cur_message++;
-            continue;
-        }
-
-        if (m_should_marshal_records) {
-            if (false == m_serializer_initialized) {
-                initialize_serializer();
-            }
-            message = generate_json_string(m_cur_message);
-
-            if (message.back() != '\n') {
-                message += '\n';
-            }
-        }
-
-        timestamp = m_get_timestamp();
-        log_event_idx = get_next_log_event_idx();
-
-        m_cur_message++;
-        return true;
+    while (m_cur_message < m_num_messages && false == filter.filter(m_cur_message)) {
+        ++m_cur_message;
     }
 
-    return false;
+    if (m_cur_message >= m_num_messages) {
+        return false;
+    }
+
+    if (m_should_marshal_records) {
+        if (false == m_serializer_initialized) {
+            initialize_serializer();
+        }
+        message = generate_json_string(m_cur_message);
+
+        if (message.back() != '\n') {
+            message += '\n';
+        }
+    }
+
+    timestamp = m_get_timestamp();
+    log_event_idx = get_next_log_event_idx();
+
+    ++m_cur_message;
+    return true;
 }
 
-void SchemaReader::initialize_filter(FilterClass* filter) {
-    filter->init(this, m_columns);
+void SchemaReader::initialize_filter(FilterClass& filter) {
+    filter.init(this, m_columns);
 }
 
-void SchemaReader::initialize_filter_with_column_map(FilterClass* filter) {
-    filter->init(this, m_column_map);
+void SchemaReader::initialize_filter_with_column_map(FilterClass& filter) {
+    filter.init(this, m_column_map);
 }
 
 void SchemaReader::generate_local_tree(int32_t global_id) {
--- a/components/core/src/clp_s/SchemaReader.hpp
+++ b/components/core/src/clp_s/SchemaReader.hpp
@@ -203,7 +203,7 @@ public:
     [[nodiscard]] auto generate_json_string(uint64_t message_index) -> std::string;
 
     /**
-     * Gets next message
+     * Gets the next message
      * @param message
      * @return true if there is a next message
      */
@@ -215,7 +215,20 @@ public:
      * @param filter
      * @return true if there is a next message
      */
-    bool get_next_message(std::string& message, FilterClass* filter);
+    bool get_next_message(std::string& message, FilterClass& filter);
+
+    /**
+     * Gets the next message as well as its timestamp and log event index.
+     * @param message
+     * @param timestamp
+     * @param log_event_idx
+     * @return true if there is a next message
+     */
+    bool get_next_message_with_metadata(
+            std::string& message,
+            epochtime_t& timestamp,
+            int64_t& log_event_idx
+    );
 
     /**
      * Gets the next message matching a filter as well as its timestamp and log event index.
@@ -229,21 +242,21 @@ public:
             std::string& message,
             epochtime_t& timestamp,
             int64_t& log_event_idx,
-            FilterClass* filter
+            FilterClass& filter
     );
 
     /**
      * Initializes the filter
      * @param filter
      */
-    void initialize_filter(FilterClass* filter);
+    void initialize_filter(FilterClass& filter);
 
     /**
      * Initializes the filter with a column map.
      * Note: the column map only contains the ordered columns in a schema.
      * @param filter
      */
-    void initialize_filter_with_column_map(FilterClass* filter);
+    void initialize_filter_with_column_map(FilterClass& filter);
 
     /**
      * Initializes all internal data structures required to serialize records.
--- a/components/core/src/clp_s/search/Output.cpp
+++ b/components/core/src/clp_s/search/Output.cpp
@@ -100,7 +100,7 @@ bool Output::filter() {
                 m_output_handler->should_output_metadata(),
                 m_should_marshal_records
         );
-        reader.initialize_filter(&m_query_runner);
+        reader.initialize_filter(m_query_runner);
 
         if (m_output_handler->should_output_metadata()) {
             epochtime_t timestamp{};
@@ -109,13 +109,13 @@ bool Output::filter() {
                     message,
                     timestamp,
                     log_event_idx,
-                    &m_query_runner
+                    m_query_runner
             ))
             {
                 m_output_handler->write(message, timestamp, archive_id, log_event_idx);
             }
         } else {
-            while (reader.get_next_message(message, &m_query_runner)) {
+            while (reader.get_next_message(message, m_query_runner)) {
                 m_output_handler->write(message);
             }
         }
--- a/components/core/tests/test-clp_s-delta-encode-log-order.cpp
+++ b/components/core/tests/test-clp_s-delta-encode-log-order.cpp
@@ -97,7 +97,7 @@ TEST_CASE("clp-s-delta-encode-log-order", "[clp-s][delta-encode-log-order]") {
     REQUIRE(cNumEntries == schema_reader->get_num_messages());
 
     SimpleFilterClass simple_filter_class;
-    schema_reader->initialize_filter(&simple_filter_class);
+    schema_reader->initialize_filter(simple_filter_class);
     clp_s::BaseColumnReader* log_event_idx_reader{nullptr};
     for (auto* column_reader : simple_filter_class.get_column_readers()) {
         if (log_event_idx_node_id == column_reader->get_id()) {


integration-tests/tests/utils/package_utils.py
```diff
@@ -23,10 +32,17 @@ def start_clp_package(package_config: PackageConfig) -> None:
...
     # fmt: on
-    run_and_assert(start_cmd, timeout=DEFAULT_CMD_TIMEOUT_SECONDS)
+
+    try:
+        run_and_assert(request, start_cmd, timeout=DEFAULT_CMD_TIMEOUT_SECONDS)
+    except SubprocessError:
+        mode_name = package_config.mode_name
+        err_msg = f"The {mode_name} package failed to start."
+        logger.error(construct_log_err_msg(err_msg))
+        pytest.fail(err_msg)
```
junhaoliao: **[Score: 100] Dead exception handling** - This `except SubprocessError` block will never execute because `run_and_assert()` catches `CalledProcessError` and `TimeoutExpired` internally and converts them to `pytest.fail()` calls. The `SubprocessError` exception will never propagate to callers.

See `run_and_assert` exception handling:
https://github.com/y-scope/clp/blob/0562f48bebfa0ab6623a6d31a39bc7c989fce6f8/integration-tests/tests/utils/asserting_utils.py#L38-L45

integration-tests/tests/utils/package_utils.py
```diff
@@ -43,4 +59,11 @@ def stop_clp_package(package_config: PackageConfig) -> None:
...
     # fmt: on
-    run_and_assert(stop_cmd, timeout=DEFAULT_CMD_TIMEOUT_SECONDS)
+
+    try:
+        run_and_assert(request, stop_cmd, timeout=DEFAULT_CMD_TIMEOUT_SECONDS)
+    except SubprocessError:
+        mode_name = package_config.mode_name
+        err_msg = f"The {mode_name} package failed to stop."
+        logger.error(construct_log_err_msg(err_msg))
+        pytest.fail(err_msg)
```
junhaoliao: **[Score: 100] Dead exception handling** - Same issue: `except SubprocessError` will never execute because `run_and_assert()` converts subprocess errors to `pytest.fail()` calls internally.

integration-tests/tests/fixtures/package_instance.py
```diff
@@ -25,15 +30,10 @@ def fixt_package_instance(fixt_package_config: PackageConfig) -> Iterator[Packag
     mode_name = fixt_package_config.mode_name
 
     try:
-        start_clp_package(fixt_package_config)
+        logger.info("Starting the '%s' package...", mode_name)
```
junhaoliao: **[Score: 85] Removed helpful error message** - The `--base-port` guidance message was removed here. While the original `except RuntimeError` was also unreachable, the helpful user-facing message about port conflicts is now entirely gone.

Consider either:
1. Fixing the exception handling to actually catch errors from `start_clp_package`, or
2. Adding the port guidance to the generic error message in `start_clp_package`

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -26,15 +29,22 @@ def run_and_assert(cmd: list[str], **kwargs: Any) -> subprocess.CompletedProcess
...
-        pytest.fail(f"Command failed: {' '.join(cmd)}: {e}")
-    except subprocess.TimeoutExpired as e:
-        pytest.fail(f"Command timed out: {' '.join(cmd)}: {e}")
-    return proc
+    log_file_path = Path(request.config.getini("log_file_path"))
+    with log_file_path.open("ab") as log_file:
+        try:
+            log_debug_msg = f"Now running command: {cmd}"
+            logger.debug(log_debug_msg)
+            proc = subprocess.run(cmd, stdout=log_file, stderr=log_file, check=True, **kwargs)
```
junhaoliao: **[Score: 25] Potential race condition with log file** - The log file is opened in append-binary mode and held open during subprocess execution. If multiple tests ever run concurrently, they could interleave output. However, pytest-xdist is not currently configured, so this is theoretical.

*Low confidence - parallel testing not enabled in this project.*

integration-tests/tests/conftest.py
```diff
@@ -23,3 +29,46 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         default="55000",
         help="Base port for CLP package integration tests.",
     )
+
+    test_run_id = str(uuid.uuid4())[-4:]
```
junhaoliao: **[Score: 25] UUID generated at config time** - `test_run_id` is generated when `pytest_addoption` is called. If parallel workers were ever spawned, each would get a separate UUID (which is actually fine). The concern about shared UUIDs is theoretical and unlikely in practice.

*Low confidence - no parallel testing configured.*

integration-tests/tests/utils/package_utils.py
```diff
@@ -1,12 +1,21 @@
...
+from tests.utils.logging_utils import construct_log_err_msg
+
+logger = logging.getLogger(__name__)
+
 
 DEFAULT_CMD_TIMEOUT_SECONDS = 120.0
 
 
-def start_clp_package(package_config: PackageConfig) -> None:
+def start_clp_package(request: pytest.FixtureRequest, package_config: PackageConfig) -> None:
```
junhaoliao: **[Score: 50] Missing param + docstring issues** -
1. The `request: pytest.FixtureRequest` parameter is not documented in the docstring
2. Docstring says `:raise: Propagates run_and_assert's errors` but the implementation catches `SubprocessError` and calls `pytest.fail()` instead - it doesn't propagate errors

*Moderate confidence - documentation issues, no functional impact.*

integration-tests/tests/utils/package_utils.py
```diff
@@ -23,10 +32,17 @@ def start_clp_package(package_config: PackageConfig) -> None:
...
+        run_and_assert(request, start_cmd, timeout=DEFAULT_CMD_TIMEOUT_SECONDS)
+    except SubprocessError:
+        mode_name = package_config.mode_name
+        err_msg = f"The {mode_name} package failed to start."
+        logger.error(construct_log_err_msg(err_msg))
+        pytest.fail(err_msg)
 
 
-def stop_clp_package(package_config: PackageConfig) -> None:
+def stop_clp_package(request: pytest.FixtureRequest, package_config: PackageConfig) -> None:
```
junhaoliao: **[Score: 50] Missing param + docstring issues** - Same issues as `start_clp_package`: missing `request` param documentation, and docstring claims error propagation but implementation catches and converts to `pytest.fail()`.

*Moderate confidence - documentation issues.*

.gitignore
```diff
@@ -6,3 +6,4 @@ __pycache__/
 .task/
 build/
 dist/
+__*
```
junhaoliao: what are we trying to exclude here?

integration-tests/.pytest.ini
```diff
@@ -1,20 +1,26 @@
 [pytest]
 addopts =
-    --capture=no
     --code-highlight=yes
     --color=yes
-    -rA
+    --show-capture=no
+    --tb=short
     --strict-config
     --strict-markers
```
junhaoliao, line 7: nit
```suggestion
    --show-capture=no
    --strict-config
    --strict-markers
    --tb=short
```

integration-tests/.pytest.ini
```diff
@@ -1,20 +1,26 @@
...
 env =
     D:CLP_BUILD_DIR=../build
     D:CLP_CORE_BINS_DIR=../build/core
     D:CLP_PACKAGE_DIR=../build/clp-package
+
 log_cli = True
-log_cli_date_format = %Y-%m-%d %H:%M:%S,%f
-log_cli_format = %(name)s %(asctime)s [%(levelname)s] %(message)s
 log_cli_level = INFO
+log_cli_format = %(levelname)-8s %(asctime)-25s %(message)s
```
junhaoliao: any reason to change this? i thought we were trying to match references like https://github.com/y-scope/clp/blob/5de8bc99ed948cf439d33c12880ebad98d5def70/components/clp-py-utils/clp_py_utils/initialize-spider-db.py#L24

integration-tests/.pytest.ini
```diff
@@ -1,20 +1,26 @@
...
+log_cli_date_format = %Y-%m-%d %H:%M:%S
+
+log_file_mode = a
+log_file_level = DEBUG
+log_file_format = %(levelname)-8s %(asctime)-30s [%(name)s:%(filename)s:%(lineno)d]: %(message)s
```
junhaoliao: any reason to use a different format string between the cli and the log file?

integration-tests/.pytest.ini
```diff
@@ -1,20 +1,26 @@
...
     D:CLP_PACKAGE_DIR=../build/clp-package
+
 log_cli = True
-log_cli_date_format = %Y-%m-%d %H:%M:%S,%f
-log_cli_format = %(name)s %(asctime)s [%(levelname)s] %(message)s
 log_cli_level = INFO
+log_cli_format = %(levelname)-8s %(asctime)-25s %(message)s
+log_cli_date_format = %Y-%m-%d %H:%M:%S
+
+log_file_mode = a
```
junhaoliao, line 19: instead of appending to the same file, can isolate the logs from different test runs instead? e.g., use `set_log_path` with the default `log_file_mode = w`. see the test code at https://github.com/pytest-dev/pytest/pull/4752 for example

integration-tests/.pytest.ini
```diff
@@ -1,20 +1,26 @@
...
+log_file_level = DEBUG
+log_file_format = %(levelname)-8s %(asctime)-30s [%(name)s:%(filename)s:%(lineno)d]: %(message)s
+log_file_date_format = %Y-%m-%d %H:%M:%S,%f
```
junhaoliao: ditto - any reason to use a different date fmt in cli and files?

integration-tests/tests/conftest.py
```diff
@@ -23,3 +29,46 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+    log_file_name = Path("__pytest_logs") / f"testrun_{test_run_id}.log"
+    parser.addini(
+        "log_file_path",
+        help="Path to the log file for this test.",
+        type="paths",
+        default=log_file_name,
+    )
```
junhaoliao: any reason not to call `set_log_path` in a hook like below ones?

integration-tests/tests/fixtures/package_instance.py
```diff
@@ -25,15 +30,10 @@ def fixt_package_instance(fixt_package_config: PackageConfig) -> Iterator[Packag
...
+        start_clp_package(request, fixt_package_config)
         instance = PackageInstance(package_config=fixt_package_config)
         yield instance
-    except RuntimeError:
-        base_port = fixt_package_config.base_port
-        pytest.fail(
-            f"Failed to start the {mode_name} package. This could mean that one of the ports"
```
junhaoliao, line 37: maybe i wasn't reading carefully - why we removed this?

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -12,12 +12,15 @@
...
 from tests.utils.docker_utils import list_running_services_in_compose_project
+from tests.utils.logging_utils import construct_log_err_msg
 from tests.utils.utils import load_yaml_to_dict
 
 logger = logging.getLogger(__name__)
 
 
-def run_and_assert(cmd: list[str], **kwargs: Any) -> subprocess.CompletedProcess[Any]:
+def run_and_assert(
+    request: pytest.FixtureRequest, cmd: list[str], **kwargs: Any
```
junhaoliao: update below docstring for the `param`

integration-tests/tests/utils/package_utils.py
junhaoliao: param docstring for `request`

integration-tests/tests/utils/package_utils.py
```diff
@@ -43,4 +59,11 @@ def stop_clp_package(package_config: PackageConfig) -> None:
         "--config", str(temp_config_file_path),
     ]
     # fmt: on
-    run_and_assert(stop_cmd, timeout=DEFAULT_CMD_TIMEOUT_SECONDS)
+
+    try:
+        run_and_assert(request, stop_cmd, timeout=DEFAULT_CMD_TIMEOUT_SECONDS)
+    except SubprocessError:
```
junhaoliao: in what case would this be thrown here? i thought we would have caught this in `run_and_assert`? like also if there are other `SubprocessError` than `CalledProcessError` that might be thrown but uncaught in `run_and_assert`, why don't we modify `run_and_assert` to handle such in the first place?

integration-tests/tests/utils/package_utils.py
```diff
@@ -23,10 +32,17 @@ def start_clp_package(package_config: PackageConfig) -> None:
         "--config", str(temp_config_file_path),
     ]
     # fmt: on
-    run_and_assert(start_cmd, timeout=DEFAULT_CMD_TIMEOUT_SECONDS)
+
+    try:
+        run_and_assert(request, start_cmd, timeout=DEFAULT_CMD_TIMEOUT_SECONDS)
+    except SubprocessError:
```
junhaoliao: ditto - 

> in what case would this be thrown here? i thought we would have caught this in `run_and_assert`? like also if there are other `SubprocessError` than `CalledProcessError` that might be thrown but uncaught in `run_and_assert`, why don't we modify `run_and_assert` to handle such in the first place?

integration-tests/tests/utils/package_utils.py
junhaoliao: param docstring for `request`

junhaoliao (changes requested):
with the understanding that this is a draft, i just took a quick look

integration-tests/tests/fixtures/package_instance.py
quinntaylormitchell: Because `start_clp_package` never raises a `RuntimeError` (`run_and_assert` raises a `SubprocessError`). I can move the bit about the ports to the section that catches the `SubprocessError` in `start_clp_package`, but with the full output always being captured in the log file, I think that it will be more helpful to just tell the dev to check the test log (as we are now doing in this PR).

integration-tests/tests/utils/asserting_utils.py
quinntaylormitchell: do you mean the `request`? if so, right on, thank you

integration-tests/tests/utils/package_utils.py
quinntaylormitchell: right on, thanks

integration-tests/tests/utils/package_utils.py
quinntaylormitchell: right on, thanks

integration-tests/tests/utils/package_utils.py
quinntaylormitchell: ditto my other comment -

integration-tests/tests/utils/package_utils.py
quinntaylormitchell: Yeah that's a good point. See my message that I sent you offline.

integration-tests/.pytest.ini
quinntaylormitchell: agreed

.gitignore
quinntaylormitchell: The `__pytest_logs` directory that stores the log file for each test run.

That being said, right now the `__pytest_logs` directory gets created/stored under the `clp/integration-tests` directory. Do you think it might be better to have it under the `build/integration-tests` directory instead (used for other integration test stuff as well)? Then we wouldn't have to modify this file.

integration-tests/tests/conftest.py
quinntaylormitchell: it's a bit convoluted because trying to customize pytest's logging to do exactly what we want it to do is not very straightforward, but basically

1. `log_file_path` is an environment variable, and it stores the log file path in a way that we can access during a test run;
2.  `set_log_path` sets the path internally for pytest's internal handling of the logging messages produced by the test code. afaik, that path can't be accessed during the test.

That's the best way that I could come up with to store all of the pytest output _and_ all the output from subprocesses in a log file, and then only selectively report the information to the dev via the command line.

integration-tests/.pytest.ini
quinntaylormitchell: My only reason was aesthetics, cause this is the format of the logs that will be reported to the user via the command line. If there is some reason we want to match references like you describe above, I can change it; I just wasn't aware we were doing that

integration-tests/.pytest.ini
quinntaylormitchell: If this isn't here, the test output file will contain all of the pytest logging messages followed by all of the output from the various subprocesses called during the test run, like this:

```text
INFO 2025-01-06 event1
INFO 2025-01-06 event2
INFO 2025-01-06 event3

<output following event 1>
<output following event 2>
<output following event 3>
```

It seems better to me to have everything written to the log file in order, so that it looks like this:

```text
INFO 2025-01-06 event1
<output following event 1>
INFO 2025-01-06 event2
<output following event 2>
INFO 2025-01-06 event3
<output following event 3>
```

The logs from different test runs are already isolated by the code in `conftest.py`.

integration-tests/.pytest.ini
quinntaylormitchell: Just to provide more information to the dev; I don't want to crowd the cli output, especially for a successful (no failure) test run. I think it's safe to assume that the main reason a dev would be looking at the test log would be if something fails, and in the event of failure, they'll need more information.

integration-tests/.pytest.ini
quinntaylormitchell: ditto as above

.gitignore
junhaoliao: yup i agree we can move to `build/`

in general we want to be more explicit about what to ignore, so if we keep the folder in the current location, it would be better to just write `__pytest_logs`. That said, it really depends on which part of the paths we want to be fuzzy about. in general wildcards like `path/*/subpath` and `*.log` are fine in my opinion

junhaoliao: (some response to a previous discussion somehow refuses to be shown in the "Files changed" tab)

duplicating the comment here for visibility - 

> yup i agree we can move to build/
> 
> in general we want to be more explicit about what to ignore, so if we keep the folder in the current location, it would be better to just write __pytest_logs. That said, it really depends on which part of the paths we want to be fuzzy about. in general wildcards like path/*/subpath and *.log are fine in my opinion

integration-tests/tests/conftest.py
```diff
@@ -23,3 +29,52 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
     )
+
+    # Sets up a unique log file for this test run, and stores the path to the file.
+    test_run_id = str(uuid.uuid4())[-4:]
+    log_file_path = Path("__pytest_logs") / f"testrun_{test_run_id}.log"
+    parser.addini(
+        "log_file_path",
+        help="Path to the log file for this test.",
+        type="paths",
+        default=log_file_path,
```
junhaoliao: https://docs.pytest.org/en/stable/reference/reference.html#pytest.Parser.addini:~:text=paths%3A%20a%20list%20of%20pathlib%2EPath%2C%20separated%20as%20in%20a%20shell

seems like with `type="paths"`, the parser returns a **list** of Path objects when we actually specify any value in the config file?

shall we use `[log_file_path]` or change to `type="string"` which might make more sense?

integration-tests/tests/conftest.py
```diff
@@ -23,3 +29,52 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+    )
```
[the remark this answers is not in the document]
junhaoliao: i agree with the rabbit about the `mkdir(parents=True)` comment especially now we want to output the logs into build/

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -12,29 +12,26 @@
...
-
-    try:
-        proc = subprocess.run(cmd, check=True, **kwargs)
-    except subprocess.CalledProcessError as e:
-        pytest.fail(f"Command failed: {' '.join(cmd)}: {e}")
-    except subprocess.TimeoutExpired as e:
-        pytest.fail(f"Command timed out: {' '.join(cmd)}: {e}")
-    return proc
+    log_file_path = Path(request.config.getini("log_file_path"))
+    with log_file_path.open("ab") as log_file:
```
junhaoliao: the pytest logging plugin writes to the same file handle, right? any concurrency concerns?

I understand the outputs are usually flushed only on newlines, but this feels a bit fragile.

would it be better to use just Python loggers to log the output of the sub processes once any content is available?

something like opening a (combined) stream then iterating line-by-line should work?
```
proc = subprocess.Popen(
    cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, **kwargs
)
for line in proc.stdout:
    logger.debug("%s", line.decode(errors="replace").rstrip())
proc.wait()
```

integration-tests/tests/utils/asserting_utils.py
```diff
@@ -12,29 +12,26 @@
...
 from tests.utils.config import PackageInstance
 from tests.utils.docker_utils import list_running_services_in_compose_project
+from tests.utils.logging_utils import construct_log_err_msg
 from tests.utils.utils import load_yaml_to_dict
 
 logger = logging.getLogger(__name__)
 
 
-def run_and_assert(cmd: list[str], **kwargs: Any) -> subprocess.CompletedProcess[Any]:
+def run_and_assert(request: pytest.FixtureRequest, cmd: list[str], **kwargs: Any) -> None:
```
junhaoliao: we no longer assets right? would `run_subprocess` or `run_and_log_to_file` be better?

integration-tests/tests/utils/asserting_utils.py
junhaoliao: btw, it might be worth using a prefix in the logs otherwise we easily get mixed contents from different commands if we ever call this multiple times concurrently.

e.g.
`logger.debug("[%s] %s", cmd, line.decode...)`

integration-tests/tests/conftest.py
```diff
@@ -23,3 +29,52 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         default="55000",
         help="Base port for CLP package integration tests.",
     )
+
+    # Sets up a unique log file for this test run, and stores the path to the file.
+    test_run_id = str(uuid.uuid4())[-4:]
```
junhaoliao: benefit of using shorter UUIDs? note simply cropping the string harms the distribution of the random string (though i know this anti-pattern already exists in some of our code). would including a date string in the log file name be easier to be looked up by any developers?

integration-tests/tests/utils/package_utils.py
```diff
@@ -23,10 +34,19 @@ def start_clp_package(package_test_config: PackageTestConfig) -> None:
...
```
junhaoliao: `run_and_assert` propagates `subprocess.run`'s exceptions ->

> [The most common exception raised is OSError.](https://docs.python.org/3/library/subprocess.html#subprocess.Popen:~:text=The%20most%20common%20exception%20raised%20is%20OSError%2E)

so probably we should write this instead?

```suggestion
    except (SubprocessError, OSError):
```

integration-tests/tests/utils/package_utils.py
junhaoliao: ditto - [The most common exception raised is OSError.](https://docs.python.org/3/library/subprocess.html#subprocess.Popen:~:text=The%20most%20common%20exception%20raised%20is%20OSError%2E)

integration-tests/pyproject.toml
```diff
@@ -43,6 +43,9 @@ show_error_end = true
 [tool.ruff]
 extend = "../tools/yscope-dev-utils/exports/lint-configs/python/ruff.toml"
 
+[tool.ruff.lint]
+ignore = ["TRY400"]
```
junhaoliao: can we avoid this global override and if really needed use inline `# noqa: TRY400` instead? btw, where do we need such suppressions?

integration-tests/tests/utils/package_utils.py
junhaoliao: maybe worth including the exception (`e.g. str(e)`) in the logs as well

integration-tests/tests/test_identity_transformation.py
```diff
@@ -140,6 +140,7 @@ def test_clp_s_identity_transform(
 
 
 def _clp_s_compress_and_decompress(
+    request: pytest.FixtureRequest,
     clp_core_path_config: ClpCorePathConfig,
     test_paths: CompressionTestPathConfig,
 ) -> None:
```
junhaoliao, line 145: just found this guy does have a docstring. shall we add one or was it too trivial?

integration-tests/.pytest.ini
```diff
@@ -1,20 +1,27 @@
...
 log_cli = True
-log_cli_date_format = %Y-%m-%d %H:%M:%S,%f
-log_cli_format = %(name)s %(asctime)s [%(levelname)s] %(message)s
 log_cli_level = INFO
+log_cli_format = %(asctime)s.%(msecs)03d %(levelname)s %(message)s
+log_cli_date_format = %Y-%m-%d %H:%M:%S
+
+log_file_mode = a
+log_file_level = DEBUG
+log_file_format = %(asctime)s.%(msecs)03d %(levelname)s [%(name)s:%(filename)s:%(lineno)d]: %(message)s
```
junhaoliao: continuing from some of our offline discussion - this format is not consistent the loggers config used by the rest of the project. that said, i'm fine keeping whatever's been proposed like you said it improves readability in the consoles. 

in the long run, we should add infra for test CIs, where we should allow picking a JSON logger / some format string that's compatible with GH actions' problem matchers https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md instead

integration-tests/tests/conftest.py
```diff
@@ -23,3 +29,52 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+
+
+def pytest_itemcollected(item: pytest.Item) -> None:
+    """
+    Prettifies the name of the test for output purposes.
+
+    :param item:
+    """
+    item._nodeid = f"{BOLD}{BLUE}{item.name}{RESET}"  # noqa: SLF001
+
```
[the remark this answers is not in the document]
junhaoliao: i think the rabbit's comment about "Use item.nodeid instead of item.name to preserve test identity." is valid

integration-tests/tests/utils/package_utils.py
```diff
@@ -1,12 +1,23 @@
...
+
+logger = logging.getLogger(__name__)
+
 
 DEFAULT_CMD_TIMEOUT_SECONDS = 120.0
 
 
-def start_clp_package(package_test_config: PackageTestConfig) -> None:
+def start_clp_package(
+    request: pytest.FixtureRequest, package_test_config: PackageTestConfig
```
junhaoliao: missing docstring

integration-tests/tests/utils/package_utils.py
```diff
@@ -23,10 +34,19 @@ def start_clp_package(package_test_config: PackageTestConfig) -> None:
...
+    except SubprocessError:
+        mode_name = package_test_config.mode_config.mode_name
+        err_msg = f"The '{mode_name}' package failed to start."
+        logger.error(construct_log_err_msg(err_msg))
+        pytest.fail(err_msg)
 
 
-def stop_clp_package(package_test_config: PackageTestConfig) -> None:
+def stop_clp_package(
+    request: pytest.FixtureRequest, package_test_config: PackageTestConfig
```
junhaoliao: missing docstring

integration-tests/tests/utils/logging_utils.py
```diff
@@ -0,0 +1,14 @@
+"""Utilities for logging during the test run."""
+
+BLUE = "\033[34m"
+BOLD = "\033[1m"
+GREEN = "\033[32m"
+RESET = "\033[0m"
+
+
+def construct_log_err_msg(err_msg: str) -> str:
```
junhaoliao: missing param / return dosctring

integration-tests/tests/conftest.py
```diff
@@ -23,3 +29,52 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+    )
+
+
+def pytest_itemcollected(item: pytest.Item) -> None:
+    """
+    Prettifies the name of the test for output purposes.
+
+    :param item:
+    """
+    item._nodeid = f"{BOLD}{BLUE}{item.name}{RESET}"  # noqa: SLF001
```
junhaoliao: just a heads up -

The ANSI constants are terminal (`print("...")`) formatting concerns, not logging (`logger.info("...")`) concerns.

We usually don't want to mix them together, though here the complexity is fine to be tolerated as i'm fine keeping it if you think it helps visualization. Normally, if we want facier colorful outputs, we should look into creating a custom LoggerAdapter like explained by https://docs.python.org/3/howto/logging-cookbook.html

(anyways, don't bother until it becomes more complex like we don't only add colorful prefixes but also colorful postfixes, lol

integration-tests/tests/utils/asserting_utils.py
quinntaylormitchell: sure, I like `run_and_log_to_file`.

it does beg the question though: should this function be in a different file now that it isn't asserting anything?

integration-tests/tests/utils/asserting_utils.py
quinntaylormitchell: I don't really have any concurrency concerns here, because `pytest` only runs sequentially (the only way to run it in parallel is to use plugins that we don't use). With that in mind, it seems to me like there is no way that the logging file would be accessed by multiple processes at the same time, and by the same token, there's no way that this would get called multiple times concurrently

integration-tests/tests/utils/logging_utils.py
quinntaylormitchell: thank u

integration-tests/tests/utils/package_utils.py
quinntaylormitchell: right on

integration-tests/tests/utils/package_utils.py
quinntaylormitchell: agree with suggestion to include `OSError`; the traceback of the exception will be reported in the test summary

integration-tests/tests/utils/package_utils.py
quinntaylormitchell: right on

integration-tests/tests/conftest.py
quinntaylormitchell: I was just going off of similar logic to how we assign a random instance ID to the package when it runs, but I could put in a date string instead. that would help when looking back at multiple log files for multiple test runs, good idea

integration-tests/tests/conftest.py
quinntaylormitchell: I'll just store it as a string

integration-tests/tests/conftest.py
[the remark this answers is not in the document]
quinntaylormitchell: I'm pretty sure that the existence of the parent directory will be handled by the logger; I've tried it from scratch and there is no `FileNotFoundError` when the log file is first written

integration-tests/tests/conftest.py
quinntaylormitchell: thanks, I'll keep that in mind

integration-tests/tests/conftest.py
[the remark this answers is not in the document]
quinntaylormitchell: sure

integration-tests/tests/test_identity_transformation.py
quinntaylormitchell: it should probably have one; maybe once you approve of the overarching approach of this PR then I'll go through the rest of the code and change the logging/docstrings (I stuck to just the package testing code for now)

integration-tests/.pytest.ini
quinntaylormitchell: agreed

integration-tests/pyproject.toml
quinntaylormitchell: `TRY400` forces you to use `logger.exception()` instead of `logger.error()` inside an exception handler. I wanted to avoid using `logger.exception()` because it prints the whole traceback in the log message, and it looks cluttered (not to mention the exception is printed in the testing summary). There are four lines in the code that raise `TRY400`; should I use `# noqa: TRY400` or keep it as is?

.gitignore
quinntaylormitchell: moved to `build/`

integration-tests/tests/conftest.py
```diff
@@ -11,15 +17,122 @@
...
+    :param items:
+    :return: A formatted string listing each test name, or a warning if no tests were collected.
+    """
+    report: str = ""
+    if len(items) == 0:
+        report = f"{BOLD}No tests match the specified parameters.{RESET}\n"
+    else:
+        report = f"{BOLD}The following tests will run:{RESET}\n"
+        for item in items:
+            report += f"\t{BOLD}{BLUE}{item.name}{RESET}\n"
```
junhaoliao:
```suggestion
            report += f"\t{BOLD}{BLUE}{item.nodeid}{RESET}\n"
```

integration-tests/tests/conftest.py
```diff
@@ -11,15 +17,122 @@
...
+
+
+def pytest_itemcollected(item: pytest.Item) -> None:
+    """
+    Applies ANSI bold and blue formatting to each collected test's node ID, for the purposes of
+    test output readability.
+
+    :param item:
+    """
+    item._nodeid = f"{BOLD}{BLUE}{item.name}{RESET}"  # noqa: SLF001
```
junhaoliao:
```suggestion
    item._nodeid = f"{BOLD}{BLUE}{item.nodeid}{RESET}"  # noqa: SLF001
```

integration-tests/tests/utils/package_utils.py
```diff
@@ -1,20 +1,18 @@
...
 from tests.utils.config import (
     PackageCompressionJob,
     PackageTestConfig,
 )
-
-DEFAULT_CMD_TIMEOUT_SECONDS = 120.0
+from tests.utils.subprocess_utils import run_and_log_subprocess
 
 
 def start_clp_package(package_test_config: PackageTestConfig) -> None:
```
junhaoliao, line 10: note in `package_instnace.py` we still catch `RuntimeError` which is outdated?

if we don't catch OSError and TimeoutExpired in run_and_log_subprocess() and rethrow with RuntimeError, i believe we want to catch those two in `package_instnace.py`

integration-tests/tests/conftest.py
quinntaylormitchell: agreed

integration-tests/tests/conftest.py
quinntaylormitchell: agreed

integration-tests/tests/utils/package_utils.py
quinntaylormitchell: You're right, we don't need to catch `RuntimeError` in `package_instance.py` anymore. I've removed that check.

For our purposes here, I feel that we don't need to catch `OSError` or `TimeoutExpired`, because either of those errors would indicate conditions or circumstances that are beyond the scope of the testing system. We can just let them be handled automatically; for example, I just forced a timeout in a testing subprocess by setting `timeout=0.1`, and the timeout was reported to the user in the `ERROR` summary at the end of the test.

What do you think?

junhaoliao: sure sgtm

PR #1808, opened 2025-12-19
fix(ci): Prevent stale Docker image cache on self-hosted runners (fixes #1807).

### LINKED ISSUE #1807: Intermittent workflow job failures due to stale Docker image cache on self-hosted runners
## Bug

Multiple jobs in the `clp-artifact-build` workflow intermittently fail due to stale Docker image caches on self-hosted runners. Observed failures include, but are not limited to:

- `ubuntu-jammy-lint`: missing `python3-venv`
- `centos-stream-9-static-linked-bins`: https://github.com/y-scope/clp/actions/runs/20347055261/job/58462788977

Example error from lint job:
```
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt install python3.10-venv
```

This occurs despite `python3-venv` being included in the Docker image since commit 36140bf4.

- Intermittent failure: https://github.com/y-scope/clp/actions/runs/20347055261/job/58463692356
- Expected behavior: https://github.com/y-scope/clp/actions/runs/20347907187/job/58465817371

**Affected hosts:** baker7, baker16 (newly provisioned GitHub runners from existing hosts that were previously used for development, thus had outdated images in the cache)

**Root cause:** The `run-on-image` action uses `docker run` without `--pull=always`. On self-hosted runners, Docker caches images locally and reuses stale cached images even when the remote `:main` tag has been updated. Different runners have different cached versions, causing intermittent failures depending on which host runs the job.

This issue could also manifest in the future whenever dependency images are upgraded, as runners with cached older versions would not automatically pull the updated images.

**Expected behavior:** The workflow should always use the latest published Docker image.

### Concurrency considerations

Using `--pull=always` means the workflow always pulls the latest `:main` image from the registry. One might think this could create a race condition where a concurrent workflow updates the image mid-run. However, this is prevented by the workflow's concurrency rule:

```yaml
concurrency:
  group: "${{github.workflow}}-${{github.ref}}"
  cancel-in-progress: "${{github.ref != 'refs/heads/main'}}"
```

- For the `main` branch: `cancel-in-progress: false`, so workflow runs queue and execute sequentially
- For other branches: each branch has its own concurrency group, and `cancel-in-progress: true` cancels older runs

This ensures that on `main`, only one `clp-artifact-build` workflow runs at a time, preventing race conditions when publishing/pulling images.

### Future improvement: content-addressed image tags

Currently, images are tagged with the branch name only (`:main`) via `clp-core-build-containers/action.yaml`:
```yaml
tags: |
  type=raw,value=${{steps.get_image_props.outputs.branch}}
```

A future improvement could compute a hash of the files that generate each dependency image and tag images with that hash. Jobs would then reference images by their content hash rather than `:main`, enabling:
- Parallel workflow runs without race conditions
- Immutable, reproducible builds
- Cache invalidation only when image-generating files actually change

## CLP version

Main branch at commit a61596df

## Environment

- GitHub Actions self-hosted runners (baker7, baker16)
- Docker with local image caching
- Images: `ghcr.io/y-scope/clp/clp-core-dependencies-x86-*:main`

## Reproduction steps

1. A self-hosted runner has an older version of a `clp-core-dependencies-x86-*:main` image in its local Docker cache (e.g., from prior development use or before a dependency upgrade)
2. The corresponding deps-image job is skipped (no changes to image-related files)
3. A build/lint job runs with `use_published_image: true`
4. `docker run` uses the stale cached image instead of pulling the latest
5. The job fails due to missing packages or outdated dependencies

**Proposed fix:** Add `--pull=always` flag to `docker run` in `.github/actions/run-on-image/action.yaml` when `use_published_image == 'true'`.

jackluo923:
# Description

Add `--pull=always` flag to `docker run` in the `run-on-image` action when using published images. This ensures self-hosted runners always pull the latest Docker image from the registry instead of using potentially stale cached versions.

Resolves #1807

# Checklist

  breaking change.

# Validation performed

This issue is intermittent and difficult to reproduce since we don't control GitHub's job scheduling across self-hosted runners. However, the root cause is clear (stale cached Docker images) and the solution is straightforward (`--pull=always` ensures the latest image is used).

## Summary by CodeRabbit

* **Chores**
  * Optimized image-pulling behaviour in build workflows to ensure published images are always up-to-date.

### DIFF
--- a/tools/deployment/package-helm/Chart.yaml
+++ b/tools/deployment/package-helm/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: "v2"
 name: "clp"
-version: "0.1.2-dev.9"
+version: "0.1.2-dev.10"
 description: "A Helm chart for CLP's (Compressed Log Processor) package deployment"
 type: "application"
 appVersion: "0.7.1-dev"
--- a/tools/deployment/package-helm/templates/query-scheduler-deployment.yaml
+++ b/tools/deployment/package-helm/templates/query-scheduler-deployment.yaml
@@ -70,6 +70,13 @@ spec:
           ports:
             - name: "query-scheduler"
               containerPort: 7000
+          readinessProbe:
+            {{- include "clp.readinessProbeTimings" . | nindent 12 }}
+            tcpSocket: &query-scheduler-health-check
+              port: 7000
+          livenessProbe:
+            {{- include "clp.livenessProbeTimings" . | nindent 12 }}
+            tcpSocket: *query-scheduler-health-check
           volumeMounts:
             - name: "config"
               mountPath: "/etc/clp-config.yaml"


.github/actions/run-on-image/action.yaml
```diff
@@ -45,6 +45,7 @@ runs:
     # `--env npm_config_cache` overrides the default `/.npm` to avoid permission issues.
     - run: >-
         docker run
+        ${{inputs.use_published_image == 'true' && '--pull=always' || ''}}
```
junhaoliao, line 48: not a bug, but worth noting an empty space is added when inputs.use_published_image is false

junhaoliao (approved):
for the title, how about:
```
fix(ci): Prevent stale Docker image cache on self-hosted runners (fixes #1807).
```

PR #1824, opened 2025-12-19
fix(ci): Improve workflow efficiency by optimizing runner usage and eliminating duplicate runs

jackluo923:
## Description

Improve CI workflow efficiency with two changes:

### 1. Use GitHub-hosted runner for lightweight path filtering job
The `filter-relevant-changes` job only performs git operations to detect changed paths. Moving it to a GitHub-hosted runner avoids consuming self-hosted runner resources needed for heavier build jobs.

### 2. Use skip-duplicate-actions to prevent duplicate workflow runs
When pushing to a PR branch, both `push` and `pull_request` events trigger, causing duplicate workflow runs. This PR adds [fkirc/skip-duplicate-actions](https://github.com/fkirc/skip-duplicate-actions) to detect and skip duplicate runs.

**How it works:**
- When both `push` and `pull_request` trigger on the same commit, the duplicate is skipped
- Scheduled builds and manual triggers always run (`do_not_skip`)
- Downstream jobs are skipped automatically when `filter-relevant-changes` is skipped

**Why this action:**
- Used by 9,234+ repositories including [haskell-language-server](https://github.com/haskell/haskell-language-server), [CNCF TOC](https://github.com/cncf/toc), and [MegaLinter](https://github.com/oxsecurity/megalinter)
- Note: step-security/skip-duplicate-actions is a fork that requires a paid subscription; we use the original fkirc version which is free

## Validation performed

- Verified YAML syntax
- Reviewed fkirc/skip-duplicate-actions documentation

## Summary by CodeRabbit

* **Chores**
  * Added a reusable check to detect and skip duplicate workflow runs, reducing unnecessary builds.
  * Made downstream pipeline steps conditional so jobs run only when appropriate via the new skip output.
  * Reworked runner configuration to prefer self-hosted when available with a GitHub-hosted fallback for reliability.
  * Clarified workflow comments and alignment for maintainability.

---
  - https://app.asana.com/0/0/1212854499651194

### DIFF
--- a/.github/workflows/clp-artifact-build.yaml
+++ b/.github/workflows/clp-artifact-build.yaml
@@ -38,8 +38,10 @@ jobs:
           # Skip duplicate runs for the same content (e.g., when both push and pull_request trigger)
           concurrent_skipping: "same_content"
           skip_after_successful_duplicate: "true"
-          # Always run for scheduled builds and manual triggers
-          do_not_skip: '["schedule", "workflow_dispatch"]'
+          # Always run for scheduled builds and manual triggers.
+          # NOTE: This is a JSON array string (required by the action), with escaped quotes for
+          # yamllint compliance.
+          do_not_skip: "[\"schedule\", \"workflow_dispatch\"]"
 
   filter-relevant-changes:
     name: "filter-relevant-changes"
--- a/.github/workflows/clp-artifact-build.yaml
+++ b/.github/workflows/clp-artifact-build.yaml
@@ -8,8 +8,6 @@ on:
       - "components/core/tools/scripts/lib_install/macos/**"
       - "docs/**"
   push:
-    branches:
-      - "main"
     paths-ignore: *ignored_paths
   schedule:
     # Run daily at 00:15 UTC (the 15 is to avoid periods of high load)
@@ -28,8 +26,25 @@ concurrency:
   cancel-in-progress: "${{github.ref != 'refs/heads/main'}}"
 
 jobs:
+  skip-duplicate-check:
+    name: "skip-duplicate-check"
+    runs-on: "ubuntu-24.04"
+    outputs:
+      should_skip: "${{steps.skip_check.outputs.should_skip}}"
+    steps:
+      - id: "skip_check"
+        uses: "step-security/skip-duplicate-actions@v5"
+        with:
+          # Skip duplicate runs for the same content (e.g., when both push and pull_request trigger)
+          concurrent_skipping: "same_content"
+          skip_after_successful_duplicate: "true"
+          # Always run for scheduled builds and manual triggers
+          do_not_skip: '["schedule", "workflow_dispatch"]'
+
   filter-relevant-changes:
     name: "filter-relevant-changes"
+    needs: "skip-duplicate-check"
+    if: "needs.skip-duplicate-check.outputs.should_skip != 'true'"
     # This job only performs git operations to detect changed paths, so we use a GitHub-hosted
     # runner to avoid consuming self-hosted runner resources needed for heavier build jobs.
     runs-on: "ubuntu-24.04"
--- a/.github/workflows/clp-artifact-build.yaml
+++ b/.github/workflows/clp-artifact-build.yaml
@@ -33,7 +33,7 @@ jobs:
       should_skip: "${{steps.skip_check.outputs.should_skip}}"
     steps:
       - id: "skip_check"
-        uses: "step-security/skip-duplicate-actions@v5"
+        uses: "fkirc/skip-duplicate-actions@v5"
         with:
           # Skip duplicate runs for the same content (e.g., when both push and pull_request trigger)
           concurrent_skipping: "same_content"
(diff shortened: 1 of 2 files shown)


.github/workflows/clp-artifact-build.yaml
```diff
@@ -8,6 +8,8 @@ on:
       - "components/core/tools/scripts/lib_install/macos/**"
       - "docs/**"
   push:
+    branches:
+      - "main"
```
junhaoliao: Does this mean workflows won't run on non-main branches in forks unless a pull request is opened? Though, i guess users can manually `workflow_dispatch`?

junhaoliao: @kirkrodrigues is this something you mentioned that we want to avoid? though i vaguely remember seeing something similar in our workflows previously

junhaoliao: assuming the triggering condition change is ok, we may prefer
```suggestion
    branches: ["main"]
```

jackluo923: Yah. CI only triggers when pushed to their fork's main branch after this change if they don't submit a PR. Otherwise, if they PR via their own branch into our repo, our CI would be triggered. I can revert this if that's not something we want.

kirkrodrigues: > @kirkrodrigues is this something you mentioned that we want to avoid? though i vaguely remember seeing something similar in our workflows previously

Correct, I would prefer we didn't disable workflow runs on pushes to branches. I feel like if we want to get rid of the duplication, we should employ a different solution of some sort.

.github/workflows/skip-duplicate-check.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - id: "skip_check"
+        uses: "fkirc/skip-duplicate-actions@v5"
+        with:
+          # Skip duplicate runs for the same content (e.g., when both push and pull_request trigger)
+          concurrent_skipping: "same_content"
+          skip_after_successful_duplicate: "true"
+          # Always run for scheduled builds and manual triggers.
+          # NOTE: This is a JSON array string (required by the action), with escaped quotes for
+          # yamllint compliance.
+          do_not_skip: "[\"schedule\", \"workflow_dispatch\"]"
```
junhaoliao, line 40:
```suggestion
          do_not_skip: >-
            ["schedule", "workflow_dispatch"]
```

.github/workflows/skip-duplicate-check.yaml
```diff
@@ -0,0 +1,40 @@
+name: "skip-duplicate-check"
+
+# Reusable workflow to skip duplicate runs when both push and pull_request events trigger on the
```
junhaoliao, line 3: in this docstring, can we also explain how this workflow interacts / may conflict with the caller workflow's `concurrency.cancel-in-progress` settings?

.github/workflows/skip-duplicate-check.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  check:
+    name: "skip-duplicate-check"
+    runs-on: "ubuntu-24.04"
+    outputs:
+      should_skip: "${{steps.skip_check.outputs.should_skip}}"
+    steps:
+      - id: "skip_check"
+        uses: "fkirc/skip-duplicate-actions@v5"
```
junhaoliao, line 32:
```suggestion
        uses: "fkirc/skip-duplicate-actions@f75f66ce1886f00957d99748a42c724f4330bdcf"  # v5.3.1
```

for security / reproducibility concerns, in case the publisher republish the version with different contents

.github/workflows/skip-duplicate-check.yaml
```diff
@@ -0,0 +1,40 @@
...
+        value: "${{jobs.check.outputs.should_skip}}"
+
+jobs:
+  check:
+    name: "skip-duplicate-check"
+    runs-on: "ubuntu-24.04"
+    outputs:
+      should_skip: "${{steps.skip_check.outputs.should_skip}}"
+    steps:
+      - id: "skip_check"
```
junhaoliao, line 31: Can we omit this `id` specification? If we really want to avoid showing "uses: "fkirc/skip-duplicate-actions" as a workflow step on the workflow status page, we could specify something like `name: "Evaluate whether this workflow run should be skipped"` instead

.github/workflows/clp-artifact-build.yaml
```diff
@@ -26,14 +26,16 @@ concurrency:
   cancel-in-progress: "${{github.ref != 'refs/heads/main'}}"
 
 jobs:
+  skip-duplicate-check:
```
junhaoliao, line 29: maybe renaming this as `check-duplicate-run`? like other steps place the verb at the beginning of the ids

junhaoliao (changes requested):
lgtm. mostly just nitpicking except that one request about clarification on `concurrency:` conflicts

Issue #883, opened 2025-05-07
Add docs for `--tags` option in compress.sh / search.sh for CLP-Text


junhaoliao:
### Request

In the CLP package, the compress.sh script accepts a `--tags` option which is a comma-separated list of tags to apply to the compressed archives, and the search.sh script can also accept a `--tags` option to filter for specified tags which were applied during compression. However, those are not documented in the current user guide in the docs.
Similarly, the `--file-path` option in search.sh can be documented.

### Possible implementation

See above.

gibber9809:
No longer necessary after #1842

PR #1842, opened 2026-01-06
feat(clp-package)!: Remove the archive tags feature.

gibber9809:
# Description

The archive tags feature is being removed as part of the ongoing effort to refactor and stabilize the archive metadata representation in the package.

The general rationale behind removing this feature is that it has some overlap with the datasets feature, and the remainder of the use-cases for archive tags should be covered by our plans to allow users to associate structured metadata with each file they ingest into the package. Since we also know that our major open source users aren't relying on archive tags, the simplest solution is to drop support for this feature.

# Checklist

  breaking change.

# Validation performed

* Validated that package build succesfully
* Validated that basic package flow works correctly by testing that:
  * JSONL file can be compressed with/without dataset from cli
  * Data can be searched as expected from cli and webui with/without dataset
  * Tag options are no longer presented in the cli

## Summary by CodeRabbit

* **Refactor**
  * Removed tags support across compression, ingestion and search flows.
  * Dropped tags CLI options and tag propagation through tasks and scheduler.
  * Removed tags from job/output/search configuration schemas and public APIs.
  * Eliminated tags metadata tables and tag persistence/update logic.
  * Updated tests, UI enums and default configs to reflect tags removal.

---
  - https://app.asana.com/0/0/1212888451178058
  - https://app.asana.com/0/0/1212888453877648

### DIFF
--- a/components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
+++ b/components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
@@ -614,7 +614,8 @@ def dispatch_job_and_update_db(
         db_conn, new_job, target_archives, clp_metadata_db_conn_params, results_cache_uri
     )
     start_time = datetime.datetime.now()
-    new_job.start_time = start_time
+    if new_job.start_time is None:
+        new_job.start_time = start_time
     set_job_or_task_status(
         db_conn,
         QUERY_JOBS_TABLE_NAME,


gibber9809:
> The Package changes lgtm. a few minor sugguestions:
> 
> 1. it does seem there are some more references in the webui that should be updated. let's also remove these references to "tags":
>    
>    * https://github.com/y-scope/clp/blob/ff6f76a5419034937368a3ed0a60c9a7b461baab/components/webui/common/src/schemas/compression.ts#L117-L118
>    * https://github.com/y-scope/clp/blob/ff6f76a5419034937368a3ed0a60c9a7b461baab/components/webui/server/src/routes/api/compress/index.ts#L33
> 2. since this change is BREAKING, are there migration strategies for the metadata DB?
> 3. i believe #883 can be closed once this PR is merged.
> 4. for the commit message, i think a `feat!` or a `refactor!` is probably more appropriate?

1. Sure, will do
2. Migration strategy is to drop archive tags table followed by tags table for each dataset -- already confirmed with our largest open source user that they aren't using this feature, so best to drop support for it now before other users start relying on it. Mid to long term, the planned feature for per-ingested-file structured metadata will offer a superset of the functionality of this tags feature.
3. Yep, looks like it
4. Let's go with `refactor!`

junhaoliao (approved):
for the title, how about:
```
refactor(clp-package)!: Remove archive tags feature.
```

PR #1843, opened 2026-01-06
feat(integration-tests): Modularize package tests with per-mode directories and module-scoped fixtures.

quinntaylormitchell:
# Description

Modularization of the design for the package integration tests will make it easier to work with in the long run. This PR sets up everything needed so that devs will be able to write independent test functions.

# Checklist

  breaking change.

# Validation performed

```bash
uv run pytest -m 'package and startup'
```

Both the `clp-json` and `clp-text` startup tests passed.

## Summary by CodeRabbit

* **Tests**
  * Added integration suites for clp_json and clp_text (startup, compression, search validations).
  * Removed a legacy general startup test in favour of package‑specific test modules.
  * Reworked fixtures to module scope with test‑specific configuration objects for clearer, isolated runs.

* **Chores**
  * Expanded pytest markers (clp_json, clp_text, compression, search, startup) and refined descriptions.
  * Improved test utilities with guarded package‑instance validation and clearer failure messages.

* **Refactor**
  * Split mode vs test configuration and removed an older mode helper API.

---
  - https://app.asana.com/0/0/1212888451178040

### DIFF
--- a/tools/deployment/package-helm/Chart.yaml
+++ b/tools/deployment/package-helm/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: "v2"
 name: "clp"
-version: "0.1.3-dev.1"
+version: "0.1.3-dev.2"
 description: "A Helm chart for CLP's (Compressed Log Processor) package deployment"
 type: "application"
 appVersion: "0.8.1-dev"
--- a/tools/deployment/package-helm/templates/log-ingestor-deployment.yaml
+++ b/tools/deployment/package-helm/templates/log-ingestor-deployment.yaml
@@ -1,4 +1,6 @@
-{{- if .Values.clpConfig.log_ingestor }}
+{{- if and
+  (.Values.clpConfig.log_ingestor)
+  (eq .Values.clpConfig.logs_input.type "s3") }}
 apiVersion: "apps/v1"
 kind: "Deployment"
 metadata:
--- a/tools/deployment/package-helm/templates/log-ingestor-logs-pv.yaml
+++ b/tools/deployment/package-helm/templates/log-ingestor-logs-pv.yaml
@@ -1,4 +1,6 @@
-{{- if .Values.clpConfig.log_ingestor }}
+{{- if and
+  (.Values.clpConfig.log_ingestor)
+  (eq .Values.clpConfig.logs_input.type "s3") }}
 {{- include "clp.createStaticPv" (dict
   "root" .
   "component_category" "log-ingestor"
--- a/tools/deployment/package-helm/templates/log-ingestor-logs-pvc.yaml
+++ b/tools/deployment/package-helm/templates/log-ingestor-logs-pvc.yaml
@@ -1,4 +1,6 @@
-{{- if .Values.clpConfig.log_ingestor }}
+{{- if and
+  (.Values.clpConfig.log_ingestor)
+  (eq .Values.clpConfig.logs_input.type "s3") }}
 {{- include "clp.createPvc" (dict
   "root" .
   "component_category" "log-ingestor"
--- a/tools/deployment/package-helm/templates/log-ingestor-service.yaml
+++ b/tools/deployment/package-helm/templates/log-ingestor-service.yaml
@@ -1,4 +1,6 @@
-{{- if .Values.clpConfig.log_ingestor }}
+{{- if and
+  (.Values.clpConfig.log_ingestor)
+  (eq .Values.clpConfig.logs_input.type "s3") }}
 apiVersion: "v1"
 kind: "Service"
 metadata:
--- a/tools/deployment/package-helm/values.yaml
+++ b/tools/deployment/package-helm/values.yaml
@@ -157,17 +157,17 @@ clpConfig:
   #   logging_level: "INFO"
 
   # log-ingestor config. Currently, the config is applicable only if `logs_input.type` is "s3".
-  log_ingestor: null
-  #   port: 30302
-  #   # The timeout (in seconds) after which the log buffer is flushed for compression if no new
-  #   # input arrives.
-  #   buffer_flush_timeout: 300
-  #   # The log buffer size (in bytes) that triggers a flush for compression.
-  #   buffer_flush_threshold: 268435456  # 256 MiB
-  #   # The capacity of the internal channel used for communication between an ingestion job and the
-  #   # log buffer.
-  #   channel_capacity: 10
-  #   logging_level: "INFO"
+  log_ingestor:
+    port: 30302
+    # The timeout (in seconds) after which the log buffer is flushed for compression if no new input
+    # arrives.
+    buffer_flush_timeout: 300
+    # The log buffer size (in bytes) that triggers a flush for compression.
+    buffer_flush_threshold: 268435456  # 256 MiB
+    # The capacity of the internal channel used for communication between an ingestion job and the
+    # log buffer.
+    channel_capacity: 10
+    logging_level: "INFO"
 
   # Where archives should be output to
   archive_output:


integration-tests/tests/utils/config.py
```diff
@@ -120,66 +120,77 @@ def stop_script_path(self) -> Path:
...
 
-@dataclass(frozen=True)
-class PackageConfig:
-    """Metadata for a specific configuration of the CLP package."""
-
-    #: Path configuration for this package.
-    path_config: PackagePathConfig
+@dataclass
+class PackageModeConfig:
+    """Mode configuration for the CLP package."""
```
Bill-hbrhbr, line 125:
```suggestion
    """Defines a specific CLP package flavor, capturing its mode of operation."""
```
The term `flavor` is taken from the CLP docs here: https://docs.yscope.com/clp/main/dev-docs/building-package.html#build

integration-tests/tests/utils/config.py
```diff
@@ -120,66 +120,77 @@ def stop_script_path(self) -> Path:
...
-    #: The Pydantic representation of a CLP package configuration.
-    clp_config: ClpConfig
 
-    #: The base port from which all ports for the components are derived.
+@dataclass
+class PackageTestConfig:
+    """Metadata for a specific test of the CLP package."""
+
+    #: Path configuration for this package test.
+    path_config: PackagePathConfig
```
Bill-hbrhbr, line 142: Unless we see a `pytest` invocation that uses multiple `PackagePathConfig` instances, I suggest splitting this out of `PackageTestConfig` in a future PR.
At the moment, `fixt_package_path_config` is also scoped to `session`, which implies that only a single instance is used per `pytest` run.
Also, just to confirm: are the compression job list and search job list expected to be added here in the future? Right now this class seems to me a rather empty middle ground between mode config and instance config.

integration-tests/tests/utils/config.py
```diff
@@ -120,66 +120,77 @@ def stop_script_path(self) -> Path:
...
+    #: The base port from which all port assignments are derived.
     base_port: int
 
     def __post_init__(self) -> None:
-        """Write the temporary config file for this package."""
+        """Write the temporary config file for this package test."""
         self._write_temp_config_file()
 
     @property
     def temp_config_file_path(self) -> Path:
```
Bill-hbrhbr, line 155: I may be mistaken, but a temporary config file seems more appropriately associated with a CLP package instance being instantiated. I probably need to look a bit further ahead at how this class is expected to be used.

integration-tests/tests/fixtures/package_test_config.py
```diff
@@ -4,29 +4,24 @@
...
-def fixt_package_config(
+@pytest.fixture(scope="module")
+def fixt_package_test_config(
     request: pytest.FixtureRequest,
     fixt_package_path_config: PackagePathConfig,
-) -> Iterator[PackageConfig]:
+) -> Iterator[PackageTestConfig]:
     """
-    Creates and maintains a PackageConfig object for a specific CLP mode.
+    Creates and maintains a PackageTestConfig object for a specific CLP mode.
```
Bill-hbrhbr: I think it’s worth documenting the initialization procedure for this fixture, in particular that it selects the `PackageModeConfig` via `pytest.mark.parametrize` and `request` param on a per-module basis (since the
fixture has `module` scope). 
We should also document that any fixture depending on this one (for example, `fixt_package_instance`) must have a scope less than or equal to `module`, and that when it is referenced in pytest test function arguments, it will automatically be initialized using this `fixt_package_test_config` fixture, thereby using the same `PackageModeConfig`.

integration-tests/tests/fixtures/package_test_config.py
```diff
@@ -4,29 +4,24 @@
...
 
     :param request:
```
Bill-hbrhbr:
```suggestion
    :param request: Provides access to `PackageModeConfig` selection.
```

integration-tests/tests/fixtures/package_test_config.py
Bill-hbrhbr: I’ve sketched out the idea for the docstring, but it’s not quite ready to use yet. If we can put together a clear explanatory paragraph, it would be helpful to see whether Junhao and Kirk can follow what’s going on without needing to dig into the pytest internals.
This setup uses a more advanced pattern of fixture chaining and indirect parametrization, so the goal would be for the docstring itself to make the initialization flow and configuration selection easy to understand. I can foresee future developers making use of this fixture system in their own tests.

integration-tests/.pytest.ini
```diff
@@ -17,6 +17,11 @@ log_cli_format = %(name)s %(asctime)s [%(levelname)s] %(message)s
 log_cli_level = INFO
 markers =
     clp: mark tests that use the CLP storage engine
+    clp_json: mark tests that use the clp-json package
     clp_s: mark tests that use the CLP-S storage engine
+    clp_text: mark tests that use the clp-text package
```
Bill-hbrhbr, line 22: I personally like `clp_text` and `clp_json` much more than `clp` and `clp_s`. Let's leave a TODO saying that we are going to deprecate the latter in a future PR in favor of the new marker terms.

integration-tests/tests/package_tests/clp_json/test_clp_json.py
```diff
@@ -0,0 +1,88 @@
...
+    component_list=[*CLP_BASE_COMPONENTS, CLP_API_SERVER_COMPONENT],
+)
+
+
+# Pytest markers for this module.
+pytestmark = [
+    pytest.mark.package,
+    pytest.mark.clp_json,
+    pytest.mark.parametrize("fixt_package_test_config", [CLP_JSON_MODE], indirect=True),
+]
```
Bill-hbrhbr, line 40: Ideally we put module-level pytest markers as far up as possible, but I understand it is not possible here due to the reliance on `CLP_JSON_MODE`. 
That being said, `CLP_JSON_MODE` can actually be turned into a session-scoped fixture that returns a `PackageModeConfig` (I don't see its contents altering/get affected by any test executions).
I'm fine with the things are right now, but if you turn `CLP_JSON_MODE` and `CLP_TEXT_MODE` into fixtures, that would make the file cleaner. The only other change you might need is to change `request.param` to `request.get_fixture_value()` during `fixt_package_test_config` initialization.

integration-tests/tests/utils/config.py
```diff
@@ -120,66 +120,77 @@ def stop_script_path(self) -> Path:
...
+    package_test_config: PackageTestConfig
 
     #: The instance ID of the running package.
     clp_instance_id: str = field(init=False, repr=True)
 
     #: The path to the .clp-config.yaml file constructed by the package during spin up.
     shared_config_file_path: Path = field(init=False, repr=True)
 
+    #: Switch indicating whether or not this instance has been validated yet.
+    instance_validated: bool = False
```
Bill-hbrhbr: How do we plan to use this variable beyond the initial stages of package testing?

integration-tests/tests/package_tests/clp_json/__init__.py
```diff
@@ -0,0 +1 @@
+"""Top-level package for clp-json package integration tests."""
```
Bill-hbrhbr: `__init__.py` for the `package_tests` directory level as well?

integration-tests/.pytest.ini
quinntaylormitchell: I think `clp` and `clp_s` are still necessary, as they refer to the binaries. `clp-text` and `clp-json` are terms used to describe the configuration of the package.

integration-tests/tests/utils/config.py
quinntaylormitchell: We want to be able to let devs run each test individually, and so that means that each test function will need a call to `validate_package_instance()` to validate that the package is running correctly for that test run. But if multiple tests are run at the same time, we don't need to validate the same package over and over again. `instance_validated` is set to `True` once that `PackageInstance` has been validated, and then it doesn't need to be validated again (note that the `PackageInstance` fixture now has the `module` scope).

integration-tests/tests/utils/config.py
quinntaylormitchell: iirc we talked about this; `PackageInstance` holds information about an instance that is currently running, while `PackageTestConfig` (old name: `PackageConfig`) holds information that the package needs before it starts. with that in mind, I think this belongs here.

integration-tests/tests/utils/config.py
quinntaylormitchell: 1. Sure, will keep that in mind for a future PR.
2. `fixt_package_path_config` is indeed scoped to `session`.
3. The compression job list and search job list approach isn't necessary if we're taking this modular approach.

integration-tests/tests/utils/config.py
quinntaylormitchell: iirc we talked about this; some of the modes will not be flavors, i.e. `clp-presto` and `clp-json-s3` are not flavors, but they will eventually be modes.

integration-tests/tests/package_tests/clp_json/__init__.py
quinntaylormitchell: afaik I don't think it needs one, because the `package_tests` directory doesn't have any python code files in it (all the code files are in sub-directories of `package_tests`)

integration-tests/tests/fixtures/package_test_config.py
quinntaylormitchell: agreed

integration-tests/tests/fixtures/package_test_config.py
quinntaylormitchell: agreed; I'll write up something and push it. let me know what you think.

We can also write about it in the integration test docs (when they are up).

integration-tests/tests/package_tests/clp_json/test_clp_json.py
quinntaylormitchell: I'll keep an eye on this; I can see how changing `CLP_JSON_MODE` to a session-scoped fixture might be valuable.

integration-tests/tests/utils/config.py
Bill-hbrhbr: Sure I'm good with either terms. Let's keep using `modes` for differentiations.

integration-tests/tests/package_tests/clp_json/__init__.py
Bill-hbrhbr: I remember running into linter issues for this type of situations, but if the current linters don't complain I guess things are good.

integration-tests/.pytest.ini
Bill-hbrhbr: I think this is also where some of the overlap between the concepts of `flavor` and `mode` starts to show up, at least from a naming perspective. But if you limit `clp_json` to package tests only (i.e. excluding binary tests), I agree that the current set up is fine.

integration-tests/tests/utils/config.py
Bill-hbrhbr: I see where this is coming from. From a more rigorous standpoint, we would also need some form of locking to make this fully functional. I’m also not entirely sure when pytest runs tests in parallel by default. For now, I believe our tests are executed sequentially.

integration-tests/tests/package_tests/clp_json/__init__.py
quinntaylormitchell: yeah I've experienced that too, no complaints here though.

integration-tests/tests/utils/config.py
quinntaylormitchell: afaik, `pyetst` always runs sequentially unless you install some other plugins, so I'm not sure locking would be necessary here.

integration-tests/tests/utils/config.py
Bill-hbrhbr: Need @junhaoliao  ’s input here. I don’t think there’s much harm in creating multiple identical temporary config files, even if multiple package instances are spun up using the exact same config. We won’t be running that many test instances in practice. 
More generally, I’m not convinced that we necessarily need to separate `PackageInstance` and `PackageTestConfig` at this point, as I don’t see many concrete benefits from introducing an additional layer of abstraction.

-------------

A good argument to keep them separated would be if we expect to add things to `PackageTestConfig` in the future that won't naturally belong to a running instance.

quinntaylormitchell: if the fixtures are scoped to `module` as they are in this PR, then there won't be multiple identical temporary config files; just one file every time this fixture gets used (which, as we're confining all tests about a given mode to a single module, will be only once per test run).

as for separate `PackageInstance` and `PackageTestConfig`, I'm less convinced than I used to be (i.e., I'm starting to think that they don't actually need to be separate). will keep an eye on that; if we change it then it will be in a future PR.

@junhaoliao what do you think

integration-tests/tests/utils/config.py
```diff
@@ -120,66 +120,77 @@ def stop_script_path(self) -> Path:
...
-    package_config: PackageConfig
+    package_test_config: PackageTestConfig
 
     #: The instance ID of the running package.
     clp_instance_id: str = field(init=False, repr=True)
 
     #: The path to the .clp-config.yaml file constructed by the package during spin up.
     shared_config_file_path: Path = field(init=False, repr=True)
 
+    #: Switch indicating whether or not this instance has been validated yet.
```
Bill-hbrhbr:
```suggestion
    #: Flag indicating whether this package instance has completed all validation checks. When set,
    #: subsequent tests reusing the same instance can skip redundant validation.
```

integration-tests/tests/utils/config.py
Bill-hbrhbr: I get it now. Posted a comment reword suggestion.

integration-tests/tests/fixtures/package_test_config.py
```diff
@@ -0,0 +1,50 @@
...
+    request: pytest.FixtureRequest,
+    fixt_package_path_config: PackagePathConfig,
+) -> Iterator[PackageTestConfig]:
+    """
+    Creates and maintains a PackageTestConfig object for a specific CLP mode.
+
+    This fixture is initialized on a per-module basis using `request.param`. Test code that uses
+    this fixture should set this parameter with `pytest.mark.parametrize()`.
+
+    Any fixture that depends on this one must have a scope less than or equal to `module`.
```
Bill-hbrhbr:
```suggestion
    This fixture is initialized from a `PackageModeConfig` provided via `request.param`. Tests that
    use this fixture, or any fixture that depends on it, should apply marker generated by
    `pytest.mark.parametrize` to select which package mode to run for the current test config.
    For exmaple:
    
    pytest.mark.parametrize(
        "fixt_package_test_config",
        [CLP_JSON_MODE],
        indirect=True,
    )

    Because this fixture has `module` scope, all tests in the same module share a single
    `PackageTestConfig` instance. Tests in different modules use different configs and therefore
    spin up separate package instances. The `PackageInstance` fixture, or other fixture depending on
    this one, must have a scope less than or equal to `module`.
```

quinntaylormitchell: I feel like an explanation of this verbosity belongs better in the docs, doesn't it?

Bill-hbrhbr: I agree. We just need to document this somewhere; otherwise, I know I’d probably forget how it works after not looking at the code for a while.

integration-tests/tests/utils/config.py
```diff
@@ -120,66 +120,78 @@ def stop_script_path(self) -> Path:
         return self.clp_package_dir / "sbin" / "stop-clp.sh"
 
 
-@dataclass(frozen=True)
-class PackageConfig:
-    """Metadata for a specific configuration of the CLP package."""
-
-    #: Path configuration for this package.
-    path_config: PackagePathConfig
+@dataclass
```
junhaoliao: possible to make it `frozen=True`? we might need to make `component_list` a tuple instead of the list though

integration-tests/tests/fixtures/package_test_config.py
junhaoliao: on the contrary, i think the docstring can be more concise, as in it should avoid any educational details that can be found via official pytest docs / known by anyone who's familiar with module-level fixtures. (i mean, it's certainly not someone without the knowledge would understand lol. i also just learned more about the related concepts via docs. i personally find it quite Pythonic and intuitive.)

how about:

```
    Creates and maintains a module-level PackageTestConfig object for a specific CLP mode. For
    efficiency, group all tests for a given mode in the same module.

    :param request: Provides `PackageModeConfig` via `request.param`.
```

note:
1. how passing the PackageModeConfig via `[]` is reflected in the updated docstring for param `request`
2. the rationale of the module-level fixture is now explained: "For efficiency, group all tests for a given mode in the same module." or if we want to be more explicit, "the fixture with the specified mode is cached within each module".

integration-tests/tests/utils/config.py
junhaoliao: i'm still a bit confused about the scope of validity of this flag. it sounds like:
1. `fixt_package_instance` is created per module. Each module contains different tests
2. Every test in the module seems to be calling `validate_package_instance()` to validate the package is healthy.
3. However, any subsequent call to `validate_package_instance()` will always return True if any previous test has run `validate_package_instance()` and it returned True.

What if, some previous test crashed the package instance for whatever reason. Then if the subsequent tests trust this flag and proceed to the test logic, we can be getting false positives / failed test results with inaccurate attribution to the direct cause?

if my understanding is correct, i think we should consider these:
1. If the validations are not so heavy (at least it doesn't seem heavy to me at the moment), is it possible to remove caching? (If not possible to remove caching, for record keeping, please respond with the time the validations take so we better understand the overhead.)
2. Otherwise, is it possible to drop some health checks to make the validations lighter weight?
3. If not possible to perform validation on every test and also not possible to drop certain checks, i bet we have to accept the risks of inaccurate cause attributions / false positives among tests in a module, as mentioned in the analysis. Then we should really document such risks in the docstring of `validate_package_instance()`.

integration-tests/tests/package_tests/clp_json/test_clp_json.py
```diff
@@ -0,0 +1,88 @@
...
+@pytest.mark.compression
+def test_clp_json_compression(fixt_package_instance: PackageInstance) -> None:
+    """
+    Validate that the `clp-json` package successfully compresses some dataset.
+
+    :param fixt_package_instance:
+    """
+    validate_package_instance(fixt_package_instance)
+
+    # TODO: compress some dataset and check the correctness of compression.
```
junhaoliao, line 65: can we link any WIP PRs for these TODOs here? if not, let's create GH issues to track them

integration-tests/tests/package_tests/clp_text/test_clp_text.py
```diff
@@ -0,0 +1,53 @@
...
+    pytest.mark.parametrize("fixt_package_test_config", [CLP_TEXT_MODE], indirect=True),
+]
+
+
+@pytest.mark.startup
+def test_clp_text_startup(fixt_package_instance: PackageInstance) -> None:
+    """Tests package startup."""
+    validate_package_instance(fixt_package_instance)
+
+    log_msg = "test_clp_text_startup is running successfully."
```
junhaoliao: fwiw, this log_msg seems to use a different wording / verb tense than the one in `test_clp_json_startup`. i also found the verb tenses are inconsistent with the docstrings too. we should make all consistent if it makes sense to do so

integration-tests/tests/utils/clp_mode_utils.py
```diff
@@ -52,7 +31,7 @@ def _to_docker_compose_service_name(name: str) -> str:
     return name.replace("_", "-")
 
 
-# These component lists should be maintained alongside the CLP_MODE_CONFIGS list.
+# These component lists should agree with the component lists of each package operating mode.
```
junhaoliao:
```suggestion
# Base components shared by all package modes. Test modules define mode-specific component lists
# by extending this (see tests/package_tests/*/test_*.py).
```

would this be more clear?

integration-tests/tests/package_tests/clp_text/__init__.py
```diff
@@ -0,0 +1 @@
+"""Top-level package for clp-text package integration tests."""
```
junhaoliao:
```suggestion
"""Integration tests for the clp-text package."""
```

integration-tests/tests/package_tests/clp_json/__init__.py
junhaoliao:
```suggestion
"""Integration tests for the clp-json package."""
```

integration-tests/tests/utils/config.py
junhaoliao: if we keep them separate, i think we should come up with better names to differentiate the purposes of those two classes. (they were not immediately obvious to me, though now after i read the code a few more times, i think i understand better now:
1. PackageInstance is for tracking intermediate states / artifacts
2. PackageTestConfig is for supplying input configs

i don't find harm in merging them into a single class. Even if we end up with more attributes in the class, we may still be able to group the attributes and put high-level docstrings to explain each group.

so yeah i vote +1 for merging

quinntaylormitchell: sounds good, will make a note to put up a PR to merge the classes.

integration-tests/tests/utils/config.py
quinntaylormitchell: ah @junhaoliao this is a very good point, I hadn't realized that. the full validation process only takes ~100ms on my system, which is far less than compression or search jobs will take to run on the package. so with that in mind I think the best course of action is to remove the `instance_validated` flag and perform the validation every time. I'm not sure what you mean by "remove caching" but was that it?

integration-tests/tests/fixtures/package_test_config.py
quinntaylormitchell: @junhaoliao yep, I like this.

integration-tests/tests/package_tests/clp_json/__init__.py
quinntaylormitchell: agreed

integration-tests/tests/package_tests/clp_json/test_clp_json.py
quinntaylormitchell: these TODOs will be fulfilled and removed by the PRs in the project PR plan; I can create gh issues if you'd like

integration-tests/tests/package_tests/clp_text/__init__.py
quinntaylormitchell: agreed

integration-tests/tests/package_tests/clp_text/test_clp_text.py
quinntaylormitchell: agreed that the different wording should be standardized; will change. most meaningful logging and reporting changes will come in #1802; making a note to standardize wording/verb tense there.

integration-tests/tests/utils/clp_mode_utils.py
quinntaylormitchell: I like the gist of that, but I want the explanation to encompass `CLP_API_SERVER_COMPONENT` as well. I'll write

```suggestion
# Names of components that may comprise a given package mode. Test modules use these lists to
# assemble mode-specific component lists (see tests/package_tests/*/test_*.py).
```

integration-tests/tests/utils/config.py
quinntaylormitchell: is there value in making it `frozen=True`?

junhaoliao: That makes it clear that the objects are catalogues rather than regular objects like data stores.

integration-tests/tests/package_tests/clp_json/test_clp_json.py
junhaoliao: sounds good. let's ask the rabbit for help.

@coderabbitai can you create issues to track both the test_clp_xxxx_compression TODO and the test_clp_xxxx_search TODO?

integration-tests/tests/utils/clp_mode_utils.py
junhaoliao: sounds good

integration-tests/tests/utils/config.py
junhaoliao: Right, removing `instance_validated` was what i meant. sounds good then

integration-tests/tests/utils/config.py
junhaoliao: @coderabbitai submit a GH issue to track that we will merge PackageInstance and PackageTestConfig into one class

integration-tests/tests/utils/config.py
quinntaylormitchell: ok, sounds good.

Bill-hbrhbr (approved):

PR #1847, opened 2026-01-07
feat(clp-s::timestamp_parser): Add support for ignoring quotes in quoted timestamp patterns.

gibber9809:
# Description

This PR changes the parsing behaviour of the timestamp parser to skip the beginning and end quote in a quoted timestamp pattern when parsing raw UTF-8 timestamps. This change will allow some of the code in #1788 to be simplified.

This change allows code that uses this library to handle UTF-8 and JSON literals in a simpler way. E.g., the pattern `"\Y"` can be used to parse both the JSON literal `"2024"` and the raw UTF-8 string `2024` by simply passing the right flags to `timestamp_parser::parse_timestamp`. The previous approach would have required library users to use a different timestamp pattern (here, `\Y`) for the UTF-8 case, which is cumbersome.

# Checklist

  breaking change.

# Validation performed

* Updated tests to exercise all valid combinations of parsing quoted/unquoted content with quoted/unquoted patterns.

## Release Notes

* **New Features**
  * Added support for quoted timestamp patterns, allowing timestamps enclosed in quotation marks to be properly recognized and parsed.
  * Enhanced timestamp parsing to correctly interpret and handle JSON escape sequences and special characters.

* **Tests**
  * Expanded test coverage to include verification of quoted timestamp patterns and multiple parsing scenarios, including edge cases with JSON escape sequences.

### DIFF
--- a/components/core/src/clp/ffi/ir_stream/search/test/test_QueryHandlerImpl.cpp
+++ b/components/core/src/clp/ffi/ir_stream/search/test/test_QueryHandlerImpl.cpp
@@ -127,6 +127,9 @@ auto generate_matchable_kql_expressions(
             matchable_kql_expressions.emplace_back(
                     fmt::format("{}: {}", column_query_with_namespace, cRefTestInt)
             );
+            matchable_kql_expressions.emplace_back(
+                    fmt::format("{}: timestamp(\"{}\")", column_query_with_namespace, cRefTestInt)
+            );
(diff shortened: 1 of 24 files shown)


LinZhihao-723 (approved):
The change lgtm!
For the PR title, how about:
```
feat(clp-s::timestamp_parser): Add support for ignoring quotes in quoted timestamp patterns.
```
to make it a bit more concise

PR #1848, opened 2026-01-07
feat(kv-ir)!: Add support for matching KQL timestamp literals against integer and float values.

gibber9809:
# Description

This PR updates the mapping between kv-ir stream schema tree nodes and literal types in order to allow timestamp literals to match floats, integers, and nothing else.

In practice for the current way ASTs are generated from KQL and refined this was already possible (since timestamp literals can be converted to integer and float literals), but this PR makes the mapping more explicit and adds tests.

# Checklist

  breaking change.

# Validation performed

* Added tests to ensure that timestamp literals work as expected for column resolution and search.
* Manually validated that timestamps in kv-ir streams can be searched as expected using timestamp literals in the `clp-s` cli.

## Release Notes

* **Tests**
  * Enhanced timestamp expression support in search query validation.
  * Extended test coverage for timestamp literal handling with various filter operations.

* **Internal Updates**
  * Refined type mapping for timestamp handling in search utilities.

### DIFF
--- a/tools/deployment/package-helm/Chart.yaml
+++ b/tools/deployment/package-helm/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: "v2"
 name: "clp"
-version: "0.1.2-dev.15"
+version: "0.1.2-dev.16"
 description: "A Helm chart for CLP's (Compressed Log Processor) package deployment"
 type: "application"
 appVersion: "0.7.1-dev"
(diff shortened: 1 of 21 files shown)


gibber9809:
> The code changes lgtm. My only concern is that this implementation will no longer allow comparing timestamp literal against an unparsed string representation. We should probably create an issue to keep track of this. For the PR title, how about:
> 
> ```
> feat(kv-ir)!: Add support for matching KQL timestamp literals against integer and float values.
> ```
> 
> I feel like we should mark this as a breaking change since it changes the search behavior.

Makes sense. Yeah I figured it was better to just drop the string comparison support for now until we have time to do it properly -- created #1851 to track.

PR #1829, opened 2025-12-26
feat(helm): Add multi-node deployment support with configurable scheduling.

junhaoliao:
# Description

> [!NOTE]
> This PR is part of the ongoing work for #1309. More PRs will be submitted until the Helm chart is complete and fully functional.

Adds support for multi-node deployments with configurable worker scheduling and automatic
control-plane toleration.

## Key Features

- **Deployment mode flag (`distributedDeployment`)**: Controls PersistentVolume type and scheduling behavior
  - `distributedDeployment=false` (default): Single-node mode with local PVs bound to control-plane node;
    pods automatically tolerate control-plane taints
  - `distributedDeployment=true`: Multi-node mode with hostPath PVs (assumes shared storage like NFS/CephFS
    mounted at the same path on all nodes)

- **Per-component scheduling**: Each worker type can be configured with:
  - `nodeSelector`: Target specific node labels
  - `affinity`: Pod/node affinity rules
  - `tolerations`: Tolerate node taints
  - `topologySpreadConstraints`: Spread pods across failure domains

- **Configurable worker replicas**: `compressionWorker.replicas` and `queryWorker.replicas`

- **StorageClass management**: Chart creates `local-storage` StorageClass with
  `WaitForFirstConsumer` binding mode by default

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

- Ran `test.sh` - single-node deployment with all pods running
- Ran `test-multi-shared.sh` - multi-node deployment with shared worker nodes
- Ran `test-multi-dedicated.sh` - multi-node deployment with dedicated node pools; verified
  compression-worker scheduled on compression node, query-worker scheduled on query node

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **New Features**
  * Configurable replicas for compression, query and reducer workers.
  * New scheduling options (node selectors, affinities, tolerations, topologySpread) with distribution-aware defaults.
  * Conditional static/local persistent volumes and optional StorageClass for local deployments.
  * New setup scripts to automate single- and multi-node KinD test clusters.

* **Chores**
  * Added deployment flags for distributed mode and worker concurrency; chart metadata version updated.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

Issue #1760, opened 2025-12-10
Centralized operational logging architecture for multi-host CLP deployments


junhaoliao:
### Background

Currently, CLP's deployment architecture requires container-to-host volume mounts for all service logs. Each service writes logs to files within `/var/log/<component>/` which are mounted from `${CLP_LOGS_DIR_HOST:-./var/log}` on the host. While this approach has been convenient for single-host deployments (allowing users to easily provide logs by archiving host files), it creates significant challenges for multi-host deployments:

1. **Multi-host incompatibility**: In Kubernetes or multi-node Docker Compose deployments, logs are scattered across different hosts, making centralized log access difficult
2. **Storage overhead**: Each host requires dedicated storage for log retention
3. **Operational complexity**: Admins must access (e.g. SSH into) individual hosts or set up additional log aggregation infrastructure
4. **Scaling limitations**: Adding nodes increases operational burden exponentially

With planned support for multi-host deployments through Kubernetes, and addition of the log-ingestor component, there's an opportunity to modernize the operational logging architecture to leverage:
1. **CLP's own compression technology** - operational logs can benefit from the same high-compression-ratio storage that user logs enjoy
2. **Container-native logging** - Docker/Kubernetes native log drivers eliminate the need for host mounts
3. **Centralized access** - all logs accessible from a single control node
4. **WebUI integration** - operational logs viewable alongside user logs in the existing CLP interface

### Requirements

The new operational logging solution must satisfy the following requirements:

#### R1: Multi-host Support
- **R1.1**: Support both Kubernetes (multi-node) and Docker Compose (single or multi-host) deployments
- **R1.2**: All logs accessible from a central control node, regardless of where services are running
- **R1.3**: No per-host log file access required for normal operations

#### R2: Tiered Access
- **R2.1 (Hot)**: Real-time access to recent logs (0-X minutes) for debugging active/crashed services
    - Maximum acceptable lag: < 30 seconds
    - Must support live tailing
- **R2.2 (Warm)**: Recent historical logs (X minutes - Y hours) available uncompressed for immediate grep/analysis
    - Maximum acceptable lag: < 5 minutes
    - Must support live tailing
- **R2.3 (Cold)**: Older logs compressed using CLP for long-term storage and efficient retrieval
    - Maximum acceptable lag: < 24 hours
    - Must support full-text search

#### R3: Admin Access & Export
- **R3.1**: Deployment admins can view all logs from all services
[...]

PR #1603, opened 2025-11-14
WIP - feat(deployment): Add k8s Helm chart for CLP package deployment.

junhaoliao:
this PR is meant to be reference only and should not be merged.
## Summary - PR Checklist

- [x] #1814 (`0.1.2-dev.8`) - Refactor `database.name` to `database.names.clp`
- [x] #1815 (`0.1.2-dev.9`) - Remove unused `query_scheduler.port`
- [x] #1816 (`0.1.2-dev.10`) - Add query-scheduler probes
- [x] #1817 (`0.1.2-dev.11`) - Add garbage_collector deployment (incl. retention_period)
- [x] #1818 (`0.1.2-dev.12`) - Add api_server deployment
- [x] #1819 (`0.1.2-dev.13`) - Add mcp_server deployment
- [x] #1822 (`0.1.2-dev.14`) - Template formatting improvements
- [x] #1846 (`0.1.2-dev.15`) - Add missing database.names.spider which prevented the API server from starting.
- [x] #1825 (`0.1.2-dev.16`) - S3/AWS storage support (depends on #1817, #1818)
- [x] #1856 (`0.1.2-dev.17`) - Add log-ingestor for S3 (depends on #1825)
- [x] #1829 (`0.1.2-dev.18`) - Add node affinity and topology spread for multinode worker deployments (depends on #1825)
- [x] #1882 (`0.1.2-dev.19`) - fix(helm): Ensure log_ingestor is explicitly null in configmap when disabled (fixes #1881).
- [ ] #1913 (`0.1.3-dev.2`) - fix(helm): Enable log_ingestor by default (fixes https://github.com/y-scope/clp/issues/1912); gate log_ingestor templates to S3 logs_input only.
- [ ] #1834 (`0.1.3-dev.3`) - Use tmpfs for temporary directory mounts in Docker Compose.
- [ ] #1827 - Documentation updates (Kubernetes and Docker Compose deployment guides)
- [ ] PR 13 (`0.1.2-dev.21`) - Spider support (code + docs, de-prioritized)

PR #1856, opened 2026-01-08
feat(helm): Add `log-ingestor` deployment.

junhaoliao:
# Description

> [!NOTE]

Add Helm chart support for deploying the log-ingestor component. The log-ingestor enables ingestion of logs from S3 buckets via periodic scanning or SQS queue notifications.

This PR adds:
- Deployment template for log-ingestor with health checks (readiness/liveness probes)
- Service template exposing the log-ingestor on NodePort 30302
- PersistentVolume and PersistentVolumeClaim templates for log-ingestor logs
- ConfigMap configuration for log-ingestor settings
- values.yaml configuration options for log-ingestor

The log-ingestor is conditionally deployed only when `clpConfig.log_ingestor` is configured in values.yaml.

# Checklist

  breaking change.

# Validation performed

```
cd tools/deployment/package-helm
./test.sh
# Output: All jobs completed and services are ready.

curl http://localhost:30302/health
# Output: log-ingestor is running

curl --location 'localhost:30302/s3_scanner' \
--header 'Content-Type: application/json' \
--data '{"bucket_name": "xxxx", "key_prefix": "/", "region": "us-east-2"}'
# Output: {
#     "id": "8c5f6817-2166-44d7-93ea-4d26282412d4"
# }

kubectl logs test-clp-log-ingestor-66656c445-8qdt8
# Output: ...
# "timestamp":"2026-01-08T21:53:29.153277Z","level":"INFO","fields":{"message":"Scanned new object metadata on S3.","object":"ObjectMetadata { bucket: NonEmptyString(\"yscope-junhao-test\"), key: NonEmptyString(\"/fb0befb7-a8ee-4de4-b2f6-60d2284d2c97_0_30.jsonl\"), size: 8164 }"},"filename":"components/log-ingestor/src/ingestion_job/s3_scanner.rs","line_number":121}

```

## Summary by CodeRabbit

* **New Features**
  * Introduced a configurable log-ingestor component for log ingestion and processing.
  * Added optional settings for buffer thresholds, timeouts, channel capacity and logging verbosity.
  * Added support for persistent log storage and a dedicated service endpoint on port 3002.

* **Chores**
  * Updated Helm chart version to 0.1.2-dev.17.
  * Adjusted local cluster port mappings and expanded runtime log directories.


tools/deployment/package-helm/templates/log-ingestor-deployment.yaml
LinZhihao-723, line 1: Cross-compared with `api-server-deployment.yaml`. The changes lgtm as:
- log-ingestor only waits for DB creation; no need to wait for result cache to start.
- log-ingestor doesn't need storage mount at this stage.
- log-ingestor doesn't need stream storage.

tools/deployment/package-helm/templates/log-ingestor-logs-pv.yaml
LinZhihao-723, line 1: Cross-compared with `api-server-logs-pv.yaml`. The changes lgtm.
My only concern is the 5Gi capacity might be too small. Current log ingestor generates a log even whenever a new object is ingested, with the full S3 key and metadata tracked. Assuming each log event is 150-byte long on average, this log capacity would only allow 35k files ingested. What's the behavior if the log size exceeds the capacity?

tools/deployment/package-helm/templates/log-ingestor-logs-pvc.yaml
LinZhihao-723, line 1: Cross-compared with `api-server-logs-pvc.yaml` and the changes lgtm.

tools/deployment/package-helm/templates/log-ingestor-service.yaml
LinZhihao-723, line 1: Cross-compared with `api-server-srervice.yaml` and the changes lgtm.

tools/deployment/package-helm/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: "v2"
 name: "clp"
-version: "0.1.2-dev.16"
+version: "0.1.2-dev.18"
```
LinZhihao-723: What's the reason for boosting 16 to 18 but not 17?

tools/deployment/package-helm/values.yaml
LinZhihao-723, line 1: Cross-compared with the current package config and this lgtm.

LinZhihao-723 (commented):
In general I don't have any questions on changes. Will run validations on my end with necessary set up before approval.

LinZhihao-723 (dismissed):
For the PR title, how about:
```
feat(helm): Add `log-ingestor` deployment.
``` 
This should match how we refer to the log ingestor service in previous PR titles.

tools/deployment/package-helm/Chart.yaml
junhaoliao: 1856 is waiting on #1829 according to the PR submission plan posted on #1603

LinZhihao-723: I see, make sense

tools/deployment/package-helm/templates/log-ingestor-logs-pv.yaml
junhaoliao: > What's the behavior if the log size exceeds the capacity?

if i'm not wrong, logs are dropped silently.

Let's raise it to 10Gi to match the compression scheduler log volume. In the long run, we should implement #1760 instead of requesting host volumes specifically for logs.

tools/deployment/package-helm/Chart.yaml
junhaoliao: plan changed: we will merge log ingestor first
```suggestion
version: "0.1.2-dev.17"
```

LinZhihao-723 (approved):
🚀

tools/deployment/package-helm/values.yaml
```diff
@@ -90,6 +90,19 @@ clpConfig:
   #   port: 30800
   #   logging_level: "INFO"
 
+  # log-ingestor config. Currently, the config is applicable only if `logs_input.type` is "s3".
+  log_ingestor: null
```
junhaoliao, line 94: We should explicitly set `log_ingestor: null` when it is so in `values.yaml`; otherwise, the Pydantic model parser would fallback to default values which leaves the log-ingestor configured. When `clpConfig.package.storage` is configured as `"clp"` (CLP-Text which doesn't support the log-ingestor), the config is deemed invalid and most Helm chart deployments won't start.

junhaoliao: #1882 sets `log_ingestor: null` in the `clp_config.yaml` ConfigMap to address #1881. However, the Rust utils cannot handle null values for this key and will fail with a parsing error.

As a result, if the API server is enabled while `log_ingestor` is disabled by setting `log_ingestor: null`, the API server will fail to start.

More importantly, this configuration ("enable the API server but disable log_ingestor by setting `log_ingestor: null`") is currently the default behavior in our Helm `values.yaml`. This means the v0.8.0 chart deploys an API server that continuously crashes and restarts.

I think we need to address two issues:
* Update the Rust utils to correctly handle a nullable log_ingestor.
* Enable log_ingestor by default in Helm values.yaml, so it aligns with the default clp_config.yaml (and clp_config.template.json.yaml).

@coderabbitai create those new two issues as GH issues

PR #1865, opened 2026-01-13
feat(core): Optimize subquery generation by narrowing wildcard token type inference (fixes #1864).

### LINKED ISSUE #1864: feat(core): Eliminate impossible token types for non-definite variable tokens to skip invalid subquery interpretations that are causing unnecessary decompression and scan.
### Request

For a query like `"*NonDFS*"` we end up treating the query as a single non-definite variable token containing wildcards. The current logic in `QueryToken.cpp` will treat this kind of token as any possible variable type, without checking whether each interpretation is actually possible.

For `*NonDFS*`, the token clearly could not ever be an `IntVar` or `FloatVar`, because it contains characters that could not possibly appear in those kinds of variables.

The current logic will end up turning this into four subqueries against logtypes matching `*NonDFS*`, `*\x11*`, `*\x12*`, `*\x13*` (i.e. any logtypes containing NonDFS, or ANY integer, float, or dictionary variable placeholder), which effectively results in decompression and scan for the entire dataset.

This particular example comes from an internally reported performance bug where this particular query takes approximately ~12m30s single-threaded over a ~300GB dataset because the generated subqueries require full decompression and scan. Hacking a fix into subquery generation to ignore the impossible `IntVar` and `FloatVar` brings query completion time down to ~3s.

### Possible implementation

Instead of treating non-definite variable tokens containing wildcards as any possible type, we should try to determine which types are actually possible.

This can be achieved by writing helper functions that use heuristics to determine whether it is possible for a wildcard string to be interpreted as a representable `IntVar` or `FloatVar`. Note that a heuristic approach can lead to us still interpreting some query tokens as possibly corresponding to `IntVar`/`FloatVar` when they're not, but it should significantly cut down on invalid query interpretations in the common case.

gibber9809:
# Description

This PR fixes a performance bug described in #1864. The issue is that during subquery generation we treat non-definite variable tokens containing wildcards as any possible type, which can lead to us considering some impossible query interpretations that in turn lead to unnecessary decompression and scan.

The fix is to add some heuristic helpers that try to determine if a wildcard string could possibly correspond to a representable `IntVar` or `FloatVar`. We use these helpers to skip these query interpretations when they're definitely impossible.

For the query and dataset provided by @rishikeshdevsot, this change improves single-threaded query performance by about ~250x because decompression and scan is reduced from ~300GB to ~500MB as a result of not pursuing the impossible query interpretations.

# Checklist

  breaking change.

# Validation performed

* Added tests for helpers that determine whether wildcard strings could correspond to `IntVar` or `FloatVar`.
* Benchmarked performance improvement in example provided by @rishikeshdevsot.

## Summary by CodeRabbit

* **New Features**
  * Added wildcard validation for determining whether wildcarded strings could represent integers or floats, improving numeric inference during encoding and matching.

* **Bug Fixes**
  * Refined type inference for tokens with wildcards so numeric possibilities are only retained when wildcard patterns can validly represent numbers.

* **Tests**
  * Added parameterized tests covering a range of integer and float wildcard representability cases.


components/core/src/clp/EncodedVariableInterpreter.hpp
```diff
@@ -118,6 +118,19 @@ class EncodedVariableInterpreter {
      */
     static void convert_encoded_float_to_string(encoded_variable_t encoded_var, std::string& value);
 
+    /**
+     * @param value
+     * @return Whether the given wildcard string could correspond to a representable integer
+     * variable.
+     */
+    static auto wildcard_string_could_be_representable_integer_var(std::string_view value) -> bool;
```
LinZhihao-723:
```suggestion
    [[nodiscard]] static auto wildcard_string_could_be_representable_integer_var(
            std::string_view value
    ) -> bool;
```

components/core/src/clp/EncodedVariableInterpreter.hpp
```diff
@@ -118,6 +118,19 @@ class EncodedVariableInterpreter {
...
+
+    /**
+     * @param value
+     * @return Whether the given wildcard string could correspond to a representable float variable.
+     */
+    static auto wildcard_string_could_be_representable_float_var(std::string_view value) -> bool;
```
LinZhihao-723:
```suggestion
    [[nodiscard]] static auto wildcard_string_could_be_representable_float_var(
            std::string_view value
    ) -> bool;
```

components/core/src/clp/EncodedVariableInterpreter.cpp
```diff
@@ -196,6 +196,28 @@ void EncodedVariableInterpreter::convert_encoded_float_to_string(
...
 
+auto EncodedVariableInterpreter::wildcard_string_could_be_representable_integer_var(
+        std::string_view value
+) -> bool {
+    for (char const c : value) {
+        if (false == (('0' <= c && c <= '9') || '-' == c || '?' == c || '*' == c)) {
+            return false;
+        }
+    }
+    return true;
```
LinZhihao-723:
```suggestion
    return false == std::ranges::any_of(value, [](char const c) -> bool {
               return false == (('0' <= c && c <= '9') || c == '-' || c == '?' || c == '*');
           });
```
Could be simplified using `any_of`. We could also use `all_of` for better readability but I guess this form is more efficient.

components/core/src/clp/EncodedVariableInterpreter.cpp
```diff
@@ -196,6 +196,28 @@ void EncodedVariableInterpreter::convert_encoded_float_to_string(
...
+
+auto EncodedVariableInterpreter::wildcard_string_could_be_representable_float_var(
+        std::string_view value
+) -> bool {
+    for (char const c : value) {
+        if (false == (('0' <= c && c <= '9') || '-' == c || '.' == c || '?' == c || '*' == c)) {
+            return false;
+        }
+    }
+    return true;
```
LinZhihao-723:
```suggestion
    return false == std::ranges::any_of(value, [](char const c) -> bool {
               return false
                      == (('0' <= c && c <= '9') || c == '-' || c == '.' || c == '?' || c == '*');
           });
```

components/core/src/clp/EncodedVariableInterpreter.cpp
LinZhihao-723, line 1: Sorry for missing this, we should also include `<algorithm>`

components/core/src/clp/EncodedVariableInterpreter.cpp
```diff
@@ -196,6 +197,23 @@ void EncodedVariableInterpreter::convert_encoded_float_to_string(
     value[value_length - 1 - decimal_pos] = '.';
 }
 
+auto EncodedVariableInterpreter::wildcard_string_could_be_representable_integer_var(
+        std::string_view value
+) -> bool {
+    return false == std::ranges::any_of(value, [](char const c) -> bool {
+               return false == (('0' <= c && c <= '9') || c == '-' || c == '?' || c == '*');
+           });
+}
```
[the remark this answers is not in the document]
LinZhihao-723: @gibber9809 I think rabbit is probably right. Even in our case, it is impossible to have such a case, but from this helper's perspective we should probably check whether the give string is empty.

LinZhihao-723 (changes requested):

components/core/src/clp/EncodedVariableInterpreter.cpp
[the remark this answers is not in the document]
gibber9809: Sure -- yeah I was ignoring it because it can't happen for our case, but it's basically free, so I'm fine adding it.

gibber9809:
> 
> ```
> feat(core): Optimize subquery generation by narrowing wildcard type inference (fixes #1864).
> ```

Going to use "wildcard token type inference" as a slight edit.

PR #1627, opened 2025-11-19
style(core): Update clang-format include priority to separate project headers using angle brackets; Refactor JsonParser to use angle bracket includes for project headers.

davidlion:
# Description

This PR updates our clang-format configuration to reflect our latest include guidelines and refactors JsonParser.{cpp,hpp} to provide an example.

Functionally, this change only tells clang tidy to separate angled includes of the project (clp, clp_s, etc.) from standard library includes. Since clp has been using relative includes there are no automated changes made to existing code.
Going forward reviewers should ensure PRs now use angled project includes over relative includes as per the guidelines.

The guidelines can be found here:
https://www.notion.so/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?source=copy_link#25304e4d9e6b801481fbe508b428af20

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

CI passing.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **Chores**
  * Standardized header grouping and comments to unify include ordering.
  * Consolidated local/relative includes into public-style include paths to clarify dependency boundaries.
* **Refactor**
  * Reorganized header imports and grouping for consistency; removed now-unused standard headers.
  * No changes to public APIs or runtime behaviour.

PR #1868, opened 2026-01-14
build(cmake): Add include directories for all targets to include project headers using angle brackets; Define `CLP_ENABLE_TESTS` to enable test targets.

SharafMohamed:
# Description
- Adds `components/core/src` directory to target include directories for all targets to allow for the usage of system include such as `<clp/Def.h>` over `"Defs.h"` as per https://github.com/y-scope/clp/pull/1627.
- Added default testing if CLP is the top level project. This also allows for testing flags to be set in the future, which will be used to create friend classes to test private methods without exposing them in production code.

# Validation Performed
- Tested swapping headers in various files to use system includes instead of local includes. Successfully compiles.    

## Summary by CodeRabbit

* **Chores**
  * Broadened header include resolution across multiple components and executables to reduce build fragility and improve header discovery.
* **Tests**
  * Introduced CTest integration and a new test enablement flag so optional unit tests are gated and activated more consistently during builds.

### DIFF
--- a/components/core/CMakeLists.txt
+++ b/components/core/CMakeLists.txt
@@ -135,6 +135,7 @@ if(PROJECT_IS_TOP_LEVEL)
         endif()
     endif()
 
+    # If previously undefined, `BUILD_TESTING` will be set to ON.
     include(CTest)
 endif()
 


components/core/CMakeLists.txt
```diff
@@ -735,7 +735,15 @@ set(SOURCE_FILES_unitTest
         tests/test-Utils.cpp
         )
 
-if(CLP_BUILD_TESTING)
+if(CLP_IS_TOP_LEVEL)
+    include(CTest)
+endif()
```
davidlion: In general we want to move these to the top of the file to ensure everything is defined early to avoid anyone adding code that uses `CLP_ENABLE_TESTS` before it is defined.

Move into `PROJECT_IS_TOP_LEVEL` block at line ~121.

components/core/CMakeLists.txt
```diff
@@ -735,7 +735,15 @@ set(SOURCE_FILES_unitTest
...
+
+if(BUILD_TESTING AND CLP_BUILD_TESTING)
+    set(CLP_ENABLE_TESTS ON)
+endif()
```
davidlion: This block can be moved up just after the `PROJECT_IS_TOP_LEVEL` block.

components/core/CMakeLists.txt
SharafMohamed: Moved. I'm assumed `CLP_IS_TOP_LEVEL` is redundant with `PROJECT_IS_TOP_LEVEL` so modified it accordingly.

components/core/CMakeLists.txt
SharafMohamed: Moved.

components/core/CMakeLists.txt
```diff
@@ -134,6 +134,12 @@ if(PROJECT_IS_TOP_LEVEL)
             cmake_policy(SET "CMP0144" "NEW")
         endif()
     endif()
+
+    include(CTest)
```
davidlion, line 139: Sorry one last nit. I think it is worth copying this comment from log-surgeon/ystdlib-cpp since this behaviour isn't obvious unless you're very familiar with CMake.

```suggestion
    # If previously undefined, `BUILD_TESTING` will be set to ON.
    include(CTest)
```

kirkrodrigues (approved):
Approving last commit. Updated PR title based on offline discussion.

PR #1884, opened 2026-01-19
chore: Bump version for 0.8.0 release and remove edit page button from docs.

kirkrodrigues:
# Description

This PR updates any version references from 0.7.1-dev to 0.8.0 in preparation for the release.

# Checklist

  breaking change.

# Validation performed

Validated `task` succeeded.


docs/conf/conf.py
```diff
@@ -85,25 +86,15 @@
     "navbar_start": ["navbar-logo", "version-switcher"],
     "navbar_end": ["navbar-icon-links", "theme-switcher"],
     "primary_sidebar_end": [],
-    "secondary_sidebar_items": ["page-toc", "edit-this-page"],
```
junhaoliao, line 88: do we want to instead keep the edit button but let it point to docs on the main branch?

kirkrodrigues: The issue is that the docs on the main branch may change significantly, so having an edit button could be confusing. Alternatively, we could make it point at the release branch, but I would prefer not to encourage edits to release docs until we have a better workflow for updating such docs.

docs/conf/conf.py
```diff
@@ -85,25 +86,15 @@
...
+    "secondary_sidebar_items": ["page-toc"],
     "show_prev_next": False,
     "switcher": {
         "json_url": "https://docs.yscope.com/_static/clp-versions.json",
-        "version_match": CLP_GIT_REF,
+        "version_match": CLP_VERSION,
     },
     "use_edit_page_button": True,
```
junhaoliao:
```
    "use_edit_page_button": True,
```
what's the effect of keeping this? shall we remove?

kirkrodrigues: Yeah, good point, we can.

kirkrodrigues:
> the title lgtm. (i suppose it's intentional not to mention that we are removing the edit button, since we will apply similar changes on the release branches for every release)

Not intentional. Since I follow a runbook for creating the release, I didn't think too much about the title. What do you think of the updated title?

junhaoliao:
> we should also lock the Helm Chart version?

and the image ref in values.yaml

junhaoliao (approved):

PR #1368, opened 2025-10-02
feat(clp-package): Add CLP Model Context Protocol (MCP) Server:

20001020ycx:
* Add `clp-mcp-server` Python project.
* Add tasks to build and package `clp-mcp-server` as a part of the CLP package.

# Description

**1. MCP Server Source Code Setup**
Introduces the MCP server as a new component under the CLP project, with the following directory structure:
clp/components/clp-mcp-server/
├── pyproject.toml & uv.lock
├── README.md
└── clp_mcp_server/
&nbsp;&nbsp;├── server
&nbsp;&nbsp;&nbsp;&nbsp;└──  server.py (implementation of  MCP API tool calls)
&nbsp;&nbsp;└── clp_mcp_server.py (main)
- pyproject.tom/uv.lock: Configured Poetry with uv.lock for managing dependencies required for the MCP server.
- Package name: clp-mcp-server. This is the Python package executed when Docker Compose starts the MCP server container: python3 -m clp_mcp_server

**2. Taskfile Integration**
Adds infrastructure for building and packaging the MCP server.
- Create task clp-mcp-server: Generate a Python wheel at clp-mcp-server/dist/clp-mcp-server.whl. 
- Update task package: Install the MCP server wheel into the CLP package’s Python environment. After installation, we should have:
  - MCP server’s source files in build/clp-package/lib/python3/site-packages/clp_mcp_server/
  - MCP server’s metadata and dependencies in build/clp-package/lib/python3/site-packages/clp-mcp-server-0.5.2.dev0.dist-info/
 - Documentation and linter updates of the new component for developer usability

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

1. Ran `task package` to build the package.
2. Ran `cd build/clp-package/lib/python3/site-packages && python3 -m clp_mcp_server.clp_mcp_server --host 127.0.0.1 --port 8000`
4. Confirmed the MCP server started successfully.
5. Connected Claude Desktop agent and issued a “hello world” query, receiving the expected greeting and server status response.
Screenshot 2025-10-04 at 12 37 07 PM

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

- **New Features**
[...]

PR #1888, opened 2026-01-20
fix(yscope-clp-core): Fix Python wheel build workflow failure by disabling uv cache; Move the temporary build directory under `G_BUILD_DIR`.

LinZhihao-723:
# Description

The task introduced in #1732 involves `uv` calls where caching is not disabled. This causes the wheel build workflow to fail when running inside GitHub runner as reported:

- https://github.com/kirkrodrigues/clp/actions/runs/21166625748/job/60872856714
- https://github.com/kirkrodrigues/clp/actions/runs/21166506200/job/60872866386

The root cause of the failure is that when `uv` tries to create a cache, the default path of the cache depends on `$HOME` env variable, which is not set in this workflow setup (causing by https://github.com/y-scope/clp/blob/1843abd9026d9b7040ef641a6491afd916382b6f/.github/actions/run-on-image/action.yaml#L49 not executing properly inside the GitHub runner).

We saw similar problems in #1368 and back to that time I decided to disable the cache as discussed in https://github.com/y-scope/clp/pull/1368#discussion_r2404123388. It's my bad that I forgot this issue over this long and it's been triggered again.

To properly fix this caching issue, we should set the cache directory manually whenever a uv call is needed. This is tracked in issue https://github.com/y-scope/clp/issues/1382.

# Checklist

  breaking change.

# Validation performed

* [x] Reprodueced the issue locally and ensured that the wheel can be properly built by disabling the uv cache.

## Summary by CodeRabbit

* **Chores**
  * Temporary build directory moved under the main build output directory for clearer artifact placement.
  * Python wheel build and repair steps now run with dependency caching disabled to force fresh resolution each run.
  * Temporary-directory creation and cleanup commands updated for more consistent invocation syntax.

---
  - https://app.asana.com/0/0/1212889533484534

### DIFF
--- a/taskfiles/toolchains.yaml
+++ b/taskfiles/toolchains.yaml
@@ -103,12 +103,13 @@ tasks:
         EOF
 
       # Disable Rustup's auto self update at the end to avoid unexpected checksum mismatches.
+      # NOTE: Temporarily lock `nextest` to v0.9.122 as a workaround for #1906.
       - |-
         . "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
         rustup toolchain install nightly --allow-downgrade --component clippy,rustfmt
         rustup toolchain install stable
         rustup default stable
-        cargo install cargo-nextest
+        cargo install cargo-nextest --version 0.9.122
         rustup set auto-self-update disable
 
       # This command must be last


taskfile.yaml
junhaoliao, line 1: kinda unrelated to the PR, but can we place TMP_DIR into the build dir so all build assets are consistently inside build/? e.g.,
```
TMP_DIR: "{{.G_BUILD_DIR}}/tmp-{{.TASK}}"
```

taskfile.yaml
```diff
@@ -609,8 +609,9 @@ tasks:
       # Build the wheel into the temporary directory
       - |-
         export YSCOPE_CLP_CORE_CLP_S_EXE="{{.G_CORE_COMPONENT_BUILD_DIR}}/clp-s"
-        uv run python3 -m build --wheel --outdir "{{.TMP_DIR}}"
+        uv run --no-cache python3 -m build --wheel --outdir "{{.TMP_DIR}}"
```
junhaoliao, line 612: Would it work if we directly override the UV_CACHE_DIR as suggested in the description?

e.g., simply add
```
    env:
      UV_CACHE_DIR: "{{.G_BUILD_DIR}}/uv-cache"
```
and keep everything as original.

taskfile.yaml
LinZhihao-723: Sure, do u want me to do it in this PR?

taskfile.yaml
LinZhihao-723: This can be an alternative and I can try it out.
However, as a quick fix to our CI, do u think we should defer this to another PR to make it more self-contained, as we also need to update other uv calls as well.

junhaoliao: discussed offline - we should consistently override this env wherever uv is used, rather than simply this task, to avoid unexpected behaviours / not utilizing caches properly. we will defer it as the PR description explains then

taskfile.yaml
junhaoliao: discussed offline. now resolving this given the changes have been applied.

LinZhihao-723:
CI failure due to: https://github.com/nextest-rs/nextest/issues/2990

LinZhihao-723:
https://github.com/y-scope/clp/issues/1906

Issue #1897, opened 2026-01-22
clp-package: Scheduling jitter is much larger than the scheduling interval, resulting in slow search.


gibber9809:
### Bug

Our query scheduler is currently designed to run a scheduling loop on a fixed polling interval, where on each interval we gather finished batches of tasks and dispatch new batches. This design has some fundamental problems, but that will be addressed in a separate design doc going over a proper solution.

The immediate issue, then, is that this scheduling loop currently has so much jitter that the effective rate at which we can dispatch new batches of task for each job is regulated by this jitter instead of the configured polling interval.

The following chart shows the breakdown of how time is spent in the main part of the scheduling loop over a small illustrative time slice of a longer search job as batches of task are completed, and illustrates how task completion affects jitter. (Apologies for the ad-hoc chart).

<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/543d50c8-6560-43c6-929e-e280277300cc" />

As you can see from the graph, each time a batch of tasks is completed we spend a suspicious amount of time waiting for results from celery.

As it turns out, this is because even after `task.ready()` is true, celery still seems to go into a polling loop to retrieve results from redis -- since the default polling interval is 0.5s we seem to always experience this 0.5s delay when retrieving results. Reducing this polling interval directly reduces the time we spend in `task.get()` (and experimentally even when reduced significantly we seem to always spend ~`polling_interval` time in `get()`).

Besides this issue with retrieving the results of celery tasks, we can also reduce jitter by changing how we sleep in our main scheduling loop.

Currently the loop looks something like
```
while True:
    # do stuff
    await asyncio.sleep(polling_interval)
```
but to reduce jitter we really want something like
```
while True:
    # do stuff
    await asyncio.sleep(polling_interval - time_spent_doing_stuff)
```

### CLP version

0.8.0

### Environment

Package build started with docker-compose.

### Reproduction steps

1. Compress enough data to form at least a few archives
2. Make sure the configured batch size is less than the total number of archives
3. Dispatch any search across all archives (note that because of another issue, command line searches that don't invoke the reducer end up with all tasks in a single batch).

gibber9809:
<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/c8ea5ed8-e61f-45f0-8883-4b95b5aefedd" />
After #1899, and with some code moved around to make measurement easier, this is what a small time slice of a batched query looks like.

PR #1899, opened 2026-01-22
perf(clp-package): Reduce batch scheduling jitter in the query scheduler.

gibber9809:
# Description

The query scheduler has extreme jitter in the main task dispatch loop, making task dispatch frequency a direct function of this jitter as described in #1897.

This PR significantly reduces the jitter, bringing the pacing of the main scheduling loop much closer to the configured polling interval, and likewise pulling the maximum task dispatch rate closer to the polling interval. The following chart shows the jitter and breakdown of time spent over a small time slice of a search query as batches of search tasks are completed.

ReducedJitter

When compared to the chart in #1897, you can see that the main source of the jitter (long waits for retrieving celery results) has been significantly reduced, and pacing for the main scheduling loop is much closer to the configured rate (configured to once every 0.1 seconds in this case).

Note that while the pacing still isn't perfect, the part of this diff that attempts to explicitly pace the scheduling loop does help (without this part of the diff excursions from the configured rate are even larger). This means that there are likely other sources of jitter in the query scheduler that we could attempt to eliminate.

Also note that reducing jitter in this PR results directly in improved query performance in the package (since batches of tasks can now be dispatched more frequently).  For some queries consisting of very short running tasks I measured a decrease in query time from ~1minute to ~18seconds in my single-machine docker-compose setup.

# Checklist

  breaking change.

# Validation performed

* Ran searches with extra manually inserted instrumentation in the package, and directly measured decrease in jitter as a result of this change.

## Summary by CodeRabbit

**Bug Fixes**
* Enhanced job orchestration stability by refining task result retrieval and polling intervals, resulting in more consistent and reliable job status updates regardless of task execution time variations.

**Performance**
* Optimized polling mechanism to maintain uniform check intervals, reducing delays and improving responsiveness in job scheduling operations.


components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1087,9 +1087,13 @@ async def check_job_status_and_update_db(db_conn_pool, results_cache_uri):
...
 async def handle_job_updates(db_conn_pool, results_cache_uri: str, jobs_poll_delay: float):
     while True:
+        interval_start_time = datetime.datetime.now()
         await handle_cancelling_search_jobs(db_conn_pool)
         await check_job_status_and_update_db(db_conn_pool, results_cache_uri)
-        await asyncio.sleep(jobs_poll_delay)
+        interval_end_time = datetime.datetime.now()
+        await asyncio.sleep(
+            jobs_poll_delay - (interval_end_time - interval_start_time).total_seconds()
+        )
```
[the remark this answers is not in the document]
gibber9809: The behaviour of asyncio.sleep for negative durations is well-documented, so handling this as a special case in our own code decreases clarity imo.

gibber9809:
Doing one more experiment to see if I can get the jitter even lower -- noticed on a second look that the task retiring is in the polling loop here, but the task dispatching is in our other polling loop. Going to see if there are any easy improvements to that polling loop that further reduce jitter.

gibber9809:
> Doing one more experiment to see if I can get the jitter even lower -- noticed on a second look that the task retiring is in the polling loop here, but the task dispatching is in our other polling loop. Going to see if there are any easy improvements to that polling loop that further reduce jitter.

Yeah seems like there's potentially at least one or two more obvious improvements worth trying out. Going to temporarily put this PR in draft, then take it out once I've finished trying them.

gibber9809:
ReducedJitterMoreDetail
For reference, this is what the overheads look like when I move all of the code related to dispatching batches into the same flow so it's a bit easier to measure.

gibber9809:
Marking ready for review but removing "fixes #1897" since this PR doesn't completely remove the jitter, it just significantly reduces it.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -859,7 +859,7 @@ def handle_pending_query_jobs(
 def try_getting_task_result(async_task_result):
     if not async_task_result.ready():
         return None
-    return async_task_result.get()
+    return async_task_result.get(interval=0.005)
```
sitaowang1998, line 862: I am a little confused about the interval. If the result is already ready, why do we need an interval?

gibber9809: Because celery still polls for the result for some reason and never seems to actually retrieve it on the first request. This polling interval is 0.5 seconds by default, so what was happening previously was that this call would always wait for 0.5 seconds then retrieve the result -- you can see this in the breakdown shown in #1897.

As for why celery works this way, I'm not sure -- some sources online imply that task state (running, failed, successful) is tracked separately from task result payload, which could explain this behaviour.

Issue #1906, opened 2026-01-24
CI failure caused by nextest compilation failure


LinZhihao-723:
Failure: https://github.com/y-scope/clp/actions/runs/21307314336/job/61337558248?pr=1888
Related issue: https://github.com/nextest-rs/nextest/issues/2990
Related fix PR: https://github.com/nextest-rs/nextest/pull/2991
nextest version: 0.9.123

PR #1907, opened 2026-01-24
fix(toolchains): Temporarily lock `nextest` to v0.9.122 as a workaround for #1906.

LinZhihao-723:
# Description

A CI failure has been reported in #1906.

As a workaround, we Temporarily lock `nextest` to v0.9.122 to by pass the issue, since it seems like v0.9.124 won't be released very soon.

# Checklist

  breaking change.

# Validation performed

### DIFF
--- a/taskfiles/toolchains.yaml
+++ b/taskfiles/toolchains.yaml
@@ -103,6 +103,7 @@ tasks:
         EOF
 
       # Disable Rustup's auto self update at the end to avoid unexpected checksum mismatches.
+      # NOTE: Temporarily lock `nextest` to v0.9.122 as a workaround for #1906.
       - |-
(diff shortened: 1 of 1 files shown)


LinZhihao-723:
> for record-keeping, mind explaining why we want to use the latest version of `cargo-nextest` rather than locking it to some specific version to promote reproducibility?

This is a test runner that is more efficient than the default Rust test runner (also printing the test results in a better-looking way), not any particular library we strictly rely on. We can still run all tests with the default Rust runner.
Ideally, it should always be preferred to use the latest as:
* It should always support whatever is supported in the default Rust test runner.
* None of our existing test code relies on any specific feature from nextest.

Issue #390, opened 2024-05-10
clp-s: Implicit wildcard behaviour is unintuitive


gibber9809:
### Request

For clp-s the way that we treat wildcards is unintuitive.

The rules are that if a user adds any explicit wildcards we don't add any implicit ones, and we only add implicit wildcards at the start and end if the predicate contains a space.

To make things simpler for users we should make the user use explicit wildcards for everything.

### Possible implementation

Instead of conditionally adding wildcards when processing a query we will simply not add implicit wildcards.

PR #1917, opened 2026-01-27
feat(presto-clp): Add support for custom S3 endpoint URLs and buckets.

20001020ycx:
<!-- markdownlint-disable MD012 -->

<!--
Set the PR title to a meaningful commit message that:

* is in imperative form.
* follows the Conventional Commits specification (https://www.conventionalcommits.org).
  * See https://github.com/commitizen/conventional-commit-types/blob/master/index.json for possible
    types.

Example:

fix: Don't add implicit wildcards ('*') at the beginning and the end of a query (fixes #390).
-->

# Description

<!-- Describe what this request will change/fix and provide any details necessary for reviewers. -->

  ## Summary                                                                                                                                                                                                 
                                                                                                                                                              
  Previously, the config script always constructed the S3 endpoint as `https://{bucket}.s3.{region}.amazonaws.com/`, which only works for AWS S3. This change allows configuring S3-compatible storage (e.g., MinIO) by reading the `endpoint_url` and `bucket` field from `clp-config.yaml`. 

# Checklist

<!-- Ensure each item below is satisfied and indicate so by inserting an `x` within each `[ ]`. -->

  breaking change.

# Validation performed

<!-- Describe what tests and validation you performed on the change. -->

- End to end test 
  -  Query: `SELECT CLP_GET_JSON_STRING() from clp.default.default limit 100`
  -  Set up: official clp package v0.8.0, OSS version of Presto, velox with changes in this PR, and a locally set minio. The below velox clp.properties is the one generated using the script updated in this PR.
  - The result is returned as expected.

CLP package's clp-config.yaml:
```
package:
  storage_engine: "clp-s"
  query_engine: "presto"

archive_output:
  storage:
    # type: "fs"
    type: "s3"
    staging_directory: "var/data/staged-archives"
    s3_config:
      region_code: "us-east-1"
      endpoint_url: "http://172.26.105.44:9000"
      bucket: "logs"
      key_prefix: "archives/"
      aws_authentication:
        type: "credentials"
        credentials:
          access_key_id: "minioadmin"
          secret_access_key: "minioadmin"
```

Velox's clp.properties:
```
  connector.name=clp                                                                                                                                                                                                             
  clp.storage-type=s3                                                                                                                                                                                                            
  clp.s3-auth-provider=clp_package                                                                                                                                                                                               
  clp.s3-end-point=http://172.26.105.44:9000                                                                                                                                                                                     
  clp.s3-bucket=logs                                                                                                                                                                                                
  clp.s3-access-key-id=minioadmin                                                                                                                                                                                                
  clp.s3-secret-access-key=minioadmin   
```


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Support for configuring a custom S3 endpoint for worker properties, with automatic fallback when none is provided.
  * S3 bucket is now exposed as an environment variable for worker configuration.

* **Bug Fixes**
  * Improved S3 endpoint resolution to avoid unreliable region-only derivation across regions/providers.

* **Chores**
  * Deployment initialization updated to align with the new S3 configuration approach.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/components/core/CMakeLists.txt
+++ b/components/core/CMakeLists.txt
@@ -575,6 +575,8 @@ set(SOURCE_FILES_unitTest
         src/clp/GrepCore.hpp
         src/clp/hash_utils.cpp
         src/clp/hash_utils.hpp
+        src/clp/SchemaSearcher.cpp
+        src/clp/SchemaSearcher.hpp
         src/clp/ir/constants.hpp
         src/clp/ir/EncodedTextAst.cpp
         src/clp/ir/EncodedTextAst.hpp
@@ -704,6 +706,11 @@ set(SOURCE_FILES_unitTest
         tests/clp_s_test_utils.cpp
         tests/clp_s_test_utils.hpp
         tests/LogSuppressor.hpp
+        tests/MockLogTypeDictionary.hpp
+        tests/MockVariableDictionary.hpp
+        tests/SchemaSearcherTest.hpp
+        tests/search_test_utils.cpp
+        tests/search_test_utils.hpp
         tests/TestOutputCleaner.hpp
         tests/test-BoundedReader.cpp
         tests/test-BufferedReader.cpp
@@ -731,6 +738,7 @@ set(SOURCE_FILES_unitTest
         tests/test-ParserWithUserSchema.cpp
         tests/test-query_methods.cpp
         tests/test-regex_utils.cpp
+        tests/test-SchemaSearcher.cpp
         tests/test-Segment.cpp
         tests/test-SQLiteDB.cpp
         tests/test-Stopwatch.cpp
@@ -789,5 +797,8 @@ if(CLP_ENABLE_TESTS)
             )
     target_compile_features(unitTest
             PRIVATE cxx_std_20
+            )
+    target_compile_definitions(unitTest
+            PRIVATE CLP_ENABLE_TESTS
             )
 endif()
--- a/components/core/src/clp/GrepCore.cpp
+++ b/components/core/src/clp/GrepCore.cpp
@@ -1,17 +1,12 @@
 #include "GrepCore.hpp"
 
 #include <cstddef>
-#include <set>
 #include <string>
 #include <string_view>
 
-#include <log_surgeon/Constants.hpp>
 #include <string_utils/string_utils.hpp>
 
-#include "ir/parsing.hpp"
-#include "LogSurgeonReader.hpp"
-#include "QueryToken.hpp"
-#include "StringReader.hpp"
+#include <clp/ir/parsing.hpp>
 
 using clp::ir::is_delim;
 using clp::string_utils::is_alphabet;
@@ -19,17 +14,6 @@ using clp::string_utils::is_wildcard;
 using std::string;
 
 namespace clp {
-namespace {
-/**
- * Wraps the tokens returned from the log_surgeon lexer, and stores the variable ids of the tokens
- * in a search query in a set. This allows for optimized search performance.
- */
-class SearchToken : public log_surgeon::Token {
-public:
-    std::set<int> m_type_ids_set;
-};
-}  // namespace
-
 bool GrepCore::get_bounds_of_next_potential_var(
         string const& value,
         size_t& begin_pos,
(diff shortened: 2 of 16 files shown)


tools/deployment/presto-clp/scripts/init.py
```diff
@@ -281,13 +281,16 @@ def _add_clp_s3_env_vars(
         s3_secret_access_key = _get_required_config_value(
             clp_config, f"{s3_credentials_key}.secret_access_key", clp_config_file_path
         )
+        s3_endpoint_url = _get_required_config_value(
+            clp_config, f"{s3_config_key}.endpoint_url", clp_config_file_path
+        )
```
gibber9809: Isn't the endpoint URL optional?

I think this needs to be `_get_config_value`, with some handling for the case where `s3_endpoint_url` is `NULL`? For context `s3_endpoint_url` is currently always null when the endpoint is actually s3 if I'm remembering correctly.

More specifically, looks like the right thing to do here might be to either:
1. Use the definition of s3_end_point value from the old version of the code here whenever `s3_endpoint_url` is null, or
2. Update `s3_utils.py` to cover this use-case -- I think we basically want what the `generate_s3_url` function does except
a. When the endpoint is null/s3 we want all of the output except the part with the object key
b. Otherwise we want all of the output except the part with the bucket and object key?

20001020ycx: Good catch, following the logic in s3_endpoint_url, here is the implemented logic:
- required config: bucket
- optional config: region and endpoint_url
- when endpoint_url is empty, constructs corresponding aws endpoint pending the existence of region.
- when endpoint_url is not empty, use the user provided url directly

tools/deployment/presto-clp/scripts/init.py
```diff
@@ -284,15 +284,38 @@ def _add_clp_s3_env_vars(
...
+
+    :param endpoint_url:
+    :param region_code: AWS region code, or None.
+    :param bucket_name: S3 bucket name (used to construct AWS S3 URL).
+    :return: The S3 endpoint URL.
+    """
+    if endpoint_url is not None:
+        return endpoint_url
+
+    if region_code is None:
```
junhaoliao, line 314: I understand this is needed to support non-AWS S3 endpoints. However, omitting a region is non-standard for AWS S3. Historically, some AWS SDKs would fall back to `us-east-1` when no region was specified, but this behavior should not be relied upon.

Should we add a log message when neither `endpoint_url` nor `region_code` is provided (here)?

tools/deployment/presto-clp/scripts/init.py
```diff
@@ -284,15 +284,38 @@ def _add_clp_s3_env_vars(
...
+    """
+    Returns the S3 endpoint URL, or constructs an AWS S3 URL if not provided.
+
+    :param endpoint_url:
+    :param region_code: AWS region code, or None.
+    :param bucket_name: S3 bucket name (used to construct AWS S3 URL).
+    :return: The S3 endpoint URL.
+    """
+    if endpoint_url is not None:
+        return endpoint_url
```
junhaoliao: is the URL guaranteed to end with `/`? if not, the URL format may not be consistent with others who end with `/`?

we should either normalize trailing slashes or document the expected format

tools/deployment/presto-clp/scripts/init.py
junhaoliao: e.g. 
Neither 'endpoint_url' nor 'region_code' specified in S3 config. Using region-less S3 endpoint which may not work for all buckets.

tools/deployment/presto-clp/scripts/init.py
```diff
@@ -284,15 +284,38 @@ def _add_clp_s3_env_vars(
...
     return True
 
 
+def _handle_s3_endpoint_url(
+    endpoint_url: str | None, region_code: str | None, bucket_name: str
+) -> str:
+    """
+    Returns the S3 endpoint URL, or constructs an AWS S3 URL if not provided.
+
+    :param endpoint_url:
```
junhaoliao:
```suggestion
    :param endpoint_url: Custom S3 endpoint URL.
```

tools/deployment/presto-clp/scripts/init.py
```diff
@@ -284,15 +284,38 @@ def _add_clp_s3_env_vars(
...
+    :param region_code: AWS region code, or None.
```
junhaoliao:
```suggestion
    :param region_code: AWS region code.
```

already reflected in the `| None` type

tools/deployment/presto-clp/scripts/init.py
```diff
@@ -21,6 +21,9 @@
 PRESTO_QUERY_MEMORY_RATIO = 0.5
 PRESTO_SYSTEM_MEMORY_RATIO = 0.9
 
+# S3 URL constant
+AWS_ENDPOINT = "amazonaws.com"
```
junhaoliao: how about `AWS_S3_DOMAIN`?

tools/deployment/presto-clp/scripts/init.py
```diff
@@ -284,15 +284,38 @@ def _add_clp_s3_env_vars(
...
     env_vars["PRESTO_WORKER_CLPPROPERTIES_S3_ACCESS_KEY_ID"] = s3_access_key_id
-    s3_end_point = f"https://{s3_bucket}.s3.{s3_region_code}.amazonaws.com/"
+    env_vars["PRESTO_WORKER_CLPPROPERTIES_S3_BUCKET"] = s3_bucket
     env_vars["PRESTO_WORKER_CLPPROPERTIES_S3_END_POINT"] = s3_end_point
     env_vars["PRESTO_WORKER_CLPPROPERTIES_S3_SECRET_ACCESS_KEY"] = s3_secret_access_key
 
     return True
 
 
+def _handle_s3_endpoint_url(
```
junhaoliao: "handle" isn't too meaningful in this context. how about `_resolve_s3_endpoint_url`

tools/deployment/presto-clp/scripts/init.py
20001020ycx: As suggested by @gibber9809 , this behavior is followed by [generate_s3_url](https://github.com/y-scope/clp/blob/01d35a672c2f6fb892912121800b7a85f7448b23/components/clp-py-utils/clp_py_utils/s3_utils.py#L257) in s3_utils.py so that CLP package always make consistent assumption.

I am okay with adding the log message when neither 'endpoint_url' nor 'region_code' is specified in S3 config as a warnning. But should we keep the same behavior as the changes in this PR?

And thanks for the insight on this S3 support!

tools/deployment/presto-clp/scripts/init.py
20001020ycx: This is a good point, the new code has reflected with always deleting trailing slashes when present. This is for aligning the velox convention as implemented in this [PR](https://github.com/y-scope/velox/pull/48)

tools/deployment/presto-clp/scripts/init.py
junhaoliao: Let's see what @gibber9809 decides on

tools/deployment/presto-clp/scripts/init.py
```diff
@@ -284,15 +284,38 @@ def _add_clp_s3_env_vars(
...
 
+def _resolve_s3_endpoint_url(
+    endpoint_url: str | None, region_code: str | None, bucket_name: str
+) -> str:
+    """
+    Returns the resolved S3 endpoint URL, or constructs an AWS S3 URL if not provided.
+
+    :param endpoint_url: Custom S3 endpoint URL.
+    :param region_code: AWS region code.
+    :param bucket_name: S3 bucket name (used to construct AWS S3 URL).
```
junhaoliao:
```suggestion
    :param bucket_name:
```

junhaoliao (dismissed):

junhaoliao (approved):
Let's wait for @gibber9809 's response before merging: https://github.com/y-scope/clp/pull/1917#discussion_r2734254388

tools/deployment/presto-clp/scripts/init.py
gibber9809: Sorry for slow response. Yeah logging a warning in this case sounds reasonable. It probably also makes sense to log the same warning during the package startup whenever 'endpoint_url' and 'region_code' are missing from the package config for the same reason.

gibber9809: I would be fine tackling this in a separate PR though that adds the warning both here and in the package startup script.

junhaoliao: @coderabbitai we will merge the current PR first and add the warning messages later. let's create an issue about adding a warning message both here and in the package startup script

20001020ycx:
Let's merge this PR until this [PR](https://github.com/y-scope/velox/pull/48) is merged. It is pending on Velox CI reviewing for publishing the right image to CLP package.

PR #1927, opened 2026-01-28
fix(clp-rust-utils): Make `api_server` and `log_ingestor` config values optional (fixes #1911).

### LINKED ISSUE #1911: Update Rust utils to handle nullable log_ingestor configuration
## Problem

The Rust utils cannot handle null values for the `log_ingestor` key in `clp_config.yaml` and fail with a parsing error.

## Impact

When the API server is enabled while `log_ingestor` is disabled by setting `log_ingestor: null`, the API server fails to start. This is particularly problematic because the current default configuration in Helm `values.yaml` has `log_ingestor: null`, causing the v0.8.0 chart to deploy an API server that continuously crashes and restarts.

## Solution

Update the Rust configuration parser to correctly handle nullable `log_ingestor` values, treating `null` as an indication that the log-ingestor feature should be disabled.

## References

- PR: https://github.com/y-scope/clp/pull/1856
- PR: https://github.com/y-scope/clp/pull/1882
- Related issue: #1881
- Comment: https://github.com/y-scope/clp/pull/1856#discussion_r2703116768
- Reported by: @junhaoliao

## Files Affected

- Rust configuration parser (likely in components utilizing clp_config.yaml)

hoophalab:
# Description

- Make api_server and log_ingestor fields optional in Config struct to allow  null values.
- Update all usages in api-server and log-ingestor to handle Option with explicit panics when config is missing.

# Checklist

  breaking change.

# Validation performed

1. Set `log_ingestor: null` in `clp-config.yaml`. The package starts correctly after the fix.
2. Successfully ingested logs using log ingestor and submited queries using the api server

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration error messaging, added a new error for missing API config, and added explicit fail-fast checks when required server/ingestor sections are absent.

* **Refactor**
  * Made server and ingestor configuration fields optional and adjusted default initializations to reflect potentially absent components.

* **Documentation**
  * Clarified panic/failure conditions and documented the new missing-configuration behavior.


components/api-server/src/bin/api_server.rs
```diff
@@ -24,10 +24,8 @@ fn read_config_and_credentials(
     args: &Args,
 ) -> anyhow::Result<(package::config::Config, package::credentials::Credentials)> {
     let config_path = std::path::Path::new(args.config.as_str());
-    let config: package::config::Config = yaml::from_path(config_path).context(format!(
-        "Config file {} does not exist",
-        config_path.display()
-    ))?;
+    let config: package::config::Config = yaml::from_path(config_path)
+        .context(format!("Cannot load config file {}", config_path.display()))?;
```
```suggestion
        .context(format!("cannot load config file {}", config_path.display()))?;
```

components/api-server/src/client.rs
```diff
@@ -130,8 +134,12 @@ impl Client {
             }
         }
         if search_job_config.max_num_results == 0 {
-            search_job_config.max_num_results =
-                self.config.api_server.default_max_num_query_results;
+            search_job_config.max_num_results = self
+                .config
+                .api_server
+                .as_ref()
```
LinZhihao-723: How about making a private member `get_api_server_config` to wrap:
```rust
self.config.api_server.as_ref.expect("...")
```
By doing this, we're assuming the config should always be `Some` for a valid client instance. That said, we should assert the API server config is not `None` in the factory function.

components/api-server/src/client.rs
```diff
@@ -102,11 +106,13 @@ impl Client {
...
         let mongo_client = mongodb::Client::with_uri_str(mongo_uri).await?;
 
-        Ok(Self {
+        let client = Self {
             config: config.clone(),
             mongodb_client: mongo_client,
             sql_pool,
-        })
+        };
+        client.get_api_server_config();
```
LinZhihao-723: I agree with the rabbit. I think we should make the check explicit and return as an error maybe?

components/api-server/src/client.rs
```diff
@@ -121,6 +127,10 @@ impl Client {
     ///
     /// * Forwards [`rmp_serde::to_vec_named`]'s return values on failure.
     /// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `self.config.api_server` is `None`.
```
LinZhihao-723: I don't think we need to document the implicit panics because it can easily get convoluted...
Let's drop them all except for `get_api_server_config`.

components/api-server/src/client.rs
```diff
@@ -442,6 +448,20 @@ impl Client {
         });
         Ok(mapped)
     }
+
+    /// Returns a reference to the API server configuration.
```
```suggestion
    /// # Returns
    ///
    /// A reference to the API server configuration.
```

PR #1948, opened 2026-02-02
fix(log-surgeon): Fix schema regex patterns in the heuristic schema file and `LogConverter`; Add a unit test to build a parser for the heuristic schema file.

davidlion:
# Description

#1273 updated log-surgeon which included a few syntax changes. Unfortunately, some locations that need updating were missed.

This PR:

* Updates the delimiter escape characters in the heuristic schema file (`components/core/config/schemas.txt`) and `LogConverter`.
* Updates the `LogConverter` time stamp pattern to use the `header` + `timestamp` capture group syntax.
* Groups part of the `LogConverter` time stamp pattern in a sub expression due to the change in alteration behaviour.
* Adds a unit test to build the heuristic schema file to change any syntax breaks (the accuracy of the heuristic schema will be tested in the integration tests).

Technically, this isn't a breaking change as the old syntax behaviour was already broken by #1273, this PR just updates the syntax.

Updating the syntax exposed what appears to be an issue with log-surgeon, as the schema variable `equals:.*=(?<var>.*[a-zA-Z0-9].*)` causes the parser to fail to finish building. The work around in this PR (`equals:[a-zA-Z0-9]+=(?<var>.*[a-zA-Z0-9].*)`) should cover the vast majority of cases as `[a-zA-Z0-9]+` should be very close to "any non-delimiter characters" (i.e. `.*`).

The issue in log surgeon is tracked by https://github.com/y-scope/log-surgeon/issues/207.

# Checklist

  breaking change.

# Validation performed

CI passing including new test. LogConverter tested manually.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timestamp recognition by adjusting parsing to more reliably locate timestamps.
  * Refined delimiter handling to accept additional literal characters.
  * Tightened key–value parsing to require an identifier before '=' for more accurate captures.

* **Tests**
  * Added test coverage for creating a log parser from configuration-based schemas.


components/core/config/schemas.txt
```diff
@@ -16,4 +16,4 @@ float:\-{0,1}[0-9]+\.[0-9]+
 // Dictionary variables
 hex:[a-fA-F]+
 hasNumber:.*\d.*
-equals:.*=(?<var>.*[a-zA-Z0-9].*)
+equals:[a-zA-Z0-9]+=(?<var>.*[a-zA-Z0-9].*)
```
gibber9809, line 19: If I'm understanding correctly, doesn't this stop this rule from working with most unicode?

Would be fine with this as a temporary workaround as long as we create an issue to track fixing this case.

davidlion: You're right, although afaik log-surgeon doesn't work properly with unicode in general.

I made https://github.com/y-scope/log-surgeon/issues/207 (I forgot to add it to the PR description sorry). I'm not sure if we also want to add an issue to clp to track the log surgeon issue (lmk if you'd like that).

gibber9809: As long as it's tracked somewhere that's fine with me.

components/core/tests/test-ParserWithUserSchema.cpp
```diff
@@ -147,6 +153,11 @@ TEST_CASE("Test creating log parser with delimiters", "[LALR1Parser][LogParser]"
     generate_log_parser(schema_file_path.string());
 }
 
+TEST_CASE("Test creating log parser from config schema", "[LALR1Parser][LogParser]") {
+    auto const schema_file_path = get_config_schema_files_dir() / "schemas.txt";
+    generate_log_parser(schema_file_path.string());
```
gibber9809:
```suggestion
    REQUIRE_NOTHROW(generate_log_parser(schema_file_path.string()));
```
Nit

gibber9809 (approved):
New changes LGTM and PR title is fine as well.

Also manually validated that the log-converter is still working for a few important cases.

I found another issue with the timestamp regex where it doesn't handle some timezones correctly, but that's just an existing issue with the way the regex is written, not anything to do with log-surgeon changes, so that can be addressed in a separate PR.

PR #1964, opened 2026-02-09
feat(api-server): Add DELETE endpoint to cancel query jobs; Treat `Cancelling` as a terminal state when polling results.

hoophalab:
# Description

This PR adds the ability to cancel previously submitted query jobs through the API server.

- Add a `DELETE /query/{search_job_id}` endpoint that submits a cancellation request for a search
  job by updating its status to `Cancelling` if the job is currently `Pending` or `Running`.
- Treat `Cancelling` as a terminal state in the query results polling loop so that the client stops
  waiting for results from a job that is being cancelled.
- Register the new endpoint in the OpenAPI documentation generated by `utoipa`.

# Checklist

  breaking change.

# Validation performed

1. Submit a job
```
curl -X POST http://localhost:3001/query \
  -d '{"query_string": "*", "dataset": "default", "buffer_results_in_mongodb": true}'
```
2. Cancel a query using `DELETE http://localhost:3001/query/10`
3. The status is set to `4` then `5` in `clp-db.query_jobs` table

## Summary by CodeRabbit

* **New Features**
  * Added an API endpoint to cancel in-progress search queries.

* **Behavior Changes**
  * Cancelling is now treated as a non-success terminal state; only succeeded jobs report as Succeeded.
  * Polling/backoff focuses on Running/Pending; Cancelling is surfaced as a failure when observed.

* **Error Handling**
  * Cancel returns Not Found when no cancellable job exists.

* **Documentation**
  * API docs updated to include the new cancel endpoint.

### DIFF
--- a/components/clp-py-utils/clp_py_utils/clp_config.py
+++ b/components/clp-py-utils/clp_py_utils/clp_config.py
@@ -774,7 +774,7 @@ class LogIngestor(BaseModel):
     host: DomainStr = "localhost"
     port: Port = 3002
     buffer_flush_timeout: PositiveInt = 300  # seconds
-    buffer_flush_threshold: PositiveInt = 256 * 1024 * 1024  # 256 MiB
+    buffer_flush_threshold: PositiveInt = 4096 * 1024 * 1024  # 4 GiB
     channel_capacity: PositiveInt = 10
     logging_level: LoggingLevelRust = "INFO"
 
--- a/components/clp-rust-utils/src/clp_config/package/config.rs
+++ b/components/clp-rust-utils/src/clp_config/package/config.rs
@@ -239,7 +239,7 @@ impl Default for LogIngestor {
             host: "localhost".to_owned(),
             port: 3002,
             buffer_flush_timeout_sec: 300,
-            buffer_flush_threshold: 256 * 1024 * 1024,
+            buffer_flush_threshold: 4096 * 1024 * 1024, // 4 GiB
             channel_capacity: 10,
             logging_level: "INFO".to_owned(),
         }
--- a/components/core/src/clp/ReadOnlyMemoryMappedFile.cpp
+++ b/components/core/src/clp/ReadOnlyMemoryMappedFile.cpp
@@ -2,36 +2,43 @@
 
 #include <sys/mman.h>
 
-#include <string>
+#include <cerrno>
 #include <string_view>
+#include <system_error>
 
-#include "ErrorCode.hpp"
+#include <ystdlib/error_handling/Result.hpp>
+
 #include "FileDescriptor.hpp"
+#include "TraceableException.hpp"
 
 namespace clp {
-ReadOnlyMemoryMappedFile::ReadOnlyMemoryMappedFile(std::string_view path) {
-    FileDescriptor const fd{path, FileDescriptor::OpenMode::ReadOnly};
-    auto const file_size{fd.get_size()};
-    if (0 == file_size) {
-        // `mmap` doesn't allow mapping an empty file, so we don't need to call it.
-        return;
+auto ReadOnlyMemoryMappedFile::create(std::string_view path)
+        -> ystdlib::error_handling::Result<ReadOnlyMemoryMappedFile> {
+    try {
+        FileDescriptor const fd{path, FileDescriptor::OpenMode::ReadOnly};
+        auto const file_size{fd.get_size()};
+        if (0 == file_size) {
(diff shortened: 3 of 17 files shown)


components/api-server/src/client.rs
```diff
@@ -218,6 +221,32 @@ impl Client {
...
+    /// (e.g., if the job was already completed, no rows would be updated).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
+    pub async fn cancel_search_job(&self, search_job_id: u64) -> Result<(), ClientError> {
+        sqlx::query(&format!(
+            "UPDATE `{QUERY_JOBS_TABLE_NAME}` SET status = ? WHERE id = ? AND status IN (?, ?)"
```
LinZhihao-723, line 239: Just realized this: I think we can leverage `const_format` to format SQL at compile time. If this sounds good to u, we can ask coderabbitai to create an issue for it.

components/api-server/src/client.rs
```diff
@@ -218,6 +221,32 @@ impl Client {
             .map(|s| SearchResultStream::Mongo { inner: s })
     }
 
+    /// Submits a cancellation request for a search job.
+    ///
+    /// # Returns
+    ///
+    /// Returns `Ok(())` on success, regardless of whether the job was actually updated
+    /// (e.g., if the job was already completed, no rows would be updated).
```
```suggestion
    /// Returns `Ok(())` on success, regardless of whether the job was actually updated  (e.g., if
    /// the job was already completed, no rows would be updated).
```

components/api-server/src/routes.rs
```diff
@@ -180,6 +181,42 @@ async fn query_results(
...
+)]
+async fn cancel_query(
+    State(client): State<Client>,
+    Path(search_job_id): Path<u64>,
+) -> Result<StatusCode, HandlerError> {
+    tracing::info!("Cancelling search job ID: {}", search_job_id);
+    match client.cancel_search_job(search_job_id).await {
+        Ok(()) => {
+            tracing::info!(
+                "Successfully submitted cancellation for search job ID: {}",
```
```suggestion
                "Successfully submitted cancellation request for search job ID: {}",
```

LinZhihao-723 (changes requested):
One issue is, if we cancel a search job that doesn't exist, it won't complain and return on success. Actually, it's the same when attempting to retrieve results from a nonexistent query. Shall we improve this as well (in a future PR)?

LinZhihao-723:
> One issue is, if we cancel a search job that doesn't exist, it won't complain and return on success. Actually, it's the same when attempting to retrieve results from a nonexistent query. Shall we improve this as well (in a future PR)?

Confirmed with @kirkrodrigues  that we should probably try to fix this. I think the fix should be just adding a SQL query to check whether the job exists.

components/api-server/src/client.rs
```diff
@@ -218,6 +221,33 @@ impl Client {
             .map(|s| SearchResultStream::Mongo { inner: s })
     }
 
+    /// Submits a cancellation request for a search job.
+    ///
```
LinZhihao-723, line 225:
```suggestion
    ///
    /// # Returns
    ///
    /// `Ok(())` on success.
    ///
```

components/api-server/src/error.rs
```diff
@@ -24,6 +24,9 @@ pub enum ClientError {
 
     #[error("api_server configuration is missing")]
     ConfigIsNone,
+
+    #[error("No cancellable search job found")]
+    SearchJobNotFound,
```
LinZhihao-723:
```suggestion
    #[error("Search job ID found: {0}")]
    SearchJobIdNotFound(u64),
```
How about:
- Update the description accordingly, since it might not just be used for cancellation. This should make the error message general, and it's reusable in result retrieval.
- Also store the ID of the search job.

docs/src/_static/generated/api-server-openapi.json
LinZhihao-723, line 1: This is outdated.

LinZhihao-723 (changes requested):
I guess we need to update `get_status` to return `SearchJobNotFound` as well, but presumably in another PR.

components/api-server/src/error.rs
hoophalab: I update the doc and add the id, but keep the name. SearchJobNotFound sounds more concise.

hoophalab:
> I guess we need to update get_status to return SearchJobNotFound as well, but presumably in another PR.

@coderabbitai Can you create an issue?

PR #1965, opened 2026-02-09
feat(clp-package)!: Increase `buffer_flush_threshold` default from 256MiB to 4GiB.

jonathan-imanu:
# Description

The current log-ingestor buffering is too conservative, making the submitted compression jobs not buffer enough input logs. This PR increases the `buffer_flush_threshold` to be 4.096 GiB from 256 MiB (a 16x increase).

# Checklist

  breaking change.

# Validation performed

- Verified that the log-ingestor builds and works as expected.

## Summary by CodeRabbit

* **Chores**
  * Increased default log ingestor buffer flush threshold from 256 MiB to 4 GiB.
  * Bumped release/chart version for deployment.

### DIFF
--- a/components/webui/client/src/pages/IngestPage/Jobs/index.tsx
+++ b/components/webui/client/src/pages/IngestPage/Jobs/index.tsx
@@ -36,7 +36,7 @@ const Jobs = () => {
                 dataSource={jobs}
                 pagination={false}
                 rowKey={(record) => record.key}
-                scroll={{y: 400, x: true}}
+                scroll={{y: 400, x: 1}}
                 tableLayout={"fixed"}/>
         </DashboardCard>
     );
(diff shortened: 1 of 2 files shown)


components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao, line 1: btw, this change might be BREAKING cuz now we consume more memory? on memory-constrained systems (VPC / cgroupv2 resource-constraint containers), if users don't increase the memory limit of their resources, the log-ingestor may crash?

either way, i feel the scope of the PR should be `feat` than `chore` cuz the PR does bring benefits to users

LinZhihao-723: I'm not sure if this should be categorized as a BREAKING change:
- The buffer threshold is the total size of the metadata buffered, not the actual buffer size. I suspect this would cause visible difference to users. This PR doesn't imply the memory usage will increase from 256MB to 4GB (per buffer).
- Since we didn't change the config interface, this is not "breaking". But we probably need to draw a line on whether changing the default value should be considered as breaking.

PR #1989, opened 2026-02-13
feat(log-ingestor): Allow an SQS listener job to spawn multiple concurrent tasks to consume notifications from the same queue (resolves #1977).

### LINKED ISSUE #1977: [log-ingestor] Support multiple SQS listening coroutines in an SQS listener job.
### Request

The current SQS listener job (in log-ingestor) only uses one coroutine to listen to the SQS. As requested by our user, the ingestion throughput expected would be 15k SQS messages per minute. To adapt this throughput, we need more coroutines to process SQS messages concurrently inside one SQS listener job.

### Possible implementation

* Allow users to specify the number of coroutines per job.
* All coroutines execute the same ingestion task, while the tasks are the multi-source (sender) of the same buffer for compression.

LinZhihao-723:
# Description

This PR adds support for spawning multiple coroutines to process a single SQS queue. This enables higher message-processing throughput.

To adapt the multi-listener design, this PR also exposes the max waiting time as a configurable option so that coroutines can avoid frequent void responses.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests to cover multiple listener tasks.
* [x] Ensure invalid concurrent config will be rejected.
* [x] Ensure that with 16 tasks per job, the ingestion speed is significantly improved.

## Summary by CodeRabbit

* **New Features**
  * SQS listener supports configurable concurrency and wait time with per-task validation and improved per-task logging.

* **Bug Fixes / Behaviour**
  * Listener shutdown now completes cleanly even when individual tasks fail; invalid listener configs yield clear Bad Request responses.

* **Tests**
  * Tests expanded to exercise multiple concurrency levels and noise-object scenarios.

* **Chores**
  * Client wrappers made clonable and API surfaces adjusted to accept validated config references.

### DIFF
--- a/components/core/src/clp_s/indexer/CMakeLists.txt
+++ b/components/core/src/clp_s/indexer/CMakeLists.txt
@@ -37,6 +37,8 @@ set(
         ../../clp/ffi/SchemaTree.cpp
         ../../clp/ffi/SchemaTree.hpp
         ../../clp/ffi/StringBlob.hpp
+        ../../clp/FileDescriptor.cpp
+        ../../clp/FileDescriptor.hpp
         ../../clp/FileReader.cpp
         ../../clp/FileReader.hpp
         ../../clp/GlobalMetadataDBConfig.cpp
@@ -75,6 +77,8 @@ set(
         ../../clp/time_types.hpp
         ../../clp/TraceableException.hpp
         ../../clp/type_utils.hpp
+        ../../clp/utf8_utils.cpp
+        ../../clp/utf8_utils.hpp
         ../../clp/VariableDictionaryEntryReq.hpp
         ../../clp/VariableDictionaryReaderReq.hpp
         ../../clp/VariableDictionaryWriterReq.hpp
@@ -92,6 +96,8 @@ set(
         ../FileReader.hpp
         ../FileWriter.cpp
         ../FileWriter.hpp
+        ../FloatFormatEncoding.cpp
+        ../FloatFormatEncoding.hpp
         ../InputConfig.cpp
         ../InputConfig.hpp
         ../PackedStreamReader.cpp
--- a/components/log-ingestor/src/compression/compression_job_submitter.rs
+++ b/components/log-ingestor/src/compression/compression_job_submitter.rs
@@ -165,9 +165,17 @@ async fn submit_clp_compression_job_and_wait_for_completion(
                 match status {
                     CompressionJobStatus::Failed => {
                         tracing::error!(
-                            "Compression job {} failed. Status message: {:?}",
+                            "Compression job {} failed. Status message: {}",
                             job_id,
-                            status_message
+                            status_message.as_deref().unwrap_or("None")
+                        );
+                        return;
+                    }
+                    CompressionJobStatus::Killed => {
+                        tracing::error!(
+                            "Compression job {} was killed. Status message: {}",
+                            job_id,
+                            status_message.as_deref().unwrap_or("None")
                         );
                         return;
                     }


components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -37,6 +37,9 @@ pub enum Error {
     #[error("Custom endpoint URL not supported: {0}")]
     CustomEndpointUrlNotSupported(String),
 
+    #[error("Invalid `num_concurrent_listener_tasks`: {0}`")]
```
hoophalab:
```suggestion
    #[error("Invalid `num_concurrent_listener_tasks`: {0}")]
```

components/clp-rust-utils/src/job_config/ingestion.rs
```diff
@@ -50,6 +50,30 @@ pub mod s3 {
...
+        pub num_concurrent_listener_tasks: u16,
+
+        /// The long polling wait time, in seconds, for receiving messages from the SQS queue.
+        ///
+        /// This value controls how long an SQS `ReceiveMessage` call will wait for a message to
+        /// arrive before returning a response. Enabling long polling can reduce empty responses
+        /// and lower overall request costs.
+        ///
+        /// AWS SQS enforces a maximum wait time of 20 seconds. Any configured value greater than
+        /// 20 seconds will be truncated to 20 seconds.
```
hoophalab: Coderabbit is probably correct? We reject values > 20 rather than truncate them.
```suggestion
        /// AWS SQS enforces a maximum wait time of 20 seconds.
```

components/log-ingestor/src/ingestion_job_manager.rs
hoophalab:
```suggestion
    #[error("Invalid `num_concurrent_listener_tasks`: {0}")]
```

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -154,66 +160,140 @@ impl<SqsClientManager: AwsClientManagerType<Client>> Task<SqsClientManager> {
...
+        sqs_client_manager: &SqsClientManager,
+        config: &SqsListenerConfig,
+        sender: &mpsc::Sender<ObjectMetadata>,
+    ) -> Self {
+        let mut task_handles = Vec::with_capacity(config.num_concurrent_listener_tasks as usize);
+        assert!(
+            config.num_concurrent_listener_tasks != 0,
+            "The number of concurrent listener tasks must be a positive integer. The upper level \
+             caller should ensure the configuration is valid."
+        );
```
hoophalab: assert before create Vec?
```suggestion
        assert!(
            config.num_concurrent_listener_tasks != 0,
            "The number of concurrent listener tasks must be a positive integer. The upper level \
             caller should ensure the configuration is valid."
        );
        let mut task_handles = Vec::with_capacity(config.num_concurrent_listener_tasks as usize);
```

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -154,66 +160,140 @@ impl<SqsClientManager: AwsClientManagerType<Client>> Task<SqsClientManager> {
...
+            let task_handle = TaskHandle::spawn(task, id);
+            task_handles.push(task_handle);
+            tracing::info!(job_id = ? id, task_id = ? task_id, "Spawned SQS listener task.");
+        }
+
+        Self { id, task_handles }
+    }
+
+    /// Shuts down and waits for the underlying tasks to complete.
+    pub async fn shutdown_and_join(self) {
```
hoophalab, line 255: Should we still return a `Result<()>` to keep consistent with s3_scanner? Just return the last Err if multiple tasks exit with `Err`s?

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -153,6 +159,12 @@ impl IngestionJobManagerState {
                  URL: {endpoint_url}"
             )));
         }
+        // NOTE: This limit must be kept in sync with the limit in the config definition.
+        if config.num_concurrent_listener_tasks < 1 || config.num_concurrent_listener_tasks > 32 {
+            return Err(Error::InvalidNumConcurrentListenerTasks(
+                config.num_concurrent_listener_tasks,
+            ));
+        }
```
hoophalab: Can we move this check to the top of this function so that it returns earlier?

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -154,66 +160,140 @@ impl<SqsClientManager: AwsClientManagerType<Client>> Task<SqsClientManager> {
...
+                        }
+                    }
+                }
+                Err(err) => {
+                    tracing::warn!(
+                        error = ? err,
+                        job_id = ? self.id,
+                        task_id = ? task_id,
+                        "SQS listener task failed to cancel."
+                    );
```
hoophalab: 1, In the error messages, "failed to cancel" is misleading. join_handle.await returns an Err if the task panics
2. Maybe we can make a better use of match
```suggestion
            match task_handle.join_handle.await {
                Ok(Ok(())) => {
                    tracing::info!(
                        job_id = ? self.id,
                        task_id = ? task_id,
                        "SQS listener task completed successfully."
                    );
                }
                Ok(Err(_)) => {
                    // We don't need to log the error here because the underlying task will
                    // log it.
                    tracing::warn!(
                        job_id = ? self.id,
                        task_id = ? task_id,
                        "SQS listener task completed with an error."
                    );
                }
                Err(err) => {
                    tracing::warn!(
                        error = ? err,
                        job_id = ? self.id,
                        task_id = ? task_id,
                        "SQS listener task panicked."
                    );
                }
            }
```

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -154,66 +160,140 @@ impl<SqsClientManager: AwsClientManagerType<Client>> Task<SqsClientManager> {
...
+            config.num_concurrent_listener_tasks != 0,
+            "The number of concurrent listener tasks must be a positive integer. The upper level \
+             caller should ensure the configuration is valid."
+        );
+        for task_id in 0..config.num_concurrent_listener_tasks {
+            let task = Task {
+                sqs_client_manager: sqs_client_manager.clone(),
+                config: config.clone(),
+                sender: sender.clone(),
+                id: task_id as TaskId,
```
hoophalab:
```suggestion
                id: TaskId::from(task_id),
```

components/log-ingestor/src/ingestion_job/sqs_listener.rs
```diff
@@ -154,66 +160,140 @@ impl<SqsClientManager: AwsClientManagerType<Client>> Task<SqsClientManager> {
...
+    /// domain constraints—such as upper bounds on the number of concurrent listener tasks—is the
+    /// responsibility of the caller.
+    #[must_use]
+    pub fn spawn<SqsClientManager: AwsClientManagerType<Client>>(
+        id: Uuid,
+        sqs_client_manager: &SqsClientManager,
+        config: &SqsListenerConfig,
+        sender: &mpsc::Sender<ObjectMetadata>,
+    ) -> Self {
+        let mut task_handles = Vec::with_capacity(config.num_concurrent_listener_tasks as usize);
```
hoophalab:
```suggestion
        let mut task_handles = Vec::with_capacity(usize::from(config.num_concurrent_listener_tasks));
```

hoophalab (commented):
Some nitpicks.

Validations:
1. Successfully ingested logs from AWS SQS
2. `tests:rust-all` passes

components/log-ingestor/src/ingestion_job/sqs_listener.rs
LinZhihao-723: I will update S3 scanner to not return an error. With DB features added, there's no need to return an error in this call anymore.
@coderabbitai can you create an issue for it?

components/log-ingestor/src/ingestion_job_manager.rs
LinZhihao-723: I think it's cleaner if we group all config validation logic together. Are you suggesting we should return before getting a clone of the base config?

components/clp-rust-utils/src/job_config/ingestion.rs
LinZhihao-723: This is tricky. With our current set up, there's no way to enforce the config to be valid values all the way down to the actual listener job spawn. That means we need to do validations manually. However, this is different from the other two validations we perform at present:
- The other two validations we perform are on the top-level, however it won't affect the actual ingestion job execution if an "invalid" value is given.
  - A custom endpoint won't propagate into the SQS job execution, since it's handled by the client manager.
  - A number larger than 32 given to the job also won't affect the execution. The SQS job can handle any arbitrary number of coroutines; it's our top-level decision to not allow a number > 32.
-  The wait time is different. A wait time larger than 20 will cause the operation to fail. Since there's no way to enforce config validation all the way down to the SQS job execution, we need to truncate inside the actual task to make sure it's under 20 anyway. That said, I think as long as we clearly document this truncation in the OpenAPI schema, it should be fair.

I can add the logic to reject >20 sec config on the top-level, but I will keep the truncation logic inside the SQS job just in case.

hoophalab: Maybe we should just adopt c philosophy: if the input isn't valid, then the behavior is undefined. You cannot validate everything in every function/using types.

So we should just reject >20 in routes.rs, and add a debug assert when receive message

LinZhihao-723: I don't agree on the C philosophy in general, lol. UB is the root of evil.
Actually, I have a better idea for handling this. We can wrap the validated config with a special type and pass this type all the way down to the ingestion job.

hoophalab (approved):

components/clp-rust-utils/src/job_config/ingestion.rs
hoophalab: Type is not the solution to everything. Even tokio implementation has undefined behaviors in their documents.

PR #1991, opened 2026-02-13
feat(docker): Support corporate proxies and package mirror overrides for base image builds.

jackluo923:
# Description

Docker base image builds fail in corporate environments where TLS-intercepting proxies (e.g., Zscaler, Fortinet, Palo Alto) replace upstream SSL certificates with the organization's own CA. Package managers (dnf, apk, apt, pip) reject the proxy's certificate because it isn't in the container's trust store.

This PR adds proper corporate proxy support to all 6 base image variants by:

1. **CA certificate injection** — `build.sh` auto-detects the host's CA bundle and copies it into the Docker build context. An in-container script (`setup-corporate-proxy.sh`) installs it into the system trust store. When no corporate CA is present, this is a no-op.

2. **Python/pip trust store override** — `ENV` vars (`SSL_CERT_FILE`, `CURL_CA_BUNDLE`, `REQUESTS_CA_BUNDLE`, `PIP_CERT`) point pip/pipx at the system trust store instead of certifi's bundled CA store.

3. **Proxy env var forwarding** — `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, `ALL_PROXY` are forwarded as Docker build args.

4. **Package mirror overrides** — Each distro supports an env var to override default mirrors (for organization-internal or regional mirrors):
   - `DNF_MIRROR_BASE_URL` for manylinux_2_28 (AlmaLinux) and centos-stream-9
   - `APK_MIRROR_URL` for musllinux_1_2 (Alpine)
   - `APT_MIRROR_URL` for ubuntu-jammy

5. **Docker network mode** — Auto-enables `--network host` when proxy URLs point to localhost/127.0.0.1. Explicit override via `DOCKER_NETWORK` env var.

6. **`DOCKER_PULL=true`** — Opt-in flag to force-pull the latest base image.

7. **Error hints** — On `install-all.sh` failure, displays a helpful banner suggesting mirror override if the error is network-related.

All changes are backward-compatible. Without proxy env vars or mirror overrides, builds work identically to before.

# Checklist

# Validation performed

- Built `manylinux_2_28-aarch64` image successfully with mirror override (`DNF_MIRROR_BASE_URL=<url> ./build.sh`)
- Verified `setup-corporate-proxy.sh` is a no-op when no CA bundle is provided (CI-safe — the GitHub Actions workflow uses `docker/build-push-action` directly, bypassing `build.sh`)
- Verified error hint banner displays on `install-all.sh` failure
- Verified `--network host` is only added when proxy URLs contain localhost/127.0.0.1

## Release Notes

* **New Features**
  * Added corporate proxy support to Docker image builds with automatic CA certificate handling
  * Introduced optional mirror URL overrides for package managers (DNF, APK, APT) to support restricted network environments
  * Added support for custom CA certificates to enable TLS verification behind corporate proxies

* **Improvements**
  * Enhanced installation error handling with actionable guidance when package installation fails

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -891,6 +891,7 @@ dependencies = [
  "hex",
  "non-empty-string",
  "num_enum",
+ "regex",
  "rmp-serde",
  "secrecy",
  "serde",
@@ -2767,9 +2768,9 @@ dependencies = [
 
 [[package]]
 name = "regex"
-version = "1.12.2"
+version = "1.12.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "843bc0191f75f3e22651ae5f1e72939ab2f72a4bc30fa80a066bd66edefc24d4"
+checksum = "e10754a14b9137dd7b1e3e5b0493cc9171fdd105e0ab477f51b72e7f3ac0e276"
 dependencies = [
  "aho-corasick",
  "memchr",
--- a/components/api-server/src/client.rs
+++ b/components/api-server/src/client.rs
@@ -483,6 +483,52 @@ impl Client {
         Ok(mapped)
     }
 
+    /// Retrieves timestamp column names for a given dataset.
+    ///
+    /// # Returns
+    ///
+    /// A vector of timestamp column name strings on success.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`ClientError::InvalidDatasetName`] if the dataset name contains invalid characters.
+    /// * [`ClientError::DatasetNotFound`] if the dataset's column metadata table doesn't exist.
+    /// * Forwards [`sqlx::query::Query::fetch_all`]'s return values on failure.
+    pub async fn get_timestamp_column_names(
+        &self,
+        dataset_name: &str,
+    ) -> Result<Vec<String>, ClientError> {
+        // Must be kept in sync with `NodeType::Timestamp` in
+        // `components/core/src/clp_s/SchemaTree.hpp`.
+        const TIMESTAMP_NODE_TYPE: i8 = 14;
+        // MySQL error number for "Table doesn't exist".
+        const MYSQL_TABLE_NOT_FOUND: u16 = 1146;
+
+        if !clp_rust_utils::dataset::VALID_DATASET_NAME_REGEX.is_match(dataset_name) {
+            return Err(ClientError::InvalidDatasetName);
+        }
+        let table_name = format!("clp_{dataset_name}_column_metadata");
+        let names: Vec<String> =
+            sqlx::query_scalar(&format!("SELECT name FROM `{table_name}` WHERE type = ?"))
+                .bind(TIMESTAMP_NODE_TYPE)
+                .fetch_all(&self.sql_pool)
+                .await
+                .map_err(|err| {
+                    if let sqlx::Error::Database(db_err) = &err
+                        && let Some(mysql_err) =
+                            db_err.try_downcast_ref::<sqlx::mysql::MySqlDatabaseError>()
+                        && mysql_err.number() == MYSQL_TABLE_NOT_FOUND
+                    {
+                        return ClientError::DatasetNotFound(dataset_name.to_owned());
+                    }
+                    err.into()
+                })?;
+
+        Ok(names)
+    }
+
     /// # Returns
     ///
     /// A reference to the API server configuration.
--- a/components/api-server/src/error.rs
+++ b/components/api-server/src/error.rs
@@ -27,6 +27,12 @@ pub enum ClientError {
 
     #[error("Search job not found: {0}")]
     SearchJobNotFound(u64),
+
+    #[error("Invalid dataset name")]
+    InvalidDatasetName,
+
+    #[error("Dataset not found: {0}")]
+    DatasetNotFound(String),
 }
 
 /// Empty trait to mark errors that indicate malformed data.
--- a/components/api-server/src/routes.rs
+++ b/components/api-server/src/routes.rs
@@ -36,6 +36,10 @@ pub fn from_client(client: Client) -> Result<axum::Router, serde_json::Error> {
         .routes(routes!(query))
         .routes(routes!(query_results))
         .routes(routes!(cancel_query))
+        .route(
+            "/column_metadata/{dataset_name}/timestamp",
+            get(get_timestamp_column_names),
+        )
         .with_state(client)
         .split_for_parts();
     let api_json = api.to_json()?;
@@ -221,6 +225,24 @@ async fn cancel_query(
     }
 }
 
+async fn get_timestamp_column_names(
+    State(client): State<Client>,
+    Path(dataset_name): Path<String>,
+) -> Result<Json<Vec<String>>, HandlerError> {
+    let names = client
+        .get_timestamp_column_names(&dataset_name)
+        .await
+        .map_err(|err| {
+            tracing::error!(
+                "Failed to get timestamp column names for dataset '{}': {:?}",
+                dataset_name,
+                err
+            );
+            HandlerError::from(err)
+        })?;
+    Ok(Json(names))
+}
+
 /// Generic errors for request handlers.
 #[derive(Error, Debug)]
 enum HandlerError {
@@ -228,6 +250,8 @@ enum HandlerError {
     InternalServer,
     #[error("Not found")]
     NotFound,
+    #[error("Bad request: {0}")]
+    BadRequest(String),
 }
 
 impl From<axum::Error> for HandlerError {
@@ -239,7 +263,8 @@ impl From<axum::Error> for HandlerError {
 impl From<ClientError> for HandlerError {
     fn from(err: ClientError) -> Self {
         match err {
-            ClientError::SearchJobNotFound(_) => Self::NotFound,
+            ClientError::SearchJobNotFound(_) | ClientError::DatasetNotFound(_) => Self::NotFound,
+            ClientError::InvalidDatasetName => Self::BadRequest(format!("{err}")),
             _ => Self::InternalServer,
         }
     }
@@ -251,6 +276,7 @@ impl IntoResponse for HandlerError {
         match self {
             Self::NotFound => StatusCode::NOT_FOUND.into_response(),
             Self::InternalServer => StatusCode::INTERNAL_SERVER_ERROR.into_response(),
+            Self::BadRequest(msg) => (StatusCode::BAD_REQUEST, msg).into_response(),
         }
     }
 }
--- a/components/clp-rust-utils/Cargo.toml
+++ b/components/clp-rust-utils/Cargo.toml
@@ -10,6 +10,7 @@ aws-sdk-sqs = "1.92.0"
 brotli = "8.0.2"
 num_enum = "0.7.5"
 non-empty-string = { version = "0.2.6", features = ["serde"] }
+regex = "1.12.3"
 rmp-serde = "1.3.1"
 secrecy = { version = "0.10.3", features = ["serde"] }
 serde = { version = "1.0.228", features = ["derive"] }
--- a/components/clp-rust-utils/src/dataset.rs
+++ b/components/clp-rust-utils/src/dataset.rs
@@ -1,0 +1,6 @@
+use std::sync::LazyLock;
+
+use regex::Regex;
+
+pub static VALID_DATASET_NAME_REGEX: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9_]+$").unwrap());
--- a/components/clp-rust-utils/src/lib.rs
+++ b/components/clp-rust-utils/src/lib.rs
@@ -1,6 +1,7 @@
 pub mod aws;
 pub mod clp_config;
 pub mod database;
+pub mod dataset;
 mod error;
 pub mod job_config;
 pub mod logging;
--- a/components/core/cmake/Options/options.cmake
+++ b/components/core/cmake/Options/options.cmake
@@ -95,6 +95,12 @@ option(
     ON
 )
 
+option(
+    CLP_BUILD_CLP_S_ENABLE_CURL
+    "Include libcurl support for clp-s."
+    ON
+)
+
 # Validates that the `CLP_BUILD_` options required by `TARGET_CLP_BUILD_OPTION` are `ON`.
 #
 # @param {string} TARGET_CLP_BUILD_OPTION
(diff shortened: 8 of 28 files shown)


jackluo923:
Addressed both review comments:

1. **Ubuntu Dockerfile sed patterns** — Updated to use `https\?://` to match both http and https protocols for future compatibility.

2. **`detect_ca_bundle` SSL_CERT_FILE support** — Now checks `${SSL_CERT_FILE:-}` as the first priority before scanning hardcoded paths, so user/corporate CA bundle overrides are respected.

jackluo923:
Addressed the second round of review comments:

1. **`git rev-parse` stderr suppression** — Applied. Added `2>&1` to suppress error output when building from a source tarball outside a git repo. (4372040)

2. **sed `|` delimiter collision in APK_MIRROR_URL** — Dismissed. The pipe character `|` is not a valid URL character, so this scenario cannot occur in practice. Adding a guard would be over-defensive for an impossible input.

components/core/tools/docker-images/proxy-lib.sh
```diff
@@ -0,0 +1,150 @@
...
+    )
+
+    local has_localhost_proxy=false
+    for var in "${proxy_vars[@]}"; do
+        if [[ -n "${!var:-}" ]]; then
+            _build_cmd+=(--build-arg "${var}=${!var}")
+            if [[ "${!var}" =~ (localhost|127\.0\.0\.1) ]]; then
+                has_localhost_proxy=true
+            fi
+        fi
```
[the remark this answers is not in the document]
jackluo923: Applied. Regex now requires `://` prefix (avoids substring false-matches) and covers IPv6 loopback `[::1]`. (315eb4c)

components/core/tools/docker-images/proxy-lib.sh
```diff
@@ -0,0 +1,153 @@
...
+
+    local ca_bundle
+    if ca_bundle="$(detect_ca_bundle)"; then
+        echo "Corporate proxy support: copying CA bundle from ${ca_bundle}"
+        cp "$ca_bundle" "$dest"
+    else
+        echo "Corporate proxy support: no CA bundle found on host (no-op)"
+        touch "$dest"
+    fi
+}
```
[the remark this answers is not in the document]
jackluo923: Not needed — all `build.sh` callers run with `set -o errexit`, so a failing `cp` already terminates the script immediately with a non-zero exit. Adding an explicit check would be redundant.

components/core/tools/scripts/lib_install/setup-corporate-proxy.sh
```diff
@@ -0,0 +1,47 @@
...
+    update-ca-certificates
+    echo "setup-corporate-proxy: updated trust store via update-ca-certificates."
+else
+    echo "setup-corporate-proxy: WARNING: no recognized CA trust update tool found." >&2
+    exit 1
+fi
+
+# Remove the staging file — it's no longer needed after installation into the
+# trust store.
+rm "$ca_cert"
```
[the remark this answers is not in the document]
jackluo923: False positive — the file already ends with a trailing newline (`0a`). Verified with `xxd`.

components/core/tools/docker-images/proxy-lib.sh
```diff
@@ -0,0 +1,153 @@
...
+
+    add_proxy_build_args "$1"
+
+    if [[ -n "$mirror_var" ]] && [[ -n "${!mirror_var:-}" ]]; then
+        _build_cmd_ref+=(--build-arg "${mirror_var}=${!mirror_var}")
+    fi
+
+    if [[ "${DOCKER_PULL:-false}" == "true" ]]; then
+        _build_cmd_ref+=(--pull)
+    fi
```
junhaoliao: we already set this --pull flag - why do we want to make this opt-in?

components/core/tools/docker-images/proxy-lib.sh
```diff
@@ -0,0 +1,153 @@
...
+
+# Finalizes the build command array and executes the Docker build.
+# Adds proxy build args, an optional mirror override, DOCKER_PULL support,
+# git metadata labels, then runs the build.
+# Arguments:
+#   $1 - name of the bash array variable holding the build command
+#   $2 - script_dir (used for git label metadata)
+#   $3 - (optional) name of the mirror override env var (e.g., DNF_MIRROR_BASE_URL)
+finalize_build() {
+    local -n _build_cmd_ref=$1
```
junhaoliao: how about renaming as just `_build_cmd`?

components/core/tools/docker-images/proxy-lib.sh
```diff
@@ -0,0 +1,153 @@
...
+    if [[ "${DOCKER_PULL:-false}" == "true" ]]; then
+        _build_cmd_ref+=(--pull)
+    fi
+
+    if command -v git >/dev/null \
+        && git -C "$script_dir" rev-parse --is-inside-work-tree >/dev/null 2>&1
+    then
+        _build_cmd_ref+=(
+            --label "org.opencontainers.image.revision=$(git -C "$script_dir" rev-parse HEAD 2>/dev/null)"
+        )
```
junhaoliao: think this is longer than 100 char. maybe extract `revision="$(git -C "$script_dir" rev-parse HEAD 2>/dev/null)`

components/core/tools/docker-images/proxy-lib.sh
```diff
@@ -0,0 +1,153 @@
...
+# (curl, dnf, pip, etc.) reject these certificates because the corporate CA
+# isn't in the container's trust store.
+#
+# Solution: This library detects the host's CA bundle, copies it into the Docker
+# build context, and forwards proxy environment variables as build args. A
+# companion script (setup-corporate-proxy.sh) runs inside the container to
+# install the CA bundle into the system trust store.
+#
+# When no CA bundle is found on the host, everything is a no-op — builds work
+# identically to before.
```
junhaoliao:
```suggestion
# When no CA bundle is found on the host, an empty file is staged so that the
# Dockerfile's COPY succeeds; the in-container setup script detects the empty
# file and skips trust-store installation.
```

components/core/tools/scripts/lib_install/setup-corporate-proxy.sh
junhaoliao, line 1: the max-length of some scripts seem to be 80, which is inconsistent with the 100 limit in the guideline

components/core/tools/docker-images/proxy-lib.sh
junhaoliao, line 1: do you know if this works on mac? if not, we should better document it

junhaoliao: in fact, if this doesn't work on mac, we might better exclude this from being sourced, or we might break the dependent scripts on mac

jackluo923: It works on mac

components/core/tools/docker-images/proxy-lib.sh
jackluo923: Good catch — flipped the default to `true` so `--pull` is always passed (matching prior behavior). Can be opted out with `DOCKER_PULL=false` for offline/air-gapped builds. (d1bf4db)

components/core/tools/docker-images/proxy-lib.sh
jackluo923: Done — renamed to `_build_cmd` in `finalize_build`. (d1bf4db)

components/core/tools/docker-images/proxy-lib.sh
jackluo923: Done — extracted into a local `revision` variable. (d1bf4db)

components/core/tools/scripts/lib_install/setup-corporate-proxy.sh
jackluo923: You're right — rewrapped all comments in both `proxy-lib.sh` and `setup-corporate-proxy.sh` to consistently fill to ~100 chars instead of wrapping at 80.

docs/src/dev-docs/tooling-containers.md
```diff
@@ -189,6 +189,49 @@ The container image sets the following environment variables and configurations,
...
+certificates without disabling SSL.
+
+The following proxy environment variables are forwarded as Docker build args when set:
+
+| Variable | Description |
+|---|---|
+| `HTTP_PROXY` / `http_proxy` | HTTP proxy URL |
+| `HTTPS_PROXY` / `https_proxy` | HTTPS proxy URL |
+| `ALL_PROXY` / `all_proxy` | SOCKS/catch-all proxy URL |
+| `NO_PROXY` / `no_proxy` | Comma-separated list of hosts to bypass |
```
junhaoliao:
```suggestion
| Variable                      | Description                             |
|-------------------------------|-----------------------------------------|
| `HTTP_PROXY` / `http_proxy`   | HTTP proxy URL                          |
| `HTTPS_PROXY` / `https_proxy` | HTTPS proxy URL                         |
| `ALL_PROXY` / `all_proxy`     | SOCKS/catch-all proxy URL               |
| `NO_PROXY` / `no_proxy`       | Comma-separated list of hosts to bypass |
```

docs/src/dev-docs/tooling-containers.md
```diff
@@ -189,6 +189,49 @@ The container image sets the following environment variables and configurations,
...
+
+### Package mirror overrides
+
+Each distro supports an environment variable to override the default package mirror:
+
+| Variable | Distro | Example |
+|---|---|---|
+| `DNF_MIRROR_BASE_URL` | manylinux_2_28 (AlmaLinux), centos-stream-9 | `https://internal.example.com/almalinux` |
+| `APK_MIRROR_URL` | musllinux_1_2 (Alpine) | `https://internal.example.com/alpine` |
+| `APT_MIRROR_URL` | ubuntu-jammy | `https://internal.example.com/ubuntu` |
```
junhaoliao:
```suggestion
| Variable              | Distro                                      | Example                                  |
|-----------------------|---------------------------------------------|------------------------------------------|
| `DNF_MIRROR_BASE_URL` | manylinux_2_28 (AlmaLinux), centos-stream-9 | `https://internal.example.com/almalinux` |
| `APK_MIRROR_URL`      | musllinux_1_2 (Alpine)                      | `https://internal.example.com/alpine`    |
| `APT_MIRROR_URL`      | ubuntu-jammy                                | `https://internal.example.com/ubuntu`    |
```

docs/src/dev-docs/tooling-containers.md
```diff
@@ -189,6 +189,49 @@ The container image sets the following environment variables and configurations,
...
+
+### Other build flags
+
+| Variable | Default | Description |
+|---|---|---|
+| `DOCKER_PULL` | `true` | Pull the latest base image before building. Set to `false` for offline builds. |
+| `DOCKER_NETWORK` | (auto) | Override Docker's network mode (e.g., `host`, `bridge`). |
```
junhaoliao:
```suggestion
| Variable         | Default | Description                                                                    |
|------------------|---------|--------------------------------------------------------------------------------|
| `DOCKER_PULL`    | `true`  | Pull the latest base image before building. Set to `false` for offline builds. |
| `DOCKER_NETWORK` | (auto)  | Override Docker's network mode (e.g., `host`, `bridge`).                       |
```

junhaoliao:
for the title, how about:
```
feat(docker): Support corporate proxies and package mirror overrides for base image builds.
```

junhaoliao:
@coderabbitai create a bug report

`components/core/tools/docker-images/clp-env-base-ubuntu-jammy/build.sh` does not run successfully and the error seems related to the changes in this PR:
```bash
junhao@ASUS-X870E:~/workspace/8-clp/components/core/tools/docker-images/clp-env-base-ubuntu-jammy$ ./build.sh 
Corporate proxy support: copying CA bundle from /etc/ssl/certs/ca-certificates.crt
Running: docker build --tag clp-core-dependencies-x86-ubuntu-jammy:dev /home/junhao/workspace/8-clp/components/core/tools/docker-images/clp-env-base-ubuntu-jammy/../../../ --file /home/junhao/workspace/8-clp/components/core/tools/docker-images/clp-env-base-ubuntu-jammy/Dockerfile --pull --label org.opencontainers.image.revision=e3d23ee49a84b6eb3862d173c77d50611ff45bad --label org.opencontainers.image.source=https://github.com/y-scope/clp.git
[+] Building 1.4s (15/18)                                                            docker:default
 => [internal] load build definition from Dockerfile                                           0.0s
 => => transferring dockerfile: 2.82kB                                                         0.0s
 => resolve image config for docker-image://docker.io/docker/dockerfile:1                      0.4s
 => [auth] docker/dockerfile:pull token for registry-1.docker.io                               0.0s
 => CACHED docker-image://docker.io/docker/dockerfile:1@sha256:4a43a54dd1fedceb30ba47e76cfcf2  0.0s
 => => resolve docker.io/docker/dockerfile:1@sha256:4a43a54dd1fedceb30ba47e76cfcf2b47304f4161  0.0s
 => [internal] load metadata for docker.io/library/ubuntu:jammy                                0.3s
 => [auth] library/ubuntu:pull token for registry-1.docker.io                                  0.0s
 => [internal] load .dockerignore                                                              0.0s
 => => transferring context: 2B                                                                0.0s
 => [base 1/8] FROM docker.io/library/ubuntu:jammy@sha256:ce4a593b4e323dcc3dd728e397e0a866a1b  0.0s
 => => resolve docker.io/library/ubuntu:jammy@sha256:ce4a593b4e323dcc3dd728e397e0a866a1bf516a  0.0s
 => [internal] load build context                                                              0.1s
 => => transferring context: 246.91kB                                                          0.0s
 => CACHED [base 2/8] WORKDIR /root                                                            0.0s
 => [base 3/8] RUN mkdir -p ./tools/scripts/lib_install                                        0.2s
 => [base 4/8] COPY --link ./tools/scripts/lib_install ./tools/scripts/lib_install             0.1s
 => => merging                                                                                 0.0s
 => ERROR [base 5/8] RUN ./tools/scripts/lib_install/setup-corporate-proxy.sh                  0.1s
------                                                                                              
 > [base 5/8] RUN ./tools/scripts/lib_install/setup-corporate-proxy.sh:
0.126 setup-corporate-proxy: installing corporate CA certificates...
0.126 setup-corporate-proxy: WARNING: no recognized CA trust update tool found.
------
Dockerfile:13
--------------------
  11 |     # proxies (e.g., Zscaler, Fortinet, Palo Alto). No-op when not behind a
  12 |     # corporate proxy. See proxy-lib.sh for details.
  13 | >>> RUN ./tools/scripts/lib_install/setup-corporate-proxy.sh
  14 |     
  15 |     # Point pip/pipx and curl at the system trust store. By default, Python's
--------------------
ERROR: failed to build: failed to solve: process "/bin/sh -c ./tools/scripts/lib_install/setup-corporate-proxy.sh" did not complete successfully: exit code: 1
```

PR #1992, opened 2026-02-14
feat!: Add multi-dataset query support (resolves #1933).

### LINKED ISSUE #1933: Support for Selecting multiple datasets while searching in CLP UI
### Request

Hi Team

We require a feature where SDEs can select  multiple datasets while searching in CLP UI. 

### Possible implementation

submit parallel jobs to different datasets and then collate search results based on timestamp.

junhaoliao:
### Impact Assessment

- **What is affected:** All query entry points (Web UI, Rust API server, CLI), the query scheduler,
  query workers, C++ `clp-s` binary output handler, and result rendering in the Web UI.
- **Why the change is needed:** Users need to search across multiple datasets simultaneously without
  submitting separate queries for each dataset. This enables cross-dataset log correlation and
  analysis workflows.
- **Breaking changes:**
  - The CLI `--dataset` flag keeps its name but now uses `action="append"` (can be specified
    multiple times, e.g., `--dataset ds1 --dataset ds2`). Old single-value usage
    (`--dataset <name>`) still works for the single-dataset case.
  - The Rust API server's `QueryConfig.dataset: Option<String>` becomes
    `QueryConfig.datasets: Option<Vec<String>>`.
  - The WebUI server's `QueryJobCreationSchema` changes from `dataset: Union(Null, String)` to
    `datasets: Array(String)`.
  - Existing API clients and stored queries must update their JSON payloads accordingly.
  - The `SearchJobConfig.dataset` field is replaced with `SearchJobConfig.datasets` (a list) in the
    Python scheduler, Rust API server, and TypeScript schema.
  - `QueryJobConfig` no longer owns a `dataset` field. Instead, `SearchJobConfig` owns
    `datasets: list[str] | None` and `ExtractJsonJobConfig` owns `dataset: str | None`.
- **New configuration:** `max_datasets_per_query` (default 10, `null` for unlimited) is added to the
  `QueryScheduler` config.

### Key changes by layer

1. **Job config schema** -- Split `QueryJobConfig.dataset` into type-specific fields.
   - `QueryJobConfig` base class: `dataset` removed (empty `pass` body).
   - `SearchJobConfig`: new `datasets: list[str] | None` field.
   - `ExtractJsonJobConfig`: new `dataset: str | None` field (extraction targets a single dataset).
   - Rust `SearchJobConfig`: `dataset: Option<String>` -> `datasets: Option<Vec<String>>`.
   - TypeScript `QueryJobCreationSchema`: `dataset: Union(Null, String)` ->
     `datasets: Array(String)`.
   - Files:
     - `components/job-orchestration/job_orchestration/scheduler/job_config.py`
     - `components/clp-rust-utils/src/job_config/search.rs`
     - `components/webui/common/src/schemas/search.ts`

2. **Configuration** -- Add `max_datasets_per_query` (default 10, `null` for unlimited) to
   `QueryScheduler` config. Also exposed to the Web UI via `settings.json` as
   `MaxDatasetsPerQuery`.
   - Files:
     - `components/clp-py-utils/clp_py_utils/clp_config.py`
     - `components/clp-package-utils/clp_package_utils/controller.py`
     - `components/package-template/src/etc/clp-config.template.json.yaml`
     - `components/package-template/src/etc/clp-config.template.text.yaml`
     - `components/webui/client/public/settings.json`
     - `components/webui/client/src/config/index.ts`
     - `components/webui/client/src/settings.ts`

3. **Scheduler** -- Refactored `handle_pending_query_jobs` by extracting two helpers:
   `_handle_new_search_job` and `_handle_new_extraction_job`.
   - `_handle_new_search_job`: validates datasets (deduplication, `max_datasets_per_query` limit,
     existence check), then fetches archives and creates the `SearchJob`.
   - `_handle_new_extraction_job`: handles IR/JSON extraction jobs (logic moved from the main loop
     without functional change, except `dataset` is now passed as an explicit task argument).
   - `get_archives_for_search` split into two functions:
     - `_get_archives_for_search_without_datasets` -- for CLP-Text (no datasets).
     - `get_archives_for_search` -- for CLP-S, uses `UNION ALL` across per-dataset archive tables
       and returns `{archive_id, end_timestamp, dataset}` dicts.
   - `dispatch_query_job` and `get_task_group_for_job` now accept `list[dict]` (with `archive_id`
     and `dataset` keys) instead of `list[str]` of archive IDs. `dataset` is passed as an explicit
     Celery task kwarg to `search` and `extract_stream`.
   - Dataset existence check moved from the main loop (which checked `QueryJobConfig.dataset`) into
     `_handle_new_search_job` (which checks `SearchJobConfig.datasets`).
   - Logging changed from f-strings to `%s`-style structured logging.
   - File: `components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py`

4. **C++ core -- results attribution** -- Added `--dataset` CLI flag to `clp-s` and a `dataset`
   field in each BSON result document written to the results cache, enabling per-result dataset
   attribution.
   - `ResultsCacheOutputHandler` constructor takes a new `std::string dataset` parameter, stored as
     `m_dataset`. Each `QueryResult` struct gains a `dataset` member that is written into the BSON
     output.
   - Files:
     - `components/core/src/clp_s/CommandLineArguments.{cpp,hpp}`
     - `components/core/src/clp_s/OutputHandlerImpl.{cpp,hpp}`
     - `components/core/src/clp_s/archive_constants.hpp`
     - `components/core/src/clp_s/clp-s.cpp`

5. **Worker execution** -- `fs_search_task.py` and `extract_stream_task.py` accept `dataset` as an
   explicit Celery task parameter and pass it to the `clp-s` binary via `--dataset`. The dataset is
   no longer read from `search_config.dataset` / `extract_json_config.dataset` inside the worker;
   instead, it comes from the scheduler via the task arguments.
   - Files:
     - `components/job-orchestration/job_orchestration/executor/query/fs_search_task.py`
     - `components/job-orchestration/job_orchestration/executor/query/extract_stream_task.py`

6. **Web UI** -- The dataset selector supports both single-select (Presto) and multi-select
   (Native/clp-s) modes via the `isMultiSelect` prop on `DatasetSelect`.
   - State changes:
     - `selectDataset: string | null` -> `selectedDatasets: string[]`
     - `cachedDataset: Nullable<string>` -> `queriedDatasets: string[]` (datasets submitted with the
       most recent query, separate from UI selection state).
   - `LogViewerLink` reads `dataset` from each individual search result document (returned by the
     C++ core) instead of the global cached state. This is necessary because multi-dataset results
     can come from different datasets.
   - The multi-select dropdown enforces `MaxDatasetsPerQuery` on the client side and uses responsive
     tag collapsing. Clearing all selections falls back to the `default` dataset.
   - SQL queries for time range and query speed updated to use `UNION ALL` across multiple dataset
     tables.
   - `SearchResult` type gains a `dataset: string` field.
   - Fastify routes updated to accept a `datasets` array.
   - Files:
     - `components/webui/client/src/pages/SearchPage/SearchControls/Dataset/DatasetSelect.tsx`
     - `components/webui/client/src/pages/SearchPage/SearchControls/Dataset/index.{tsx,module.css}`
     - `components/webui/client/src/pages/SearchPage/SearchControls/Native/SearchButton/SubmitButton/index.tsx`
     - `components/webui/client/src/pages/SearchPage/SearchControls/Presto/Guided/GuidedControls/From.tsx`
     - `components/webui/client/src/pages/SearchPage/SearchControls/Presto/Guided/presto-guided-search-requests.ts`
     - `components/webui/client/src/pages/SearchPage/SearchControls/Presto/SqlSearchButton/RunButton/GuidedRunButton.tsx`
     - `components/webui/client/src/pages/SearchPage/SearchControls/QueryStatus/QuerySpeed/{index.tsx,utils.ts}`
     - `components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/{sql.ts,utils.tsx}`
     - `components/webui/client/src/pages/SearchPage/SearchControls/TimeRangeInput/Presto/TimeRangeFooter/TimestampKeySelect/index.tsx`
     - `components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Native/Message/{LogViewerLink.tsx,index.tsx}`
     - `components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Native/SearchResultsVirtualTable/typings.tsx`
     - `components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTimeline/Native/NativeResultsTimeline.tsx`
     - `components/webui/client/src/pages/SearchPage/SearchState/{index.tsx,Presto/useTimestampKeyInit/index.tsx}`
     - `components/webui/server/src/routes/api/search/index.ts`
     - `components/webui/common/src/schemas/search.ts`

7. **API server (Rust)** -- `QueryConfig.dataset: Option<String>` ->
   `QueryConfig.datasets: Option<Vec<String>>`. Default fallback for CLP-S changed from
   `Some("default")` to `Some(vec!["default"])`.
   - Files:
     - `components/api-server/src/client.rs`
     - `components/api-server/src/routes.rs`

8. **CLI tools** -- The `--dataset` flag keeps its name but now uses `action="append"` to accept
   multiple values (e.g., `--dataset ds1 --dataset ds2`).
   `validate_dataset_exists` -> `validate_datasets_exist` to validate all specified datasets before
   submission. The CLI also enforces `max_datasets_per_query` before submitting.
   - Files:
     - `components/clp-package-utils/clp_package_utils/scripts/search.py`
     - `components/clp-package-utils/clp_package_utils/scripts/native/search.py`
     - `components/clp-package-utils/clp_package_utils/scripts/native/utils.py`

9. **Docs** -- Updated API server guide, quick-start guide, and regenerated OpenAPI spec to reflect
   `datasets` (plural) in the API payload. Added CLI documentation for multi-dataset usage.
   - Files:
     - `docs/src/_static/generated/api-server-openapi.json`
     - `docs/src/user-docs/guides-using-the-api-server.md`
     - `docs/src/user-docs/quick-start/clp-json.md`

## Checklist

* [x] The PR satisfies the contribution guidelines.
* [x] This is a breaking change and that has been indicated in the PR title.
* [x] Necessary docs have been updated.

### 0. Build CLP package

**Task:** Verify the project builds successfully with all changes (C++, Rust, Python, TypeScript).

**Command:**
```bash
task
```

Build completed successfully -- all C++ targets, Rust binaries, Python packages, and TypeScript
client/server compiled without errors.

---

### T1. Start CLP (text engine) and compress sample data

**Commands:**
```bash
cd build/clp-package
mv etc/clp-config.template.text.yaml etc/clp-config.yaml
./sbin/start-clp.sh
./sbin/compress.sh ~/samples/hive-24hr/
```

**Output:**
```
Compressed 1.99GB into 44.02MB (46.25x). Speed: 697.29MB/s.
```

### T2. Search (no `--dataset` flag)

**Command:**
```bash
./sbin/search.sh "*" --count
```

**Output:**
```
tags: [] count: 6864529
```

### T3. Search for specific term

**Command:**
```bash
./sbin/search.sh "ERROR" --count
```

**Output:**
```
tags: [] count: 132582
```

### T4. Search page loads (Web UI)

Navigated to `http://localhost:4000/search`. The page loads correctly. No dataset selector is shown
(CLP-Text does not support datasets).

### T5. Wildcard search (Web UI)

Searched with `*` query. Returns **6,864,529 results** (matches the CLI count from T2). Search
results display correctly with timestamps, log messages, and original file links. No console errors.

### T6. Stop CLP (text engine)

```bash
./sbin/stop-clp.sh
```

---

### 1. Start CLP and compress sample data

**Task:** Start CLP (with API server enabled) and compress sample data into two datasets.

**Commands:**
```bash
task  # rebuild to restore default JSON config
cd build/clp-package
./sbin/start-clp.sh
./sbin/compress.sh ~/samples/postgresql.log
./sbin/compress.sh --dataset test_ds ~/samples/postgresql.log
```

Each dataset contains 1,000,000 log events from the same PostgreSQL log file (`385.21 MB`
uncompressed -> `10.22 MB` compressed).

---

### 2. Single dataset search

**Command:**
```bash
./sbin/search.sh "*" --dataset default --count
```

**Output:**
```
tags: [] count: 1000000
```

### 3. No `--dataset` flag (defaults to `default`)

**Command:**
```bash
./sbin/search.sh "*" --count
```

**Output:**
```
tags: [] count: 1000000
```

Without `--dataset`, the CLI defaults to `["default"]`.

### 4. Multi-dataset search

**Command:**
```bash
./sbin/search.sh "*" --dataset default --dataset test_ds --count
```

**Output:**
```
tags: [] count: 2000000
```

The search fans out across both datasets, correctly returning 2,000,000 total events.

### 5. Single non-default dataset

**Command:**
```bash
./sbin/search.sh "*" --dataset test_ds --count
```

**Output:**
```
tags: [] count: 1000000
```

### 6. Non-existent dataset

**Command:**
```bash
./sbin/search.sh "*" --dataset nonexistent --count
```

**Output:**
```
ERROR [search] Dataset(s) ['nonexistent'] don't exist.
ERROR [search] Search failed.
```

### 7. Mixed valid and invalid datasets

**Command:**
```bash
./sbin/search.sh "*" --dataset default --dataset nonexistent --count
```

**Output:**
```
ERROR [search] Dataset(s) ['nonexistent'] don't exist.
ERROR [search] Search failed.
```

The validator identifies only the non-existent dataset and prevents the query from being submitted.

---

### 8. Single dataset search (buffered)

**Command:**
```bash
curl -s -X POST http://localhost:3001/query \
  -H 'Content-Type: application/json' \
  -d '{"query_string":"*","datasets":["default"],"max_num_results":10,
       "buffer_results_in_mongodb":true}'
```

Returns query ID. SSE stream returns results from the default dataset.

### 9. Single dataset search (unbuffered / SSE)

**Command:**
```bash
curl -s -X POST http://localhost:3001/query \
  -H 'Content-Type: application/json' \
  -d '{"query_string":"*","datasets":["default"],"max_num_results":10,
       "buffer_results_in_mongodb":false}'
```

Returns SSE stream with results. Verified correct operation.

### 10. Multi-dataset search (buffered)

**Command:**
```bash
curl -s -X POST http://localhost:3001/query \
  -H 'Content-Type: application/json' \
  -d '{"query_string":"*","datasets":["default","test_ds"],"max_num_results":10,
       "buffer_results_in_mongodb":true}'
```

Returns results across both datasets.

### 11. Multi-dataset search (unbuffered / SSE)

Same as above with `buffer_results_in_mongodb: false`. Verified SSE stream returns results from both
datasets.

### 12. Non-existent dataset via API

```bash
curl -s -X POST http://localhost:3001/query \
  -H 'Content-Type: application/json' \
  -d '{"query_string":"*","datasets":["nonexistent"]}'
```

Scheduler rejects the non-existent dataset (logged: `Datasets {'nonexistent'} don't exist.`).

### 13. Cancel query via API

```bash
curl -s -X DELETE http://localhost:3001/query/{id}
```

Successfully cancels a running query.

---

#### Web UI validation (Playwright)

Automated browser tests using Playwright (headless Chromium). Screenshots saved to
`validation-screenshots/`.

### 14. Search page loads

Navigated to `http://localhost:4000/search`. The page loads correctly with the dataset selector
showing `default` pre-selected.

### 15. Dataset selector shows both datasets

The multi-select dropdown displays both `default` and `test_ds`.

### 16. Single dataset search (Web UI)

Searched with `*` query and `default` dataset selected. Returns **1,000,000 results**.

### 17. Extraction job -- `default` dataset

Clicked an "Original file" link in search results. The EXTRACT_JSON job was submitted, completed,
and the yscope-log-viewer loaded with the extracted JSONL stream content.

### 18. Extraction job -- `test_ds` dataset

Switched to `test_ds` only, searched, and clicked an "Original file" link. The extraction job
correctly used `ExtractJsonJobConfig.dataset = "test_ds"` to locate the archive and the log viewer
loaded successfully.

### 19. Multi-dataset search (Web UI)

Selected both `default` and `test_ds`, searched with `*`. Returns **2,000,000 results**.

### 20. No console errors (Web UI)

Loaded a fresh page and monitored the browser console. No infinite re-render errors or other
critical console errors detected.

---

### 21. Stop CLP

**Command:**
```bash
./sbin/stop-clp.sh
```


components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -605,6 +605,10 @@ CommandLineArguments::parse_arguments(int argc, char const** argv) {
                     po::value<uint64_t>(&m_max_num_results)->value_name("MAX")->
                             default_value(m_max_num_results),
                     "The maximum number of results to output"
+            )(
+                    "dataset",
+                    po::value<std::string>(&m_dataset)->value_name("DATASET"),
+                    "The dataset name to include in each result document"
             );
```
junhaoliao, line 612: The dataset name needs to be include in the query results before dumping them into the result cache. i know this feels like a hack. a more future proof interface might to to accept a JSON object string `--extra-result-metadata` and merge such objects with the results before dumping into the results cache. what do you think?

junhaoliao: alternatively, we can add multi-dataset support to the stream extraction flow as well. then we just pass an array of `datasets` in the stream extraction config, avoiding touching the clp-s binary

i think this may be more aligned with the future plan of using a single table for all datasets (maybe i haven't had a correct understanding of how we are going to refactor the datasets feature though

components/clp-package-utils/clp_package_utils/scripts/native/utils.py
```diff
@@ -87,8 +87,10 @@ def validate_dataset_exists(db_config: Database, dataset: str) -> None:
         closing(sql_adapter.create_connection(True)) as db_conn,
         closing(db_conn.cursor(dictionary=True)) as db_cursor,
     ):
-        if dataset not in fetch_existing_datasets(db_cursor, table_prefix):
-            raise ValueError(f"Dataset `{dataset}` doesn't exist.")
+        existing_datasets = fetch_existing_datasets(db_cursor, table_prefix)
+        for dataset in datasets:
+            if dataset not in existing_datasets:
+                raise ValueError(f"Dataset `{dataset}` doesn't exist.")
```
sitaowang1998: Shall we report all datasets that do not exist in the metadata database?

components/job-orchestration/job_orchestration/executor/query/extract_stream_task.py
```diff
@@ -95,7 +96,8 @@ def _make_clp_s_command_and_env_vars(
         "x",
     ]
 
-    dataset = extract_json_config.dataset
+    if dataset is None:
+        dataset = extract_json_config.datasets[0] if extract_json_config.datasets else None
```
sitaowang1998: If using the first dataset is the intended behavior for backward compatibility, add comment to avoid confusion.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -675,6 +702,25 @@ def handle_pending_query_jobs(
                         logger.error(f"Failed to set job {job_id} as failed.")
                     continue
 
+                # NOTE: This assumes we never delete a dataset.
+                missing = set(datasets) - existing_datasets
+                if missing:
```
sitaowang1998: Prefer to use explicit check, even in Python.
```suggestion
                if len(missing) == 0:
```

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -675,6 +702,25 @@ def handle_pending_query_jobs(
...
+                    existing_datasets.update(fetch_existing_datasets(db_cursor, table_prefix))
+                    missing = set(datasets) - existing_datasets
+                    if missing:
```
sitaowang1998: Same as above.
```suggestion
                    if len(missing) == 0:
```

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -657,12 +666,30 @@ def handle_pending_query_jobs(
...
 
             table_prefix = clp_metadata_db_conn_params["table_prefix"]
-            dataset = QueryJobConfig.model_validate(job_config).dataset
-            if dataset is not None and dataset not in existing_datasets:
-                # NOTE: This assumes we never delete a dataset.
-                existing_datasets.update(fetch_existing_datasets(db_cursor, table_prefix))
-                if dataset not in existing_datasets:
-                    logger.error(f"Dataset `{dataset}` doesn't exist.")
+            datasets = QueryJobConfig.model_validate(job_config).datasets
+            if datasets is not None:
```
sitaowang1998: This if branch is pretty large, should we extract it into a separate function?

sitaowang1998 (changes requested):
I have reviewed the package side. Should we also check if the `datasets` exceeds `max_datasets_per_query` in search script?

components/clp-package-utils/clp_package_utils/scripts/native/search.py
```diff
@@ -297,10 +297,17 @@ def main(argv):
...
-    if dataset is not None:
+    datasets = parsed_args.dataset
+    if datasets is not None:
+        max_datasets_per_query = clp_config.query_scheduler.max_datasets_per_query
+        if max_datasets_per_query is not None and len(datasets) > max_datasets_per_query:
+            logger.error(
+                f"Number of datasets ({len(datasets)}) exceeds"
+                f" max_datasets_per_query={max_datasets_per_query}."
+            )
+            return -1
```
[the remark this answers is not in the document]
sitaowang1998: Sorry. Should have caught this in the first review. CodeRabbit is correct about [Ruff G004](https://docs.astral.sh/ruff/rules/logging-f-string/). We should use `"%s"` to format in logging.

components/clp-package-utils/clp_package_utils/scripts/native/utils.py
```diff
@@ -87,8 +87,10 @@ def validate_dataset_exists(db_config: Database, dataset: str) -> None:
         closing(sql_adapter.create_connection(True)) as db_conn,
         closing(db_conn.cursor(dictionary=True)) as db_cursor,
     ):
-        if dataset not in fetch_existing_datasets(db_cursor, table_prefix):
-            raise ValueError(f"Dataset `{dataset}` doesn't exist.")
+        existing_datasets = fetch_existing_datasets(db_cursor, table_prefix)
+        missing = [ds for ds in datasets if ds not in existing_datasets]
+        if len(missing) > 0:
+            raise ValueError(f"Dataset(s) {missing} don't exist.")
```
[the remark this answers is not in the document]
sitaowang1998: 1. We prefer to use explicit length check.
2. Apart from the long message, we should not also not directly use format string in `raise`. Check https://docs.astral.sh/ruff/rules/f-string-in-exception/.
3. This one makes sense.

components/clp-package-utils/clp_package_utils/scripts/search.py
```diff
@@ -113,16 +113,17 @@ def main(argv):
...
-            validate_dataset_name(clp_db_connection_params["table_prefix"], dataset)
+            for ds in datasets:
+                validate_dataset_name(clp_db_connection_params["table_prefix"], ds)
         except Exception as e:
             logger.error(e)
             return -1
-    elif dataset is not None:
+    elif datasets is not None:
         logger.error(f"Dataset selection is not supported for storage engine: {storage_engine}.")
         return -1
```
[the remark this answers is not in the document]
sitaowang1998: 1. Make sense.
2. Same as above. Avoid using f-string in logs.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -839,18 +720,16 @@ def handle_pending_query_jobs(
...
                 job,
-                archive_ids_for_search,
+                archives_for_search,
                 clp_metadata_db_conn_params,
                 results_cache_uri,
                 job.num_archives_to_search,
             )
             logger.info(
-                f"Dispatched job {job_id} with {len(archive_ids_for_search)} archives to search."
+                f"Dispatched job {job_id} with {len(archives_for_search)} archives to search."
```
sitaowang1998: Same. Avoid using f-string in logs.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+    if job_id in active_jobs:
+        return
+
+    search_config = SearchJobConfig.model_validate(job_config)
+    datasets = search_config.datasets
+    if datasets is not None:
+        # Deduplicate
+        datasets = list(dict.fromkeys(datasets))
+        if len(datasets) == 0:
+            logger.error(f"Job {job_id} has an empty datasets list.")
```
sitaowang1998: Same. Avoid using f-string in log.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+            if not set_job_or_task_status(
+                db_conn,
+                QUERY_JOBS_TABLE_NAME,
+                job_id,
+                QueryJobStatus.FAILED,
+                QueryJobStatus.PENDING,
+                start_time=datetime.datetime.now(),
+                duration=0,
+            ):
+                logger.error(f"Failed to set job {job_id} as failed.")
```
sitaowang1998: Same. Avoid using f-string in log.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+            return
+
+        # Enforce max_datasets_per_query limit
+        if max_datasets_per_query is not None and len(datasets) > max_datasets_per_query:
+            logger.error(
+                f"Job {job_id} requests {len(datasets)} datasets,"
+                f" exceeding max_datasets_per_query={max_datasets_per_query}."
+            )
```
sitaowang1998, line 1212: Same. Avoid using f-string in log.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+            if not set_job_or_task_status(
+                db_conn,
+                QUERY_JOBS_TABLE_NAME,
+                job_id,
+                QueryJobStatus.FAILED,
+                QueryJobStatus.PENDING,
+                start_time=datetime.datetime.now(),
+                duration=0,
+            ):
+                logger.error(f"Failed to set job {job_id} as failed.")
```
sitaowang1998: Same. Avoid using f-string in log.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+            return
+
+        # NOTE: This assumes we never delete a dataset.
+        missing = set(datasets) - existing_datasets
+        if missing:
```
sitaowang1998:
```suggestion
        if length(missing) > 0:
```

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+            existing_datasets.update(fetch_existing_datasets(db_cursor, table_prefix))
+            missing = set(datasets) - existing_datasets
+            if missing:
```
sitaowang1998:
```suggestion
            if length(missing) > 0:
```

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+                logger.error(f"Datasets {missing} don't exist.")
```
sitaowang1998: Avoid using f-string in log.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+                if not set_job_or_task_status(
+                    db_conn,
+                    QUERY_JOBS_TABLE_NAME,
+                    job_id,
+                    QueryJobStatus.FAILED,
+                    QueryJobStatus.PENDING,
+                    start_time=datetime.datetime.now(),
+                    duration=0,
+                ):
+                    logger.error(f"Failed to set job {job_id} as failed.")
```
sitaowang1998: Avoid using f-string in log.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+            db_conn,
+            QUERY_JOBS_TABLE_NAME,
+            job_id,
+            QueryJobStatus.SUCCEEDED,
+            QueryJobStatus.PENDING,
+            start_time=datetime.datetime.now(),
+            num_tasks=0,
+            duration=0,
+        ):
+            logger.info(f"No matching archives, skipping job {job_id}.")
```
sitaowang1998: Avoid using f-string in log.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+    global active_jobs
+
+    job_handle: StreamExtractionHandle
+    try:
+        if QueryJobType.EXTRACT_IR == job_type:
+            job_handle = IrExtractionHandle(job_id, job_config, db_conn, table_prefix)
+        else:
+            job_handle = JsonExtractionHandle(job_id, job_config, db_conn, table_prefix)
+    except ValueError:
+        logger.exception("Failed to initialize extraction job handle")
```
sitaowang1998, line 1330: Avoid using f-string in log.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+            db_conn,
+            QUERY_JOBS_TABLE_NAME,
+            job_id,
+            QueryJobStatus.FAILED,
+            QueryJobStatus.PENDING,
+            start_time=datetime.datetime.now(),
+            num_tasks=0,
+            duration=0,
+        ):
+            logger.error(f"Failed to set job {job_id} as failed")
```
sitaowang1998: Avoid using f-string in log.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+    # aforementioned logic.
+
+    # Check if the required streams are currently being extracted; if so, add the job ID
+    # to the list of jobs waiting for it.
+    if job_handle.is_stream_extraction_active():
+        job_handle.mark_job_as_waiting()
+        logger.info(
+            f"Stream {job_handle.get_stream_id()} is already being extracted,"
+            f" so mark job {job_id} as running."
+        )
```
sitaowang1998, line 1368: Avoid using f-string in log.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+        if not set_job_or_task_status(
+            db_conn,
+            QUERY_JOBS_TABLE_NAME,
+            job_id,
+            QueryJobStatus.RUNNING,
+            QueryJobStatus.PENDING,
+            start_time=datetime.datetime.now(),
+            num_tasks=0,
+        ):
+            logger.error(f"Failed to set job {job_id} as running")
```
sitaowang1998: Same. Avoid using f-string in log.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+        return
+
+    # Check if a required stream file has already been extracted
+    if job_handle.is_stream_extracted(results_cache_uri, stream_collection_name):
+        logger.info(
+            f"Stream {job_handle.get_stream_id()} already extracted,"
+            f" so mark job {job_id} as succeeded."
+        )
```
sitaowang1998, line 1387: Avoid using f-string in log.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+            db_conn,
+            QUERY_JOBS_TABLE_NAME,
+            job_id,
+            QueryJobStatus.SUCCEEDED,
+            QueryJobStatus.PENDING,
+            start_time=datetime.datetime.now(),
+            num_tasks=0,
+            duration=0,
+        ):
+            logger.error(f"Failed to set job {job_id} as succeeded")
```
sitaowang1998: Avoid using f-string in log.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+        new_stream_extraction_job,
+        [{"archive_id": archive_id, "dataset": job_handle.get_dataset()}],
+        clp_metadata_db_conn_params,
+        results_cache_uri,
+        1,
+    )
+
+    job_handle.mark_job_as_waiting()
+    active_jobs[job_id] = new_stream_extraction_job
+    logger.info(f"Dispatched stream extraction job {job_id} for archive: {archive_id}")
```
sitaowang1998: Avoid using f-string in log.

sitaowang1998 (changes requested):
The package side have two problems:
1. Avoid using f-string for logging messages. https://docs.astral.sh/ruff/rules/logging-f-string/
2. Avoid using f-string directly in exception. Store it inside a variable and then use it for exception. https://docs.astral.sh/ruff/rules/f-string-in-exception/

These two should have been caught by `ruff`. I think we are ignoring `ruff` in the CI workflow because of old violations. Maybe we should double check if there are any new violations caught by `ruff`.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+
+    archive_end_ts_lower_bound: int | None = None
+    if archive_retention_period is not None:
+        archive_end_ts_lower_bound = SECOND_TO_MILLISECOND * (
+            job_creation_time - archive_retention_period * MIN_TO_SECONDS
+        )
+
+    archives_for_search = get_archives_for_search(
+        db_conn, table_prefix, search_config, archive_end_ts_lower_bound, datasets
+    )
```
gibber9809: I think we need to handle the `datasets=None` case either here or in `get_archives_for_search`.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1238,5 +1120,228 @@ async def main(argv: list[str]) -> int:
...
+    db_cursor,
+    job_id: str,
+    job_config: dict,
+    job_creation_time: float,
+    table_prefix: str,
+    max_datasets_per_query: int | None,
+    existing_datasets: set[str],
+    archive_retention_period: int | None,
+    pending_search_jobs: list,
+    reducer_acquisition_tasks: list[asyncio.Task],
```
gibber9809, line 1156: Do you think it makes sense to just return `tuple[list, list[asyncio.Task] | None` from this function instead of using output parameters? Maybe the newly added search job as well.

I think it might make the code easier to read since the effects of this function would more obvious. Alternatively, a brief docstring would help a lot.

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -70,11 +70,13 @@ ResultsCacheOutputHandler::ResultsCacheOutputHandler(
...
         uint64_t batch_size,
         uint64_t max_num_results,
+        string dataset,
         bool should_output_timestamp
 )
         : ::clp_s::search::OutputHandler(should_output_timestamp, true),
           m_batch_size(batch_size),
-          m_max_num_results(max_num_results) {
+          m_max_num_results(max_num_results),
+          m_dataset(std::move(dataset)) {
```
gibber9809:
```suggestion
          m_max_num_results{max_num_results},
          m_dataset{std::move(dataset)} {
```
Nit: coding guidelines.

components/core/src/clp_s/archive_constants.hpp
```diff
@@ -58,6 +58,7 @@ constexpr char cLogEventIx[]{"log_event_ix"};
 constexpr char cTimestamp[]{"timestamp"};
 constexpr char cMessage[]{"message"};
 constexpr char cArchiveId[]{"archive_id"};
+constexpr char cDataset[]{"dataset"};
```
gibber9809:
```suggestion
constexpr std::string_view cDataset{"dataset"};
```
To comply with our current coding guidelines.

gibber9809: We should probably also move all of these results cache related constants out of `archive_constants.hpp` at some point, but not as part of this PR. @coderabbitai can you create an issue to track?

components/core/src/clp_s/OutputHandlerImpl.hpp
```diff
@@ -128,19 +128,22 @@ class ResultsCacheOutputHandler : public ::clp_s::search::OutputHandler {
...
+                int64_t log_event_idx,
+                std::string_view dataset
         )
                 : original_path(original_path),
                   message(message),
                   timestamp(timestamp),
                   archive_id(archive_id),
-                  log_event_idx(log_event_idx) {}
+                  log_event_idx(log_event_idx),
+                  dataset(dataset) {}
```
gibber9809:
```suggestion
                  log_event_idx{log_event_idx},
                  dataset{dataset} {}
```

gibber9809 (changes requested):
Did an initial review of query scheduler, c++ core, and CLI tools. Still need to take a look at worker execution.

components/core/src/clp_s/CommandLineArguments.cpp
junhaoliao: discussed offline - the `--dataset` CLI flag is fine. we will remove it when we refactor the CLI

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
junhaoliao: the function does mutate (append to) a few lists. making this helper to return the lists may result in the appends happening in the caller anyways. i tried grouping the "output" parameter towards the end of the function signature and adding docstrings to mark them as `[out]` parameters. do you think that will be more readable already, or splitting those parameters into return values would give benefits?

gibber9809: I think with the docstring it is clear enough

gibber9809:
Looked at the search/decompression executors + looked over the revisions and they seem fine.

I think I'm more or less ready to approve after a response on that `datasets=None` comment.

gibber9809 (dismissed):
LGTM after most recent set of changes. PR title fine as well. As long as you tested the package flow again after those changes, then everything looks good to me.

components/webui/client/src/pages/SearchPage/SearchControls/Presto/SqlSearchButton/RunButton/GuidedRunButton.tsx
```diff
@@ -34,7 +34,10 @@ const GuidedRunButton = () => {
     const updateTimelineConfig = useSearchStore((state) => state.updateTimelineConfig);
     const timeRangeOption = useSearchStore((state) => state.timeRangeOption);
     const timeRange = useSearchStore((state) => state.timeRange);
-    const {selectDataset: from} = useSearchStore.getState();
+    const {selectedDatasets} = useSearchStore.getState();
+    const from = 0 < selectedDatasets.length ?
+        selectedDatasets[0] :
```
hoophalab, line 39: It looks like it’s selecting the first dataset without a warning.

Ideally, we’d use the old Presto box, but if that’s too tedious, can we add a message box when multiple datasets are selected? Something like: 'Only the first dataset and timestamp key will be used.'

Alternatively, joining the dataset name with comma probably will also work, but I cannot verify this because Presto is broke.

components/clp-package-utils/clp_package_utils/scripts/native/search.py
hoophalab, line 1: I'm not sure if this is intended or if it's just me seeing this, but search.sh is only showing a comma prefix instead of the actual dataset name.

```
$ sudo sbin/search.sh --dataset spark --dataset sql 'pid > 8000 OR Timestamp > 100'
: {"Event":"SparkListenerBlockManagerAdded","Block Manager ID":{"Executor ID":"driver","Host":"hostb7.dsrg.utoronto.ca","Port":41037},"Maximum Memory":2101975449,"Timestamp":1633875632527,"Maximum Onheap Memory":2101975449,"Maximum Offheap Memory":0}
: {"Event":"SparkListenerBlockManagerAdded","Block Manager ID":{"Executor ID":"10","Host":"10.1.0.16","Port":37693},"Maximum Memory":2101975449,"Timestamp":1633875634570,"Maximum Onheap Memory":2101975449,"Maximum Offheap Memory":0}
: {"Event":"SparkListenerBlockManagerAdded","Block Manager ID":{"Executor ID":"5","Host":"10.1.0.19","Port":39827},"Maximum Memory":2101975449,"Timestamp":1633875634573,"Maximum Onheap Memory":2101975449,"Maximum Offheap Memory":0}
...
```

hoophalab (commented):
Reviewed API server and WebUI. LGTM in general. Two comments.

Validations:
1. WebUI shows joined query results from multiple datasets.
2. API server executes multi-dataset query correctly.

components/webui/client/src/pages/SearchPage/SearchControls/Presto/SqlSearchButton/RunButton/GuidedRunButton.tsx
junhaoliao: good catch. i've updated the code to explicitly avoid handling multi-dataset with the presto flow

components/clp-package-utils/clp_package_utils/scripts/native/search.py
junhaoliao: good catch... supposedly the "original file path" should be before the `:` if i understand correctly. we just never added such to the network writer's outputs? since the issue is a pre-existing one, i created a bug report to track: #2052

hoophalab: Sorry, I didn't realize that this is a pre-existing issue.

hoophalab (approved):

Issue #1978, opened 2026-02-12
[log-ingestor] Fault-tolerance support.


LinZhihao-723:
# Introduction

Currently, the log-ingestor relies entirely on in-memory buffers to track ingestion jobs and the state of ingested files. This approach creates reliability issues during failure and recovery scenarios. Specifically:
- If the log-ingestor crashes and restarts, all in-flight ingestion jobs are lost.
- Any files ingested by a failed job that have not yet been submitted to a compression job may be permanently lost, resulting in data loss.

This feature request aims to improve the reliability of log-ingestor.

# Objective

With fault tolerance implemented at the log-ingestor level, the CLP system should be able to:
- Recover the log-ingestor from failures without losing ingestion data.
  - For S3 scanning, resume ingestion from the last successfully ingested object key.
  - For SQS-based ingestion, ensure that all relevant message metadata is durably persisted before the message is deleted from the queue.
- Ensure that all ingested files are eventually submitted to a compression job, even after recovery from failures.
- Provide a clear and traceable association between any failed-to-compress files and their corresponding compression jobs.
- Enable improved monitoring and observability for all ingestion jobs across the system (possibly demonstrating in a WebUI).

# Fault-tolerance model

We will partially implement the fault-tolerance model designed for CLP-Spider integration. Some modifications are made to adapt to components that are not yet available in the current system.

![Image](https://github.com/user-attachments/assets/b621c73f-963a-4228-bee1-08bcf81937e1)

- Steps 1–4: Implemented as specified in the Spider design doc.
- Steps 5–6 (Job submission): Instead of submitting compression jobs to Spider (which is not ready yet), the log-ingestor will submit compression jobs directly to the CLP DB.
- Step 7: Implemented as specified in the Spider design doc.
- Steps 8–9 (Job completion tracking): Rather than relying on Spider-side job state callbacks, the log-ingestor will poll the CLP DB until the associated compression job reaches a terminal state.
- Step 10 (Post-compression updates): The log-ingestor will update only the ingested-file status. It will not update archive-specific metadata (e.g., archive IDs), since that information is not currently available from CLP core.

# Storage schema

## Table for ingestion jobs

| Column | Type | NOT NULL | Key     | Auto Inc. | NOTE                                                                                     |
|--------|------|----------|---------|-----------|------------------------------------------------------------------------------------------|
| id     | INTEGER (i64) | ☑ | Primary | ☑ |                                                                                          |
[...]

PR #2001, opened 2026-02-17
feat(log-ingestor): Introduce an abstract `State` layer for fault-tolerance implementation:

LinZhihao-723:
* Define state-related traits to formalize fault-tolerance contracts.
* Refactor `SqsListener` and `S3Scanner` to integrate with state operations.
* Update listener and buffer components to support batched object metadata handling.
* Add `ZeroFaultToleranceIngestionJobState`, an implementation of the state traits that preserves the existing zero fault-tolerance behavior.

# Description

This PR introduces an abstract layer called `State` for fault-tolerance implementation. It abstracts all the state transitions required when persisting ingested files and the related ingestion job state. This abstraction allows us to implement a no-op state to preserve the zero-fault-tolerance mode, while keeping the tester runnable without introducing the persistence layer.

For more details, check the multi-line commit message followed by the PR title.

# Checklist

  breaking change.

# Validation performed

* [x] Verified all unit tests pass after adding `--nocapture` option.
* [x] Tested S3 scanner and SQS listener ingestion jobs work as normal.

## Summary by CodeRabbit

* **Refactor**
  * End-to-end ingestion now handles metadata as batched payloads and threads per-job state through job lifecycle and shutdown.

* **New Features**
  * Pluggable per-job ingestion state abstraction added, with a default "zero-fault" state implementation for start/stop and ingestion hooks.

* **Tests**
  * Test suite updated to exercise batched payloads and the new per-job state paths.

### DIFF
--- a/components/log-ingestor/src/ingestion_job/s3_scanner.rs
+++ b/components/log-ingestor/src/ingestion_job/s3_scanner.rs
@@ -73,8 +73,8 @@ impl<S3ClientManager: AwsClientManagerType<Client>, State: S3ScannerState>
         }
     }
 
-    /// Scans the configured S3 bucket and prefix for new objects and sends their metadata to the
-    /// underlying channel sender.
+    /// Scans the configured S3 bucket and prefix for new objects and ingests their metadata into
+    /// the provided state.
     ///
     /// # Note
     ///
@@ -168,7 +168,7 @@ impl<State: S3ScannerState> S3Scanner<State> {
     /// Creates and spawns a new [`S3Scanner`] backed by a [`Task`].
     ///
     /// This function spawns a [`Task`]. The spawned task will periodically scan the configured S3
-    /// bucket and prefix for new objects and send their metadata to the provided channel sender.
+    /// bucket and prefix for new objects and ingest their metadata into the provided state.
     ///
     /// # Type parameters
     ///
--- a/components/log-ingestor/src/ingestion_job/state.rs
+++ b/components/log-ingestor/src/ingestion_job/state.rs
@@ -68,7 +68,7 @@ pub trait S3ScannerState: Send + Sync + Clone + 'static {
     ///
     /// Implementations **must document** the specific error variants they may return and the
     /// conditions under which those errors occur.
-    async fn ingest<'object_metadata_lifetime>(
+    async fn ingest(
         &self,
         objects: Vec<ObjectMetadata>,
         last_ingested_key: &str,
@@ -119,7 +119,7 @@ impl S3ScannerState for ZeroFaultToleranceIngestionJobState {
     /// Returns an error if:
     ///
     /// * Forwards [`mpsc::Sender::send`]'s return values on failure.
-    async fn ingest<'object_metadata_lifetime>(
+    async fn ingest(
         &self,
         objects: Vec<ObjectMetadata>,
         _last_ingested_key: &str,
--- a/components/log-ingestor/src/ingestion_job/sqs_listener.rs
+++ b/components/log-ingestor/src/ingestion_job/sqs_listener.rs
@@ -37,7 +37,7 @@ impl<SqsClientManager: AwsClientManagerType<Client>, State: SqsListenerState>
     Task<SqsClientManager, State>
 {
     /// Runs the SQS listener task to listen to SQS messages and extract S3 object metadata. The
-    /// extracted metadata is sent to the provided channel sender.
+    /// extracted metadata is ingested into the provided state.
     ///
     /// # Returns
     ///
@@ -77,7 +77,8 @@ impl<SqsClientManager: AwsClientManagerType<Client>, State: SqsListenerState>
         }
     }
 
-    /// Processes the SQS response to extract S3 object metadata and send it to the channel sender.
+    /// Processes the SQS response to extract S3 object metadata and ingests it into the provided
+    /// state.
     ///
     /// # NOTE
     ///
@@ -263,7 +264,7 @@ impl<State: SqsListenerState> SqsListener<State> {
     /// Creates and spawns a new [`SqsListener`] backed by a [`Task`].
     ///
     /// This function spawns a series of [`Task`]. Each spawned task will listen to SQS messages,
-    /// extract relevant S3 object metadata, and send the metadata to the provided channel sender.
+    /// extract relevant S3 object metadata, and ingest the metadata into the provided state.
     ///
     /// # Type parameters
     ///


components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -136,8 +136,8 @@ impl IngestionJobManagerState {
             config.base.endpoint_url.as_ref(),
         )
         .await;
-        self.create_s3_ingestion_job(config.base.clone(), move |job_id, sender| {
-            let scanner = S3Scanner::spawn(job_id, s3_client_manager, config, sender);
+        self.create_s3_ingestion_job(config.base.clone(), move |job_id, sender, state| {
+            let scanner = S3Scanner::spawn(job_id, s3_client_manager, config, sender, state);
             IngestionJob::S3Scanner(scanner)
         })
         .await
```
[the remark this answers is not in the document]
LinZhihao-723: This is expected.

components/log-ingestor/src/ingestion_job/s3_scanner.rs
```diff
@@ -119,10 +126,25 @@ impl<S3ClientManager: AwsClientManagerType<Client>> Task<S3ClientManager> {
...
+            .key
+            .clone()
+            .into();
+        self.state
+            .ingest(&object_metadata_to_ingest, last_ingested_key.as_str())
+            .await?;
+        self.start_after = Some(last_ingested_key);
+        self.sender.send(object_metadata_to_ingest).await?;
+
         Ok(response.is_truncated.unwrap_or(false))
```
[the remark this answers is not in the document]
LinZhihao-723: This is OK in our failure model.

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -284,16 +295,19 @@ impl IngestionJobManagerState {
         // designed to be shared among multiple ingestion jobs in the future.
         let job_listener = self.create_listener(&ingestion_job_config);
         let sender = job_listener.get_new_sender();
+        let state = NoopIngestionJobState::default();
+        state.start().await?;
```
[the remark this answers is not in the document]
LinZhihao-723: This must be in the mutex section.

components/log-ingestor/src/ingestion_job/state.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// # Errors
+    ///
+    /// Implementations **must document** the specific error variants they may return and the
+    /// conditions under which those errors occur.
+    async fn ingest<'object_metadata_lifetime>(
+        &self,
+        objects: &'object_metadata_lifetime [ObjectMetadata],
+        last_ingested_key: &'object_metadata_lifetime str,
+    ) -> anyhow::Result<()>;
+}
```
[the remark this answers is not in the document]
LinZhihao-723: This is because the last ingested key is a reference in the given object view. Being explicit for readability.

components/log-ingestor/src/compression/listener.rs
```diff
@@ -83,7 +79,7 @@ impl<Submitter: BufferSubmitter + Send + 'static> ListenerTask<Submitter> {
 /// Represents a listener that accepts S3 object metadata from multiple senders and buffers them
 /// for submission.
 pub struct Listener {
-    sender: mpsc::Sender<ObjectMetadata>,
+    sender: mpsc::Sender<Vec<ObjectMetadata>>,
```
hoophalab, line 82: Passing a Vec through the channel feels like a duplication of the buffer's functionality.

More critically, it's posiible that `s3_scanner`/`sqs_listener` successfully adds objects to the buffer and deletes the messages, but the buffer isn't full so they are not compressed immediately. The ingestor could crash after message delete and before buffer flushes, and there will be data loss.

It might be best to delete the buffer.rs, or move the buffer's functionality to jobs.

components/log-ingestor/src/ingestion_job/state.rs
```diff
@@ -0,0 +1,114 @@
+use async_trait::async_trait;
+use clp_rust_utils::s3::ObjectMetadata;
+
+/// An abstract layer for managing ingestion job states.
+#[async_trait]
+pub trait IngestionJobState:
```
hoophalab: 1. IngestionJobManager uses `NoopIngestionJobState` directly. `IngestionJobState` isn't actually used.

2. I cannot see why `IngestionJobState: SqsListenerState + S3ScannerState`. If `IngestionJobManager` needs to call state structs defined in `s3_scanner.rs`  and `sqs_listener.rs`, `IngestionJobState` should be simple trait with `fn start` and `fn end`.

3. On an even higher level, there would be only one struct that implements `trait S3ScannerState` in the foreseeable future. It might be best to use two structs in both `s3_scanner.rs`  and `sqs_listener.rs`. and remove `trait SqsListenerState` and `trait S3ScannerState`.

components/log-ingestor/src/compression/listener.rs
hoophalab: outdated docstring

hoophalab (commented):
some comments.

components/log-ingestor/src/compression/listener.rs
LinZhihao-723: * "Passing a Vec through the channel feels like a duplication of the buffer's functionality" is not true. Buffer is operating on a single file/metadata object to trigger size-based flush. Passing a vector through the channel is intended to improve the current batched-receiving design, which is more efficient when multiple SQS listener tasks are concurrently sending things to the channel, since it requires fewer locking operations.
* In the fault-tolerance design, the buffer is recoverable, and there will be no data loss. I will share the design doc with you offline.
* In general, buffering is still an important feature we need to implement inside the log-ingestor. Please review the design doc to see if you have any more comments.

components/log-ingestor/src/ingestion_job/state.rs
LinZhihao-723: 1. `IngestionJobManager` is the server-side implementation, and I do expect it to use a concrete state instead of a generic.
2. `IngestionJobState:  SqsListenerState + S3ScannerState` requires the state to implement all methods required by all supported ingestion job types:
    - This implements all general job state APIs; for example, in the future, we may want to add APIs to get the number of compression jobs submitted vs. the number of files ingested. These general methods will be implemented in `IngestionJobState` and be used in the job level.
    - When going down to the task level, we narrow the trait to the actual job state implementation so that: (1) It avoids job type A to call job type B's specific job methods. (2) It allows different job states to have the same methods but different implementations (like the `ingest` method in the current two job states).

    I can see your concern about `IngestionJobState: SqsListenerState + S3ScannerState`. I think it might make more sense to add this constraint in the job level, for example, the state for an SQS listener might be `State: IngestionJobState + SqsListenerState`.

3. Besides the reason for using a trait as explained above, using a trait allows us to implement a trivial noop state for testing, so that the unit testing won't require an external DB service. I do agree that we can still achieve this by using two structs (which is probably even simpler to implement). But I believe using the trait implementation is a more formal approach for better documentation and type safety.

components/log-ingestor/src/ingestion_job/s3_scanner.rs
hoophalab, line 1: We probably need update the doc strings. How about

``` diff
diff --git a/components/log-ingestor/src/ingestion_job/s3_scanner.rs b/components/log-ingestor/src/ingestion_job/s3_scanner.rs
index 71f0a418..65e4deb0 100644
--- a/components/log-ingestor/src/ingestion_job/s3_scanner.rs
+++ b/components/log-ingestor/src/ingestion_job/s3_scanner.rs
@@ -73,8 +73,8 @@ impl<S3ClientManager: AwsClientManagerType<Client>, State: S3ScannerState>
         }
     }
 
-    /// Scans the configured S3 bucket and prefix for new objects and sends their metadata to the
-    /// underlying channel sender.
+    /// Scans the configured S3 bucket and prefix for new objects and ingests their metadata through
+    /// the underlying state.
     ///
     /// # Note
     ///
@@ -168,7 +168,7 @@ impl<State: S3ScannerState> S3Scanner<State> {
     /// Creates and spawns a new [`S3Scanner`] backed by a [`Task`].
     ///
     /// This function spawns a [`Task`]. The spawned task will periodically scan the configured S3
-    /// bucket and prefix for new objects and send their metadata to the provided channel sender.
+    /// bucket and prefix for new objects and ingest their metadata through the provided state.
     ///
     /// # Type parameters
     ///
diff --git a/components/log-ingestor/src/ingestion_job/sqs_listener.rs b/components/log-ingestor/src/ingestion_job/sqs_listener.rs
index a3a8314b..015d4fb1 100644
--- a/components/log-ingestor/src/ingestion_job/sqs_listener.rs
+++ b/components/log-ingestor/src/ingestion_job/sqs_listener.rs
@@ -37,7 +37,7 @@ impl<SqsClientManager: AwsClientManagerType<Client>, State: SqsListenerState>
     Task<SqsClientManager, State>
 {
     /// Runs the SQS listener task to listen to SQS messages and extract S3 object metadata. The
-    /// extracted metadata is sent to the provided channel sender.
+    /// extracted metadata is ingested through the underlying state.
     ///
     /// # Returns
     ///
@@ -77,7 +77,8 @@ impl<SqsClientManager: AwsClientManagerType<Client>, State: SqsListenerState>
         }
     }
 
-    /// Processes the SQS response to extract S3 object metadata and send it to the channel sender.
+    /// Processes the SQS response to extract S3 object metadata and ingest it through the
+    /// underlying state.
     ///
     /// # NOTE
     ///
@@ -263,7 +264,7 @@ impl<State: SqsListenerState> SqsListener<State> {
     /// Creates and spawns a new [`SqsListener`] backed by a [`Task`].
     ///
     /// This function spawns a series of [`Task`]. Each spawned task will listen to SQS messages,
-    /// extract relevant S3 object metadata, and send the metadata to the provided channel sender.
+    /// extract relevant S3 object metadata, and ingest the metadata through the provided state.
     ///
     /// # Type parameters
     ///
```

components/log-ingestor/src/ingestion_job/state.rs
```diff
@@ -0,0 +1,130 @@
...
+    ///
+    /// # Returns
+    ///
+    /// `Ok(())` on success.
+    ///
+    /// # Errors
+    ///
+    /// Implementations **must document** the specific error variants they may return and the
+    /// conditions under which those errors occur.
+    async fn ingest<'object_metadata_lifetime>(
```
hoophalab: Am I missing something, this lifetime is no longer used? `trait SqsListenerState` also has this lifetime.

```suggestion
    async fn ingest(
```

components/log-ingestor/src/ingestion_job/s3_scanner.rs
LinZhihao-723: Committed with a bit rewording.

hoophalab (approved):

PR #2009, opened 2026-02-19
feat(clp-s): add variable-dictionary bloom filter build flow

ShangDanLuXian:
# Description

This PR adds variable-dictionary Bloom filter building during `clp-s` compression (build-only flow, no packing/query changes).

Main changes:
- Add a `clp_s/filter` module for Bloom filter construction, filter config/type parsing, and filter file I/O.
- Wire filter config through CLI and job orchestration (`filter_type`, `filter_fpr`).
- Build `.var.dict.filter` files during archive write when enabled.
- Keep filter outputs under the data mount (`<data>/filters`) with staging layout for temporary job outputs.
- Add package/deployment plumbing for filters directory config/mount handling.
- Integrate MurmurHash from ANTLR runtime for Bloom hashing and wire required include/source paths in CMake.
- Add tests:
  - C++ unit tests for Bloom filter behavior (positive, mixed, and false-positive-rate sanity checks).
  - Integration test validating filter files are generated during `clp-s` compression.

Scope notes:
- This PR is build-only for filters.
- No filter packing/query-time behavior changes are included.

# Checklist

  breaking change.

# Validation performed

* [x] Unit tests are added and pass.
* [x] Integration tests are added and pass.
* [x] Lint/build checks for touched paths pass in the development environment.

## Release Notes

* **New Features**
  * Variable dictionary filtering using Bloom filters is now supported to optimize archive output
  * New configuration options enable users to specify filter type and false-positive rate for dictionary compression

* **Tests**
  * Added comprehensive Bloom filter unit tests and integration tests to validate filter generation and functionality

### DIFF
--- a/components/core/src/clp_s/clp-s.cpp
+++ b/components/core/src/clp_s/clp-s.cpp
@@ -7,9 +7,9 @@
 #include <string>
 #include <system_error>
 #include <utility>
+#include <vector>
 
 #include <mongocxx/instance.hpp>
-#include <nlohmann/json.hpp>
 #include <spdlog/sinks/stdout_sinks.h>
 #include <spdlog/spdlog.h>
 
@@ -29,7 +29,6 @@
 #include "search/ast/Expression.hpp"
 #include "search/ast/NarrowTypes.hpp"
 #include "search/ast/OrOfAndForm.hpp"
-#include "search/ast/SearchUtils.hpp"
 #include "search/ast/SetTimestampLiteralPrecision.hpp"
 #include "search/ast/TimestampLiteral.hpp"
 #include "search/EvaluateRangeIndexFilters.hpp"
@@ -107,6 +106,8 @@ bool compress(CommandLineArguments const& command_line_arguments) {
     option.single_file_archive = command_line_arguments.get_single_file_archive();
     option.structurize_arrays = command_line_arguments.get_structurize_arrays();
     option.record_log_order = command_line_arguments.get_record_log_order();
+    option.filter_config = command_line_arguments.get_filter_config();
+    option.filter_output_dir = command_line_arguments.get_var_filter_output_dir();
 
     clp_s::JsonParser parser(option);
     if (false == parser.ingest()) {
--- a/components/core/src/clp_s/filter/BloomFilter.cpp
+++ b/components/core/src/clp_s/filter/BloomFilter.cpp
@@ -1,11 +1,11 @@
 #include "BloomFilter.hpp"
 
+#include <misc/MurmurHash.h>
+
 #include <algorithm>
 #include <cmath>
 #include <limits>
 #include <utility>
-
-#include <misc/MurmurHash.h>
 
 #include "../clp/ErrorCode.hpp"
 #include "../clp/ReaderInterface.hpp"
@@ -16,13 +16,11 @@ constexpr uint32_t kMinNumHashFunctions{1};
 constexpr uint32_t kMaxNumHashFunctions{20};
 
 auto compute_murmur_hash(std::string_view value, uint64_t seed) -> uint64_t {
-    return static_cast<uint64_t>(
-            antlr4::misc::MurmurHash::hashCode(
-                    value.data(),
-                    value.size(),
-                    static_cast<size_t>(seed)
-            )
-    );
+    return static_cast<uint64_t>(antlr4::misc::MurmurHash::hashCode(
+            value.data(),
+            value.size(),
+            static_cast<size_t>(seed)
+    ));
 }
 }  // namespace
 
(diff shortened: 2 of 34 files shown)


components/core/src/clp_s/clp-s.cpp
```diff
@@ -7,9 +7,9 @@
 #include <string>
 #include <system_error>
 #include <utility>
+#include <vector>
 
 #include <mongocxx/instance.hpp>
-#include <nlohmann/json.hpp>
```
ShangDanLuXian, line 12: Not sure why these two imports are touched, maybe because I run the cpp format fixer?
But the project can be compiled

components/core/src/clp_s/clp-s.cpp
```diff
@@ -29,7 +29,6 @@
 #include "search/ast/Expression.hpp"
 #include "search/ast/NarrowTypes.hpp"
 #include "search/ast/OrOfAndForm.hpp"
-#include "search/ast/SearchUtils.hpp"
```
ShangDanLuXian, line 32: Same as above

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include "../clp/ReaderInterface.hpp"
+
+namespace clp_s::filter {
+namespace {
+constexpr uint32_t kMinNumHashFunctions{1};
+constexpr uint32_t kMaxNumHashFunctions{20};
+
+auto compute_murmur_hash(std::string_view value, uint64_t seed) -> uint64_t {
+    return static_cast<uint64_t>(antlr4::misc::MurmurHash::hashCode(
+            value.data(),
```
ShangDanLuXian, line 20: I used murmurhash in antlr4 - I see that there is only sha256 in hash_utils and it is not suitable for bloom filters because it is slow.

I searched that murmurhash is the industry standard for bloom filters, I did this way because it the most easy way, but I feel it's not good dependency cycle.

Maybe better ways are include some git submodules (e.g. SMHasher), or a static cpp file of murmurhash, but thats hundreds of loc

gibber9809 (changes requested):
Some initial high level feedback:
1. This PR is trying to do too many things at once, which makes it hard to review -- I think that this could pretty easily be split into a series of at least 3 PRs (Just the contents of src/clp_s/filter as an independent library with tests -> integration into clp-s with tests -> package integration).
2. You're right that we need to think a bit more about pulling in or writing our own hash function -- pulling in a murmur implementation from antlr internals doesn't really seem maintainable.

For that first point, could we reduce the scope of this first PR to just creating the library for creating/working with bloom filters? Either shrinking this first PR, or closing it and opening a new PR with that smaller set of changes would be fine.

For the second point, I think that one way or another we need to build confidence that the hash function implementation we're using for our bloom filters is fast enough, won't change without our knowledge, and will work across different deployment contexts (x86, arm, embedded, etc.).

If we go the route of pulling in a library, you can reference how we manage library dependencies [here](https://github.com/y-scope/clp/blob/main/taskfiles/deps/main.yaml). From looking around though it seems like most big c++ projects just implement their own version of murmur3 or cityhash instead of pulling in a library though -- honestly since it's only ~200 lines it could be worth copying an implementation from a project with a permissive license with proper credit instead of pulling in a library, just so that we can be confident that the hash won't subtly change across library versions or something. Either way, just need to make some well-reasoned choice for this.

ShangDanLuXian:
> Some initial high level feedback:
> 
> 1. This PR is trying to do too many things at once, which makes it hard to review -- I think that this could pretty easily be split into a series of at least 3 PRs (Just the contents of src/clp_s/filter as an independent library with tests -> integration into clp-s with tests -> package integration).
> 2. You're right that we need to think a bit more about pulling in or writing our own hash function -- pulling in a murmur implementation from antlr internals doesn't really seem maintainable.
> 
> For that first point, could we reduce the scope of this first PR to just creating the library for creating/working with bloom filters? Either shrinking this first PR, or closing it and opening a new PR with that smaller set of changes would be fine.
> 
> For the second point, I think that one way or another we need to build confidence that the hash function implementation we're using for our bloom filters is fast enough, won't change without our knowledge, and will work across different deployment contexts (x86, arm, embedded, etc.).
> 
> If we go the route of pulling in a library, you can reference how we manage library dependencies [here](https://github.com/y-scope/clp/blob/main/taskfiles/deps/main.yaml). From looking around though it seems like most big c++ projects just implement their own version of murmur3 or cityhash instead of pulling in a library though -- honestly since it's only ~200 lines it could be worth copying an implementation from a project with a permissive license with proper credit instead of pulling in a library, just so that we can be confident that the hash won't subtly change across library versions or something. Either way, just need to make some well-reasoned choice for this.

Thanks for your prompt review.
I will close this PR and open a new PR for the filter library only.
Regarding MurmurHash, I’d prefer to copy the implementation. This gives us full control and ensures the hashing logic remains immutable. It's critical that the hash behavior stays consistent to maintain the integrity of existing filters.

PR #2010, opened 2026-02-20
feat(clp_s::filter): Add library for representing bloom filters; Add xxHash v0.8.3 as a dependency

ShangDanLuXian:
# Description

This is the first PR for the issue #2090 
This PR is intentionally narrowed to a standalone `clp_s` Bloom filter library + xxHash integration + unit tests.

Main changes:
- Add `BloomFilter` implementation under `components/core/src/clp_s/filter/`:
  - `BloomFilter.cpp/.hpp`
  - `ErrorCode.cpp/.hpp`
  - `XxHash.cpp/.hpp`
- Add dependency plumbing for `xxHash` (`v0.8.3`) in `taskfiles/deps/main.yaml`.
- Add/adjust CMake wiring for `clp_s::filter`:
  - `components/core/src/clp_s/filter/CMakeLists.txt`
  - `components/core/src/clp_s/CMakeLists.txt`
  - `components/core/CMakeLists.txt`
  - `components/core/cmake/Options/options.cmake`
- Add unit tests under `components/core/src/clp_s/filter/tests/`:
  - `test-clp_s-bloom_filter.cpp`
  - `test-clp_s-xxhash.cpp`
- Pin xxHash usage/version expectations in tests via seeded sanity vectors (`XXH3_64bits_withSeed`, `v0.8.3`).

Why `xxHash`:
- Performance: Bloom filtering is hash-heavy; `xxHash` is fast and suitable for this workload.
- Portability: upstream implementation is widely used across architectures/toolchains.
- Stability: dependency is pinned in our deps flow, and test vectors are included to detect drift in hash behavior.

Out of scope in this PR:
- integration work for other filter-library components (beyond BloomFilter/xxHash/unit tests)
- archive writer integration
- package/runtime plumbing
- query/runtime behavior changes

# Checklist

  breaking change.

# Validation performed

- Added unit tests:
  - `components/core/src/clp_s/filter/tests/test-clp_s-bloom_filter.cpp`
  - `components/core/src/clp_s/filter/tests/test-clp_s-xxhash.cpp`
- Verified build/dependency wiring updates for `clp_s::filter` and `xxHash` integration.

### DIFF
--- a/components/core/src/clp_s/filter/HashAlgorithm.cpp
+++ b/components/core/src/clp_s/filter/HashAlgorithm.cpp
@@ -8,12 +8,11 @@
 
 namespace clp_s::filter {
 auto try_parse_hash_algorithm(uint8_t hash_algorithm) -> std::optional<HashAlgorithm> {
-    switch (hash_algorithm) {
-        case static_cast<uint8_t>(HashAlgorithm::Xxh364):
-            return HashAlgorithm::Xxh364;
-        default:
-            return std::nullopt;
+    if (static_cast<uint8_t>(HashAlgorithm::Xxh364) == hash_algorithm) {
+        return HashAlgorithm::Xxh364;
     }
+
+    return std::nullopt;
 }
 
 auto hash64(HashAlgorithm hash_algorithm, std::string_view value, uint64_t seed) -> uint64_t {


components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,162 @@
...
+#include "../../clp/ErrorCode.hpp"
+#include "../../clp/ReaderInterface.hpp"
+#include "XxHash.hpp"
+
+namespace clp_s::filter {
+namespace {
+constexpr size_t kDefaultBitArraySize{64};
+constexpr uint32_t kDefaultNumHashFunctions{1};
+constexpr uint32_t kMinNumHashFunctions{1};
+constexpr uint32_t kMaxNumHashFunctions{20};
```
ShangDanLuXian: Correspond to 1-e6 FPR, which will be set as the min FPR in the integration

components/core/src/clp_s/filter/FilterBuilder.cpp
```diff
@@ -0,0 +1,34 @@
...
+bool FilterBuilder::write(std::string const& filter_path) const {
+    if (false == m_enabled) {
+        return false;
+    }
+    FileWriter writer;
+    writer.open(filter_path, FileWriter::OpenMode::CreateForWriting);
+    write_filter_file(writer, m_config.type, m_filter);
+    writer.close();
+    return true;
+}
```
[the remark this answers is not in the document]
ShangDanLuXian: I am intentionally writing like this because I checked that in JsonConstructor.cpp and SchemaMap.cpp there is also no protection against writer.close(), so make it consistent. Also that the loss of data in the writer is acceptable in the context of filter.

I am actually considerring add try catch in the caller(integration) but not here

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,162 @@
...
+constexpr uint64_t kPrimaryHashSeed{0};
+constexpr uint64_t kSecondaryHashSeed{0x9e37'79b9'7f4a'7c15ULL};
```
gibber9809:
```suggestion
constexpr size_t cDefaultBitArraySize{64};
constexpr uint32_t cDefaultNumHashFunctions{1};
constexpr uint32_t cMinNumHashFunctions{1};
constexpr uint32_t cMaxNumHashFunctions{20};
constexpr uint64_t cPrimaryHashSeed{0};
constexpr uint64_t cSecondaryHashSeed{0x9e37'79b9'7f4a'7c15ULL};
```
Nit: style for constants in the repo is to use the prefix `c` -- sorry, this is one of the details that is only in the private part of our style guide.

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,162 @@
...
+}  // namespace
+
+BloomFilter::BloomFilter()
+        : m_bit_array_size(kDefaultBitArraySize),
+          m_num_hash_functions(kDefaultNumHashFunctions),
+          m_bit_array((kDefaultBitArraySize + 7) / 8, 0) {}
```
gibber9809:
```suggestion
        : m_bit_array_size{cDefaultBitArraySize},
          m_num_hash_functions{cDefaultNumHashFunctions},
          m_bit_array{(cDefaultBitArraySize + 7) / 8, 0} {}
```
Nit: preferred style is to use brace initializers wherever possible -- sorry again, this also is only documented in the private part of our style guide.

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+BloomFilter::BloomFilter(size_t expected_num_elements, double false_positive_rate) {
+    auto [bit_array_size, num_hash_functions]
+            = compute_optimal_parameters(expected_num_elements, false_positive_rate);
+
+    m_bit_array_size = bit_array_size;
+    m_num_hash_functions = num_hash_functions;
+
+    size_t const num_bytes = (m_bit_array_size + 7) / 8;
```
gibber9809: This code seems to use (x + 7)/8 to get the number of bytes that contain a given number of bits in several places. Might be worth adding a helper called `min_bytes_containing_bits` in the anonymonous namespace in this file or something.

Might also be worth defining a constant `cNumBitsInByte` to replace all of the instances of 8 in this file for readability.

components/core/src/clp_s/filter/XxHash.cpp
```diff
@@ -0,0 +1,10 @@
+#include "XxHash.hpp"
+
+#define XXH_INLINE_ALL
```
gibber9809: What does this do?

components/core/src/clp_s/CMakeLists.txt
```diff
@@ -232,6 +232,45 @@ if(CLP_BUILD_CLP_S_IO)
...
+                ${CLP_XXHASH_INCLUDE_DIRECTORY}
+        )
+        target_link_libraries(
+                clp_s_filter
+                PUBLIC
+                clp_s::io
+                PRIVATE
+                clp::string_utils
+        )
+endif()
```
gibber9809, line 255: Could we move all of this stuff for defining the library into a new `filter/CMakeLists.txt`, then just use `add_subdirectory` from this `CMakeLists` file to add that subdirectory?

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,162 @@
+#include "BloomFilter.hpp"
+
+#include <algorithm>
+#include <cmath>
+#include <limits>
+#include <utility>
+
+#include "../../clp/ErrorCode.hpp"
+#include "../../clp/ReaderInterface.hpp"
```
gibber9809:
```suggestion
#include <clp/ErrorCode.hpp>
#include <clp/ReaderInterface.hpp>
```
Nit: `<>` is our preferred include style for pulling in other parts of the clp project.

components/core/src/clp_s/filter/BloomFilter.hpp
```diff
@@ -0,0 +1,42 @@
+#ifndef CLP_S_FILTER_BLOOM_FILTER_HPP
+#define CLP_S_FILTER_BLOOM_FILTER_HPP
+
+#include <cstddef>
+#include <cstdint>
+#include <string_view>
+#include <utility>
+#include <vector>
+
+#include "../FileWriter.hpp"
```
gibber9809:
```suggestion
#include <clp_s/FileWriter.hpp>
```
Nit

taskfiles/deps/main.yaml
```diff
@@ -558,6 +559,26 @@ tasks:
...
+      - task: "yscope-dev-utils:remote:download-and-extract-tar"
+        vars:
+          CHECKSUM_FILE: "{{.G_DEPS_CPP_CHECKSUMS_DIR}}/{{.LIB_NAME}}.md5"
+          FILE_SHA256: "baee0c6afd4f03165de7a4e67988d16f0f2b257b51d0e3cb91909302a26a79c4"
+          OUTPUT_DIR: "{{.XXHASH_OUTPUT_DIR}}"
+          URL: "https://github.com/Cyan4973/xxHash/archive/refs/tags/v0.8.2.tar.gz"
+      - >-
+        echo "set(
+        CLP_XXHASH_INCLUDE_DIRECTORY \"{{.XXHASH_OUTPUT_DIR}}\"
+        )" > "{{.G_DEPS_CPP_CMAKE_SETTINGS_DIR}}/{{.LIB_NAME}}.cmake"
```
gibber9809:
```suggestion
  xxhash:
    internal: true
    vars:
      LIB_NAME: "xxHash"
      XXHASH_OUTPUT_DIR: "{{.G_DEPS_CPP_DIR}}/{{.LIB_NAME}}-src"
    run: "once"
    cmds:
      - task: "utils:install-remote-cmake-lib"
        vars:
          CMAKE_GEN_ARGS:
            - "-DXXHASH_BUILD_XXHSUM=OFF"
            - "-DBUILD_SHARED_LIBS=ON"
            - '-DCMAKE_INSTALL_PREFIX={{.G_DEPS_CPP_DIR}}/{{.LIB_NAME}}-install'
          LIB_NAME: "{{.LIB_NAME}}"
          TARBALL_SHA256: "aae608dfe8213dfd05d909a57718ef82f30722c392344583d3f39050c7f29a80"
          TARBALL_URL: "https://github.com/Cyan4973/xxHash/archive/refs/tags/v0.8.3.tar.gz"
          CMAKE_SOURCE_DIR: "cmake_unofficial"
```
Changing this to pull in v0.8.3 as well as use the cmake install flow.

Now instead of the `CLP_XXHASH_INCLUDE_DIRECTORY` stuff you're doing, you can directly use `xxHash::xxhash` as a link target in `target_link_libraries` for the relevant targets.

You will also need to add a block like:
```
if(CLP_NEED_XXHASH)
    find_package(xxHash 0.8.3 CONFIG REQUIRED)
    if(xxHash_FOUND)
        message(STATUS "Found xxHash ${xxHash_VERSION}")
    endif()
endif()
```
To `components/core/CMakeLists.txt`
and you'll need to add functions/flags that set `CLP_NEED_XXHASH` in `components/core/cmake/Options/options.cmake` -- sorry that this isn't documented, but you'll pretty much want to copy what we do for other targets. That means you'll have to add an option flag like

```
option(
    CLP_BUILD_CLP_S_FILTER
    "Build clp_s::filter."
    ON
)
```
Alphabetically sorted in the right position in the long list of options at the start of the file + add a function
`validate_clp_s_filter_dependencies` which asks for `CLP_BUILD_CLP_STRING_UTILS`+`CLP_BUILD_CLP_S_IO` and `set_clp_s_filter_dependencies` which sets `CLP_NEED_XXHASH` + call these two functions in `validate_and_setup_all_clp_dependency_flags` like we do for other dependencies + add `CLP_NEED_XXHASH` to the list in `convert_clp_dependency_properties_to_variables`.

Hopefully you can figure that out by copying what we do for other dependencies, but if not let me know and I can help -- we really should have documented how to set up these flags/refactored them by now, but haven't yet, so it's more confusing than it needs to be.

components/core/src/clp_s/CMakeLists.txt
```diff
@@ -232,6 +232,45 @@ if(CLP_BUILD_CLP_S_IO)
...
+        filter/FilterConfig.hpp
+        filter/FilterFile.cpp
+        filter/FilterFile.hpp
+        filter/FilterReader.cpp
+        filter/FilterReader.hpp
+        filter/XxHash.cpp
+        filter/XxHash.hpp
+)
+
+if(CLP_BUILD_CLP_S_IO)
```
gibber9809:
```suggestion
if(CLP_BUILD_CLP_S_FILTER)
```
See my other comment about how to set up this flag properly.

components/core/src/clp_s/CMakeLists.txt
```diff
@@ -232,6 +232,45 @@ if(CLP_BUILD_CLP_S_IO)
...
+                ${CLP_S_FILTER_SOURCES}
+        )
+        add_library(clp_s::filter ALIAS clp_s_filter)
+        target_compile_features(clp_s_filter PRIVATE cxx_std_20)
+        target_include_directories(
+                clp_s_filter
+                PUBLIC
+                ../
+                PRIVATE
+                ${CLP_XXHASH_INCLUDE_DIRECTORY}
```
gibber9809: Per other comment, should be able to remove this once you change how the dependency is pulled in.

components/core/src/clp_s/CMakeLists.txt
```diff
@@ -232,6 +232,45 @@ if(CLP_BUILD_CLP_S_IO)
...
+        )
+        target_link_libraries(
+                clp_s_filter
+                PUBLIC
+                clp_s::io
```
gibber9809: Per other comment, will need to link against `xxHash::xxhash`.

components/core/CMakeLists.txt
```diff
@@ -761,6 +774,7 @@ if(CLP_ENABLE_TESTS)
     target_include_directories(unitTest
             PRIVATE
             ${CLP_SQLITE3_INCLUDE_DIRECTORY}
+            ${CLP_XXHASH_INCLUDE_DIRECTORY}
```
gibber9809:
```suggestion
```
Per other comment, can get rid of this when we update how we pull in the xxhash dependency.

components/core/CMakeLists.txt
```diff
@@ -366,6 +366,18 @@ set(SOURCE_FILES_clp_s_unitTest
...
+    src/clp_s/filter/FilterBuilder.cpp
+    src/clp_s/filter/FilterBuilder.hpp
+    src/clp_s/filter/FilterConfig.cpp
+    src/clp_s/filter/FilterConfig.hpp
+    src/clp_s/filter/FilterFile.cpp
+    src/clp_s/filter/FilterFile.hpp
+    src/clp_s/filter/FilterReader.cpp
+    src/clp_s/filter/FilterReader.hpp
+    src/clp_s/filter/XxHash.cpp
+    src/clp_s/filter/XxHash.hpp
```
gibber9809: Instead of pulling in every source file from `clp_s/filter` could we just link `unitTest` against `clp_s::filter`?

components/core/tests/test-clp_s-bloom_filter.cpp
```diff
@@ -0,0 +1,71 @@
+#include <cmath>
+#include <cstddef>
+#include <cstdint>
+#include <string>
+#include <vector>
+
+#include <catch2/catch_test_macros.hpp>
+
+#include "../src/clp_s/filter/BloomFilter.hpp"
```
gibber9809:
```suggestion
#include <clp_s/filter/BloomFilter.hpp>
```
Since this test is linking against the `clp_s::filter` library.

components/core/tests/test-clp_s-bloom_filter.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+namespace {
+constexpr size_t kInsertions = 10'000;
+constexpr size_t kQueries = 10'000;
+constexpr double kFalsePositiveRate = 0.001;
```
gibber9809:
```suggestion
constexpr size_t cInsertions{10'000};
constexpr size_t cQueries{10'000};
constexpr double cFalsePositiveRate{0.001};
```
Nit: style.

components/core/tests/test-clp_s-bloom_filter.cpp
```diff
@@ -0,0 +1,71 @@
...
+}  // namespace
+
+TEST_CASE("BloomFilter empty filter has no matches", "[clp_s][filter]") {
+    clp_s::filter::BloomFilter filter{0, kFalsePositiveRate};
+
+    for (size_t i = 0; i < kQueries; ++i) {
+        auto const value = make_odd(i);
+        REQUIRE_FALSE(filter.possibly_contains(std::to_string(value)));
+    }
+}
```
gibber9809: Do you think this test is required? It doesn't seem to actually test anything besides that our bit vector is initialized to zero, which the other test would catch anyway.

components/core/tests/test-clp_s-bloom_filter.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+uint64_t make_odd(uint64_t value) {
+    return value * 2 + 1;
+}
+
+uint64_t make_even(uint64_t value) {
+    return value * 2;
+}
+}  // namespace
+
```
gibber9809: Do you think it's worth adding a test that runs xxhash on a handful of inputs and checks that the hash value is what we expect? Just to help avoid a situation where someone somehow links this library against a pre 0.8.0 version of xxhash and gets different hash results.

components/core/tests/test-clp_s-bloom_filter.cpp
```diff
@@ -0,0 +1,71 @@
+#include <cmath>
+#include <cstddef>
+#include <cstdint>
+#include <string>
+#include <vector>
+
+#include <catch2/catch_test_macros.hpp>
```
gibber9809:
```suggestion
#include <catch2/catch_message.hpp>
#include <catch2/catch_test_macros.hpp>
```
Since you're also using `INFO`, etc

components/core/tests/test-clp_s-bloom_filter.cpp
```diff
@@ -0,0 +1,71 @@
...
+            ++false_positives;
+        }
+    }
+
+    auto const false_positive_rate
+            = static_cast<double>(false_positives) / static_cast<double>(kQueries);
+    double const sigma = std::sqrt(
+            kFalsePositiveRate * (1.0 - kFalsePositiveRate) / static_cast<double>(kQueries)
+    );
+    double const allowed_false_positive_rate = kFalsePositiveRate + 3.0 * sigma;
```
gibber9809:
```suggestion
    auto const false_positive_rate
            = static_cast<double>(false_positives) / static_cast<double>(kQueries);
    double const sigma{std::sqrt(
            kFalsePositiveRate * (1.0 - kFalsePositiveRate) / static_cast<double>(kQueries)
    )};
    double const allowed_false_positive_rate{kFalsePositiveRate + 3.0 * sigma};
```
Nit: style -- the main situations where we don't require `{}` initialization are for `auto xyz =` or situations where it changes the semantics of the initialization (like with `nlohmann::json`), otherwise it's required.

components/core/tests/test-clp_s-bloom_filter.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    size_t false_negatives = 0;
+    for (auto const& value : inserted) {
+        if (false == filter.possibly_contains(value)) {
+            ++false_negatives;
+        }
+    }
+    REQUIRE(false_negatives == 0);
+
+    size_t false_positives = 0;
```
gibber9809:
```suggestion
    size_t false_positives{};
```

components/core/tests/test-clp_s-bloom_filter.cpp
```diff
@@ -0,0 +1,71 @@
...
+    std::vector<std::string> inserted;
+    inserted.reserve(kInsertions);
+
+    for (size_t i = 0; i < kInsertions; ++i) {
+        auto const value = make_odd(i);
+        inserted.emplace_back(std::to_string(value));
+        filter.add(inserted.back());
+    }
+
+    size_t false_negatives = 0;
```
gibber9809:
```suggestion
    size_t false_negatives{};
```

components/core/tests/test-clp_s-bloom_filter.cpp
```diff
@@ -0,0 +1,71 @@
...
+constexpr size_t kQueries = 10'000;
+constexpr double kFalsePositiveRate = 0.001;
+
+uint64_t make_odd(uint64_t value) {
+    return value * 2 + 1;
+}
+
+uint64_t make_even(uint64_t value) {
+    return value * 2;
+}
```
gibber9809:
```suggestion
[[nodiscard]] auto make_odd(uint64_t value) -> uint64_t {
    return value * 2 + 1;
}

[[nodiscard]] auto make_even(uint64_t value) -> uint64_t {
    return value * 2;
}
```
Nit: our style guide requires trailing return syntax.

components/core/src/clp_s/filter/XxHash.hpp
```diff
@@ -0,0 +1,11 @@
+#ifndef CLP_S_FILTER_XX_HASH_HPP
+#define CLP_S_FILTER_XX_HASH_HPP
+
+#include <cstdint>
+#include <string_view>
+
+namespace clp_s::filter::xxhash {
+[[nodiscard]] uint64_t hash64(std::string_view value, uint64_t seed);
```
gibber9809:
```suggestion
/**
 * @param value
 * @param seed
 * @return The 64-bit seeded XXH3 hash of `value`.
 */
[[nodiscard]] auto hash64(std::string_view value, uint64_t seed) -> uint64_t ;
```
Nit: trailing return syntax.
Nit: missing docstring.

components/core/src/clp_s/filter/XxHash.cpp
```diff
@@ -0,0 +1,9 @@
+#include "XxHash.hpp"
+
+#include <xxhash.h>
```
gibber9809, line 6:
```suggestion
#include <cstdint>
#include <string_view>

#include <xxhash.h>
```
Nit: IWYU -- need to include headers for all symbols used in a file.

components/core/src/clp_s/filter/XxHash.cpp
```diff
@@ -0,0 +1,9 @@
...
+
+namespace clp_s::filter::xxhash {
+uint64_t hash64(std::string_view value, uint64_t seed) {
```
gibber9809:
```suggestion
auto hash64(std::string_view value, uint64_t seed) -> uint64_t {
```
Nit: trailing return syntax

ShangDanLuXian:
> Going to go through more later today, but leaving another set of comments on tests + `XxHash.{cpp,hpp}`.
> 
> The updates to cmake scripts look good!
> 
> Besides those comments could you:
> 
> 1. Move your `tests` subdirectory to `filter/tests` -- our convention changed a few months ago to putting tests in a `tests` subdirectory next to the thing they're testing so that it's easier to find tests related to a given library or feature.
> 2. Run clang-tidy on all of your c++ code? I'm finding a lot of details that our clang-tidy config would complain about. If the `task` flow for running `clang-tidy` isn't bringing up the issues, you may have to manually run clang-tidy like `clang-tidy -p your/build/path/compile_commands.json components/core/src/clp_s/filter/BloomFilter.cpp` etc.

Thanks for the review. Sorry that forgot to look into the test files and change the styling as well. Also forgot to run clang-tidy again after last commit, will fix it in the next commit.

components/core/tests/test-clp_s-bloom_filter.cpp
ShangDanLuXian: I was thinking before that make some pre loaded input and output for testing but that will lead to a large number of loc - not a large file but just loc. Or do you think a very small set (like 10 sparse cases) for senity check is enough?

components/core/src/clp_s/filter/XxHash.hpp
ShangDanLuXian: Also see some other missing docstrings, will add all.

components/core/src/clp_s/filter/BloomFilter.hpp
```diff
@@ -0,0 +1,42 @@
...
+#include <cstdint>
+#include <string_view>
+#include <utility>
+#include <vector>
+
+#include <clp_s/FileWriter.hpp>
+
+namespace clp {
+class ReaderInterface;
+}  // namespace clp
```
gibber9809: For the write_to_file  interface I think it might make sense to rely on `clp::WriterInterface &` to make it more generic -- that way if we decide to e.g. compress this bloom filter using zstd or something we don't need to change any code here.

Also, for these includes it's fine to just include `<clp/ReaderInterface.hpp>` and `<clp/WriterInterface.hpp>` here -- no need for this forward declaration of `ReaderInterface` to avoid pulling in that header.

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+    double const ln2 = std::log(2.0);
+    double const ln2_squared = ln2 * ln2;
+    auto const ideal_bit_array_size
+            = (-static_cast<double>(expected_num_elements) * std::log(false_positive_rate)
+               / ln2_squared);
+    if (false == std::isfinite(ideal_bit_array_size)
+        || ideal_bit_array_size > static_cast<double>(std::numeric_limits<size_t>::max()))
+    {
+        return {cDefaultBitArraySize, cDefaultNumHashFunctions};
```
gibber9809: It seems like this and the other default return above are hiding errors for a few kinds of invalid inputs.

I think we're better off just forwarding the error in this kind of situation instead of silently falling back to a default. The easiest way to do that might be to use result types for returns here like we've started using in other parts of the codebase, e.g. `ystdlib::error_handling::Result<std::pair<size_t, uint32_t>>`.

You can take a look at `clp_s/timestamp_parser/TimestampParser.{cpp,hpp}` to see a somewhat recent example of how we use result types in the codebase.

components/core/src/clp_s/filter/BloomFilter.hpp
```diff
@@ -0,0 +1,42 @@
...
+private:
+    [[nodiscard]] static std::pair<size_t, uint32_t>
+    compute_optimal_parameters(size_t expected_num_elements, double false_positive_rate);
+
+    void set_bit(size_t bit_index);
+    [[nodiscard]] bool test_bit(size_t bit_index) const;
+
+    size_t m_bit_array_size{0};
+    uint32_t m_num_hash_functions{0};
+    std::vector<uint8_t> m_bit_array;
```
gibber9809: We have a class `ystdlib::containers::Array<T>` to represent fixed sized arrays with size configured at construction time that might be a better fit than an `std::vector` here.

components/core/src/clp_s/filter/BloomFilter.hpp
```diff
@@ -0,0 +1,42 @@
...
+public:
+    BloomFilter(size_t expected_num_elements, double false_positive_rate);
+
+    BloomFilter();
+
+    void add(std::string_view value);
+    [[nodiscard]] bool possibly_contains(std::string_view value) const;
+
+    void write_to_file(FileWriter& writer) const;
+    [[nodiscard]] bool read_from_file(clp::ReaderInterface& reader);
```
gibber9809: I think this might make more sense as a static function that returns `ystdlib::error_handling::Result<BloomFilter>` -- kind of matching the `create()` pattern used in other parts of the codebase. Mainly my reasoning is that it avoids having to default initialize a `BloomFilter` first before attempting to initialize it from some `ReaderInterface`, avoiding some unnecessary allocations + cleans up the error handling since it makes it so that if you're able to create the bloom filter it is guaranteed to be valid. 

Does that make sense, or do you think the current approach is better?

components/core/src/clp_s/filter/BloomFilter.hpp
```diff
@@ -0,0 +1,42 @@
...
+namespace clp {
+class ReaderInterface;
+}  // namespace clp
+
+namespace clp_s::filter {
+class BloomFilter {
+public:
+    BloomFilter(size_t expected_num_elements, double false_positive_rate);
+
+    BloomFilter();
```
gibber9809: What's the use case for the default constructor?

components/core/tests/test-clp_s-bloom_filter.cpp
gibber9809: Yeah, 5-10 cases as a sanity check is enough I think.

components/core/src/clp_s/filter/BloomFilter.hpp
ShangDanLuXian: The default constructor was only needed in the old flow where we default-constructed a BloomFilter as a placeholder and then populated.
With the new Result<BloomFilter>-based API (create / try_read_from_file) and std::optional at call sites, we no longer need placeholder construction, I have removed the default constructor.

components/core/src/clp_s/filter/BloomFilter.hpp
ShangDanLuXian: I feel this is better, also replied in the default constructor comment that the default constructor is not needed with this approach.

components/core/src/clp_s/filter/BloomFilter.cpp
ShangDanLuXian: Currently I treat expected_num_elements == 0 as valid input.
In the variable-dictionary path this should be impossible, but for the Bloom filter library itself it is still a valid state (empty set), so the implementation returns a default empty filter instead of an error.

Please let me know if you’d prefer this to be treated as an error instead.

components/core/src/clp_s/CMakeLists.txt
ShangDanLuXian: should we not add the package into this cmakelist in this PR - it should be part of the integration right?

gibber9809: It's fine to add the cmake library in this PR since we aren't linking it into clp-s. It's also got a flag to disable if we need, so no real risk of adverse side effects.

components/core/src/clp_s/filter/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+                clp_s_filter
+                ${CLP_S_FILTER_SOURCES}
+        )
+        add_library(clp_s::filter ALIAS clp_s_filter)
+        target_compile_features(clp_s_filter PRIVATE cxx_std_20)
+        target_include_directories(clp_s_filter PUBLIC ../..)
+        target_link_libraries(
+                clp_s_filter
+                PUBLIC
+                clp_s::io
```
gibber9809:
```suggestion
                clp_s::clp_dependencies
```

Since this is now using `clp::ReaderInterface` and `clp::WriterInterface` I think we don't need to depend on `clp_s::io` anymore, but still need the clp-related dependencies.

components/core/cmake/Options/options.cmake
```diff
@@ -270,6 +277,17 @@ function(set_clp_s_clp_dependencies_dependencies)
     )
 endfunction()
 
+function(validate_clp_s_filter_dependencies)
+    validate_clp_dependencies_for_target(CLP_BUILD_CLP_S_FILTER)
```
gibber9809:
```suggestion
    validate_clp_dependencies_for_target(CLP_BUILD_CLP_S_FILTER
        CLP_BUILD_CLP_S_CLP_DEPENDENCIES
    )
```
Same as my other comment about depending on clp_s_clp_dependencies + I missed that this wasn't checking for deps previously.

gibber9809 (changes requested):
Could you re-run clang-tidy on `BloomFilter.cpp` and `BloomFilter.hpp`? I still see some clang-tidy errors when I pull your branch and run it on those files.

For the error about `reinterpret_cast` I think you can use `clp::size_checked_pointer_cast` instead (from `clp/type_utils.hpp`).

I'm still taking a look at whether we need to modify how we pull in xxhash in the taskfile deps flow -- I'll get back to you on that once I've figured it out.

ShangDanLuXian:
> Could you re-run clang-tidy on `BloomFilter.cpp` and `BloomFilter.hpp`? I still see some clang-tidy errors when I pull your branch and run it on those files.
> 
> For the error about `reinterpret_cast` I think you can use `clp::size_checked_pointer_cast` instead (from `clp/type_utils.hpp`).
> 
> I'm still taking a look at whether we need to modify how we pull in xxhash in the taskfile deps flow -- I'll get back to you on that once I've figured it out.

Sorry that I misunderstood how clang-tidy works previously, now should all be fixed :)

components/core/CMakeLists.txt
```diff
@@ -329,6 +329,13 @@ if(CLP_NEED_SQLITE)
     FindDynamicLibraryDependencies(sqlite "${sqlite_DYNAMIC_LIBS}")
 endif()
 
+if(CLP_NEED_XXHASH)
+    find_package(xxHash 0.8.3 EXACT CONFIG REQUIRED)
```
gibber9809:
```suggestion
    find_package(xxHash 0.8.3 CONFIG REQUIRED)
```
Removing EXACT per sidebar.

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,192 @@
...
+}  // namespace
+
+BloomFilter::BloomFilter(
+        size_t bit_array_size,
+        uint32_t num_hash_functions,
+        ystdlib::containers::Array<uint8_t> bit_array
+)
+        : m_bit_array_size(bit_array_size),
+          m_num_hash_functions(num_hash_functions),
+          m_bit_array(std::move(bit_array)) {}
```
gibber9809:
```suggestion
        : m_bit_array_size{bit_array_size},
          m_num_hash_functions{num_hash_functions},
          m_bit_array{std::move(bit_array)} {}
```
Nit: Use `{}` for initialization

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+auto BloomFilter::create(size_t expected_num_elements, double false_positive_rate)
+        -> ystdlib::error_handling::Result<BloomFilter> {
+    auto const optimal_parameters
+            = compute_optimal_parameters(expected_num_elements, false_positive_rate);
+    if (optimal_parameters.has_error()) {
+        return optimal_parameters.error();
+    }
```
gibber9809:
```suggestion
    auto const optimal_parameters{YSTDLIB_ERROR_HANDLING_TRYX(compute_optimal_parameters(expected_num_elements, false_positive_rate)};
```
You can use `YSTDLIB_ERROR_HANDLING_TRYX` for results containing a value and `YSTDLIB_ERROR_HANDLING_TRYV` for void results as nice shorthands for forwarding an error if it exists, and unwrapping the result otherwise.

With this diff you can then use `optimal_parameters` directly, since it represents the unwrapped value, instead of `optimal_parameters.value()`.

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,192 @@
...
+    auto const bit_array_size
+            = std::max<size_t>(1, static_cast<size_t>(std::ceil(ideal_bit_array_size)));
+
+    auto const num_hash_functions = static_cast<uint32_t>(
+            static_cast<double>(bit_array_size) / static_cast<double>(expected_num_elements) * ln2
+    );
+
+    uint32_t const capped_num_hash_functions
+            = std::clamp(num_hash_functions, cMinNumHashFunctions, cMaxNumHashFunctions);
```
gibber9809:
```suggestion
    uint32_t const capped_num_hash_functions{
            std::clamp(num_hash_functions, cMinNumHashFunctions, cMaxNumHashFunctions)};
```
Nit: we require `{}` initialization with the make exception being `auto const xyz =`

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,192 @@
...
+}
+
+auto BloomFilter::test_bit(size_t bit_index) const -> bool {
+    size_t const byte_index{bit_index / cNumBitsInByte};
+    size_t const bit_offset{bit_index % cNumBitsInByte};
+    return (m_bit_array.at(byte_index) & static_cast<uint8_t>(1U << bit_offset)) != 0;
+}
+
+auto BloomFilter::write_to_file(clp::WriterInterface& writer) const -> void {
+    writer.write_numeric_value<uint32_t>(m_num_hash_functions);
```
gibber9809, line 132: Sorry I didn't think about this earlier, but do you think it makes sense to include an extra byte in this format indicating what hash algorithm was used for this bloom filter (e.g., by defining an enum for hash functions, with 0x00 representing xxhash)? That way if we ever change the hash algorithm in the future it's easy to keep things backwards compatible.

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,192 @@
...
+        if (clp::ErrorCode_Success
+            != reader.try_read_exact_length(
+                    clp::size_checked_pointer_cast<char>(bit_array.data()),
+                    static_cast<size_t>(bit_array_bytes)
+            ))
+        {
+            return ErrorCode{ErrorCodeEnum::ReadFailure};
+        }
+    }
+    return BloomFilter(bit_array_size, num_hash_functions, std::move(bit_array));
```
gibber9809:
```suggestion
    return BloomFilter{bit_array_size, num_hash_functions, std::move(bit_array)};
```

components/core/src/clp_s/filter/BloomFilter.hpp
```diff
@@ -0,0 +1,83 @@
...
+    auto add(std::string_view value) -> void;
+
+    /**
+     * @param value
+     * @return true if the value may be present, false if definitely not present.
+     */
+    [[nodiscard]] auto possibly_contains(std::string_view value) const -> bool;
+
+    /**
+     * Writes Bloom filter payload fields to a writer.
```
gibber9809:
```suggestion
     * Writes a serialized representation of a bloom filter to a writer.
```
I think this wording is slightly clearer.

components/core/src/clp_s/filter/BloomFilter.hpp
```diff
@@ -0,0 +1,83 @@
...
+     * - ErrorCodeEnum::InvalidFalsePositiveRate if the false-positive rate is not in [1e-6, 1).
+     * - ErrorCodeEnum::ParameterComputationOutOfRange if parameter computation overflows.
+     */
+    [[nodiscard]] static auto create(size_t expected_num_elements, double false_positive_rate)
+            -> ystdlib::error_handling::Result<BloomFilter>;
+
+    /**
+     * Reads Bloom filter payload fields from a reader.
+     * @param reader
+     * @return A result containing a parsed BloomFilter on success, or an error code indicating
```
gibber9809:
```suggestion
     * Reads a serialized representation of a bloom filter from a reader.
     * @param reader
     * @return A result containing an initialized BloomFilter on success, or an error code indicating
```
I think this wording is slightly clearer.

taskfiles/deps/main.yaml
```diff
@@ -558,6 +559,23 @@ tasks:
...
+    internal: true
+    vars:
+      LIB_NAME: "xxHash"
+    run: "once"
+    cmds:
+      - task: "utils:install-remote-cmake-lib"
+        vars:
+          CMAKE_GEN_ARGS:
+            - "-DXXHASH_BUILD_XXHSUM=OFF"
+            - "-DBUILD_SHARED_LIBS=ON"
```
gibber9809:
```suggestion
            - "-DBUILD_SHARED_LIBS=OFF"
```
After further thought I think it's just simpler to only build + link the static version of xxhash. Main justification being that if we wanted to build a static + shared version I think we'd need to build them as separate install targets in separate directories + write a `Findxyz.cmake` script like we do for LZMA, which adds some maintenance burden and makes it harder to support things like using an xxhash library installed at the system level.

components/core/src/clp_s/filter/BloomFilter.cpp
ShangDanLuXian: This function writes payload of a bloom filter. There will be other metadata bytes written in the FilterFile (which does not belong to this PR), including the type of filter.
My current implementation (out of scope of this PR again) treats bloom filters using different hash algorithms as different "filter type", unless we have a bunch of hash algorithms to use then it will make a lot of combinations. I feel it would be not very much difference comparing to your suggestion.

But yeah definitely I need to integrate a mapper for hash algorithms in bloomfilter.cpp , do you recommand add in this PR or next

ShangDanLuXian: Thanks for the suggestion — after thinking it through again, I agree to add a hash algorithm ID in the BloomFilter payload - it  makes mapping easier than my previous approach.

I’ve already implemented a small `HashAlgorithm` mapper (ID parsing + `hash64` dispatch), plus the corresponding BloomFilter.cpp/hpp updates. In total it’s ~80 LOC, but most of that is headers/guards/includes/docstrings; the core logic is quite small and self-contained.

I haven’t pushed it yet to keep scope tight, but I can include it in this PR if you’re okay with that.

Please let me know and I’ll push it. :)

gibber9809 (dismissed):
LGTM! Thanks for going through all the revisions.

Before we merge, could we:
1. Change the PR title to:
```
feat(clp_s::filter): Add library for representing bloom filters; Add xxHash v0.8.3 as a dependency.
```
2. Update the PR description to reference that this is one of the PRs for #2090 -- I can see that you manually reference this PR in that issue, but referencing the issue from the PR makes github automatically create a link to the PR from the issue, which makes it easy to track related PRs.

taskfiles/deps/main.yaml
```diff
@@ -558,6 +559,23 @@ tasks:
         CLP_UTFCPP_INCLUDE_DIRECTORY \"{{.UTFCPP_OUTPUT_DIR}}\"
         )" > "{{.G_DEPS_CPP_CMAKE_SETTINGS_DIR}}/{{.LIB_NAME}}.cmake"
 
+  xxhash:
```
LinZhihao-723, line 562: A few things missing for adding a new dep:
- We need to update the clang-format config to include `xxhash` as an external library header. Check [here](https://github.com/y-scope/clp/blob/b27425abcbc899776a05d1d1b7f8e2d6734ddcae/components/core/.clang-format#L8).
- We need to update the table in our dev-doc [here](https://github.com/y-scope/clp/blob/b27425abcbc899776a05d1d1b7f8e2d6734ddcae/docs/src/dev-docs/components-core/index.md?plain=1#L36).

components/core/src/clp_s/filter/HashAlgorithm.hpp
```diff
@@ -0,0 +1,32 @@
...
+namespace clp_s::filter {
+/**
+ * Hash algorithms supported by the filter serialization format.
+ */
+enum class HashAlgorithm : uint8_t {
+    Xxh364 = 0x00,
+};
+
+/**
+ * @param hash_algorithm Encoded algorithm id from serialized payload.
```
LinZhihao-723:
```suggestion
 * @param hash_algorithm Encoded algorithm ID from serialized payload.
```

components/core/src/clp_s/filter/HashAlgorithm.cpp
```diff
@@ -0,0 +1,24 @@
...
+        return HashAlgorithm::Xxh364;
+    }
+    return std::nullopt;
+}
+
+auto hash64(HashAlgorithm hash_algorithm, std::string_view value, uint64_t seed) -> uint64_t {
+    if (HashAlgorithm::Xxh364 == hash_algorithm) {
+        return xxhash::hash64(value, seed);
+    }
+
```
LinZhihao-723:
```suggestion
    switch (hash_algorithm) {
        case HashAlgorithm::Xxh364:
            return xxhash::hash64(value, seed);
        default:
            break;
    }
```
Since we're using enum, I think it's cleaner if we use a switch statement.
In terms of the default case, I think we should have sth equivalent to [`unreachable`](https://doc.rust-lang.org/std/macro.unreachable.html) in Rust. But this is out of the scope for this PR, let's just keep it as is.

components/core/src/clp_s/filter/ErrorCode.cpp
```diff
@@ -0,0 +1,32 @@
...
+        case ErrorCodeEnum::ParameterComputationOutOfRange:
+            return "Bloom filter parameter computation overflowed or produced invalid values.";
+        case ErrorCodeEnum::UnsupportedHashAlgorithm:
+            return "Bloom filter hash algorithm is unsupported.";
+        case ErrorCodeEnum::CorruptFilterPayload:
+            return "Bloom filter payload is malformed or inconsistent.";
+        case ErrorCodeEnum::ReadFailure:
+            return "Failed to read Bloom filter payload from reader.";
+    }
+    return "Unknown error code enum";
```
LinZhihao-723:
```suggestion
    switch (error_enum) {
        case ErrorCodeEnum::InvalidFalsePositiveRate:
            return "false-positive rate must be in the range [1e-6, 1)";
        case ErrorCodeEnum::ParameterComputationOutOfRange:
            return "bloom filter parameter computation overflowed or produced invalid values";
        case ErrorCodeEnum::UnsupportedHashAlgorithm:
            return "bloom filter hash algorithm is unsupported";
        case ErrorCodeEnum::CorruptFilterPayload:
            return "bloom filter payload is malformed or inconsistent";
        case ErrorCodeEnum::ReadFailure:
            return "failed to read Bloom filter payload from reader";
    }
    return "unknown error code enum";
```
Let's:
- Don't capitalize the error message
- Don't add the period

This is because the error enum is usually added to a format string when we log the error. We should make it more like an error string and ask the log statement to decide how to format it.

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,206 @@
...
+constexpr uint32_t cDefaultNumHashFunctions{1};
+constexpr uint32_t cMinNumHashFunctions{1};
+constexpr uint32_t cMaxNumHashFunctions{20};
+constexpr double cMinFalsePositiveRate{1e-6};
+constexpr HashAlgorithm cDefaultHashAlgorithm{HashAlgorithm::Xxh364};
+constexpr uint64_t cPrimaryHashSeed{0};
+constexpr uint64_t cSecondaryHashSeed{0x9e37'79b9'7f4a'7c15ULL};
+constexpr size_t cNumBitsInByte{8};
+
+auto min_bytes_containing_bits(size_t num_bits) -> size_t {
```
LinZhihao-723:
```suggestion
[[nodiscard]] auto min_bytes_containing_bits(size_t num_bits) -> size_t {
```

components/core/src/clp_s/filter/BloomFilter.hpp
```diff
@@ -0,0 +1,88 @@
...
+    BloomFilter(
+            size_t bit_array_size,
+            uint32_t num_hash_functions,
+            HashAlgorithm hash_algorithm,
+            ystdlib::containers::Array<uint8_t> bit_array
+    );
+
+    [[nodiscard]] static auto
+    compute_optimal_parameters(size_t expected_num_elements, double false_positive_rate)
+            -> ystdlib::error_handling::Result<std::pair<size_t, uint32_t>>;
```
LinZhihao-723: According to our [guideline](https://docs.yscope.com/dev-guide/contrib-guides-cpp.html#declaration-order):
- Static methods should come before constructors.
- The definition should match the declaration order.

These should be quick fixes, can you check the source file as well to make sure they're properly ordered?

LinZhihao-723: Since this function doesn't explain what the returned parameters are in its name, and also it may return errors, I think a docstring is necessary and cannot be omitted. I've drafted one for u, feel free to apply or adjust if it's inappropriate.

```suggestion
    /**
     * Computes the optimal Bloom filter parameters from the given inputs.
     * @param expected_num_elements
     * @param false_positive_rate
     * @return A result containing a pair of Bloom filter parameters on success, or an error code
     * indicating the failure:
     * - The pair:
     *  - The number of bytes needed for the Bloom filter's bit array.
     *  - The number of hash functions to apply when mapping an element to positions in the bit
     *    array.
     * - Possible error codes:
     *   - ErrorCodeEnum::InvalidFalsePositiveRate if `false_positive_rate` is out of the supported
     *     range [1e-6, 1).
     *   - ErrorCodeEnum::ParameterComputationOutOfRange if the computed parameters overflow.
     */
    [[nodiscard]] static auto
    compute_optimal_parameters(size_t expected_num_elements, double false_positive_rate)
            -> ystdlib::error_handling::Result<std::pair<size_t, uint32_t>>;
```

components/core/src/clp_s/filter/BloomFilter.hpp
```diff
@@ -0,0 +1,88 @@
...
+class BloomFilter {
+public:
+    // Factory functions
+    /**
+     * @param expected_num_elements Expected number of inserted values.
+     * @param false_positive_rate Target false-positive rate in the range [1e-6, 1).
+     * @return A result containing a constructed BloomFilter on success, or an error code
+     * indicating the failure:
+     * - ErrorCodeEnum::InvalidFalsePositiveRate if false_positive_rate is not in [1e-6, 1).
+     * - ErrorCodeEnum::ParameterComputationOutOfRange if parameter computation overflows.
```
LinZhihao-723: We don't directly return these errors, but forward from another function. In this case, we should document it as:
```suggestion
     * - Forwards `compute_optimal_parameters`'s return values on failure.
```

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,206 @@
...
+          m_num_hash_functions{num_hash_functions},
+          m_hash_algorithm{hash_algorithm},
+          m_bit_array{std::move(bit_array)} {}
+
+auto BloomFilter::create(size_t expected_num_elements, double false_positive_rate)
+        -> ystdlib::error_handling::Result<BloomFilter> {
+    auto const optimal_parameters{YSTDLIB_ERROR_HANDLING_TRYX(
+            compute_optimal_parameters(expected_num_elements, false_positive_rate)
+    )};
+    auto const [bit_array_size, num_hash_functions]{optimal_parameters};
```
LinZhihao-723: fyi u can do:

```suggestion
    auto const [bit_array_size, num_hash_functions] = YSTDLIB_ERROR_HANDLING_TRYX(
            compute_optimal_parameters(expected_num_elements, false_positive_rate)
    );
```

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+    return std::make_pair(bit_array_size, capped_num_hash_functions);
+}
+
+auto BloomFilter::add(std::string_view value) -> void {
+    uint64_t const h1{hash64(m_hash_algorithm, value, cPrimaryHashSeed)};
+    uint64_t h2{hash64(m_hash_algorithm, value, cSecondaryHashSeed)};
+    if (0 == h2) {
+        h2 = 1;
+    }
```
LinZhihao-723:
```suggestion
    auto const h2{std::max<uint64_t>(hash64(m_hash_algorithm, value, cSecondaryHashSeed), 1ULL)};
```
Might be a cleaner way to achieve this logic

components/core/src/clp_s/filter/BloomFilter.cpp
```diff
@@ -0,0 +1,206 @@
...
+        set_bit(static_cast<size_t>((h1 + i * h2) % m_bit_array_size));
+    }
+}
+
+auto BloomFilter::possibly_contains(std::string_view value) const -> bool {
+    uint64_t const h1{hash64(m_hash_algorithm, value, cPrimaryHashSeed)};
+    uint64_t h2{hash64(m_hash_algorithm, value, cSecondaryHashSeed)};
+    if (0 == h2) {
+        h2 = 1;
+    }
```
LinZhihao-723:
```suggestion
    auto const h2{std::max<uint64_t>(hash64(m_hash_algorithm, value, cSecondaryHashSeed), 1ULL)};
```
Similarly

components/core/src/clp_s/filter/tests/test-clp_s-xxhash.cpp
```diff
@@ -0,0 +1,38 @@
...
+            {.input = "CLP", .seed = 0ULL, .expected_hash = 0x70da'c239'055f'38a5ULL},
+            {.input = "The quick brown fox jumps over the lazy dog",
+             .seed = 42ULL,
+             .expected_hash = 0xb4a3'f3c3'6b3c'7d26ULL},
+    }};
+
+    for (auto const& [input, seed, expected_hash] : test_cases) {
+        INFO("input=" << input << ", seed=" << seed);
+        REQUIRE(clp_s::filter::xxhash::hash64(input, seed) == expected_hash);
+    }
```
LinZhihao-723: Here's a more maintainable way to structure the test cases:

```suggestion
    using TestCase = std::tuple<std::string_view, uint64_t, uint64_t>;

    auto const [input, seed, expected_hash] = GENERATE(
            TestCase{"", 0ULL, 0x2d06'8005'38d3'94c2ULL},
            TestCase{"", 1ULL, 0x4dc5'b0cc'826f'6703ULL},
            TestCase{"a", 0ULL, 0xe6c6'32b6'1e96'4e1fULL},
            TestCase{"a", 1ULL, 0xd2f6'd099'6f37'a720ULL},
            TestCase{"hello", 0ULL, 0x9555'e855'5c62'dcfdULL},
            TestCase{"hello", 42ULL, 0xbafa'072f'07db'7937ULL},
            TestCase{"CLP", 0ULL, 0x70da'c239'055f'38a5ULL},
            TestCase{"The quick brown fox jumps over the lazy dog", 42ULL, 0xb4a3'f3c3'6b3c'7d26ULL}
    );

    CAPTURE(input, seed, expected_hash);
    REQUIRE(clp_s::filter::xxhash::hash64(input, seed) == expected_hash);
```

On error (and only on error), `CAPTURE` will print:
```
  REQUIRE(clp_s::filter::xxhash::hash64(input, seed) == expected_hash)
with expansion:
  8132025627329902757 (0x70dac239055f38a5)
==
8132025627329902756 (0x70dac239055f38a4)
with message:
  input := "CLP"
  seed := 0
  expected_hash := 8132025627329902756 (0x70dac239055f38a4)
```

Don't forget to drop `<array>` and include `<tuple>`

components/core/src/clp_s/filter/tests/test-clp_s-bloom_filter.cpp
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    auto const false_positive_rate
+            = static_cast<double>(false_positives) / static_cast<double>(cQueries);
+    double const sigma{std::sqrt(
+            cFalsePositiveRate * (1.0 - cFalsePositiveRate) / static_cast<double>(cQueries)
+    )};
+    double const allowed_false_positive_rate{cFalsePositiveRate + (3.0 * sigma)};
+    INFO("False positive rate: " << false_positive_rate);
+    INFO("Allowed max FPR: " << allowed_false_positive_rate);
```
LinZhihao-723: Similarly:
```suggestion
    CAPTURE(false_positive_rate, allowed_false_positive_rate);
```

components/core/src/clp_s/filter/BloomFilter.hpp
LinZhihao-723, line 1: Just curious: what's the real size of the bitmap in practice? Is it going to be mostly the default (64 * 8)?
If the size can become big, shall we try/benchmark [this](https://github.com/RoaringBitmap/RoaringBitmap) compressed bitmap? It also supports serde so writing to or reading from a file shouldn't be a problem.

ShangDanLuXian: I ran a simple benchmark comparing a RoaringBitmap-backed Bloom filter against the current conventional bitmap implementation. For this workload, RoaringBitmap used slightly more space due to its container/metadata overhead, and it was also slower for both build and query.

The main reason is that Bloom filters are configured near their optimal parameters, which gives a bit density around 50%. Combined with the fact that the set bits are distributed by xxhash, the bitmap is relatively dense and random-looking, with very little local structure or long runs. That makes it a poor fit for RoaringBitmap-style compression.

I think this matches the case described in the RoaringBitmap README where a conventional bitset is preferred:

> You have dense set of random values that never form runs of continuous values. For example, consider the set 0,2,4,...,10000. If this is typical of your application, you might be better served with a conventional bitset (e.g., Java's BitSet class).

ShangDanLuXian:
> A bit more style-wise catches and tips/suggestions. Otherwise, we should be good to go.

Thanks for your review!
I split the updates into 2 commits for easier review. You can look at the first commit for the style/doc changes only, and then the second commit is just the BloomFilter.hpp/.cpp reordering.

Edit:
Also made one small follow-up wording fix for consistency: changed hash algorithm id to hash algorithm ID in the BloomFilter::try_read_from_file docstring to match the wording in HashAlgorithm.hpp

components/core/src/clp_s/filter/HashAlgorithm.cpp
ShangDanLuXian: Thanks — I also applied the same switch style to `try_parse_hash_algorithm` for consistency. However, the `case static_cast<uint8_t>(HashAlgorithm::Xxh364)` looks a little awkward to me. Please let me know if you’d prefer a different style there.

components/core/src/clp_s/filter/BloomFilter.hpp
LinZhihao-723: Make sense. Thanks for your investigation!

components/core/src/clp_s/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 add_subdirectory(ffi)
 add_subdirectory(indexer)
 add_subdirectory(log_converter)
+add_subdirectory(filter)
```
LinZhihao-723:
```suggestion
add_subdirectory(ffi)
add_subdirectory(filter)
add_subdirectory(indexer)
add_subdirectory(log_converter)
```
ops, missed in the last round

components/core/src/clp_s/filter/HashAlgorithm.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "XxHash.hpp"
+
+namespace clp_s::filter {
+auto try_parse_hash_algorithm(uint8_t hash_algorithm) -> std::optional<HashAlgorithm> {
+    switch (hash_algorithm) {
+        case static_cast<uint8_t>(HashAlgorithm::Xxh364):
+            return HashAlgorithm::Xxh364;
+        default:
+            return std::nullopt;
+    }
```
LinZhihao-723: As a reply to [this](https://github.com/y-scope/clp/pull/2010#discussion_r2938606816) comment, I don't think we need to use switch here (as u suggested, it just looks weird, lol). Essentially this is because we're working on the underlying integer of the enum value instead of the enum value itself. I think it doesn't make too much sense if we're not working directly on the enum type when using switch.
I'm fine if u'd like to rollback or just leave it as is.

PR #2013, opened 2026-02-20
perf(clp-s::timestamp_parser): Improve marshalling speed for padded integer fields (resolves #1968).

### LINKED ISSUE #1968: perf(clp-s::timestamp_parser): Improve marshalling speed for format specifiers representing padded integer fields.
### Request

Currently `clp_s::timestamp_parser` is slower at marshalling certain kinds of timestamp formats than the older `clp_s::TimestampPattern`, which can lead to slower overall decompression speeds as shown in #1788.

After some profiling, this slowdown seems to mostly come from calls to `fmt::format` for marshalling zero and space-padded integer fields (judging by the profiling results, most of the time is spent dynamically parsing format strings).

### Possible implementation

Replace all or the majority of the `fmt::format` calls in the marshalling path for `clp_s::timestamp_parser` with calls to an optimized function which appends padded integers to a buffer, similar to what we do in the old `clp_s::TimestampPattern` marshalling code.

gibber9809:
# Description

This PR significantly improves decompression speed for padded integer fields in timestamps, leading to much higher decompression speeds overall. As identified in #1968 most of the overhead for timestamp marshalling was coming from `fmt::format()` calls, which appeared to be dynamically parsing the format strings on each call.

Decompression speedup for open source datasets compared to decompression speed without this optimization:

| dataset | speedup % |
| ------ | ------ |
| mongodb | +23.14% |
| cockroachdb | +3.75% |
| elasticsearch | +16.81% |

# Checklist

  breaking change.

# Validation performed
* Validated that unit tests still pass
* Benchmarked improvement in decompression speed

## Summary by CodeRabbit

* **Refactor**
  * Centralized numeric padding and validation for timestamp components for consistent formatting.
  * Consolidated date/time formatting paths to standardize fractional seconds and timezone output.

* **Bug Fixes**
  * More reliable error propagation for invalid date/time values, reducing risk of malformed timestamps.
  * Improved handling of padding-related edge cases to prevent incorrect timestamp fields.

### DIFF
--- a/components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
+++ b/components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
@@ -955,7 +955,7 @@ auto append_positive_left_padded_integer(
             value_str_buffer.data() + value_str_buffer.size(),
             value
     );
-    if (auto err{std::make_error_code(ec)}; err) {
+    if (auto const err{std::make_error_code(ec)}; err) {
         return err;
     }
 


components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -294,6 +296,23 @@ find_first_matching_prefix(std::string_view str, std::span<std::string_view cons
...
+ * Appends a positive integer to a buffer, left-padding it to a minimum length with a padding
+ * character.
+ * @param buffer
+ * @param length The minimum length that the integer will be padded to.
+ * @param value
+ * @param padding_character
+ * @return A void result on success, or an error code indicating the failure:
+ * - ErrorCodeEnum::InvalidDate if `value` is negative.
+ */
+auto append_positive_left_padded_integer(
```
```suggestion
[[nodiscard]] auto append_positive_left_padded_integer(
```

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -294,6 +296,23 @@ find_first_matching_prefix(std::string_view str, std::span<std::string_view cons
...
+        std::string& buffer,
+        size_t length,
+        int value,
+        char padding_character
```
LinZhihao-723: How about:
```C++
        int value,
        size_t min_padded_length,
        char padding_character,
        std::string& buffer
```
- `length` -> `min_padded_length` to be more explicit about what this length is.
- Put `buffer` to the end since it's an output param.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -857,6 +932,25 @@ auto is_repeatable_escape_sequence(char c) -> bool {
...
+    if (value < 0) {
+        return ErrorCode{ErrorCodeEnum::InvalidDate};
+    }
+
+    auto value_str{std::to_string(value)};
+    size_t const num_padding_characters{
+            value_str.length() >= length ? size_t{0} : length - value_str.length()
+    };
+    buffer.append(num_padding_characters, padding_character);
+    buffer.append(value_str);
```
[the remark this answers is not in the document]
LinZhihao-723: I'm actually curious whether this would make a difference. Since you probably already have the archives set up, shall we try it and see if it's actually better?
```C++
    if (value < 0) {
        return ErrorCode{ErrorCodeEnum::InvalidDate};
    }

    std::array<char, std::numeric_limits<typeof(value)>::digits10 + 1> value_str_buffer{};
    auto const [end_ptr, ec] = std::to_chars(
            value_str_buffer.data(),
            value_str_buffer.data() + value_str_buffer.size(),
            value
    );
    if (auto err{std::make_error_code(ec)}; err) {
        return err;
    }

    std::string_view const value_str{value_str_buffer.data(), end_ptr};
    size_t const num_padding_characters{
            value_str.length() >= length ? size_t{0} : length - value_str.length()
    };
    buffer.append(num_padding_characters, padding_character);
    buffer.append(value_str);
    return ystdlib::error_handling::success();
```

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -857,6 +932,25 @@ auto is_repeatable_escape_sequence(char c) -> bool {
...
+        std::string& buffer,
+        size_t length,
+        int value,
+        char padding_character
+) -> ystdlib::error_handling::Result<void> {
+    if (value < 0) {
+        return ErrorCode{ErrorCodeEnum::InvalidDate};
+    }
+
+    auto value_str{std::to_string(value)};
```
```suggestion
    auto const value_str{std::to_string(value)};
```

LinZhihao-723 (changes requested):

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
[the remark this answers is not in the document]
gibber9809: Sure I can test it quickly -- I was being a bit lazy and relying on the small string optimization since most of the padded fields are short.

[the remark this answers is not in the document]
gibber9809: It ends up helping a bit (~3%) for some datasets, with no real difference for others. Will include this change.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -294,6 +298,23 @@ find_first_matching_prefix(std::string_view str, std::span<std::string_view cons
...
 
+/**
+ * Appends a positive integer to a buffer, left-padding it to a minimum length with a padding
+ * character.
+ * @param value
+ * @param min_padded_length The minimum length that the integer will be padded to.
+ * @param padding_character
+ * @param buffer
+ * @return A void result on success, or an error code indicating the failure:
+ * - ErrorCodeEnum::InvalidDate if `value` is negative.
```
LinZhihao-723, line 309:
```suggestion
 * - ErrorCodeEnum::InvalidDate if `value` is negative.
 * - Forwards `std::to_chars`' return values on failure.
```

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -857,6 +937,36 @@ auto is_repeatable_escape_sequence(char c) -> bool {
...
+        return ErrorCode{ErrorCodeEnum::InvalidDate};
+    }
+
+    std::array<char, std::numeric_limits<typeof(value)>::digits10 + 1> value_str_buffer{};
+    auto const [end_ptr, ec] = std::to_chars(
+            value_str_buffer.data(),
+            value_str_buffer.data() + value_str_buffer.size(),
+            value
+    );
+    if (auto err{std::make_error_code(ec)}; err) {
```
```suggestion
    if (auto const err{std::make_error_code(ec)}; err) {
```

PR #2014, opened 2026-02-22
feat(log-ingestor): Add DB operations for ingestion job orchestration:

LinZhihao-723:
* Add SQL statements to create necessary tables for log-ingestor status tracking in CLP DB.
* Add support for registering new ingestion jobs in CLP DB.
* Add support for ingesting S3 object metadata into CLP DB through implementing ingestion job state traits.

# Description

This PR implements DB operations for:

* Ingestion job table creation (three tables will be created as described in #1978).
* Ingestion job creation.
* S3 object metadata ingestion.

At the top level, we introduce `ClpDbIngestionConnector` for job creation. It returns an ingestion job state implementation, `ClpIngestionState`, which is a per-job state that enables ingesting S3 object metadata into the CLP DB.

The ingestion operation uses transactions to update both the object metadata table and the ingestion job table for consistency.

# Checklist

  breaking change.

# Validation performed

* [x] Run the following tests.
  * Step 1: Run [this](https://github.com/y-scope/spider/blob/main/tools/scripts/mariadb/start.py) script to start a mariaDB instance on localhost:
    ```shell
    ./start.py --name "clp-db-test" --port 3306 --database "clp-db" --username "ROOT" --password "ROOT"
    ```
  * Step 2: Add the following code to the test section in `clp_ingestion.rs`:
    ```Rust
    #[tokio::test]
    async fn test_table_creation() {
        use clp_rust_utils::{
            job_config::ingestion::s3::S3ScannerConfig,
            types::non_empty_string::ExpectedNonEmpty,
        };
        use serde_json::json;

        let mysql_options = sqlx::mysql::MySqlConnectOptions::new()
            .host("localhost")
            .port(3306)
            .database("clp-db")
            .username("ROOT")
            .password("ROOT");

        let db_pool = sqlx::mysql::MySqlPoolOptions::new()
            .max_connections(100)
            .connect_with(mysql_options)
            .await
            .expect("Failed to connect to mysql");

        sqlx::query(
            r"
                CREATE TABLE IF NOT EXISTS `compression_jobs` (
                    `id` INT NOT NULL AUTO_INCREMENT,
                    `status` INT NOT NULL DEFAULT '0',
                    `status_msg` VARCHAR(512) NOT NULL DEFAULT '',
                    `creation_time` DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
                    `start_time` DATETIME(3) NULL DEFAULT NULL,
                    `update_time` DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP(),
                    `duration` FLOAT NULL DEFAULT NULL,
                    `original_size` BIGINT NOT NULL DEFAULT '0',
                    `uncompressed_size` BIGINT NOT NULL DEFAULT '0',
                    `compressed_size` BIGINT NOT NULL DEFAULT '0',
                    `num_tasks` INT NOT NULL DEFAULT '0',
                    `num_tasks_completed` INT NOT NULL DEFAULT '0',
                    `clp_binary_version` INT NULL DEFAULT NULL,
                    `clp_config` VARBINARY(60000) NOT NULL,
                    PRIMARY KEY (`id`) USING BTREE,
                    INDEX `JOB_STATUS` (`status`) USING BTREE,
                    INDEX `JOB_UPDATE_TIME` (`update_time`) USING BTREE
                ) ROW_FORMAT=DYNAMIC
        ",
        )
            .execute(&db_pool)
            .await
            .expect("Failed to create compression jobs table");

        let connector = match ClpDbIngestionConnector::new(db_pool).await {
            Ok(c) => c,
            Err(e) => {
                panic!("Failed to create CLP DB connector: {e}");
            }
        };

        let json_config = json!(
            {
                "bucket_name": "test",
                "key_prefix": "test-",
                "region": "us-east-2",
            }
        );

        let config: S3ScannerConfig = serde_json::from_str(json_config.to_string().as_str())
            .expect("Failed to deserialize S3 ingestion job config");

        let state = match connector
            .create_ingestion_job(S3IngestionJobConfig::S3Scanner(config))
            .await
        {
            Ok(s) => s,
            Err(e) => {
                panic!("Failed to create ingestion job: {e}");
            }
        };

        state.start().await.expect("Failed to start ingestion job");

        let mut objects = Vec::new();
        for i in 0..15000 {
            objects.push(ObjectMetadata {
                bucket: NonEmptyString::from_static_str("test"),
                key: NonEmptyString::from_string(format!("test-{i}")),
                size: 100 + i,
            });
        }

        let last_ingested_key = objects.last().unwrap().key.clone();
        S3ScannerState::ingest(&state, objects, last_ingested_key.as_str())
            .await
            .expect("Failed to ingest");

        state.end().await.expect("Failed to end ingestion job");

        let objects = vec![ObjectMetadata {
            bucket: NonEmptyString::from_static_str("test"),
            key: NonEmptyString::from_static_str("test-10000"),
            size: 8,
        }];
        let last_ingested_key = objects.last().unwrap().key.clone();
        assert!(
            S3ScannerState::ingest(&state, objects, last_ingested_key.as_str())
                .await
                .is_err()
        );
    }
    ```
  * Step 3: Run the tester (inside the IDE or through the terminal) and make sure:
    * The DB contains the expected tables with the correct schema.
    * One ingestion job is created.
    * 37 files are ingested into the created ingestion job (on both the ingestion job table and S3 object metadata table).
    * The ingestion job ends with a `finished` state.
    * All above indicates the last file ingestion (`test-10000`) didn't go through since the job has finished.

## Summary by CodeRabbit

* **New Features**
  * Enhanced S3 ingestion configuration to support multiple ingestion strategies.
  * Added database-backed ingestion job management for persistent job state, lifecycle control, transactional metadata ingest and status tracking.

* **Chores**
  * Updated build dependencies to enable additional derive/macros support.

### DIFF
--- a/components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
+++ b/components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
@@ -1,8 +1,11 @@
 use async_trait::async_trait;
-use clp_rust_utils::{job_config::ingestion::s3::S3IngestionJobConfig, s3::ObjectMetadata};
+use clp_rust_utils::{
+    database::mysql::MySqlEnumFormat,
+    job_config::ingestion::s3::S3IngestionJobConfig,
+    s3::ObjectMetadata,
+};
 use const_format::formatcp;
 use sqlx::MySqlPool;
-use strum::IntoEnumIterator;
 use strum_macros::{AsRefStr, Display, EnumIter, EnumString};
 
 use crate::{
@@ -170,6 +173,7 @@ impl ClpIngestionState {
             r"INSERT INTO `{table}` (`bucket`, `key`, `size`, `ingestion_job_id`) VALUES ",
             table = INGESTED_S3_OBJECT_METADATA_TABLE_NAME,
         );
+
         assert!(
             !objects.is_empty(),
             "Cannot build S3 object metadata ingestion query with empty objects"
@@ -303,11 +307,23 @@ impl S3ScannerState for ClpIngestionState {
         );
 
         let mut tx = self.db_pool.begin().await?;
-        sqlx::query(UPDATE_S3_SCANNER_STATE_QUERY)
+
+        if sqlx::query(UPDATE_S3_SCANNER_STATE_QUERY)
             .bind(last_ingested_key)
             .bind(self.job_id)
             .execute(&mut *tx)
-            .await?;
+            .await?
+            .rows_affected()
+            == 0
+        {
+            const ERROR_MSG: &str = "Failed to update last ingested key for S3 scanner state.";
+            tracing::error!(
+                job_id = ? self.job_id,
+                last_ingested_key = ? last_ingested_key,
+                ERROR_MSG
+            );
+            return Err(anyhow::anyhow!(ERROR_MSG));
+        }
 
         self.ingest_s3_object_metadata(&mut tx, &objects)
             .await
@@ -429,23 +445,6 @@ const INGESTION_JOB_TABLE_NAME: &str = "ingestion_job";
 const INGESTION_JOB_S3_SCANNER_STATE_TABLE_NAME: &str = "ingestion_job_s3_scanner_state";
 const INGESTED_S3_OBJECT_METADATA_TABLE_NAME: &str = "ingested_s3_object_metadata";
 
-/// A trait for formatting Rust enums as SQL `ENUM(...)` declarations.
-trait MySqlEnumFormat: IntoEnumIterator + Sized + ToString
-where
-    Self::Iterator: Iterator<Item = Self>, {
-    /// # Returns
-    ///
-    /// A string representing the SQL enum definition for this enum.
-    #[must_use]
-    fn format_as_sql_enum() -> String {
-        let inner = Self::iter()
-            .map(|v| format!("'{}'", v.to_string()))
-            .collect::<Vec<_>>()
-            .join(", ");
-        format!("ENUM({inner})")
-    }
-}
-
 /// The query to create the table for CLP ingestion jobs.
 #[must_use]
 fn ingestion_job_table_creation_query() -> String {
@@ -453,7 +452,7 @@ fn ingestion_job_table_creation_query() -> String {
         r"
 CREATE TABLE IF NOT EXISTS `{table}` (
     `id` BIGINT unsigned NOT NULL AUTO_INCREMENT,
-    `config` VARCHAR(1024) NOT NULL,
+    `config` TEXT NOT NULL,
     `status` {status_enum} NOT NULL DEFAULT '{default_status}',
     `num_files_ingested` BIGINT unsigned NOT NULL DEFAULT '0',
     `num_files_compressed` BIGINT unsigned NOT NULL DEFAULT '0',
@@ -462,7 +461,7 @@ CREATE TABLE IF NOT EXISTS `{table}` (
     `last_update_ts` DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3)
         ON UPDATE CURRENT_TIMESTAMP(3),
     PRIMARY KEY (`id`)
-) ROW_FORMAT=DYNAMIC;",
+);",
         table = INGESTION_JOB_TABLE_NAME,
         status_enum = ClpIngestionJobStatus::format_as_sql_enum(),
         default_status = ClpIngestionJobStatus::Requested,
@@ -481,11 +480,12 @@ CREATE TABLE IF NOT EXISTS `{INGESTION_JOB_S3_SCANNER_STATE_TABLE_NAME}` (
     CONSTRAINT `{INGESTION_JOB_S3_SCANNER_STATE_TABLE_NAME}_ingestion_job_id_ref`
         FOREIGN KEY (`id`) REFERENCES `{INGESTION_JOB_TABLE_NAME}` (`id`)
         ON DELETE RESTRICT ON UPDATE RESTRICT
-) ROW_FORMAT=DYNAMIC;",
+);",
     )
 }
 
 /// The query to create the table for ingested S3 object metadata.
+#[must_use]
 fn ingested_s3_object_metadata_table_creation_query() -> String {
     format!(
         r"
@@ -506,7 +506,7 @@ CREATE TABLE IF NOT EXISTS `{table}` (
     CONSTRAINT `{table}_compression_job_id_ref`
         FOREIGN KEY (`compression_job_id`) REFERENCES `{compression_job_table}` (`id`)
         ON DELETE RESTRICT ON UPDATE RESTRICT
-) ROW_FORMAT=DYNAMIC;",
+);",
         table = INGESTED_S3_OBJECT_METADATA_TABLE_NAME,
         job_table = INGESTION_JOB_TABLE_NAME,
         status_enum = IngestedS3ObjectMetadataStatus::format_as_sql_enum(),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -899,6 +899,7 @@ dependencies = [
  "serde_json",
  "serde_yaml",
  "sqlx",
+ "strum",
  "thiserror",
  "utoipa",
 ]
--- a/components/clp-rust-utils/Cargo.toml
+++ b/components/clp-rust-utils/Cargo.toml
@@ -15,6 +15,7 @@ secrecy = { version = "0.10.3", features = ["serde"] }
 serde = { version = "1.0.228", features = ["derive"] }
 serde_yaml = "0.9.34"
 sqlx = { version = "0.8.6", features = ["runtime-tokio", "mysql"] }
+strum = "0.28.0"
 thiserror = "2.0.18"
 utoipa = { version = "5.4.0" }
 
--- a/components/clp-rust-utils/src/database/mysql.rs
+++ b/components/clp-rust-utils/src/database/mysql.rs
@@ -1,9 +1,27 @@
 use secrecy::ExposeSecret;
+use strum::IntoEnumIterator;
 
 use crate::clp_config::package::{
     config::Database as DatabaseConfig,
     credentials::Database as DatabaseCredentials,
 };
+
+/// Trait for formatting Rust enums as SQL `ENUM(...)` declarations.
+pub trait MySqlEnumFormat: IntoEnumIterator + Sized + ToString
+where
+    Self::Iterator: Iterator<Item = Self>, {
+    /// # Returns
+    ///
+    /// A string representing the SQL enum definition for this enum.
+    #[must_use]
+    fn format_as_sql_enum() -> String {
+        let inner = Self::iter()
+            .map(|v| format!("'{}'", v.to_string()))
+            .collect::<Vec<_>>()
+            .join(", ");
+        format!("ENUM({inner})")
+    }
+}
 
 /// Creates a new `MySQL` connection pool to the CLP DB using the provided configuration and
 /// credentials.


components/log-ingestor/Cargo.toml
```diff
@@ -29,6 +29,8 @@ secrecy = "0.10.3"
 serde = { version = "1.0.228", features = ["derive"] }
 serde_json = "1.0.149"
 sqlx = { version = "0.8.6", features = ["runtime-tokio", "mysql"] }
+strum = "0.28.0"
+strum_macros = "0.28.0"
```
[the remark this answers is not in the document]
LinZhihao-723: u're outdated. It's a new release.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -0,0 +1,518 @@
...
+    `creation_ts` DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
+    `last_update_ts` DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3)
+        ON UPDATE CURRENT_TIMESTAMP(3),
+    PRIMARY KEY (`id`)
+) ROW_FORMAT=DYNAMIC;",
+        table = INGESTION_JOB_TABLE_NAME,
+        status_enum = ClpIngestionJobStatus::format_as_sql_enum(),
+        default_status = ClpIngestionJobStatus::Requested,
+    )
+}
```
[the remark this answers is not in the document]
LinZhihao-723: It should be fine for now.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -0,0 +1,537 @@
...
+    /// A string representing the SQL enum definition for this enum.
+    #[must_use]
+    fn format_as_sql_enum() -> String {
+        let inner = Self::iter()
+            .map(|v| format!("'{}'", v.to_string()))
+            .collect::<Vec<_>>()
+            .join(", ");
+        format!("ENUM({inner})")
+    }
+}
```
sitaowang1998: This should probably be put in another SQL helper files. Also, create a custom derive macro to automatically generate the `imple MySqlEnumForat for ... {}` would be a plus.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -0,0 +1,537 @@
...
+    `config` VARCHAR(1024) NOT NULL,
+    `status` {status_enum} NOT NULL DEFAULT '{default_status}',
+    `num_files_ingested` BIGINT unsigned NOT NULL DEFAULT '0',
+    `num_files_compressed` BIGINT unsigned NOT NULL DEFAULT '0',
+    `error_msg` TEXT NULL DEFAULT NULL,
+    `creation_ts` DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
+    `last_update_ts` DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3)
+        ON UPDATE CURRENT_TIMESTAMP(3),
+    PRIMARY KEY (`id`)
+) ROW_FORMAT=DYNAMIC;",
```
sitaowang1998: Since dynamic is the default row format https://mariadb.com/docs/server/server-usage/storage-engines/innodb/innodb-row-formats/innodb-dynamic-row-format, I don't think need to explicitly specify this. If the user deploys a `MariaDB` instance with other settings, we should probably honor that.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -0,0 +1,537 @@
...
+fn ingestion_job_s3_scanner_state_table_creation_query() -> String {
+    format!(
+        r"
+CREATE TABLE IF NOT EXISTS `{INGESTION_JOB_S3_SCANNER_STATE_TABLE_NAME}` (
+    `id` BIGINT unsigned NOT NULL,
+    `last_ingested_key` VARCHAR(1024) NULL DEFAULT NULL,
+    PRIMARY KEY (`id`),
+    CONSTRAINT `{INGESTION_JOB_S3_SCANNER_STATE_TABLE_NAME}_ingestion_job_id_ref`
+        FOREIGN KEY (`id`) REFERENCES `{INGESTION_JOB_TABLE_NAME}` (`id`)
+        ON DELETE RESTRICT ON UPDATE RESTRICT
```
sitaowang1998, line 482: Why restrict the deletion in the job table?

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -0,0 +1,537 @@
...
+        self.update_job_status(ClpIngestionJobStatus::Running).await
+    }
+
+    /// # Errors
+    ///
+    /// * Forwards [`Self::update_job_status`]'s return values on failure.
+    async fn end(&self) -> anyhow::Result<()> {
+        self.update_job_status(ClpIngestionJobStatus::Finished)
+            .await
+    }
```
sitaowang1998, line 285: Should we check if the job is in correct state before updating it?

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -0,0 +1,537 @@
...
+        self.update_ingestion_stats_and_commit(tx, objects.len())
+            .await
+            .inspect_err(|err| {
+                tracing::error!(
+                    error = ? err,
+                    job_id = ? self.job_id,
+                    "Failed to commit ingestion."
+                );
+            })
+    }
```
sitaowang1998, line 347: The transaction is not handled correctly on failure. On failure, the transaction needs to be explicitly aborted by calling `tx.rollback`. Maybe we can unify the transaction logic in a helper function that accept a function which takes a transaction as an argument. The helper function creates a transaction, call the function passed in, rollback and forward the error on failure and commit on success.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -0,0 +1,537 @@
...
```
sitaowang1998, line 380: Same transaction problem as above.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
LinZhihao-723: I think it's fine. We should be able to restart or stop a job from any state for now.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
LinZhihao-723: iiuc we don't need to call it explicitly since `rollback` is called inside `drop` if not committed: https://docs.rs/sqlx/latest/sqlx/struct.Transaction.html

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
LinZhihao-723: iiuc, adding "restrict" means the job can't be removed if there are still rows in other tables that reference it. That's probably the behavior we want: we don't want these ingested rows to be accidentally deleted because the job gets removed. Instead, they should be persisted in the DB until they have been ingested.
They plan (not in this release, probably) to clean the ingested metadata by having the GC periodically clean up the ones already cleaned up. And an ingestion job can only be removed if the referenced metadata has been removed by GC, or has been explicitly removed by users.
To achieve this behavior, do you have better suggestions other than using "restrict"?

sitaowang1998: As long as the intended behaviour is to make deleting job harder, `RESTRICT` is a good choice.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -0,0 +1,537 @@
...
+        objects: &[ObjectMetadata],
+    ) -> anyhow::Result<()> {
+        const BASE_INGESTION_QUERY: &str = formatcp!(
+            r"INSERT INTO `{table}` (`bucket`, `key`, `size`, `ingestion_job_id`) VALUES ",
+            table = INGESTED_S3_OBJECT_METADATA_TABLE_NAME,
+        );
+        assert!(
+            !objects.is_empty(),
+            "Cannot build S3 object metadata ingestion query with empty objects"
+        );
```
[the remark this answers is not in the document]
LinZhihao-723: This is intended.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -0,0 +1,537 @@
...
+#[sqlx(rename_all = "snake_case")]
+#[strum(serialize_all = "snake_case")]
+pub enum IngestedS3ObjectMetadataStatus {
+    Buffered,
+    Submitted,
+    Compressed,
+    Failed,
+}
+
+impl MySqlEnumFormat for IngestedS3ObjectMetadataStatus {}
```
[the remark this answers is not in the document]
LinZhihao-723: This is because it's not used in this PR. We will add it in the next PR where we will add more operations to update the metadata status.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -0,0 +1,537 @@
...
+    format!(
+        r"
+CREATE TABLE IF NOT EXISTS `{table}` (
+    `id` BIGINT unsigned NOT NULL AUTO_INCREMENT,
+    `bucket` VARCHAR(1024) NOT NULL,
+    `key` VARCHAR(1024) NOT NULL,
+    `size` BIGINT unsigned NOT NULL,
+    `status` {status_enum} NOT NULL DEFAULT '{default_status}',
+    `ingestion_job_id` BIGINT unsigned NOT NULL,
+    `compression_job_id` INT NULL DEFAULT NULL,
```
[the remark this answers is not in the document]
LinZhihao-723: It should match, as the ID is defined [here](https://github.com/y-scope/clp/blob/13b47f069e16e1c70e2575080edbd970c82f9c8f/components/clp-py-utils/clp_py_utils/initialize-orchestration-db.py#L65).

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -0,0 +1,537 @@
...
+            r"UPDATE `{table}` SET `last_ingested_key` = ? WHERE `id` = ?;",
+            table = INGESTION_JOB_S3_SCANNER_STATE_TABLE_NAME,
+        );
+
+        let mut tx = self.db_pool.begin().await?;
+        sqlx::query(UPDATE_S3_SCANNER_STATE_QUERY)
+            .bind(last_ingested_key)
+            .bind(self.job_id)
+            .execute(&mut *tx)
+            .await?;
```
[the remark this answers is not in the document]
LinZhihao-723: Fixed.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
LinZhihao-723: Moved to `clp_rust_utils::db::mysql`.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
LinZhihao-723: Removed.

PR #2015, opened 2026-02-23
feat(integration-tests): Add end-to-end test workflow for downstream `clp-ffi-js`; Bump `yscope-dev-utils` to y-scope/yscope-dev-utils@e2a1aed.

Bill-hbrhbr:
# Description

Since `clp-ffi-js` builds CLP from source using Emscripten and Clang as part of its WASM toolchain, this PR introduces end to end testing to ensure that changes in the CLP repository do not introduce build failures downstream in `clp-ffi-js`.

Additionally, this PR bumps the `yscope-dev-utils` version to incorporate the fix for the broken Boost argument handling described in https://github.com/y-scope/yscope-dev-utils/pull/103

# Checklist

  breaking change.

# Validation performed

* [X] `task tests:clp-ffi-js` succeeds locally.

## Summary by CodeRabbit

* **New Features**
  * Added a taskfile to integrate and set up the CLP FFI JavaScript workspace with configurable source/build flows.

* **Tests**
  * Added end-to-end automation: download/extract sources with checksum validation, cmake build targets, Node and browser test flows (browser tasks require elevated privileges), and support for local symlinked dependency integration.

* **Chores**
  * Updated development tooling reference to a newer revision and adjusted the tooling download/extract setup.


taskfiles/tests/clp-ffi-js.yaml
```diff
@@ -0,0 +1,95 @@
...
+# - `task tests:clp-ffi-js:build-and-test-node`
+# - `task tests:clp-ffi-js:build-and-test-browser`
+# as needed. Note that `sudo` access is required for the latter.
+tasks:
+  default:
+    dir: "{{.G_CLP_FFI_JS_SRC_DIR}}"
+    deps: ["download-src"]
+    cmds:
+      # Expose `clp-ffi-js` tasks for test setup
+      - task: "symlink-yscope-dev-utils"
```
hoophalab, line 32: Why do we make `symlink-yscope-dev-utils` a separate sub-task while keep
```
        rm -rf build/deps/clp
        ln -s "{{.ROOT_DIR}}" build/deps/clp
```
inline? Might be cleaner to make both inline.

hoophalab (approved):
LGTM. One style comment.

This isn't actually executed within GitHub workflow, right? Might be best to change this PR title to `end-to-end test task for downstream...`

Validation: The task successfully downloads clp-ffi-js and executes the tests.

taskfiles/tests/clp-ffi-js.yaml
```diff
@@ -0,0 +1,95 @@
...
+      - |-
+        sed -i 's/^\(\s*internal:\s*\)true\b/\1false/' taskfile.yaml
```
junhaoliao, line 34: given the tasks in clp-ffi-js is now publicly (external to the clp-ffi-js) used, maybe we should change the tasks in clp-ffi-js to non-internal?

so long we dont document those tasks in clp-ffi-js docs, it shouldn't cause confusion to developers about what tasks they should run? if you really want, i think we can add inline docstrings to those tasks to explain thse are needed to set up the right dependencies for integration tests in the clp project. either way, it should be cleaner / more future-proof than this sed here

taskfiles/tests/clp-ffi-js.yaml
```diff
@@ -0,0 +1,95 @@
...
+    deps:
+      - "::init"
+    cmds:
+      - task: "yscope-dev-utils:remote:download-and-extract-tar"
+        vars:
+          CHECKSUM_FILE: "{{.G_CLP_FFI_JS_CHECKSUM_FILE}}"
+          FILE_SHA256: "5ab1c27031caafb014198d4db711a71b3fbb1661b396bcd4e46764194e2b1f61"
+          OUTPUT_DIR: "{{.G_CLP_FFI_JS_SRC_DIR}}"
+          TAR_FILE: "{{.G_BUILD_DIR}}/clp-ffi-js.tar.gz"
+          URL: "https://github.com/y-scope/clp-ffi-js/archive/4cc6a7c.tar.gz"
```
junhaoliao, line 87: do you think it's worth adding a docstring to explain when this commit was available? / in the future if we pin those commit ids at specific clp-ffi-js releases, it's better to comment with the release versions as well

junhaoliao: something like
```
# clp-ffi-js @ v0.X.Y (March 2, 2026)
```

taskfiles/tests/clp-ffi-js.yaml
```diff
@@ -0,0 +1,95 @@
...
+  build-and-test-node:
+    dir: "{{.G_CLP_FFI_JS_SRC_DIR}}"
+    deps: ["build-targets"]
+    cmds:
+      - |-
+        VITE_NODE_MODULE_ABS_PATH="{{.G_CLP_FFI_JS_NODE_MODULE_PATH}}" \
+          npm run test -- --project node
+
+  # NOTE: Browser tests require installing browsers and system dependencies. The task is skipped if
+  # sudo is unavailable.
```
junhaoliao, line 53:
```suggestion
  # passwordless sudo is unavailable.
```

taskfiles/tests/clp-ffi-js.yaml
```diff
@@ -0,0 +1,95 @@
...
+  build-and-test-browser:
+    dir: "{{.G_CLP_FFI_JS_SRC_DIR}}"
+    deps: ["build-targets"]
+    status:
+      - "! (command -v sudo >/dev/null && sudo -n true)"
+    cmds:
+      - |-
```
junhaoliao, line 60:
```suggestion
    cmds:
      - |-
        if ! (command -v sudo >/dev/null && sudo -n true 2>/dev/null); then
          echo "WARNING: Skipping browser tests (passwordless sudo is unavailable)."
          exit 0
        fi
```

taskfiles/tests/clp-ffi-js.yaml
```diff
@@ -0,0 +1,95 @@
...
+    deps: ["download-src"]
+    cmds:
+      # Expose `clp-ffi-js` tasks for test setup
+      - task: "symlink-yscope-dev-utils"
+      - |-
+        sed -i 's/^\(\s*internal:\s*\)true\b/\1false/' taskfile.yaml
+        task node-modules
+        task config-cmake-project
+        rm -rf build/deps/clp
+        ln -s "{{.ROOT_DIR}}" build/deps/clp
```
junhaoliao, line 38:
```suggestion
        rm -rf "build/deps/clp"
        ln -s "{{.ROOT_DIR}}" "build/deps/clp"
```

taskfiles/tests/clp-ffi-js.yaml
junhaoliao, line 1: if there's no easy way to set up `clp-ffi-js` as one of `deps` for `build-and-test-node` and `build-and-test-browser`, how do you feel about adding this to both tasks:
```yaml
    preconditions:
      - msg: "clp-ffi-js is not set up. Run `task tests:clp-ffi-js` first."
        sh: "test -d {{.G_CLP_FFI_JS_BUILD_DIR}}"
```

junhaoliao (changes requested):
the code lgtm

at a high level, do note that we are hardcoding the clp-ffi-js commit ref here, which means we may not be testing with the latest clp-ffi-js code. we probably shouldn't use the latest commit from clp-ffi-js anyways, like otherwise this flow that blocks PR submission cannot run deterministically. i wonder if it would be cleaner to instead set up some nightly flow in clp-ffi-js which builds clp-ffi-js against the latest version of CLP. if anything breaks, it will be clp-ffi-js' maintainers' (some of whom are clp's maintainers) onus to fix the issues / request fixes from the clp maintainers. what do you think?

PR #2017, opened 2026-02-24
feat(log-ingestor): Add DB operations for compression job submission, result retrieval, and ingestion metadata status updates.

LinZhihao-723:
# Description

This PR implements DB operations for compression job submission, result retrieval, and ingestion metadata status updates. Essentially, it's an upgraded implementation for the logic in [compression job submitter](https://github.com/y-scope/clp/blob/main/components/log-ingestor/src/compression/compression_job_submitter.rs) with the following improvements:

- In a single transaction, a compression job is submitted, and its ID will be assigned to all submitted S3 object metadata.
- When the compression job completes, its result will be updated to the submitted S3 object metadata. And on a successful compression, the ingestion job's `num_files_compressed` will be updated.

The plan for the next release is to update `compression_job_submitter` implementation to use `ClpIngestionState` for compression job submission and result retrieval.

The long-term plan for compression job submission is tracked in #2018.

# Checklist

  breaking change.

# Validation performed

* [x] Run the following tests locally:
  * Step 1: Run [this](https://github.com/y-scope/spider/blob/main/tools/scripts/mariadb/start.py) script to start a mariaDB instance on localhost:
    ```shell
    ./start.py --name "clp-db-test" --port 3306 --database "clp-db" --username "ROOT" --password "ROOT"
    ```
  * Step 2: Add the following code to the test section in `clp_ingestion.rs`:
    ```Rust
    #[tokio::test]
    async fn test_table_creation() {
        use clp_rust_utils::{
            clp_config::{AwsAuthentication::Credentials, AwsCredentials, S3Config},
            job_config::{S3InputConfig, OutputConfig, ingestion::s3::S3ScannerConfig},
            types::non_empty_string::ExpectedNonEmpty,
        };
        use serde_json::json;

        let mysql_options = sqlx::mysql::MySqlConnectOptions::new()
            .host("localhost")
            .port(3306)
            .database("clp-db")
            .username("ROOT")
            .password("ROOT");

        let db_pool = sqlx::mysql::MySqlPoolOptions::new()
            .max_connections(100)
            .connect_with(mysql_options)
            .await
            .expect("Failed to connect to mysql");

        sqlx::query(
            r"
                CREATE TABLE IF NOT EXISTS `compression_jobs` (
                    `id` INT NOT NULL AUTO_INCREMENT,
                    `status` INT NOT NULL DEFAULT '0',
                    `status_msg` VARCHAR(512) NOT NULL DEFAULT '',
                    `creation_time` DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
                    `start_time` DATETIME(3) NULL DEFAULT NULL,
                    `update_time` DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP(),
                    `duration` FLOAT NULL DEFAULT NULL,
                    `original_size` BIGINT NOT NULL DEFAULT '0',
                    `uncompressed_size` BIGINT NOT NULL DEFAULT '0',
                    `compressed_size` BIGINT NOT NULL DEFAULT '0',
                    `num_tasks` INT NOT NULL DEFAULT '0',
                    `num_tasks_completed` INT NOT NULL DEFAULT '0',
                    `clp_binary_version` INT NULL DEFAULT NULL,
                    `clp_config` VARBINARY(60000) NOT NULL,
                    PRIMARY KEY (`id`) USING BTREE,
                    INDEX `JOB_STATUS` (`status`) USING BTREE,
                    INDEX `JOB_UPDATE_TIME` (`update_time`) USING BTREE
                ) ROW_FORMAT=DYNAMIC
        ",
        )
            .execute(&db_pool)
            .await
            .expect("Failed to create compression jobs table");

        let connector = match ClpDbIngestionConnector::new(db_pool.clone()).await {
            Ok(c) => c,
            Err(e) => {
                panic!("Failed to create CLP DB connector: {e}");
            }
        };

        let json_config = json!(
            {
                "bucket_name": "test",
                "key_prefix": "test-",
                "region": "us-east-2",
            }
        );

        let config: S3ScannerConfig = serde_json::from_str(json_config.to_string().as_str())
            .expect("Failed to deserialize S3 ingestion job config");

        let (sender, mut receiver) = mpsc::channel::<Vec<ObjectMetadata>>(100);

        let state = match connector
            .create_ingestion_job(S3IngestionJobConfig::S3Scanner(config), sender.clone())
            .await
        {
            Ok(s) => s,
            Err(e) => {
                panic!("Failed to create ingestion job: {e}");
            }
        };

        state.start().await.expect("Failed to start ingestion job");

        let mut objects = Vec::new();
        for i in 0..12345 {
            objects.push(ObjectMetadata {
                bucket: NonEmptyString::from_static_str("test"),
                key: NonEmptyString::from_string(format!("test-{i}")),
                size: 100 + i,
                id: None,
            });
        }

        let last_ingested_key = objects.last().unwrap().key.clone();
        S3ScannerState::ingest(&state, objects, last_ingested_key.as_str())
            .await
            .expect("Failed to ingest");

        state.end().await.expect("Failed to end ingestion job");

        let objects = vec![ObjectMetadata {
            bucket: NonEmptyString::from_static_str("test"),
            key: NonEmptyString::from_static_str("test-10000"),
            size: 8,
            id: None,
        }];
        let last_ingested_key = objects.last().unwrap().key.clone();
        assert!(
            S3ScannerState::ingest(&state, objects, last_ingested_key.as_str())
                .await
                .is_err()
        );

        // Compression
        let objects_to_submit = receiver.recv().await.expect("sth");
        assert_eq!(objects_to_submit.len(), 12345);

        for chunk in objects_to_submit.chunks(3000) {
            let s3_input_config = S3InputConfig {
                s3_config: S3Config {
                    bucket: NonEmptyString::from_static_str("test"),
                    region_code: Some(NonEmptyString::from_static_str("us-east-2")),
                    key_prefix: NonEmptyString::from_static_str("test-"),
                    endpoint_url: None,
                    aws_authentication: Credentials {
                        credentials: AwsCredentials {
                            access_key_id: "ROOT".to_owned(),
                            secret_access_key: "ROOT".to_owned(),
                        },
                    },
                },
                keys: None,
                dataset: Some(NonEmptyString::from_static_str("default")),
                timestamp_key: None,
                unstructured: false,
            };
            let output_config = OutputConfig {
                target_archive_size: 1000,
                target_dictionaries_size: 1000,
                target_encoded_file_size: 1000,
                target_segment_size: 1000,
                compression_level: 3,
            };
            let io_config_template = ClpIoConfig {
                input: InputConfig::S3InputConfig {
                    config: s3_input_config,
                },
                output: output_config,
            };

            let compression_job_id = state
                .submit_for_compression(io_config_template, &chunk)
                .await
                .expect("Failed to submit for compression");
            let state_copy = state.clone();
            let num_objects_submitted = chunk.len();
            let handle = tokio::spawn(async move {
                state_copy
                    .wait_for_compression_and_update_submitted_metadata(
                        compression_job_id,
                        num_objects_submitted,
                    )
                    .await
                    .expect("Failed to wait for compression result")
            });

            // Manually mark the compression as successful
            sqlx::query(
                format!(
                    r#"UPDATE `{table}` SET `status` = ? WHERE `id` = ?"#,
                    table = CLP_COMPRESSION_JOB_TABLE_NAME
                )
                    .as_str(),
            )
                .bind(2)
                .bind(compression_job_id)
                .execute(&db_pool)
                .await
                .expect("Failed to mark compression job as successful");

            let (status, status_msg) = handle.await.expect("Failed to join handle");
            assert_eq!(status, CompressionJobStatus::Succeeded);
        }
    }
    ```
  * Step 3: Run the tester (inside the IDE or through the terminal) and make sure:
    * The DB contains the expected tables with the correct schema.
    * One ingestion job is created.
    * 37 files are ingested into the created ingestion job (on both the ingestion job table and S3 object metadata table).
    * The ingestion job ends with a `finished` state.
    * Five compression jobs are submitted and finished successfully.
    * All metadata rows are marked as `compressed` across 5 compression jobs.
    * All files are compressed in the ingestion job table while the job status is "finished".

## Summary by CodeRabbit

* **New Features**
  * Compression job submission and status tracking for ingested S3 objects.
  * Asynchronous metadata processing with channel-based forwarding.
  * Object metadata now includes an optional identifier.

* **Refactor**
  * Database indexes adjusted to support ingestion and compression tracking.
  * Compression status polling enhanced with backoff-based wait logic.

* **Tests**
  * Test fixtures updated to include the new optional metadata id.

### DIFF
--- a/components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
+++ b/components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
@@ -1,6 +1,7 @@
 use async_trait::async_trait;
 use clp_rust_utils::{
     database::mysql::MySqlEnumFormat,
+    impl_sqlx_type,
     job_config::{
         ClpIoConfig,
         CompressionJobId,
@@ -200,8 +201,12 @@ impl ClpIngestionState {
                 anyhow::anyhow!("The retrieved ID overflows: {}", result.last_insert_id())
             })?;
 
-        // Update compression job ID for ingested objects
-        for chunk in objects.chunks(1000) {
+        // Update compression job ID for ingested objects.
+        // NOTE: We batch the update to avoid hitting the maximum placeholder limit of MySQL. The
+        // batch size is chosen to be 10000, which is conservative enough to avoid hitting the limit
+        // while also minimizing the number of batches for typical use cases. If the number of
+        // placeholders per update changes, we may need to adjust the batch size accordingly.
+        for chunk in objects.chunks(10000) {
             let mut query_builder = sqlx::QueryBuilder::<sqlx::MySql>::new(formatcp!(
                 r"UPDATE `{table}` ",
                 table = INGESTED_S3_OBJECT_METADATA_TABLE_NAME,
@@ -699,15 +704,7 @@ pub enum ClpIngestionJobStatus {
 
 impl MySqlEnumFormat for ClpIngestionJobStatus {}
 
-impl sqlx::Type<sqlx::MySql> for ClpIngestionJobStatus {
-    fn type_info() -> <sqlx::MySql as sqlx::Database>::TypeInfo {
-        <str as sqlx::Type<sqlx::MySql>>::type_info()
-    }
-
-    fn compatible(ty: &<sqlx::MySql as sqlx::Database>::TypeInfo) -> bool {
-        <str as sqlx::Type<sqlx::MySql>>::compatible(ty)
-    }
-}
+impl_sqlx_type!(ClpIngestionJobStatus => str);
 
 /// Enum for CLP ingestion S3 object metadata status.
 #[derive(
@@ -734,15 +731,7 @@ pub enum IngestedS3ObjectMetadataStatus {
 
 impl MySqlEnumFormat for IngestedS3ObjectMetadataStatus {}
 
-impl sqlx::Type<sqlx::MySql> for IngestedS3ObjectMetadataStatus {
-    fn type_info() -> <sqlx::MySql as sqlx::Database>::TypeInfo {
-        <str as sqlx::Type<sqlx::MySql>>::type_info()
-    }
-
-    fn compatible(ty: &<sqlx::MySql as sqlx::Database>::TypeInfo) -> bool {
-        <str as sqlx::Type<sqlx::MySql>>::compatible(ty)
-    }
-}
+impl_sqlx_type!(IngestedS3ObjectMetadataStatus => str);
 
 const INGESTION_JOB_TABLE_NAME: &str = "ingestion_job";
 const INGESTION_JOB_S3_SCANNER_STATE_TABLE_NAME: &str = "ingestion_job_s3_scanner_state";
--- a/components/clp-rust-utils/src/database/mysql.rs
+++ b/components/clp-rust-utils/src/database/mysql.rs
@@ -6,6 +6,28 @@ use crate::clp_config::package::{
     credentials::Database as DatabaseCredentials,
 };
 
+/// Implements [`sqlx::Type<sqlx::MySql>`] for `$ty` by delegating to `$delegate`.
+///
+/// # Examples
+///
+/// ```rust
+/// impl_sqlx_type!(IngestedS3ObjectMetadataStatus => str);
+/// ```
+#[macro_export]
+macro_rules! impl_sqlx_type {
+    ($ty:ty => $delegate:ty $(,)?) => {
+        impl ::sqlx::Type<::sqlx::MySql> for $ty {
+            fn type_info() -> <::sqlx::MySql as ::sqlx::Database>::TypeInfo {
+                <$delegate as ::sqlx::Type<::sqlx::MySql>>::type_info()
+            }
+
+            fn compatible(ty: &<::sqlx::MySql as ::sqlx::Database>::TypeInfo) -> bool {
+                <$delegate as ::sqlx::Type<::sqlx::MySql>>::compatible(ty)
+            }
+        }
+    };
+}
+
 /// Trait for formatting Rust enums as SQL `ENUM(...)` declarations.
 pub trait MySqlEnumFormat: IntoEnumIterator + Sized + ToString
 where


components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -265,6 +468,125 @@ impl ClpIngestionState {
...
+                    job_id = ? self.job_id,
+                    "Failed to commit ingestion."
+                );
+            })?;
+
+        for (chunk_id, chunk) in objects.chunks_mut(chunk_size).enumerate() {
+            let mut next_metadata_id = *last_inserted_ids.get(chunk_id).expect("invalid chunk ID");
+            for object in chunk {
+                object.id = Some(next_metadata_id);
+                next_metadata_id += 1;
```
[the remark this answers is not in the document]
LinZhihao-723: We've explained this assumption in `ingest_s3_object_metadata`'s documentation.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -110,9 +120,202 @@ impl ClpDbIngestionConnector {
...
+
+        // Submit compression job
+        let result = sqlx::query(COMPRESSION_JOB_SUBMISSION_QUERY)
+            .bind(clp_rust_utils::serde::BrotliMsgpack::serialize(&io_config)?)
+            .execute(&mut *tx)
+            .await?;
+        let compression_job_id =
+            CompressionJobId::try_from(result.last_insert_id()).map_err(|_| {
+                anyhow::anyhow!("The retrieved ID overflows: {}", result.last_insert_id())
+            })?;
```
sitaowang1998, line 202: Since the id is auto increment, if a previous id overflows, all later ids are bound to overflow. Maybe we should panic instead?

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -110,9 +120,202 @@ impl ClpDbIngestionConnector {
...
+
+        // Update compression job ID for ingested objects
+        for chunk in objects.chunks(1000) {
```
sitaowang1998: 1. Why this chunk size is just 1000?
2. Since we need chunk size more than once, should we create a const function to compute chunk size given the number of query parameters per entry?

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -441,6 +734,16 @@ pub enum IngestedS3ObjectMetadataStatus {
...
+impl sqlx::Type<sqlx::MySql> for IngestedS3ObjectMetadataStatus {
+    fn type_info() -> <sqlx::MySql as sqlx::Database>::TypeInfo {
+        <str as sqlx::Type<sqlx::MySql>>::type_info()
+    }
+
+    fn compatible(ty: &<sqlx::MySql as sqlx::Database>::TypeInfo) -> bool {
+        <str as sqlx::Type<sqlx::MySql>>::compatible(ty)
+    }
+}
+
```
sitaowang1998: I have seen this code twice in this file. Is there another way to do it with less boilerplate code?

sitaowang1998 (changes requested):

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
LinZhihao-723: I think this should be fine, considering it's implementing sqlx::Type trait, where we may need different underlying types (in the enum case it's `str`) accordingly for sqlx to interpret.

sitaowang1998: Can we write a macro for this, where the underlying type is a parameter?

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
LinZhihao-723: We only use panic to catch internal errors (for example, all object metadata passed to the compression should have the ingestion metadata ID set). And when the panic happens, it means something wrong inside the log-ingestor implementation.
However, the overflow here is not a part of log-ingestor's responsibility, so we just return an error and rely on the top-level caller to log the error (and no retry logic).
To fix this in a legit way, I think we should improve the compression job side to use `BIGINT unsigned` IDs (not `INT`). And in that case, we should be good to panic here since that overflow will become a really "unrecoverable" issue.

sitaowang1998: @coderabbit Please open an issue about using `BIGINT UNSIGNED` for all ids.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
LinZhihao-723: 1. I chose 1000 'cuz I thought the IDs are in-place values, not placeholders. But I just realized we're actually using placeholders here since IDs are uploaded through `bind`. I elevate this number to 10000 with a similar comment. This value can be smaller than the one in ingestion to be conservative, since it's not in a critical path, so no aggressive batching would be needed.
2. I think it's meaningless to create such a helper if the number of placeholders can't be derived directly from the statement. Also, depends on the use case we may choose the chunk size accordingly like explained in (1). Therefore, I think a hardcoded value here might be ok for now, as long as we clearly document the limitation in the comment.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
LinZhihao-723: Added in the latest commit.

sitaowang1998 (approved):

PR #2031, opened 2026-02-25
feat(clp-s): Add `--remove-path-prefix`, `--remove-leading-slash`, and `--normalize-paths` command line options (resolves #2011).

### LINKED ISSUE #2011: Add --remove-prefix option at compression in clp-s
### Request

We have IR files stored in S3 that need to be consolidated into an archive. Their S3 keys follow the pattern abc/def/ghi-1.ir, abc/def/ghi-2.ir, and so on.

The consolidation service downloads all consolidatable IRs from S3 to the local filesystem (e.g. /tmp/ir-20260220), performs consolidation, then uploads the resulting archive back to S3.

However, this causes $_filePath stored in archive to be set to the full local path /tmp/ir-20260220/abc/def/ghi-1.ir. The /tmp/ir-20260220 prefix must be stripped, as it is not persisted in the metadata database — leaving it in place breaks the lookup that maps file-level information to its associated metadata.

### Possible implementation

Should be a simple feature, any implementation is okay.

gibber9809:
# Description

This PR adds the new `--remove-path-prefix` and `--normalize-paths` flags to the compression flow. Both of these flags only apply to files ingested from the filesystem, and are silently ignored for network paths.

`--normalize-paths` will convert all paths passed to compression into absolute paths relative to the root without extra `..` or `.` -- these absolute paths are what get stored as the filename in the archive metadata.

`--remove-path-prefix <prefix>` will remove a prefix from file paths. When used on its own, the prefix path must strictly be a string prefix of file paths passed to compression, but when used in conjunction with `--normalize-paths` the prefix path only needs to represent a directory prefix and doesn't necessarily need to be a string prefix as well. Note: when removing path prefixes we always leave `/` as the start of the file path.

For example, with the directory structure `/root/a/b/c.jsonl`, when running commands from the `/root/` subdirectory:
`clp-s c arch a/ --remove-path-prefix a/` -> results in path `/b/c.jsonl` in metadata
`clp-s c arch a --remove-path-prefix ../root/a` -> doesn't work since `../root/a` isn't a string prefix of `a` even though we're referring to the same directory
`clp-s c arch a --remove-path-prefix ../root/a --normalize-paths` -> results in `/b/c.jsonl` in metadata
`clp-s c arch a --normalize-paths` -> results in `/root/a/b/c.jsonl` in metadata

I'm open to feedback for whether `--normalize-paths` should be on by default for compression (since it gives more predictable file names in the metadata no matter how compression is invoked + makes the semantics of `--remove-path-prefix simpler`), and make a `--no-normalize-paths` flag instead. My only hesitation is that that would technically be a breaking change (though probably not a harmful one).

# Checklist

  breaking change.

# Validation performed

* Added unit tests for path prefix removal utility
* Manually validated that file-level metadata is correct after compression when using the `--remove-path-prefix` and `--normalize-paths` flags

## Summary by CodeRabbit

* **New Features**
  * Added options to canonicalize input file paths and to optionally strip a configured path prefix; ingested files record filenames after these transformations.

* **Bug Fixes / Improvements**
  * Input ingestion and metadata consistently use transformed/canonical filenames; per-input pairing of source path and canonical filename improves accuracy.

* **Tests**
  * Added unit tests verifying path-prefix removal behaviour.

* **Chores**
  * Included the new test in the unit test suite.


20001020ycx:
### Code review

Found 3 issues:

1. `--remove-path-prefix` unnecessarily aborts compression when `--disable-log-order` is also set

When `--disable-log-order` is active (`m_record_log_order == false`), `file_name_in_metadata` is never actually written to the archive — `initialize_fields_for_archive` returns immediately. However, the prefix-removal check in `ingest()` still runs unconditionally and returns `false` on any mismatch. This means combining `--disable-log-order` with `--remove-path-prefix` can cause compression to abort even when the prefix feature has no effect on the archive output.

https://github.com/y-scope/clp/blob/8ca683180078971f58a12612cfc6d98d3186669d/components/core/src/clp_s/JsonParser.cpp#L650-L665

2. Error format for normalization failures omits `ec.category().name()`

The new error messages use only `ec.message()`, deviating from the established pattern throughout the codebase (e.g. `JsonParser.cpp`, `ColumnWriter.cpp`, `TimestampDictionaryWriter.cpp`) which consistently includes both `ec.category().name()` and `ec.message()` for `std::error_code` failures.

https://github.com/y-scope/clp/blob/8ca683180078971f58a12612cfc6d98d3186669d/components/core/src/clp_s/CommandLineArguments.cpp#L345-L371

3. Validation error messages for `--remove-path-prefix` do not include the offending path

`"Specified prefix to remove does not exist."` and `"Specified prefix to remove is not a directory."` omit the value of `path_prefix_to_remove`, making these errors harder to diagnose. Nearby error messages in the same file (e.g. `"Invalid input path \"{}\"."` at line 376) consistently include the path value.

https://github.com/y-scope/clp/blob/8ca683180078971f58a12612cfc6d98d3186669d/components/core/src/clp_s/CommandLineArguments.cpp#L333-L339

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -647,6 +648,22 @@ void JsonParser::parse_line(
 bool JsonParser::ingest() {
     auto archive_creator_id = boost::uuids::to_string(m_generator());
     for (auto const& path : m_input_paths) {
+        std::string file_name_in_metadata{path.path};
```
20001020ycx: Is there a specific reason we put such validation in the ingestion flow? Can we move this to CommandLineArguments.cpp? I saw there's a [for loop](https://github.com/y-scope/clp/pull/2031/changes#diff-6b02224344ae7558e1bcaa4b74ecb0271db8ec129c6ee0d7da25be0db1b92a2aL321) that does validation for path normalization. Can we throw this validation logic there, and here would just be simply extracting the removed file path, all the error handlings are hoisted at the command line.

components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -248,6 +250,16 @@ CommandLineArguments::parse_arguments(int argc, char const** argv) {
...
+            )(
+                    "normalize-paths",
+                    po::bool_switch(&normalize_file_paths),
+                    "Make all paths for ingested files absolute relative to the root directory."
+            )(
+                    "remove-path-prefix",
+                    po::value<std::string>(&path_prefix_to_remove)
+                            ->value_name("DIR")
+                            ->default_value(path_prefix_to_remove),
+                    "Remove the given path prefix from each compressed file."
```
jackluo923: After thinking this through, enforcing `/` will make it consistent for user, however it would mean it would be impossible to represent object storage keys. Could we add one of the two flags to achieve this purpose? 
```
--relative-paths <path prefix directory>   # Strip the prefix directory and '/' with only the relative path intact, follows linux-style relative path behavior, probably should be exclusive w/ remove-path-prefix

--no-leading-slash    # Self explanatory
```

jackluo923 (changes requested):
Identified a requirement gap w.r.t to representing object storage key, or simply a representing a relative path in the general sense.

components/core/src/clp_s/JsonParser.cpp
gibber9809: We can, and it probably makes sense to move it there -- I was just trying to avoid some refactoring. I'll update the code.

It's a little bit awkward, since it makes it so we need pass around paths + canonical representations in the resulting metadata for the compression flow, but not too bad.

PR #2033, opened 2026-02-25
feat(clp-s): Add `get_metadata_for_log_event` function to `clp_s::ArchiveReader` (resolves #2012).

### LINKED ISSUE #2012: `ArchiveReader::get_metadata_for_log_event` API
### Request

**Background**

At Presto, there's a need to project file-level (IR) metadata such as `$_filename` in a given archive, this is for the sake of metadata projection.  Given that `$_filename` is not part of the log record schema — it lives in the archive range index. Rather than implementing `$_filename` as a schema column in the log event (which we agreed doesn't make sense), we'd like an API on `ArchiveReader` to resolve file-level metadata using `log_event_idx`, this would allow us to extract `$_filename`, as well as other file-level metadata preserved in the archive.

**Timeline:**
We'd need this by early the week after next. No rush to merge into main — the snapshot branch works for now if that's easier.

### Possible implementation

Subject to implementer's decision

gibber9809:
# Description

This PR implements the `get_metadata_for_log_event` function as requested in #2012. The implementation is simply to create a map of `end_index -> metadata` containing entries for each range with at least one record, and look up the metadata for a given index using the `upper_bound()` function.

This PR also adds some assertions in `test-clp_s-range-index` to confirm that `get_metadata_for_log_event` works as expected.

# Checklist

  breaking change.

# Validation performed

* Added tests to validate that `get_metadata_for_log_event` retrieves the correct metadata for a given index, and throws for invalid indexes.

## Summary by CodeRabbit

* **New Features**
  * Added ability to retrieve metadata for individual log events, exposed as JSON.

* **Tests**
  * Added tests validating per-log-event metadata retrieval, including boundary and invalid-index error handling.

### DIFF
--- a/components/core/.clang-format
+++ b/components/core/.clang-format
@@ -7,7 +7,7 @@ IncludeCategories:
   # NOTE: clang-format retains leading white-space on a line in violation of the YAML spec.
   - Regex: "<(absl|antlr4|archive|boost|bsoncxx|catch2|curl|date|fmt|log_surgeon|lzma\
 |mongocxx|msgpack|mysql|nlohmann|openssl|outcome|regex_utils|simdjson|spdlog|sqlite3|string_utils\
-|yaml-cpp|ystdlib|zstd)"
+|xxhash|yaml-cpp|ystdlib|zstd)"
     Priority: 3
 
   # Project library headers
--- a/components/core/CMakeLists.txt
+++ b/components/core/CMakeLists.txt
@@ -329,6 +329,13 @@ if(CLP_NEED_SQLITE)
     FindDynamicLibraryDependencies(sqlite "${sqlite_DYNAMIC_LIBS}")
 endif()
 
+if(CLP_NEED_XXHASH)
+    find_package(xxHash 0.8.3 CONFIG REQUIRED)
+    if(xxHash_FOUND)
+        message(STATUS "Found xxHash ${xxHash_VERSION}")
+    endif()
+endif()
+
 add_subdirectory(src/clp/regex_utils)
 add_subdirectory(src/clp/string_utils)
 
@@ -359,6 +366,8 @@ set(SOURCE_FILES_clp_s_unitTest
     src/clp_s/FileReader.hpp
     src/clp_s/FileWriter.cpp
     src/clp_s/FileWriter.hpp
+    src/clp_s/filter/tests/test-clp_s-bloom_filter.cpp
+    src/clp_s/filter/tests/test-clp_s-xxhash.cpp
     src/clp_s/FloatFormatEncoding.cpp
     src/clp_s/FloatFormatEncoding.hpp
     src/clp_s/ffi/sfa/ClpArchiveReader.cpp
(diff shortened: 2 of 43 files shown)


components/core/src/clp_s/ArchiveReader.hpp
```diff
@@ -182,6 +183,16 @@ class ArchiveReader {
         return get_header().has_deprecated_timestamp_format();
     }
 
+    /**
+     * @param log_event_idx
+     * @return The file-level metadata associated with the record at `log_event_idx`.
+     * @throws ArchiveReaderAdaptor::OperationFailed when `log_event_idx` cannot be mapped to
+     * any metadata.
+     */
+    [[nodiscard]] auto get_metadata_for_log_event(int64_t log_event_idx) -> nlohmann::json const& {
```
Bill-hbrhbr, line 192: Should log event simply take an unsigned type like `uint64_t`?

components/core/src/clp_s/ArchiveReaderAdaptor.cpp
```diff
@@ -362,4 +365,13 @@ void ArchiveReaderAdaptor::checkin_reader_for_section(std::string_view section)
 
     m_current_reader_holder.reset();
 }
+
+auto ArchiveReaderAdaptor::get_metadata_for_log_event(int64_t log_event_idx)
+        -> nlohmann::json const& {
+    auto const it{m_non_empty_range_metadata_map.upper_bound(log_event_idx)};
+    if (m_non_empty_range_metadata_map.end() == it || log_event_idx < 0) {
```
Bill-hbrhbr, line 372: Logic seems like you don't really need to look up anything if the log even idx is negative, which makes me feel more strongly about using `uint64_t`

components/core/src/clp_s/ArchiveReaderAdaptor.cpp
```diff
@@ -168,6 +168,9 @@ auto ArchiveReaderAdaptor::try_read_range_index(ZstdDecompressor& decompressor,
                 end_index,
                 std::move(range_index_entry.at(RangeIndexWriter::cMetadataFieldsName))
         );
+        if (start_index != end_index) {
+            m_non_empty_range_metadata_map.emplace(end_index, m_range_index.back().fields);
+        }
```
Bill-hbrhbr, line 173: I guess I've always had the question if the range index accepts files with empty ranges, or gaps between adjacent ranges.

components/core/src/clp_s/ArchiveReader.hpp
gibber9809: It can't since `log_event_idx` is `int64_t`.

components/core/src/clp_s/ArchiveReaderAdaptor.cpp
gibber9809: Ditto per other comment

components/core/src/clp_s/ArchiveReaderAdaptor.cpp
gibber9809: Yeah we allow ranges where `start_index` == `end_index` to represent empty files, and those are excluded from this map since this map is for an API to look up the file-level metadata for a specific log record.

PR #2034, opened 2026-02-25
feat(clp-s): Add initial `clp_s::ffi::sfa::ArchiveReader` implementation with typed error codes and unit test coverage.

Bill-hbrhbr:
# Description

This PR introduces an initial `clp_s` SFA FFI surface for opening and inspecting single file archives from either filesystem paths or in memory buffers. The new `ClpArchiveReader` is a thin wrapper around `clp_s::ArchiveReader` that is designed to handle single file archives directly in C++, bypassing the CLI `clp-s` tool, while also providing a clean surface for FFI frontends such as JavaScript and Python.

Currently, the API exposes only basic archive metadata such as `archive_id` and the total number of log events. Corresponding C++ unit tests have been added to validate both path-based and in memory SFA opening flows against fixture archives.

The PR also introduces an SFA specific error code domain and adds build wiring for a standalone `clp_s_ffi_sfa` target.

# Checklist

  breaking change.

# Validation performed

* [X] Unit tests pass and downstream clp-ffi-js compiles and passes.

## Summary by CodeRabbit

* **New Features**
  * Added an FFI-facing single-file archive reader with file and in-memory creation, safe teardown, and metadata queries (archive ID, event count).
  * Introduced SFA-specific error codes with user-facing messages for clearer error reporting.
* **Tests**
  * Added integration tests validating disk and in-memory archive reads, event counts, and archive IDs.
* **Chores**
  * Build configuration updated to include the new FFI archive reader component.

### DIFF
--- a/components/api-server/src/client.rs
+++ b/components/api-server/src/client.rs
@@ -3,12 +3,9 @@ use std::pin::Pin;
 use async_stream::stream;
 use clp_rust_utils::{
     aws::AWS_DEFAULT_REGION,
-    clp_config::{
-        AwsAuthentication,
-        package::{
-            config::{Config, StorageEngine, StreamOutputStorage},
-            credentials::Credentials,
-        },
+    clp_config::package::{
+        config::{Config, StorageEngine, StreamOutputStorage},
+        credentials::Credentials,
     },
     database::mysql::create_clp_db_mysql_pool,
     job_config::{QUERY_JOBS_TABLE_NAME, QueryJobStatus, QueryJobType, SearchJobConfig},
@@ -374,8 +371,6 @@ impl Client {
             unreachable!();
         };
 
-        let AwsAuthentication::Credentials { credentials } = &s3_config.aws_authentication;
-
         let s3_config = s3_config.clone();
         if s3_config.region_code.is_none() && s3_config.endpoint_url.is_none() {
             return Err(ClientError::Aws {
@@ -384,15 +379,14 @@ impl Client {
             });
         }
 
-        let credentials = credentials.clone();
+        let region_str = s3_config
+            .region_code
+            .as_ref()
+            .map_or(AWS_DEFAULT_REGION, non_empty_string::NonEmptyString::as_str);
         let s3_client = clp_rust_utils::s3::create_new_client(
-            credentials.access_key_id.as_str(),
-            credentials.secret_access_key.as_str(),
-            s3_config
-                .region_code
-                .as_ref()
-                .map_or(AWS_DEFAULT_REGION, non_empty_string::NonEmptyString::as_str),
+            region_str,
             s3_config.endpoint_url.as_ref(),
+            &s3_config.aws_authentication,
         )
         .await;
 
--- a/components/clp-py-utils/clp_py_utils/clp_config.py
+++ b/components/clp-py-utils/clp_py_utils/clp_config.py
@@ -165,10 +165,10 @@ class StorageType(LowercaseStrEnum):
 
 
 class AwsAuthType(LowercaseStrEnum):
+    default = auto()
     credentials = auto()
     profile = auto()
     env_vars = auto()
-    ec2 = auto()
 
 
 AwsAuthTypeStr = Annotated[AwsAuthType, StrEnumSerializer]
@@ -588,7 +588,7 @@ class AwsAuthentication(BaseModel):
             raise ValueError(f"profile must be set when type is '{auth_enum}.'")
         if AwsAuthType.credentials == auth_enum and not credentials:
             raise ValueError(f"credentials must be set when type is '{auth_enum}.'")
-        if auth_enum in [AwsAuthType.ec2, AwsAuthType.env_vars] and (profile or credentials):
+        if auth_enum in [AwsAuthType.default, AwsAuthType.env_vars] and (profile or credentials):
             raise ValueError(f"profile and credentials must not be set when type is '{auth_enum}.'")
         return data
 
@@ -951,7 +951,6 @@ class ClpConfig(BaseModel):
             raise ValueError(f"tmp_directory is invalid: {ex}")
 
     def validate_aws_config_dir(self, use_host_mount: bool = False):
-        profile_auth_used = False
         auth_configs = []
 
         if StorageType.S3 == self.logs_input.type:
@@ -961,15 +960,22 @@ class ClpConfig(BaseModel):
         if StorageType.S3 == self.stream_output.storage.type:
             auth_configs.append(self.stream_output.storage.s3_config.aws_authentication)
 
-        for auth in auth_configs:
-            if AwsAuthType.profile == auth.type:
-                profile_auth_used = True
-                break
+        auth_types_used = {auth.type for auth in auth_configs}
+        default_auth_used = AwsAuthType.default in auth_types_used
+        profile_auth_used = AwsAuthType.profile in auth_types_used
+        config_dir_allowed = profile_auth_used or default_auth_used
 
         if profile_auth_used:
             if self.aws_config_directory is None:
                 raise ValueError(
                     "aws_config_directory must be set when using profile authentication"
+                )
+
+        if self.aws_config_directory is not None:
+            if not config_dir_allowed:
+                raise ValueError(
+                    "aws_config_directory is only supported with 'profile' or 'default'"
+                    " authentication"
                 )
             resolved_aws_config_dir = (
                 resolve_host_path_in_container(self.aws_config_directory)
@@ -980,10 +986,6 @@ class ClpConfig(BaseModel):
                 raise ValueError(
                     f"aws_config_directory does not exist: '{self.aws_config_directory}'"
                 )
-        if not profile_auth_used and self.aws_config_directory is not None:
-            raise ValueError(
-                "aws_config_directory should not be set when profile authentication is not used"
-            )
 
     def validate_api_server(self):
         if StorageEngine.CLP == self.package.storage_engine and self.api_server is not None:
--- a/components/clp-py-utils/clp_py_utils/s3_utils.py
+++ b/components/clp-py-utils/clp_py_utils/s3_utils.py
@@ -90,10 +90,10 @@ def get_credential_env_vars(auth: AwsAuthentication) -> dict[str, str]:
         if aws_credentials is None:
             raise ValueError(f"Failed to authenticate with profile {auth.profile}")
 
-    elif AwsAuthType.ec2 == auth.type:
+    elif AwsAuthType.default == auth.type:
         aws_credentials = _get_session_credentials()
         if aws_credentials is None:
-            raise ValueError("Failed to authenticate with EC2 metadata.")
+            raise ValueError("Failed to authenticate with the default credential provider chain.")
     else:
         raise ValueError(f"Unsupported authentication type: {auth.type}")
 
@@ -206,7 +206,7 @@ def _create_s3_client(
             region_name=region_code,
             aws_session_token=credentials.session_token,
         )
-    elif AwsAuthType.env_vars == s3_auth.type or AwsAuthType.ec2 == s3_auth.type:
+    elif AwsAuthType.env_vars == s3_auth.type or AwsAuthType.default == s3_auth.type:
         # Use default session which will use environment variables or instance role
         aws_session = boto3.Session(region_name=region_code)
     else:
--- a/components/clp-rust-utils/src/clp_config/package/config.rs
+++ b/components/clp-rust-utils/src/clp_config/package/config.rs
@@ -345,7 +345,34 @@ mod tests {
                     assert_eq!(credentials.access_key_id, ACCESS_KEY_ID);
                     assert_eq!(credentials.secret_access_key, SECRET_ACCESS_KEY);
                 }
+                crate::clp_config::AwsAuthentication::Default => {
+                    panic!("Expected credentials, got `default`")
+                }
             },
+            LogsInput::Fs { .. } => panic!("Expected S3"),
+        }
+    }
+
+    #[test]
+    fn deserialize_logs_input_s3_default_config() {
+        let logs_input_config_json = serde_json::json!({
+            "type": "s3",
+            "aws_authentication": {
+                "type": "default",
+            }
+        });
+
+        let deserialized =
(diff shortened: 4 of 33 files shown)


components/core/src/clp_s/ffi/sfa/ClpArchiveReader.hpp
```diff
@@ -0,0 +1,92 @@
...
+    // Default move constructor and assignment operator
+    ClpArchiveReader(ClpArchiveReader&&) = default;
+    [[nodiscard]] auto operator=(ClpArchiveReader&&) -> ClpArchiveReader& = default;
+
+    /**
+     * @return A result containing the archive ID on success, or an error code indicating the
+     * failure:
+     * - `SfaErrorCodeEnum::NotInit` if the reader is not initialized or has already been closed.
+     */
+    [[nodiscard]] auto get_archive_id() const -> ystdlib::error_handling::Result<std::string>;
```
junhaoliao: we briefly discussed this previously on a call but i still don't get why we wanna add this method. like why would a caller wants the archive ID?

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+auto ClpArchiveReader::get_event_count() const -> Result<uint64_t> {
+    if (nullptr == m_archive_reader) {
+        return SfaErrorCode{SfaErrorCodeEnum::NotInit};
+    }
+
+    uint64_t event_count{0};
+    for (auto const& range : m_archive_reader->get_range_index()) {
+        event_count += static_cast<uint64_t>(range.end_index - range.start_index);
+    }
```
junhaoliao: * can we consider caching the count?
* shall we move this into the factory function?

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+        return ClpArchiveReader{std::move(archive_reader), std::move(archive_data_owner)};
+    } catch (std::exception const& ex) {
+        SPDLOG_ERROR("Exception while creating ClpArchiveReader: {}", ex.what());
+        return SfaErrorCode{SfaErrorCodeEnum::IoFailure};
+    }
+}
+
+ClpArchiveReader::~ClpArchiveReader() {
+    if (nullptr == m_archive_reader) {
```
junhaoliao: i'm not sure if we just wanna make the linter happy here, but the null check doesn't seem necessary if the archive reader is guaranteed to be assigned in the factory function, and nobody else would `move()` it?

components/core/src/clp_s/ffi/sfa/SfaErrorCode.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include <cstdint>
+
+#include <ystdlib/error_handling/ErrorCode.hpp>
+
+namespace clp_s::ffi::sfa {
+/**
+ * Error code enum for SFA API operations.
+ */
+enum class SfaErrorCodeEnum : uint8_t {
+    EndOfStream,
```
junhaoliao: note to self: this guy seems unused

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <string>
+#include <string_view>
+#include <utility>
+#include <vector>
+
+#include <spdlog/spdlog.h>
+#include <ystdlib/error_handling/Result.hpp>
+
+#include <clp/BufferReader.hpp>
+#include <clp_s/ArchiveReader.hpp>
```
junhaoliao, line 15: do you think we have to `#include <clp_s/ArchiveReader.hpp>`? instead, if we simply write `namespace clp_s { class ArchiveReader; }`, we can avoid pulling other header `#include`s from clp_s/ArchiveReader.hpp which helps depedency setups in FFI use cases in general?

junhaoliao (changes requested):
this is the initial implementation which matches the full picture imo. the rest lgtm

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
Bill-hbrhbr: - `ClpArchiveReader.hpp` now forward declares `clp_s::ArchiveReader` instead of including
`clp_s/ArchiveReader.hpp`.
- The actual include stays in `ClpArchiveReader.cpp`.
- Since `ClpArchiveReader` holds a `std::unique_ptr<clp_s::ArchiveReader>`, I moved the move constructors into .cpp  so incomplete type handling remains safe.

Result:
- Improves dependency hygiene and reduces build coupling for FFI consumers by avoiding
unnecessary transitive includes.
- Binary size impact should be negligible. This is mainly about improving the compile and
dependency surface rather than optimizing code size.

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
```diff
@@ -0,0 +1,142 @@
...
+)
+        : m_archive_reader{std::move(reader)},
+          m_archive_data{std::move(archive_data)},
+          m_event_count{event_count} {}
+
+// Default move constructor
+ClpArchiveReader::ClpArchiveReader(ClpArchiveReader&&) noexcept = default;
+
+// Default move assignment operator
+auto ClpArchiveReader::operator=(ClpArchiveReader&&) noexcept -> ClpArchiveReader& = default;
```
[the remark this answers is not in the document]
Bill-hbrhbr: review again. does issue still exist

components/core/tests/test-clp_s-ffi_sfa_reader.cpp
```diff
@@ -0,0 +1,103 @@
...
+}  // namespace
+
+TEST_CASE("clp_s_ffi_sfa_reader_matches_test_no_floats_sorted", "[clp-s][ffi][sfa]") {
+    assert_archive_event_count_matches_log(cArchiveNoFloats);
+    assert_archive_event_count_matches_log_in_memory(cArchiveNoFloats);
+}
+
+TEST_CASE("clp_s_ffi_sfa_reader_matches_test_search_float_timestamp", "[clp-s][ffi][sfa]") {
+    assert_archive_event_count_matches_log(cArchiveFloatTimestamp);
+    assert_archive_event_count_matches_log_in_memory(cArchiveFloatTimestamp);
```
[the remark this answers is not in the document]
Bill-hbrhbr: it's not essential for now. Unit tests are for sanity checking that `ClpArchiveReader` works under normal operations.

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
```diff
@@ -0,0 +1,130 @@
...
+                archive_data_owner->size()
+        )};
+
+        archive_reader = std::make_unique<clp_s::ArchiveReader>();
+        archive_reader->open(reader, cDefaultArchiveId);
+        return ClpArchiveReader{std::move(archive_reader), std::move(archive_data_owner)};
+    } catch (std::bad_alloc const&) {
+        SPDLOG_ERROR(
+                "Failed to create ClpArchiveReader for archive {}: out of memory.",
+                cDefaultArchiveId
```
junhaoliao:
```suggestion
                "Failed to create ClpArchiveReader: out of memory."
```

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
```diff
@@ -0,0 +1,130 @@
...
+        SPDLOG_ERROR("Exception while creating ClpArchiveReader: {}", ex.what());
+        return SfaErrorCode{SfaErrorCodeEnum::IoFailure};
+    }
+}
+
+auto ClpArchiveReader::create_from_bytes(std::span<char const> archive_data)
+        -> Result<ClpArchiveReader> {
+    std::unique_ptr<clp_s::ArchiveReader> archive_reader;
+    std::shared_ptr<std::vector<char>> archive_data_owner;
+    constexpr std::string_view cDefaultArchiveId{"default"};
```
junhaoliao:
```suggestion
    // `clp_s::ArchiveReader` requires an archive ID, but
    // `clp_s::ffi::sfa::ClpArchiveReader` never uses it. Provide a dummy value
    // solely to satisfy the constructor.
    constexpr std::string_view cDefaultArchiveId{"default"};
```

junhaoliao (changes requested):
then it should be good to go

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
```diff
@@ -0,0 +1,132 @@
...
+        -> Result<ClpArchiveReader> {
+    std::unique_ptr<clp_s::ArchiveReader> archive_reader;
+    std::shared_ptr<std::vector<char>> archive_data_owner;
+    // `clp_s::ArchiveReader` requires an archive ID, but
+    // `clp_s::ffi::sfa::ClpArchiveReader` never uses it. Provide a dummy value
+    // solely to satisfy the constructor.
+    constexpr std::string_view cDefaultArchiveId{"default"};
+
+    try {
+        archive_data_owner
```
hoophalab: Why do we copy the bytes into a vector? If we would like to take the ownership of the underlying data, then we shouldn't take span as the argument. It's best to use span::data and size directly here

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
```diff
@@ -0,0 +1,132 @@
...
+                archive_data_owner->size()
+        )};
+
+        archive_reader = std::make_unique<clp_s::ArchiveReader>();
+        archive_reader->open(reader, cDefaultArchiveId);
+        return ClpArchiveReader{std::move(archive_reader), std::move(archive_data_owner)};
+    } catch (std::bad_alloc const&) {
+        SPDLOG_ERROR(
+                "Failed to create ClpArchiveReader: out of memory."
+        );
```
junhaoliao: to fix lint
```suggestion
        SPDLOG_ERROR("Failed to create ClpArchiveReader: out of memory.");
```

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
```diff
@@ -0,0 +1,132 @@
...
+ClpArchiveReader::ClpArchiveReader(ClpArchiveReader&& rhs) noexcept {
+    move_from(rhs);
+}
+
+auto ClpArchiveReader::operator=(ClpArchiveReader&& rhs) noexcept -> ClpArchiveReader& {
+    if (this == &rhs) {
+        return *this;
+    }
+
+    close();
```
hoophalab, line 88: Why do we `close()` the archive? Shouldn't `operator=` do the same thing as `ClpArchiveReader(ClpArchiveReader&& rhs)`, and the free of `lhs`'s resources is done in the destructor?

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
```diff
@@ -0,0 +1,132 @@
...
+    m_event_count = std::exchange(rhs.m_event_count, 0);
+}
+
+auto ClpArchiveReader::precompute_archive_metadata() -> void {
+    auto const& range_index{m_archive_reader->get_range_index()};
+
+    for (auto const& range : range_index) {
+        auto const start_idx{static_cast<uint64_t>(range.start_index)};
+        auto const end_idx{static_cast<uint64_t>(range.end_index)};
+        m_event_count += end_idx - start_idx;
```
hoophalab: 1. I'm not familiar with archive format, but does this procedure really yield total event count in non-trivial cases? Might need to confirm with @gibber9809
 
2. Style: Is moving static cast after the subtraction more reasonable?
```suggestion
        m_event_count += static_cast<uint64_t>(range.end_index - range.start_index);
```

components/core/tests/test_single_file_archives/test_no_floats_sorted.clp
hoophalab, line 1: To be confirmed with @gibber9809: I'm not familiar with core's guidelines, but it doesn't look correct to include binary test files in repo. The existing unittests compress logs into temporary archive files during the test.

hoophalab (commented):
some comments

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
Bill-hbrhbr: I think in `lhs = std::move(rhs)`;, `lhs` remains the same object rather than becoming `rhs`, so `lhs`’s destructor is not called at assignment time. Need an explicit cleanup of resources previously held by `lhs` before it is overwritten.

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
Bill-hbrhbr: 2. sure but later in Junhao's [prototype](https://github.com/junhaoliao/clp-ffi-js/blob/cba10972cf4c149751913c355273050d87a4f801/src/clp_ffi_js/sfa/types.ts#L25) there would be explicit fields for start and end indices.

Bill-hbrhbr: 1. should be the case? Haven't encountered counter-examples yet.

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
Bill-hbrhbr: sure, I'll update the make the owning intention more explicit by passing std::vector&&

components/core/tests/test_single_file_archives/test_no_floats_sorted.clp
Bill-hbrhbr: I'll see if I can remove the need for committing binary files.

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.cpp
hoophalab: I see. Wrote too many rust and didn't realize Rust move is different from C++ move xD

PR #2035, opened 2026-02-25
feat(log-ingestor)!: Back ingestion jobs with CLP-DB–persisted states:

### LINKED ISSUE #1733: [log-ingestor] Improve ingestion job design
### Request

As introduced in #1704, we need to pass a callback to create and spawn an ingestion job to adapt different type of job creation. This design complicates things. We should avoid using the callback in a future refactoring.

### Possible implementation

Instead of creating and spawning ingestion jobs in the same method, we could split the creation and spawn so that:
- Ingestion jobs can be created in advance.
- We spawn the job after all validations passed.

In this way, no callback will be needed.
Splitting the spawn as a separate function may also help us handling failures: we don't need to create a new job instance but just need to re-spawn after log-ingestor detects and retries from a failure.

### LINKED ISSUE #1995: Update S3 scanner's shutdown_and_join to not return Result<()>
## Context

Following PR #1989, the `SqsListener::shutdown_and_join` method was updated to not return a `Result<()>` since with DB features added, there's no need to return an error in this call anymore.

For consistency, the S3 scanner's `shutdown_and_join` method should be updated similarly.

## References

- PR: https://github.com/y-scope/clp/pull/1989
- Comment: https://github.com/y-scope/clp/pull/1989#discussion_r2806894535
- Requested by: @LinZhihao-723

LinZhihao-723:
* Replace UUIDv4 job IDs with CLP-DB–generated IDs.
* Refactor `IngestionJobManager` APIs to formalize the distinction between ingestion jobs and job instances.
* Update `S3Scanner` and `Listener::shutdown_and_join` to no longer return errors (resolves #1995).
* Generalize job instance creation via `S3IngestionJobConfig` (resolves #1733).
* Mark ingestion jobs as `failed` when task coroutines return errors or panic.
* Refactor `CompressionJobSubmitter` to use `ClpCompressionState` for submission and result retrieval.
* Replace the job `DELETE` API with `POST {job_id}/terminate` to reflect termination semantics.
* Make job termination idempotent.

# Description

This PR is a part of the implementation of #1978, and it depends on #2038 and #2045.

This is a breaking change because:

* It's introducing new tables in CLP-DB.
* It changes log-ingestor's API.

This PR integrates `clp_ingestion.rs` to back ingestion jobs with CLP-DB-persisted states. With this PR, all ingestion jobs will have their ingested S3 object metadata persisted in CLP DB, as well as their job status.

However, this PR doesn't implement failure recovery yet. A future PR is needed to walk through the CLP DB to restart all ingestion job instances at log-ingestor initialization time.

The detailed changes are listed as follows:

* Replace UUIDv4 job IDs with CLP-DB–generated IDs since we use CLP DB to register ingestion jobs now.
* Rewrite job creation and termination APIs in `IngestionJobManager` to reflect the semantics changes in terms of ingestion job vs. ingestion job instance.
  * An ingestion job instance is a running instance hosted by the log-ingestor, while an ingestion job is a job registered in CLP-DB, and it's not necessarily running in log-ingestor.
  * As a part of this rewriting, we made the following changes:
    * Update `S3Scanner` and `Listener::shutdown_and_join` to no longer return errors (resolves #1995) to make the shutdown behavior consistent with `SqsListener`.
    * Generalize job instance creation via `S3IngestionJobConfig` (resolves #1733) so that the API is reusable for job recovery at restart.
* We now mark ingestion jobs `failed` if the underlying task coroutine fails. This failure will be reflected in CLP-DB with the error message uploaded.
* Don't log ingested S3 object metadata since they're now persisted in CLP-DB.
* Refactor `CompressionJobSubmitter` to make use of `ClpCompressionState` for compression. With this change,  all CLP DB interactions are grouped in `clp_ingestion` module.
* As discussed [here](https://github.com/y-scope/clp/pull/2035#discussion_r2857210014), we replace the job `DELETE` API with `POST {job_id}/terminate` to reflect termination semantics.
  * This new terminate API is idempotent.

# Checklist

  breaking change.

# Validation performed

* [x] Ensure ingestion jobs can be properly terminated (and idempotent).
* [x] Ensure coroutine failure will be forwarded to CLP DB with the job marked as failed. All tasks will terminate if a failure happens.
* [x] Proofread OpenAPI doc and ensure everything is properly rendered in SwaggerUI.
* [x] Ensure S3 scanner and SQS listener test works using [this](https://github.com/LinZhihao-723/LogIngestorServer/blob/main/scripts/generate_and_upload.py) script.
* [x] Check SQS listener performance with the following setup. Ensure ingestion speed is real-time (SQS messages are consumed as fast as they're generated).
  * 8 SQS listener jobs, each listens to an SQS queue with 15000 messages per minute.
  * 16 tasks per listener job.
  

## Summary by CodeRabbit

* **New Features**
  * Creation and termination endpoints now use numeric job IDs and return terminal status (Finished/Failed).

* **Bug Fixes**
  * Shutdown paths no longer propagate internal errors; outcomes are logged and failures signalled without panics.

* **Refactor**
  * Unified job identifier across APIs; job registration/creation flow reworked and compression submission moved to a state-driven flow; listener/scanner APIs updated to accept composite state bounds.

* **Tests**
  * Tests updated to use the numeric job ID and adjusted shutdown handling.


components/log-ingestor/src/compression/compression_job_submitter.rs
```diff
@@ -15,33 +15,39 @@ use clp_rust_utils::{
...
+            .collect();
+        let io_config_template = self.io_config_template.clone();
+        let state = self.state.clone();
         tokio::spawn(submit_clp_compression_job_and_wait_for_completion(
-            self.db_pool.clone(),
-            io_config,
+            state,
+            io_config_template,
+            id_and_key_pairs,
         ));
```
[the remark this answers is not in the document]
LinZhihao-723: In this case, the log should be sufficient.

components/log-ingestor/src/compression/compression_job_submitter.rs
```diff
@@ -106,91 +112,72 @@ impl CompressionJobSubmitter {
...
+            compression_job_id = ? compression_job_id,
+            compression_job_status = ? compression_job_status,
+            compression_job_status_msg = ? message,
+            "Compression job failed."
+        ),
+        _ => unreachable!(
+            "Unknown compression job status: {:?}",
+            compression_job_status
+        ),
     }
```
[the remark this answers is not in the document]
LinZhihao-723: It's ok. These errors should be caught in testing.

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -103,7 +89,7 @@ impl IngestionJobManagerState {
         let mysql_pool = clp_rust_utils::database::mysql::create_clp_db_mysql_pool(
             &clp_config.database,
             &clp_credentials.database,
-            10,
+            100,
         )
         .await?;
```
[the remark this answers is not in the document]
LinZhihao-723: Help me create an issue to track this.
We will use a hardcoded value in this PR, but we want this to be a configurable value in log-ingestor's config.

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -112,122 +98,97 @@ impl IngestionJobManagerState {
...
-                entry.ingestion_job.shutdown_and_join().await?;
-                tracing::debug!("Ingestion job {} shut down.", job_id);
-                entry.listener.shutdown_and_join().await?;
-                tracing::debug!("Ingestion job {}'s listener shut down.", job_id);
-                Ok(())
+                Ok(false)
             }
             None => Err(Error::JobNotFound(job_id)),
         }
     }
```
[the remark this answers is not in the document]
LinZhihao-723: This method should return as long as the instance has been removed.

components/log-ingestor/src/routes.rs
```diff
@@ -219,42 +236,37 @@ async fn create_sqs_listener_job(
...
-    path = "/job/{job_id}",
+    post,
+    path = "/job/{job_id}/terminate",
     tags = ["IngestionJob"],
-    description = "Deletes an existing ingestion job by its ID. This operation stops the job if it \
-        is currently running and removes all associated resources.",
+    description = "Terminates the ingestion job instance identified by the given job ID.\n\n\
+        This operation only terminates the job instance; the job record, including its status and \
+        statistics, is preserved and can still be accessible using the same job ID.\n\n\
+        This operation is idempotent.",
```
LinZhihao-723, line 256: @junhaoliao Since we don't actually delete the job but terminate it, do you think this change makes sense?
It's no longer a `delete` method.
We may implement `/job/{job_id}/pause` and `/job/{job_id}/restart` in the future.

junhaoliao: i think `POST` is most appropriate as you have currently written.

`DELETE` will remove resources -> i don't think we're removing the job record here.

`PUT` can be used for replacement of the whole thing -> it would make sense if the url is like /job/{job_id}/state

then i think `POST`, to modify parts of a resource, is most appropriate here

LinZhihao-723: Thx. Let's use `POST` then.

components/log-ingestor/src/routes.rs
sitaowang1998, line 160: I assume it is the responsibility of the user to ensure immutability.
```suggestion
       2. Immutability: Objects under the specified prefix are immutable. Once an object is \
       created, it should not be modified or overwritten.",
```

LinZhihao-723: Hmmm, this section is under the tone of assumptions, so I think using "it is not" (which is like what we expect) is more declarative than "it should not" (which is like what we enforce).

PR #2036, opened 2026-02-25
fix(core): Avoid mixing iterators and raw pointers in `std::string_view` usages to ensure portable builds.

Bill-hbrhbr:
# Description

Emscripten’s libc++ implements `std::string_view::const_iterator` as a wrapped iterator type rather than a raw pointer. Code that implicitly converts iterators to pointers can therefore fail to compile under the Emscripten/wasm32 toolchain.
This change uses `data()` when interacting with APIs that require raw pointers and avoids implicit iterator to pointer assumptions. 

# Checklist

  breaking change.

# Validation performed

* [X] Works with clp-ffi-js.

## Summary by CodeRabbit

* **Refactor**
  * Improved robustness of search, wildcard handling, query tokenization, string utilities and schema serialization to reduce parsing/memory-access issues and increase stability.
* **New Features**
  * Added public schema data accessors to safely expose underlying schema storage for direct access.

### DIFF
--- a/components/log-ingestor/Cargo.toml
+++ b/components/log-ingestor/Cargo.toml
@@ -35,10 +35,10 @@ tokio = { version = "1.49.0", features = ["macros", "rt-multi-thread", "signal",
 tokio-util = "0.7.18"
 tracing = "0.1.44"
 tower-http = { version = "0.6.8", features = ["cors"] }
-uuid = { version = "1.20.0", features = ["v4"] }
 thiserror = "2.0.18"
 utoipa = { version = "5.4.0", features = ["axum_extras"] }
 utoipa-axum = "0.2.0"
 
 [dev-dependencies]
 serial_test = "3.3.1"
+uuid = { version = "1.20.0", features = ["v4"] }
--- a/components/log-ingestor/src/compression/compression_job_submitter.rs
+++ b/components/log-ingestor/src/compression/compression_job_submitter.rs
@@ -15,33 +15,39 @@ use clp_rust_utils::{
         S3InputConfig,
         ingestion::s3::BaseConfig,
     },
-    s3::ObjectMetadata,
+    s3::{ObjectMetadata, S3ObjectMetadataId},
 };
-use const_format::formatcp;
-use sqlx::MySqlPool;
+use non_empty_string::NonEmptyString;
 
-use crate::compression::BufferSubmitter;
+use crate::{compression::BufferSubmitter, ingestion_job_manager::ClpCompressionState};
 
 /// The CLP compression job table name.
 pub const CLP_COMPRESSION_JOB_TABLE_NAME: &str = "compression_jobs";
 
 /// A compression submitter that implements [`BufferSubmitter`] to submit compression jobs to CLP.
 pub struct CompressionJobSubmitter {
-    db_pool: MySqlPool,
     io_config_template: ClpIoConfig,
+    state: ClpCompressionState,
 }
 
 #[async_trait]
 impl BufferSubmitter for CompressionJobSubmitter {
     async fn submit(&self, buffer: &[ObjectMetadata]) -> Result<()> {
-        let mut io_config = self.io_config_template.clone();
-        let s3_input_config = match &mut io_config.input {
-            InputConfig::S3InputConfig { config } => config,
-        };
-        s3_input_config.keys = Some(buffer.iter().map(|obj| obj.key.clone()).collect());
+        let id_and_key_pairs = buffer
+            .iter()
+            .map(|object_metadata| {
+                (
+                    object_metadata.id.expect("the ingestion ID must be set"),
+                    object_metadata.key.clone(),
+                )
+            })
+            .collect();
+        let io_config_template = self.io_config_template.clone();
+        let state = self.state.clone();
         tokio::spawn(submit_clp_compression_job_and_wait_for_completion(
-            self.db_pool.clone(),
-            io_config,
+            state,
+            io_config_template,
+            id_and_key_pairs,
         ));
         Ok(())
     }
@@ -53,10 +59,10 @@ impl CompressionJobSubmitter {
     /// # Returns
     ///
     /// A newly created instance with the given parameters, dedicated to submitting compression jobs
-    /// created by the given ingestion job specification.
+    /// using the given [`ClpCompressionState`] for its underlying ingestion job.
     #[must_use]
     pub fn new(
-        db_pool: MySqlPool,
+        clp_compression_state: ClpCompressionState,
         aws_credentials: AwsCredentials,
         archive_output_config: &ArchiveOutput,
         ingestion_job_config: &BaseConfig,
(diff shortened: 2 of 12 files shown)


components/core/src/clp/string_utils/string_utils.cpp
```diff
@@ -154,8 +154,8 @@ auto clean_up_wildcard_search_string(string_view str) -> string {
     string cleaned_str;
 
     bool is_escaped{false};
-    auto const* const str_end = str.cend();
-    for (auto const* current = str.cbegin(); current != str_end;) {
+    std::string_view::const_iterator const str_end = str.cend();
```
Bill-hbrhbr:
```suggestion
    std::string_view::const_iterator const str_end{str.cend()};
```

davidlion:
I could be wrong, but shouldn't most of these cases be using https://en.cppreference.com/w/cpp/string/basic_string_view/substr.html?

In general, we should be avoiding pointer arithmetic, so it would be ideal if there was a solution to these cases that didn't involve adding a position to `data()`.

Bill-hbrhbr:
> I could be wrong, but shouldn't most of these cases be using https://en.cppreference.com/w/cpp/string/basic_string_view/substr.html?
> 
> In general, we should be avoiding pointer arithmetic, so it would be ideal if there was a solution to these cases that didn't involve adding a position to `data()`.

You are right. Updated relevant code sections.

components/core/src/clp_s/SchemaTree.hpp
```diff
@@ -80,7 +80,7 @@ class SchemaNode {
               m_type(type),
               m_count(0),
               m_depth(depth) {
-        memcpy(m_key_name_buf.get(), key_name.begin(), key_name.size());
+        memcpy(m_key_name_buf.get(), key_name.data(), key_name.size());
```
junhaoliao, line 83: shall we use `std::memcpy()` and `#include <cstring>`? like why did we revert in 248a777632a1411a7c8f1d0e0867547513dd3184 ?

components/core/src/clp/string_utils/string_utils.cpp
```diff
@@ -154,8 +154,8 @@ auto clean_up_wildcard_search_string(string_view str) -> string {
     string cleaned_str;
 
     bool is_escaped{false};
-    auto const* const str_end = str.cend();
-    for (auto const* current = str.cbegin(); current != str_end;) {
+    std::string_view::const_iterator const str_end{str.cend()};
+    for (std::string_view::const_iterator current = str.cbegin(); current != str_end;) {
```
junhaoliao: would it be cleaner to rewrite this with `for (size_t i{0}; i < str.size();) {}` like index-based accesses? that way we can get rid of the `// NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)` suppresions too

components/core/src/clp_s/SchemaTree.hpp
Bill-hbrhbr: I think it's not creating an issue with downstream clp-ffi-js, and there more more memcpy and other cstring functions to replace than this one.

junhaoliao: i see. right, we can do it properly in a future PR. @coderabbitai create an issue to track

components/core/src/clp/ffi/search/CompositeWildcardToken.cpp
```diff
@@ -252,7 +252,7 @@ void CompositeWildcardToken<encoded_variable_t>::try_add_wildcard_variable(
                 end_pos
         );
     } else {
-        string_view var(m_query.cbegin() + begin_pos, end_pos - begin_pos);
+        auto const var = m_query.substr(begin_pos, end_pos - begin_pos);
```
junhaoliao, line 255:
```suggestion
        auto const var{m_query.substr(begin_pos, end_pos - begin_pos)};
```

components/core/src/clp/ffi/search/WildcardToken.cpp
```diff
@@ -152,7 +152,7 @@ WildcardToken<encoded_variable_t>::WildcardToken(
         : QueryToken(query, begin_pos, end_pos),
           m_has_prefix_star_wildcard('*' == query[begin_pos]),
           m_has_suffix_star_wildcard('*' == query[end_pos - 1]) {
-    auto token = string_view(query.cbegin() + begin_pos, end_pos - begin_pos);
+    auto const token = query.substr(begin_pos, end_pos - begin_pos);
```
junhaoliao, line 155:
```suggestion
    auto const token{query.substr(begin_pos, end_pos - begin_pos)};
```

components/core/src/clp/ffi/search/query_methods.cpp
```diff
@@ -205,7 +205,7 @@ void tokenize_query(
                 composite_wildcard_token_indexes.push_back(tokens.size() - 1);
             }
         } else {
-            string_view variable(wildcard_query.cbegin() + begin_pos, end_pos - begin_pos);
+            auto const variable = wildcard_query.substr(begin_pos, end_pos - begin_pos);
```
junhaoliao, line 208:
```suggestion
            auto const variable{wildcard_query.substr(begin_pos, end_pos - begin_pos)};
```

components/core/src/glt/ffi/search/CompositeWildcardToken.cpp
```diff
@@ -252,7 +252,7 @@ void CompositeWildcardToken<encoded_variable_t>::try_add_wildcard_variable(
                 end_pos
         );
     } else {
-        string_view var(m_query.cbegin() + begin_pos, end_pos - begin_pos);
+        auto const var = m_query.substr(begin_pos, end_pos - begin_pos);
```
junhaoliao, line 255:
```suggestion
        auto const var{m_query.substr(begin_pos, end_pos - begin_pos)};
```

components/core/src/glt/ffi/search/WildcardToken.cpp
```diff
@@ -152,7 +152,7 @@ WildcardToken<encoded_variable_t>::WildcardToken(
         : QueryToken(query, begin_pos, end_pos),
           m_has_prefix_star_wildcard('*' == query[begin_pos]),
           m_has_suffix_star_wildcard('*' == query[end_pos - 1]) {
-    auto token = string_view(query.cbegin() + begin_pos, end_pos - begin_pos);
+    auto const token = query.substr(begin_pos, end_pos - begin_pos);
```
junhaoliao, line 155:
```suggestion
    auto const token{query.substr(begin_pos, end_pos - begin_pos)};
```

components/core/src/glt/ffi/search/query_methods.cpp
```diff
@@ -205,7 +205,7 @@ void tokenize_query(
                 composite_wildcard_token_indexes.push_back(tokens.size() - 1);
             }
         } else {
-            string_view variable(wildcard_query.cbegin() + begin_pos, end_pos - begin_pos);
+            auto const variable = wildcard_query.substr(begin_pos, end_pos - begin_pos);
```
junhaoliao, line 208:
```suggestion
            auto const variable{wildcard_query.substr(begin_pos, end_pos - begin_pos)};
```

Bill-hbrhbr (commented):
https://www.notion.so/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?source=copy_link#907f63e5fe324d71a2b9003904f07f50

The brace init doesn't apply to auto variables per our guideline.

junhaoliao:
> [notion.so/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?source=copy_link#907f63e5fe324d71a2b9003904f07f50](https://www.notion.so/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?source=copy_link#907f63e5fe324d71a2b9003904f07f50)
> 
> The brace init doesn't apply to auto variables per our guideline.

right. if there's no harm in using the initializer list style (e.g., we don't want the variable type to be `std::initializer_list<int>` which i believe we have to write `auto x = {1, 2};`), i believe using `{}` would be better for consistency? anyways, i can see there's no benefits in using `{}` and given our guidelines allows the exception, whatever's been in the PR is fine to leave as-is then

Issue #2018, opened 2026-02-24
[log-ingestor] Improve compression job submission flow.


LinZhihao-723:
#2017 introduces DB operations to submit ingested S3 object metadata into a compression job, as a part of the implementation for #1978.

However, this is not a perfect compression job submission flow due to the following issues:

* In #2017, we changed `ObjectMetadata` to include an ID in the DB for compression job submission. The way of setting this ID is not legit as it mutates `ObjectMetadata` while the information duplicates what's been persisted in CLP DB.
* In fact, the root of inefficiency comes from how compression jobs are processed. Before we have a log-ingestor, the compression scheduler needs to access S3 to fetch the size information for compression task partitioning. However, with log-ingestor, this is no longer needed because the size information has been fetched during compression. Instead of referring to a list of keys for compression, it would be more efficient to directly submit a list of S3 object metadata ID since the compression scheduler can fetch the information needed from the DB without talking to the external S3 service.

Therefore, the following action items are required for the overall compression job flow efficiency improvements:

* Update the compression job config so that the S3 config can refer to the S3 object metadata table directly without sending any keys explicitly inside the config.
* Update the compression scheduler to handle the compression by:
  * Fetching relevant S3 object metadata
  * Process the metadata and partition for compression tasks
* Update log-ingestor's buffer and compression job submitter to receive S3 object metadata with an ID and a size only. The bucket and key information will no longer be needed once they're ingested into the CLP DB.
* Update log-ingestor's unit tests and component-wise integration tests so that the ingested files are buffered and checked directly through the ingestion job state implementation, and no buffering logic will be involved.

jonathan-imanu:
After some discussion with @LinZhihao-723 offline,  we’re opening the config design for input before implementing.

Here are some options we have come up with:

- Have an optional field with the list of S3 metadata IDs.
- Alternatively, we could make the keys field an enum which can take a list of keys or a list of S3 metadata IDs instead. This is to make sure keys and IDs can't be given at the same time.
- We could also leave the keys field totally empty. In this case, we'll attempt to read directly from the S3 metadata table to select all rows that has the compression job ID marked as the current one. This might be a more efficient way.

junhaoliao:
Sounds that we have two types of information that are exclusive to each other, or at least one set of the info would (/could?) be redundant while the other set presents.

> Alternatively, we could make the keys field an enum which can take a list of keys or a list of S3 metadata IDs instead. This is to make sure keys and IDs can't be given at the same time.

[...]

PR #2053, opened 2026-03-02
feat(log-ingestor): Add support for recovering `log-ingestor` on restart (resolves #1978).

### LINKED ISSUE #1978: [log-ingestor] Fault-tolerance support.

LinZhihao-723:
# Description

This PR is a part of the implementation for #1978, and it is the last PR to implement this feature to resolve #1978.

This PR adds recovery logic for log-ingestor on a restart. The restart procedure is as follows:

1. Recover all coroutines that should wait for a compression job to finish and thus update the status of submitted object metadata.
2. Recover all recoverable ingestion jobs. "Recoverable" here means we should recreate an ingestion job instance.
3. Recover all inactive ingestion jobs (including paused, failed, and finished ones). This recovery only refills the ingestion buffer, so already-ingested files can be submitted for compression.

| Ingestion job type | Recoverable | Inactive | Notes |
|----------|----------|----------|----------|
| Waiting for compression to finish  | Yes | Yes | By step 1 in general. |
| Refill ingestion buffer  | Yes  | Yes | |
| Create ingestion job insatance | Yes | No | |

To make ingestion job handling generic on creation and on restart, we introduce `ClpIngestionContext` and let `IngestionJobManager` operate on it:

* On success, the ingestion manager owns the ingestion context, which holds the listener as a single owner of the ingestion buffer.
* On failure, the context is dropped, so the listener will be closed. No resource leak.

This PR rewrites `IngestionJobManager` partially to make use of `ClpIngestionContext`.

# Checklist

  breaking change.

# Validation performed

* [x] Run the following test (for both MySQL and MariaDB):
  1. Start log-ingestor.
  2. Submit 8 SQS listener jobs at once.
    * Ensure they are all running.
  3. When the first compression job is created, restart log-ingestor.
    * Ensure all ingestion jobs are resumed.
  4. Wait until all files are ingested.
  5. Close 4 ingestion jobs.
  6. Restart log-ingestor.
    * Ensure closed jobs are not resumed.
    * Ensure the buffered files for the closed jobs are submitted.
    * Ensure the rest ingestion jobs are resumed.
  7. Ensure all files are in `compressed` status eventually.
    * This implies the buffer is recovered properly in steps 3 and 6.
    * This implies the waiting coroutines are successfully recovered in steps 3 and 6.
    * This implies that no compression jobs are submitted repeatedly.
* [x] Start an S3 ingestion job; restart if necessary multiple times during the file ingestion, and ensure all files are eventually ingested without duplication.

## Summary by CodeRabbit

* **Bug Fixes**
  * Listener channels now close gracefully instead of signaling an unexpected error.

* **New Features**
  * Startup recovery restores unfinished compression work, rebuilds in‑progress ingestion jobs, and refills buffered metadata so processing resumes after restarts.
  * Termination responses now report explicit terminal status (Finished/Failed) for clearer shutdown outcomes.

* **Chores**
  * Reduced duplicated wait-and-log logic and introduced job/context abstractions to improve lifecycle management and reliability.


components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -21,11 +28,134 @@ use strum_macros::{AsRefStr, Display, EnumIter, EnumString};
...
+    /// * `true` indicates the job ends in [`ClpIngestionJobStatus::Failed`] status.
+    /// * `false` indicates the job ends in [`ClpIngestionJobStatus::Finished`] status.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`ClpIngestionState::get_job_status`]'s return values on failure.
+    /// * Forwards [`ClpIngestionState::end`]'s return values on failure.
+    pub async fn shutdown(self) -> anyhow::Result<bool> {
```
sitaowang1998: Avoid returning `bool` for different states, even binary states.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -37,63 +167,86 @@ pub struct ClpDbIngestionConnector {
...
-            buffer_flush_threshold,
-        })
+            archive_output_config: clp_config.archive_output.clone(),
+            buffer_flush_timeout: Duration::from_secs(log_ingestor_config.buffer_flush_timeout_sec),
+            buffer_flush_threshold: log_ingestor_config.buffer_flush_threshold,
+        };
+
+        let unfinished_compression_jobs = connector.get_unfinished_compression_jobs().await?;
+        let (recoverable_ingestion_jobs, inactive_ingestion_jobs) =
+            connector.load_ingestion_jobs().await?;
```
sitaowang1998, line 232: I am a little worried about these two functions under failure. If one db connection fails, or config parse fails, we just stop halfway. Should we try our best to handle all jobs and only give up on failed ones?

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -87,30 +83,31 @@ impl IngestionJobManagerState {
...
+            recovery_context.recoverable_ingestion_jobs,
+            recovery_context.inactive_ingestion_jobs,
+        );
+        try_recover_waiting_coroutines_for_unfinished_compression_jobs(unfinished_compression_jobs);
+        try_recover_ingestion_job_instances(
+            ingestion_job_manager.clone(),
+            recoverable_ingestion_jobs,
+        )
+        .await;
+        try_recover_inactive_ingestion_jobs(inactive_ingestion_jobs).await;
```
sitaowang1998, line 118: Same concern for failing halfway.

components/log-ingestor/src/ingestion_job_manager.rs
```diff
@@ -369,3 +349,124 @@ struct IngestionJobTableEntry {
...
+) {
+    let mut num_recovered_jobs = 0;
+    for ingestion_job_context in recoverable_ingestion_jobs {
+        let job_id = ingestion_job_context.get_job_id();
+        tracing::info!(
+            job_id = ? job_id,
+            "Recovering ingestion job."
+        );
+
+        try_refill_ingestion_buffer(&ingestion_job_context).await;
```
sitaowang1998, line 402: So we continue creating job instance even if refill ingestion buffer fails?

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
LinZhihao-723: I think the current behavior is that if a connection error happens, we return the error directly to the top-level caller. And that would trigger another restart on the log-ingestor to retry everything.

components/log-ingestor/src/ingestion_job_manager.rs
LinZhihao-723: These functions are best-effort. It won't return any errors but only log them and keep proceeding. That's why it's called "try_xxx".
In the future, we plan expose APIs to retry a failed job explicitly.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
sitaowang1998: What if the error is not connection error?

components/log-ingestor/src/ingestion_job_manager.rs
LinZhihao-723: Yeah, I think it's fair to still allow the ingestion job to ingest more files into the system as:
- These new logs will likely be ingested into the buffer properly for compression.
- As planned for kv-ir, ingesting them into the system would allow early-stage IR-level search.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
LinZhihao-723: Yeah I also realized this is a bit ugly... how about returning https://github.com/y-scope/clp/blob/dcc105c574795fadfbb348a1fe119e8044e28a40/components/log-ingestor/src/routes.rs#L117?

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
LinZhihao-723: We only forward `fetch_all` errors which are from DB operations. Whether these errors are connection errors probably doesn't matter? The next round of retry would be triggered on a restart before the recovery procedure starts.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -206,6 +338,330 @@ impl ClpDbIngestionConnector {
...
+        &self,
+    ) -> anyhow::Result<Vec<ClpCompressionJobContext>> {
+        const QUERY: &str = formatcp!(
+            "SELECT `ingestion_job_id`, `compression_job_id`, COUNT(*) as `num_submitted` FROM \
+             `{table}` WHERE `compression_job_id` IS NOT NULL AND `status` = ? GROUP BY \
+             `compression_job_id` ORDER BY `compression_job_id` ASC;",
+            table = INGESTED_S3_OBJECT_METADATA_TABLE_NAME,
+        );
+        let all_rows: Vec<(IngestionJobId, CompressionJobId, i64)> = sqlx::query_as(QUERY)
+            .bind(IngestedS3ObjectMetadataStatus::Submitted)
```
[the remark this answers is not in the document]
LinZhihao-723: @sitaowang1998 Can u help check this?

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
sitaowang1998: I think it is fine for now. Maybe we can distinguish different errors later.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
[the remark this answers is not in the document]
sitaowang1998: This is valid, and make a lot of sense. If for each `compression_job_id`, there is only one `ingestion_job_id`, we can work around by either adding it to `GROUP BY` or by using `MIN/MAX` on `ingestion_job_id`.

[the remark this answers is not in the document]
LinZhihao-723: Yes, for now, each compression job should only have one ingestion job associated with it.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -248,6 +705,52 @@ impl ClpIngestionState {
...
+            object_metadata_vec.push(ObjectMetadata {
+                id: Some(id),
+                bucket,
+                key,
+                size,
+            });
+        }
+
+        Ok(object_metadata_vec)
+    }
```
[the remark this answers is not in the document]
LinZhihao-723: - This was recovered from a previous failure, which means everything could fit into memory before the failure.
- We can optimize it by batching in a later PR if really needed.
- We should touch this code soon once we implement #2018.

PR #2071, opened 2026-03-05
fix(clp-s): Replace uses of `size_t` with `uint64_t` in archive serialization; check that offsets are representable as `size_t` during archive deserialization (fixes #1955).

### LINKED ISSUE #1955: Incorrect use of `size_t` for 64-bit serialized metadata fields breaks CLP-S archives reads on 32-bit platforms.
### Bug

Several `clp_s` deserialization routines use `size_t` for reading numeric values from archives via `try_read_numeric_value()`. On 64-bit platforms this works correctly because `size_t` is 8 bytes, matching the serialized format. However, on wasm32 (Emscripten), `size_t` is 4 bytes, causing `try_read_numeric_value<size_t>()` to read only 4 bytes instead of 8, corrupting the stream position and producing incorrect values.

This affects any attempt to read CLP-S archives (created on 64-bit systems) from a WASM build.

**Affected locations:**

1. `PackedStreamReader.cpp:
   1. https://github.com/y-scope/clp/blob/670f0e43169992eee0fd1a6940a9d061d3289bc6/components/core/src/clp_s/PackedStreamReader.cpp#L17
   2. https://github.com/y-scope/clp/blob/670f0e43169992eee0fd1a6940a9d061d3289bc6/components/core/src/clp_s/PackedStreamReader.cpp#L24
   3. https://github.com/y-scope/clp/blob/670f0e43169992eee0fd1a6940a9d061d3289bc6/components/core/src/clp_s/PackedStreamReader.cpp#L25

2. `ReaderUtils.cpp`
   1. https://github.com/y-scope/clp/blob/670f0e43169992eee0fd1a6940a9d061d3289bc6/components/core/src/clp_s/ReaderUtils.cpp#L17
   2. https://github.com/y-scope/clp/blob/670f0e43169992eee0fd1a6940a9d061d3289bc6/components/core/src/clp_s/ReaderUtils.cpp#L26
   3. https://github.com/y-scope/clp/blob/670f0e43169992eee0fd1a6940a9d061d3289bc6/components/core/src/clp_s/ReaderUtils.cpp#L90

3. `ArchiveReader.cpp`: `num_separate_column_schemas`, `num_schemas`

4. `ColumnReader.cpp`: `encoded_vars_length`

5. `ArchiveReaderAdaptor.cpp`: `m_archive_header.compressed_size` (a `uint64_t`) is implicitly narrowed when assigned to `size_t next_file_offset` without a cast.

### CLP version

670f0e43169992eee0fd1a6940a9d061d3289bc6

### Environment

Emscripten (wasm32) tested with emsdk 4.0.3.

The issue affects any 32-bit platform where `sizeof(size_t) == 4`.

### Reproduction steps

1. Create a CLP-S single-file archive on a 64-bit system (e.g., Linux x86_64).
2. Build `clp_s` with Emscripten targeting wasm32.
3. Attempt to open and read the archive via `ArchiveReader`.
4. Observe corruption: metadata reads produce incorrect values because only 4 bytes are consumed from the stream where 8 were written. Subsequent reads are misaligned, causing cascading failures (e.g., "corrupt archive" errors, out-of-bounds reads, or incorrect record counts).

Bill-hbrhbr:
# Description

This PR fixes numeric width mismatches in CLP-S archive serialization and deserialization.

Previously, several metadata fields were serialized using host-sized types. Archives written on 64-bit systems could therefore contain values that do not fit `size_t` on narrower targets such as `wasm32`. This could lead to silent truncation during decode.

This PR standardizes serialized metadata fields on `uint64_t` and adds checked conversions when deserializing values that must be used as `size_t`.
- On the write path, metadata counts that are part of the archive format are now explicitly emitted as `uint64_t`.
- On the read path, deserialized `uint64_t` values are validated before conversion to `size_t`.
As a result, CLP-S archive decoding now fails safely when values exceed the platform `size_t` range instead of proceeding with truncated values.

Fixes #1955 

# Checklist

  breaking change.

# Validation performed

* [X] Works with clp-ffi-js prototypes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling during archive metadata reading to detect and report failures gracefully
  * Added bounds validation to prevent integer overflow when processing archive records
  * Improved consistency checks for archive structure validation to catch corruption early
  * Strengthened safety of numeric conversions during archive processing operations

### DIFF
--- a/components/core/src/clp_s/ArchiveReader.cpp
+++ b/components/core/src/clp_s/ArchiveReader.cpp
@@ -112,7 +112,7 @@ auto ArchiveReader::read_single_schema_metadata()
 
     return std::make_pair(
             schema_id,
-            SchemaReader::SchemaMetadata{stream_id, stream_offset, num_messages, 0}
+            SchemaReader::SchemaMetadata{stream_id, stream_offset, num_messages}
     );
 }
 
@@ -157,23 +157,27 @@ auto ArchiveReader::read_metadata() -> ystdlib::error_handling::Result<void> {
         };
         m_schema_ids.push_back(schema_id);
 
-        if (metadata.stream_id != prev_metadata.stream_id) {
-            prev_metadata.uncompressed_size
-                    = m_stream_reader.get_uncompressed_stream_size(prev_metadata.stream_id)
-                      - prev_metadata.stream_offset;
-        } else if (metadata.stream_offset < prev_metadata.stream_offset) {
+        if (metadata.stream_id() != prev_metadata.stream_id()) {
+            prev_metadata.set_uncompressed_size(
+                    m_stream_reader.get_uncompressed_stream_size(prev_metadata.stream_id())
+                    - prev_metadata.stream_offset()
+            );
+        } else if (metadata.stream_offset() < prev_metadata.stream_offset()) {
             throw OperationFailed(ErrorCodeCorrupt, __FILENAME__, __LINE__);
         } else {
-            prev_metadata.uncompressed_size = metadata.stream_offset - prev_metadata.stream_offset;
+            prev_metadata.set_uncompressed_size(
+                    metadata.stream_offset() - prev_metadata.stream_offset()
+            );
         }
         m_id_to_schema_metadata[prev_schema_id] = prev_metadata;
 
         prev_schema_id = schema_id;
         prev_metadata = metadata;
     }
-    prev_metadata.uncompressed_size
-            = m_stream_reader.get_uncompressed_stream_size(prev_metadata.stream_id)
-              - prev_metadata.stream_offset;
+    prev_metadata.set_uncompressed_size(
+            m_stream_reader.get_uncompressed_stream_size(prev_metadata.stream_id())
+            - prev_metadata.stream_offset()
+    );
     m_id_to_schema_metadata[prev_schema_id] = prev_metadata;
     m_table_metadata_decompressor.close();
 
@@ -211,10 +215,13 @@ SchemaReader& ArchiveReader::read_schema_table(
             should_marshal_records
     );
 
-    auto& schema_metadata = m_id_to_schema_metadata[schema_id];
-    auto stream_buffer = read_stream(schema_metadata.stream_id, true);
-    m_schema_reader
-            .load(stream_buffer, schema_metadata.stream_offset, schema_metadata.uncompressed_size);
+    auto const& schema_metadata = m_id_to_schema_metadata[schema_id];
+    auto stream_buffer = read_stream(schema_metadata.stream_id(), true);
+    m_schema_reader.load(
+            stream_buffer,
+            schema_metadata.stream_offset(),
+            schema_metadata.uncompressed_size()
+    );
     return m_schema_reader;
 }
 
@@ -224,12 +231,12 @@ std::vector<std::shared_ptr<SchemaReader>> ArchiveReader::read_all_tables() {
     for (auto schema_id : m_schema_ids) {
         auto schema_reader = std::make_shared<SchemaReader>();
         initialize_schema_reader(*schema_reader, schema_id, true, true);
-        auto& schema_metadata = m_id_to_schema_metadata[schema_id];
-        auto stream_buffer = read_stream(schema_metadata.stream_id, false);
+        auto const& schema_metadata = m_id_to_schema_metadata[schema_id];
+        auto stream_buffer = read_stream(schema_metadata.stream_id(), false);
         schema_reader->load(
                 stream_buffer,
-                schema_metadata.stream_offset,
-                schema_metadata.uncompressed_size
+                schema_metadata.stream_offset(),
+                schema_metadata.uncompressed_size()
         );
         readers.push_back(std::move(schema_reader));
     }
@@ -363,7 +370,7 @@ void ArchiveReader::initialize_schema_reader(
             m_projection,
             schema_id,
             schema.get_ordered_schema_view(),
-            m_id_to_schema_metadata[schema_id].num_messages,
+            m_id_to_schema_metadata[schema_id].num_messages(),
             should_marshal_records
     );
     auto timestamp_column_ids
--- a/components/core/src/clp_s/SchemaReader.hpp
+++ b/components/core/src/clp_s/SchemaReader.hpp
@@ -56,11 +56,50 @@ public:
                 : TraceableException(error_code, filename, line_number) {}
     };
 
-    struct SchemaMetadata {
-        size_t stream_id{0};
-        size_t stream_offset{0};
-        uint64_t num_messages{0};
-        size_t uncompressed_size{0};
+    /**
+     * Metadata describing one schema table entry.
+     *
+     * Fields (except `num_messages`) are stored as `size_t` instead of their serialized type
+     * `uint64_t` because they are passed to buffer and index APIs that operate on pointer sized
+     * offsets. Deserialized metadata must therefore be converted before constructing this object.
+     */
+    class SchemaMetadata {
+    public:
+        // Constructors
+        SchemaMetadata() = default;
+
+        /**
+         * Constructs metadata without `uncompressed_size`, which is derived later after reading
+         * neighboring schema entries.
+         *
+         * @param stream_id
+         * @param stream_offset
+         * @param num_messages
+         */
+        SchemaMetadata(size_t stream_id, size_t stream_offset, uint64_t num_messages)
+                : m_stream_id{stream_id},
+                  m_stream_offset{stream_offset},
+                  m_num_messages{num_messages} {}
+
+        // Methods
+        [[nodiscard]] auto stream_id() const -> size_t { return m_stream_id; }
+
+        [[nodiscard]] auto stream_offset() const -> size_t { return m_stream_offset; }
+
+        [[nodiscard]] auto num_messages() const -> uint64_t { return m_num_messages; }
+
+        [[nodiscard]] auto uncompressed_size() const -> size_t { return m_uncompressed_size; }
+
+        auto set_uncompressed_size(size_t uncompressed_size) -> void {
+            m_uncompressed_size = uncompressed_size;
+        }
+
+    private:
+        // Members
+        size_t m_stream_id{0};
+        size_t m_stream_offset{0};
+        uint64_t m_num_messages{0};
+        size_t m_uncompressed_size{0};
     };
 
     // Constructor


components/core/src/clp_s/ArchiveReader.cpp
```diff
@@ -54,8 +55,8 @@ void ArchiveReader::read_metadata() {
 
     m_stream_reader.read_metadata(m_table_metadata_decompressor);
 
-    size_t num_separate_column_schemas;
-    if (auto error
+    uint64_t num_separate_column_schemas{0};
```
gibber9809, line 128: Alongside this change we also need to change the write side to use `uint64_t` -- basically just need to update `ArchiveWriter.cpp:453` and `ArchiveWriter.cpp:456` to write `uint64_t`.

components/core/src/clp_s/ReaderUtils.cpp
```diff
@@ -14,29 +16,34 @@ std::shared_ptr<SchemaTree> ReaderUtils::read_schema_tree(ArchiveReaderAdaptor&
             = adaptor.checkout_reader_for_section(constants::cArchiveSchemaTreeFile);
     schema_tree_decompressor.open(*schema_tree_reader, cDecompressorFileReadBufferCapacity);
 
-    size_t num_nodes;
+    uint64_t num_nodes{0};
```
gibber9809, line 19: Similar to my other comment, we also currently directly write `size_t` instead of `uint64_t` sized values on the write path for the schema tree -- need to modify `SchemaTree.hpp:104` and `SchemaTree.hpp:109` to write `uint64_t`.

components/core/src/clp_s/ReaderUtils.hpp
```diff
@@ -67,6 +73,32 @@ class ReaderUtils {
...
+     * running on a 64-bit platform serialized values using `size_t`, a reader on a platform with a
+     * narrower pointer type (e.g. wasm32) also using `size_t` could misinterpret the value due to
+     * truncation.
+     *
+     * This helper assumes a 64-bit archive writer and verifies that a deserialized 64-bit value can
+     * be safely represented as `size_t` before it is used for local buffer or container sizes.
+     *
+     * In the long term, the archive format should record explicit numeric serialization widths
+     * rather than relying on platform-sized types.
+     *
```
gibber9809:
```suggestion
```
The archive format is the same regardless of whether it is written by a 64 bit writer or not -- a few write/read sites were still (incorrectly) using size_t since we haven't really supported 32 bit platforms before. Fixing those sites is a prerequisite to supporting 32 bit platforms, which we will do in this PR, so this part of the comment isn't really needed.

components/core/src/clp_s/PackedStreamReader.cpp
```diff
@@ -14,28 +20,50 @@ void PackedStreamReader::read_metadata(ZstdDecompressor& decompressor) {
...
+        ErrorCodeSuccess != error)
+    {
         throw OperationFailed(error, __FILE__, __LINE__);
     }
+
+    auto const num_streams_result = ReaderUtils::try_uint64_to_size_t(num_streams_u64);
+    if (num_streams_result.has_error()) {
+        throw OperationFailed(ErrorCodeOutOfBounds, __FILENAME__, __LINE__);
+    }
+    auto const num_streams = num_streams_result.value();
```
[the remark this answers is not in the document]
Bill-hbrhbr: @coderabbitai
Add an issue to replace all `__FILE__` with `__FILENAME__`.

components/core/src/clp_s/PackedStreamReader.hpp
```diff
@@ -40,7 +42,8 @@ class PackedStreamReader {
      * reading packed streams.
      * @param decompressor an open ZstdDecompressor pointing to the packed stream metadata
      */
-    void read_metadata(ZstdDecompressor& decompressor);
+    [[nodiscard]] auto read_metadata(ZstdDecompressor& decompressor)
```
gibber9809, line 47: Nit: Just to apply our standard for return value documentation for results the docstring above should have:
```
* @return A void result on success, or an error code indicating the failure:
* - Forwards `ReaderUtils::try_uint64_to_size_t`'s return values on failure.
```

components/core/src/clp_s/ReaderUtils.hpp
```diff
@@ -67,6 +73,20 @@ class ReaderUtils {
             ArchiveReaderAdaptor& adaptor
     );
 
+    /**
+     * Converts a serialized 64-bit numeric value into `size_t` with bounds checking.
+     * @param value The 64-bit value deserialized from archive metadata.
+     * @return Converted `size_t` on success.
+     * @return std::errc::value_too_large if the value cannot fit in `size_t`.
```
gibber9809:
```suggestion
     * @return A result containing the converted value on success, or an error code indicating the
     * failure:
     * - std::errc::value_too_large if the value cannot fit in `size_t`.
```
Nit: same as my other comment, just applying the standard way of documenting result return values.

Bill-hbrhbr:
@coderabbitai Review

components/core/src/clp_s/SchemaReader.hpp
```diff
@@ -55,10 +57,10 @@ class SchemaReader {
     };
 
     struct SchemaMetadata {
-        uint64_t stream_id;
-        uint64_t stream_offset;
-        uint64_t num_messages;
-        uint64_t uncompressed_size;
+        size_t stream_id{0};
```
gibber9809: I think I agree that it makes sense to use `size_t` here so that we can pre-validate that this works on a 32-bit platform, but I think it's confusing that we're using `size_t` on the read-side despite the fields being 64-bit in the format.

Could you add a comment above the struct explaining why we use `size_t` for these fields?

components/core/src/clp_s/search/Output.cpp
```diff
@@ -37,7 +37,11 @@ bool Output::filter() {
     bool has_array = false;
     bool has_array_search = false;
 
-    m_archive_reader->read_metadata();
+    if (auto const result{m_archive_reader->read_metadata()}; result.has_error()) {
+        SPDLOG_ERROR("Failed to read archive metadata.");
```
gibber9809: I know these errors aren't super meaningful right now, but we should probably include the `.error().message()`  in these error logs anyway.

components/core/src/clp_s/ArchiveReader.cpp
```diff
@@ -100,61 +144,34 @@ void ArchiveReader::read_metadata() {
...
+          prev_metadata]{YSTDLIB_ERROR_HANDLING_TRYX(read_single_schema_metadata())};
+    m_schema_ids.push_back(prev_schema_id);
+    for (uint64_t i{1}; i < num_schemas; ++i) {
+        auto const [schema_id, metadata]{
+                YSTDLIB_ERROR_HANDLING_TRYX(read_single_schema_metadata())
+        };
+        m_schema_ids.push_back(schema_id);
 
-        if (stream_offset > m_stream_reader.get_uncompressed_stream_size(stream_id)) {
+        if (metadata.stream_offset < prev_metadata.stream_offset) {
```
gibber9809: This needs to be conditional on them being part of the same stream, otherwise it's guaranteed to throw for archives with more than one packed stream of ERTs.

components/core/src/clp_s/SchemaReader.hpp
gibber9809: Could we add a constructor that takes `stream_id`, `stream_offset`, and `num_messages`? Would help communicate that we're intentionally not initializing `uncompressed_size` in `read_single_schema_metadata` I think.

components/core/src/clp_s/ArchiveReader.hpp
```diff
@@ -53,6 +55,7 @@ class ArchiveReader {
 
     /**
      * Reads the dictionaries and metadata.
+     * @throws OperationFailed if reading or decompressing metdata fails.
```
gibber9809:
```suggestion
     * @throws OperationFailed if reading or decompressing metadata fails.
```

components/core/src/clp_s/ArchiveReader.hpp
```diff
@@ -93,8 +96,15 @@ class ArchiveReader {
 
     /**
      * Reads the metadata from the archive.
+     * @return A void result on success, or an error code indicating the failure:
+     * - Forwards `ArchiveReader::read_single_schema_metadata`'s return values on failure.
+     * - Forwards `PackedStreamReader::read_metadata`'s return values on failure.
+     * - Forwards `ReaderUtils::try_uint64_to_size_t`'s return values on failure.
+     * @throws OperationFailed if archive metadata is empty or corrupt.
+     * @throws OperationFailed if archive metadata stream offset is not strictly incremental.
+     * @throws OperationFailed if reading or decompressing metadata fails.
```
gibber9809, line 104: I _really_ don't like the mix of different error handling styles here after this change, but the scope of fixing that goes a bit beyond what you're doing here, so we can fix it in a separate PR.

We should get rid of the throws here, and use a custom error category/error codes instead of relying on the `std` ones. As it stands the mix of styles makes it difficult for a caller to address and forward errors correctly.

@coderabbitai can you create an issue to track?

components/core/src/clp_s/ArchiveReader.hpp
```diff
@@ -93,8 +96,15 @@ class ArchiveReader {
 
     /**
      * Reads the metadata from the archive.
+     * @return A void result on success, or an error code indicating the failure:
+     * - Forwards `ArchiveReader::read_single_schema_metadata`'s return values on failure.
+     * - Forwards `PackedStreamReader::read_metadata`'s return values on failure.
+     * - Forwards `ReaderUtils::try_uint64_to_size_t`'s return values on failure.
```
gibber9809:
```suggestion
```
As far as I can tell this function doesn't directly call `try_uint64_to_size_t`?

components/core/src/clp_s/TimestampEntry.cpp
```diff
@@ -47,34 +49,40 @@ void TimestampEntry::write_to_stream(std::stringstream& stream) const {
...
+        int32_t id{0};
         error_code = decompressor.try_read_numeric_value<int32_t>(id);
         if (ErrorCodeSuccess != error_code) {
             return error_code;
         }
         m_column_ids.insert(id);
     }
 
-    uint64_t encoding;
+    uint64_t encoding{0};
```
gibber9809:
```suggestion
```
Yeah rabbit is right that this is unused since we read directly to `m_encoding` -- can get rid of it.

PR #2073, opened 2026-03-05
fix(deployment): Pin Presto image tags to versioned defaults and allow overrides via env vars (fixes #2072).

### LINKED ISSUE #2072: Presto docker-compose may use stale `:dev` images due to missing `pull_policy`
### Bug

The Presto docker-compose file (`tools/deployment/presto-clp/docker-compose.yaml`) uses mutable
`:dev` image tags for both `presto-coordinator` and `presto-worker` services without setting
`pull_policy: "always"`. Docker Compose's default pull policy is `missing`, which means it only
pulls an image if one doesn't already exist locally. As a result, once a `:dev` image is cached
locally, subsequent `docker compose up` invocations will reuse the stale local copy even if a newer
version has been pushed to the registry.

**Expected:** Running `docker compose up` should always use the latest `:dev` images from the
registry.

**Actual:** Docker reuses the locally cached `:dev` images, which may be outdated.

**Affected services:**
- `presto-coordinator`: `ghcr.io/y-scope/presto/coordinator:dev`
- `presto-worker`: `ghcr.io/y-scope/presto/prestissimo-worker:dev`

### CLP version

dbc1799b

### Environment

Any environment using Docker Compose with the Presto deployment (`tools/deployment/presto-clp/docker-compose.yaml`) that has previously pulled an older Presto image.

### Reproduction steps

1. Run `docker compose up` in `tools/deployment/presto-clp/` to pull the `:dev` images.
2. Wait for a newer `:dev` image to be pushed to the registry.
3. Run `docker compose up` again.
4. Observe that the locally cached (stale) images are used instead of the latest ones from the
   registry.

junhaoliao:
# Description

The Presto `docker-compose.yaml` hardcodes `:dev` image tags for both the `presto-coordinator` and `presto-worker` services:

https://github.com/y-scope/clp/blob/dbc1799b187e034bec4578c058e9dfa1c87f05fb/tools/deployment/presto-clp/docker-compose.yaml#L3

https://github.com/y-scope/clp/blob/dbc1799b187e034bec4578c058e9dfa1c87f05fb/tools/deployment/presto-clp/docker-compose.yaml#L27

This causes a problem: it's impossible to pin specific image versions on release branches, and the mutable `:dev` tag can silently change underneath users.

This PR:
- Replaces the hardcoded `:dev` tags with configurable environment variables that default to
  `clp-v0.10.0`:
  - `CLP_PRESTO_COORDINATOR_IMAGE_TAG` (default: `clp-v0.10.0`) for the coordinator service.
  - `CLP_PRESTO_WORKER_IMAGE_TAG` (default: `clp-v0.10.0`) for the worker service.

The versioned default (`clp-v0.10.0`) follows the convention used by other third-party services in the project that pin specific version tags (e.g., `redis:7.2.4`, `mongo:7.0.1`, `rabbitmq:3.9.8`).
The configurable env var approach follows the existing convention where Docker image references are overridable (e.g., `CLP_DB_CONTAINER_IMAGE_REF` in `docker-compose-all.yaml`).

# Checklist

  breaking change.

# Validation performed

## Scenario 1: Default image tags (no env vars set)

**Task:** Verify that with no env vars set, both images resolve to the `:clp-v0.10.0` tag and no
`pull_policy` is configured (matching how other third-party services are defined).

**Command:**
```bash
cd tools/deployment/presto-clp
touch .env
docker compose config 2>&1 | grep -E "(image:|pull_policy:)"
rm .env
```

**Output:**
```
    image: ghcr.io/y-scope/presto/coordinator:clp-v0.10.0
    image: ghcr.io/y-scope/presto/prestissimo-worker:clp-v0.10.0
```

**Explanation:** Both services correctly default to `:clp-v0.10.0`. No `pull_policy` is present,
which matches the convention used by other third-party services (`redis`, `mongo`, `rabbitmq`).
Since the version tag is immutable, Docker's default `pull_policy: missing` is sufficient.

## Scenario 2: Custom image tags via environment variables

**Task:** Verify that setting `CLP_PRESTO_COORDINATOR_IMAGE_TAG` and `CLP_PRESTO_WORKER_IMAGE_TAG`
overrides the default tag.

**Command:**
```bash
cd tools/deployment/presto-clp
echo -e "CLP_PRESTO_COORDINATOR_IMAGE_TAG=clp-v0.9.0\nCLP_PRESTO_WORKER_IMAGE_TAG=clp-v0.9.0" > .env
docker compose config 2>&1 | grep -E "(image:|pull_policy:)"
rm .env
```

**Output:**
```
    image: ghcr.io/y-scope/presto/coordinator:clp-v0.9.0
    image: ghcr.io/y-scope/presto/prestissimo-worker:clp-v0.9.0
```

**Explanation:** Both services correctly use the custom tag `clp-v0.9.0`, confirming that release
branches can pin specific image versions via these environment variables.

## Summary by CodeRabbit

* **Chores**
  * Updated Presto deployment to use environment-driven image versioning for coordinator and worker services.
  * Enables specifying image tags at deploy time and ensures images refresh automatically during deployments.


junhaoliao:
Converting this to draft until I confirm the changes are working with the latest code

@hoophalab @20001020ycx i'll ping you when it's ready for review. thanks in advance

hoophalab (dismissed):
LGTM. Though I'm not sure if `pull_policy: always` is needed. The user should decide whether to pull a new image of same tag.

junhaoliao:
> LGTM. Though I'm not sure if `pull_policy: always` is needed. The user should decide whether to pull a new image of same tag.

i think the using the `latest` tag with the always pull policy is fine on the `main` branch as it's supposed to be a staging (for development) than a release branch. 

that said, using the `latest` tag does differ from the specifications of other third-party images where fixed version tags are used. the plan going forward is to allow "presto" to be configurable in the `clpConfig.bundled[]` and therefore the Presto services should be consider "third-party" from the CLP package's perspective, and i think their handling should match. how do you feel about replacing the `latest` tag with a version tag then?

cc @kirkrodrigues

kirkrodrigues:
> how do you feel about replacing the `latest` tag with a version tag then?

Seems reasonable. I guess by `latest`, you really mean `dev` in this context?

junhaoliao:
> > how do you feel about replacing the `latest` tag with a version tag then?
> 
> Seems reasonable. I guess by `latest`, you really mean `dev` in this context?

right i meant `dev`. sorry

---

i'll update the PR to use the version tag then

hoophalab (approved):

PR #2082, opened 2026-03-06
feat(job-orchestration): Add `S3ObjectMetadataInputConfig` for compression of `log-ingestor`-ingested S3 objects.

jonathan-imanu:
# Description

This PR is the first step in implementing the proposal in #2018. A the relevant section of the proposal is below: 

**Current Problem**

The Compression Scheduler makes a redundant trip to S3 for jobs submitted by the log-ingestor.  

Regardless of job source, the scheduler uses the `key` field in S3InputConfig to get the metadata of files by reaching out to S3. However, for jobs submitted by the log-ingestor, object size is already stored in ingested_s3_object_metadata. 

**The Fix** 

The current configuration used by the scheduler (`S3InputConfig`) for ingestor submitted jobs has no reference to the `ingested_s3_object_metadata`. This PR adds support for a new input that contains the `ingestor_job_id` & `metadata_ids` so the scheduler can directly query the `ingested_s3_object_metadata` table for the metadata.

This currently does nothing since the log-ingestor is still submitting `S3InputConfig`. This will be addressed in the next PR.

# Checklist

  breaking change.

# Validation performed

- Workflows
- A one-time validation of the job scheuduler detailed below

### Setup

- Ensure `minio` is installed.
- Create a folder called `test-data` and a subfolder `test-data/elasticsearch/`
- Download one of the structured datasets [here](https://docs.yscope.com/clp/main/user-docs/resources-datasets.html#datasets) & untar. I used [elasticsearch](https://zenodo.org/records/10516226) and my commands assume you did as well.

```bash
# in a separate terminal in the parent of the test-data folder
minio server ./test-data
```

```bash
task
# IMPORTANT: UPDATE YOUR CREDENTIALS AND CLP.CONFIG WITH MINIO CREDS!
# ALSO MAKE SURE THE LOG-INGESTOR IS ENABLED!
build/clp-package/sbin/start-clp.sh
```

### Testing

Create a S3 Scanner job

```bash
curl -X POST "http://127.0.0.1:3002/s3_scanner" \
  -d  '{"bucket_name":"elasticsearch","key_prefix":"elasticsearch","endpoint_url":"http://<minio-host>:9000","scanning_interval_sec":30}'
```

Verify it worked:

```bash
docker exec -it $(docker ps --filter name=database --format '{{.Names}}' | head -1) \
  mysql -u clp-user -p clp-db -e "SELECT id, status, status_msg, num_files_compressed, creation_ts, last_update_ts FROM ingestion_job WHERE id = 1;"
```

Once the status has been compressed, reset the entry to submitted:

```bash
docker exec -it $(docker ps --filter name=database --format '{{.Names}}' | head -1) \
  mysql -u clp-user -p clp-db -e "UPDATE ingested_s3_object_metadata SET status = 'submitted' WHERE id = 1;"
```

Now we need to manually create a submission with the metadata_ids to the `compression_jobs` table. We can't use raw SQL to do this because the job scheduler expects this config to be brotli-compressed msgpack. So create the following script in `tools/`, make sure to replace the `MINIO_ENDPOINT_URL`!

#### tools/insert-compression-job-from-metadata.py

```python
# disclaimer: this is fully AI generated lol 

import argparse
import os
import sys
from pathlib import Path

if str(Path(__file__).resolve().parent.parent) not in sys.path:
    sys.path.insert(0, str(Path(__file__).resolve().parent.parent))

import brotli
import msgpack
import mysql.connector

from clp_py_utils.clp_config import COMPRESSION_JOBS_TABLE_NAME
from job_orchestration.scheduler.job_config import (
    ClpIoConfig,
    InputType,
    OutputConfig,
    S3ObjectMetadataInputConfig,
)

# Hardcoded defaults
MINIO_ENDPOINT_URL = "http://<minio-host>:9000"  
OUTPUT = OutputConfig(
    target_archive_size=268435456,
    target_dictionaries_size=33554432,
    target_segment_size=268435456,
    target_encoded_file_size=268435456,
    compression_level=3,
)
AWS_AUTH = {"type": "credentials", "credentials": {"access_key_id": "minioadmin", "secret_access_key": "minioadmin"}}


def main():
    p = argparse.ArgumentParser(description="Insert compression job from ingested_s3_object_metadata IDs.")
    p.add_argument("--ingestion-job-id", type=int, required=True)
    p.add_argument("--metadata-ids", type=int, nargs="*", default=None, help="Default: fetch all from DB")
    p.add_argument("--sql-only", action="store_true", help="Print INSERT only; requires --bucket")
    p.add_argument("--bucket", type=str, default=None)
    p.add_argument("--key-prefix", type=str, default="")
    p.add_argument("--dataset", type=str, default="default", help="Dataset name (default: default)")
    args = p.parse_args()

    if args.sql_only and not args.bucket:
        p.error("--bucket required with --sql-only")
    if args.sql_only and not args.metadata_ids:
        p.error("--metadata-ids required with --sql-only")

    bucket = args.bucket
    key_prefix = args.key_prefix or ""
    metadata_ids = args.metadata_ids

    conn = None
    if not args.sql_only:
        conn = mysql.connector.connect(
            host=os.environ.get("CLP_DB_HOST", "127.0.0.1"),
            port=int(os.environ.get("CLP_DB_PORT", "3306")),
            user=os.environ.get("CLP_DB_USER", "clp-user"),
            password=os.environ.get("CLP_DB_PASS", ""),
            database=os.environ.get("CLP_DB_NAME", "clp-db"),
        )
        cur = conn.cursor(dictionary=True)
        cur.execute(
            "SELECT id, bucket, `key` FROM ingested_s3_object_metadata WHERE ingestion_job_id = %s",
            (args.ingestion_job_id,),
        )
        rows = cur.fetchall()
        if not rows:
            print("No rows for that ingestion_job_id. Ingest objects first.", file=sys.stderr)
            conn.close()
            return 1
        if metadata_ids is None:
            metadata_ids = [r["id"] for r in rows]
        if bucket is None:
            bucket = rows[0]["bucket"]

    cfg = ClpIoConfig(
        input=S3ObjectMetadataInputConfig(
            type=InputType.S3_OBJECT_METADATA.value,
            ingestion_job_id=args.ingestion_job_id,
            s3_object_metadata_ids=metadata_ids,
            bucket=bucket,
            key_prefix=key_prefix,
            aws_authentication=AWS_AUTH,
            endpoint_url=MINIO_ENDPOINT_URL,
            region_code=None,
            dataset=args.dataset,
        ),
        output=OUTPUT,
    )
    blob = brotli.compress(msgpack.packb(cfg.model_dump()))

    if args.sql_only:
        print(f"INSERT INTO {COMPRESSION_JOBS_TABLE_NAME} (clp_config, status) VALUES (0x{blob.hex()}, 0);")
        return 0

    cur = conn.cursor()
    cur.execute(f"INSERT INTO {COMPRESSION_JOBS_TABLE_NAME} (clp_config, status) VALUES (%s, 0)", (blob,))
    conn.commit()
    print(f"Inserted compression job id={cur.lastrowid} (status=PENDING).")
    conn.close()
    return 0


if __name__ == "__main__":
    sys.exit(main())
```

Then, run the script from the repo root:

```bash
uv run --project components/job-orchestration tools/insert-compression-job-from-metadata.py --ingestion-job-id 1 --metadata-ids 1
```

If you are trying to use this from a dev container, use `--sql-only`:

```bash
uv run --project components/job-orchestration tools/insert-compression-job-from-metadata.py --ingestion-job-id 1 --metadata-ids 1 --bucket elasticsearch --sql-only
```

Then execute the query in the DB

Verify the job succeeded: 

```bash
docker logs $(docker ps -a --filter name=compression-scheduler --format '{{.Names}}' | head -1) 2>&1 | tail -80
```

## Summary by CodeRabbit

* **New Features**
  * Add support for ingest-registered S3 object metadata as a new input type for compression jobs.
  * Metadata inputs are validated, fetched, and enqueued for compression.
  * Metadata-based inputs are handled like S3 for authentication and log-list generation.

* **New Features**
  * Expose new output settings: target segment size, target encoded file size, and compression level.

* **Bug Fixes**
  * Failures when fetching metadata are logged and mark the job as failed.

* **Chores**
  * Configuration and constants updated to recognise the new metadata input type.

### DIFF
--- a/components/core/.clang-format
+++ b/components/core/.clang-format
@@ -7,7 +7,7 @@ IncludeCategories:
   # NOTE: clang-format retains leading white-space on a line in violation of the YAML spec.
   - Regex: "<(absl|antlr4|archive|boost|bsoncxx|catch2|curl|date|fmt|log_surgeon|lzma\
 |mongocxx|msgpack|mysql|nlohmann|openssl|outcome|regex_utils|simdjson|spdlog|sqlite3|string_utils\
-|yaml-cpp|ystdlib|zstd)"
+|xxhash|yaml-cpp|ystdlib|zstd)"
     Priority: 3
 
   # Project library headers
--- a/components/core/CMakeLists.txt
+++ b/components/core/CMakeLists.txt
@@ -329,6 +329,13 @@ if(CLP_NEED_SQLITE)
     FindDynamicLibraryDependencies(sqlite "${sqlite_DYNAMIC_LIBS}")
 endif()
 
+if(CLP_NEED_XXHASH)
+    find_package(xxHash 0.8.3 CONFIG REQUIRED)
+    if(xxHash_FOUND)
+        message(STATUS "Found xxHash ${xxHash_VERSION}")
+    endif()
+endif()
+
 add_subdirectory(src/clp/regex_utils)
 add_subdirectory(src/clp/string_utils)
 
@@ -359,6 +366,8 @@ set(SOURCE_FILES_clp_s_unitTest
     src/clp_s/FileReader.hpp
     src/clp_s/FileWriter.cpp
     src/clp_s/FileWriter.hpp
+    src/clp_s/filter/tests/test-clp_s-bloom_filter.cpp
+    src/clp_s/filter/tests/test-clp_s-xxhash.cpp
     src/clp_s/FloatFormatEncoding.cpp
     src/clp_s/FloatFormatEncoding.hpp
     src/clp_s/ffi/sfa/ClpArchiveReader.cpp
@@ -778,6 +787,7 @@ if(CLP_ENABLE_TESTS)
             clp_s::search::sql
             clp_s::timestamp_parser
             clp_s::timestamp_pattern
+            clp_s::filter
             date::date
             fmt::fmt
             log_surgeon::log_surgeon
--- a/components/core/cmake/Options/options.cmake
+++ b/components/core/cmake/Options/options.cmake
@@ -101,6 +101,12 @@ option(
     ON
 )
 
+option(
+    CLP_BUILD_CLP_S_FILTER
+    "Build clp_s::filter."
+    ON
+)
+
 # Validates that the `CLP_BUILD_` options required by `TARGET_CLP_BUILD_OPTION` are `ON`.
 #
 # @param {string} TARGET_CLP_BUILD_OPTION
@@ -168,6 +174,7 @@ function(validate_clp_tests_dependencies)
         CLP_BUILD_CLP_REGEX_UTILS
         CLP_BUILD_CLP_STRING_UTILS
         CLP_BUILD_CLP_S_ENABLE_CURL
+        CLP_BUILD_CLP_S_FILTER
         CLP_BUILD_CLP_S_SEARCH_AST
         CLP_BUILD_CLP_S_SEARCH_KQL
         CLP_BUILD_CLP_S_SEARCH_SQL
@@ -267,6 +274,19 @@ function(set_clp_s_clp_dependencies_dependencies)
         CLP_NEED_SPDLOG
         CLP_NEED_YSTDLIB
         CLP_NEED_ZSTD
+    )
+endfunction()
+
+function(validate_clp_s_filter_dependencies)
+    validate_clp_dependencies_for_target(CLP_BUILD_CLP_S_FILTER
+        CLP_BUILD_CLP_S_CLP_DEPENDENCIES
+    )
+endfunction()
+
+function(set_clp_s_filter_dependencies)
+    set_clp_need_flags(
+        CLP_NEED_XXHASH
+        CLP_NEED_YSTDLIB
     )
 endfunction()
 
@@ -435,6 +455,11 @@ function(validate_and_setup_all_clp_dependency_flags)
         set_clp_s_clp_dependencies_dependencies()
     endif()
 
+    if (CLP_BUILD_CLP_S_FILTER)
+        validate_clp_s_filter_dependencies()
+        set_clp_s_filter_dependencies()
+    endif()
+
     if (CLP_BUILD_CLP_S_IO)
         validate_clp_s_io_dependencies()
         set_clp_s_io_dependencies()
@@ -504,6 +529,7 @@ function (convert_clp_dependency_properties_to_variables)
         CLP_NEED_SIMDJSON
         CLP_NEED_SPDLOG
         CLP_NEED_SQLITE
+        CLP_NEED_XXHASH
         CLP_NEED_YAMLCPP
         CLP_NEED_YSTDLIB
         CLP_NEED_ZSTD
--- a/components/core/src/clp_s/CMakeLists.txt
+++ b/components/core/src/clp_s/CMakeLists.txt
@@ -1,4 +1,5 @@
 add_subdirectory(ffi)
+add_subdirectory(filter)
 add_subdirectory(indexer)
 add_subdirectory(log_converter)
 add_subdirectory(search)
--- a/components/core/src/clp_s/filter/BloomFilter.cpp
+++ b/components/core/src/clp_s/filter/BloomFilter.cpp
@@ -1,0 +1,199 @@
+#include "BloomFilter.hpp"
+
+#include <algorithm>
+#include <cmath>
+#include <cstddef>
+#include <cstdint>
+#include <limits>
+#include <numbers>
+#include <string_view>
+#include <utility>
+
+#include <ystdlib/containers/Array.hpp>
+#include <ystdlib/error_handling/Result.hpp>
+
+#include <clp/ErrorCode.hpp>
+#include <clp/ReaderInterface.hpp>
+#include <clp/type_utils.hpp>
+#include <clp/WriterInterface.hpp>
+
+#include "ErrorCode.hpp"
+#include "HashAlgorithm.hpp"
+
+namespace clp_s::filter {
+namespace {
+constexpr size_t cDefaultBitArraySize{64};
+constexpr uint32_t cDefaultNumHashFunctions{1};
+constexpr uint32_t cMinNumHashFunctions{1};
+constexpr uint32_t cMaxNumHashFunctions{20};
+constexpr double cMinFalsePositiveRate{1e-6};
+constexpr HashAlgorithm cDefaultHashAlgorithm{HashAlgorithm::Xxh364};
+constexpr uint64_t cPrimaryHashSeed{0};
+constexpr uint64_t cSecondaryHashSeed{0x9e37'79b9'7f4a'7c15ULL};
+constexpr size_t cNumBitsInByte{8};
+
+[[nodiscard]] auto min_bytes_containing_bits(size_t num_bits) -> size_t {
+    return (num_bits / cNumBitsInByte) + ((0 != (num_bits % cNumBitsInByte)) ? 1 : 0);
+}
+}  // namespace
+
+auto BloomFilter::create(size_t expected_num_elements, double false_positive_rate)
+        -> ystdlib::error_handling::Result<BloomFilter> {
+    auto const [bit_array_size, num_hash_functions] = YSTDLIB_ERROR_HANDLING_TRYX(
+            compute_optimal_parameters(expected_num_elements, false_positive_rate)
+    );
+
+    size_t const num_bytes{min_bytes_containing_bits(bit_array_size)};
+    return BloomFilter{
+            bit_array_size,
+            num_hash_functions,
+            cDefaultHashAlgorithm,
+            ystdlib::containers::Array<uint8_t>(num_bytes)
+    };
+}
+
+auto BloomFilter::try_read_from_file(clp::ReaderInterface& reader)
+        -> ystdlib::error_handling::Result<BloomFilter> {
+    uint8_t hash_algorithm_u8{};
+    if (clp::ErrorCode_Success != reader.try_read_numeric_value(hash_algorithm_u8)) {
+        return ErrorCode{ErrorCodeEnum::ReadFailure};
+    }
+    auto const optional_hash_algorithm{try_parse_hash_algorithm(hash_algorithm_u8)};
+    if (false == optional_hash_algorithm.has_value()) {
+        return ErrorCode{ErrorCodeEnum::UnsupportedHashAlgorithm};
+    }
+    auto const hash_algorithm{optional_hash_algorithm.value()};
+
+    uint32_t num_hash_functions{};
+    if (clp::ErrorCode_Success != reader.try_read_numeric_value(num_hash_functions)) {
+        return ErrorCode{ErrorCodeEnum::ReadFailure};
+    }
+    if (num_hash_functions < cMinNumHashFunctions || num_hash_functions > cMaxNumHashFunctions) {
+        return ErrorCode{ErrorCodeEnum::CorruptFilterPayload};
+    }
+
+    uint64_t bit_array_size_u64{};
+    if (clp::ErrorCode_Success != reader.try_read_numeric_value(bit_array_size_u64)) {
+        return ErrorCode{ErrorCodeEnum::ReadFailure};
+    }
+    if (0 == bit_array_size_u64 || bit_array_size_u64 > std::numeric_limits<size_t>::max()) {
+        return ErrorCode{ErrorCodeEnum::CorruptFilterPayload};
+    }
+    size_t const bit_array_size{static_cast<size_t>(bit_array_size_u64)};
+
+    uint64_t bit_array_bytes{};
+    if (clp::ErrorCode_Success != reader.try_read_numeric_value(bit_array_bytes)) {
+        return ErrorCode{ErrorCodeEnum::ReadFailure};
+    }
+    if (bit_array_bytes > std::numeric_limits<size_t>::max()) {
+        return ErrorCode{ErrorCodeEnum::CorruptFilterPayload};
+    }
+    size_t const expected_bit_array_bytes{min_bytes_containing_bits(bit_array_size)};
+    if (bit_array_bytes != expected_bit_array_bytes) {
+        return ErrorCode{ErrorCodeEnum::CorruptFilterPayload};
+    }
+
+    ystdlib::containers::Array<uint8_t> bit_array(static_cast<size_t>(bit_array_bytes));
+    if (false == bit_array.empty()) {
+        if (clp::ErrorCode_Success
+            != reader.try_read_exact_length(
+                    clp::size_checked_pointer_cast<char>(bit_array.data()),
+                    static_cast<size_t>(bit_array_bytes)
+            ))
+        {
+            return ErrorCode{ErrorCodeEnum::ReadFailure};
+        }
+    }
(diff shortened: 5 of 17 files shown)


components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -392,7 +395,7 @@ def run_clp(
     logs_list_path = tmp_dir / f"{instance_id_str}-log-paths.txt"
     if InputType.FS == input_type:
         _generate_fs_logs_list(logs_list_path, paths_to_compress)
-    elif InputType.S3 == input_type:
+    elif InputType.S3 == input_type or InputType.INGESTOR == input_type:
         _generate_s3_logs_list(logs_list_path, paths_to_compress, clp_config.input)
```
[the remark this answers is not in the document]
jonathan-imanu: Seems to be a pre-existing bug no? Would fixing this in a separate PR be best?

LinZhihao-723 (commented):
As a response to review the naming for this new config:
How about `S3ObjectMetadataInputConfig`?
This is to match the table for storing the S3 object metadata: the table is called `ingested_s3_object_metadata`, and we're essentially referring to object metadata IDs in this table. @junhaoliao Please also check.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -183,6 +188,75 @@ def _process_s3_input(
...
+    given metadata_ids and ingestion_job_id.
+
+    :param metadata_ids: IDs to fetch.
+    :param ingestion_job_id: Ingestion job to filter by.
+    :param db_cursor: Database cursor for the query.
+    :return: List of row dicts with "id", "key", and "size".
+    :raises RuntimeError: If no rows are found, or if any requested metadata_id is missing.
+    """
+    if not metadata_ids:
+        return []
```
junhaoliao: is this dead code given we already
```
raise ValueError("metadata_ids cannot be an empty list")
```

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -313,7 +316,7 @@ def _make_log_converter_command_and_env(
     # fmt: on
 
     conversion_env_vars = dict(os.environ)
-    if InputType.S3 == clp_config.input.type:
+    if InputType.S3 == clp_config.input.type or InputType.INGESTOR == clp_config.input.type:
```
junhaoliao: this is repeated quite a few times - shall we create a helper like
```
def _is_s3_based_input(input_type: InputType) -> bool:
    return input_type in {InputType.S3, InputType.INGESTOR}
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -183,6 +188,75 @@ def _process_s3_input(
         paths_to_compress_buffer.add_file(object_metadata)
 
 
+def _fetch_ingested_s3_object_metadata(
+    metadata_ids: list[int],
+    ingestion_job_id: int,
+    db_cursor: Any,
+) -> list[dict[str, Any]]:
```
junhaoliao: consider returning a list of `FileMetadata` directly from this function (or a small dataclass/`NamedTuple`) instead of passing raw dicts through. this would also eliminate the separate loop in `_process_log_ingestor_submitted_s3_input` and make the code more self-documenting

components/job-orchestration/job_orchestration/scheduler/job_config.py
```diff
@@ -11,6 +11,7 @@
 class InputType(LowercaseStrEnum):
     FS = auto()
     S3 = auto()
+    METADATA = auto()
```
LinZhihao-723: Using `METADATA` as the type here can be very confusing, since the metadata as an input doesn't clarify what exactly metadata it is.
How about:
`S3OBJECTMETADATA`, or `S3_OBJECT_METADATA` if that works for @junhaoliao (I'm not too sure about our Pydentic side naming)?

components/job-orchestration/job_orchestration/scheduler/job_config.py
```diff
@@ -44,6 +45,22 @@ def validate_keys(cls, value):
         return value
 
 
+class S3ObjectMetadataInputConfig(S3Config):
+    type: Literal[InputType.METADATA.value] = InputType.METADATA.value
+    ingestion_job_id: int
+    dataset: str | None = None
+    timestamp_key: str | None = None
+    unstructured: bool = False
+    metadata_ids: list[int]
```
LinZhihao-723: Similarly, prefer `s3_object_metadata_ids`.

components/job-orchestration/job_orchestration/scheduler/job_config.py
```diff
@@ -44,6 +45,22 @@ def validate_keys(cls, value):
...
+
+    @field_validator("metadata_ids")
+    @classmethod
+    def validate_metadata_ids_non_empty(cls, value: list[int]) -> list[int]:
+        if not value:
```
LinZhihao-723:
```suggestion
        if len(value) == 0:
```
I understand this is equivalent to `not value` in Python. However, we usually don't do this in our code base because it's not explicitly checking the size of the list. We usually prefer to program Python like other languages without using too many syntax sugars.

components/job-orchestration/job_orchestration/scheduler/utils.py
```diff
@@ -17,6 +17,11 @@
     QueryTaskStatus,
     SchedulerType,
 )
+from job_orchestration.scheduler.job_config import InputType
+
+
+def _is_s3_based_input(input_type: InputType) -> bool:
```
LinZhihao-723: I think we should use `is_s3_based_input`.
`_` prefix is used for local helper functions. This method is exported from this mod.

components/job-orchestration/job_orchestration/scheduler/utils.py
```diff
@@ -17,6 +17,11 @@
...
+    return input_type in {InputType.S3, InputType.METADATA}
```
LinZhihao-723:
```suggestion
    return InputType.S3 != input_type and InputType.METADATA != input_type
```
I understand this is suggested by Junhao, but it's too Python and in practice we shouldn't program in a way where each comparison requires constructing a tuple, lol

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -183,6 +188,52 @@ def _process_s3_input(
...
 
 
+def _process_s3_object_metadata_input(
+    s3_object_metadata_input_config: S3ObjectMetadataInputConfig,
+    paths_to_compress_buffer: PathsToCompressBuffer,
+    db_context: DbContext,
+) -> None:
+    """
+    Fetches S3 object metadata rows from the INGESTED_S3_OBJECT_METADATA_TABLE_NAME table for the
+    given metadata_ids and ingestion_job_id, and adds the metadata to paths_to_compress_buffer.
```
LinZhihao-723:
```suggestion
    Fetches S3 object metadata rows from the `INGESTED_S3_OBJECT_METADATA_TABLE_NAME` table for the
    given `metadata_ids` and `ingestion_job_id`, and adds the metadata to
    `paths_to_compress_buffer`.
```
We should mark symbols with `.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -183,6 +188,52 @@ def _process_s3_input(
...
+
+    :param s3_object_metadata_input_config:
+    :param paths_to_compress_buffer:
+    :param db_context:
+    :raises RuntimeError: If no rows are found, or if any requested metadata_id is missing.
+    """
+    metadata_ids = s3_object_metadata_input_config.metadata_ids
```
LinZhihao-723: We should probably make sure there's no duplicate IDs as a part of validation. Probably worth to put `requested_ids` to the beginning and make sure its size matches `metadata_ids`.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -183,6 +188,52 @@ def _process_s3_input(
...
+    ingestion_job_id = s3_object_metadata_input_config.ingestion_job_id
+
+    placeholders = ", ".join(["%s"] * len(metadata_ids))
+    query = (
+        f"SELECT id, `key`, size FROM {INGESTED_S3_OBJECT_METADATA_TABLE_NAME} "
```
LinZhihao-723:
```suggestion
        f"SELECT `id`, `key`, `size` FROM {INGESTED_S3_OBJECT_METADATA_TABLE_NAME} "
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -183,6 +188,52 @@ def _process_s3_input(
...
+        f"WHERE id IN ({placeholders}) AND ingestion_job_id = %s"
+    )
+    params = (*metadata_ids, ingestion_job_id)
+    db_context.cursor.execute(query, params)
+    metadata_list = db_context.cursor.fetchall()
+    if not metadata_list:
```
LinZhihao-723: Same, please use the explicit size check.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -183,6 +188,52 @@ def _process_s3_input(
...
+    returned_ids = {row["id"] for row in metadata_list}
+    requested_ids = set(metadata_ids)
+    missing_ids = requested_ids - returned_ids
+    if missing_ids:
+        raise RuntimeError(
+            f"Missing metadata rows in {INGESTED_S3_OBJECT_METADATA_TABLE_NAME} for "
+            f"ingestion_job_id {ingestion_job_id}: {sorted(missing_ids)}."
+        )
+
+    for metadata in metadata_list:
```
LinZhihao-723, line 233: For each key, we should also ensure it contains the expected prefix specified in `S3Config` (the base class of `S3ObjectMetadataInputConfig`).

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -183,6 +188,57 @@ def _process_s3_input(
...
+    :param db_context:
+    :raises RuntimeError: If no rows are found, or if any requested metadata_id is missing.
+    """
+    s3_object_metadata_ids = s3_object_metadata_input_config.s3_object_metadata_ids
+    ingestion_job_id = s3_object_metadata_input_config.ingestion_job_id
+
+    placeholders = ", ".join(["%s"] * len(s3_object_metadata_ids))
+    query = (
+        f"SELECT `id`, `key`, `size` FROM {INGESTED_S3_OBJECT_METADATA_TABLE_NAME} "
+        f"WHERE id IN ({placeholders}) AND ingestion_job_id = %s"
```
LinZhihao-723:
```suggestion
        f"WHERE `id` IN ({placeholders}) AND `ingestion_job_id` = %s"
```

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -183,6 +188,57 @@ def _process_s3_input(
...
+    )
+    params = (*s3_object_metadata_ids, ingestion_job_id)
+    db_context.cursor.execute(query, params)
+    metadata_list = db_context.cursor.fetchall()
+    if len(metadata_list) == 0:
+        raise RuntimeError(
+            f"No rows found in {INGESTED_S3_OBJECT_METADATA_TABLE_NAME} for the given "
+            f"s3_object_metadata_ids and ingestion_job_id {ingestion_job_id}."
+        )
+    # Validate that all requested IDs are present.
```
LinZhihao-723, line 223:
```suggestion

    # Validate that all requested IDs are present.
```
Generally the rule is if we have an inline comment, leave an empty line before it.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -183,6 +188,57 @@ def _process_s3_input(
...
+    s3_object_metadata_ids = s3_object_metadata_input_config.s3_object_metadata_ids
+    ingestion_job_id = s3_object_metadata_input_config.ingestion_job_id
+
+    placeholders = ", ".join(["%s"] * len(s3_object_metadata_ids))
+    query = (
+        f"SELECT `id`, `key`, `size` FROM {INGESTED_S3_OBJECT_METADATA_TABLE_NAME} "
+        f"WHERE id IN ({placeholders}) AND ingestion_job_id = %s"
+    )
+    params = (*s3_object_metadata_ids, ingestion_job_id)
+    db_context.cursor.execute(query, params)
```
LinZhihao-723, line 215: @sitaowang1998 Can u help me check:
- iiuc we should use `?` placeholders instead of `%s`, as we did [here](https://github.com/y-scope/spider/blob/57ffe409b0ada2caf0127717e8d75911fb4b9409/python/spider-py/src/spider_py/storage/mariadb_storage.py#L82) for security purposes
- We should be able to use `fetchall`'s results as a tuple without key indexing, right?

sitaowang1998: - Python uses `%s` instead of `?` for SQL queries. As long as all parameters are not directly baked into the query and is feed into `execute` as second parameter, it is safe.
- `fetchall` typically returns a list of tuples. You can treat it as tuple of tuples, but I hardly see any benefits.

LinZhihao-723: - Sure, let's use `%s` then.
- Treat the results as a tuple should avoid paying dictionary indexing. @jonathan-imanu Let's try to read the fetched row as tuple then.

jonathan-imanu: Discussed offline, sticking with dictionary access due to the following:

1. In the compression scheduler, we create the cursor with `closing(db_conn.cursor(dictionary=True)) as db_cursor` ([ref](https://github.com/y-scope/clp/blob/b27425abcbc899776a05d1d1b7f8e2d6734ddcae/components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py#L470)). 
   - The [docs](https://dev.mysql.com/doc/connector-python/en/connector-python-api-mysqlconnection-cursor.html) says "If dictionary is True, the cursor returns rows as dictionaries"
   - This pattern is used elsewhere too. Such as [query_scheduler.py](https://github.com/y-scope/clp/blob/b27425abcbc899776a05d1d1b7f8e2d6734ddcae/components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py#L527)
2. We also use the dictionary pattern access elsewhere in the codebase. Such as
[Getting jobs in Compression Scheduler](https://github.com/y-scope/clp/blob/b27425abcbc899776a05d1d1b7f8e2d6734ddcae/components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py#L250)

PR #2087, opened 2026-03-09
feat(log-ingestor): Use `S3ObjectMetadataInputConfig` for compression job submission (resolves #2018).

### LINKED ISSUE #2018: [log-ingestor] Improve compression job submission flow.

jonathan-imanu:
# Description

This PR builds off of #2082 and completes the proposal in #2018. 

In #2082, we added support for a new input that contains the ingestor_job_id & metadata_ids so the scheduler can directly query the ingested_s3_object_metadata table for the metadata. This input was unused when #2082 was merged since the`log-ingestor` was still submitting S3InputConfig. 

This PR updates the `log-ingestor` to submit the new `S3ObjectMetadataInputConfig`. To achieve this, this PR:

- updates the buffer to hold a list of metadata ids as opposed to complete metadata objects
- introduces a new type `CompressionBufferEntry` to associate object size with metadata_id and updates interfaces of listener and submitter to use this type
- updates the unit tests of `log-ingestor`

# Checklist

  breaking change.

# Validation performed

- `task tests:rust-all`
- Workflows
- Verification of full flow & Compression Scheduler perf. comparisons against mainline (see below).

### Results

The following performance changes are not improvements that are gained from this PR. They are from #2082, I'm just documenting them here because testing is a little easier.

## Before

| job_id | scheduler_discovery_to_scheduling_seconds | full compression duration (s) |
|---:|---:|---:|
| 1 | 0.742398 | 155.92 |
| 2 | 0.275263 | 165.578 |
| 3 | 0.849841 | 166.084 |
| 4 | 0.258035 | 127.227 |
| 5 | 1.874552 | 212.654 |
| 6 | 1.511318 | 112.755 |
| 7 | 0.746475 | 112.168 |
| 8 | 0.358103 | 112.835 |
| 9 | 0.783230 | 205.532 |
| 10 | 0.774677 | 205.212 |
| 11 | 0.447144 | 236.192 |

- **Average `scheduler_discovery_to_scheduling_seconds`**: 0.7837305455
- **Average full compression duration (seconds)**: 164.7415454545

## After

| job_id | scheduler_discovery_to_scheduling_seconds | full compression duration (s) |
| ------ | ----------------------------------------- | ----------------------------- |
| 1      | 0.600464                                  | 118.73                        |
| 2      | 0.083286                                  | 119.076                       |
| 3      | 0.066629                                  | 165.608                       |
| 4      | 0.141759                                  | 122.536                       |
| 5      | 0.108384                                  | 241.382                       |
| 6      | 0.183196                                  | 125.094                       |
| 7      | 0.449663                                  | 265.933                       |
| 8      | 1.558069                                  | 345.997                       |
| 9      | 0.091492                                  | 425.42                        |
| 10     | 0.074876                                  | 438.922                       |
| 11     | 0.174953                                  | 421.748                       |

- **Average** `scheduler_discovery_to_scheduling_seconds`: 0.321161
- **Average full compression duration (seconds)**: 253.6769090909

The drop in `scheduler_discovery_to_scheduling_seconds` makes sense, we no longer make a trip to S3 per job.
The increase in `full compression duration (s)` is strange but certainly not a result of this PR. This metric comes from the `duration` column in `compression_jobs`.

### Methodology

1. 11 copies of the [elasticsearch](https://zenodo.org/records/10516226) structured dataset were loaded into a Minio bucket called `elasticsearch`. 
2. On both branches, I changed the following lines to `components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py`:

On line 311:

```python
t0 = time.perf_counter()
```

On line 418:

```python
     duration_sec = time.perf_counter() - t0
            logger.info(
                "scheduler_discovery_to_scheduling_seconds=%.6f job_id=%s input_type=%s",
                duration_sec,
                job_id,
                input_config.type,
            )
```

These lines were added since our main focus is measuring the time from when the compression job starts to the file metadata all resolved.

### Setup

**Prereqs**: To follow the commands below, you MUST have `minio` downloaded & have >90 GB of free space. 

- Download the[elasticsearch](https://zenodo.org/records/10516226) structured dataset
- Create a folder called `test-data` 
- Using the minio webui create a bucket called `elasticsearch/`
- Put the uncompressed dataset in this bucket
- Make copies:
   -  `for i in $(seq 1 10); do cp -r elasticsearch.log "elasticsearch${i}.log"; done`
   - The command above assumes your dataset is also called `elasticsearch.log` and is a folder (this should be correct if you uploaded through the webui).

If you haven't done this already:

```bash
# in a separate terminal in the parent of the test-data folder
minio server ./test-data
```

- Create a CLP config that allows S3 AND sets up the log-ingestor AND you have updated your credentials with minio creds!

### Verifying

```bash
task
# IMPORTANT: UPDATE YOUR CREDENTIALS AND CLP.CONFIG WITH MINIO CREDS!
# ALSO MAKE SURE THE LOG-INGESTOR IS ENABLED!
build/clp-package/sbin/start-clp.sh
# alternatively you can do:
# python -m clp_package_utils.scripts.start_clp --config "PATH TO YOU CONFIG"
```

Create a S3 Scanner job (replace `<minio-host>`):

```bash
curl -X POST "http://127.0.0.1:3002/s3_scanner" \
  -d  '{"bucket_name":"elasticsearch","key_prefix":"elasticsearch","endpoint_url":"http://<minio-host>:9000","scanning_interval_sec":30}'
```

Verify it worked:

```bash
# see table entry
docker exec -it $(docker ps --filter name=database --format '{{.Names}}' | head -1) \
  mysql -u clp-user -p clp-db -e "SELECT id, status, status_msg, creation_time, start_time, update_time, duration, num_tasks, num_tasks_completed FROM compression_jobs;"
  
 # see worker logs
docker logs $(docker ps -a --filter name=compression-scheduler --format '{{.Names}}' | head -1) 2>&1 | tail -80
```

## Summary by CodeRabbit

* **New Features**
  * Added a structured S3 object‑metadata input config and a public ingestion‑job ID type.

* **Refactor**
  * Compression/ingestion now use lightweight buffer entries (id+size) and metadata IDs instead of full object records; submission APIs updated accordingly.

* **Tests**
  * Updated tests to use compression buffer entries and metadata‑ID workflows, and adjusted expected serialized payloads.


components/clp-rust-utils/src/job_config/clp_io_config.rs
```diff
@@ -10,14 +10,17 @@ pub struct ClpIoConfig {
     pub output: OutputConfig,
 }
 
+/// Represents the unique identifier for an ingestion job in CLP DB.
+pub type IngestionJobId = u64;
```
LinZhihao-723: I don't think this ID should belong to this mod.
Let's do the following:
- Move and rename this ID to `clp_rust_utils::job_config::ingestion::JobId`
- In `log_ingestor::ingestion_job`, add `pub use clp_rust_utils::job_config::ingestion::JobId as IngestionJobId`

components/clp-rust-utils/src/job_config/clp_io_config.rs
```diff
@@ -10,14 +10,17 @@ pub struct ClpIoConfig {
...
+
 /// An enum representing CLP input config.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize)]
 #[serde(tag = "type")]
 pub enum InputConfig {
-    #[serde(rename = "s3")]
-    S3InputConfig {
+    #[serde(rename = "s3_object_metadata")]
+    S3ObjectMetadataInputConfig {
```
LinZhihao-723, line 28:
```suggestion
    #[serde(rename = "s3")]
    S3InputConfig {
        #[serde(flatten)]
        config: S3InputConfig,
    },

    #[serde(rename = "s3_object_metadata")]
    S3ObjectMetadataInputConfig {
```
We shoudn't drop `S3InputConfig` yet, since it's a valid enum in our package, right?

components/clp-rust-utils/tests/clp_config_test.rs
```diff
@@ -42,20 +43,25 @@ fn test_clp_io_config_serialization() {
...
+
+    let expected = "1bdc0100c4aa350b081365c242113820d5873cbb21492ac6023d91a3f\
+        23f7593f5f195c76e4c5a03e42fdb9869162f8b4349b77947a7a45a53e95a85ab8362b8db53\
+        5eba172ae2a7000dce0626b90c8f12976d951d179058379f0af6f585537abbfd7\
+        4e16aab0b60697d67cb483bbe4ed93ef912f9a26c2852467682cc92d89dea296f\
+        cd9f8247cecf8b3823881cdb84997211e8d566c29bcc49daf198293a2447ab7b\
+        3015a2be9bca7da952a9f57ae8d42668559f51dd40ede0b957ab746b7da8e64f\
+        a6f792ff1fbe07cb97602af18f731b3015fa39495352d91d5389a5877bdf5828\
+        847b99c51df7c04e95bf6375c4dbd78f4b6c24e809642ed5ff542c16fb5a5cd\
+        1495253250d3841af440db6c9ca9e39a66d0034cfb2efd6083a204eb64a02";
```
LinZhihao-723:
```suggestion
    let expected = "1bdc0100c4aa350b081365c242113820d5873cbb21492ac6023d91a3f23f7593f5f195c76e\
        4c5a03e42fdb9869162f8b4349b77947a7a45a53e95a85ab8362b8db535eba172ae2a7000dce0626b90c8f12976\
        d951d179058379f0af6f585537abbfd74e16aab0b60697d67cb483bbe4ed93ef912f9a26c2852467682cc92d89d\
        ea296fcd9f8247cecf8b3823881cdb84997211e8d566c29bcc49daf198293a2447ab7b3015a2be9bca7da952a9f\
        57ae8d42668559f51dd40ede0b957ab746b7da8e64fa6f792ff1fbe07cb97602af18f731b3015fa39495352d91d\
        5389a5877bdf5828847b99c51df7c04e95bf6375c4dbd78f4b6c24e809642ed5ff542c16fb5a5cd1495253250d3\
        841af440db6c9ca9e39a66d0034cfb2efd6083a204eb64a02";
```

components/clp-rust-utils/src/job_config/clp_io_config.rs
```diff
@@ -33,6 +36,18 @@ pub struct S3InputConfig {
     pub unstructured: bool,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub struct S3ObjectMetadataInputConfig {
+    #[serde(flatten)]
+    pub s3_config: S3Config,
+
+    pub ingestion_job_id: IngestionJobId,
+    pub s3_object_metadata_ids: Option<Vec<S3ObjectMetadataId>>,
```
LinZhihao-723: Why would we allow this field to be optional?
Notice that in Python definition, this field is not nulllable: https://github.com/y-scope/clp/blob/aac93a487a5b47851b347be7a8d83ad47a4cc2a9/components/job-orchestration/job_orchestration/scheduler/job_config.py#L54

components/log-ingestor/src/compression/buffer.rs
```diff
@@ -1,6 +1,14 @@
 use anyhow::Result;
 use async_trait::async_trait;
-use clp_rust_utils::s3::ObjectMetadata;
+use clp_rust_utils::s3::S3ObjectMetadataId;
+use sqlx::FromRow;
+
+/// One entry of data fed into the compression buffer
```
LinZhihao-723:
```suggestion
/// Represents an entry of a buffered object metadata.
```

components/log-ingestor/src/compression/buffer.rs
```diff
@@ -14,18 +22,19 @@ pub trait BufferSubmitter {
     /// # Errors:
     ///
     /// Returns an [`anyhow::Error`] on failure.
-    async fn submit(&self, buffer: &[ObjectMetadata]) -> Result<()>;
+    async fn submit(&self, buffer: &[S3ObjectMetadataId]) -> Result<()>;
 }
 
-/// A buffer that accumulates object metadata and submits it when a size threshold is reached.
+/// A buffer that accumulates object metadata IDs and submits when the
+/// size threshold is reached.
```
LinZhihao-723:
```suggestion
/// A buffer that accumulates object metadata IDs and submits when the size threshold is reached.
```
Hmmm, we don't need wrap this comment. Might be worth to check ur IDE setting to see if this is triggered automatically by the IDE formatting.

components/log-ingestor/src/compression/compression_job_submitter.rs
```diff
@@ -67,15 +57,17 @@ impl CompressionJobSubmitter {
...
             s3_config: S3Config {
                 bucket: ingestion_job_config.bucket_name.clone(),
                 region_code: ingestion_job_config.region.clone(),
                 key_prefix: ingestion_job_config.key_prefix.clone(),
                 endpoint_url: ingestion_job_config.endpoint_url.clone(),
                 aws_authentication,
             },
-            keys: None,
+            ingestion_job_id,
+            s3_object_metadata_ids: None,
```
LinZhihao-723: Since we don't want to make this field optional, probably leave it as an empty vector.

components/log-ingestor/src/compression/compression_job_submitter.rs
```diff
@@ -164,14 +156,14 @@ pub async fn wait_for_compression_job_completion_and_update_metadata(
...
+    object_metadata_ids: Vec<S3ObjectMetadataId>,
 ) {
     let ingestion_job_id = state.get_ingestion_job_id();
-    let num_objects_submitted = id_and_key_pairs.len();
+    let num_objects_submitted = object_metadata_ids.len();
     tracing::info!(ingestion_job_id = ? ingestion_job_id, "Submitting CLP compression job.");
 
     let compression_job_id = match state
-        .submit_for_compression(io_config_template, id_and_key_pairs)
+        .submit_for_compression(io_config_template, &object_metadata_ids)
```
LinZhihao-723: I don't think we need to pass object metadata IDs as a reference: it's not efficient.
- It's been copied from a slice into a vector when calling `submit_clp_compression_job_and_wait_for_completion`.
- The reference here is passed all the way down to `ClpCompressionState::submit_for_compression` to reconstruct a vector.
  - Looks like u could just pass the ownership all the way down to the compression state instead.

LinZhihao-723: And in fact, why not construct the io config here (outside of the CLP compression state)?
Originally, we put the logic inside `ClpCompressionState::submit_for_compression` because the object metadata -> ID mapping behavior wasn't well-formed.
Now, I think we could create the io-config by embedding `object_metadata_ids` directly into the field (it can happen even before this function, so this call only takes `io_config: ClpIoConfig` and submits it).
And what we pass into `ClpCompressionState` will be a well-formed config at the first place, and it can decide how to handle different types of input configs. For example, if we need to add Fs compression config, the same method should also be able to handle.

LinZhihao-723 (changes requested):
Reviewd:
- clp_io_config (clp-rust-utils)
- s3 (clp-rust-utils)
- clp_config_test (clp-rust-utils)
- buffer (log-ingestor)
- compression_job_submitter (log-ingestor)

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -902,15 +887,18 @@ impl ClpIngestionState {
...
             {
-                object.id = Some(next_metadata_id);
+                buffer_entries.push(CompressionBufferEntry {
+                    id: next_metadata_id,
+                    size: object.size,
+                });
             }
         }
-        self.sender.send(objects).await?;
+        self.sender.send(buffer_entries).await?;
```
LinZhihao-723, line 884: This piece of code doesn't make sense anymore with the changes in this PR:
- We had to do it this way because previously if we let `ingest_s3_object_metadata` to reconstruct a vector of object metadata it would be too inefficient.
- However, for now, it makes more sense in terms of efficiency and ownership to let `ingest_s3_object_metadata` take the ownership of `object`, and return a vector of `CompressionBufferEntry`. There's no need to know the chunk size outside of `ingest_s3_object_metadata` anymore.

components/log-ingestor/src/compression/listener.rs
```diff
@@ -1,32 +1,32 @@
...
 impl<Submitter: BufferSubmitter + Send + 'static> ListenerTask<Submitter> {
-    /// Runs the listener task to buffer and submit S3 object metadata. Submission can be triggered
-    /// in three ways:
+    /// Runs the listener task to buffer and submit [`CompressionBufferEntry`] values. Submission
+    /// can be triggered in three ways:
     /// * Receiving a cancellation signal via the provided [`cancel_token`].
-    /// * The buffer capacity is reached after receiving a new object metadata.
-    /// * A timeout occurs without receiving new object metadata.
+    /// * The buffer's size threshold is reached after receiving new entries.
+    /// * A timeout occurs without receiving new entries.
```
LinZhihao-723, line 31:
```suggestion
    /// Runs the listener task to buffer and submit [`CompressionBufferEntry`] values. Submission
    /// can be triggered if:
    ///
    /// * Receiving a cancellation signal via the provided [`cancel_token`].
    /// * All senders are closed.
    /// * The buffer's size threshold is reached after receiving new entries.
    /// * A timeout occurs without receiving new entries.
```
Just realized:
- We were missing an empty line for the bullet points.
- We're missing one submission condition.

components/log-ingestor/src/compression/listener.rs
```diff
@@ -1,32 +1,32 @@
...
     time::{Instant, Sleep, sleep_until},
 };
 use tokio_util::sync::CancellationToken;
 
-use crate::compression::{Buffer, BufferSubmitter};
+use crate::compression::{Buffer, BufferSubmitter, CompressionBufferEntry};
 
-/// Represents a listener task that buffers and submits S3 object metadata.
+/// Represents a listener task that buffers incoming [`CompressionBufferEntry`] values and submits
+/// when full or on timeout.
```
LinZhihao-723:
```suggestion
/// when a certain size threshold is reached or on timeout.
```

components/log-ingestor/tests/test_compression_listener.rs
```diff
@@ -76,24 +82,18 @@ async fn test_compression_listener() -> Result<()> {
...
-    let objs1 = bucket_0_objects.clone();
-    let objs2 = bucket_1_objects[..100].to_vec();
-    let objs3 = bucket_1_objects[100..].to_vec();
-
-    let h1 = tokio::spawn(async move { send_to_listener(objs1, sender1).await });
-    let h2 = tokio::spawn(async move { send_to_listener(objs2, sender2).await });
-    let h3 = tokio::spawn(async move { send_to_listener(objs3, sender3).await });
+    let h1 = tokio::spawn(async move { send_to_listener(entries1, sender1).await });
+    let h2 = tokio::spawn(async move { send_to_listener(entries2, sender2).await });
+    let h3 = tokio::spawn(async move { send_to_listener(entries3, sender3).await });
```
LinZhihao-723: Might be a better way to refactor this:
```suggestion
    let expected_ids: Vec<S3ObjectMetadataId> = (1..=300).collect();
    let mut handlers = Vec::new();
    for ids in expected_ids.chunks(100) {
        let entries = create_test_buffer_entries(ids);
        let sender = listener.get_new_sender();
        let h = tokio::spawn(async move { send_to_listener(entries, sender).await });
        handlers.push(h);
    }

    // Wait for all sender tasks to finish
    for handler in handlers {
        handler.await.expect("sender task panicked");
    }
```
- We move `expected_ids` to the front.
- We change `create_test_buffer_entries` to the following:
  ```rust
  fn create_test_buffer_entries(
      ids: &[S3ObjectMetadataId],
  ) -> Vec<CompressionBufferEntry> {
      ids.iter()
          .map(|id| CompressionBufferEntry {
              id: *id,
              size: TEST_OBJECT_SIZE,
          })
          .collect()
  }
  ```

components/log-ingestor/tests/test_ingestion_job.rs
LinZhihao-723, line 1: This file needs a bit more refactoring (I should have said this in the original problem statement for #2018 but I didn't make it clear). These test cases were written in this way because we used to have no "state" concept for an ingestion job. So when these test cases were written, the ingestion flow is: ingestion jobs -> channel senders. Later when we add the state, we didn't up this test case on time so it looks like ingestion jobs -> state -> channel senders, so we can still do all the check inside the channel receiver end. I think this is inefficient and confusing and we should change it to directly use the state as a sort of shared buffer for checking the ingested files. A possible implementation would be to implement `SqsListenerState` and `S3ScannerState` in this test file to hold a shared buffer protected by a mutex, and use this version of state implementation to replace the current use of `ZeroFaultToleranceIngestionJobState`.
That said, I do realize this is out of the scope for the current PR. But I think we should do this ASAP before we all forget about it, lol. Can you do a follow-up PR after this one to fix it?

LinZhihao-723 (changes requested):

components/log-ingestor/tests/test_ingestion_job.rs
LinZhihao-723: @coderabbitai Can you create an issue to track this?

components/log-ingestor/tests/test_compression_listener.rs
```diff
@@ -110,16 +102,13 @@ async fn test_compression_listener() -> Result<()> {
...
-    actual_total_submitted.sort();
+    let mut actual_ids: Vec<S3ObjectMetadataId> =
+        submitted_buffers.iter().flatten().copied().collect();
+    actual_ids.sort_unstable();
 
-    let mut expected_total_submitted = Vec::new();
-    expected_total_submitted.extend_from_slice(&bucket_0_objects);
-    expected_total_submitted.extend_from_slice(&bucket_1_objects);
-    expected_total_submitted.sort();
+    let expected_ids: Vec<S3ObjectMetadataId> = (1..=300).collect();
```
LinZhihao-723:
```suggestion
```

components/log-ingestor/tests/test_compression_listener.rs
```diff
@@ -21,50 +26,48 @@ impl TestBufferSubmitter {
...
         .await
-        .expect("Failed to send objects to listener");
+        .expect("Failed to send buffer entries to listener");
 }
 
-/// Creates a vector of [`ObjectMetadata`] objects for a given bucket. Each object will have a
-/// unique key and a fixed size of [`TEST_OBJECT_SIZE`].
+/// Creates [`CompressionBufferEntry`] values for testing. IDs start at `id_start`.
+/// [`TEST_OBJECT_SIZE`].
 ///
```
LinZhihao-723:
```suggestion
```
Should be ok to just drop them all.

LinZhihao-723 (changes requested):
Will do some package-level testing before approving.

LinZhihao-723 (changes requested):
Verified that for both the S3 scanner and the SQS listener, the ingested objects can be submitted for compression, and the compression results will be reflected properly after the job finishes.
However, there's an issue in the WebUI where the compression jobs are not showing properly. This is because the new S3 object metadata compression job config is not added into the WebUI component. Already located the root cause and had a quick fix with @junhaoliao. Need to decide whether we want to submit WebUI changes as a separate PR or as a part of this PR.

LinZhihao-723 (commented):
Pushed WebUI changes directly. @junhaoliao Can you help check?

junhaoliao (commented):
now that we have created this new job type, do you think we can remove the `keys[]` field in `S3InputConfig`?

components/webui/common/src/schemas/compression.ts
```diff
@@ -110,6 +111,18 @@ const ClpIoS3InputConfigSchema = Type.Object({
     unstructured: Type.Boolean(),
 });
 
+/**
+ * Matching `S3ObjectMetadataInputConfig` in `job_orchestration.scheduler.job_config`.
```
junhaoliao, line 115: (just a note for awareness - no code changes are needed in this PR)

The "matching" comment is misleading because what we have here is only a subset of job_orchestration.scheduler.job_config fields. That said, i do realize that has been a preexisting issue for the other types as well.

Currently this Type object is only used to derive a TypeScript type for type hints, so we shouldn't run into any issues. However, once we add the input form for s3 then the other type objects will be used to validate against inputs, and validation errors will be raised on additional fields that are not listed. anyways, ClpIoS3ObjectMetadataInputConfig is an internal input config and will not be used in any input form validation in the foreseeable future so no code changes are needed

junhaoliao (approved):
the webui changes lgtm

LinZhihao-723:
> now that we have created this new job type, do you think we can remove the `keys[]` field in `S3InputConfig`?

The `keys` field is actually used for key ingestion path: https://docs.yscope.com/clp/main/user-docs/guides-using-object-storage/aws-s3/using-clp-with-aws-s3.html#s3-object-compression-mode. I think we can drop it once we move all S3 ingestion into log-ingestor.

components/webui/common/src/schemas/compression.ts
LinZhihao-723: I agree. We should probably create a minimum schema that contains the fields needed.
Shall we ask rabbit to create an issue to track this renaming?

LinZhihao-723 (approved):
For the PR title, how about:
```
feat(log-ingestor): Use `S3ObjectMetadataInputConfig` for compression job submission (resolves #2018).
```

PR #2005, opened 2026-02-17
feat(kv-ir): Improve error handling in `clp::ffi::ir_stream::ir_unit_deserialization_methods`:

jonathan-imanu:
# Description

This PR is the first step in improving the error handling in `clp::ffi::ir_stream::Deserializer.hpp`. Following the suggestion of #904, the end goal is to use the `Result<T, IrErrorCode>` as the return type instead of using `std::errc`. 

Since this is a larger refactor so the first step is changing functions in `components/core/src/clp/ffi/ir_stream/ir_unit_deserialization_methods.cpp` to return `Result<T, IrErrorCode>` instead of the C-style `IRErrorCode`.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

- Unit tests and integration tests

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **New Features**
  * Introduced a dedicated IR deserialization error type with expanded error kinds (including CorruptedIR and UnsupportedFormat).

* **Bug Fixes**
  * Improved detection and reporting for truncated, corrupted or unsupported IR data; failure modes mapped to specific deserialization error codes.

* **Refactor**
  * Deserialization routines migrated to a uniform Result-based error flow and now propagate the new deserialization error values.

* **Documentation**
  * Clarified and corrected wording in deserialization docs and error descriptions.

* **Chores**
  * Build configuration updated to include the new deserialization error implementation.

Issue #904, opened 2025-05-15
Improve error handling in `clp::ffi::ir_stream::Serializer/Deserializer`


LinZhihao-723:
### Request

The error handling system for `clp::ffi::ir_stream::Serializer/Deserializer` can be improved with the new error code system introduced in `ystdlib::error_handling`.

### Possible implementation

* For serializer: We should improve the error handling by changing the API to return a void result instead of a simple boolean indicating success/failure.
* For deserializer: We should integrate our own error code instead of using `std::errc`.

PR #2089, opened 2026-03-11
feat(kv-ir): Improve error handling in `clp::ffi::ir_stream::decoding_methods`:

jonathan-imanu:
# Description

This PR is the second step in improving the error handling in `clp::ffi::ir_stream::Deserializer.hpp`. Following the suggestion of #904, the end goal is to use the `Result<T, IrErrorCode>` as the return type instead of using `std::errc`. This PR builds off of the work done in #2005. 

Since this is a larger refactor, this step focuses changing functions in `components/core/src/clp/ffi/ir_stream/decoding_methods.cpp` to return `Result<T, IrErrorCode>` instead of the C-style `IRErrorCode`.

# Checklist

  breaking change.

# Validation performed

- Workflows
- Unit tests

## Summary by CodeRabbit

* **Refactor**
  * Internal deserialization was modernized for more consistent error handling and clearer failure propagation.

* **Bug Fixes**
  * Improved detection and handling of incomplete or corrupted IR streams to reduce silent failures.
  * Added a distinct error case for invalid magic numbers.

* **User-facing**
  * Error messages for malformed IR data are clearer and more informative, improving diagnostics and test reliability.


components/core/src/clp/ffi/ir_stream/decoding_methods.hpp
```diff
@@ -54,20 +54,13 @@ class DecodingException : public TraceableException {
...
- * @return IRErrorCode_Success on success
- * @return IRErrorCode_Incomplete_IR if reader doesn't contain enough data to deserialize
+ * @return A void result on success, or an error code indicating the failure:
+ * - IrDeserializationErrorEnum::IncompleteStream if reader doesn't contain enough data to
+ *   deserialize.
+ * - IrDeserializationErrorEnum::InvalidMagicNumber if reader contains an invalid magic number.
  */
-[[nodiscard]] IRErrorCode deserialize_tag(ReaderInterface& reader, encoded_tag_t& tag);
+auto get_encoding_type(ReaderInterface& reader, bool& is_four_bytes_encoding)
+        -> ystdlib::error_handling::Result<void>;
```
LinZhihao-723: This would be a bad practice we should try to avoid: if we're using result-based error handling, then we should use the result to return `is_four_bytes_encoding`.
- The easy fix is to make the return type `ystdlib::error_handling::Result<bool>`
- However, it's still a bit hacky and personally I'd create an enum named `EncodingType` which contains the encoding type we support.

Depends on u whether u want to keep this PR to step 1, or go further to step 2.

LinZhihao-723 (changes requested):
Will stop reading as I realized many API changes in this PR don't make sense as explained in the comment.
And in fact, I realized most of these APIs aren't needed anymore once we support using the kv-ir deserializer to deserialize the old IR format.
That said, I think the right way to do is probably:
- Leave those APIs not used by kv-ir deserializer untouched in this PR, and only add result-style versions for those used by kv-ir deserializer (for example, `deserialize_int`).
- Work on supporting the old IR format in the kv-ir deserializer next.
- Use kv-ir deserializer to implement `ir::LogEventDeserializer`.
- Move the rest unused APIs to a new namespace for deprecation.

Let me know ur thoughts.

components/core/src/clp/ffi/ir_stream/decoding_methods.cpp
```diff
@@ -605,44 +591,54 @@ auto deserialize_tag(ReaderInterface& reader) -> ystdlib::error_handling::Result
     return tag;
 }
 
-IRErrorCode deserialize_preamble(
+auto deserialize_preamble(
```
jonathan-imanu: This is only used in tests. Should we remove it and update the tests? If not, is it worth changing the return type to a struct to avoid what u mentioned [here](https://github.com/y-scope/clp/pull/2089#discussion_r2934543109)

components/core/src/clp/ffi/ir_stream/decoding_methods.hpp
```diff
@@ -53,12 +58,13 @@ class DecodingException : public TraceableException {
...
- * @return ErrorCode_Corrupted_IR if reader contains invalid IR
- * @return ErrorCode_Incomplete_IR if reader doesn't contain enough data to decode
+ * @return A result containing the encoding type on success, or an error code indicating the
+ * failure:
+ * - IrDeserializationErrorEnum::IncompleteStream if reader doesn't contain enough data to decode
+ * - IrDeserializationErrorEnum::InvalidMagicNumber if the magic number is invalid
  */
-IRErrorCode get_encoding_type(ReaderInterface& reader, bool& is_four_bytes_encoding);
+[[nodiscard]] auto get_encoding_type(ReaderInterface& reader)
+        -> ystdlib::error_handling::Result<EncodingType>;
```
LinZhihao-723, line 78: I'd suggest to add a new signature for the result-based `get_encoding_type`:
- The current implementation is very problematic (for some legacy reason), lol. I will give u suggestion in another comment.
- This method is used by some of the ffi libraries. Keep the old signature untouched would make it easier when we need to upgrade CLP versions in ffi library build.

components/core/src/clp/ffi/ir_stream/decoding_methods.hpp
```diff
@@ -53,12 +58,13 @@ class DecodingException : public TraceableException {
...
  * Deserializes the IR stream's encoding type
  * @param reader
- * @param is_four_bytes_encoding Returns the encoding type
- * @return ErrorCode_Success on success
- * @return ErrorCode_Corrupted_IR if reader contains invalid IR
- * @return ErrorCode_Incomplete_IR if reader doesn't contain enough data to decode
+ * @return A result containing the encoding type on success, or an error code indicating the
+ * failure:
+ * - IrDeserializationErrorEnum::IncompleteStream if reader doesn't contain enough data to decode
+ * - IrDeserializationErrorEnum::InvalidMagicNumber if the magic number is invalid
```
LinZhihao-723: This docstring's style is also outdated. Use the following rewrite and keep the original untouched.
```C++
/**
 * Reads the IR stream's encoding type from the magic number.
 * @param reader
 * @return A result containing the encoding type on success, or an error code indicating the
 * failure:
 * - IrDeserializationErrorEnum::IncompleteStream if `reader` doesn't contain enough data to decode
 *   the magic number.
 * - IrDeserializationErrorEnum::InvalidMagicNumber if the magic number is invalid.
 */
```

components/core/src/clp/ffi/ir_stream/decoding_methods.hpp
```diff
@@ -186,16 +192,17 @@ void generic_decode_message(
...
+ * - Forwards `deserialize_metadata`'s return values on failure.
  */
-IRErrorCode deserialize_preamble(
+[[nodiscard]] auto deserialize_preamble(
         ReaderInterface& reader,
         encoded_tag_t& metadata_type,
         size_t& metadata_pos,
         uint16_t& metadata_size
-);
+) -> ystdlib::error_handling::Result<void>;
```
LinZhihao-723: Similarly, let's keep this method untouched.

components/core/src/clp/ffi/ir_stream/decoding_methods.hpp
```diff
@@ -212,6 +219,18 @@ IRErrorCode deserialize_preamble(
...
 
+/**
+ * Deserializes the preamble for an IR stream.
+ * @param reader
+ * @return A result containing a pair with the metadata type and metadata size on success, or an
+ * error code indicating the failure:
+ * - IrDeserializationErrorEnum::IncompleteStream if reader doesn't contain enough data to
+ *   deserialize
+ * - Forwards `deserialize_metadata`'s return values on failure.
+ */
```
LinZhihao-723, line 246:
```suggestion
/**
 * Deserializes the preamble for an IR stream.
 * @param reader
 * @return A result containing a pair on success, or an error code indicating the failure:
 * - The pair:
 *   - The tag indicating the type of the deserialized metadata.
 *   - The payload of the deserialized metadata.
 * - The possible error codes:
 *   - IrDeserializationErrorEnum::IncompleteStream if `reader` doesn't contain enough data to
 *     deserialize.
 *   - Forwards `deserialize_metadata`'s return values on failure.
 */
```

components/core/src/clp/ffi/ir_stream/decoding_methods.hpp
```diff
@@ -212,6 +219,18 @@ IRErrorCode deserialize_preamble(
...
+auto deserialize_preamble(ReaderInterface& reader)
```
LinZhihao-723:
```suggestion
[[nodiscard]] auto deserialize_preamble(ReaderInterface& reader)
```

components/core/src/clp/ffi/ir_stream/decoding_methods.cpp
```diff
@@ -416,39 +412,30 @@ static IRErrorCode generic_deserialize_log_event(
     return IRErrorCode_Success;
 }
 
-static IRErrorCode deserialize_metadata(
-        ReaderInterface& reader,
-        encoded_tag_t& metadata_type,
-        uint16_t& metadata_size
-) {
+auto deserialize_metadata(ReaderInterface& reader)
```
LinZhihao-723: As metioned in previous, we probably want to leave the old signature untouched. Let's move the logic inside this helper to the new result-based version and fix some linter errors. I will comment the expected version in another comment.

components/core/src/clp/ffi/ir_stream/decoding_methods.cpp
```diff
@@ -605,44 +591,54 @@ auto deserialize_tag(ReaderInterface& reader) -> ystdlib::error_handling::Result
...
     if (ErrorCode_Success
         != reader.try_read_exact_length(
                 size_checked_pointer_cast<char>(metadata.data()),
                 metadata_size
         ))
     {
-        return IRErrorCode_Incomplete_IR;
+        return IrDeserializationError{IrDeserializationErrorEnum::IncompleteStream};
+    }
+    return std::pair{metadata_type, std::move(metadata)};
```
LinZhihao-723: Suggested implementation:
```suggestion
    auto const metadata_type{YSTDLIB_ERROR_HANDLING_TRYX(deserialize_tag(reader))};
    auto const encoded_tag{YSTDLIB_ERROR_HANDLING_TRYX(deserialize_tag(reader))};
    size_t metadata_size{};
    switch (encoded_tag) {
        case cProtocol::Metadata::LengthUByte: {
            metadata_size = YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int<uint8_t>(reader));
            break;
        }
        case cProtocol::Metadata::LengthUShort: {
            metadata_size = YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int<uint16_t>(reader));
            break;
        }
        default:
            return IrDeserializationError{IrDeserializationErrorEnum::UnsupportedMetadataFormat};
    }

    std::vector<int8_t> metadata(metadata_size, 0);
    if (ErrorCode_Success
        != reader.try_read_exact_length(
                size_checked_pointer_cast<char>(metadata.data()),
                metadata_size
        ))
    {
        return IrDeserializationError{IrDeserializationErrorEnum::IncompleteStream};
    }
    return {metadata_type, std::move(metadata)};
```
You might need to update the docstring's error section accordingly.

components/core/src/clp/ffi/ir_stream/decoding_methods.cpp
LinZhihao-723: It might be used in ffi libraries (not in this repo).

components/core/src/clp/ffi/ir_stream/decoding_methods.cpp
```diff
@@ -568,26 +555,25 @@ template <ir::EncodedVariableTypeReq encoded_variable_t>
...
                         cProtocol::MagicNumberLength
                 )))
     {
-        is_four_bytes_encoding = false;
+        return EncodingType::EightByte;
     } else {
-        return IRErrorCode_Corrupted_IR;
+        return IrDeserializationError{IrDeserializationErrorEnum::InvalidMagicNumber};
     }
-    return IRErrorCode_Success;
```
LinZhihao-723, line 590: We have some legacy issues of not using constexpr `std::array` for `cProtocol::FourByteEncodingMagicNumber/EightByteEncodingMagicNumber`. This is what I can think of the best if we don't change them:
```C++
    std::array<char, cProtocol::MagicNumberLength> buffer{};
    if (ErrorCode_Success != reader.try_read_exact_length(buffer.data(), buffer.size())) {
        return IrDeserializationError{IrDeserializationErrorEnum::IncompleteStream};
    }
    auto comparator = [](char lhs, int8_t rhs) -> bool { return static_cast<int8_t>(lhs) == rhs; };
    if (std::ranges::equal(buffer, std::span{cProtocol::FourByteEncodingMagicNumber}, comparator)) {
        return EncodingType::FourByte;
    }
    if (std::ranges::equal(buffer, std::span{cProtocol::EightByteEncodingMagicNumber}, comparator))
    {
        return EncodingType::EightByte;
    }
    return IrDeserializationError{IrDeserializationErrorEnum::InvalidMagicNumber};
```
It might be worth to open an issue to track of this through. @coderabbitai, can you create an issue to keep track of `cProtocol::FourByteEncodingMagicNumber/EightByteEncodingMagicNumber` using C-style array? We should replace it with constexpr std::array.

components/core/src/clp/ffi/ir_stream/decoding_methods.cpp
LinZhihao-723, line 1: I think we should revert the changes except the methods I commented.

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -238,26 +241,16 @@ auto Deserializer<IrUnitHandlerType, QueryHandlerType>::create_generic(
...
-    encoded_tag_t metadata_type{};
-    if (auto const err{deserialize_preamble(reader, metadata_type, metadata)};
-        IRErrorCode::IRErrorCode_Success != err)
-    {
-        return ir_error_code_to_errc(err);
-    }
+    [[maybe_unused]] auto const encoding_type{
+            YSTDLIB_ERROR_HANDLING_TRYX(get_encoding_type(reader))
+    };
+    auto const metadata_pair{YSTDLIB_ERROR_HANDLING_TRYX(deserialize_preamble(reader))};
```
LinZhihao-723:
```suggestion
    auto const [metadata_type, metadata]{YSTDLIB_ERROR_HANDLING_TRYX(deserialize_preamble(reader))};
```
In 99% percent of the case you should use struct binding to avoid using `.first` and `.second`, which makes the readability really bad.

components/core/src/clp/ffi/ir_stream/ir_unit_deserialization_methods.cpp
```diff
@@ -225,58 +228,33 @@ auto deserialize_schema_tree_node_key_name(ReaderInterface& reader, std::string&
...
-        }
+        uint16_t length{YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int<uint16_t>(reader))};
         str_length = static_cast<size_t>(length);
     } else if (cProtocol::Payload::StrLenUInt == tag) {
-        uint32_t length{};
-        if (false == deserialize_int(reader, length)) {
-            return IrDeserializationError{IrDeserializationErrorEnum::IncompleteStream};
-        }
+        uint32_t length{YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int<uint32_t>(reader))};
         str_length = static_cast<size_t>(length);
```
LinZhihao-723:
```C++
        str_length = static_cast<size_t>(
                YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int<uint8_t>(reader))
        );
    } else if (cProtocol::Payload::StrLenUShort == tag) {
        str_length = static_cast<size_t>(
                YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int<uint16_t>(reader))
        );
    } else if (cProtocol::Payload::StrLenUInt == tag) {
        str_length = static_cast<size_t>(
                YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int<uint32_t>(reader))
        );
```
They should be refactored in this way to improve readability.

components/core/src/clp/ffi/ir_stream/ir_unit_deserialization_methods.cpp
```diff
@@ -363,17 +341,13 @@ auto deserialize_value_and_insert_to_node_id_value_pairs(
         case cProtocol::Payload::ValueInt16:
         case cProtocol::Payload::ValueInt32:
         case cProtocol::Payload::ValueInt64: {
-            value_int_t value_int{};
-            YSTDLIB_ERROR_HANDLING_TRYV(deserialize_int_val(reader, tag, value_int));
+            value_int_t value_int{YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int_val(reader, tag))};
```
LinZhihao-723:
```suggestion
            auto const value_int{YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int_val(reader, tag))};
```
- Use `auto` whenever possible.
- `const` is missing in many places, please be aware of the const-correctness practice.

components/core/src/clp/ffi/ir_stream/ir_unit_deserialization_methods.cpp
```diff
@@ -363,17 +341,13 @@ auto deserialize_value_and_insert_to_node_id_value_pairs(
...
             break;
         }
         case cProtocol::Payload::ValueFloat: {
-            uint64_t val{};
-            if (false == deserialize_int(reader, val)) {
-                return IrDeserializationError{IrDeserializationErrorEnum::IncompleteStream};
-            }
-            node_id_value_pairs.emplace(node_id, Value{bit_cast<value_float_t>(val)});
+            auto const raw_bits{YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int<uint64_t>(reader))};
+            node_id_value_pairs.emplace(node_id, Value{bit_cast<value_float_t>(raw_bits)});
```
LinZhihao-723:
```suggestion
            node_id_value_pairs.emplace(
                    node_id,
                    Value{bit_cast<value_float_t>(
                            YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int<uint64_t>(reader))
                    )}
            );
```

components/core/src/clp/ffi/ir_stream/IrDeserializationError.hpp
LinZhihao-723: Sorry for missing this (wrote too much Rust and I forgot C++ practice, lol). We should make sure the first enum in the error code starts from 1. Otherwise it will be considered as a "success" code when converting into a boolean var.
We need to do it for `IrSerializationErrorEnum` as well.
```suggestion
    DuplicateKey = 1,
```

components/core/src/clp/ffi/ir_stream/utils.hpp
```diff
@@ -53,6 +53,19 @@ auto serialize_int(integer_t value, std::vector<int8_t>& output_buf) -> void;
 template <IntegerType integer_t>
 [[nodiscard]] auto deserialize_int(ReaderInterface& reader, integer_t& value) -> bool;
 
+/**
+ * Deserializes an integer from the given reader
+ * @tparam integer_t Type of the integer to deserialize
```
LinZhihao-723:
```suggestion
 * Deserializes an integer from the given reader.
 * @tparam integer_t The type of the integer to deserialize
```
Period

components/core/src/clp/ffi/ir_stream/utils.hpp
```diff
@@ -175,6 +188,25 @@ auto deserialize_int(ReaderInterface& reader, integer_t& value) -> bool {
     return true;
 }
 
+template <IntegerType integer_t>
+auto deserialize_int(ReaderInterface& reader) -> ystdlib::error_handling::Result<integer_t> {
+    integer_t value_little_endian;
```
LinZhihao-723:
```suggestion
    integer_t value_little_endian{};
```

components/core/src/clp/ffi/ir_stream/utils.hpp
```diff
@@ -175,6 +188,25 @@ auto deserialize_int(ReaderInterface& reader, integer_t& value) -> bool {
...
+    constexpr auto cReadSize = sizeof(integer_t);
+    if constexpr (cReadSize == 1) {
+        return value_little_endian;
+    } else if constexpr (cReadSize == 2) {
+        return bswap_16(value_little_endian);
+    } else if constexpr (cReadSize == 4) {
+        return bswap_32(value_little_endian);
+    } else if constexpr (cReadSize == 8) {
+        return bswap_64(value_little_endian);
+    }
```
LinZhihao-723, line 213:
```suggestion
    } else {
        static_assert(false, "unsupported integer size");
    }
```

LinZhihao-723 (changes requested):
Didn't check the test updates yet. Let's fix these issues first.

components/core/src/clp/ffi/ir_stream/decoding_methods.hpp
```diff
@@ -212,6 +228,26 @@ IRErrorCode deserialize_preamble(
...
+ * - The possible error codes:
+ *   - IrDeserializationErrorEnum::UnsupportedMetadataFormat if the metadata format is not
+ *supported.
+ *   - IrDeserializationErrorEnum::IncompleteStream if `reader` doesn't contain enough data to
+ *     deserialize.
+ *   - Forwards `deserialize_int`'s return values on failure.
+ *   - Forwards `deserialize_metadata`'s return values on failure.
+ *   - Forwards `deserialize_tag`'s return values on failure.
+
+ */
```
LinZhihao-723:
```suggestion
 *     supported.
 *   - IrDeserializationErrorEnum::IncompleteStream if `reader` doesn't contain enough data to
 *     deserialize.
 *   - Forwards `deserialize_int`'s return values on failure.
 *   - Forwards `deserialize_metadata`'s return values on failure.
 *   - Forwards `deserialize_tag`'s return values on failure.
 */
```
Fix formatting.

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -293,11 +286,7 @@ auto Deserializer<IrUnitHandler, QueryHandlerType>::deserialize_next_ir_unit(
         return std::errc::operation_not_permitted;
     }
 
-    encoded_tag_t tag{};
-    if (auto const err{deserialize_tag(reader, tag)}; IRErrorCode::IRErrorCode_Success != err) {
-        return ir_error_code_to_errc(err);
-    }
-
+    encoded_tag_t tag{YSTDLIB_ERROR_HANDLING_TRYX(deserialize_tag(reader))};
```
LinZhihao-723:
```suggestion
    auto const tag{YSTDLIB_ERROR_HANDLING_TRYX(deserialize_tag(reader))};
```

components/core/src/clp/ffi/ir_stream/decoding_methods.cpp
```diff
@@ -646,6 +662,36 @@ IRErrorCode deserialize_preamble(
...
+    std::vector<int8_t> metadata(metadata_size, 0);
+    if (ErrorCode_Success
+        != reader.try_read_exact_length(
+                size_checked_pointer_cast<char>(metadata.data()),
+                metadata_size
+        ))
+    {
+        return IrDeserializationError{IrDeserializationErrorEnum::IncompleteStream};
+    }
+    return std::pair<encoded_tag_t, std::vector<int8_t>>{metadata_type, std::move(metadata)};
```
LinZhihao-723:
```suggestion
    return std::pair{metadata_type, std::move(metadata)};
```
Compiler should be able to deduct types for u.

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
LinZhihao-723: Just realize we need to update the docstring for `deserialize_next_ir_unit`, but let's defer it to the next PR then.

LinZhihao-723 (changes requested):
Some last minute catches otherwise we're good.

LinZhihao-723 (approved):
For the commit message, how about:

```
feat(kv-ir): Improve error handling in `clp::ffi::ir_stream::decoding_methods`:
* Add `Result`-based error propagation variants of decoding APIs.
* Refactor decoding API implementation to apply C++20 standard.
```

PR #2101, opened 2026-03-17
feat(clp-s::filter): Add abstraction layer for filter building, serialization, and deserialization.

ShangDanLuXian:
# Description

This PR adds a filter serialization layer for `clp_s::filter`.

It introduces:
- `FilterBuilder` for constructing a filter from input values and serializing it
- `FilterReader` for reading a serialized filter and performing membership checks
- shared filter metadata/helpers in `FilterOptions`, including:
  - filter type parsing
  - filter normalization parsing
  - shared filter file magic
  - shared normalization logic

The current implementation is still backed by `BloomFilter`. Making this layer fully generic would
likely require a parent abstraction representing filter implementations, but since we currently only
have one filter type, this PR intentionally skips that refactor to keep the scope small.

This PR also:
- adds filter-specific error codes for unsupported filter type / normalization and corrupt filter
  file metadata
- wires the new filter library files into the build
- adds the required `clp::string_utils` dependency for shared normalization logic

# Checklist

  breaking change.

# Validation performed

Since this change mainly adds file I/O for the filter serialization layer, I am leaving the main
validation to integration tests.

## Summary by CodeRabbit

* **New Features**
  * Full filter support: create, populate, serialize and read Bloom filters; membership and wildcard-aware query handling; string normalization (None, Lowercase) and parsing utilities.

* **Bug Fixes**
  * Improved input/read validation and clearer, specific error responses for unsupported filter types/normalizations and read failures.

* **Chores**
  * Build updated to include new filter components and string utilities.


components/core/src/clp_s/filter/FilterOptions.cpp
```diff
@@ -0,0 +1,51 @@
...
+namespace clp_s::filter {
+auto try_parse_filter_type_id(uint8_t filter_type_id) -> std::optional<FilterType> {
+    if (static_cast<uint8_t>(FilterType::Bloom) == filter_type_id) {
+        return FilterType::Bloom;
+    }
+
+    return std::nullopt;
+}
+
+auto try_parse_filter_type(std::string_view filter_type_string) -> std::optional<FilterType> {
```
ShangDanLuXian, line 14: This one will be used during integration to reject filter writing when processing command line arguments.

Probably I also need a method to return all the available filter type strings so that can show in help.

components/core/src/clp_s/filter/FilterBuilder.hpp
```diff
@@ -0,0 +1,67 @@
...
+     * Adds a value to the configured filter after applying the configured normalization strategy.
+     * @param value
+     */
+    auto add(std::string_view value) -> void;
+
+    /**
+     * Writes a serialized filter representation to a writer.
+     * @param writer
+     */
+    auto write_to_file(clp::WriterInterface& writer) const -> void;
```
gibber9809: Since we're writing to a `WriterInterface` maybe we should just call this method `write` instead of `write_to_file`?

components/core/src/clp_s/filter/FilterBuilder.hpp
```diff
@@ -0,0 +1,67 @@
...
+    [[nodiscard]] auto get_type() const -> FilterType { return m_type; }
+
+    [[nodiscard]] auto get_normalization() const -> FilterNormalization { return m_normalization; }
+
+private:
+    FilterBuilder(FilterType type, FilterNormalization normalization, BloomFilter bloom_filter);
+
+    FilterType m_type;
+    FilterNormalization m_normalization;
+    BloomFilter m_bloom_filter;
```
gibber9809:
```suggestion
    FilterType m_type{};
    FilterNormalization m_normalization{};
    BloomFilter m_bloom_filter;
```
Nit: Generally best to make sure all class members get default initialized to something reasonable, that way if more constructors get added/constructors change over time all of the members are still initialized properly.

components/core/src/clp_s/filter/FilterBuilder.hpp
```diff
@@ -0,0 +1,67 @@
...
+     * @param writer
+     */
+    auto write_to_file(clp::WriterInterface& writer) const -> void;
+
+    /**
+     * @return The configured filter type.
+     */
+    [[nodiscard]] auto get_type() const -> FilterType { return m_type; }
+
+    [[nodiscard]] auto get_normalization() const -> FilterNormalization { return m_normalization; }
```
gibber9809: Don't necessarily need docstrings for simple get methods that are described by their name, but if you put it on one here you should put it on both. Also fine to just get rid of the docstring on `get_type`.

components/core/src/clp_s/filter/FilterBuilder.cpp
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    switch (type) {
+        case FilterType::Bloom: {
+            auto bloom_filter{YSTDLIB_ERROR_HANDLING_TRYX(
+                    BloomFilter::create(expected_num_elements, false_positive_rate)
+            )};
+            return FilterBuilder{type, normalization, std::move(bloom_filter)};
+        }
+    }
```
gibber9809: It reads a bit strangely to use `if` for conditions on one enum and `switch` for the conditions on the other enum. For consistency, could we use `switch` for both?

components/core/src/clp_s/filter/FilterBuilder.cpp
```diff
@@ -0,0 +1,61 @@
...
+    return ErrorCode{ErrorCodeEnum::UnsupportedFilterType};
+}
+
+auto FilterBuilder::add(std::string_view value) -> void {
+    auto const normalized_value{normalize_string(value, m_normalization)};
+    m_bloom_filter.add(normalized_value);
+}
+
+auto FilterBuilder::write_to_file(clp::WriterInterface& writer) const -> void {
+    writer.write(cFilterFileMagic.data(), cFilterFileMagic.size());
```
gibber9809: Do you think that this needs file magic?

Within an archive we generally don't use file magic, and instead rely on some archive-level metadata about the different sections of the archive. I can see the argument for when this is external to an archive, but I'm not sure if this file format is actually independent as it is right now anyway (in that it doesn't indicate which archive it's a filter for), so to use it externally to an archive we have to have some external way to tell that this is a filter for a given archive anyway. Potentially open to keeping it if you can think of a use-case though.

When we get to doing packed filters though I can definitely see file magic making sense.

components/core/src/clp_s/filter/FilterBuilder.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include <ystdlib/error_handling/Result.hpp>
+
+#include <clp/WriterInterface.hpp>
+
+#include "BloomFilter.hpp"
+#include "FilterOptions.hpp"
+
+namespace clp_s::filter {
+/**
+ * Builds variable-dictionary filters and writes them to a writer.
```
gibber9809: Do we need to restrict ourselves to just the variable dictionary with this class? I could see this being useful for string filtering in other contexts as well (e.g. string filter on a specific column or something as well).

If we did want to extend this beyond just variable dictionary filtering I think we could bake that into the `FilterType` pretty easily (e.g. `FilterType::VariableDictionaryBloomFilter`, `FilterType::LogTypeDictionaryNGramFilter`,  `FilterType::ColumnFilter`.), possibly with some types writing some extra metdata when needed (e.g. column filter also recording the column name or something).

gibber9809: I guess this really depends on whether you think that we need a significantly different interface for string filtering in other cases.

components/core/src/clp_s/filter/FilterOptions.cpp
gibber9809: I'd maybe leave it for the integration PR then since it isn't directly related, but not a huge deal either way.

If you do keep it though `"bloom"` needs to be a `constexpr std::string_view` per our style guide.

components/core/src/clp_s/filter/FilterOptions.hpp
```diff
@@ -0,0 +1,63 @@
...
+        -> std::optional<FilterType>;
+
+/**
+ * Parses a serialized filter normalization ID.
+ * @param filter_normalization_id Encoded filter normalization ID from serialized payload.
+ * @return Parsed filter normalization for known IDs, or std::nullopt otherwise.
+ */
+[[nodiscard]] auto try_parse_filter_normalization_id(uint8_t filter_normalization_id)
+        -> std::optional<FilterNormalization>;
+
```
gibber9809: Besides the command line related method, are there any users for these parse methods besides `FilterReader`? I'm wondering whether it makes sense for these to just be private methods in that class.

components/core/src/clp_s/filter/FilterReader.hpp
```diff
@@ -0,0 +1,59 @@
...
+    [[nodiscard]] static auto try_read_from_file(clp::ReaderInterface& reader)
+            -> ystdlib::error_handling::Result<FilterReader>;
+
+    /**
+     * Checks whether the filter may contain the value after applying the configured
+     * normalization strategy.
+     * @param value
+     * @return true if the filter may contain `value`; false if it definitely does not.
+     */
+    [[nodiscard]] auto possibly_contains(std::string_view value) const -> bool;
```
gibber9809, line 40: Do you think it makes sense to have another method here that accepts a wildcard search string and returns whether it is possibly contained in the data based on the filter?

For the current bloom filter based approach it'll just be always true unless the search string contains no wildcards, in which case we actually do the lookup, but even that could be useful. Definitely more useful for other kinds of filters though.

components/core/src/clp_s/filter/FilterReader.hpp
```diff
@@ -0,0 +1,59 @@
...
+    [[nodiscard]] auto get_type() const -> FilterType { return m_type; }
+
+    [[nodiscard]] auto get_normalization() const -> FilterNormalization { return m_normalization; }
+
+private:
+    FilterReader(FilterType type, FilterNormalization normalization, BloomFilter bloom_filter);
+
+    FilterType m_type;
+    FilterNormalization m_normalization;
+    BloomFilter m_bloom_filter;
```
gibber9809:
```suggestion
    FilterType m_type{};
    FilterNormalization m_normalization{};
    BloomFilter m_bloom_filter;
```
Ditto about default initialization.

components/core/src/clp_s/filter/FilterReader.hpp
```diff
@@ -0,0 +1,59 @@
...
+     * Reads a serialized filter from a reader.
+     * @param reader
+     * @return A result containing an initialized FilterReader on success, or an error code
+     * indicating the failure:
+     * - ErrorCodeEnum::CorruptFilterFile for malformed filter file metadata.
+     * - ErrorCodeEnum::UnsupportedFilterType if the filter type ID is not supported.
+     * - ErrorCodeEnum::UnsupportedFilterNormalization if the filter normalization ID is not
+     *   supported.
+     * - ErrorCodeEnum::ReadFailure for truncated/failed reads.
+     * @return Forwards the underlying filter implementation's return values on payload read
```
gibber9809:
```suggestion
     * - Forwards the underlying filter implementation's return values on payload read
```

gibber9809 (changes requested):
Did an initial pass and looks good overall, just some details to discuss. I think I generally agree that it's fine to not make the interface completely generic to different kinds of filters at the moment -- mostly looks like it will just require a bit of refactoring for the `create` methods down the line.

components/core/src/clp_s/filter/FilterBuilder.cpp
ShangDanLuXian: There is intentionally no archive-to-filter binding encoded in this library itself, since this filter serialization layer may be used in different ways in the future.

My reason for keeping the file magic is narrower: it gives us a lightweight validation that the bytes being packed are actually a serialized filter file, which helps guard against malformed files or unrelated junk in the staging area.

That said, I agree this is arguable. In my prototype implementation, the packing flow identified staged filters by filename convention:
<filter_output_dir>/<archive_id>.var.dict.filter

So in that flow, the archive association comes from the filename/path convention rather than from the filter payload itself. That already filters out most unrelated files in the staging directory (if there is any), unless something else happens to match the same naming pattern. Do you think we need to protect that case?

components/core/src/clp_s/filter/FilterBuilder.hpp
ShangDanLuXian: This is just stale wording in the docstring — the class itself is intended to be generic, not specific to variable-dictionary filtering.

My current thinking is that this layer should only encode the filter representation itself, e.g. filter type plus normalization. The caller is responsible for deciding what the filter is being built for, such as a variable dictionary versus some future column-level string filter. If we later need filter-specific metadata for those other use cases, I think that should be added explicitly at the wrapper/format level rather than folded into FilterType now. Please also see my other reply about how the variable-dictionary association was handled in the earlier prototype for context.

components/core/src/clp_s/filter/FilterOptions.hpp
ShangDanLuXian: I grouped these together intentionally since they are closely related, but I agree this is probably exposing a bit too much. I think it makes sense to keep the string-based one shared, and move the serialized-ID ones into FilterReader.

components/core/src/clp_s/filter/FilterReader.hpp
ShangDanLuXian: Yeah, it makes sense. I’ll add a wildcard-check function with a switch on filter type; for Bloom, it will always return true, as this type is for exact match only.

components/core/src/clp_s/filter/FilterBuilder.cpp
gibber9809: I think your reason for keeping it as lightweight validation when it's staged makes sense, so lets keep it for now. We can always revisit whether the file magic part is required for filters in archives when we get to that.

components/core/src/clp_s/filter/FilterBuilder.hpp
gibber9809: I see your point, but we should probably think a little bit about how that external metadata documenting what the filter is for works then. Reason being that I'm not sure we'll be able to rely entirely on file naming conventions in all cases, so the approach we're taking for staging bloom filters on the variable dictionary might not work directly for all types of filters. Point being I guess that for the general case we need to explicitly store the mapping between staged filters -> parts of the archive that they apply to _somewhere_ alongside the staged filter(s).

components/core/src/clp_s/filter/FilterReader.cpp
```diff
@@ -0,0 +1,105 @@
...
+    return FilterReader{
+            optional_filter_type.value(),
+            optional_filter_normalization.value(),
+            std::move(bloom_filter)
+    };
+}
+
+auto FilterReader::possibly_contains(std::string_view value) const -> bool {
+    auto const normalized_value{normalize_string(value, m_normalization)};
+    return m_bloom_filter.possibly_contains(normalized_value);
```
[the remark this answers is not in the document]
gibber9809: Yeah I think the proper way to address this is to add a normalizer class that internally maintains a buffer and has a method `normalize(std::string_view) -> std::string_view` that will return an `std::string_view` of its internal buffer when it has to do normalization and return the input `std::string_view` otherwise.

Note that even when performing a given normalization you can sometimes scan and then directly return the input string without copying it into the internal buffer (e.g. when the input string is already lowercase).

That said, since we don't expect to have to check too many values for a given bloom filter + we expect that normalization will be required in the common case, optimizing to occasionally avoid a copy here probably isn't worthwhile at the moment.

components/core/src/clp_s/filter/FilterOptions.cpp
```diff
@@ -0,0 +1,51 @@
...
+}
+
+auto normalize_string(std::string_view value, FilterNormalization normalization) -> std::string {
+    std::string normalized{value};
+    switch (normalization) {
+        case FilterNormalization::None:
+            return normalized;
+        case FilterNormalization::Lowercase:
+            clp::string_utils::to_lower(normalized);
+            return normalized;
```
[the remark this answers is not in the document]
gibber9809: Incorrect -- the behaviour of `clp::string_utils::to_lower` doesn't depend on locale. Generally the approach for handling lower-casing for different locales will probably be to treat them as different kinds of normalization.

components/core/src/clp_s/filter/FilterReader.hpp
```diff
@@ -0,0 +1,65 @@
...
+     * normalization strategy.
+     * @param value
+     * @return true if the filter may contain `value`; false if it definitely does not.
+     */
+    [[nodiscard]] auto possibly_contains(std::string_view value) const -> bool;
+
+    /**
+     * Checks whether the filter may contain the wildcard search string.
+     * The caller is responsible for determining whether `value` should be treated as a wildcard
+     * query.
```
gibber9809, line 46:
```suggestion
     * Checks whether the filter may contain a wildcard query string.
     * The caller is responsible for ensuring `value` is a valid wildcard query string.
```

components/core/src/clp_s/filter/FilterReader.hpp
```diff
@@ -0,0 +1,65 @@
...
+     * @param value
+     * @return true if the filter may contain `value`; false if it definitely does not.
```
gibber9809, line 48:
```suggestion
     * @return Whether the filter may contain the query string `value`.
```

components/core/src/clp_s/filter/FilterReader.hpp
```diff
@@ -0,0 +1,65 @@
...
+     */
+    [[nodiscard]] auto possibly_contains_wildcard(std::string_view value) const -> bool;
```
gibber9809, line 50: Could we rename to `possibly_contains_query_string`, or another more specific name? `possibly_contains_wildcard` reads to me like we're checking if the filter contains string values with wildcards.

components/core/src/clp_s/filter/FilterReader.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+auto FilterReader::possibly_contains_wildcard(std::string_view) const -> bool {
+    switch (m_type) {
+        case FilterType::Bloom:
+            return true;
+    }
+
+    return true;
+}
```
gibber9809, line 95:
```suggestion
auto FilterReader::possibly_contains_wildcard(std::string_view value) const -> bool {
    switch (m_type) {
        case FilterType::Bloom:
            if (clp_s::search::ast::has_unescaped_wildcards(value)) {
                return true;
            }
            auto const unescaped_value{clp::string_utils::unescape_string(value)};
            return possibly_contains(unescaped_value);
    }

    return true;
}
```
^I think it makes sense to handle the case where we can still use the bloom filter because the query string happens to not contain any wildcards.

we would need to include/ depend on `<clp/string_utils.hpp` and `<clp_s/search/ast/SearchUtils.hpp>` as well.

components/core/src/clp_s/filter/FilterBuilder.cpp
gibber9809: Sorry, revisiting this one more time since the new plan doesn't involve staging anymore -- do you think we still need the file magic here?

I'm also happy to leave this for now and decide whether we should change it when we get to writing the indexing flow to avoid holding up this PR any longer.

gibber9809 (changes requested):
Apologies for taking a long time to get back to this. Just a few details I think are worth changing, but otherwise looks like this PR is looking good.

gibber9809 (commented):
Changes look good, but will hold off on marking approved until we finish cutting a release from main.

For the PR title, how about:
```
feat(clp-s::filter): Add abstraction layer for filter building, serialization, and deserialization.
```

gibber9809 (approved):
Marking approved now that the release is done.

LinZhihao-723 (approved):
Took a skim without checking details. Approved on behalf of the code owner approval.

PR #2105, opened 2026-03-18
feat(clp-s): Add `SchemaReader` API to iterate over messages with metadata without a filter; make `SchemaReader` message iteration null-filter safe.

Bill-hbrhbr:
# Description

This change improves the robustness of `SchemaReader` message iteration by forcing a filter reference to be provided when doing search related iterations.

This PR also adds a filter-less version of `get_next_message_with_metadata`.

# Checklist

  breaking change.

# Validation performed

* [X] Existing tests and CI pass.

## Summary by CodeRabbit

* **New Features**
  * Added a retrieval variant that returns message metadata (timestamp and event index) alongside message content.

* **Bug Fixes**
  * Improved filtering and iteration so non-matching entries are reliably skipped and exhaustion is correctly signalled.
  * Fixed off-by-one/unconditional-advance issues to ensure message emission and metadata assignment are consistent.

* **Tests**
  * Updated unit tests to cover the revised filter and message-retrieval behaviour.

### DIFF
--- a/components/core/src/clp_s/ArchiveReader.hpp
+++ b/components/core/src/clp_s/ArchiveReader.hpp
@@ -10,6 +10,7 @@
 #include <utility>
 #include <vector>
 
+#include <nlohmann/json_fwd.hpp>
 #include <ystdlib/error_handling/Result.hpp>
 
 #include <clp_s/ArchiveReaderAdaptor.hpp>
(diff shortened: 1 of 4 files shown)


components/core/src/clp_s/SchemaReader.cpp
```diff
@@ -224,76 +225,99 @@ bool SchemaReader::get_next_message(std::string& message) {
...
+bool SchemaReader::get_next_message_with_metadata(
+        std::string& message,
+        epochtime_t& timestamp,
+        int64_t& log_event_idx
+) {
+    if (m_cur_message >= m_num_messages) {
+        return false;
+    }
+
+    if (false == m_serializer_initialized) {
```
gibber9809: I know it's a weird design, but just for the sake of consistency with the rest of the class we should respect `m_should_marshal_records` here.

gibber9809 (changes requested):
Mostly looks good to me after the most recent change, just one detail I think we should change.

PR #2106, opened 2026-03-18
refactor(integration-tests): Refactor and augment data class design.

quinntaylormitchell:
# Description

This PR implements a data class redesign for the integration test system.

# Checklist

  breaking change.

# Validation performed

The data classes introduced in this PR are not yet used elsewhere in the code.

## Summary by CodeRabbit

* **Tests**
  * Added integration-test infrastructure for CLP: configurable path and directory management with automated cache and log creation, static-path existence checks, dataset metadata utilities, helpers to locate CLP binaries and package scripts, and facilities to load and validate package runtime configs and instance identifiers used during integration runs.

### DIFF
--- a/components/core/src/clp/BufferReader.cpp
+++ b/components/core/src/clp/BufferReader.cpp
@@ -37,7 +37,7 @@ auto BufferReader::try_read_to_delimiter(
     // Find the delimiter
     auto const* buffer_head = m_internal_buf + m_internal_buf_pos;
     auto const* delim_ptr
-            = static_cast<char const*>(memchr(buffer_head, delim, remaining_data_size));
+            = static_cast<char const*>(std::memchr(buffer_head, delim, remaining_data_size));
 
     size_t append_length{0};
     if (delim_ptr != nullptr) {
--- a/components/core/src/clp/LibarchiveFileReader.cpp
+++ b/components/core/src/clp/LibarchiveFileReader.cpp
@@ -53,7 +53,7 @@ LibarchiveFileReader::try_read(char* buf, size_t num_bytes_to_read, size_t& num_
                     (size_t)(m_data_block_pos_in_file - m_pos_in_file),
                     num_bytes_to_read - num_bytes_read
             );
-            memset(&buf[num_bytes_read], '\0', num_zeros_to_append);
+            std::memset(&buf[num_bytes_read], '\0', num_zeros_to_append);
             num_bytes_read += num_zeros_to_append;
             m_pos_in_file += num_zeros_to_append;
 
@@ -68,7 +68,7 @@ LibarchiveFileReader::try_read(char* buf, size_t num_bytes_to_read, size_t& num_
             size_t data_length = m_data_block_length - m_pos_in_data_block;
 
             size_t num_bytes_to_append = std::min(data_length, num_bytes_to_read - num_bytes_read);
-            memcpy(&buf[num_bytes_read], data, num_bytes_to_append);
+            std::memcpy(&buf[num_bytes_read], data, num_bytes_to_append);
             num_bytes_read += num_bytes_to_append;
             m_pos_in_data_block += num_bytes_to_append;
             m_pos_in_file += num_bytes_to_append;
@@ -141,7 +141,8 @@ ErrorCode LibarchiveFileReader::try_read_to_delimiter(
             char const* data = reinterpret_cast<char const*>(m_data_block) + m_pos_in_data_block;
             size_t data_length = m_data_block_length - m_pos_in_data_block;
 
-            char const* delim_ptr = reinterpret_cast<char const*>(memchr(data, delim, data_length));
+            char const* delim_ptr
+                    = reinterpret_cast<char const*>(std::memchr(data, delim, data_length));
             if (nullptr == delim_ptr) {
                 // Add the remaining data to the string
                 str.append(data, data_length);
--- a/components/core/src/clp/LibarchiveReader.cpp
+++ b/components/core/src/clp/LibarchiveReader.cpp
@@ -1,5 +1,7 @@
 #include "LibarchiveReader.hpp"
 
+#include <cstring>
+
 #include <archive_entry.h>
 
 #include "Defs.h"
@@ -156,7 +158,7 @@ la_ssize_t LibarchiveReader::libarchive_read_callback(
                 archive_set_error(archive, ENOMEM, "Unknown error.");
                 return -1;
             case ErrorCode_errno:
-                archive_set_error(archive, errno, "%s", strerror(errno));
+                archive_set_error(archive, errno, "%s", std::strerror(errno));
                 return -1;
             case ErrorCode_EndOfFile:
                 return 0;
--- a/components/core/src/clp/MySQLParamBindings.cpp
+++ b/components/core/src/clp/MySQLParamBindings.cpp
@@ -15,7 +15,7 @@ void MySQLParamBindings::resize(size_t num_fields) {
     m_statement_binding_lengths.resize(num_fields);
     for (size_t i = 0; i < num_fields; ++i) {
         auto& binding = m_statement_bindings[i];
-        memset((void*)&binding, 0, sizeof(binding));
+        std::memset((void*)&binding, 0, sizeof(binding));
         binding.length = &m_statement_binding_lengths[i];
     }
 }
--- a/components/core/src/clp/TimestampPattern.cpp
+++ b/components/core/src/clp/TimestampPattern.cpp
@@ -354,7 +354,7 @@ bool TimestampPattern::parse_timestamp(
                     case 'B': {  // Month name
                         bool match_found = false;
                         for (int month_ix = 0; !match_found && month_ix < cNumMonths; ++month_ix) {
-                            size_t const length = strlen(cMonthNames[month_ix]);
+                            size_t const length = std::strlen(cMonthNames[month_ix]);
                             if (0 == line.compare(line_ix, length, cMonthNames[month_ix])) {
                                 month = month_ix + 1;
                                 match_found = true;
@@ -369,7 +369,7 @@ bool TimestampPattern::parse_timestamp(
                     case 'b': {  // Abbreviated month name
                         bool match_found = false;
                         for (int month_ix = 0; !match_found && month_ix < cNumMonths; ++month_ix) {
-                            size_t const length = strlen(cAbbrevMonthNames[month_ix]);
+                            size_t const length = std::strlen(cAbbrevMonthNames[month_ix]);
                             if (0 == line.compare(line_ix, length, cAbbrevMonthNames[month_ix])) {
                                 month = month_ix + 1;
                                 match_found = true;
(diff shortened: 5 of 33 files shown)


integration-tests/tests/package_tests/__init__.py
```diff
@@ -0,0 +1 @@
+"""Top-level package for CLP binary integration tests."""
```
Bill-hbrhbr:
```suggestion
"""Top-level package for CLP package integration tests."""
```

integration-tests/tests/utils/classes.py
```diff
@@ -0,0 +1,74 @@
...
+    clp_build_dir: Path
+
+    #: Default integration test project root directory.
+    integration_tests_project_root: Path
+
+    #: The cache directory for integration tests.
+    test_cache_dir: Path = field(init=False)
+
+    #: The test log directory for integration tests.
+    test_log_dir: Path = field(init=False)
```
Bill-hbrhbr: Instead of initializing these fields inside `__post_init__`, let's just turn them into property functions.

We can still make sure to call `mkdir` + property functions for them inside `__post_init__`.

integration-tests/tests/utils/classes.py
```diff
@@ -0,0 +1,74 @@
...
+        self.test_cache_dir.mkdir(parents=True, exist_ok=True)
+
+        # Create and set `test_log_dir` within `test_cache_dir`.
+        object.__setattr__(self, "test_log_dir", self.test_cache_dir / "test_logs")
+        self.test_log_dir.mkdir(parents=True, exist_ok=True)
+
+    @property
+    def test_data_path(self) -> Path:
+        """:return: The absolute path to the sample dataset directory."""
+        return self.integration_tests_project_root / "tests" / "data"
```
Bill-hbrhbr, line 49: If we are not keeping a centralized data directory, I think we should remove this property.

integration-tests/tests/package_tests/classes.py
```diff
@@ -0,0 +1,221 @@
...
+
+        # Ensure that the CLP package directory exists and that it is structured correctly.
+        clp_package_dir = self.clp_package_dir
+        validate_dir_exists(clp_package_dir)
+        required_dirs = ["etc", "sbin"]
+        missing_dirs = [d for d in required_dirs if not (clp_package_dir / d).is_dir()]
+        if len(missing_dirs) > 0:
+            err_msg = f"The CLP package has missing directories: {', '.join(missing_dirs)}"
+            pytest.fail(err_msg)
+
```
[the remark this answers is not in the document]
Bill-hbrhbr: I agree with the rabbit's comment here. I think we can iterate over all the property function return values, and make sure they exist (either as a file or as a dir). Here's the pseudo code:
```
  missing_paths: list[str] = []

  for name in dir(type(self)):
      attr = getattr(type(self), name, None)
      if isinstance(attr, property):
          path = getattr(self, name)
          if not path_exists(path):
              missing_paths.append(name)
```
Even better, we can make this code a method of `IntegrationTestPathConfig`, so that we don't have to duplicate the utility in either derived path classes.

integration-tests/tests/package_tests/classes.py
```diff
@@ -0,0 +1,221 @@
...
+        super().__post_init__()
+
+        # Ensure that the CLP package directory exists and that it is structured correctly.
+        clp_package_dir = self.clp_package_dir
+        validate_dir_exists(clp_package_dir)
+        required_dirs = ["etc", "sbin"]
+        missing_dirs = [d for d in required_dirs if not (clp_package_dir / d).is_dir()]
+        if len(missing_dirs) > 0:
+            err_msg = f"The CLP package has missing directories: {', '.join(missing_dirs)}"
+            pytest.fail(err_msg)
```
Bill-hbrhbr:
```suggestion
```
replace with property check

integration-tests/tests/package_tests/classes.py
```diff
@@ -0,0 +1,221 @@
...
+    """Path configuration for CLP package integration tests."""
+
+    #: Default CLP package directory.
+    clp_package_dir: Path
+
+    #: Directory to store temporary package config files.
+    temp_config_dir: Path = field(init=False)
+
+    #: Directory to store temporary decompressed files.
+    package_decompression_dir: Path = field(init=False)
```
Bill-hbrhbr: Similarly let's turn these into properties. I don't see the benefits of `init=False` anymore.

integration-tests/tests/utils/classes.py
```diff
@@ -0,0 +1,74 @@
...
+
+    #: Path to the dataset logs.
+    path_to_dataset_logs: Path
+
+    #: A dictionary of metadata describing the dataset.
+    metadata_dict: dict[str, Any]
+
+
+@dataclass
+class IntegrationTestExternalAction:
```
Bill-hbrhbr: Let's remove all the external action classes from this PR. I will post my analysis in the group channel. Ideally, we also remove all the non-path related classes in this PR, but that's up to you.
Whichever you choose, I believe the next immediate PR is to replace the old **path** config classes and fixtures with the new ones you've added here.

integration-tests/tests/package_tests/__init__.py
quinntaylormitchell: good catch thank you

integration-tests/tests/utils/classes.py
quinntaylormitchell: agreed

integration-tests/tests/utils/classes.py
quinntaylormitchell: discussing this offline; in progress

integration-tests/tests/utils/classes.py
quinntaylormitchell: agreed

integration-tests/tests/package_tests/classes.py
quinntaylormitchell: agreed

integration-tests/tests/package_tests/classes.py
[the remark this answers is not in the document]
quinntaylormitchell: @Bill-hbrhbr sure, I see your point

integration-tests/tests/package_tests/classes.py
quinntaylormitchell: agreed

integration-tests/tests/utils/classes.py
```diff
@@ -0,0 +1,98 @@
...
+    clp_build_dir: Path
+
+    #: Default integration test project root directory.
+    integration_tests_project_root: Path
+
+    def __post_init__(self) -> None:
+        """Create and validate directories."""
+        # Validate that init directories exist.
+        validate_dir_exists(self.clp_build_dir)
+        validate_dir_exists(self.integration_tests_project_root)
```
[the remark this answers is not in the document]
Bill-hbrhbr: I think these directories will be normalized before dataclass parameter passing.

integration-tests/tests/utils/classes.py
```diff
@@ -0,0 +1,98 @@
...
+_STATIC_PATH_ATTR = "_is_static_path"
+
+
+def static_path(func: Callable[..., Path]) -> Callable[..., Path]:
+    """
+    Decorator that marks a method as a static (non-runtime) path to be validated by
+    `IntegrationTestPathConfig.validate_paths()`. Must be placed below `@property`.
+    """
+    setattr(func, _STATIC_PATH_ATTR, True)
+    return func
```
junhaoliao: while this is smart, it seems there are quite a few restrictions in using this decorator, and things might easily go wrong:
1. "Must be placed below `@property`." - i suppose it won't work if we happen to place `@property` below it. We can add a check on the type of `func` though that is extra effort:
    ```py
        if isinstance(func, property):
            err_msg = "@static_path must be placed below @property, not above it."
            raise TypeError(err_msg)
    ```
2. We have to do type reflection with `inspect.getmembers()` - nothing wrong but it's less intuitive to those who don't know about the API and haven't read the manual
3. While considerably this may improve dev ergonomics, one must read the decorator / the `validate_static_paths()` method to understand how the whole validation flow works, especially when their entry point is one of the subclasses - we would be in a "write easy, read hard" situation


how do you feel about simply structuring the code as:
```py
@dataclass
class IntegrationTestPathConfig:
    def _static_paths(self) -> list[Path]:
        """:return: Paths that must exist on disk at construction time."""
        return [self.test_data_path]

        ...
        for path in self._static_paths():
            if not path.exists():
                pytest.fail(f"Expected path does not exist: '{path}'")

@dataclass
class ClpPackageTestPathConfig(IntegrationTestPathConfig):
    def _static_paths(self) -> list[Path]:
        return super()._static_paths() + [
            self.archive_manager_path,
            self.compress_from_s3_path,
            self.compress_path,
            self.dataset_manager_path,
            self.decompress_path,
            self.search_path,
            self.start_clp_path,
            self.stop_clp_path,
        ]
        validate_dir_exists(self.clp_package_dir)
        super().__post_init__()
        ...

@dataclass
class ClpBinaryTestPathConfig(IntegrationTestPathConfig):
    def _static_paths(self) -> list[Path]:
        return super()._static_paths() + [self.clp_binary_path, self.clp_s_binary_path]

        """
        Validates that the CLP core binaries directory exists and contains all required
        executables.
        """
        # Validate that init directory exists.
        validate_dir_exists(self.clp_core_bins_dir)

        super().__post_init__()
```

on my end i found a 33-line reduction in code after making all changes

integration-tests/tests/utils/classes.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # Create `test_cache_dir`.
+        self.test_cache_dir.mkdir(parents=True, exist_ok=True)
+
+        # Create `test_log_dir`.
+        self.test_log_dir.mkdir(parents=True, exist_ok=True)
+
+        # Validate all static path properties.
+        if type(self) is IntegrationTestPathConfig:
+            self.validate_static_paths()
```
junhaoliao: it might be more of a personal preference but i'd validate everything before making write operations, just so nothing changes on the disk if things may have gone wrong. e.g. place `self.validate_static_paths()` before the `mkdir()`

integration-tests/tests/package_tests/classes.py
```diff
@@ -0,0 +1,208 @@
...
+        """Validate data members."""
+        log_msg = f"Validating the ClpConfig pydantic object for the '{self.mode_name}' package."
+        logger.info(log_msg)
+        try:
+            ClpConfig.model_validate(self.clp_config)
+        except ValidationError as err:
+            fail_msg = (
+                f"The `ClpConfig` pydantic object for the CLP package could not be validated: {err}"
+            )
+            pytest.fail(fail_msg)
```
junhaoliao, line 159: i might be wrong about this `ClpConfig.model_validate()` already run on any ClpConfig instances when the instances are created. maybe this is redudant? or maybe i have not understood the flow of `ClpPackage` construction correctly

if the validation is intentional, some docstring of the intention would be appreciated

quinntaylormitchell: This is the only place that the `ClpConfig` object is validated; I'll give it a more explicit comment.

Also, fwiw `ClpPackage` isn't used yet in the code; that will come in a later PR that merges the `fixt_package_test_config` and the `fixt_package_instance` fixtures into a single fixture that uses `ClpPackage`.

integration-tests/tests/utils/classes.py
quinntaylormitchell: I like that idea, thanks Junhao!

integration-tests/tests/utils/classes.py
quinntaylormitchell: I don't have a preference, I will defer to you

PR #2117, opened 2026-03-24
refactor(log-ingestor): Use mock ingestion state for testing S3 ingestion jobs (resolves #2114).

### LINKED ISSUE #2114: refactor(log-ingestor): Refactor test_ingestion_job.rs to use state-based shared buffers
## Overview

The tests in `components/log-ingestor/tests/test_ingestion_job.rs` were originally written when there was no "state" concept for an ingestion job. The original flow was:

```
ingestion jobs → channel senders
```

When the state concept was later added, the tests were not updated accordingly, resulting in the current (confusing) flow:

```
ingestion jobs → state → channel senders
```

This makes the tests inefficient and harder to understand.

## Proposed Refactor

- Implement `SqsListenerState` and `S3ScannerState` directly in the test file.
- Each state implementation should hold a shared buffer protected by a `Mutex`.
- Replace the current use of `ZeroFaultToleranceIngestionJobState` with these new test-local state implementations.
- Use the state as a shared buffer to check ingested files, instead of routing through a channel receiver.

## References

- Raised in PR #2087 (comment: https://github.com/y-scope/clp/pull/2087#discussion_r2970205586)
- Requested by @LinZhihao-723

jonathan-imanu:
# Description

Fixes #2114 by:

- Implement TestSqsListenerState and TestS3ScannerState directly in the test file.
- Each state implementation should hold a shared buffer protected by a Mutex.
- Replaced current use of ZeroFaultToleranceIngestionJobState with these new test-local state implementations.
- Used the state as a shared buffer to check ingested files, instead of routing through a channel receiver.

# Checklist

  breaking change.

# Validation performed

- `task tests:rust-all` you need to modify the script to add `LOCALSTACK_ACKNOWLEDGE_ACCOUNT_REQUIREMENT=1`

## Summary by CodeRabbit

* **Tests**
  * Improved ingestion tests to use a shared in-memory buffer with synchronized polling to await ingested objects. Added lightweight test-state helpers and a simplified upload helper; replaced the previous receiver-based collection. Adjusted noise-object handling for more accurate test expectations.
* **Chores**
  * No user-facing changes.


components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -17,16 +21,104 @@ use clp_rust_utils::{
...
+        objects: Vec<ObjectMetadata>,
+        _last_ingested_key: &str,
+    ) -> anyhow::Result<()> {
+        self.ingested_objects
+            .lock()
+            .expect("poisoned mutex")
+            .extend(objects);
+        Ok(())
+    }
+}
```
LinZhihao-723, line 118: We should move this after the corresponded struct impl:
```rust
struct X {}

impl X {
  ...
}

impl Y for X {
  ...
}
```

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -17,16 +21,104 @@ use clp_rust_utils::{
...
+struct SqsListenerTestState {
+    ingested_objects: Arc<Mutex<Vec<ObjectMetadata>>>,
+}
+
+impl SqsListenerTestState {
+    fn new() -> Self {
+        Self::default()
+    }
+
+    fn shared_buffer(&self) -> Arc<Mutex<Vec<ObjectMetadata>>> {
```
LinZhihao-723: `shared_buffer` -> `get_shared_buffer`
We shouldn't omit `get_` prefix for clarity that this is a getter method.

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -17,16 +21,104 @@ use clp_rust_utils::{
...
-const RECEIVER_TIMEOUT_SEC: u64 = 30;
+const WAIT_FOR_INGESTED_OBJECTS_TIMEOUT_SEC: u64 = 30;
+const INGESTED_OBJECT_POLL_INTERVAL_MS: u64 = 100;
 const NUM_TEST_OBJECTS: usize = 100;
 const NUM_NOISE_OBJECTS: usize = 100;
-const TEST_CHANNEL_CAPACITY: usize = 10;
+
+#[derive(Clone, Default)]
+struct SqsListenerTestState {
+    ingested_objects: Arc<Mutex<Vec<ObjectMetadata>>>,
```
LinZhihao-723: `ingested_objects` -> `shared_ingestion_buffer`?

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -17,16 +21,104 @@ use clp_rust_utils::{
...
+
+#[async_trait]
+impl IngestionJobState for SqsListenerTestState {
+    async fn start(&self) -> anyhow::Result<()> {
+        Ok(())
+    }
+
+    async fn end(&self) -> anyhow::Result<()> {
+        Ok(())
+    }
```
LinZhihao-723:
```suggestion
    async fn start(&self) -> Result<()> {
        Ok(())
    }

    async fn end(&self) -> Result<()> {
        Ok(())
    }
```
We use `anyhow::Result` in this mod.

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -17,16 +21,104 @@ use clp_rust_utils::{
...
+    ingested_objects: Arc<Mutex<Vec<ObjectMetadata>>>,
+}
+
+impl SqsListenerTestState {
+    fn new() -> Self {
+        Self::default()
+    }
+
+    fn shared_buffer(&self) -> Arc<Mutex<Vec<ObjectMetadata>>> {
+        Arc::clone(&self.ingested_objects)
```
LinZhihao-723:
```suggestion
        self.ingested_objects.clone()
```
Same for S3 Scanner

components/log-ingestor/tests/test_ingestion_job.rs
LinZhihao-723: We should probably use `tokio::sync::Mutex` since the critical section is called inside a tokio coroutine.

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -59,46 +151,46 @@ async fn create_s3_objects(
...
+    tokio::time::timeout(
+        Duration::from_secs(WAIT_FOR_INGESTED_OBJECTS_TIMEOUT_SEC),
+        async {
+            loop {
+                {
+                    let ingested_objects = shared_buffer.lock().expect("poisoned mutex");
+                    if ingested_objects.len() >= min_num_objects {
+                        return ingested_objects.clone();
+                    }
+                }
```
LinZhihao-723:
```suggestion
                let ingested_objects = shared_buffer.lock().await;
                if ingested_objects.len() >= min_num_objects {
                    return ingested_objects.clone();
                }
                drop(ingested_objects);
```
With the change to `tokio::sync::Mutex` you should be able to write it this way, which minimizes the critical section and passes all linter checks.

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -186,12 +264,11 @@ async fn run_sqs_listener_test(
     )
     .await;
 
-    let upload_and_receive_handle = tokio::spawn(upload_and_receive(
+    let test_upload_handle = tokio::spawn(upload_test_objects(
```
LinZhihao-723: How about just `upload_handle`?

LinZhihao-723 (changes requested):

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -17,16 +18,99 @@ use clp_rust_utils::{
...
+    async fn start(&self) -> Result<()> {
+        Ok(())
+    }
+
+    async fn end(&self) -> Result<()> {
+        Ok(())
+    }
+
+    async fn fail(&self, _msg: String) {}
+}
```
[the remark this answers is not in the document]
LinZhihao-723: I think it's worth to add an assert inside `fail` to make sure it's never been triggered (in the current implementation it should be unreachable, but adding it would be a bit more future-proof)

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -17,16 +18,103 @@ use clp_rust_utils::{
...
+    async fn start(&self) -> Result<()> {
+        Ok(())
+    }
+
+    async fn end(&self) -> Result<()> {
+        Ok(())
+    }
+
+    async fn fail(&self, msg: String) {
+        unreachable!("SqsListenerTestState::fail should be unreachable: {msg}");
```
LinZhihao-723:
```suggestion
        panic!("SqsListenerTestState::fail should be unreachable: {msg}");
```
Even though the underlying implementation are the same, `panic` and `unreachable` have very different semantic meanings, as `unreachable` is particularly used to explicitly point out that a path is not possible under any control flow but the branch is added by compiler requirement: https://doc.rust-lang.org/std/macro.unreachable.html.

LinZhihao-723 (changes requested):
One nit pick otherwise lgtm.

components/log-ingestor/tests/test_ingestion_job.rs
```diff
@@ -18,16 +19,103 @@ use clp_rust_utils::{
...
+        Ok(())
+    }
+
+    async fn end(&self) -> Result<()> {
+        Ok(())
+    }
+
+    async fn fail(&self, msg: String) {
+        unreachable!("S3ScannerTestState::fail should be unreachable: {msg}");
+    }
```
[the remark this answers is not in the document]
LinZhihao-723: Please use `panic` here as well.

LinZhihao-723 (changes requested):

LinZhihao-723 (approved):
For the PR title, how about:
```
refactor(log-ingestor): Use mock ingestion state for testing S3 ingestion jobs (resolves #2114).
```
Don't forget the period.

PR #2128, opened 2026-03-25
fix(core): Update `libarchive` download URL to use GitHub releases.

jackluo923:
# Description

The libarchive source tarball download URL (`https://www.libarchive.org/downloads/...`) no longer
works. This PR switches the URL to the GitHub releases page
(`https://github.com/libarchive/libarchive/releases/download/v<version>/...`), which hosts the same
tarballs.

This affects all platforms that build libarchive from source (ubuntu-jammy with v3.5.1,
manylinux_2_28 and musllinux_1_2 with v3.8.0), since they all use the same shared install script.

# Checklist

  breaking change.

# Validation performed

* Verified the new URL returns a valid response for both v3.5.1 and v3.8.0.
* Ran the install script in an Ubuntu Jammy Docker container — libarchive 3.5.1 downloaded, built,
  and installed successfully.

## Summary by CodeRabbit

* **Chores**
  * Updated the source used for downloading a build dependency to retrieve assets from the project's GitHub Releases instead of the previous download site, improving reliability of fetching archives.


components/core/tools/scripts/lib_install/libarchive.sh
```diff
@@ -54,7 +54,7 @@ extracted_dir=${temp_dir}/${package_name}-${version}
 if [ ! -e ${extracted_dir} ] ; then
   tar_filename=${package_name}-${version}.tar.gz
   if [ ! -e ${tar_filename} ] ; then
-    curl -fsSL https://www.libarchive.org/downloads/${tar_filename} -o ${tar_filename}
+    curl -fsSL https://github.com/libarchive/libarchive/releases/download/v${version}/${tar_filename} -o ${tar_filename}
```
junhaoliao: otherwise the unquoted string fails shellcheck; also 100 char limit

```suggestion
    curl -fsSL \
      "https://github.com/libarchive/libarchive/releases/download/v${version}/${tar_filename}" \
      -o "${tar_filename}"
```

junhaoliao (approved):

junhaoliao (approved):
for the title how about
```
fix(core): Update `libarchive` download URL to use GitHub releases.
```

PR #2134, opened 2026-03-26
fix(clp-s::log_converter): Make timezone offset optional in timezone regex (fixes #1952).

### LINKED ISSUE #1952: log-converter: Some timezones not extracted correctly.
### Bug

In the timestamp regex used by log-converter the regex for extracting the timestamp offset from UTC is non-optional, meaning that simple timezones like Z can not be extracted properly.

E.g. the regex won't accept a timestamp like `2026-01-01T11:12:13.145Z`.

### CLP version

0.8.0

### Environment

Ubuntu jammy docker environment.

### Reproduction steps

1. Invoke `log-converter` on a test log file containing timestamps like the one mentioned in the issue description 
2. Observe that timestamps are not perfectly extracted in the resulting IR

gibber9809:
# Description

This PR fixes an issue in log-converter where the timezone `Z` wasn't accepted as a valid timezone on its own since the timezone regex required a timezone offset in matching timezones.

The fix is simply to update the regex to make the offset part of the timezone optional.

# Checklist

  breaking change.

# Validation performed

1. Processed the following dataset into `log-converter`:
```
2026-03-26T07:03:14,788Z INFO [main] One test line.
2026-03-26T07:03:17,222Z INFO [main] Another line.
```
2. Ingested the resulting converted log into `clp-s` using the `--timestamp-key timestamp` option.
3. Observed that the search query `timestamp < timestamp("2026-03-26T07:03:17,222Z")` returned the following result, as expected:
```
{"message":" INFO [main] One test line.\n","timestamp":"2026-03-26T07:03:14,788Z"}
```

## Release Notes

* **Bug Fixes**
  * Improved timestamp parsing flexibility in log conversion to better support varying fractional-seconds formats.


components/core/src/clp_s/log_converter/LogConverter.cpp
```diff
@@ -29,8 +29,8 @@ constexpr std::string_view cTimestampSchema{
         R"(header:(?<timestamp>((\d{2,4}[ /\-]{0,1}[ 0-9]{2}[ /\-][ 0-9]{2})|([ 0-9]{2}[ /\-])"
         R"(((Jan(uary){0,1})|(Feb(ruary){0,1})|(Mar(ch){0,1})|(Apr(il){0,1})|(May)|(Jun(e){0,1})|)"
         R"((Jul(y){0,1})|(Aug(ust){0,1})|(Sep(tember){0,1})|(Oct(ober){0,1})|(Nov(ember){0,1})|)"
-        R"((Dec(ember){0,1}))[ /\-]\d{2,4}))[ T:][ 0-9]{2}:[ 0-9]{2}:[ 0-9]{2})"
-        R"(([,\.:]\d{1,9}){0,1}([ ]{0,1}(UTC){0,1}[\+\-]\d{2}(:{0,1}\d{2}){0,1}Z{0,1}){0,1}))"
+        R"((Dec(ember){0,1}))[ /\-]\d{2,4}))[ T:][ 0-9]{2}:[ 0-9]{2}:[ 0-9]{2}([,\.:]\d{1,9}){0,1})"
+        R"(([ ]{0,1}(UTC){0,1}([\+\-]\d{2}(:{0,1}\d{2}){0,1}){0,1}Z{0,1}){0,1}))"
```
LinZhihao-723, line 33: iiuc, `Z` is a short term for UTC and it should be mutually exclusive to the UTC offset. Does it make more sense to write in this way (or similar)?

```suggestion
        R"((Dec(ember){0,1}))[ /\-]\d{2,4}))[ T:][ 0-9]{2}:[ 0-9]{2}:[ 0-9]{2})"
        R"(([,\.:]\d{1,9}){0,1}(([ ]{0,1}(UTC){0,1}[\+\-]\d{2}(:{0,1}\d{2}))|(Z)){0,1}))"
```

gibber9809: In practice it isn't mutually exclusive. We see timezones like -00Z in the wild. (Edited to a more common case).

PR #2140, opened 2026-03-26
fix(clp-package): Allow `sbin/search.sh` TCP queries to be split into multiple batches (fixes #1898).

### LINKED ISSUE #1898: clp-package: sbin/search.sh searches get dispatched as a single batch.
### Bug

Currently, searches dispatched `sbin/search.sh` have all search sub-tasks dispatched as a single batch by our query-scheduler. This is because the query scheduler seems to be hardcoded to do this whenever the output destination for a task is a network address, which is the case for regular `sbin/search.sh` queries.

When such a query targets a large number of archives, this can can make the query hog all of the cluster resources until completion.

### CLP version

0.8.0

### Environment

Regular package build docker-compose setup.

### Reproduction steps

1. Compress enough data to produce #archives > batch size
2. Dispatch a regular search with `sbin/search.sh`
3. Observe that the number of tasks dispatched in the first batch of the job is equal to the number of searched archives by checking the query scheduler logs.

gibber9809:
# Description

This PR removes the special case in the query scheduler which allowed search jobs to be processed in a single large batch, regardless of the number of archives searched, whenever the output destination was a network address. 

Fixing this issue will help improve fairness in the query scheduler, and shouldn't add untenable amounts of overhead now that the query scheduler has reduced jitter/overhead.

# Checklist

  breaking change.

# Validation performed

1. Ingested enough data into the package to produce more than a single batch worth of archives
2. Issued search via `sbin/search.sh` script and observed that the search job was split into several batches based on query scheduler logs

## Summary by CodeRabbit

* **Refactor**
  * Streamlined the pending job dispatching decision when splitting work into sub-jobs, keeping behaviour the same while simplifying the criteria used for partitioning and dispatch.


junhaoliao (approved):
it looks like the `job.search_config.network_address is None` condition applies to the following use cases:

* `sbin/search.sh` with aggregation: `--count` or `--count-by-time`
* WebUI native search
* Rust API server queries
* MCP server

in other words, it applies to every use case except when `sbin/search.sh` is used with a self-listening TCP server to receive results

removing this condition check correctly allows queries to be split into multiple batches

for the tile, would this be more clear?
```
fix(clp-package): Allow non-TCP search jobs to split into multiple batches (fixes #1898).
```

gibber9809:
> it looks like the `job.search_config.network_address is None` condition applies to the following use cases:
> 
> * `sbin/search.sh` with aggregation: `--count` or `--count-by-time`
> * WebUI native search
> * Rust API server queries
> * MCP server
> 
> in other words, it applies to every use case except when `sbin/search.sh` is used with a self-listening TCP server to receive results
> 
> removing this condition check correctly allows queries to be split into multiple batches
> 
> for the tile, would this be more clear?
> 
> ```
> fix(clp-package): Allow non-TCP search jobs to split into multiple batches (fixes #1898).
> ```

It's actually the opposite -- the if side of the condition is guarding the logic that allows the search tasks to be batched, so previously the `job.search_config.network_address is None` condition meant that _only_ non-TCP jobs would be batched. Now all jobs can be batched.

I think for title maybe we could go with:
```
fix(clp-package): Allow TCP search jobs to be split into multiple batches (fixes #1898).
```

junhaoliao:
> It's actually the opposite -- the if side of the condition is guarding the logic that allows the search tasks to be batched, so previously the `job.search_config.network_address is None` condition meant that _only_ non-TCP jobs would be batched. Now all jobs can be batched.


you're right. sorry for misreading the if-else

junhaoliao:
maybe we can name it 

```
fix(clp-package): Allow `sbin/search.sh` TCP queries to be split into multiple batches (fixes #1898).
```

to make the message more user-interfacing. your call

PR #2150, opened 2026-03-31
feat(log-ingestor)!: Switch buffer submission from soft timeout to hard timeout; Move buffer config from package config to job config (resolves #2136).

### LINKED ISSUE #2136: [log-ingestor] Switch buffer submission from soft timeout to hard timeout
### Request

The current log-ingestor implements buffer submission using a soft timeout mechanism. Specifically, whenever a new file is ingested, the timeout timer is reset. As a result, timeout-based buffer submission is only triggered when no new files are ingested within the configured interval.

However, user feedback indicates that this behavior is suboptimal in scenarios involving continuous ingestion of small files. In such cases:

- Size-based buffer submission is not triggered because individual files are too small to accumulate sufficient data.
- Timeout-based buffer submission is also not triggered, since each newly ingested file resets the timer.

This combination can lead to buffers remaining unsubmitted indefinitely under sustained low-volume ingestion.

After internal discussion, we propose switching to a hard timeout mechanism for buffer submission:

- The timeout timer is not reset upon new file ingestion.
- The timer is only reset after a buffer submission is triggered.

This approach ensures that buffers are flushed within a bounded time window, regardless of ingestion patterns, and better aligns with real-world workloads involving continuous streams of small files.

### Possible implementation

As suggested above.

LinZhihao-723:
# Description

This PR resolves #2136.

This PR changes the compression listener's timeout behavior from soft to hard timeout, and moves buffer configuration from the per-package `LogIngestor` config to the per-job `BaseConfig`.

## Hard timeout

Previously, the listener reset its timer on every `select!` loop iteration (soft timeout), meaning the buffer would only flush after a full timeout period of inactivity. Now, the timer starts when the first entry is added to an empty buffer and is **not** reset by subsequent entries. This guarantees that the oldest buffered object is submitted within the configured timeout duration, regardless of ingestion rate.

Key implementation changes:
- `Buffer::add` now returns whether a size-threshold flush was triggered.
- `Buffer::is_empty` is added to track buffer state transitions.
- `ListenerTask::run` uses a `timer_active` flag and only resets the timer when the buffer transitions from empty to non-empty, or after a size-threshold flush leaves the buffer non-empty.

## Per-job buffer config

Buffer settings (`flush_threshold_bytes`, `timeout_sec`, `channel_capacity`) are moved from the package-level `LogIngestor` config into a new `BufferConfig` struct embedded in `BaseConfig` (per-job). This allows different ingestion jobs to use different buffer configurations.

The following per-package config fields are removed:
- `buffer_flush_timeout` / `buffer_flush_timeout_sec`
- `buffer_flush_threshold`
- `channel_capacity`

These are removed from the Rust `LogIngestor` struct, the Python `LogIngestor` class, the `ClpDbIngestionConnector` struct, the config template, and the Helm chart values/configmap.

# Checklist

# Validation performed

- All workflows passed.
- Added new test cases for hard timeout behavior in `test_compression_listener.rs`.
- Verified no remaining references to the removed per-package buffer fields across `components/` and `tools/`.

## Summary by CodeRabbit

* **Configuration Updates**
  * Removed global buffer/channel tuning options and moved buffering configuration to per-job settings with sensible defaults; templates and deployment values updated.

* **Behavior Changes**
  * Buffering and listener timing refined: flush triggers, timer enable/disable and submission signalling adjusted for more predictable, timely flushes; improved startup/error handling for listener creation.

* **Tests**
  * Added and updated tests validating timing, threshold and flush scenarios, including hard-timeout behaviour.

* **Chores**
  * Updated Helm chart version to 0.2.1-dev.5.

### DIFF
--- a/components/clp-rust-utils/src/job_config/ingestion.rs
+++ b/components/clp-rust-utils/src/job_config/ingestion.rs
@@ -166,6 +166,7 @@ pub mod s3 {
 
     /// Configuration for buffer behavior.
     #[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+    #[serde(default)]
     pub struct BufferConfig {
         /// Size-based flush threshold in bytes.
         ///
@@ -224,3 +225,34 @@ pub mod s3 {
 
 /// Represents the unique identifier for an ingestion job in CLP DB.
 pub type JobId = u64;
+
+#[cfg(test)]
+mod tests {
+    use super::s3::BufferConfig;
+
+    #[test]
+    fn test_buffer_config_partial_override() -> Result<(), serde_json::Error> {
+        let config: BufferConfig = serde_json::from_str(r#"{"timeout_sec": 60}"#)?;
+        let default_config = BufferConfig::default();
+        assert_eq!(config.timeout_sec, 60);
+        assert_eq!(
+            config.flush_threshold_bytes,
+            default_config.flush_threshold_bytes
+        );
+        assert_eq!(config.channel_capacity, default_config.channel_capacity);
+        Ok(())
+    }
+
+    #[test]
+    fn test_buffer_config_empty_object_uses_defaults() -> Result<(), serde_json::Error> {
+        let config: BufferConfig = serde_json::from_str("{}")?;
+        let default_config = BufferConfig::default();
+        assert_eq!(
+            config.flush_threshold_bytes,
+            default_config.flush_threshold_bytes
+        );
+        assert_eq!(config.timeout_sec, default_config.timeout_sec);
+        assert_eq!(config.channel_capacity, default_config.channel_capacity);
+        Ok(())
+    }
+}
--- a/components/clp-rust-utils/src/job_config/ingestion.rs
+++ b/components/clp-rust-utils/src/job_config/ingestion.rs
@@ -65,6 +65,10 @@ pub mod s3 {
         /// Whether to treat the ingested objects as unstructured logs. Defaults to `false`.
         #[serde(default = "default_unstructured")]
         pub unstructured: bool,
+
+        /// Per-job ingestion buffer config.
+        #[serde(default)]
+        pub buffer_config: BufferConfig,
     }
 
     /// Configuration for a SQS listener job.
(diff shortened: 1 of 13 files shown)


components/clp-rust-utils/src/job_config/ingestion.rs
```diff
@@ -160,6 +164,47 @@ pub mod s3 {
...
+
+        /// Time-based flush threshold in seconds.
+        ///
+        /// This is a hard timeout. The buffer is flushed to create a compression job when the
+        /// oldest buffered object has remained in the buffer for at least this duration,
+        /// regardless of the total buffered size. The timer is not reset by newly ingested
+        /// objects.
+        ///
+        /// Defaults to 300 seconds (5 minutes).
+        pub timeout_sec: u64,
```
hoophalab, line 187: Just to confirm this is the intended behavior:

without `#[serde(default = "...")]` on each field, the user must either omit `buffer_config` entirely or specify all three fields. `{"buffer_config": {"timeout_sec": 100}}` would be invalid.

hoophalab (dismissed):
LGTM. One question.

Validations:
1. Listen to an SQS queue and set the timeout to 5 seconds.
2. Upload a small file. ~5 seconds later, a compression job is created.
3. Continuously upload small files. The first compression job is created ~5 seconds after the first file is uploaded.
4. Stop uploading files. The last file is eventually compressed in all my tests.

hoophalab:
`per-package` sounds like an adjective. How about this title:
```
feat(log-ingestor)!: Switch buffer submission from soft timeout to hard timeout; Move buffer config from package config to job config (resolves #2136).
```

components/clp-rust-utils/src/job_config/ingestion.rs
LinZhihao-723: Fixed by adding `#[serde(default)]` to `BufferConfig`. Added unit tests to make sure it works. Also tested myself with a partially overwritten buffer config when submitting ingestion jobs.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -425,18 +414,19 @@ impl ClpDbIngestionConnector {
...
+            base_config,
         );
 
         let listener = Listener::spawn(
-            Buffer::new(submitter, self.buffer_flush_threshold),
-            self.buffer_flush_timeout,
-            self.channel_capacity,
+            Buffer::new(submitter, base_config.buffer_config.flush_threshold_bytes),
+            Duration::from_secs(base_config.buffer_config.timeout_sec),
+            base_config.buffer_config.channel_capacity,
```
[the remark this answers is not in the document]
LinZhihao-723: That should be fine. Restart should be equivalent to a buffer reset. Things won't be submitted earlier than expected.

LinZhihao-723 (commented):
Found an issue: if `channel_capacity` is 0, the channel creation will panic and thus kill the coroutine. In the current flow, we need to create the job entry in the ingestion job table before we create the channel. That job will be silently failed. And even worse, if log-ingestor restarts, it will fail forever because the panic will kill the restart coroutine.
To address this, we need to add a check to make sure the channel capacity is a positive integer. In the latest push, this is a solved problem.
The only issue left is: if channel size 0 is given, the user response would be "internal error" becuase we use anyhow result that can't differentiate it from real internal errors from other config errors. Two solutions:
1. In `clp_ingestion`, we return a thiserror instead of using `anyhow::Result`.
2. We pre-validate the buffer config as we did for the SQS listener config.

However, either way seems out of the scope for this PR. In the OpenAPI doc, we added a notice that a non-positive capacity will result in undefined behavior. That is probably enough for the current PR and release cycle, and the improvements should be tracked in an issue and planned for a future release.

hoophalab (approved):

PR #2163, opened 2026-04-01
feat(clp-s::ffi::sfa): Add `ClpArchiveDecoder` for iterating log events from decompression and search.

Bill-hbrhbr:
# Description

This PR adds `ClpArchiveDecoder` as the utility class for iterating log events decoded from archives. The decoder is created from `ClpArchiveReader`, and supports both in-order and out-of-order decoding.

Once log events are decoded, they are cached inside the decoder to avoid repeating decoding work.

# Checklist

  breaking change.

# Validation performed

## Release Notes

* **New Features**
  * Added archive decoding functionality to extract and retrieve log events from archives
  * Implemented log event data structure containing message, timestamp, and event index information
  * Added support for both ordered and sequential log event consumption

* **Bug Fixes**
  * Improved error handling with additional failure diagnostics

### DIFF
--- a/components/core/src/clp_s/ffi/sfa/ClpArchiveDecoder.cpp
+++ b/components/core/src/clp_s/ffi/sfa/ClpArchiveDecoder.cpp
@@ -4,6 +4,8 @@
 #include <exception>
 #include <memory>
 #include <new>
+#include <optional>
+#include <span>
 #include <string>
 #include <utility>
 
@@ -12,15 +14,22 @@
 
 #include <clp_s/ArchiveReader.hpp>
 #include <clp_s/Defs.hpp>
+#include <clp_s/SchemaReader.hpp>
 
 #include "ClpArchiveReader.hpp"
+#include "LogEvent.hpp"
 #include "SfaErrorCode.hpp"
 
 namespace clp_s::ffi::sfa {
-auto ClpArchiveDecoder::create(ClpArchiveReader& reader)
-        -> ystdlib::error_handling::Result<ClpArchiveDecoder> {
+template <typename ReturnType>
+using Result = ystdlib::error_handling::Result<ReturnType>;
+
+auto ClpArchiveDecoder::create(ClpArchiveReader& reader) -> Result<ClpArchiveDecoder> {
     try {
-        return ClpArchiveDecoder{reader.m_archive_reader->read_all_tables()};
+        return ClpArchiveDecoder{
+                reader.m_archive_reader->read_all_tables(),
+                reader.m_archive_reader->has_log_order()
+        };
     } catch (std::bad_alloc const&) {
         SPDLOG_ERROR("Failed to create ClpArchiveDecoder: out of memory.");
         return SfaErrorCode{SfaErrorCodeEnum::NoMemory};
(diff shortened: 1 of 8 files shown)


components/core/src/clp_s/ffi/sfa/ClpArchiveDecoder.cpp
```diff
@@ -0,0 +1,145 @@
...
+    return std::span<LogEvent const>{m_log_events};
+}
+
+auto ClpArchiveDecoder::append_next_log_event(clp_s::SchemaReader& table) -> bool {
+    std::string message;
+    epochtime_t timestamp{0};
+    int64_t log_event_idx{0};
+
+    if (table.get_next_message_with_metadata(message, timestamp, log_event_idx)) {
+        m_log_events.emplace_back(message, timestamp, log_event_idx);
```
Bill-hbrhbr, line 102:
```suggestion
        m_log_events.emplace_back(std::move(message), timestamp, log_event_idx);
```

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.hpp
```diff
@@ -101,7 +103,19 @@ class ClpArchiveReader {
...
     [[nodiscard]] auto get_file_infos() const -> std::vector<FileInfo> { return m_file_infos; }
 
+    /**
+     * Decodes all log events from the archive.
+     *
+     * @return A result containing the newly constructed `ClpArchiveDecoder` on success, or an
+     * error code indicating the failure:
+     * - Forwards `ClpArchiveDecoder::create`'s return values on failure.
+     */
+    [[nodiscard]] auto decode_all() -> ystdlib::error_handling::Result<ClpArchiveDecoder>;
```
hoophalab, line 113: 1. Is this function const?
2. Is the argument of create const?
    [[nodiscard]] static auto create(ClpArchiveReader const& reader)
3. Can we return
```
ClpArchiveDecoder{
                m_archive_reader->read_all_tables(),
                m_archive_reader->has_log_order()
}
```
inside `auto decode_all()` directly to avoid `friend class ClpArchiveDecoder`?

components/core/src/clp_s/ffi/sfa/ClpArchiveDecoder.cpp
```diff
@@ -0,0 +1,145 @@
...
+auto ClpArchiveDecoder::create(ClpArchiveReader& reader) -> Result<ClpArchiveDecoder> {
+    try {
+        return ClpArchiveDecoder{
+                reader.m_archive_reader->read_all_tables(),
+                reader.m_archive_reader->has_log_order()
+        };
+    } catch (std::bad_alloc const&) {
+        SPDLOG_ERROR("Failed to create ClpArchiveDecoder: out of memory.");
+        return SfaErrorCode{SfaErrorCodeEnum::NoMemory};
+    } catch (std::exception const& ex) {
```
hoophalab, line 36: Not sure if catching `bad_alloc` here and in other functions, and having a dedicated `SfaErrorCodeEnum::NoMemory` is necessarily.

components/core/src/clp_s/ffi/sfa/ClpArchiveReader.hpp
```diff
@@ -9,8 +9,10 @@
 
 #include <ystdlib/error_handling/Result.hpp>
 
+#include "ClpArchiveDecoder.hpp"
```
hoophalab, line 12: should this be a forward declaration instead?

components/core/src/clp_s/ffi/sfa/LogEvent.hpp
```diff
@@ -0,0 +1,29 @@
...
+#define CLP_S_FFI_SFA_LOGEVENT_HPP
+
+#include <cstdint>
+#include <string>
+#include <utility>
+
+namespace clp_s::ffi::sfa {
+class LogEvent {
+public:
+    LogEvent(std::string message, int64_t timestamp, int64_t log_event_idx)
```
hoophalab, line 11:
```suggestion
    explicit LogEvent(std::string message, int64_t timestamp, int64_t log_event_idx)
```

components/core/src/clp_s/ffi/sfa/SfaErrorCode.hpp
```diff
@@ -10,6 +10,7 @@ namespace clp_s::ffi::sfa {
  * Error code enum for SFA API operations.
  */
 enum class SfaErrorCodeEnum : uint8_t {
+    Failure,
```
hoophalab, line 13: How about
```suggestion
    DecodeFailure,
```

components/core/src/clp_s/ffi/sfa/ClpArchiveDecoder.cpp
```diff
@@ -0,0 +1,145 @@
...
+        if (false == found_next_table || table_log_event_idx < next_log_event_idx) {
+            next_table = table;
+            next_log_event_idx = table_log_event_idx;
+            found_next_table = true;
+        }
+    }
+
+    if (false == found_next_table) {
+        return false;
+    }
```
hoophalab, line 141:
```suggestion
    std::shared_ptr<clp_s::SchemaReader> next_table;
    int64_t next_log_event_idx{INT64_MAX};

    for (auto const& table : m_tables) {
        if (table->done()) {
            continue;
        }

        auto const table_log_event_idx{table->get_next_log_event_idx()};
        if (table_log_event_idx < next_log_event_idx) {
            next_table = table;
            next_log_event_idx = table_log_event_idx;
        }
    }

    if (nullptr == next_table) {
        return false;
    }
```

1. Does removing `found_next_table` make this function more understandable?
2. The function looks correct to me but I'm not fully confident. Best to ask @gibber9809 to take a look.

hoophalab (commented):
Some comments.

One high level question:

It might be best to transform the decoder to an iterator. That way, `get_next_log_event` doesn't need to copy and return a `LogEvent` from the internal vector.

The behaviour of `collect_log_events` is also a concern. I haven't looked into it in detail, but it appears that if a user calls `get_next_log_event` first and then `collect_log_events`, and `get_next_log_event` fails, `collect_log_events` might return a span of `LogEvent`s that silently skips the failed event.

With an iterator, the user could simply collect `LogEvent`s into a vector themselves, avoiding this issue.

PR #2168, opened 2026-04-02
fix(taskfile): Isolate log-viewer dependency tasks to fix race condition in webui build (fixes #2167); Bump yscope-dev-utils to y-scope/yscope-dev-utils@2facad8.

### LINKED ISSUE #2167: Race condition in `webui-node-modules` between `deps:log-viewer` extraction and checksum validation
## Bug

A race condition exists in the `webui-node-modules` task (`taskfile.yaml:354`) caused by its concurrently-running deps `deps:log-viewer` and `utils:checksum:validate` for `yscope-log-viewer/node_modules`. Both deps operate on overlapping paths under `components/webui/yscope-log-viewer/`, and since Task runs deps concurrently by default, their operations can interleave in ways that produce incorrect build state.

`webui-node-modules` has four concurrent deps:

```yaml
deps:
  - "deps:log-viewer"                                            # (A)
  - "nodejs-22"                                                  # (B)
  - task: "utils:checksum:validate"                              # (C)
    vars:
      CHECKSUM_FILE: "{{.G_WEBUI_LOG_VIEWER_NODE_MODULES_CHECKSUM_FILE}}"
      INCLUDE_PATTERNS: ["{{.LOG_VIEWER_OUTPUT_DIR}}"]
  - task: "utils:checksum:validate"                              # (D)
    vars:
      CHECKSUM_FILE: "{{.G_WEBUI_PACKAGE_NODE_MODULES_CHECKSUM_FILE}}"
      INCLUDE_PATTERNS: ["{{.PACKAGE_OUTPUT_DIR}}"]
```

Dep (A) calls `yscope-dev-utils:remote:download-and-extract-tar` (`tools/yscope-dev-utils/exports/taskfiles/utils/remote.yaml:66`), which when the stored checksum is stale, runs:

```bash
rm -rf "components/webui/yscope-log-viewer"     # deletes EVERYTHING including node_modules/
mkdir -p "components/webui/yscope-log-viewer"
tar --extract ... --directory "components/webui/yscope-log-viewer" ...
```

Dep (C) calls `utils:checksum:validate` (`tools/yscope-dev-utils/exports/taskfiles/utils/checksum.yaml:82`), which computes a checksum of `yscope-log-viewer/node_modules/` by running `tar --create` piped into `md5sum`.

Because (A) and (C) run concurrently with no ordering between them:

1. (C) starts: `checksum:compute` reads the existing `yscope-log-viewer/node_modules/` (from a previous build) and computes a checksum that matches the stored checksum file, so `checksum:validate` keeps the checksum file intact.
2. (A) starts: `download-and-extract-tar` decides it needs to re-extract. It runs `rm -rf yscope-log-viewer/`, then extracts the tar (which doesn't include `node_modules/`).
3. After both deps finish: `yscope-log-viewer/node_modules/` no longer exists, but the checksum file from (C) still indicates it's valid.
4. `webui-node-modules` may be considered up-to-date (both `generates` files exist and `sources` haven't changed), so it skips `cmds`, leaving `yscope-log-viewer/node_modules/` missing.
5. Downstream tasks fail because the `webui` build expects `yscope-log-viewer/node_modules/` to exist.

Alternatively, if (A) starts deleting `yscope-log-viewer/` while (C) is mid-way through `tar --create | md5sum`, the `tar` process encounters files that vanish mid-read. Since `IGNORE_ERROR` is set to `"true"` inside `checksum:validate`, the error is swallowed and the resulting checksum file is unreliable.

This is analogous to the race condition described in #1477 where concurrent tasks accessed shared cmake settings files.

## CLP version

22c1275f9cb32309067865ba07fd696831d7e937

## Environment

All platforms (Linux, macOS). The issue depends on task scheduling timing, so it may manifest intermittently.

## Reproduction steps

The issue is timing-dependent and more likely to trigger on a clean or stale build where `deps:log-viewer` decides to re-extract:

1. Run a full build: `task webui`
2. Make a change that causes `deps:log-viewer` to re-extract (e.g., change the `FILE_SHA256` or `URL` in `taskfiles/deps/main.yaml`, or delete `build/deps/log-viewer.md5`)
3. Run `task webui` again
4. Observe that the build intermittently fails because `yscope-log-viewer/node_modules/` is missing or incomplete

The race is easiest to reproduce by adding an artificial delay in `download-and-extract-tar` between `rm -rf` and `tar --extract`, which widens the window for `checksum:validate` to observe the deleted state.

junhaoliao:
# Description

Isolate the log-viewer's source download, `node_modules` installation, and production build into a
self-contained `deps:yscope-log-viewer` taskfile (`taskfiles/deps/log-viewer.yaml`). This fixes a
race condition where `download-and-extract-tar` (which does `rm -rf` on the source directory) could
run concurrently with `checksum:validate` for the log-viewer's `node_modules`, causing intermittent
build failures (fixes #2167).

Key changes:

* **New `taskfiles/deps/log-viewer.yaml`**: Contains `source`, `node-modules`, and `build` tasks
  with proper dependency ordering. `node-modules-checksum-validate` ensures
  `download-and-extract-tar` completes before `node_modules` is validated (eliminating the race).
  The `build` task recomputes checksums after vite's build to account for `node_modules/.vite-temp`.
* **Move `nodejs-22` to `taskfiles/toolchains.yaml`**: Use `yscope-dev-utils`'s
  `download-and-extract-tar` with pinned per-platform SHA256 hashes, replacing the old
  `download-and-extract-tar` and `nodejs` tasks in `taskfile.yaml`.
* **Decouple `webui-node-modules` from log-viewer**: The webui's node-modules task no longer
  installs or checksums the log-viewer's `node_modules`; it only manages the webui package's own
  workspace dependencies.
* **Update `clean-webui`**: Only cleans the webui package's `node_modules` and build output; the
  log-viewer's cleanup is handled by its own `clean-node-modules` task.

# Checklist

  breaking change.

# Validation performed

## 1. `deps:yscope-log-viewer` caching after clean build

Verifies that after `rm -rf .task/ build/`, the second run is fully cached (no re-extraction,
no re-install, no rebuild).

**Commands**:
```
rm -rf .task/ build/
task deps:yscope-log-viewer   # Run 1
task deps:yscope-log-viewer   # Run 2
```

**Run 2 output**:
```
task: Task "deps:yscope-dev-utils:remote:download-and-extract-tar-/.../yscope-log-viewer" is up to date
task: Task "deps:yscope-log-viewer:node-modules" is up to date
task: Task "deps:yscope-log-viewer:build" is up to date
```

## 2. `webui` build succeeds and caches correctly

Verifies the full `webui` pipeline builds from a clean state and that log-viewer tasks are cached
on the second run.

**Commands**:
```
rm -rf .task/ build/
task webui   # Run 1
task webui   # Run 2
```

**Run 2 output** (log-viewer tasks):
```
task: Task "deps:yscope-dev-utils:remote:download-and-extract-tar-/.../yscope-log-viewer" is up to date
task: Task "deps:yscope-log-viewer:node-modules" is up to date
task: Task "deps:yscope-log-viewer:build" is up to date
task: Task "webui" is up to date
```

**Note**: `webui-node-modules` re-runs on the second invocation. This is a pre-existing issue
(confirmed on `main`) unrelated to this PR.

## 3. YAML lint passes

**Command**:
```
task lint:check-yaml
```

**Output**:
```
task: [lint:yaml] . ".../build/lint-venv/bin/activate"
yamllint --strict \
  --config-file "tools/yscope-dev-utils/exports/lint-configs/.yamllint.yml" \
  ".github" \
  "components/core/.clang-format" \
  ...
  "taskfile.yaml" \
  "taskfiles" \
  ...
```

No errors.

## Summary by CodeRabbit

* **Chores**
  * Moved log‑viewer dependency to an external yscope‑log‑viewer flow and simplified the Web UI node_modules pipeline.
  * Centralized checksum handling to validate only the Web UI top‑level node_modules.
  * Added a cleanup task that removes top‑level Web UI node_modules (client/common/server/top) without touching embedded log‑viewer modules.
  * Introduced Node.js 22 toolchain support for the node‑modules build flow and updated rebuild triggers accordingly.


Bill-hbrhbr (changes requested):
I think the fundamental design issue is that we should put `yscope-log-viewer` along with its `node-modules` initialization into a separate module. What I suggest is the following refactor:
- Put `nodejs22` install into `taskfiles/toolchain.yaml`
- Move `deps:yscope-log-viewer` into a standalone task file like `taskfiles/deps/yscope-log-viewer.yaml` and handle the log-viewer-specific node modules initialization there. We should also do the checksum over there, rather than inside the main taskfile.

taskfiles/deps/yscope-log-viewer.yaml
```diff
@@ -0,0 +1,68 @@
...
+        vars:
+          CHECKSUM_FILE: "{{.G_WEBUI_LOG_VIEWER_NODE_MODULES_CHECKSUM_FILE}}"
+          INCLUDE_PATTERNS: ["{{.OUTPUT_DIR}}"]
+
+  source:
+    internal: true
+    sources:
+      - "{{.TASKFILE}}"
+    generates:
+      - "{{.G_DEPS_LOG_VIEWER_CHECKSUM_FILE}}"
```
Bill-hbrhbr:
```suggestion
```
We should let `yscope-dev-utils:remote:download-and-extract-tar` handle checksum and status checking.

taskfile.yaml
```diff
@@ -101,12 +97,17 @@ tasks:
       - "rm -rf '{{.G_RUST_BUILD_DIR}}'"
 
   clean-webui:
+    cmds:
+      - task: "clean-webui-package-node-modules"
+      - task: "deps:yscope-log-viewer:clean"
```
Bill-hbrhbr:
```suggestion
      - task: "deps:yscope-log-viewer:clean-node-modules"
```

taskfiles/deps/yscope-log-viewer.yaml
```diff
@@ -0,0 +1,68 @@
+version: "3"
+
+vars:
+  G_DEPS_LOG_VIEWER_CHECKSUM_FILE: "{{.G_BUILD_DIR}}/deps/log-viewer.md5"
+  G_WEBUI_LOG_VIEWER_NODE_MODULES_CHECKSUM_FILE:
+    "{{.G_BUILD_DIR}}/webui-log-viewer-node-modules.md5"
+  G_WEBUI_LOG_VIEWER_SRC_DIR: "{{.G_WEBUI_SRC_DIR}}/yscope-log-viewer"
+
+tasks:
+  clean:
```
Bill-hbrhbr:
```suggestion
  clean-node-modules:
```

taskfiles/deps/yscope-log-viewer.yaml
```diff
@@ -0,0 +1,68 @@
...
+    cmds:
+      - "rm -rf '{{.G_WEBUI_LOG_VIEWER_SRC_DIR}}/node_modules'"
+
+  default:
+    deps: ["node-modules"]
+
+  node-modules-checksum:
+    internal: true
+    vars:
+      OUTPUT_DIR: "{{.G_WEBUI_LOG_VIEWER_SRC_DIR}}/node_modules"
```
Bill-hbrhbr:
```suggestion
```
Maybe extract this to a global variable? `G_WEBUI_LOG_VIEWER_NODE_MODULES_DIR`?

taskfiles/deps/yscope-log-viewer.yaml
```diff
@@ -0,0 +1,68 @@
...
+    deps:
+      - "source"
+    cmds:
+      - task: ":yscope-dev-utils:checksum:validate"
+        vars:
+          CHECKSUM_FILE: "{{.G_WEBUI_LOG_VIEWER_NODE_MODULES_CHECKSUM_FILE}}"
+          INCLUDE_PATTERNS: ["{{.OUTPUT_DIR}}"]
```
Bill-hbrhbr:
```suggestion
          INCLUDE_PATTERNS: ["{{.G_WEBUI_LOG_VIEWER_NODE_MODULES_DIR}}"]
```

taskfiles/deps/yscope-log-viewer.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+  node-modules:
+    internal: true
+    vars:
+      OUTPUT_DIR: "{{.G_WEBUI_LOG_VIEWER_SRC_DIR}}/node_modules"
```
Bill-hbrhbr:
```suggestion
```

taskfiles/deps/yscope-log-viewer.yaml
```diff
@@ -0,0 +1,68 @@
...
+    deps:
+      - "node-modules-checksum"
+      - "::toolchains:nodejs-22"
+    cmds:
+      - "PATH='{{.G_NODEJS_22_BIN_DIR}}':$PATH npm clean-install"
+      # This command must be last
+      - task: ":yscope-dev-utils:checksum:compute"
+        vars:
+          CHECKSUM_FILE: "{{.G_WEBUI_LOG_VIEWER_NODE_MODULES_CHECKSUM_FILE}}"
+          INCLUDE_PATTERNS: ["{{.OUTPUT_DIR}}"]
```
Bill-hbrhbr:
```suggestion
          INCLUDE_PATTERNS: ["{{.G_WEBUI_LOG_VIEWER_NODE_MODULES_DIR}}"]
```

taskfiles/deps/yscope-log-viewer.yaml
```diff
@@ -0,0 +1,68 @@
...
+tasks:
+  clean:
+    internal: true
+    cmds:
+      - "rm -rf '{{.G_WEBUI_LOG_VIEWER_SRC_DIR}}/node_modules'"
+
+  default:
+    deps: ["node-modules"]
+
+  node-modules-checksum:
```
Bill-hbrhbr:
```suggestion
  node-modules-checksum-validate:
```
Feels like this is a more fitting name.

taskfiles/deps/yscope-log-viewer.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - "{{.G_DEPS_LOG_VIEWER_CHECKSUM_FILE}}"
+      - "{{.G_NODEJS_22_CHECKSUM_FILE}}"
+      - "{{.G_WEBUI_LOG_VIEWER_SRC_DIR}}/package-lock.json"
+      - "{{.G_WEBUI_LOG_VIEWER_SRC_DIR}}/package.json"
+      - "{{.TASKFILE}}"
+    dir: "{{.G_WEBUI_LOG_VIEWER_SRC_DIR}}"
+    generates:
+      - "{{.G_WEBUI_LOG_VIEWER_NODE_MODULES_CHECKSUM_FILE}}"
+    deps:
+      - "node-modules-checksum"
```
Bill-hbrhbr:
```suggestion
      - "node-modules-checksum-validate"
```

taskfile.yaml
```diff
@@ -301,130 +301,45 @@ tasks:
...
+  # NOTE: The webui build depends on two different node_modules directories:
+  # * log-viewer submodule, handled by `deps:yscope-log-viewer:node-modules`
+  # * the top-level one we call "package", handled here
   webui-node-modules:
     internal: true
     vars:
-      LOG_VIEWER_OUTPUT_DIR: "{{.G_WEBUI_SRC_DIR}}/yscope-log-viewer/node_modules"
       PACKAGE_OUTPUT_DIR: "{{.G_WEBUI_SRC_DIR}}/node_modules"
     sources:
-      - "{{.G_DEPS_LOG_VIEWER_CHECKSUM_FILE}}"
```
Bill-hbrhbr, line 363: Maybe we should put this back? Will there be cases where log viewer source code changes, but nothing affects the node modules and its checksum.

taskfiles/deps/yscope-log-viewer.yaml
Bill-hbrhbr: or `G_WEBUI_LOG_VIEWER_OUTPUT_DIR` (similar to what you've used before).

Bill-hbrhbr (changes requested):

taskfile.yaml
junhaoliao: right. let me double check and add it back

junhaoliao: actually i remember this is intentional. `webui-node-modules` should only be responsible for ensuring the webui components's node_modules are cleanly built.

do note that the log-viewer source code is consumed at
- task `webui`, which depends on 
- task webui-node-modules, which depends on
- task `deps:yscope-log-viewer:node-modules`, which depends on
- task `deps:yscope-log-viewer:node-modules-checksum-validate`, which depends on
`deps:yscope-log-viewer:source`.

we have seen that
- `webui` includes `G_WEBUI_PACKAGE_NODE_MODULES_CHECKSUM_FILE` in its sources
- `webui-node-modules` includes `G_WEBUI_LOG_VIEWER_NODE_MODULES_CHECKSUM_FILE` in its sources, and generates `G_WEBUI_PACKAGE_NODE_MODULES_CHECKSUM_FILE`
- `deps:yscope-log-viewer:node-modules` includes `G_DEPS_LOG_VIEWER_CHECKSUM_FILE` in its sources, and generates `G_WEBUI_LOG_VIEWER_NODE_MODULES_CHECKSUM_FILE`
- `source` generates `G_DEPS_LOG_VIEWER_CHECKSUM_FILE`

So any change to the log viewer source already propagates through the checksum chain. Additionally, `webui` has `**/*` (excluding `node_modules` and `dist`) in its sources, which directly catches log viewer source file changes.

---

Now i think about it, maybe we should move the log-viewer building to deps/yscope-log-viewer.yaml so the concerns will be cleanly isolated. let me try making the changes now

taskfiles/deps/log-viewer.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+  source:
+    internal: true
+    run: "once"
+    deps:
+      - ":utils:init"
+    cmds:
+      - task: ":yscope-dev-utils:remote:download-and-extract-tar"
+        vars:
+          CHECKSUM_EXCLUDE_PATTERNS:
```
junhaoliao, line 98: @Bill-hbrhbr 
this isn't a valid parameter in the `":yscope-dev-utils:remote:download-and-extract-tar"` task if i'm not wrong

junhaoliao: please ignore the above comment. i believe this can be resolved once https://github.com/y-scope/yscope-dev-utils/pull/113#pullrequestreview-4166940714 is merged

taskfiles/deps/log-viewer.yaml
```diff
@@ -0,0 +1,105 @@
+version: "3"
+
+vars:
+  G_DEPS_LOG_VIEWER_SRC_DIR: "{{.G_WEBUI_SRC_DIR}}/yscope-log-viewer"
+  G_DEPS_LOG_VIEWER_NODE_MODULES_DIR: "{{.G_DEPS_LOG_VIEWER_SRC_DIR}}/node_modules"
+  G_DEPS_LOG_VIEWER_BUILD_DIR: "{{.G_DEPS_DIR}}/log-viewer-build"
```
junhaoliao:
```suggestion
  G_DEPS_LOG_VIEWER_BUILD_DIR: "{{.G_DEPS_DIR}}/log-viewer-build"
  G_DEPS_LOG_VIEWER_SRC_DIR: "{{.G_WEBUI_SRC_DIR}}/yscope-log-viewer"
  G_DEPS_LOG_VIEWER_NODE_MODULES_DIR: "{{.G_DEPS_LOG_VIEWER_SRC_DIR}}/node_modules"
```

junhaoliao:
i approve 4b65567 and 585cf3e

Bill-hbrhbr (approved):

PR #1686, opened 2025-11-28
docs(integration-tests): Restructure `dev-docs` directory to accommodate package integration test docs.

quinntaylormitchell:
# Description

This PR restructures the `dev-docs` directory in preparation for docs on package integration tests and binary integration tests.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

Built and served docs.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **Documentation**
  * Restructured integration tests documentation with dedicated sections for testing CLP package and binaries.
  * Updated testing documentation navigation structure and cross-references for improved content organization.
  * Added new reference pages for CLP package and binary testing procedures (currently under construction).

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1212853123119364
  - https://app.asana.com/0/0/1212888468915965
  - https://app.asana.com/0/0/1212851383828818

PR #2181, opened 2026-04-07
feat(integration-tests): Introduce `IntegrationTestDataset` fixtures; move test data to central directory.

quinntaylormitchell:
# Description

This PR stores test datasets in a centralized `/data` directory, and makes them available to all tests through session-scoped fixtures in `datasets.py`. Each test dataset directory carries a `metadata.json` file which specifies certain information about the dataset needed for testing. The information in this file is loaded into a pydantic object (`IntegrationTestDatasetMetadata`) which validates the format of the file.

NOTE: the old test datasets located at `integration-tests/tests/package_tests/clp_json/data/json-multifile` and `integration-tests/tests/package_tests/clp_text/data/text-multifile` are not removed in this PR, and they are still referenced in code. They will be removed when appropriate as the rest of the package testing code is updated and developed.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **Tests**
  * Added reusable session-scoped dataset fixtures and extensive sample log files (JSON/text) for integration testing
  * Introduced metadata-driven validation for datasets and stronger runtime checks of dataset contents

* **Documentation**
  * Added a README describing dataset layout and metadata schema for integration tests

* **Chores**
  * Test configuration now auto-loads the fixtures module for smoother test runs


integration-tests/tests/utils/classes.py
```diff
@@ -58,15 +59,83 @@ def _static_paths(self) -> list[Path]:
...
+    Metadata for a sample dataset. All `<dataset_name>/metadata.json` files must conform to this
+    schema.
+    """
+
+    dataset_name: str
+    unstructured: bool
+    timestamp_key: str | None
+    begin_ts: int
+    end_ts: int
+    logs_subdir: str
```
Bill-hbrhbr, line 73:
```suggestion
    logs_dir: str
```
would this be more straigtforward?

integration-tests/tests/utils/classes.py
```diff
@@ -58,15 +59,83 @@ def _static_paths(self) -> list[Path]:
...
+            file_path_abs = self.logs_path / file_path
+            validate_file_exists(file_path_abs)
+
+    @property
+    def metadata_file_path(self) -> Path:
+        """:return: The absolute path to the file containing metadata for the dataset."""
+        return self.path_to_dataset_root / "metadata.json"
+
+    @property
+    def logs_path(self) -> Path:
```
Bill-hbrhbr, line 122:
```suggestion
    def logs_dir(self) -> Path:
```
Dir is more clear right? Also matches the metadata key.

integration-tests/tests/utils/classes.py
```diff
@@ -58,15 +59,83 @@ def _static_paths(self) -> list[Path]:
...
+        """:return: The absolute path to the subdirectory containing logs."""
+        return self.path_to_dataset_root / self.metadata.logs_subdir
```
Bill-hbrhbr, line 124:
```suggestion
        return self.path_to_dataset_root / self.metadata.logs_dir
```
ditto

integration-tests/tests/utils/classes.py
```diff
@@ -58,15 +59,83 @@ def _static_paths(self) -> list[Path]:
...
+    columns_file_name: str | None
+
+
 @dataclass
 class IntegrationTestDataset:
-    """Metadata for a sample dataset."""
+    """Path layout and metadata storage for a sample dataset."""
 
     #: The name of the dataset (for logging purposes).
     dataset_name: str
```
Bill-hbrhbr: Should `dataset_name` be parsed from metadata?

integration-tests/tests/utils/classes.py
```diff
@@ -58,15 +59,83 @@ def _static_paths(self) -> list[Path]:
         return [self.test_data_path]
 
 
+class IntegrationTestDatasetMetadata(BaseModel):
+    """
+    Metadata for a sample dataset. All `<dataset_name>/metadata.json` files must conform to this
```
Bill-hbrhbr, line 64: If by design, all future test data metadata should conform to this schema, maybe it's also worthwhile to add a README inside `integration-tests/tests/data`.

integration-tests/tests/data/json_multifile/logs/sts-135-2011-07-08.jsonl
```diff
@@ -0,0 +1,8 @@
+{"timestamp":1310138944000,"mission":"STS-135","mission_day_index":0,"event":"SRB_IGNITION_CONFIRMED","subsystem":"PROP","level":"INFO","detail":"Solid rocket boosters report stable ignition and rising thrust","line_index":0}
```
Bill-hbrhbr: Check for CRLF (\r\n) line endings in all jsonl.

integration-tests/tests/data/json_multifile/columns.json
```diff
@@ -0,0 +1,76 @@
+{
+    "columns": [
+        {
+            "Column": "mission",
+            "Type": "varchar",
+            "Extra": "",
+            "Comment": "",
+            "Precision": null,
+            "Scale": null,
+            "Length": 2147483647
```
Bill-hbrhbr: is this a fixed schema? Do you think it'd be worthwhile to use pydantic model to verify schema correctness?

integration-tests/tests/utils/classes.py
```diff
@@ -58,15 +59,83 @@ def _static_paths(self) -> list[Path]:
...
+
+        # Validate metadata properties.
+        validate_dir_exists(self.logs_path)
 
-    #: A dictionary of metadata describing the dataset.
-    metadata_dict: dict[str, Any]
+        if self.columns_file_path is not None:
+            validate_file_exists(self.columns_file_path)
+
+        if self.metadata.dataset_name != self.dataset_name:
```
Bill-hbrhbr: If we parse dataset name directly from metadata. we won't have this problem.

Bill-hbrhbr (changes requested):

integration-tests/tests/data/json_multifile/logs/sts-135-2011-07-08.jsonl
quinntaylormitchell: Right on, thanks.

integration-tests/tests/data/json_multifile/columns.json
quinntaylormitchell: It is a fixed schema, and using a pydantic model to verify would be worthwhile, I agree. That being said, I think that I'm going to remove everything related to `columns.json` from this PR, because we don't actually need it until `clp-presto` testing gets put up. I shouldn't have included this when I split this PR's material off from my full implementation branch.

I'll make a note to update my approach with the pydantic model verification.

integration-tests/tests/utils/classes.py
quinntaylormitchell: Agreed; README has been added.

I've also made a personal note to mention it when I do the integration test docs.

integration-tests/tests/utils/classes.py
quinntaylormitchell: I'd like it to be clear that this parameter holds the **name** of the subdirectory that holds logs within the dataset directory; it doesn't hold the absolute path to the logs directory. To me it feels like changing it to `logs_dir` would obscure that meaning.

integration-tests/tests/utils/classes.py
quinntaylormitchell: Yeah that's fair enough, I'll do that.

When I think about it, the data class might not even need a separate parameter for `dataset_name` (we could just access it from the metadata when we need it). I'm going to leave it there for now, but I'll make a note to reassess once we have more testing code merged.

integration-tests/tests/utils/classes.py
quinntaylormitchell: Agreed.

integration-tests/tests/utils/classes.py
quinntaylormitchell: I'd like to communicate that this property returns the absolute path to the logs, so I think `logs_path` is more appropriate.

(Also see my reply to your earlier comment; this isn't the same information as the metadata key.)

integration-tests/tests/utils/classes.py
quinntaylormitchell: See reply.

joe-spring: Fixed in a1ec71dc97860b956f21703528c289c45f8f5cd2: `resolveRiskFieldId` under `customer_risks` now applies `RISK_FLAG_MAP` for alleged risks, keeps `customer_*` keys as direct columns, and prefixes everything else with `analyzed_custom_fields.` for inherent risks. Added four dedicated tests (alleged-via-map, customer-prefixed direct, inherent-prefixed, mixed).

integration-tests/tests/data/text_multifile/metadata.json
```diff
@@ -0,0 +1,16 @@
+{
+    "dataset_name": "text_multifile",
+    "unstructured": true,
+    "timestamp_key": null,
+    "begin_ts": 92554380000,
+    "end_ts": 93596040000,
```
junhaoliao: The last event's datetime string in `integration-tests/tests/data/text_multifile/logs/apollo-17_day13.txt` is `1972-12-19T06:53:27.004` (`Z` - UTC) which should be

```suggestion
    "end_ts": 93596007004,
```

it doesn't seem like any test is referencing these sample files yet, so the mismatch isn't easily discover-able. that said, shall we derive those stats directly from the logs (if we can have high confidence in the tester code timestamp parsing correctness; if not, then) / add some sanity check functions to ensure the metadata is matching the actual logs? i'm fine deferring the exercise to another PR; if you agree, let's ask the rabbit to create an issue to track

integration-tests/tests/data/README.md
junhaoliao, line 1: @kirkrodrigues 

shall we start creating integration-test-specific docs in `<project-root>/docs/` or do you think this `README.md` is fine to be kept here for now?

integration-tests/tests/data/README.md
```diff
@@ -0,0 +1,86 @@
...
+1. Each dataset must be located in its own directory.
+2. Each dataset directory must be a first-level directory within `clp/integration-tests/tests/data`,
+   i.e., dataset directories cannot be nested.
+
+## Dataset contents
+
+When adding or changing a dataset directory within `clp/integration-tests/tests/data`, the following
+rules must be observed:
+
+1. Each dataset directory must contain a first-level subdirectory that holds all log files for the
```
junhaoliao:
```suggestion
1. Each dataset directory must contain a first-level subdirectory that holds all log files for the
```

integration-tests/tests/data/README.md
```diff
@@ -0,0 +1,86 @@
...
+   <dd>The earliest timestamp present in the dataset (ms).</dd>
+   <dt><code>end_ts</code></dt>
+   <dd>The latest timestamp present in the dataset (ms).</dd>
+   <dt><code>logs_subdir</code></dt>
+   <dd>The name of the subdirectory containing logs.</dd>
+   <dt><code>file_names</code></dt>
+   <dd>A list of the files within `logs_subdir`.</dd>
+   <dt><code>single_match_wildcard_query</code></dt>
+   <dd>A wildcard query which, when searched, will match a single log message in the dataset.</dd>
+   </dl>
```
junhaoliao: this doesn't seem markdown-native - 

can a table or a list better represent this?

junhaoliao (changes requested):

integration-tests/tests/data/text_multifile/metadata.json
quinntaylormitchell: Good catch, thanks!

And good point too; I think that we should defer that to another PR. My initial reaction is that I don't want to add too much complexity or over-process the logs by parsing the timestamp or anything. I think that all in all, there will only be a handful of datasets (my guess is less than 10) ever needed for the testing suite, and they should all be small like this one is, so it'll be feasible to check them manually. I will think about it though!

integration-tests/tests/data/README.md
quinntaylormitchell: fwiw I have already started the docs but #1686 is on hold until the bulk of the testing suite is finished.

integration-tests/tests/data/README.md
quinntaylormitchell: I can do a list if you would prefer

integration-tests/tests/data/README.md
Bill-hbrhbr: 1. We have integration specific docs inside `docs/src/dev-docs/testing/integration-tests.md`.
2. I think we can move this README.md to the docs directory in a follow-up PR. Preferably after the current refactoring operation is complete.

integration-tests/tests/utils/classes.py
```diff
@@ -58,15 +59,66 @@ def _static_paths(self) -> list[Path]:
...
+        # Validate metadata properties.
+        validate_dir_exists(self.logs_path)
 
-    #: Path to the dataset logs.
-    path_to_dataset_logs: Path
+        if self.metadata.begin_ts > self.metadata.end_ts:
+            pytest.fail(
+                f"Dataset metadata failure: `begin_ts` '{self.metadata.begin_ts}' is larger than"
+                f" `end_ts` '{self.metadata.end_ts}'"
+            )
```
Bill-hbrhbr: This should be `ValueError`. `metadata.json` is not the outcome of a particular test.

integration-tests/tests/utils/classes.py
```suggestion
    def test_data_dir(self) -> Path:
```

integration-tests/tests/utils/classes.py
```diff
@@ -58,15 +59,66 @@ def _static_paths(self) -> list[Path]:
         return [self.test_data_path]
```
```suggestion
        return [self.test_data_dir]
```

integration-tests/tests/utils/classes.py
```diff
@@ -58,15 +59,66 @@ def _static_paths(self) -> list[Path]:
...
+    single_match_wildcard_query: str
+
+
 @dataclass
 class IntegrationTestDataset:
-    """Metadata for a sample dataset."""
+    """Path layout and metadata storage for a sample dataset."""
+
+    #: Absolute path to the dataset root.
+    path_to_dataset_root: Path
```
```suggestion
    dataset_root_dir: Path
```

integration-tests/tests/utils/classes.py
```diff
@@ -58,15 +59,66 @@ def _static_paths(self) -> list[Path]:
...
+    file_names: list[str]
+    single_match_wildcard_query: str
+
+
 @dataclass
 class IntegrationTestDataset:
-    """Metadata for a sample dataset."""
+    """Path layout and metadata storage for a sample dataset."""
+
+    #: Absolute path to the dataset root.
```
```suggestion
    #: Absolute path to the dataset root directory.
```

integration-tests/tests/utils/classes.py
```diff
@@ -58,15 +59,66 @@ def _static_paths(self) -> list[Path]:
...
+
+    @property
+    def metadata_file_path(self) -> Path:
+        """:return: The absolute path to the file containing metadata for the dataset."""
+        return self.path_to_dataset_root / "metadata.json"
+
+    @property
+    def logs_path(self) -> Path:
+        """:return: The absolute path to the subdirectory containing logs."""
+        return self.path_to_dataset_root / self.metadata.logs_subdir
```
```suggestion
        return self.dataset_root_dir / self.metadata.logs_subdir
```

integration-tests/tests/utils/classes.py
```diff
@@ -58,15 +59,66 @@ def _static_paths(self) -> list[Path]:
...
-    #: A dictionary of metadata describing the dataset.
-    metadata_dict: dict[str, Any]
+        for file_path in self.metadata.file_names:
+            file_path_abs = self.logs_path / file_path
+            validate_file_exists(file_path_abs)
+
+    @property
+    def metadata_file_path(self) -> Path:
+        """:return: The absolute path to the file containing metadata for the dataset."""
+        return self.path_to_dataset_root / "metadata.json"
```
```suggestion
        return self.dataset_root_dir / "metadata.json"
```

integration-tests/tests/utils/classes.py
```diff
@@ -58,15 +59,66 @@ def _static_paths(self) -> list[Path]:
...
+    #: Pydantic model of metadata describing the dataset.
+    metadata: IntegrationTestDatasetMetadata = field(init=False)
 
     #: The name of the dataset (for logging purposes).
-    dataset_name: str
+    dataset_name: str = field(init=False)
+
+    def __post_init__(self) -> None:
+        """Validate data members and load metadata."""
+        validate_dir_exists(self.path_to_dataset_root)
```
```suggestion
        validate_dir_exists(self.dataset_root_dir)
```

integration-tests/tests/utils/classes.py
Bill-hbrhbr: Instead of **name**, I think **relative path** is a better concept here. That being said, `logs_subdir` is fine.

integration-tests/tests/utils/classes.py
Bill-hbrhbr: I think it's clearer to say "this property returns the absolute path to the logs **directory**"

integration-tests/tests/fixtures/datasets.py
```diff
@@ -0,0 +1,28 @@
...
+)
+
+
+@pytest.fixture(scope="session")
+def json_multifile(
+    integration_test_path_config: IntegrationTestPathConfig,
+) -> IntegrationTestDataset:
+    """Returns an object corresponding to the `json_multifile` test dataset."""
+    return IntegrationTestDataset(
+        path_to_dataset_root=integration_test_path_config.test_data_path / "json_multifile",
```
Bill-hbrhbr:
```suggestion
        dataset_root_dir=integration_test_path_config.test_data_dir / "json_multifile",
```
Let's also update `test_data_path` to `test_data_dir`.

integration-tests/tests/fixtures/datasets.py
```diff
@@ -0,0 +1,28 @@
...
+    )
+
+
+@pytest.fixture(scope="session")
+def text_multifile(
+    integration_test_path_config: IntegrationTestPathConfig,
+) -> IntegrationTestDataset:
+    """Returns an object corresponding to the `text_multifile` test dataset."""
+    return IntegrationTestDataset(
+        path_to_dataset_root=integration_test_path_config.test_data_path / "text_multifile",
```
Bill-hbrhbr:
```suggestion
        dataset_root_dir=integration_test_path_config.test_data_dir / "text_multifile",
```

integration-tests/tests/data/README.md
```diff
@@ -0,0 +1,78 @@
...
+   and should conform to the following format:
+
+   ```python
+   @pytest.fixture(scope="session")
+   def dataset_name(
+       integration_test_path_config: IntegrationTestPathConfig,
+   ) -> IntegrationTestDataset:
+       """Returns an object corresponding to the `dataset_name` test dataset."""
+       return IntegrationTestDataset(
+           path_to_dataset_root=integration_test_path_config.test_data_path / "dataset_name",
```
Bill-hbrhbr:
```suggestion
           dataset_root_dir=integration_test_path_config.test_data_dir / "dataset_name",
```

integration-tests/tests/data/README.md
```diff
@@ -0,0 +1,78 @@
+# Integration test data
+
+This directory (`clp/integration-tests/tests/data`) contains test datasets for use throughout the
```
Bill-hbrhbr:
```suggestion
The directory (`integration-tests/tests/data`) contains test datasets for use throughout the
```

integration-tests/tests/data/README.md
```diff
@@ -0,0 +1,78 @@
...
+   ) -> IntegrationTestDataset:
+       """Returns an object corresponding to the `dataset_name` test dataset."""
+       return IntegrationTestDataset(
+           path_to_dataset_root=integration_test_path_config.test_data_path / "dataset_name",
+       )
+   ```
+
+3. While it is possible to read the logs from a dataset directly (i.e., without using a 
+   session-scoped dataset fixture), it is not recommended, as dataset metadata is needed in several
+   verification processes throughout the testing system.
```
Bill-hbrhbr:
```suggestion
3. Using dataset fixtures is encouraged over reading logs directly, since verification relies on the associated dataset metadata.
```

integration-tests/tests/data/README.md
```diff
@@ -0,0 +1,78 @@
...
+   | `logs_subdir` | The name of the subdirectory containing logs. |
+   | `file_names` | A list of the files within `logs_subdir`. |
+   | `single_match_wildcard_query` | A wildcard query which, when searched, will match a single log message in the dataset. |
+
+## Accessing datasets within the testing system
+
+To access a dataset from within the test system, the following rules should be observed:
+
+1. All datasets should have their own session-scoped fixture in
+   `clp/integration-tests/tests/fixtures/datasets.py`. Test code should access this fixture to
```
Bill-hbrhbr:
```suggestion
   `integration-tests/tests/fixtures/datasets.py`. Test code should access this fixture to
```

integration-tests/tests/data/README.md
```diff
@@ -0,0 +1,78 @@
...
+When augmenting or changing the contents of `clp/integration-tests/tests/data`, the following rules
+must be observed:
+
+1. Each dataset must be located in its own directory.
+2. Each dataset directory must be a first-level directory within `clp/integration-tests/tests/data`,
+   i.e., dataset directories cannot be nested.
+
+## Dataset contents
+
+When adding or changing a dataset directory within `clp/integration-tests/tests/data`, the following
```
Bill-hbrhbr:
```suggestion
When adding or changing a dataset directory within `integration-tests/tests/data`, the following
```

integration-tests/tests/data/README.md
```diff
@@ -0,0 +1,78 @@
...
+rules must be observed:
+
+1. Each dataset directory must contain a first-level subdirectory that holds all log files for the
```
Bill-hbrhbr:
```suggestion
1. Each dataset directory must contain a top-level subdirectory specified by `logs_subdir`. This subdirectory should hold all log files for the
```

integration-tests/tests/data/README.md
```diff
@@ -0,0 +1,78 @@
...
+   dataset.
+2. The logs subdirectory may only contain log files. (These log files may be organized in any way
+   within the logs subdirectory.)
+3. Each dataset directory must contain a file called `metadata.json`.
+4. The content of `metadata.json` must conform to the following schema:
```
Bill-hbrhbr:
```suggestion
3. Each dataset directory must contain a file called `metadata.json`, which must comform to the following schema:
```

integration-tests/tests/data/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Dataset contents
+
+When adding or changing a dataset directory within `clp/integration-tests/tests/data`, the following
+rules must be observed:
+
+1. Each dataset directory must contain a first-level subdirectory that holds all log files for the
+   dataset.
+2. The logs subdirectory may only contain log files. (These log files may be organized in any way
+   within the logs subdirectory.)
```
Bill-hbrhbr: Not sure if this point is useful.

integration-tests/tests/data/README.md
```diff
@@ -0,0 +1,78 @@
+# Integration test data
+
+This directory (`clp/integration-tests/tests/data`) contains test datasets for use throughout the
+integration test system.
+
+## Directory structure
+
+When augmenting or changing the contents of `clp/integration-tests/tests/data`, the following rules
```
Bill-hbrhbr:
```suggestion
When augmenting or changing the contents of `integration-tests/tests/data`, the following rules
```

integration-tests/tests/data/README.md
```diff
@@ -0,0 +1,78 @@
...
+must be observed:
+
+1. Each dataset must be located in its own directory.
+2. Each dataset directory must be a first-level directory within `clp/integration-tests/tests/data`,
+   i.e., dataset directories cannot be nested.
```
Bill-hbrhbr:
```suggestion
2. Each dataset directory must be a direct child of `integration-tests/tests/data`; nested dataset directories are not allowed.
```

integration-tests/tests/data/README.md
Bill-hbrhbr: We should prepare this README to be moved into `docs/src/dev-docs/testing/integration-tests.md`.
Let's remove all the leading `clp/` prefixes since that is the convention inside `docs`.

integration-tests/tests/data/README.md
```diff
@@ -0,0 +1,78 @@
...
+   | Field | Description |
+   | --- | --- |
+   | `dataset_name` | The name of the dataset directory. |
+   | `unstructured` | `True` if logs are unstructured, else `False`. |
+   | `timestamp_key` | The authoritative timestamp key, or `null` if there is no such key. |
+   | `begin_ts` | The earliest timestamp present in the dataset (ms). |
+   | `end_ts` | The latest timestamp present in the dataset (ms). |
+   | `logs_subdir` | The name of the subdirectory containing logs. |
+   | `file_names` | A list of the files within `logs_subdir`. |
+   | `single_match_wildcard_query` | A wildcard query which, when searched, will match a single log message in the dataset. |
```
Bill-hbrhbr:
```suggestion
   | `single_match_wildcard_query` | A wildcard query that should match exactly one log message in the dataset. |
```

Bill-hbrhbr (changes requested):
1. Polish README
2. Use `*_dir` instead of `*_path` for directory properties.

integration-tests/tests/data/README.md
quinntaylormitchell: Fine with removing all the leading `clp/` prefixes

integration-tests/tests/data/README.md
quinntaylormitchell: I'm going to do

```suggestion
1. Each dataset directory must contain a subdirectory that holds all log files for the dataset.
```

instead; I don't like the idea of talking about `logs_subdir` before we introduce it.

integration-tests/tests/data/README.md
quinntaylormitchell: How about


```suggestion
   | `single_match_wildcard_query` | A wildcard query that matches exactly one log message in the dataset. |
```

integration-tests/tests/data/README.md
quinntaylormitchell: I'm going to do:

```suggestion
Tests should use dataset fixtures instead of reading the logs directly, because many verification
flows rely on dataset metadata.
```

because a. I realized it isn't a rule and so it shouldn't be in the list, and b. I like this wording better

integration-tests/tests/utils/classes.py
quinntaylormitchell: Agreed

integration-tests/tests/utils/classes.py
quinntaylormitchell: True

Bill-hbrhbr (approved):
onto @junhaoliao

integration-tests/tests/utils/classes.py
```diff
@@ -55,18 +56,70 @@ def test_log_dir(self) -> Path:
...
+            file_path_abs = self.logs_path / file_path
+            validate_file_exists(file_path_abs)
+
+    @property
+    def metadata_file_path(self) -> Path:
+        """:return: The absolute path to the file containing metadata for the dataset."""
+        return self.dataset_root_dir / "metadata.json"
+
+    @property
+    def logs_path(self) -> Path:
```
Bill-hbrhbr, line 123:
```suggestion
    def logs_dir(self) -> Path:
```
Still prefer using `_dir` suffix but let's wait for @junhaoliao

junhaoliao: Let's use `logs_path` - i agree it's more clear in representing an absolute path. we should probably document this convention somewhere in our developer guidelines though

junhaoliao (approved):

PR #2182, opened 2026-04-07
feat(integration-tests): Introduce `ExternalAction` and `CmdArgs` abstractions for subprocess execution.

quinntaylormitchell:
# Description

This PR introduces two data classes: `ExternalAction` and `CmdArgs`. They provide an abstraction for how the testing system interacts with external programs (CLP, Presto, docker, etc.).

NOTE: The old `log_subprocess_output_to_file` and `run_and_log_subprocess` are not removed in this PR, and they are still referenced in code. They will be removed when appropriate as the rest of the package testing code is updated and developed.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **Tests**
  * Integration test framework now supports strongly-typed, parameterized external actions with structured command arguments
  * External actions persist per-action execution output files with timestamped summaries for easier inspection
  * Failure reporting aggregates related action log paths into concise messages to speed debugging
  * Added a lightweight subprocess runner with command validation and reliable capture of stdout/stderr


integration-tests/tests/utils/subprocess_utils.py
Bill-hbrhbr, line 32: Is there a way to directly update this to use external action inside this PR? I think this is the only place where `log_subprocess_output_to_file` is called.

integration-tests/tests/utils/subprocess_utils.py
```diff
@@ -34,3 +36,40 @@ def run_and_log_subprocess(cmd: list[str]) -> subprocess.CompletedProcess[str]:
         pytest.fail(f"Subprocess '{Path(cmd[0]).name}' returned a non-zero exit code.")
 
     return proc
+
+
+def execute_external_action(external_action: IntegrationTestExternalAction) -> None:
```
Bill-hbrhbr: I think we can turn this function into an `ExternalAction` method.

integration-tests/tests/utils/subprocess_utils.py
Bill-hbrhbr: I think we can initialize external action with cmd list[str].

integration-tests/tests/utils/classes.py
```diff
@@ -70,3 +71,17 @@ class IntegrationTestDataset:
...
+    """Metadata for an external action executed during an integration test."""
+
+    #: Command to pass to `subprocess.run()`.
+    cmd: list[str]
+
+    #: The completed process returned from `subprocess.run()`.
+    completed_proc: subprocess.CompletedProcess[str] = field(init=False)
+
+    #: Path to the file where this action's subprocess output was logged.
+    log_file_path: Path = field(init=False)
```
Bill-hbrhbr, line 93: we should have API method for updating these fields, especially if we want all subprocess invocations to go through this class.

integration-tests/tests/utils/classes.py
```diff
@@ -70,3 +71,17 @@ class IntegrationTestDataset:
...
     #: A dictionary of metadata describing the dataset.
     metadata_dict: dict[str, Any]
+
+
+@dataclass
+class IntegrationTestExternalAction:
+    """Metadata for an external action executed during an integration test."""
+
+    #: Command to pass to `subprocess.run()`.
+    cmd: list[str]
```
Bill-hbrhbr, line 87: Feels like
```
class CommandArgs(BaseModel):
    arg1: str
    arg2: str

    def to_cmd(self) -> list[str]:
        """Docstring."""
        return [arg1, arg2]
```
Could be a part of this class.

integration-tests/tests/package_tests/classes.py
```diff
@@ -210,3 +217,11 @@ def get_running_config_from_shared_config_file(self) -> ClpConfig:
             pytest.fail(fail_msg)
 
         return running_config
+
+
+@dataclass
+class ClpPackageExternalAction(IntegrationTestExternalAction, Generic[T]):
```
Bill-hbrhbr: TBH I don't see how this can be potentially useful along with `ClpBinaryExternalAction`. Maybe they differ by how they invoke `to_cmd()`? But as I commented I believe `to_cmd()` should be an abstract function for `IntegrationTestExternalAction`, which could be overriden by derived classes if needed.

Also, did you add `Generic[T]` for mypy type checking?

That being said, we can always add these classes in follow-up PRs if you can show the clear intent of their existence. For now I think we can drop them.

Bill-hbrhbr (changes requested):
For this PR, my current review comments remain high-level:
- I think the best way to introduce `ExternalAction` is to make `IntegrationTestExternalAction` a class that covers all the potential operations it needs, instead of having them scattered around in various utils files. The final goal is to make every subprocess invocation go through this class anyway.
- We can also have a simple `*ExternalAction` class derived from the base one above, implementing the most basic `to_cmd()`, which is probably joining a list of strings. Don't see benefits of including `ClpBinaryExternalAction` or `ClpPackageExternalAction` now.
- I would like to see if it's possible to deprecate `log_subprocess_output_to_file()` immediately. By the looks of it I think it's doable.

Bill-hbrhbr (changes requested):
One more comment:
What the PR description currently describes is not reflected in the code. In particular, the `CommandArgs` concept and the `to_cmd()`-driven flow do not appear in this PR.
```
"""A module that executes some action on CLP."""

class CommandArgs(BaseModel):
    arg1: str
    arg2: str

    def to_cmd(self) -> list[str]:
        """Docstring."""
        return [arg1, arg2]

# ...

    args: CommandArgs= _construct_command_args(clp_package, dataset)
    action: ClpPackageExternalAction[CommandArgs] = ClpPackageExternalAction(
        cmd=args.to_cmd(), args=args
    )
    execute_external_action(action)
```

Can you make that `CommandArgs` concept part of the base `IntegrationTestExternalAction` class itself, instead of derived classes? 

Instead of calling `to_cmd()` at construction time, it would be cleaner to call it inside whatever method `IntegrationTestExternalAction` uses to execute the action.

Also, I realize command construction and subprocess execution may not always happen immediately one after another. But if we can mostly pass around the `CommandArgs` object rather than the full `IntegrationTestExternalAction`, we should follow RAII (Resource Acquisition Is Initialization) and invoke `subprocess.run()` from `IntegrationTestExternalAction.__post_init__()`.

That way, the completed `subprocess.proc` will be immediately available.

integration-tests/tests/package_tests/classes.py
quinntaylormitchell: I've made `*CmdArgs` two generic classes and called `to_cmd()` in `*ExternalAction.__post_init__()`. `to_cmd()` is not needed in `IntegrationTestExternalAction`; see offline convo for more information.

(`Generic[T]` was there for mypy type checking, you're correct.)

integration-tests/tests/utils/classes.py
quinntaylormitchell: I've added it to `*ExternalAction` derived classes.

integration-tests/tests/utils/classes.py
quinntaylormitchell: After restructuring, the only place that these two parameters are updated is within `IntegrationTestExternalAction.execute()`, so problem solved (if you want further changes, let me know.)

integration-tests/tests/utils/subprocess_utils.py
quinntaylormitchell: See offline convo re. doing this in a different PR.

integration-tests/tests/utils/subprocess_utils.py
quinntaylormitchell: Agreed.

quinntaylormitchell (commented):
Thanks for the review @Bill-hbrhbr ; a couple high-level changes made re. your high level comments:

1. `*ExternalAction` classes now have all of the methods they need to operate (`execute()`, `_log_action_summary_to_file()`).
2. `*CmdArgs` is now a generic model from which we can derive other command argument models. `to_cmd()` is called from `*ExternalAction.__post_init__()`.

See our offline convo for more info.

integration-tests/tests/utils/subprocess_utils.py
```diff
@@ -34,3 +35,35 @@ def run_and_log_subprocess(cmd: list[str]) -> subprocess.CompletedProcess[str]:
...
+        timeout=DEFAULT_CMD_TIMEOUT_SECONDS:
+        check=False:                            Error will be handled during verification.
+        text=True:                              Output should be str for analysis purposes.
+
+    :param cmd:
+    """
+    if not cmd:
+        pytest.fail("Cannot run subprocess: `cmd` list is empty.")
+
+    log_msg = f"Running '{Path(cmd[0]).name}' subprocess. Command: {cmd}"
```
Bill-hbrhbr: Extract `Path(cmd[0]).name` to something like `exe_name` or `bin_name` since it is reused a couple times.

integration-tests/tests/utils/subprocess_utils.py
```diff
@@ -34,3 +35,35 @@ def run_and_log_subprocess(cmd: list[str]) -> subprocess.CompletedProcess[str]:
         pytest.fail(f"Subprocess '{Path(cmd[0]).name}' returned a non-zero exit code.")
 
     return proc
+
+
+def run_subprocess(cmd: list[str]) -> subprocess.CompletedProcess[str]:
```
Bill-hbrhbr: Do you think this could also be a part of `ExternalAction`? We want all subprocess invocations to go through `ExternalAction` in the future.

integration-tests/tests/utils/classes.py
```diff
@@ -70,3 +79,77 @@ class IntegrationTestDataset:
...
+        stdout_content = completed_proc.stdout or "(empty)"
+        stderr_content = completed_proc.stderr or "(empty)"
+
+        if not stdout_content.endswith("\n"):
+            stdout_content += "\n"
+        if not stderr_content.endswith("\n"):
+            stderr_content += "\n"
+
+        sep = "-" * 32
+        summary_lines = [
```
Bill-hbrhbr, line 197: Maybe use triple-quoted f-string?
```
summary_lines = textwrap.dedent(f"""\
    SUBPROCESS RUN SUMMARY
    {sep}
...

    """)
```

integration-tests/tests/utils/classes.py
```diff
@@ -70,3 +79,77 @@ class IntegrationTestDataset:
...
+            "\n\n",
+            "captured stderr\n",
+            f"{sep}\n",
+            stderr_content,
+            "\n",
+        ]
+
+        self.log_file_path.parent.mkdir(parents=True, exist_ok=True)
+        with self.log_file_path.open("w", encoding="utf-8") as log_file:
+            log_file.writelines(summary_lines)
```
Bill-hbrhbr, line 217:
```suggestion
            log_file.write(summary_lines)
```
Update if you use single f-string over list of strings.

integration-tests/tests/utils/logging_utils.py
```diff
@@ -62,3 +65,18 @@ def log_subprocess_output_to_file(
         f" '{subprocess_output_file_path}'"
     )
     logger.info(log_msg)
+
+
+def format_action_failure_msg(reason: str, *actions: ExternalAction) -> tuple[bool, str]:
```
Bill-hbrhbr: Inside the docstring, please clarify that this function can associate a single failure with multiple external actions/ subprocess calls.
Also, since this function always returns `False` as the first tuple element. Is it necessary to include that value in the return type?

Bill-hbrhbr (changes requested):

integration-tests/tests/utils/logging_utils.py
quinntaylormitchell: Agreed on both points

integration-tests/tests/utils/subprocess_utils.py
quinntaylormitchell: Agreed

integration-tests/tests/utils/classes.py
quinntaylormitchell: I think using `dedent` here isn't ideal because `stdout_content` and `stderr_content` are potentially multi-line; afaik `dedent` works after f-string interpolation has already happened, so `dedent` will see the string as something like this:

```text
            ...
            captured stdout
            --------------------------------
            line one
line two
...
```

and because it strips the shortest common whitespace prefix (in this case the empty string `""`), the formatting will be messed up

Would you be fine if I just left it as a list?

integration-tests/tests/utils/classes.py
quinntaylormitchell: see above

integration-tests/tests/utils/subprocess_utils.py
quinntaylormitchell: Ah I didn't think of that; I think I had left it separate cause I thought it'd be nice to have the option of calling it independently, but you're right, putting it in `ExternalAction` will strengthen the Action layer. Will do

integration-tests/tests/utils/classes.py
Bill-hbrhbr: Agreed

Bill-hbrhbr (dismissed):
onto @junhaoliao

integration-tests/tests/utils/logging_utils.py
```diff
@@ -62,3 +65,18 @@ def log_subprocess_output_to_file(
...
+    Formats a failure message that indicates where to find the subprocess log(s) relevant to the
+    failure. This function associates a single failure with one or more external actions.
+
+    :param reason:
+    :param actions:
+    :return: The failure message.
+    """
+    action_log_paths: list[str] = []
+    for action in actions:
+        action_log_paths.append(str(action.log_file_path))
```
junhaoliao: just a more Pythonic shorthand. feel free to reject

```suggestion
    action_log_paths = [str(action.log_file_path) for action in actions]
```

integration-tests/tests/utils/classes.py
```diff
@@ -1,14 +1,22 @@
...
 import pytest
+from pydantic import BaseModel
 
-from tests.utils.utils import (
-    validate_dir_exists,
-)
+from tests.conftest import get_test_log_dir
+from tests.utils.utils import validate_dir_exists
+
+DEFAULT_CMD_TIMEOUT_SECONDS = 120.0
```
junhaoliao, line 16: this duplicates https://github.com/y-scope/clp/blob/32d2d4301f33e8ce1b475e76cf7a0bf09b403338/integration-tests/tests/utils/subprocess_utils.py#L11 ?

possible to let `integration-tests/tests/utils/subprocess_utils.py` import from `integration-tests/tests/utils/classes.py`?

integration-tests/tests/utils/classes.py
```diff
@@ -70,3 +78,102 @@ class IntegrationTestDataset:
...
+        """
+        Passes `self.cmd` to `subprocess.run()` with preset parameters:
+            capture_output=True:                    Output will be logged and analysed later.
+            timeout=DEFAULT_CMD_TIMEOUT_SECONDS:
+            check=False:                            Error will be handled during verification.
+            text=True:                              Output should be str for analysis purposes.
+        """
+        exe_name = Path(self.cmd[0]).name
+        log_msg = f"Running '{exe_name}' subprocess. Command: {self.cmd}"
+        logger.info(log_msg)
```
junhaoliao, line 164: seems there are a few such occurrences of f-string misuses in the logging statements. i'm fine deferring the fixes to another PR though

```suggestion
    logger.info("Running '%s' subprocess. Command: %s", Path(cmd[0]).name, cmd)
```

@Bill-hbrhbr  btw, is it possible to configure the linter to catch this kind of slip off from the `%`-style string checks?

junhaoliao (dismissed):
lgtm. posted some suggestions though i think some of the fixes can be deferred. for the title, how about:

```
feat(integration-tests): Introduce `ExternalAction` and `CmdArgs` abstractions for subprocess execution.
```

integration-tests/tests/utils/logging_utils.py
quinntaylormitchell: sure!

integration-tests/tests/utils/classes.py
quinntaylormitchell: Will do in another PR, I've made a note of it

junhaoliao (approved):
lgtm

PR #2184, opened 2026-04-07
feat(clp-package)!: Dispatch batches of search tasks in parallel to improve query scheduler throughput.

gibber9809:
# Description

This PR parallelizes task dispatch across processes in a process pool executor to reduce the dispatch bottleneck when running concurrent jobs. On startup, each process in the process pool creates a database connection pool with one connection and stores some config needed for most jobs in the global scope for later use during dispatch. During dispatch task groups are dispatched to celery in the dispatch processes, then rehydrated in the main process so that results can be retrieved from the main process.

Overall, on an old Xeon server setup, this improves overall per-batch scaling (both dispatching and retiring) from ~47.3ms/batch to ~26.5ms/batch given sufficient concurrent jobs for batches of tasks to be dispatched in parallel.

This change is marked "breaking" since it updates the package config.

# Checklist

  breaking change.

# Validation performed

* Benchmarked query performance in scenario with 50 parallel jobs running set of 600 queries over 65GiB of data, and measured improved scaling
* Tested that normal package operations (sbin search, webui search, api server search, cancelling, log viewing, etc.) appear to work as expected.

## Summary by CodeRabbit

* **New Features**
  * Exposed a configurable query-scheduler concurrency setting (default: 16) to tune parallel processing.

* **Chores**
  * Scheduler dispatch now runs work in parallel processes so jobs begin running sooner and throughput is improved.

* **Documentation**
  * Deployment guides, Helm chart and config templates updated to document and surface the new scheduler parallelism setting.


components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1013,37 +1091,43 @@ async def handle_jobs(
...
-            db_conn_pool,
-            clp_metadata_db_conn_params,
-            results_cache_uri,
-            stream_collection_name,
-            num_archives_to_search_per_sub_job,
-            max_datasets_per_query,
-            existing_datasets,
-            archive_retention_period,
+    with concurrent.futures.ProcessPoolExecutor(
+        max_workers=16,
```
gibber9809: Do we want to make this configurable?

LinZhihao-723: I think we should, in package config I assume.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -88,6 +91,27 @@
 
 reducer_connection_queue: asyncio.Queue | None = None
 
+# Globals for dispatch executor pool
+executor_db_conn_pool: ConnectionPoolWrapper | None = None
+executor_clp_metadata_db_conn_params: dict[str, any] | None = None
+executor_results_cache_uri: str | None = None
```
LinZhihao-723: 1. This file is already long and has many functions serving for different purposes.
2. The current naming of `_for_executor` is a bit confusing.

How about restructuring in the following way to improve the readability and maintainability of the per-executor-process data and methods:

```Python
class DispatchExecutor:
    # Globals for dispatch executor pool
    _db_conn_pool: ConnectionPoolWrapper | None = None
    _clp_metadata_db_conn_params: dict[str, any] | None = None
    _results_cache_uri: str | None = None

    @classmethod
    def initialize(
        cls,
        database_config: Database,
        clp_metadata_db_conn_params: dict[str, any],
        results_cache_uri: str,
    ) -> None:
        cls._clp_metadata_db_conn_params = clp_metadata_db_conn_params
        sql_adapter = SqlAdapter(database_config)
        cls._db_conn_pool = sql_adapter.create_connection_pool(
            logger=logger, pool_size=1, disable_localhost_socket_connection=True
        )
        cls._results_cache_uri = results_cache_uri

    @staticmethod
    def dispatch_job_and_update_db(
        job_config: dict[str, any], job_type: QueryJobType, job_id: str, archives: list[dict]
    ) -> tuple[str, int, str]:
        if not QueryJobType.SEARCH_OR_AGGREGATION == job_type:
            raise NotImplementedError(f"Unexpected job type: {job_type}")

        archive_ids = [a["archive_id"] for a in archives]
        with contextlib.closing(SearchJobExecutor._db_conn_pool.connect()) as db_conn:
            task_ids = insert_query_tasks_into_db(db_conn, job_id, archive_ids)

        celery_task_group = celery.group(
            search.s(
                job_id=job_id,
                archive_id=archives[i]["archive_id"],
                task_id=task_ids[i],
                job_config=job_config,
                dataset=archives[i].get("dataset"),
                clp_metadata_db_conn_params=SearchJobExecutor._clp_metadata_db_conn_params,
                results_cache_uri=SearchJobExecutor._results_cache_uri,
            )
            for i in range(len(archives))
        )
        group_result = celery_task_group.apply_async()
        group_result.save()
        return job_id, len(archives), group_result.id
```

Pool initialization:
```Python
with concurrent.futures.ProcessPoolExecutor(
    max_workers=16,
    initializer=SearchJobExecutor.initialize,
    initargs=(database_config, clp_metadata_db_conn_params, results_cache_uri),
) as process_pool:
```

Pool submission:
```Python
futures.append(
    process_pool.submit(
        SearchJobExecutor.dispatch_job_and_update_db,
        job.get_config().model_dump(),
        job.get_type(),
        job_id,
        archives_for_search,
    )
)
```

LinZhihao-723 (changes requested):
The functionality changes lgtm. Leaving a code structure comment for better maintainability.

LinZhihao-723 (commented):
The code changes lgtm.
In the package deployment code: why do we use `parallelism` in the helm chart, but `concurrency` in the package config? Shall we consist the names or it's a special keyword in helm?

gibber9809:
> The code changes lgtm. In the package deployment code: why do we use `parallelism` in the helm chart, but `concurrency` in the package config? Shall we consist the names or it's a special keyword in helm?

I think I just missed renaming in one place while I was deciding -- I'll update to make these consistent.

gibber9809:
@junhaoliao do you think the way I've set up the config for query scheduler concurrency makes sense, or do you think it should work more similarly to the worker concurrency?

LinZhihao-723 (dismissed):
I don't have any more concerns. Will leave the rest to Junhao for the package config checking.
Btw, I think this seems to be a breaking change since we update the package config.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -436,6 +436,7 @@ class QueryScheduler(BaseModel):
     max_datasets_per_query: PositiveInt | None = 10
     num_archives_to_search_per_sub_job: PositiveInt = 16
     logging_level: LoggingLevel = "INFO"
+    concurrency: PositiveInt = 16
```
junhaoliao: would renaming this as `dispatch_pool_size` help resolve any potential confusion with the Celery `--concurrency` parameter and the `workConcurrency` key in values.yaml?

junhaoliao (commented):
i agree the change is breaking unless we set the `concurrency` default count to 1 - now users upgrading without touching their config would suddenly get 16 extra processes they didn't have before, and they should be aware of such resource allocation considerations.

i proposed a change to the `concurrency` config field name to avoid confusions with the (celery) worker pod / node concurrency settings

components/clp-py-utils/clp_py_utils/clp_config.py
gibber9809: I wanted to call it something a bit more generic just because we might end up using the same pool of process to parallelize some parts of retiring tasks as well, but yeah we could rename to something like `scheduler_concurrency` to further distinguish from the worker concurrency.

gibber9809:
> i agree the change is breaking unless we set the `concurrency` default count to 1 - now users upgrading without touching their config would suddenly get 16 extra processes they didn't have before, and they should be aware of such resource allocation considerations.
> 
> i proposed a change to the `concurrency` config field name to avoid confusions with the (celery) worker pod / node concurrency settings

I'll rename it to distinguish it a bit more. Also agree that this could mess with current deployments, so I think that making the default lower makes sense -- do you think it's better to default to `1`, or just something low (e.g., `2`, or `4`) so that existing users are likely to get some benefit with low odds of resource issues on existing deployments.

junhaoliao:
> ... making the default lower makes sense -- do you think it's better to default to `1`, or just something low (e.g., `2`, or `4`) so that existing users are likely to get some benefit with low odds of resource issues on existing deployments.

sg. Lowering it to `4` makes sense

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -746,18 +804,37 @@ def handle_pending_query_jobs(
...
+                )
             )
+
+        for future in concurrent.futures.as_completed(futures):
+            job_id, num_archives_for_search, group_result_id = future.result()
+            job = active_jobs[job_id]
+            job.current_sub_job_async_task_result = celery.result.GroupResult.restore(
+                group_result_id, app=app
+            )
+            job.state = InternalJobState.RUNNING
```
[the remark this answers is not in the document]
junhaoliao: sorry if i missed any previous discussions - i suppose the `update-db <-> dispatch` execution order flip is intentional? may i ask what's the rationale behind?

that said, i don't think we would end up with any stuck jobs if the scheduler crashes before the jobs are marked as DONE, because `kill_hanging_jobs` is called on scheduler startup and kills all jobs with `status=RUNNING`.

That said, if the `futures.append()` or `process_pool.submit()` may fail as pointed out by the rabbit, and the scheduler doesn't restart, the jobs might be stuck indefinitely until the scheduler is restarted?

docs/src/user-docs/guides-docker-compose-deployment.md
```diff
@@ -85,6 +85,9 @@ To configure CLP for multi-host deployment, you'll need to:
    * When using local filesystem storage (i.e., not S3), set `logs_input.directory`,
      `archive_output.storage.directory`, and `stream_output.storage.directory` to directories on the
      shared filesystem.
+   * (Optional) set `scheduler_concurrency` under `query_scheduler` to control
+     the size of the process pool used by the query scheduler. Higher
+     concurrency can help improve performance for larger clusters.
```
junhaoliao:
```suggestion
     concurrency can help improve throughput when many query jobs can be
     run concurrently in larger clusters.
```

junhaoliao (dismissed):
lgtm

left a quick question and a minor doc improvement suggestion

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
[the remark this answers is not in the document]
gibber9809: Yeah, rationale is mostly that I wanted to move this state update for marking the job running out of the per-batch dispatch path (since it really should be once per job instead of once per batch), and everything should still work correctly if any part of this submission fails -- only difference with the previous implementation is that when the scheduler restarts after an exception during dispatch we'll hit `kill_hanging_jobs` instead of picking it up again from the `PENDING` state.

I also wanted the start time for the job to actually include the time it takes to dispatch the first batch of tasks. I'd be fine moving the `RUNNING` status update into the `as_completed` loop (just keeping track of the per-job start time in the per-pending-job loop) to keep the behaviour more similar to the previous implementation if that's prefferred though.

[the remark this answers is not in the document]
junhaoliao: i see. i agree it's better to include the dispatch time in the calculation. then it's fine to keep your changes as they are

junhaoliao (approved):
the pr title lgtm

PR #2185, opened 2026-04-07
feat(webui): Add export-all and per-row copy buttons to search results table (resolves #1980).

### LINKED ISSUE #1980: Support for searching/copying data from CLP UI
As a user i would want to copy the logging data from CLP UI once my search query is over.  Also, kind of small search box to search certain keywords in those events ( similar to text editor search experience) .

goynam:
# Description

Adds an Actions column to the search results virtual table with two capabilities:

* **Export all** — An icon-only download button in the column header exports the full result set as a JSONL file (`clp-search-results-<timestamp>.jsonl`). Disabled until the query completes and results are available.

  Table header showing the export-all download button in the Actions column

  Export success toast notification after clicking the export-all button

* **Copy event** — A per-row copy icon button copies the individual log event (formatted as a single JSONL line) to the clipboard. This avoids the memory concerns of a "copy all" approach by keeping each operation O(1).

  A data row showing the per-row copy button (copy icon) on the right side of the Actions column

  Hovering shows a tooltip:

  Copy button with "Copy this event" tooltip

The JSONL format uses ISO 8601 timestamps — intentionally chosen over the human-readable display format because ISO 8601 is more immediately recognizable to both humans and automated parsers, and is the de facto standard for structured log data interchange.

Overall layout:

Full search results table with the Actions column on the right

### Key design decisions

* **Column-based placement** — Export and copy buttons live inside the table as a new rightmost column, avoiding the need for a separate container above the table (which would require adjusting `tableHeight`).
* **Store side-channel** — The column header component (`ActionsHeader`) needs to trigger an export, but only the parent `SearchResultsVirtualTable` has access to the `searchResults` cursor data. The export callback is stored in a closure variable inside the zustand `create()` callback (not zustand state), so registering it doesn't trigger re-renders.
* **Deferred blob URL revocation** — Revoking the blob URL immediately after `anchor.click()` silently aborts downloads on Firefox. Revocation is deferred by 60 seconds.

# Checklist

# Validation performed

* All webui lints pass (`npm run lint:check --workspaces` with `--max-warnings 0`).
* TypeScript compiles cleanly (`tsc --noEmit`).
* Manual testing: searched with `*` query, verified export-all downloads a JSONL file with correct content, verified per-row copy copies a single JSONL line to clipboard, verified button enable/disable states.

## Release Notes

**New Features**
* Added ability to export all search results as a JSONL file with timestamped filename.
* Added copy-to-clipboard functionality for individual search results.
* Export button appears in the search results table header and is only enabled when results are available.
* User receives success or error feedback when exporting or copying results.


components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Native/SearchResultsVirtualTable/index.tsx
```diff
@@ -35,13 +49,57 @@ const SearchResultsVirtualTable = ({tableHeight}: SearchResultsVirtualTableProps
...
+                <div className={styles["copyButtonContainer"] || ""}>
+                    <Tooltip title={"Copy all results to clipboard"}>
+                        <Button
+                            icon={<CopyOutlined/>}
+                            size={"small"}
+                            onClick={() => {
+                                handleCopyAll().catch((err: unknown) => {
+                                    throw err;
+                                });
+                            }}
```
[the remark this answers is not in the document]
goynam: Changes:

Added navigator.clipboard availability check before calling writeText()
Wrapped writeText() in try/catch with error logging and user-facing error message
Replaced .catch(() => { throw err }) with void handleCopyAll() since errors are now handled internally

junhaoliao:
Thanks for the contribution - this seems useful addition

One concern I have is that the current approach merges all results into a single string before copying to the clipboard. If there are a large number of results, this could put significant pressure on browser memory. Also, Chromium's V8 engine has a string size limit (~536.8M chars. see https://chromium.googlesource.com/v8/v8/+/ef40f0628efc7e9c15b6f28edd69aa76480f064d/src/objects/string.h#531), which we might run into in extreme cases.

Would you be open to exploring a slightly different approach by splitting this into two features?

1. Instead of a "Copy all" button, perhaps we could add an "Export all" option that streams the log events into a Blob and triggers a download. This would avoid holding everything in memory at once.
2. For copying, we could add a per-log "Copy" button that appears on hover. This keeps the UX clean while still making it easy to copy individual entries.

Happy to hear your thoughts

goynam:
> Thanks for the contribution - this seems useful addition
> 
> One concern I have is that the current approach merges all results into a single string before copying to the clipboard. If there are a large number of results, this could put significant pressure on browser memory. Also, Chromium's V8 engine has a string size limit (~536.8M chars. see https://chromium.googlesource.com/v8/v8/+/ef40f0628efc7e9c15b6f28edd69aa76480f064d/src/objects/string.h#531), which we might run into in extreme cases.
> 
> Would you be open to exploring a slightly different approach by splitting this into two features?
> 
> 1. Instead of a "Copy all" button, perhaps we could add an "Export all" option that streams the log events into a Blob and triggers a download. This would avoid holding everything in memory at once.
> 2. For copying, we could add a per-log "Copy" button that appears on hover. This keeps the UX clean while still making it easy to copy individual entries.
> 
> Happy to hear your thoughts

thanks for the feedback. updated with export all button. Skipped per log "Copy" as think anyways if someone clicks on a log row, it selects the same and normal copy action would work.

junhaoliao (changes requested):
Following up on my earlier review - thanks for addressing the clipboard load handling! After thinking about this more, I'd like to suggest a direction that I think would work even better for the overall UX. Happy to hear your thoughts.

## Suggested approach

Instead of placing the export button above the table in a separate container, I think it would be cleaner to integrate the export functionality directly into the table as a new rightmost column. This avoids the `tableHeight` adjustment concern that CodeRabbit flagged and keeps the UI compact.

Here's what I'm thinking:

Here's how the overall layout looks:

Full search results table with the Actions column on the right

### Export-all button in the column header

A small icon-only button (`DownloadOutlined`, `size="small"`, `type="text"`) in the header of a new "Actions" column on the right side of the table. The button is disabled when the query hasn't completed or there are zero results.

Table header showing the export-all download button in the Actions column

Clicking it downloads a JSONL file and shows a success toast:

Export success toast notification after clicking the export-all button

### Per-row "Copy this event" button

Instead of "Copy all" (which has the memory concerns I raised earlier), a per-row copy button in the same Actions column cell. Clicking it copies that single log event as a JSONL line to the clipboard. This keeps each operation O(1) in memory.

A data row showing the per-row copy button (copy icon) on the right side of the Actions column

Hovering over the copy button shows a tooltip:

Copy button with "Copy this event" tooltip

### Export format: JSONL

Rather than plain text, export as JSONL (`{"timestamp": "...", "message": ...}\n...`). The timestamp is formatted as an ISO 8601 string — we intentionally use ISO 8601 rather than the human-readable format used in the table display column because it's more immediately recognizable to both humans and automated parsers, and it's the de facto standard for structured log data interchange. The `message` field is included as a parsed JSON object if it's valid JSON, otherwise as a raw string. The filename would be `clp-search-results-<timestamp>.jsonl`.

---

## Implementation details

### 1. New file: `components/webui/client/src/utils/download.ts`

Shared download utilities. Three exported functions:

```ts
import dayjs from "dayjs";


// Defer revocation by this amount because revoking the blob URL immediately after `click()`
// silently aborts downloads on some browsers (e.g., Firefox).
const REVOCATION_DELAY_MS = 60_000;

/**
 * Returns a filesystem-safe timestamp string suitable for export filenames.
 *
 * @return ISO-8601-like string with colons and dots replaced by dashes.
 */
const getExportFilenameTimestamp = (): string => (
    dayjs().format("YYYY-MM-DDTHH-mm-ss-SSS[Z]")
);

/**
 * Formats a numeric timestamp of an event as an ISO 8601 string.
 *
 * @param timestamp
 * @return ISO 8601 string.
 */
const getExportEventTimestamp = (timestamp: number): string => (
    dayjs(timestamp).toISOString()
);

/**
 * Triggers a plain-text file download in the browser by creating a temporary
 * Blob URL and clicking a hidden anchor element.
 *
 * @param lines Iterable of strings to write into the file.
 * @param filename Name for the downloaded file.
 */
const downloadTextFile = (lines: Iterable<string>, filename: string): void => {
    const blob = new Blob([...lines], {type: "text/plain"});
    const url = URL.createObjectURL(blob);
    const anchor = document.createElement("a");
    anchor.href = url;
    anchor.download = filename;
    anchor.click();
    setTimeout(() => {
        URL.revokeObjectURL(url);
    }, REVOCATION_DELAY_MS);
};

export {
    downloadTextFile,
    getExportEventTimestamp,
    getExportFilenameTimestamp,
};
```

- **`getExportFilenameTimestamp()`** — Produces a filesystem-safe timestamp for filenames (colons and dots replaced with dashes).
- **`getExportEventTimestamp(timestamp)`** — Formats a numeric event timestamp as ISO 8601. We intentionally use ISO 8601 rather than the human-readable display format because it's more recognizable to both humans and automated parsers, and it's the de facto standard for structured log data interchange.
- **`downloadTextFile(lines, filename)`** — Creates a Blob from the lines, creates a temporary object URL, clicks a hidden `<a>` to trigger the download. Revokes the URL after 60 seconds (revoking immediately after `click()` silently aborts downloads on Firefox).

### 2. New file: `components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Native/SearchResultsVirtualTable/ActionsHeader.tsx`

A React component that renders the export-all button inside the Actions column header. It has no props — instead, it reads from the zustand search store. This is separate from `typings.tsx` to keep the column definitions file focused on types and column config.

```tsx
import {DownloadOutlined} from "@ant-design/icons";
import {
    Button,
    Tooltip,
} from "antd";

import useSearchStore from "../../../../SearchState/index";
import {SEARCH_UI_STATE} from "../../../../SearchState/typings";


/**
 * Column header component with an export-all button.
 *
 * @return
 */
const ActionsHeader = () => {
    const onSearchResultsExport = useSearchStore((state) => state.onSearchResultsExport);
    const numResults = useSearchStore((state) => state.numSearchResultsTable);
    const searchUiState = useSearchStore((state) => state.searchUiState);

    return (
        <Tooltip title={"Export all results as JSONL"}>
            <Button
                disabled={SEARCH_UI_STATE.DONE !== searchUiState || 0 === numResults}
                icon={<DownloadOutlined/>}
                size={"small"}
                type={"text"}
                onClick={onSearchResultsExport}/>
        </Tooltip>
    );
};

export default ActionsHeader;
```

The button reads `numSearchResultsTable` and `searchUiState` from the store. It's disabled when the query hasn't completed (`searchUiState !== SEARCH_UI_STATE.DONE`) or there are zero results. On click it calls `onSearchResultsExport()`, which is the callback registered by the parent table component via the store side-channel (see step 3).

### 3. Modified: `components/webui/client/src/pages/SearchPage/SearchState/index.tsx`

Add two new members to the `SearchState` interface:

```ts
/**
 * Triggers a file download of the current search results. The
 * implementation is registered by the component that owns the cursor
 * subscription via setOnSearchResultsExport.
 */
onSearchResultsExport: () => void;
```

```ts
setOnSearchResultsExport: (fn: (() => void) | null) => void;
```

**Important**: the implementation is stored in a **local closure variable** inside the `create()` callback, not in zustand state. This means `setOnSearchResultsExport` does *not* trigger re-renders across all consumers. Change the store creation from the current flat style to one that has a closure:

```ts
const useSearchStore = create<SearchState>((set) => {
    let onSearchResultsExportImpl: (() => void) | null = null;

    return {
        ...SEARCH_STATE_DEFAULT,
        onSearchResultsExport: () => {
            onSearchResultsExportImpl?.();
        },
        setOnSearchResultsExport: (fn) => {
            onSearchResultsExportImpl = fn;
        },
        // ... rest of the existing actions unchanged ...
    };
});
```

Also make sure `numSearchResultsTable` is in `SEARCH_STATE_DEFAULT` (it's read by `ActionsHeader`).

### 4. Modified: `components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Native/SearchResultsVirtualTable/typings.tsx`

Several additions to this file. The existing `SearchResult` interface and column definitions stay, with the following additions and adjustments:

**Extract `formatTimestamp` as a named function** (was previously an inline arrow in the timestamp column's `render` prop):

```ts
/**
 * Formats a numeric timestamp as a UTC datetime string.
 *
 * @param timestamp
 * @return The formatted datetime string.
 */
const formatTimestamp = (timestamp: number): string => (
    dayjs.utc(timestamp).format(DATETIME_FORMAT_TEMPLATE)
);
```

**Add `formatResultAsJsonl`** — serializes a search result as a single JSON line (no trailing newline). The timestamp uses `getExportEventTimestamp()` to produce an ISO 8601 string. For the `message` field: tries `JSON.parse` — if it succeeds, includes the parsed object; if it throws, includes the raw string.

```ts
/**
 * Serializes a search result as a JSONL line. If the message field is valid
 * JSON it is included as a parsed object; otherwise it is kept as a string.
 *
 * @param result
 * @return A single JSON line (without trailing newline).
 */
const formatResultAsJsonl = (result: SearchResult): string => {
    let messageValue: unknown;
    try {
        messageValue = JSON.parse(result.message);
    } catch {
        messageValue = result.message;
    }

    return JSON.stringify({
        timestamp: getExportEventTimestamp(result.timestamp),
        message: messageValue,
    });
};
```

**Add `renderActionsCell`** — a column render function for the Actions column. This is intentionally a plain function (not a React component) to satisfy the `react/destructuring-assignment` lint rule.

```ts
/**
 * Cell renderer for the Actions column. Renders a per-row copy button.
 *
 * @param _
 * @param record
 * @return
 */
const renderActionsCell = (_: unknown, record: SearchResult) => {
    const handleCopyRow = () => {
        navigator.clipboard.writeText(formatResultAsJsonl(record))
            .catch((e: unknown) => {
                console.error("Failed to copy search result to clipboard", e);
            });
    };

    return (
        <Tooltip title={"Copy this event"}>
            <Button
                icon={<CopyOutlined/>}
                size={"small"}
                type={"text"}
                onClick={handleCopyRow}/>
        </Tooltip>
    );
};
```

**Add an Actions column** at the end of `searchResultsTableColumns`:

```ts
{
    align: "right",
    key: "actions",
    render: renderActionsCell,
    title: <ActionsHeader/>,
    width: 6,
},
```

**Adjust column widths** — Timestamp: 15→12, Message: 85→82 to make room for the 6-width Actions column.

**New exports** — add `formatResultAsJsonl` and `formatTimestamp` alongside the existing `searchResultsTableColumns`.

**New imports** — add `CopyOutlined` from `@ant-design/icons`, `Button` and `Tooltip` from `antd`, `getExportEventTimestamp` from `utils/download`, and `ActionsHeader` from `./ActionsHeader`.

### 5. Modified: `components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Native/SearchResultsVirtualTable/index.tsx`

**New imports**: add `useCallback` from react; `message` from antd; `downloadTextFile` and `getExportFilenameTimestamp` from `utils/download`; `formatResultAsJsonl` from `./typings`; destructure `setOnSearchResultsExport` from the store.

**Add `message.useMessage()`** — provides `messageApi` and `contextHolder` for success/error toast notifications.

**Add `handleExport` callback** wrapped in `useCallback` (depends on `messageApi` and `searchResults`):

```ts
/**
 * Exports all search results as a JSONL file download.
 *
 * NOTE: Results are exported in the original cursor order (i.e., timestamp descending),
 * which may differ from the user's current table sort.
 */
const handleExport = useCallback(() => {
    if (null === searchResults || 0 === searchResults.length) {
        return;
    }

    try {
        downloadTextFile(
            searchResults.map((r) => `${formatResultAsJsonl(r)}\n`),
            `clp-search-results-${getExportFilenameTimestamp()}.jsonl`
        );
        messageApi.success(`Exported ${searchResults.length} results`);
    } catch (e) {
        messageApi.error("Failed to export results");
        console.error(e);
    }
}, [
    messageApi,
    searchResults,
]);
```

**Register the handler** in a `useEffect` with cleanup:

```ts
useEffect(() => {
    setOnSearchResultsExport(handleExport);

    return () => {
        setOnSearchResultsExport(null);
    };
}, [
    handleExport,
    setOnSearchResultsExport,
]);
```

**Wrap the return value** in a fragment (`<>`) with `{contextHolder}` before the `VirtualTable`.

**Remove `x: "max-content"` from `scroll`** — AntD's virtual table requires `scroll.x` to be a number; the string value causes a console warning. Just use `scroll={{y: tableHeight}}`.

### What to remove from your PR

- `index.module.css` — the separate container for the export button is no longer needed
- The `formatResultLine` function — replaced by `formatResultAsJsonl`
- The `hasResults` conditional and the button container `div` above the table

---

## Notes

- **Why the store side-channel?** The column header (`ActionsHeader`) needs to trigger an export, but only the parent `SearchResultsVirtualTable` has access to the `searchResults` cursor data. Storing the callback in a closure variable inside the zustand `create()` callback (rather than as zustand state) lets us bridge the two components without triggering re-renders on every results update.
- **Blob URL revocation**: revoking immediately after `anchor.click()` silently aborts downloads on Firefox. Deferring by 60 seconds avoids that.
- **`scroll.x: "max-content"` removed**: AntD's virtual table requires `scroll.x` to be a number.

After we are done with the changes, let's also fix the lint issues as reported by the CI before we proceed to review this again.

If you're short on time and OK with the above direction, I'm also happy to push these changes directly to your branch — just let me know.

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Native/SearchResultsVirtualTable/typings.tsx
```diff
@@ -24,14 +33,68 @@ interface SearchResult {
...
+);
+
+/**
+ * Serializes a search result as a JSONL line. If the message field is valid
+ * JSON it is included as a parsed object; otherwise it is kept as a string.
+ *
+ * @param result
+ * @return A single JSON line (without trailing newline).
+ */
+const formatResultAsJsonl = (result: SearchResult): string => {
```
hoophalab: Why are these functions put in `typings.tsx`? Shouldn't them be in `utils.tsx`?

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Native/utils.ts
```diff
@@ -16,5 +17,28 @@ const getStreamId = (result: SearchResult): string => {
...
 };
 
+/**
+ * Formats a numeric timestamp of an event as an ISO 8601 string.
+ *
+ * @param timestamp
+ * @return ISO 8601 string.
+ */
+const getExportEventTimestamp = (timestamp: number): string => (
+    dayjs(timestamp).toISOString()
```
hoophalab:
```suggestion
    dayjs.utc(timestamp).toISOString()
```

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Native/SearchResultsVirtualTable/index.tsx
```diff
@@ -35,13 +52,53 @@ const SearchResultsVirtualTable = ({tableHeight}: SearchResultsVirtualTableProps
...
+    useEffect(() => {
+        setOnSearchResultsExport(handleExport);
+
+        return () => {
+            setOnSearchResultsExport(null);
+        };
+    }, [
+        handleExport,
+        setOnSearchResultsExport,
+    ]);
```
hoophalab, line 45: We usually just put data shared by two components in SearchState. Is it better to just put `searchResults` in `SearchState`, and make `onSearchResultsExport` statically do `handleExport`? The performance penalty shouldn't be too big since zustand only checks the array.

hoophalab (commented):
LGTM. Some comments

hoophalab (approved):
LGTM

PR #2193, opened 2026-04-09
feat(clp-s::log_converter): Add `max-log-event-size` argument; Make default max log event size 512 MiB to match `clp-s` (fixes #2176).

### LINKED ISSUE #2176: CLP-JSON package fails to ingest unstructured text log files with events larger than 64 MiB
### Bug

When using the CLP-JSON package (i.e., the `clp-s` storage engine, **not** the CLP-Text package) to ingest unstructured text log files via the `--unstructured` flag, the `log-converter` binary fails if any single log event exceeds 64 MiB. The error message is:

```
[error] Failed to convert input <path> to structured representation: generic - Numerical result out of range
```

### Root cause

The `log-converter` converts unstructured text logs into KV-pair IR streams before feeding them to `clp-s` for compression. Its internal buffer starts at 64 KiB and doubles as needed to accommodate large log events, but is capped at 64 MiB ([`LogConverter.hpp:41`](https://github.com/y-scope/clp/blob/3b4d13f3/components/core/src/clp_s/log_converter/LogConverter.hpp#L41)):

```cpp
static constexpr size_t cDefaultBufferSize{64ULL * 1024ULL};  // 64 KiB
static constexpr size_t cMaxBufferSize{64ULL * 1024ULL * 1024ULL};  // 64 MiB
```

When a log event exceeds this cap, [`grow_buffer_if_full()`](https://github.com/y-scope/clp/blob/3b4d13f3/components/core/src/clp_s/log_converter/LogConverter.cpp#L133-L146) returns `std::errc::result_out_of_range`, which propagates through [`refill_buffer()`](https://github.com/y-scope/clp/blob/3b4d13f3/components/core/src/clp_s/log_converter/LogConverter.cpp#L93-L96) → [`convert_file()`](https://github.com/y-scope/clp/blob/3b4d13f3/components/core/src/clp_s/log_converter/LogConverter.cpp#L58) → and is logged by [`log_converter.cpp:71-78`](https://github.com/y-scope/clp/blob/3b4d13f3/components/core/src/clp_s/log_converter/log_converter.cpp#L71-L78) as the error shown above.

Note: the KV-pair IR format also has additional size limits (logtype ~2 GiB, dictionary variable ~2 GiB, IR metadata 64 KiB) that would become relevant if `cMaxBufferSize` is raised. These are tracked separately in #2175 

### CLP version

3b4d13f3

### Environment

Any environment using the CLP-JSON package (`clp-s` storage engine) to ingest unstructured text log files with the `--unstructured` flag (via `sbin/compress.sh` or the WebUI compression page).

### Reproduction steps

1. Create a text log file containing a single log event larger than 64 MiB. For example, a single-line Hadoop/Hive log:
   ```
   $ wc -c hive-single-event-64MB.txt
   67108866 hive-single-event-64MB.txt
   $ wc -l hive-single-event-64MB.txt
   0 hive-single-event-64MB.txt
   ```
2. Ingest the file using the CLP-JSON package with `--unstructured` (e.g., via `sbin/compress.sh` or the WebUI with the unstructured option).
3. Observe the error in the compression worker stderr log:
   ```
   [error] Failed to convert input /mnt/logs/.../hive-single-event-64MB.txt to structured representation: generic - Numerical result out of range
   ```

gibber9809:
…MiB to match clp-s.

# Description

This PR brings `log-converter` in line with `clp-s` for maximum log event size. We add a command line option to `log-converter` to make it configurable, and we set the default to 512MiB in order to match `clp-s`.

It may be worth adding a per-compression-job config option to allow users to raise this limit for datasets they know will have large log lines, but that should be part of a separate PR.

# Checklist

  breaking change.

# Validation performed

* Generated a one line and two line log file that contain log records larger than 64MiB and observed that `log-converter` now accepts these records by default.
* Observed that setting `--max-log-event-size` to a value lower than 64MiB results in the two generated files failing conversion, as expected.

## Summary by CodeRabbit

* **New Features**
  * Added a --max-log-event-size option (default 512 MiB) to configure the maximum size of individual log events; non‑positive values are rejected.

* **Behavior Change**
  * Log conversion and buffer allocation now respect the configured maximum event size, limiting growth and allocation accordingly.

### DIFF
--- a/components/core/src/clp_s/log_converter/CommandLineArguments.cpp
+++ b/components/core/src/clp_s/log_converter/CommandLineArguments.cpp
@@ -137,6 +137,11 @@ auto CommandLineArguments::parse_arguments(int argc, char const** argv)
                 "Type of authentication required for network requests (s3 | none). Authentication"
                 " with s3 requires the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment"
                 " variables, and optionally the AWS_SESSION_TOKEN environment variable."
+        )(
+                    "max-log-event-size",
+                    po::value<size_t>(&m_max_log_event_size)->value_name("LOG_EVENT_SIZE")->
+                        default_value(m_max_log_event_size),
+                    "Maximum allowed size (B) for a single log event before conversion fails."
         );
         // clang-format on
 
@@ -188,6 +193,10 @@ auto CommandLineArguments::parse_arguments(int argc, char const** argv)
         }
 
         validate_network_auth(auth, m_network_auth);
+
+        if (m_max_log_event_size <= 0) {
+            throw std::invalid_argument("Max event size must be greater than zero.");
+        }
     } catch (std::exception& e) {
         SPDLOG_ERROR("{}", e.what());
         print_basic_usage();
--- a/components/core/src/clp_s/log_converter/CommandLineArguments.cpp
+++ b/components/core/src/clp_s/log_converter/CommandLineArguments.cpp
@@ -138,10 +138,11 @@ auto CommandLineArguments::parse_arguments(int argc, char const** argv)
                 " with s3 requires the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment"
                 " variables, and optionally the AWS_SESSION_TOKEN environment variable."
         )(
-                    "max-log-event-size",
-                    po::value<size_t>(&m_max_log_event_size)->value_name("LOG_EVENT_SIZE")->
-                        default_value(m_max_log_event_size),
-                    "Maximum allowed size (B) for a single log event before conversion fails."
+                "max-log-event-size",
+                po::value<size_t>(&m_max_log_event_size)
+                    ->value_name("LOG_EVENT_SIZE")
+                    ->default_value(m_max_log_event_size),
+                "Maximum allowed size (B) for a single log event before conversion fails."
         );
         // clang-format on
 
--- a/components/core/src/clp_s/log_converter/LogConverter.hpp
+++ b/components/core/src/clp_s/log_converter/LogConverter.hpp
@@ -17,7 +17,9 @@ namespace clp_s::log_converter {
 class LogConverter {
 public:
     // Constructors
-    LogConverter() : m_buffer(cDefaultBufferSize) {}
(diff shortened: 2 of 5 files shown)


components/core/src/clp_s/log_converter/CommandLineArguments.cpp
```diff
@@ -137,6 +137,11 @@ auto CommandLineArguments::parse_arguments(int argc, char const** argv)
                 "Type of authentication required for network requests (s3 | none). Authentication"
                 " with s3 requires the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment"
                 " variables, and optionally the AWS_SESSION_TOKEN environment variable."
+        )(
+                    "max-log-event-size",
+                    po::value<size_t>(&m_max_log_event_size)->value_name("LOG_EVENT_SIZE")->
+                        default_value(m_max_log_event_size),
+                    "Maximum allowed size (B) for a single log event before conversion fails."
```
LinZhihao-723:
```suggestion
                "max-log-event-size",
                po::value<size_t>(&m_max_log_event_size)
                    ->value_name("LOG_EVENT_SIZE")
                    ->default_value(m_max_log_event_size),
                "Maximum allowed size (B) for a single log event before conversion fails."
```
To align with other parameters.

components/core/src/clp_s/log_converter/LogConverter.hpp
```diff
@@ -17,7 +17,9 @@ namespace clp_s::log_converter {
 class LogConverter {
 public:
     // Constructors
-    LogConverter() : m_buffer(cDefaultBufferSize) {}
+    explicit LogConverter(size_t max_buffer_size)
+            : m_buffer(cDefaultBufferSize),
+              m_max_buffer_size{max_buffer_size} {}
```
LinZhihao-723, line 22: Shall we check that the max buffer size is at least `cDefaultBufferSize`?

LinZhihao-723 (changes requested):

LinZhihao-723 (approved):
For the PR title, how about:
```
feat(clp-s::log_converter): Add `max-log-event-size` argument; Make default max log event size 512 MiB to match `clp-s` (fixes #2176).
```
I think this PR is adding a feature, not strictly fixing an issue...

PR #2197, opened 2026-04-11
feat(api-server): Add GET `/usage/compression` endpoint.

jackluo923:
# Description

Adds a `GET /usage/compression` endpoint to the api-server that returns per-job compression usage data (sizes, durations, task stats) within a configurable epoch-millisecond time window. Intended for reporting, dashboarding, and chargeback use cases.

**Query parameters:**

- `begin_timestamp` (required) — start of usage window, epoch ms inclusive
- `end_timestamp` (required) — end of usage window, epoch ms inclusive
- `job_status` (optional) — comma-separated, case-insensitive list of statuses (`succeeded`, `FAILED`, etc.). Defaults to all terminal states (`SUCCEEDED,FAILED,KILLED`).
- `limit` (optional) — maximum number of jobs to return. Defaults to 1000.

**Key changes:**

- `CompressionUsageParams` and validation live in `client.rs` (not the HTTP layer) so any future caller of `Client` gets the same validated interface
- Job status parsing uses `strum::EnumString` on `CompressionJobStatus` (case-insensitive, `Pending` excluded via `#[strum(disabled)]`)
- `get_compression_usage` accepts `&[CompressionJobStatus]`; empty slice means "no status filter" (IN clause omitted)
- `sqlx::FromRow` derive on `CompressionUsage` eliminates manual `try_get` mapping; `query_as` replaces manual row iteration
- Epoch-millisecond timestamps (not ISO strings); `f64` with `ROUND()` in SQL instead of `BigDecimal`
- Composite index on `compression_jobs(start_time, status)` for efficient time-range queries
- 15 unit + route tests covering validation, error paths, and edge cases

# Checklist

# Validation performed

- `task lint:check-rust` (fmt + clippy) passes
- `task codegen:openapi` — generated OpenAPI JSON committed
- `cargo test -p api-server` — 15/15 tests pass

## Summary by CodeRabbit

* **New Features**
  * Added GET /usage/compression endpoint returning compression usage over a date range with job-status filtering (comma-separated, case-insensitive), defaults, limits and OpenAPI docs.

* **Bug Fixes**
  * Invalid query parameters now yield 400 Bad Request with clearer validation (timestamps, statuses, limits); client invalid-inputs mapped to 400.

* **Chores**
  * Backfilled composite DB index to speed compression queries; updated dependencies to support case-insensitive status parsing and tooling.

* **Tests**
  * Added unit and handler tests covering validation and endpoint behaviour.


components/api-server/Cargo.toml
hoophalab, line 1: We need to run `task codegen:openapi` to re-generate https://petstore.swagger.io/?url=https://docs.yscope.com/clp/main/_static/generated/api-server-openapi.json for new endpoints;
run `task lint:fix-rust` to fix lint

components/api-server/src/client.rs
```diff
@@ -523,6 +583,89 @@ impl Client {
...
+                j.uncompressed_size, \
+                j.compressed_size, \
+                j.num_tasks, \
+                CAST(SUM(t.duration) AS DECIMAL(20, 3)) AS tasks_duration \
+             FROM compression_jobs j \
+             JOIN compression_tasks t ON t.job_id = j.id \
+             WHERE j.start_time >= FROM_UNIXTIME(? / 1000.0) \
+               AND j.start_time <= FROM_UNIXTIME(? / 1000.0){job_status_clause} \
+             GROUP BY j.id, j.status, j.creation_time, j.start_time, j.duration, \
+                      j.uncompressed_size, j.compressed_size, j.num_tasks \
```
hoophalab: 1. We've encountered an issue where WebUI runs out of memory when dumping query jobs in the last 30 days. Not sure whether we should just **aggregate all durations + compressied_size into a single value**, or use a LIMIT.
2. Why do we need to aggregate on task duration? Job duration isn't enough/accurate?
3. Why do we format dates? Shouldn't this API be machine oriented, and returning a datetime string would require the requester to re-parse the datetime?
4. Why do we cast duration from float to decimal?

components/api-server/src/client.rs
```diff
@@ -523,6 +583,89 @@ impl Client {
...
+
+        let mut query = sqlx::query(&sql).bind(begin_timestamp).bind(end_timestamp);
+        for &status in job_statuses {
+            query = query.bind(status);
+        }
+        let rows = query.fetch_all(&self.sql_pool).await?;
+
+        rows.iter()
+            .map(|r| {
+                Ok(CompressionUsageRow {
```
hoophalab: Instead of using `try_get` to map `MySqlQueryResult` to `CompressionUsageRow`, it might be easier to derive `sqlx::FromRow` in `CompressionUsageRow`, and use `sqlx::query_as::<_, CompressionUsageRow>("SELECT ...")` directly.

components/api-server/src/client.rs
```diff
@@ -18,6 +19,65 @@ use utoipa::ToSchema;
...
+/// `components/job-orchestration/job_orchestration/scheduler/constants.py`.
+///
+/// Note: `PENDING` (0) is intentionally omitted because PENDING jobs have no
+/// `start_time` and are therefore excluded by the time-range WHERE clause.
+pub const JOB_STATUS_LOOKUP: &[(&str, i32)] = &[
+    ("RUNNING", 1),
+    ("SUCCEEDED", 2),
+    ("FAILED", 3),
+    ("KILLED", 4),
+];
```
hoophalab: 1. It might be easier to use `num_enum` to replace `resolve_job_status`. This would also simplify validation greatly
2. We should be able to reuse `CompressionJobStatus` in `clp-rust-utils/src/job_config/compression.rs`

components/api-server/src/routes.rs
```diff
@@ -243,6 +249,84 @@ async fn get_timestamp_column_names(
...
+    begin_timestamp: i64,
+    /// End of usage window (epoch milliseconds, inclusive).
+    end_timestamp: i64,
+    /// Job statuses to include as a comma-separated list (e.g.
+    /// `job_status=SUCCEEDED,FAILED`). Recognized values:
+    /// `RUNNING`, `SUCCEEDED`, `FAILED`, `KILLED`.
+    /// Defaults to `SUCCEEDED,FAILED,KILLED` (all terminal states).
+    #[serde(default)]
+    job_status: Option<String>,
+}
```
hoophalab: 1. Is using url parameters a hard requirement? Writing and interpreting url parameters is very tedious. Can we just POST a json payload to keep consistent with other endpoints?
2. It might be best to move this to `Client`. This way, we validates inputs as well if we need to do some high-level operations on the package using the `Client` in the future.

components/api-server/src/routes.rs
```diff
@@ -243,6 +249,84 @@ async fn get_timestamp_column_names(
...
+
+/// Validates compression usage parameters and resolves the requested job
+/// statuses into their integer codes.
+fn validate_compression_usage_params(
```
hoophalab: 1. The conventional way to validate a serde struct is to create an `UncheckedCompressionUsageQuery` and `impl TryFrom<UncheckedCompressionUsageQuery> for CompressionUsageQuery`, but this will also show `UncheckedCompressionUsageQuery` as the schema name in the openapi docs https://petstore.swagger.io/?url=https://docs.yscope.com/clp/main/_static/generated/api-server-openapi.json#/default/health.

How about move this validate function as a method of `CompressionUsageParams`, and add a comment to explain why
```
impl CompressionUsageParams {
    // explain why not having an UncheckedCompressionUsage
    pub fn validate() -> Result<(), ClientError> {
         ....
    }
}
```

2. If you adopt `CompressionJobStatus`, then the validation of job status can be removed / simplified to only checking `empty`.

components/api-server/src/routes.rs
```diff
@@ -243,6 +249,84 @@ async fn get_timestamp_column_names(
...
+        job_statuses,
+    );
+    client
+        .get_compression_usage(params.begin_timestamp, params.end_timestamp, &job_statuses)
+        .await
+        .map(Json)
+        .map_err(|err| {
+            tracing::error!("Failed to fetch compression usage: {:?}", err);
+            HandlerError::from(err)
+        })
```
hoophalab: up to you, this might look simpler

```suggestion
    Ok(Json(
        client
            .get_compression_usage(params.begin_timestamp, params.end_timestamp, &job_statuses)
            .await
            .inspect_err(|err| {
                tracing::error!("Failed to fetch compression usage: {:?}", err);
            })?
    ))
```

components/api-server/src/client.rs
```diff
@@ -18,6 +19,65 @@ use utoipa::ToSchema;
...
+pub fn resolve_job_status(name: &str) -> Option<i32> {
+    JOB_STATUS_LOOKUP
+        .iter()
+        .find(|(n, _)| *n == name)
+        .map(|(_, code)| *code)
+}
+
+/// A single row returned by the compression usage query (one row per job).
+#[derive(Serialize, ToSchema)]
+pub struct CompressionUsageRow {
```
hoophalab:
```suggestion
pub struct CompressionUsage {
```
Technically, this isn't a row in a physical table, it might be best to just call it `CompressionUsage`

We also need to update the docs if you accept this comment.

components/api-server/src/client.rs
```diff
@@ -523,6 +583,89 @@ impl Client {
...
+                DATE_FORMAT(j.start_time, '%Y-%m-%dT%H:%i:%s') AS start_time, \
+                CAST(j.duration AS DECIMAL(20, 3)) AS duration, \
+                j.uncompressed_size, \
+                j.compressed_size, \
+                j.num_tasks, \
+                CAST(SUM(t.duration) AS DECIMAL(20, 3)) AS tasks_duration \
+             FROM compression_jobs j \
+             JOIN compression_tasks t ON t.job_id = j.id \
+             WHERE j.start_time >= FROM_UNIXTIME(? / 1000.0) \
+               AND j.start_time <= FROM_UNIXTIME(? / 1000.0){job_status_clause} \
```
hoophalab:
```suggestion
             WHERE j.start_time >= FROM_UNIXTIME(? / 1000.0) \
               AND j.start_time <= FROM_UNIXTIME(? / 1000.0) \
               AND j.status IN ({placeholders} \
```
why not just inline this AND clause?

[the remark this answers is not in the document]
hoophalab: is this concern valid?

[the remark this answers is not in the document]
jackluo923: the concern is invalid. For resource consumption, if a job is not scheduled, it did not consume resource nor did it compress any files.

components/api-server/src/routes.rs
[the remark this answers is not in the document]
jackluo923: Not a concern that we want to address at the moment.

components/api-server/src/client.rs
jackluo923: 1. I will add a default limit w/ over-ridable option. If we need aggregated results, then we can have a separate endpoint for it.

2. job duration != sum(task durations). If job runs with parallelism of 10 and scheduling is free, then the job duration is 1/10 of the sum of task duration. When computing charge back, we need the time spent on each task, which corresponds to the actual compute consumed.

3. I will switch it back to epoch
4. fixed

components/api-server/src/routes.rs
jackluo923: 1. Per the HTTP spec (RFC 9110), GET is the correct method for a read-only, idempotent query with no side effects. This property also allows the query itself to be cacheable by various intermediaries (proxies, CDNs, browsers).  This point is fairly strong motivation to use GET request. That said, if the reviewer prefers POST + JSON for ergonomics and consistency with the /query endpoint rather than aligning with common REST API semantics, then I am happy to switch.

2. Sure.

components/api-server/src/client.rs
jackluo923: nice suggestion, switch to sqlx::FromRow

components/api-server/src/client.rs
jackluo923: used strum crate to replace resolve_job_status and also reuse CompressionJobStatus

components/api-server/src/routes.rs
jackluo923: Already adopted CompressionJobStatus

components/api-server/src/client.rs
jackluo923: done

components/api-server/src/client.rs
```diff
@@ -523,6 +632,71 @@ impl Client {
...
+        );
+
+        let mut query = sqlx::query_as::<_, CompressionUsage>(&sql)
+            .bind(begin_timestamp)
+            .bind(end_timestamp);
+        for &status in job_statuses {
+            query = query.bind(i32::from(status));
+        }
+        query = query.bind(limit);
+        query.fetch_all(&self.sql_pool).await.map_err(Into::into)
```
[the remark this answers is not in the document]
jackluo923: fixed by removing the filter if no status is specified

components/api-server/src/routes.rs
```diff
@@ -280,3 +330,230 @@ impl IntoResponse for HandlerError {
...
+            .await
+            .unwrap();
+        assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+        let body = get_body(response).await;
+        assert!(
+            body.contains("at least one valid status"),
+            "expected error about empty job_status, got: {body}"
+        );
+    }
+}
```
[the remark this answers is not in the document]
jackluo923: done

tools/yscope-dev-utils
```diff
@@ -1 +1 @@
-Subproject commit 38bf51effc500a528c37052345c141a65ea88447
+Subproject commit 6ec46856cfda8cbfb65d9386affbe3bc429186cb
```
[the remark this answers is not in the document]
jackluo923: no longer relevant

components/api-server/Cargo.toml
jackluo923: done

components/api-server/src/client.rs
```diff
@@ -14,10 +15,134 @@ use futures::{Stream, StreamExt};
...
+    /// `OpenAPI` query-parameter schema. Introducing an `Unchecked` wrapper would
+    /// leak an internal implementation detail into the generated API docs.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ClientError::InvalidInput`] if:
+    /// - `begin_timestamp > end_timestamp`
+    /// - `limit <= 0`
+    /// - `job_status` contains unrecognized or empty values
+    pub fn validate(&self) -> Result<ValidatedCompressionUsageParams, ClientError> {
```
hoophalab:
```suggestion
impl TryFrom<CompressionUsageParams> for ValidatedCompressionUsageParams {
    type Error = ClientError;
    /// Validates the parameters and resolves the requested job statuses into
    /// [`CompressionJobStatus`] variants.
    ///
    /// We use a `validate()` method rather than the conventional
    /// `TryFrom<UncheckedX>` pattern because this struct also serves as the
    /// `OpenAPI` query-parameter schema. Introducing an `Unchecked` wrapper would
    /// leak an internal implementation detail into the generated API docs.
    ///
    /// # Errors
    ///
    /// Returns [`ClientError::InvalidInput`] if:
    /// - `begin_timestamp > end_timestamp`
    /// - `limit <= 0`
    /// - `job_status` contains unrecognized or empty values
    pub fn try_from(value: CompressionUsageParams) -> Result<ValidatedCompressionUsageParams, Self::Error> {
```
AI might misunderstand my comments. If we already create a `ValidatedCompressionUsageParams`, then instead of impl a validate method, it might be best to `impl TryFrom<ValidatedCompressionUsageParams> for CompressionUsageParams { ... }`.

components/clp-py-utils/clp_py_utils/initialize-orchestration-db.py
```diff
@@ -78,11 +78,27 @@ def main(argv):
...
+                scheduling_db_cursor.execute(
+                    f"""
+                    ALTER TABLE `{COMPRESSION_JOBS_TABLE_NAME}`
+                    ADD INDEX `JOB_START_TIME_STATUS`
+                    (`start_time`, `status`) USING BTREE
+                    """
+                )
+            except Exception as err:
+                if "Duplicate key name" not in str(err):
+                    raise
```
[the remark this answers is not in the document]
hoophalab: this comment is probably valid?

components/api-server/src/client.rs
```diff
@@ -523,6 +648,86 @@ impl Client {
...
+            // ONLY_FULL_GROUP_BY (MySQL 5.7.5+ default).
+            //
+            // Task durations are summed regardless of individual task status so
+            // that failed jobs still report the compute resources they consumed.
+            //
+            // INNER JOIN filters out jobs with zero tasks (e.g. killed before
+            // task creation) since they consumed no compute resources.
+            "SELECT \
+              j.id, \
+              ANY_VALUE(j.status) AS status, \
```
hoophalab: Got this error in log. Not sure whether I misconfigured anything.
```
{"timestamp":"2026-04-12T13:37:11.319164Z","level":"ERROR","fields":{"message":"Failed to fetch compression usage: Sql(Database(MySqlDatabaseError { code: Some(\"42000\"), number: 1305, message: \"FUNCTION clp-db.ANY_VALUE does not exist\" }))"},"filename":"components/api-server/src/routes.rs","line_number":284}
```

We could just remove ANY_VALUE if there is no serious compatibility concerns with other databases.

components/clp-rust-utils/src/job_config/compression.rs
```diff
@@ -1,12 +1,26 @@
...
+    Eq,
+    IntoPrimitive,
+    PartialEq,
+    Serialize,
+    TryFromPrimitive,
+)]
 #[repr(i32)]
+#[strum(ascii_case_insensitive)]
 pub enum CompressionJobStatus {
+    #[strum(disabled)]
```
hoophalab: Why disable this? Querying pending jobs is still valid.

components/api-server/src/client.rs
```diff
@@ -585,3 +790,49 @@ where
         poll
     }
 }
+
+#[cfg(test)]
+mod tests {
```
hoophalab, line 842: Those tests have become meaningless because we've switched to strum

components/api-server/src/client.rs
```diff
@@ -523,6 +648,86 @@ impl Client {
...
+            // that failed jobs still report the compute resources they consumed.
+            //
+            // INNER JOIN filters out jobs with zero tasks (e.g. killed before
+            // task creation) since they consumed no compute resources.
+            "SELECT \
+              j.id, \
+              ANY_VALUE(j.status) AS status, \
+              ANY_VALUE(CAST(UNIX_TIMESTAMP(j.creation_time) * 1000 AS SIGNED)) AS creation_time, \
+              ANY_VALUE(CAST(UNIX_TIMESTAMP(j.start_time) * 1000 AS SIGNED)) AS start_time, \
+              ANY_VALUE(ROUND(j.duration, 3)) AS duration, \
```
hoophalab: 1. Can we remove `ROUND` and just return the full precision float value?
2. We should be able to bind a `chrono::Datetime<?>` directly and avoid `FROM_UNIXTIME(? / 1000.0)`.  We can modify types in `struct ValidatedCompressionUsageParams` and `struct CompressionUsageParams` using `#[serde(with = "chrono::serde::ts_milliseconds")]`.
3. Since we are already using query_as, can we change the time fields in `struct CompressionUsage` to type chrono, and setup `#[serde(with = "chrono::serde::ts_milliseconds")]`

hoophalab (commented):
LGTM in general. Validated that the endpoint works correctly after removing `ANY_VALUE` in group by.

components/api-server/src/client.rs
jackluo923: Will fix it manually. The doc is stale and callsite need tob e updated.

components/api-server/src/client.rs
jackluo923: switched to MAX() which is equivalent to ANY_VALUE() on dependent string and works across MYSQL and MariaDB

components/clp-rust-utils/src/job_config/compression.rs
jackluo923: you are right, when it is disabled, it not only disabled the default (where it should not be used). If user really want to fetch the resource usage for pending tasks, they should be able to.

components/api-server/src/client.rs
jackluo923: you are right, removed.

components/api-server/src/client.rs
jackluo923: done

components/api-server/src/client.rs
```diff
@@ -14,10 +15,134 @@ use futures::{Stream, StreamExt};
...
+                    })
+                })
+                .collect::<Result<Vec<_>, _>>()?,
+            None => DEFAULT_JOB_STATUSES.to_vec(),
+        };
+        if job_statuses.is_empty() {
+            return Err(ClientError::InvalidInput(
+                "job_status must contain at least one valid status".to_owned(),
+            ));
+        }
```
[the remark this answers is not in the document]
jackluo923: removed dead code

components/clp-py-utils/clp_py_utils/initialize-orchestration-db.py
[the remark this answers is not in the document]
jackluo923: switched to error code instead.

jackluo923:
> **Actionable comments posted: 2**
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against the current code and only fix it if needed.
> 
> Inline comments:
> In `@components/api-server/src/routes.rs`:
> - Around line 271-298: The parsing currently allows duplicate statuses which
> leads to redundant SQL placeholders; update validate_compression_usage_params to
> deduplicate the resolved statuses before returning (e.g., after mapping with
> resolve_job_status in validate_compression_usage_params, filter out repeats
> while preserving order by tracking seen statuses in a local HashSet or Vec),
> ensuring DEFAULT_JOB_STATUSES remains the fallback and that the existing
> empty-check and error paths (and references to resolve_job_status and
> DEFAULT_JOB_STATUSES) are preserved.
> 
> In `@components/clp-py-utils/clp_py_utils/initialize-orchestration-db.py`:
> - Around line 81-82: The composite index `JOB_START_TIME_STATUS` on (start_time,
> status) is correct for get_compression_usage's query pattern but CREATE TABLE IF
> NOT EXISTS only adds it for new DBs; add a migration step in
> initialize-orchestration-db.py to ensure existing deployments get the index:
> detect if the table exists and, if so, run an ALTER TABLE to add INDEX
> `JOB_START_TIME_STATUS` (`start_time`, `status`) USING BTREE (guarded to avoid
> duplicate-creation errors), or add an explicit migration script that executes
> that ALTER TABLE; reference the table creation block and the index name
> `JOB_START_TIME_STATUS` when implementing this change.
> ```
> 
> 🪄 Autofix (Beta)
> Fix all unresolved CodeRabbit comments on this PR:
> 
> * [ ]   Push a commit to this branch (recommended)
> * [ ]   Create a new PR with the fixes
> 
> ℹ️ Review info
> ⚙️ Run configuration
> **Configuration used**: Organization UI
> 
> **Review profile**: ASSERTIVE
> 
> **Plan**: Pro
> 
> **Run ID**: `f41ac76a-326c-4e97-af35-71e7afa2beef`
> 
> 📥 Commits
> Reviewing files that changed from the base of the PR and between [9ae078e](https://github.com/y-scope/clp/commit/9ae078ee1a90f948fd5c8af334d551cbb099b43d) and [d5a0f8f](https://github.com/y-scope/clp/commit/d5a0f8fc3e189b2f14e09c467bb2d4befe533a59).
> 
> ⛔ Files ignored due to path filters (1)
> * `Cargo.lock` is excluded by `!**/*.lock`
> 
> 📒 Files selected for processing (4)

raise error when duplicates are detected. migration is already migrated..

hoophalab (dismissed):
lgtm

hoophalab (commented):
LGTM

1. Need to rerun `task codegen:openapi` and `task lint:fix-rust`
2. How about this title -- adding ` and a period
```
feat(api-server): Add GET `/usage/compression` endpoint.
```

components/api-server/src/routes.rs
```diff
@@ -280,3 +332,270 @@ impl IntoResponse for HandlerError {
...
+            .await
+            .unwrap();
+        assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+        let body = get_body(response).await;
+        assert!(
+            body.contains("limit must be > 0"),
+            "expected error about limit, got: {body}"
+        );
+    }
+}
```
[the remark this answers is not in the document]
jackluo923: added duplciate job_status rejection test

components/api-server/src/routes.rs
```diff
@@ -280,3 +332,293 @@ impl IntoResponse for HandlerError {
...
+            .await
+            .unwrap();
+        assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+        let body = get_body(response).await;
+        assert!(
+            body.contains("Duplicate job_status"),
+            "expected error about duplicate job_status, got: {body}"
+        );
+    }
+}
```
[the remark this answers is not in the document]
jackluo923: no need to cover this since it is impossible in production, we do not live that long before it overflows

hoophalab (dismissed):

components/api-server/src/client.rs
```diff
@@ -14,10 +16,155 @@ use futures::{Stream, StreamExt};
...
+    /// Total uncompressed size of input files (bytes).
+    pub uncompressed_size: i64,
+    /// Total compressed archive size (bytes).
+    pub compressed_size: i64,
+    /// Number of tasks the job was split into.
+    pub num_tasks: i32,
+    /// Sum of all task durations (CPU-seconds across all parallel workers).
+    /// `None` if all task duration values are NULL in the database.
+    pub tasks_duration: Option<f64>,
+}
```
hoophalab, line 168: Sorry, I forgot that the docstrings in this struct is used in the OpenAPI document. We need to make it user-oriented:
1. We probably shouldn't reference a `.py` file in `status`'s docstring. Instead, how about changing its type from `i32` to `CompressionJobStatus`, derive `ToSchema` for `CompressionJobStatus` , and add docstrings in `CompressionJobStatus`. OpenAPI should be able to reference `enum CompressionJobStatus` directly.
   - there is another dev-oriented docstring "Mirror of ..." on top of `CompressionJobStatus`. We should be able to hide it from OpenAPI by replacing `///` with `//`.
2. It might be worth renaming `status` as `job_status` to keep consistent with `struct CompressionUsageParams`. `sqlx::rename` should allow us to use different struct field names and column names.
3. It might be worthwhile to rename `start_time` as `time_begin_millisecs` to keep consistent with `QueryConfig`. Also `creation_time` -> `time_creation_millisecs`.
4. Recent log ingestor changes also include units in the API field names. So, it's best to rename
    - `(un)?compressed_size` as `(un)?compressed_size_bytes` or `(un)?compressed_bytes`.
    - `.*duration` as `.*duration_secs`
5. Can we not exposing database details such as `WHERE clause ` and `NULL value in database` since this struct is user oriented.
6. Can we change `/// A single row returned by the compression usage query (one row per job).` to something like `/// Resource usage statistics for the compression job with the specified id.`

components/api-server/src/client.rs
```diff
@@ -14,10 +16,155 @@ use futures::{Stream, StreamExt};
...
-use utoipa::ToSchema;
+use utoipa::{IntoParams, ToSchema};
 
 pub use crate::error::ClientError;
 
+/// Default job statuses to include when the caller does not specify `job_status`.
+/// Covers all terminal states that consumed compute resources (succeeded, failed, killed).
+///
+/// Note: `Pending` is intentionally excluded because pending jobs have no
+/// `start_time` and are therefore excluded by the time-range WHERE clause.
```
hoophalab: This NOTE duplicates with `Covers all terminal states that consumed compute resources` since `PENDING` is a non-terminal state?

components/api-server/src/client.rs
```diff
@@ -14,10 +16,155 @@ use futures::{Stream, StreamExt};
...
+];
+
+/// Query parameters for the compression usage endpoint.
+#[derive(Deserialize, IntoParams)]
+#[into_params(parameter_in = Query)]
+pub struct CompressionUsageParams {
+    /// Start of usage window (epoch milliseconds, inclusive).
+    pub begin_timestamp: i64,
+    /// End of usage window (epoch milliseconds, inclusive).
+    pub end_timestamp: i64,
```
hoophalab: 1. Thinking about this again: is it better to make `end_timestamp` exclusive? `end` in `vec[start: end]` is exclusive
2. To keep consistent with the query API, can we rename `begin_timestamp` as `time_range_begin_millisecs` and `end_timestamp` as `time_range_end_millisecs`

components/api-server/src/client.rs
```diff
@@ -14,10 +16,155 @@ use futures::{Stream, StreamExt};
...
+    /// Job statuses to include as a comma-separated list (e.g.
+    /// `job_status=succeeded,failed`). Recognized values (case-insensitive):
+    /// `PENDING`, `RUNNING`, `SUCCEEDED`, `FAILED`, `KILLED`.
+    /// Defaults to `SUCCEEDED,FAILED,KILLED` (all terminal states).
+    #[serde(default)]
+    pub job_status: Option<String>,
+    /// Maximum number of jobs to return. Must be > 0; defaults to 1000.
+    #[serde(default = "default_limit")]
+    pub limit: i64,
```
hoophalab, line 47:
```suggestion
    pub limit: u64,
```
Can we change i64 to u32 since it > 0? Also, using u64 in rust doesn't actually give us performance benefits on 64-bit systems.

components/api-server/src/client.rs
jackluo923: done

components/api-server/src/client.rs
jackluo923: done

components/api-server/src/client.rs
jackluo923: done

components/api-server/src/client.rs
jackluo923: let's keep it as i64 since it produce better error message. 
"invalid type: expected u32, found -1" error 
vs
"limit must be > 0"

components/api-server/src/client.rs
```diff
@@ -14,10 +16,157 @@ use futures::{Stream, StreamExt};
...
+/// Validated parameters produced by [`TryFrom<CompressionUsageParams>`].
+///
+/// This type is intentionally non-constructable outside of the `TryFrom` impl
+/// so that callers of [`Client::get_compression_usage`] cannot bypass validation.
+pub struct ValidatedCompressionUsageParams {
+    pub begin_timestamp: DateTime<Utc>,
+    pub end_timestamp: DateTime<Utc>,
+    pub job_statuses: Vec<CompressionJobStatus>,
+    pub limit: i64,
+}
```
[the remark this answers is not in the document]
jackluo923: fixed docstring instead.

components/api-server/src/client.rs
```diff
@@ -726,16 +728,16 @@ impl Client {
             FROM compression_jobs j \
             JOIN compression_tasks t ON t.job_id = j.id \
             WHERE j.start_time >= ? \
-              AND j.start_time <= ?\
+              AND j.start_time < ?\
```
hoophalab:
```suggestion
              AND j.start_time < ? \
```

hoophalab (commented):
lgtm. One nitpick

hoophalab (approved):

components/api-server/src/routes.rs
```diff
@@ -243,6 +260,41 @@ async fn get_timestamp_column_names(
     Ok(Json(names))
 }
 
+#[utoipa::path(
+    get,
+    path = "/usage/compression",
+    description = "Returns one row per compression job within the given \
+        epoch-millisecond time range.",
```
kirkrodrigues, line 267:
```suggestion
    description = "Gets resource usage statistics for compression jobs \
        within the given time range.",
```

kirkrodrigues (approved):

hoophalab (approved):

PR #2203, opened 2026-04-14
fix(core): Add retries with exponential backoff for transient S3 read failures before any data is read.

goynam:
When reading log files from S3, transient network errors (HTTP errors, SSL connection resets, empty replies, timeouts) during reader creation or file type deduction cause the compression task to fail permanently. Since there is no retry mechanism, the affected log files are never ingested.

### Changes

- Add `NetworkUtils::is_retryable_curl_error()` in `Utils.hpp/cpp` to detect transient curl errors (`CURLE_HTTP_RETURNED_ERROR`, `CURLE_SSL_CONNECT_ERROR`, `CURLE_GOT_NOTHING`, `CURLE_RECV_ERROR`, `CURLE_OPERATION_TIMEDOUT`).
- Add `try_create_reader_and_deduce_type_with_retries()` in `InputConfig.hpp/cpp` — a shared utility that wraps reader creation and file type deduction with retry logic (3 retries, exponential backoff starting at 1s).
- Simplify `JsonParser::ingest()` and `log_converter::convert_files()` to use the new utility, replacing inline retry loops.

### Scope

This PR only handles transient errors that occur **before** any data is ingested (i.e., during reader creation and input type deduction). Retrying after partial ingestion would require more extensive changes to the ingestion pipeline and is left for a follow-up.


gibber9809 (changes requested):
I can see this PR is aiming to handle two different failure scenarios for both `clp-s` and `log-converter`:
1. We hit a retryable network error before trying to ingest any data from some input
2. We hit a retryable network error after ingesting some data from some input

I think the first scenario is handled properly here for both `clp-s` and `log-converter`, but the second scenario will still have some issues. Specifically, in the `log-converter` path I think we'll end up with some logs being ingested more than once, and the `clp-s` path probably won't work since _most_ of the code in `JsonParser` currently assumes that any read failure leads to overall compression failure, and the error handling paths currently leave `JsonParser` in a state where it can't continue ingestion.

Since handling this second scenario properly will require some more extensive changes, I think it might make sense to reduce our scope to just handling the first scenario in this PR. For that, we could probably create a new input utility like `try_create_reader_and_deduce_input_type_with_retries` in `InputConfig` that does the whole create reader -> deduce type -> check for network failure, back off, and retry loop. Then we could use that utility in both `log_converter.cpp` and `JsonParser.cpp` to keep the changes in those files minimal.

gibber9809 (dismissed):
LGTM! Thanks for the changes. Directly edited the PR title as well. Should be good to merge once the CI finishes.

gibber9809 (approved):

kirkrodrigues (approved):
Deferring to @gibber9809's review.

For the PR title, how about:

```
fix(core): Add retries with exponential backoff for transient S3 read failures before any data is read.
```

Can we also edit the PR description to more accurately reflect the changes in this PR?

PR #2215, opened 2026-04-21
feat(job-orchestration): Cache serialized search config blob to improve scheduler throughput.

gibber9809:
# Description

This PR follows up on #2184 by improving search task serialization speed to further increase scheduler throughput. This is achieved by passing around the search job config as raw msgpack bytes to avoid reserializing it for each individual search task.

Overall this further increases scheduler throughput to ~20.5ms/batch from ~26.5ms/batch on the same Xeon server setup.

# Checklist

  breaking change.

# Validation performed

* Benchmarked query performance in scenario with 50 parallel jobs running set of 600 queries over 65GiB of data, and measured improved scaling
* Tested that normal package operations (sbin search, webui search, api server search, cancelling, log viewing, etc.) appear to work as expected

## Summary by CodeRabbit

* **Refactor**
  * Search job configuration now flows as a serialized binary payload across the execution pipeline instead of in-memory dictionaries.
  * Query jobs cache a serialized configuration payload for faster repeated access and reduced serialization overhead.
  * Job dispatching and scheduling now use the cached serialized config for search jobs, improving performance, consistency and resource efficiency.


components/job-orchestration/job_orchestration/executor/query/fs_search_task.py
```diff
@@ -213,7 +214,7 @@ def search(
     self: Task,
     job_id: str,
     task_id: int,
-    job_config: dict,
+    search_config_blob: bytes,
```
junhaoliao: i agree calling this a `_blob` is clearer

possible to name this `job_config_blob` for some consistency with the SQL table column name?

components/job-orchestration/job_orchestration/scheduler/scheduler_data.py
```diff
@@ -44,13 +45,19 @@ class QueryJob(BaseModel, ABC):
     state: InternalJobState
     start_time: datetime.datetime | None = None
     current_sub_job_async_task_result: Any | None = None
+    cached_config_blob: bytes | None = Field(default=None, exclude=True, repr=False)
```
junhaoliao: is it better to use `PrivateAttr` instead of `Field(default=None, exclude=True, repr=False)`?

junhaoliao (dismissed):
for the title, how about:
```
feat(job-orchestration): Cache serialized search config blob to improve scheduler throughput.
```

junhaoliao (approved):
> for the title, how about:
> 
> ```
> feat(job-orchestration): Cache serialized search config blob to improve scheduler throughput.
> ```

maybe this was missed - here's the proposed title change earlier

PR #2216, opened 2026-04-22
refactor(clp-s): Use `simdjson::builder::string_builder` for JSON string escaping; Bump simdjson version to 4.6.4.

ShangDanLuXian:
# Description

This PR updates the JSON string escaping path used during `clp-s` JSON marshalling.

Main changes:
- Bump the simdjson dependency used by our local deps flow from `v3.13.0` to `v4.6.4`.
- Update build/dependency wiring for simdjson:
  - `components/core/CMakeLists.txt`
  - `components/core/src/clp_s/CMakeLists.txt`
  - `taskfiles/deps/main.yaml`
  - `docs/src/dev-docs/components-core/index.md`
- Require `simdjson >= 4.0.0` at configure time.
- Add a reusable `SimdJsonStringEscaper` backed by `simdjson::builder::string_builder`.
- Update the marshalling path to use the reusable simdjson-backed escaper for string values.
- Keep the existing scalar `StringUtils::escape_json_string` implementation and use it as a fallback if the simdjson-based path fails.

Why this change:
- `StringUtils::escape_json_string` shows up as a significant hotspot when marshalling string-heavy datasets.
- Reusing a `simdjson::builder::string_builder` avoids per-call builder construction overhead while still using simdjson’s optimized escaping implementation.
- Keeping the existing scalar implementation as a fallback preserves the previous behavior if the simdjson-based escape path returns an error.

# Checklist

  breaking change.

# Validation performed

- Performed output sanity checks on `hive-24hr` and MongoDB datasets.
- Compared scalar vs simdjson-backed outputs for representative queries and confirmed matching results.
- Benchmarked `hive-24hr` decompression on my side:
  - scalar average: about `22.1s`
  - simdjson average: about `9.2s`
- Generated flame graphs for the `hive-24hr` decompression path and confirmed that JSON string escaping is no longer a dominant hotspot after the change.

## Summary by CodeRabbit

* **Chores**
  * Require simdjson >= 4.0.0 at configure time and updated dependency download metadata to v4.6.4.
* **Enhancements**
  * Added a simdjson-backed JSON string escaper and switched serializers/readers to use it, improving JSON escaping and output reliability.
  * Ensured string-processing components link against simdjson when enabled.
* **Documentation**
  * Updated components-core dependency table to reflect the simdjson version bump.

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/clp/pull/2216?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


components/core/src/clp_s/Utils.cpp
```diff
@@ -226,6 +198,17 @@ bool UriUtils::get_last_uri_component(std::string_view const uri, std::string& n
...
+    json_string_builder.clear();
+    json_string_builder.escape_and_append(source);
+    std::string_view escaped_source;
+    auto const status = json_string_builder.view().get(escaped_source);
+    if (simdjson::SUCCESS == status) {
+        destination.append(escaped_source);
+        return;
+    }
+
+    // Preserve the original implementation in case simdjson fails.
```
ShangDanLuXian: Or we may remove it completely?

gibber9809: Yeah, I would just remove the original implementation, and throw if the simdjson function fails.

components/core/CMakeLists.txt
```diff
@@ -211,7 +211,7 @@ if(CLP_NEED_NLOHMANN_JSON)
 endif()
 
 if(CLP_NEED_SIMDJSON)
-    find_package(simdjson REQUIRED)
+    find_package(simdjson 4.6.2 REQUIRED)
```
gibber9809:
```suggestion
    find_package(simdjson 4.3.0 REQUIRED)
```

To make it easier to integrate with other projects we should probably choose the earliest version that serves our needs -- it seems like [4.3.0](https://github.com/simdjson/simdjson/releases/tag/v4.3.0) is the first version with the string escaping optimizations, so let's use that as our required version.

components/core/src/clp_s/CMakeLists.txt
```diff
@@ -387,6 +387,7 @@ if(CLP_BUILD_CLP_S_ARCHIVEREADER)
                 clp_s::timestamp_pattern
                 msgpack-cxx
                 nlohmann_json::nlohmann_json
+                simdjson::simdjson
```
gibber9809, line 401: I think this should be `PRIVATE`, since it's transitive through `Utils.cpp`, not through a header file.

We should probably also split out `Utils.{cpp,hpp}` into a library in our build since those utilities are used by a few of our libraries, but I won't make you do that as part of this PR. @coderabbitai, could you create an issue to track splitting off these utilities into a separate library in our build?

taskfiles/deps/main.yaml
```diff
@@ -483,8 +483,8 @@ tasks:
             - "-DCMAKE_BUILD_TYPE=Release"
             - "-DCMAKE_INSTALL_MESSAGE=LAZY"
           LIB_NAME: "simdjson"
-          TARBALL_SHA256: "07a1bb3587aac18fd6a10a83fe4ab09f1100ab39f0cb73baea1317826b9f9e0d"
-          TARBALL_URL: "https://github.com/simdjson/simdjson/archive/refs/tags/v3.13.0.tar.gz"
+          TARBALL_SHA256: "c240d4bffcccda4fe3a2bba2872718d96fc92e56d2615bfac4f9b2bad89a6386"
+          TARBALL_URL: "https://github.com/simdjson/simdjson/archive/refs/tags/v4.6.2.tar.gz"
```
gibber9809:
```suggestion
          TARBALL_SHA256: "bde0c42f43899c4c5c48be826c09abd22500ed537b89f16b3cced5eec477c263"
          TARBALL_URL: "https://github.com/simdjson/simdjson/archive/refs/tags/v4.6.3.tar.gz"
```
Might as well pull the newest version released a few days ago.

docs/src/dev-docs/components-core/index.md
```diff
@@ -50,7 +50,7 @@ The task will download, build, and install (within the build directory) the foll
 | [mongo-cxx-driver](https://github.com/mongodb/mongo-cxx-driver)       | r4.1.1         |
 | [msgpack-cxx](https://github.com/msgpack/msgpack-c/tree/cpp_master)   | v7.0.0         |
 | [nlohmann_json](https://github.com/nlohmann/json)                     | v3.11.3        |
-| [simdjson](https://github.com/simdjson/simdjson)                      | v3.13.0        |
+| [simdjson](https://github.com/simdjson/simdjson)                      | v4.6.2         |
```
gibber9809:
```suggestion
| [simdjson](https://github.com/simdjson/simdjson)                      | v4.6.3         |
```
Per other comment.

gibber9809 (changes requested):
Nice work! Just a few things we should change.

components/core/CMakeLists.txt
gibber9809: Actually, sorry it might be better to require 4.0.0, which introduces the string builder API. Reason being that it seems like the version of velox integrate with currently pulls in simdjson 4.1.0, so requiring a newer version than that would cause problems.

gibber9809:
I did some more benchmarking on regular JSON datasets, and I noticed that this actually makes serialization a bit slower on average.

From profiling it seems like a lot of the regression comes from escaping keys -- right now we don't pre-escape keys at the leaf level, so for every field we end up calling into this escaping function, which appears to always do some heap allocation to create the string builder before doing the escaping. Since the previous implementation was non-allocating, and keys tend to not have any characters that require escaping (which means we don't benefit as much from simdjson's optimizations to the slow path), this ends up adding some meaningful overhead.

With that in mind, could we try leaving the current escaping function as is, and creating a new class which holds an instance of `simdjson::string_builder` w/ an escaping function that reuses that string builder instance? That way we can avoid the allocations on every invocation of the escaping function by just passing around an instance of that new class. It may also be worth testing whether we get better decompression speed if we only use the simdjson-based implementation for string values, or if it makes sense to use it to escape both keys and values.

components/core/src/clp_s/Utils.cpp
```diff
@@ -197,6 +198,17 @@ bool UriUtils::get_last_uri_component(std::string_view const uri, std::string& n
 }
 
 void StringUtils::escape_json_string(std::string& destination, std::string_view const source) {
+    simdjson::builtin::builder::string_builder json_string_builder{};
```
gibber9809:
```suggestion
    simdjson::builder::string_builder json_string_builder{};
```
Didn't catch this the first time I looked -- seems like this is the class they want us to use, based on the simdjson docs.

components/core/src/clp_s/Utils.cpp
gibber9809: Per other review comment -- should probably keep it actually, and create a new class for this new implementation.

taskfiles/deps/main.yaml
```diff
@@ -483,8 +483,8 @@ tasks:
             - "-DCMAKE_BUILD_TYPE=Release"
             - "-DCMAKE_INSTALL_MESSAGE=LAZY"
           LIB_NAME: "simdjson"
-          TARBALL_SHA256: "07a1bb3587aac18fd6a10a83fe4ab09f1100ab39f0cb73baea1317826b9f9e0d"
-          TARBALL_URL: "https://github.com/simdjson/simdjson/archive/refs/tags/v3.13.0.tar.gz"
+          TARBALL_SHA256: "ad238fc578a4c18d3263250d0efaacadbc6a98f9c7b6be2e70867011b627fa57"
+          TARBALL_URL: "https://github.com/simdjson/simdjson/archive/refs/tags/v4.0.0.tar.gz"
```
gibber9809: We can pull the newest version of simdjson here ([4.6.4](https://github.com/simdjson/simdjson/releases/tag/v4.6.4)) even though we only require 4.0.0 in our cmake scripts -- cmake won't complain about pulling in newer versions of the library, and it's worthwhile for us to use a version with more bug fixes & optimizations in our own builds.

docs/src/dev-docs/components-core/index.md
```diff
@@ -50,7 +50,7 @@ The task will download, build, and install (within the build directory) the foll
 | [mongo-cxx-driver](https://github.com/mongodb/mongo-cxx-driver)       | r4.1.1         |
 | [msgpack-cxx](https://github.com/msgpack/msgpack-c/tree/cpp_master)   | v7.0.0         |
 | [nlohmann_json](https://github.com/nlohmann/json)                     | v3.11.3        |
-| [simdjson](https://github.com/simdjson/simdjson)                      | v3.13.0        |
+| [simdjson](https://github.com/simdjson/simdjson)                      | v4.0.0         |
```
gibber9809: Ditto, can update to 4.6.4.

components/core/src/clp_s/Utils.hpp
```diff
@@ -128,6 +129,24 @@ class StringUtils {
...
+
+    SimdJsonStringEscaper(SimdJsonStringEscaper const&) = delete;
+    auto operator=(SimdJsonStringEscaper const&) -> SimdJsonStringEscaper& = delete;
+
+    void escape(std::string& destination, std::string_view const source);
+
+private:
+    class Implementation;
+    std::unique_ptr<Implementation> m_implementation;
+};
```
gibber9809, line 170: Let's change this class to own the `simdjson::builder::string_builder` + implement the escaping function directly. I think using this forward declaration to avoid pulling the simdjson header into this file is unnecessary.

components/core/src/clp_s/Utils.hpp
```diff
@@ -128,6 +129,24 @@ class StringUtils {
...
+class SimdJsonStringEscaper {
+public:
+    SimdJsonStringEscaper();
+    SimdJsonStringEscaper(SimdJsonStringEscaper&&) noexcept;
+    ~SimdJsonStringEscaper();
+
+    auto operator=(SimdJsonStringEscaper&&) noexcept -> SimdJsonStringEscaper&;
+
+    SimdJsonStringEscaper(SimdJsonStringEscaper const&) = delete;
+    auto operator=(SimdJsonStringEscaper const&) -> SimdJsonStringEscaper& = delete;
```
gibber9809: Since we're just explicitly defaulting/deleting these constructors and assignment operators we should do that all in this header instead of leaving some of it for the cpp file.

As well, we have a standard set of docstrings we use for different combinations of deleted/defaulted constructors that we should be using here as well. For example, take a look at `components/core/src/clp/ffi/ir_stream/KvIrDeserializerImpl.hpp`.

components/core/src/clp_s/CMakeLists.txt
```diff
@@ -392,6 +392,7 @@ if(CLP_BUILD_CLP_S_ARCHIVEREADER)
                 Boost::url
                 clp_s::clp_dependencies
                 fmt::fmt
+                simdjson::simdjson
```
gibber9809: Related to other comment about not needing to avoid pulling simdjson into that header, we'll also need to move `simdjson` into public linkage here.

components/core/src/clp_s/ColumnReader.hpp
```diff
@@ -80,6 +80,14 @@ class BaseColumnReader {
         extract_string_value_into_buffer(cur_message, buffer);
     }
 
+    virtual auto extract_escaped_string_value_into_buffer(
```
gibber9809, line 90: Missing docstring.

gibber9809 (changes requested):
Nice work! Mostly looking good now, just a few details.

gibber9809 (approved):
LGTM after the most recent changes! Also confirmed on my side that the benchmark numbers look good on some of the other JSON datasets. For PR title how about:

```
refactor(clp-s): Use `simdjson::builder::string_builder` for JSON string escaping; Bump simdjson version to 4.6.4.
```

Could you also do a quick pass on the PR description to make sure it describes the final set of changes? I think some little details like version numbers need to be updated.

LinZhihao-723 (approved):
Took a quick skim and lgtm overall.

PR #2224, opened 2026-04-24
feat(log-ingestor): Add `scan_prefix` helper for listing S3 object metadata; Refactor S3 scanner to use `scan_prefix` for scanning.

jonathan-imanu:
# Description

The first step in implementing the proposal for one-time jobs. This PR removes the `scan_once` in S3Scanner for a reusable function that exhausts a prefix using cancellation tokens.

PR Title / Commit message on merge:

```bash
feat(clp-rust-utils): Add `scan_prefix` helper with continuation token pagination:

* Add `scan_prefix` in for paginated S3 prefix scanning.
* Refactor `S3Scanner` in `log-ingestor `to use the new helper.
```

# Checklist

  breaking change.

# Validation performed

- Unit tests
- Workflows

## Summary by CodeRabbit

* **Refactor**
  * S3 scanner now fully iterates paginated results in one execution and exposes a reusable scan helper
  * Cursor handling tightened to require non-empty keys for resumption

* **Tests**
  * Added integration tests covering multi-page scanning and filtering of directory markers
  * Introduced shared test utilities for creating and uploading S3 objects and streamlined test helpers

* **Chore**
  * Suppressed unused-field warnings in test AWS config


components/clp-rust-utils/src/s3/scan.rs
```diff
@@ -0,0 +1,85 @@
...
+pub async fn scan_prefix<F, Fut>(
+    client: &Client,
+    bucket_name: &NonEmptyString,
+    key_prefix: &NonEmptyString,
+    start_after: Option<&NonEmptyString>,
+    mut on_page: F,
+) -> Result<Option<NonEmptyString>>
+where
+    F: FnMut(Vec<ObjectMetadata>) -> Fut,
+    Fut: Future<Output = Result<()>>, {
```
LinZhihao-723: We should prefer `AsynFnMut` for async closure (read [here](https://microsoft.github.io/RustTraining/async-book/ch10-async-traits.html#async-closures-rust-185) if you want to learn more). However, changing this would cause a compile-time error. This is the well-known HRTB-Send pitfall with `AsyncFnMut` closures that capture `&self` (rust-lang/rust#110338 family). The chain is:   
                             
1. `scan_prefix` (in clp-rust-utils/src/s3/scan.rs) takes `F: AsyncFnMut(Vec<ObjectMetadata>) -> Result<()>`. `AsyncFnMut`'s output future has an implicit higher-ranked lifetime over `&mut self` of the closure.
2. In s3_scanner.rs, the closure is `|page| self.ingest_scanned_page(page)`. `ingest_scanned_page` is `async fn(&self, ...)`, so the future it returns borrows `&Task<S3ClientManager, State>`.                                                                             
3. `tokio::spawn` at line 183 then asks the whole future to be `Send` for any lifetime of that borrow — but `auto_trait` only proves it for one specific lifetime, so it fails. This is purely a type-checker limitation; the code is logically Send-safe.                      
                             
The borrow of `&self` inside the async closure is the only thing forcing the HRTB. Everything else (`bucket_name`, `key_prefix`, `start_after`) is passed as a regular & argument and isn't part of the `AsyncFnMut`'s lifetime story.

I think the right action is:
- We should still use `AsyncFnMut` for clarity reason.
- Let's not pass `&self` inside the capture.

The S3-scanner side should be sth look like the following which inlines `ingest_scanned_page`:

```rust
  pub async fn scan_until_exhausted(&mut self) -> Result<()> {
      let client = self.s3_client_manager.get().await?;                                                                                
      let state = self.state.clone();
      let last_scanned_key = scan_prefix(                                                                                              
          &client,                                                                                                                     
          &self.config.base.bucket_name,
          &self.config.base.key_prefix,                                                                                                
          self.start_after.as_ref(),
          async move |page: Vec<ObjectMetadata>| -> Result<()> {
              let last_ingested_key =                                                                                                  
                  page.last().expect("`page` should not be empty").key.clone();
              state.ingest(page, last_ingested_key.as_str()).await                                                                     
          },                                                                                                                           
      )
      .await?;                                                                                                                         
      self.start_after = last_scanned_key;
      Ok(())
  }
```

components/clp-rust-utils/src/s3/scan.rs
LinZhihao-723, line 1: Thought it through, it might be better to put this under `log-ingestor` for the following reasons:
- This call should only be used by log-ingestor in the near future.
- log-ingestor already has the testing infra setup for localstack. I think we should add some unit tests to cover this particular scan function, especially for multi-paging cases.

LinZhihao-723 (changes requested):

components/clp-rust-utils/src/s3/scan.rs
LinZhihao-723: **Three style issues on this signature** (per `coding-style.md` and the skill):

1. **Missing `# Type Parameters` section.** The skill requires a `# Type Parameters` entry for every type parameter — this docstring jumps from the description straight to `# Returns` with no entry for `F` or `Fut`.
2. **Single-letter type-parameter names violate the naming rule.** `coding-style.md`: *"we use long descriptive names for generic type parameters, instead of single letters"*; the skill adds the `XxxType` suffix convention to disambiguate from concrete types.
3. **Constraints belong inline, not in a `where` clause.** Per the skill: *"Put constraints inline in the `<>` clause … not in a `where` clause -- easier to read and matches the style guide."*

Combined fix:

```suggestion
pub async fn scan_prefix<
    CallbackType: FnMut(Vec<ObjectMetadata>) -> CallbackFutureType,
    CallbackFutureType: Future<Output = Result<()>>,
>(
    client: &Client,
    bucket_name: &NonEmptyString,
    key_prefix: &NonEmptyString,
    start_after: Option<&NonEmptyString>,
    mut on_page: CallbackType,
) -> Result<Option<NonEmptyString>> {
```

The docstring also needs a `# Type Parameters` section listing both names.

components/clp-rust-utils/src/s3/scan.rs
```diff
@@ -0,0 +1,85 @@
...
+/// # Returns
+///
+/// The last scanned object key, if any non-directory object was scanned.
+///
+/// # Errors
+///
+/// * Forwards
+///   [`aws_sdk_s3::operation::list_objects_v2::builders::ListObjectsV2FluentBuilder::send`]'s
+///   return values on failure.
+/// * Forwards [`NonEmptyString::new`]'s return values if an empty object key is received.
```
LinZhihao-723: **`Forwards` is mis-used here — this error isn't forwarded.**

The body does `NonEmptyString::new(key).map_err(|_| anyhow::anyhow!(...))?;` — the original `NonEmptyString::new` error is dropped by `map_err` and a fresh `anyhow::Error` is constructed. Per the skill: *"Don't use `Forwards [Error::Variant]` for errors this function constructs and returns directly. Use `* [Error::Variant] if <condition>.` for direct errors; reserve `Forwards` for errors propagated via `?`."*

```suggestion
/// * Returns an [`anyhow::Error`] if S3 returns an empty object key.
```

components/clp-rust-utils/src/s3/scan.rs
```diff
@@ -0,0 +1,85 @@
...
+        let mut object_metadata_to_ingest = Vec::with_capacity(contents.len());
+        for content in contents {
+            let (Some(key), Some(size)) = (content.key, content.size) else {
+                continue;
+            };
+            if key.ends_with('/') {
+                continue;
+            }
+            let non_empty_key = NonEmptyString::new(key)
+                .map_err(|_| anyhow::anyhow!("An empty key is received."))?;
```
LinZhihao-723: **Error message should be lowercase with no trailing period.**

`coding-style.md`: *"For error messages inside an error enum, an `expect`, or `panic`, the error message should not capitalize the first letter, and it should not end with a period."* This codebase applies the same convention to anyhow context messages.

```suggestion
                .map_err(|_| anyhow::anyhow!("received an empty object key from S3"))?;
```

(This was already the pattern in the original `scan_once`; please fix while you're touching the new file.)

components/clp-rust-utils/src/s3.rs
```diff
@@ -1,7 +1,9 @@
 mod client;
+mod scan;
 
 pub use client::create_new_client;
 use non_empty_string::NonEmptyString;
+pub use scan::scan_prefix;
```
LinZhihao-723: **Minor: `pub use` re-exports should be grouped together.**

The new `pub use scan::scan_prefix;` is split from `pub use client::create_new_client;` by an unrelated internal `use`. Conventionally these are grouped.

```suggestion
pub use client::create_new_client;
pub use scan::scan_prefix;

use non_empty_string::NonEmptyString;
```

(The split was already on `main`; this PR adds a second `pub use` which makes fixing it cheaper to do now than later.)

components/clp-rust-utils/src/s3/scan.rs
```diff
@@ -0,0 +1,85 @@
+use std::future::Future;
+
+use anyhow::Result;
+use aws_sdk_s3::Client;
+use non_empty_string::NonEmptyString;
+
+use crate::s3::ObjectMetadata;
+
+/// Scans all objects under the given S3 prefix using continuation tokens and invokes a callback
+/// for each non-empty scanned page.
```
LinZhihao-723, line 7: **Minor: docstring doesn't mention what the function filters out.**

Beyond paginated listing, `scan_prefix` also:

* skips entries with `None` key or size,
* skips keys ending in `/` (directory-like entries),
* errors on empty keys.

A caller reading just the description wouldn't know any of this. Consider adding a `# Note` section after `# Returns` documenting the filtering, since these choices are opinionated for the ingestion use case and would surprise a non-ingestion reader.

components/log-ingestor/src/ingestion_job/s3_scanner.rs
```diff
@@ -90,71 +85,54 @@ impl<S3ClientManager: AwsClientManagerType<Client>, State: IngestionJobState + S
     ///
     /// # Returns
     ///
-    /// Whether the underlying `list_buckets` call was truncated.
+    /// `Ok(())` on success.
```
LinZhihao-723: **Minor: `# Returns` for `Result<()>` can be omitted.**

`coding-style.md`: *"This section can be omitted if the function doesn't return anything, or if it returns `()`."* The current text is harmless but redundant.

The same applies to `ingest_scanned_page`'s docstring at lines 119-121 below.

LinZhihao-723 (changes requested):
> **Auto-generated review** by Claude Code using the Rust PR review skill at
> <https://github.com/LinZhihao-723/claude-instruction/blob/main/rust/pr-review.md>.
> Findings are scoped to documented style rules from `coding-style.md` and the skill.

## Summary

A clean refactor that lifts pagination logic out of `S3Scanner::scan_once` into a reusable `scan_prefix` helper in `clp-rust-utils`. Architecture is sound.

**Five style violations** flagged inline as blocking:

1. Missing `# Type Parameters` on `scan_prefix` (`scan.rs:24-33`).
2. Single-letter type-parameter names `F`/`Fut` (`scan.rs:24-33`).
3. Constraints in `where` clause instead of inline (`scan.rs:24-33`).
4. `# Errors` mis-uses `Forwards` for an error this function constructs (`scan.rs:21`).
5. `anyhow!` message starts capitalized and ends with `.` (`scan.rs:65`).

**Three minor items** also called out:

* `pub use` re-exports split by an unrelated internal `use` (`s3.rs:4-6`).
* Docstring doesn't mention what `scan_prefix` filters out (`scan.rs:9-10`).
* `# Returns` redundant for `Result<()>` (`s3_scanner.rs:86-88` and `119-121`).

Coderabbit has separately raised: `None` contents with continuation token, the `expect(...)` invariant coupling in `s3_scanner.rs`, and the empty-key error-vs-skip choice. Those are not duplicated here.

## Suggested PR title (from the description, lightly polished)

```
feat(clp-rust-utils): Add `scan_prefix` helper with continuation token pagination:

* Add `scan_prefix` for paginated S3 prefix scanning.
* Refactor `S3Scanner` in `log-ingestor` to use the new helper.
```

(Drop the trailing space after `log-ingestor` in the second bullet of the PR description.)

components/log-ingestor/src/ingestion_job/scan.rs
```diff
@@ -0,0 +1,84 @@
+use anyhow::Result;
+use aws_sdk_s3::Client;
+use clp_rust_utils::s3::ObjectMetadata;
+use non_empty_string::NonEmptyString;
+
+/// Scans all objects under the given S3 prefix using continuation tokens and invokes a callback
+/// for each non-empty scanned page.
+///
+/// This function applies the following filters to the S3 objects:
```
LinZhihao-723, line 9:
```suggestion
///
/// This function applies the following filters to the S3 objects:
```

components/log-ingestor/src/ingestion_job/scan.rs
```diff
@@ -0,0 +1,84 @@
...
+/// The last scanned object key, if any non-directory object was scanned.
+///
+/// # Errors
+///
+/// * Forwards
+///   [`aws_sdk_s3::operation::list_objects_v2::builders::ListObjectsV2FluentBuilder::send`]'s
+///   return values on failure.
+/// * Forwards [`i64::try_into`]'s return values when failing to convert object size to [`u64`].
+/// * Forwards `on_page` callback return values on failure.
+/// * Returns an [`anyhow::Error`] if S3 returns an empty object key.
```
LinZhihao-723:
```suggestion
/// * [`anyhow::Error`] if the request returns an empty object key.
/// * Forwards
///   [`aws_sdk_s3::operation::list_objects_v2::builders::ListObjectsV2FluentBuilder::send`]'s
///   return values on failure.
/// * Forwards [`i64::try_into`]'s return values when failing to convert object size to [`u64`].
/// * Forwards `on_page` callback return values on failure.
```
For particular error type, it comes before `Forwards` section and has different format (like C++).

components/log-ingestor/src/ingestion_job/scan.rs
```diff
@@ -0,0 +1,84 @@
...
+pub async fn scan_prefix<CallbackType: AsyncFnMut(Vec<ObjectMetadata>) -> Result<()>>(
+    client: &Client,
+    bucket_name: &NonEmptyString,
+    key_prefix: &NonEmptyString,
+    start_after: Option<&NonEmptyString>,
+    mut on_page: CallbackType,
```
LinZhihao-723: How about `on_page` -> `page_callback`?

components/log-ingestor/src/ingestion_job/scan.rs
```diff
@@ -0,0 +1,84 @@
...
+            last_scanned_key = Some(non_empty_key.clone());
+            object_metadata_to_ingest.push(ObjectMetadata {
+                bucket: bucket_name.clone(),
+                key: non_empty_key,
+                size: size.try_into()?,
+            });
+        }
+
+        if !object_metadata_to_ingest.is_empty() {
+            on_page(object_metadata_to_ingest).await?;
```
LinZhihao-723: I think the return value of the page callback needs a redesign. Consider the following cases:
- For key ingestions, you may want an early exit once you have all the keys collected. In that case, the page callback needs to return an indicator on whether you want to keep going or not.
- For prefix ingestion, we might want to add a limit on how many objects to ingest. In that case, the entire page might not be entirely consumed, meaning that it should return the last processed key to indicate what is the actual last scanned key from the callback's view.

Therefore, I think the callback should return `Result<(bool, NonEmptyString)>` instead as:
- The first (bool) indicates whether the scan should proceed.
- The second (string) indicates the last key successfully processed by the callback.

components/log-ingestor/src/ingestion_job/scan.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    loop {
+        let response = client
+            .list_objects_v2()
+            .bucket(bucket_name.as_str())
+            .prefix(key_prefix.as_str())
+            .set_start_after(use_start_after.take())
+            .set_continuation_token(continuation_token)
+            .send()
+            .await?;
```
LinZhihao-723:
```suggestion
        let mut request = client
            .list_objects_v2()
            .bucket(bucket_name.as_str())
            .prefix(key_prefix.as_str());

        if let Some(continuation_token) = continuation_token.take() {
            request = request.continuation_token(continuation_token);
        } else if let Some(start_after) = start_after {
            request = request.start_after(start_after.to_string());
        }

        let response = request.send().await?;
```
I think the request should be restructured in this way:
- It sets all necessary parameters at the beginning.
- It then sets continuation token if exists, or, it sets start after. Start after marker shouldn't be set if a continuation token is given.

Let me know if this makes sense to you, or I might miss anything that is a fundamental limitation of the S3 API.

components/log-ingestor/src/ingestion_job/scan.rs
```diff
@@ -0,0 +1,84 @@
+use anyhow::Result;
+use aws_sdk_s3::Client;
+use clp_rust_utils::s3::ObjectMetadata;
+use non_empty_string::NonEmptyString;
+
+/// Scans all objects under the given S3 prefix using continuation tokens and invokes a callback
+/// for each non-empty scanned page.
+///
```
LinZhihao-723, line 8: `# Type Parameters` section is missing in the docstring.

LinZhihao-723 (changes requested):
Reviewed scan helper implementation in details.

components/log-ingestor/src/ingestion_job/scan.rs
```diff
@@ -0,0 +1,94 @@
...
+/// The last key successfully processed by `page_callback`, if any object was processed.
+///
+/// # Errors
+///
+/// * [`anyhow::Error`] if the request returns an empty object key.
+/// * Forwards
+///   [`aws_sdk_s3::operation::list_objects_v2::builders::ListObjectsV2FluentBuilder::send`]'s
+///   return values on failure.
+/// * Forwards [`i64::try_into`]'s return values when failing to convert object size to [`u64`].
+/// * Forwards `page_callback` callback return values on failure.
```
LinZhihao-723:
```suggestion
/// * Forwards `page_callback`'s return values on failure.
```

components/log-ingestor/src/ingestion_job/scan.rs
```diff
@@ -0,0 +1,94 @@
...
+        if let Some(continuation_token) = continuation_token.take() {
+            request = request.continuation_token(continuation_token);
+        } else if let Some(start_after) = start_after.take() {
+            request = request.start_after(start_after);
+        }
+
+        let response = request.send().await?;
+
+        let contents = response.contents.unwrap_or_default();
+        let mut object_metadata_to_ingest = Vec::with_capacity(contents.len());
```
LinZhihao-723: `object_metadata_to_ingest` -> `paged_object_metadata`?

components/log-ingestor/src/ingestion_job/scan.rs
```diff
@@ -0,0 +1,94 @@
...
+/// * Entries with a missing key or size are skipped.
+/// * Keys ending in `/` are skipped.
+/// * Empty keys cause an error.
+///
+/// # Type Parameters
+///
+/// * [`CallbackType`]: The type of the async callback invoked once per non-empty scanned page of
+///   object metadata. It must return (`should_continue_scanning`, `processed_last_key`), where
+///   `should_continue_scanning` controls whether to fetch more pages, and `processed_last_key` is
+///   the last key from that page.
```
LinZhihao-723: `processed_last_key` -> `last_processed_key`?
Also update it in the code.

LinZhihao-723:
```suggestion
/// * [`CallbackType`]: The type of the async callback invoked once per non-empty scanned page of
///   object metadata. It must return (`should_continue_scanning`, `last_processed_key`), where:
///   * `should_continue_scanning` indicates whether the scanning should proceed.
///   * `last_processed_key` is the last key processed by the callback.
```

components/log-ingestor/src/ingestion_job/s3_scanner.rs
```diff
@@ -79,76 +72,38 @@ impl<S3ClientManager: AwsClientManagerType<Client>, State: IngestionJobState + S
...
     ///
     /// * Forwards [`AwsClientManagerType::get`]'s return values on failure.
-    /// * Forwards
-    ///   [`aws_sdk_s3::operation::list_objects_v2::builders::ListObjectsV2FluentBuilder::send`]'s
-    ///   return values on failure.
-    /// * Forwards [`NonEmptyString::new`]'s return values on failure.
-    /// * Forwards [`S3ScannerState::ingest`]'s return values on failure.
-    /// * Forwards [`i64::try_into`]'s return values when failing to convert object size to [`u64`].
-    pub async fn scan_once(&mut self) -> Result<bool> {
+    /// * Forwards [`crate::ingestion_job::scan_prefix`]'s return values on failure.
```
LinZhihao-723:
```suggestion
    /// * Forwards [`scan_prefix`]'s return values on failure.
```
The name should be resolveable.

components/log-ingestor/tests/aws_config.rs
LinZhihao-723, line 1: Updated the test infra to avoid the use of `dead_code`.

LinZhihao-723 (changes requested):

components/log-ingestor/src/ingestion_job/scan.rs
```diff
@@ -0,0 +1,94 @@
+use anyhow::Result;
+use aws_sdk_s3::Client;
+use clp_rust_utils::s3::ObjectMetadata;
+use non_empty_string::NonEmptyString;
+
+/// Scans all objects under the given S3 prefix using continuation tokens and invokes a callback
+/// for each non-empty scanned page.
+///
```
LinZhihao-723, line 8:
```suggestion
```

components/log-ingestor/tests/test_scan.rs
```diff
@@ -0,0 +1,188 @@
...
+use uuid::Uuid;
+
+use super::{
+    aws_config::AwsConfig,
+    test_utils::{create_s3_objects, get_testing_prefix_as_non_empty_string, upload_test_objects},
+};
+
+const NUM_OBJECTS: usize = 250;
+// ListObjectsV2 returns at most 1,000 objects from a bucket
+const NUM_MULTI_PAGE_OBJECTS: usize = 1_205;
```
LinZhihao-723: How about:
- Use `MAX_NUM_OBJECTS_PER_PAGE = 1000`
- In `test_scan_prefix_multi_page_filtering`, we set the number of pages expected, `NUM_PAGES`, and then create `MAX_NUM_OBJECTS_PER_PAGE * NUM_PAGES` objects and assert `NUM_PAGES` objects received.

LinZhihao-723 (changes requested):
Nit comments. Otherwise we should be good to merge.

components/log-ingestor/tests/test_scan.rs
```diff
@@ -19,13 +18,14 @@ use uuid::Uuid;
...
 };
 
 const NUM_OBJECTS: usize = 250;
-// ListObjectsV2 returns at most 1,000 objects from a bucket
-const NUM_MULTI_PAGE_OBJECTS: usize = 1_205;
-const NUM_FILTERED_OUT_DIRECTORY_MARKERS: usize = 25;
+
+/// `ListObjectsV2` returns at most 1,000 keys per response.
+const MAX_NUM_OBJECTS_PER_PAGE: usize = 1000;
+const NUM_PAGES: usize = 2;
```
LinZhihao-723: I think this `NUM_PAGE` should be a per-test case param. Let's move it into the test case.

LinZhihao-723 (changes requested):
One more thing caught from the last commit. For the PR title, how about:
```
feat(log-ingestor): Add `scan_prefix` helper for listing S3 object metadata; Refactor S3 scanner to use `scan_prefix` for scanning.
```

LinZhihao-723 (approved):

PR #2227, opened 2026-04-25
feat(kv-ir): Add `UnstructuredIrDeserializerImpl` implementation for unstructured IR deserialization (resolves #2096).

### LINKED ISSUE #2096: Support IRv1 Streams in `clp::ffi::ir_stream::Deserializer`
## Background

CLP has two IR stream formats. The key-value pair IR format (KV-IR), version `0.1.0` (`Supported`), encodes structured log events as key–value pairs with an incremental schema tree and is read by `ffi::ir_stream::Deserializer`. An older format uses versions `v0.0.0`, `0.0.1`, and `0.0.2` (`BackwardCompatible`) and encodes one unstructured message with no schema tree. 

Differences in the formats and notes on the serialization & deserialization flows are in Appendix.

### Current Behavior

When `ffi::ir_stream::Deserializer::create_generic` creates a KV-IR deserializer, it reads the stream preamble (magic number and JSON metadata) and then checks the metadata version. It only proceeds when the version is `Supported`. If the version is `BackwardCompatible`, the stream is rejected.

As of today, the old format is only read by `clp::ir::LogEventDeserializer` in `ir::LogEventDeserializer` which is used by `FileCompressor` when compressing old IR streams. Callers that use a single entry point for IR, such as `JsonParser::ingest_kvir`, use the KV-IR deserializer only and therefore cannot ingest old IR streams.

## Proposal

We will extend the KV-IR deserializer so that it accepts `BackwardCompatible` IR versions in addition to `Supported`.

The plan to do so is the following:

1. Extend the `ffi::ir_stream::Deserializer` to accept old IR at create time and store the following information as class members:
    - Whether stream format is old IR or KV-IR
    - Encoding type, if stream format is old IR
    - `previous_timestamp`, initialized from `REFERENCE_TIMESTAMP` metadata for four-byte old IR
    - An old IR user schema tree and the node IDs for `message` and `timestamp`
2. In `ffi::ir_stream::Deserializer::deserialize_next_ir_unit` if the stream is old IR, handle these old-IR units using the existing methods in `ffi::ir_stream::decoding_methods`.
3. For old-IR events we convert each deserialized `ir::LogEvent` (composed of a timestamp and `EncodedTextAst`) into a `KeyValuePairLogEvent` with:
    - a `message` key whose value is the `EncodedTextAst`
    - a `timestamp` key whose value is the absolute event timestamp
  
We do this because existing callers already handle `KeyValuePairLogEvent` values that are `EncodedTextAst`.

Future work (whose implementation is not detailed in this proposal) will include:

1. Reimplementing `ir::LogEventDeserializer` on top of the `ffi::ir_stream::Deserializer` and convert the `KeyValuePairLogEvent` into `ir::LogEvent` for existing callers.
2. Deprecating unused APIs in `ffi::ir_stream::decoding_methods` (were previously used by `ir::LogEventDeserializer`) by moving to a new namespace.
3. `clp_s::JsonParser`, a user of `ir_stream::Deserializer`, doesn't implement `handle_utc_offset_change` in it's `IrUnitHandler`. So it will fail if a IRv1 stream has a `UtcOffsetChange`. A future PR should fix this.

This will have the following consequences:

1. The public handler interfaces for both the `ir::LogEventDeserializer` and the `ffi::ir_stream::Deserializer` will remain unchanged.
2. Users of the KV-IR deserializer such as `JsonParser::ingest_kvir` will be able to deserialize old IR streams (after 3. of future work has been done).
3. Usage and functionality of `ir::LogEventDeserializer` will remain unchanged.

### Updates to `Deserializer::create_generic` & `Deserializer` class members

In `create_generic` we will accept `validate_protocol_version(version)` equal to either `Supported` or `BackwardCompatible`. We will indicate the stream format with the following enum:

```cpp
// Defined in Deserializer.hpp
enum class StreamFormat {
    KvIr,
    IrV1,
};
```

In order to support deserializing IRv1 streams, we must also store the encoding type. For four-byte IRv1 streams we must store `previous_timestamp` so that each log event’s timestamp can be reconstructed from deltas. This is not needed for eight-byte old IR.

We also store an old IR user schema tree and the node IDs for `message` and `timestamp`. This is described in detail in **Building** `KeyValuePairLogEvent` **from old** `LogEvent`

The updated constructor and class vars are:

```cpp
// Updated private constructor
Deserializer(
        IrUnitHandlerType ir_unit_handler,
        nlohmann::json metadata,
        QueryHandlerType query_handler,
        StreamFormat stream_format,
        std::optional<bool> is_four_byte_encoded = std::nullopt,
        std::optional<epoch_time_ms_t> previous_timestamp = std::nullopt,
        std::shared_ptr<SchemaTree const> old_ir_user_schema_tree = nullptr,
        std::optional<SchemaTree::Node::id_t> old_ir_message_node_id = std::nullopt,
        std::optional<SchemaTree::Node::id_t> old_ir_timestamp_node_id = std::nullopt
)
        : m_metadata(std::move(metadata)),
          m_ir_unit_handler{std::move(ir_unit_handler)},
          m_query_handler{std::move(query_handler)},
          m_stream_format(stream_format), 
          m_is_four_byte_encoded(is_four_byte_encoded),
          m_previous_timestamp(previous_timestamp),
          m_old_ir_user_schema_tree(std::move(old_ir_user_schema_tree)),
          m_old_ir_message_node_id(old_ir_message_node_id),
          m_old_ir_timestamp_node_id(old_ir_timestamp_node_id) {}

// Updated variables. 
// Existing variables are unchanged and omitted for brevity.
StreamFormat m_stream_format{StreamFormat::KvIr};
std::optional<bool> m_is_four_byte_encoded;  // encoding type for old IR 
std::optional<epoch_time_ms_t> m_previous_timestamp;  // for four-byte old IR timestamp delta 
std::shared_ptr<SchemaTree const> m_old_ir_user_schema_tree;
std::optional<SchemaTree::Node::id_t> m_old_ir_message_node_id;
std::optional<SchemaTree::Node::id_t> m_old_ir_timestamp_node_id;
```

### Updating `deserialize_next_ir_unit` to handle old IR

In `deserialize_next_ir_unit`, when the stream format is `StreamFormat::IrV1`, the flow will be similar to that of `LogEventDeserializer<encoded_variable_t>::deserialize_log_event`:

1. Deserialize a tag
2. If the tag is `Eof`, return `IrUnitType::EndOfStream` and run the end-of-stream handler.
3. If the tag is `UtcOffsetChange`, deserialize the new UTC offset, update the deserializer’s `m_utc_offset`, call the handler’s `handle_utc_offset_change`, and return `IrUnitType::UtcOffsetChange`.
4. Otherwise we have a old-IR log event. We will:
    - Call the existing `ffi::ir_stream::deserialize_log_event`. 
      - For four-byte encoding, we compute absolute timestamp with `previous_timestamp += timestamp_delta` where `previous_timestamp` is initialized with `REFERENCE_TIMESTAMP`.
    - By this point we will have a `EncodedTextAst{ logtype, dict_vars, encoded_vars }`.
    - We will use this to build a `KeyValuePairLogEvent` as described below.
    - Evaluate query if a query handler is present.
    - Call `m_ir_unit_handler.handle_log_event`.
    - Increment `m_next_log_event_idx`
    - Finally return `IrUnitType::LogEvent`.

#### Building `KeyValuePairLogEvent` from old `LogEvent`

We call `KeyValuePairLogEvent::create` with the following:

- **User-generated schema tree**: Root plus two children:
  - `message` (`Str`)
  - `timestamp` (`Int`)
- **Auto-generated schema tree**: One empty tree 
- **Auto-generated node-id value pairs**: Empty.
- **User-generated node-id value pairs**:
  - `message_node_id -> Value(encoded_text_ast)` where `encoded_text_ast` is either `EightByteEncodedTextAst` or `FourByteEncodedTextAst` depending on stream encoding
  - `timestamp_node_id -> Value(absolute_timestamp)`
- **UTC offset**: `m_utc_offset`.

For performance, we will build this old-IR schema once during deserializer creation (for
`StreamFormat::IrV1`) and reuse it for all old-IR events in the stream. In `Deserializer::create_generic`, we will:

- Create one user-generated schema tree and insert `message` & `timestamp` nodes once.
- Store the shared pointer and the two node IDs in deserializer state.
- If query handling is enabled, we call `QueryHandler::update_partially_resolved_columns` with the node IDs.
  - We must do this here becuase old IR streams do not have SchemaTreeNodeInsertion units.
- For each event, only rebuild node-id value pairs and create the `KeyValuePairLogEvent`.

# Appendix

### Format

Starts with a preamble that contains the following:

- A 4-byte magic number that indicates whether the encoding is 4-byte or 8-byte
- A block of metadata that is serialized JSON (<64 KiB in size) that contains:
  - The version of the stream (this is `BackwardCompatible`)
  - The version and ID of the variable schema
  - Timezone and Timestamp information
  - If the encoding is four-byte, a reference timestamp is included.

Following this preamble, there is a sequence of log events. Each log event contains a message and a timestamp. It is encoded as:

- Variables in occurrence order, then logtype
- A timestamp tag and then its value.

### Serialization Flow

The implementation is in `ir::LogEventSerializer`. It writes IR into an internal buffer that is compressed with Zstd when flushed to disk. For each file, the flow is: 

- It serializes the preamble and then flushes it
- It serializes multiple log events using the following methodology (recall that a log event contains a message and timestamp):
  - A message is split into tokens based on delimiters. Delimiters are any character not in `+-.0-9A-Z\_a-z`.
  - A variable is a token that matches some criteria. An example of such criteria is "immediately after = and contains a letter". 
  - Variables are encoded in the order in which they appear. 
  - Everything else is considered a constant. These constants are encoded into the logtype.
  - After writing the variables and logtype, the timestamp is encoded with a timestamp tag and then its value. For 4-byte encodings, this value is a delta otherwise it is an absolute timestamp.

### Deserialization Flow

The implementation is in `ir::LogEventDeserializer`. It operates on the uncompressed IR stream. Each instance operates on a single file. The class is templated on the encoding type (four-byte or eight-byte). The flow is:

- The deserializer is created with a factory function. It deserializes the preamble, parses the metadata JSON, and validates that the version is `BackwardCompatible`. For four-byte encoding it reads the reference timestamp from the metadata.
- The caller then repeatedly calls `deserialize_log_event()`.  Each call reads the next tag: 
  - If Eof, the stream is done
  - If UtcOffsetChange, the deserializer updates its UTC offset and continues to the next tag
  - Otherwise it deserializes one log event:
    - Variables in order, then logtype, then timestamp.
    - For four-byte encoding, the timestamp in the stream is a delta. So the deserializer adds it to the reference or previous timestamp to obtain the absolute timestamp.
    - The method returns a `LogEvent`. This type has a timestamp and `EncodedTextAst` with logtype and variable values.

### Format

Similar to the old IR format, the KV-IR format starts with a preamble. It differs only in the metadata contents. The metadata in this preamble contains:

- A stream version of `0.1.0`
- User-defined metadata as a JSON object for stream-level attributes
- The version and ID of the variable schema

Following this preamble, there is a sequence of IR units. Each unit is preceded by a tag that identifies its type. The sequence may contain:

- Schema tree node insertions
- Key-value log events
- UTC offset change
- End-of-stream

The stream ends with an end-of-stream unit.

### Serialization Flow

The implementation is in `ffi::ir_stream::Serializer`. It operates on an uncompressed KV-IR stream. It does not do FileIO but rather maintains an in-memory buffer. Callers are responsible for writing this buffer to the real stream. The class is templated on the encoding type (four-byte or eight-byte). The flow is:

The serializer is created with a factory function that accepts additional user metadata. It then build preamble and serializes it in it's internal buffer. The caller must write that buffer to the real stream before or when adding log events.

Log events are added with `serialize_msgpack_map(auto_gen_kv_pairs_map, user_gen_kv_pairs_map)`. Both arguments are msgpack maps. 

One example usage of this method can be seen in `LogSerializer::add_message`. This class is the customer facing API and is responsbile for writing the serializer's internal buffer to a file. It builds one msgpack map per log event for user-gen and passes an empty map for auto-gen. 

As for the semantics of `serialize_msgpack_map(auto_gen_kv_pairs_map, user_gen_kv_pairs_map)`:

- The serializer maintains two internal schema trees. For each key in the maps:
  - If the key is new, it emits a schema tree node insertion 
  - Otherwise, it reuses the existing node id 
  - It then serializes the value and appends the resulting bytes are to the internal buffer.
- The caller is responsible for writing the buffer to the stream when appropriate, writing the Eof tag and closing the stream when done.

### Deserialization Flow

The implementation is in `ffi::ir_stream::Deserializer`. It operates on an uncompressed KV-IR stream supplied by the caller. This class is templated on an IR unit handler type and an optional query handler type, so callers can define how units are handled and how queries and column resolution work.

The deserializer is created with a factory function. On creation, it reads the magic number and then deserilizes the metadata. Then validates that the stream version is `0.1.0` before returning a deserializer.

Callers are expected to repeatedly call `deserialize_next_ir_unit` until either the stream ends or an error happens. Each call reads the IR Unit, dispatches by unit type and calls the supplied `ir_unit_handler`.

jonathan-imanu:
# Description

Completes the implementation of #2096. 

# Checklist

  breaking change.

# Validation performed

- Added a unit test for old IR deserilizaton.

## Summary by CodeRabbit

* **New Features**
  * Backward-compatible unstructured IR stream support with selectable encoding variants and preserved message/timestamp semantics.

* **Bug Fixes / Diagnostics**
  * Improved deserialization error reporting for invalid reference-timestamp metadata, invalid timestamp values, and missing required schema nodes.

* **Chores**
  * Build updated to include the new unstructured deserializer components.

* **Tests**
  * New end-to-end tests validating serialization/deserialization of unstructured log events across encoding variants.


components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.hpp
```diff
@@ -0,0 +1,124 @@
...
+     */
+    [[nodiscard]] auto resolve_required_node_ids(
+            std::shared_ptr<SchemaTree> const& user_gen_keys_schema_tree
+    ) -> ystdlib::error_handling::Result<std::pair<SchemaTree::Node::id_t, SchemaTree::Node::id_t>>;
+
+    // Variables
+    std::vector<SchemaTree::NodeLocator> m_pending_schema_insertions;
+    std::optional<SchemaTree::Node::id_t> m_optional_message_node_id;
+    std::optional<SchemaTree::Node::id_t> m_optional_timestamp_node_id;
+    ir::epoch_time_ms_t m_previous_timestamp{0};
```
LinZhihao-723: This member shouldn't exist for eight-byte encoded vars. Check: https://github.com/y-scope/clp/blob/32d2d4301f33e8ce1b475e76cf7a0bf09b403338/components/core/src/clp/ir/LogEventDeserializer.hpp#L78

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,300 @@
...
+    for (auto const& dict_var : dict_vars) {
+        string_blob.append(dict_var);
+    }
+    string_blob.append(logtype);
+
+    return EncodedTextAst<encoded_variable_t>::create(
+            std::move(encoded_vars),
+            std::move(string_blob)
+    );
+}
```
LinZhihao-723: We shouldn't need this. The existing encoded text AST method can be used to deserialize unstructured IR, which uses string blob directly.

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,300 @@
...
+UnstructuredIrDeserializerImpl<encoded_variable_t>::deserialize_ir_unit_schema_tree_node_insertion(
+        [[maybe_unused]] ReaderInterface& reader,
+        [[maybe_unused]] encoded_tag_t tag,
+        std::string& key_name
+) -> ystdlib::error_handling::Result<std::pair<bool, SchemaTree::NodeLocator>> {
+    auto const node_locator{m_pending_schema_insertions.back()};
+    m_pending_schema_insertions.pop_back();
+
+    key_name.clear();
+    return std::pair{false, node_locator};
```
LinZhihao-723: A design decsion I forgot to mention: the timestamp key should be considered auto-generated to match our convention in the kv-ir logging library. I think the ``pending_schema_insertions` should have a pair of `is_auto_gen, node_locator` instead.

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,300 @@
...
+                std::shared_ptr<SchemaTree> const& auto_gen_keys_schema_tree,
+                std::shared_ptr<SchemaTree> const& user_gen_keys_schema_tree,
+                UtcOffset utc_offset
+        ) -> ystdlib::error_handling::Result<KeyValuePairLogEvent> {
+    std::string logtype;
+    std::vector<ir::four_byte_encoded_variable_t> encoded_vars;
+    std::vector<std::string> dict_vars;
+    ir::epoch_time_ms_t timestamp_delta{};
+
+    if (auto const err{deserialize_log_event<ir::four_byte_encoded_variable_t>(
```
LinZhihao-723: As mentioned in the previous comment, I think we should use the latest version of `deserialize_encoded_text_ast` directly as kv-ir, and then deserialize the timestamp/timestamp-delta manually from the stream. Check [this](https://github.com/y-scope/clp/blob/32d2d4301f33e8ce1b475e76cf7a0bf09b403338/components/core/src/clp/ffi/ir_stream/decoding_methods.cpp#L456) implementation for details.

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,300 @@
...
+    auto const tag{tag_result.value()};
+
+    if (cProtocol::Eof == tag) {
+        return std::pair{IrUnitType::EndOfStream, tag};
+    }
+    if (cProtocol::Payload::UtcOffsetChange == tag) {
+        return std::pair{IrUnitType::UtcOffsetChange, tag};
+    }
+
+    return std::pair{IrUnitType::LogEvent, tag};
```
LinZhihao-723:
```suggestion
    return {IrUnitType::LogEvent, tag};
```
fyi `std::pair` is redundant. Remove it (also for other occurrences) for clarity.

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.hpp
```diff
@@ -0,0 +1,124 @@
...
+private:
+    // Constructors
+    explicit UnstructuredIrDeserializerImpl(
+            std::vector<SchemaTree::NodeLocator> initial_insertions
+    );
+
+    UnstructuredIrDeserializerImpl(
+            std::vector<SchemaTree::NodeLocator> initial_insertions,
+            ir::epoch_time_ms_t reference_timestamp
+    );
```
LinZhihao-723: We should add constraints to only enable the first method for eight-byte encoded, and the second for four-byte encoded. Check `std::enable_if_t`: https://en.cppreference.com/cpp/types/enable_if.

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,300 @@
...
+)
+        : m_pending_schema_insertions{std::move(initial_insertions)} {}
+
+template <ir::EncodedVariableTypeReq encoded_variable_t>
+UnstructuredIrDeserializerImpl<encoded_variable_t>::UnstructuredIrDeserializerImpl(
+        std::vector<SchemaTree::NodeLocator> initial_insertions,
+        ir::epoch_time_ms_t reference_timestamp
+)
+        : m_pending_schema_insertions{std::move(initial_insertions)},
+          m_previous_timestamp{reference_timestamp} {}
```
LinZhihao-723: Please pay attention to the symbol declaration ordering. The constructor is declared after other functions, then its definition should follow the declaration order.

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,300 @@
...
+) -> ystdlib::error_handling::Result<std::pair<IrUnitType, encoded_tag_t>> {
+    if (false == m_pending_schema_insertions.empty()) {
+        return std::pair{IrUnitType::SchemaTreeNodeInsertion, encoded_tag_t{0}};
+    }
+
+    auto const tag_result{deserialize_tag(reader)};
+    if (tag_result.has_error()) {
+        return tag_result.error();
+    }
+    auto const tag{tag_result.value()};
```
LinZhihao-723: Why not:
```suggestion
    auto const tag{YSTDLIB_ERROR_HANDLING_TRYX(deserialize_tag(reader))};
```

LinZhihao-723 (changes requested):

components/core/src/clp/ffi/ir_stream/decoding_methods.hpp
LinZhihao-723, line 1: Instead of reusing the existing one, can we make a result-based version for `deserialize_timestamp` that returns `Result<ir::epoch_time_ms_t>`?

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,250 @@
...
+             }},
+            {false,
+             SchemaTree::NodeLocator{SchemaTree::cRootId, cMessageKey, SchemaTree::Node::Type::Str}}
+    };
+}
+}  // namespace
+
+template <ir::EncodedVariableTypeReq encoded_variable_t>
+auto UnstructuredIrDeserializerImpl<encoded_variable_t>::create(
+        EncodingType encoding_type,
```
LinZhihao-723: This param is redundant since the encoding type is already indicated by the template parameter, right? Let's remove it.

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,250 @@
...
+) -> ystdlib::error_handling::Result<std::unique_ptr<UnstructuredIrDeserializerImpl>> {
+    auto initial_insertions{create_schema_insertions()};
+
+    if constexpr (std::is_same_v<encoded_variable_t, ir::eight_byte_encoded_variable_t>) {
+        if (EncodingType::EightByte != encoding_type) {
+            return std::errc::protocol_error;
+        }
+        return std::unique_ptr<UnstructuredIrDeserializerImpl>{
+                new UnstructuredIrDeserializerImpl{std::move(initial_insertions)}
+        };
```
LinZhihao-723: Use make unique instead. You should almost never need to use `new` explicitly.
```suggestion
        return std::make_unique<UnstructuredIrDeserializerImpl>(
                UnstructuredIrDeserializerImpl{std::move(initial_insertions)}
        );
```

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,250 @@
...
+    } else {  // std::is_same_v<encoded_variable_t, four_byte_encoded_variable_t>
+        if (EncodingType::FourByte != encoding_type) {
+            return std::errc::protocol_error;
+        }
+        auto const ref_timestamp_iter{metadata.find(cProtocol::Metadata::ReferenceTimestampKey)};
+        if (metadata.end() == ref_timestamp_iter || false == ref_timestamp_iter->is_string()) {
+            return std::errc::protocol_error;
```
LinZhihao-723: We should use our own error code, not std::errc, right?

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,250 @@
...
+            this->resolve_required_node_ids(auto_gen_keys_schema_tree, user_gen_keys_schema_tree)
+    )};
+
+    KeyValuePairLogEvent::NodeIdValuePairs auto_gen_pairs;
+    auto_gen_pairs.emplace(
+            timestamp_node_id,
+            Value{static_cast<value_int_t>(m_previous_timestamp)}
+    );
+    KeyValuePairLogEvent::NodeIdValuePairs user_gen_pairs;
+    user_gen_pairs.emplace(message_node_id, Value{std::move(encoded_text_ast)});
```
LinZhihao-723:
```suggestion
    KeyValuePairLogEvent::NodeIdValuePairs auto_gen_pairs{
            {timestamp_node_id, Value{static_cast<value_int_t>(m_previous_timestamp)}}
    };
    KeyValuePairLogEvent::NodeIdValuePairs user_gen_pairs{
            {message_node_id, Value{std::move(encoded_text_ast)}}
    };
```

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,250 @@
...
+auto
+UnstructuredIrDeserializerImpl<encoded_variable_t>::deserialize_ir_unit_schema_tree_node_insertion(
+        [[maybe_unused]] ReaderInterface& reader,
+        [[maybe_unused]] encoded_tag_t tag,
+        std::string& key_name
+) -> ystdlib::error_handling::Result<std::pair<bool, SchemaTree::NodeLocator>> {
+    auto const [is_auto_gen, node_locator]{m_pending_schema_insertions.back()};
+    m_pending_schema_insertions.pop_back();
+
+    key_name.clear();
```
LinZhihao-723: I just realized this is confusing. The original purpose of passing `key_name` is to provide a buffer for key name deserialization. In our case, it is not used. I think it might be worth renaming the interface's `key_name` to `key_name_buffer` and indicating that it might not be used, depending on the implementation.

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    auto const message_node_id{user_gen_keys_schema_tree->try_get_node_id(
+            SchemaTree::NodeLocator{SchemaTree::cRootId, cMessageKey, SchemaTree::Node::Type::Str}
+    )};
+    auto const timestamp_node_id{auto_gen_keys_schema_tree->try_get_node_id(
+            SchemaTree::NodeLocator{SchemaTree::cRootId, cTimestampKey, SchemaTree::Node::Type::Int}
+    )};
+    if (false == message_node_id.has_value() || false == timestamp_node_id.has_value()) {
+        return std::errc::protocol_error;
```
LinZhihao-723: Again, we should use our own error code.

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,250 @@
...
+    if (m_optional_message_node_id.has_value() && m_optional_timestamp_node_id.has_value()) {
+        return {m_optional_message_node_id.value(), m_optional_timestamp_node_id.value()};
+    }
+
+    auto const message_node_id{user_gen_keys_schema_tree->try_get_node_id(
+            SchemaTree::NodeLocator{SchemaTree::cRootId, cMessageKey, SchemaTree::Node::Type::Str}
+    )};
+    auto const timestamp_node_id{auto_gen_keys_schema_tree->try_get_node_id(
+            SchemaTree::NodeLocator{SchemaTree::cRootId, cTimestampKey, SchemaTree::Node::Type::Int}
+    )};
```
LinZhihao-723: These two locators appear twice in this file, how about drop the constexpr key name and make the locators themselves constexpr?

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+    template <
+            typename T = encoded_variable_t,
+            std::enable_if_t<std::is_same_v<T, ir::four_byte_encoded_variable_t>, int> = 0
+    >
+    UnstructuredIrDeserializerImpl(
+            std::vector<std::pair<bool, SchemaTree::NodeLocator>> initial_insertions,
+            ir::epoch_time_ms_t reference_timestamp
+    );
```
LinZhihao-723: Sorry for the confusion in previous, I think this should be the right way to do in C++20 (and you should drop the implementation in cpp file):
```suggestion
    explicit UnstructuredIrDeserializerImpl(
            std::vector<std::pair<bool, SchemaTree::NodeLocator>> initial_insertions
    )
    requires std::is_same_v<encoded_variable_t, ir::eight_byte_encoded_variable_t>
            : m_pending_schema_insertions{std::move(initial_insertions)} {}

    UnstructuredIrDeserializerImpl(
            std::vector<std::pair<bool, SchemaTree::NodeLocator>> initial_insertions,
            ir::epoch_time_ms_t reference_timestamp
    )
    requires std::is_same_v<encoded_variable_t, ir::four_byte_encoded_variable_t>
            : m_pending_schema_insertions{std::move(initial_insertions)},
              m_previous_timestamp{reference_timestamp} {}
```

LinZhihao-723 (changes requested):
Reviewed the details of unstructured deserializer impl.

components/core/src/clp/ffi/ir_stream/decoding_methods.hpp
```diff
@@ -96,6 +96,23 @@ IRErrorCode get_encoding_type(ReaderInterface& reader, bool& is_four_bytes_encod
...
+/**
+ * Deserializes a timestamp from the given reader.
+ * @tparam encoded_variable_t Type of the encoded variable
+ * @param reader
+ * @param encoded_tag
+ * @return A result containing the timestamp delta if encoded_variable_t ==
+ * four_byte_encoded_variable_t or the actual timestamp if encoded_variable_t ==
+ * eight_byte_encoded_variable_t on success, or an error code indicating the failure:
+ * - IrDeserializationErrorEnum::InvalidTag if the tag does not correspond to a timestamp in this
+ *   encoding.
```
LinZhihao-723:
```suggestion
 * @return A result containing the timestamp delta if `encoded_variable_t` is
 * `four_byte_encoded_variable_t` or the actual timestamp if `encoded_variable_t` is
 * `eight_byte_encoded_variable_t` on success, or an error code indicating the failure:
 * - IrDeserializationErrorEnum::InvalidTag if the tag does not correspond to a timestamp encoding.
```

components/core/src/clp/ffi/ir_stream/decoding_methods.hpp
```diff
@@ -96,6 +96,23 @@ IRErrorCode get_encoding_type(ReaderInterface& reader, bool& is_four_bytes_encod
...
+ * - IrDeserializationErrorEnum::IncompleteStream if reader doesn't contain enough data to
+ *   deserialize.
+ */
+template <ir::EncodedVariableTypeReq encoded_variable_t>
+[[nodiscard]] auto deserialize_timestamp(ReaderInterface& reader, encoded_tag_t encoded_tag)
```
LinZhihao-723: How about `deserialize_timestamp_or_timestamp_delta`?

components/core/src/clp/ffi/ir_stream/decoding_methods.cpp
```diff
@@ -628,6 +628,21 @@ auto deserialize_tag(ReaderInterface& reader) -> ystdlib::error_handling::Result
...
+    epoch_time_ms_t ts{};
+    if (auto const error_code = deserialize_timestamp<encoded_variable_t>(reader, encoded_tag, ts);
+        IRErrorCode_Success != error_code)
+    {
+        if (IRErrorCode_Corrupted_IR == error_code) {
+            return IrDeserializationError{IrDeserializationErrorEnum::InvalidTag};
+        }
+        return IrDeserializationError{IrDeserializationErrorEnum::IncompleteStream};
+    }
+    return ts;
```
LinZhihao-723:
```suggestion
    if constexpr (is_same_v<encoded_variable_t, eight_byte_encoded_variable_t>) {
        if (cProtocol::Payload::TimestampVal != encoded_tag) {
            return IrDeserializationError{IrDeserializationErrorEnum::InvalidTag};
        }
        return deserialize_int<epoch_time_ms_t>(reader);
    } else {
        if (cProtocol::Payload::TimestampDeltaByte == encoded_tag) {
            return static_cast<epoch_time_ms_t>(
                YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int<int8_t>(reader))
            );
        }
        if (cProtocol::Payload::TimestampDeltaShort == encoded_tag) {
            return static_cast<epoch_time_ms_t>(
                YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int<int16_t>(reader))
            );
        }
        if (cProtocol::Payload::TimestampDeltaInt == encoded_tag) {
            return static_cast<epoch_time_ms_t>(
                YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int<int32_t>(reader))
            );
        }
        if (cProtocol::Payload::TimestampDeltaLong == encoded_tag) {
            return static_cast<epoch_time_ms_t>(
                YSTDLIB_ERROR_HANDLING_TRYX(deserialize_int<int64_t>(reader))
            );
        }
        return IrDeserializationError{IrDeserializationErrorEnum::InvalidTag};
    }
```
I think we should just rewrite it using the latest helpers. We should do a clean-up to remove all old/outdated helpers.

LinZhihao-723: Don't forget to update the docstring since we're now forwarding the error code from `deserialize_int` instead.

components/core/src/clp/ffi/ir_stream/DeserializerImpl.hpp
```diff
@@ -16,8 +16,8 @@
 
 namespace clp::ffi::ir_stream {
 /**
- * Virtual base class defining methods for deserializing IR units and log events called by
- * `Deserializer`.
+ * Virtual base class defining methods for deserializing IR units and log events
+ * called by `Deserializer`.
```
LinZhihao-723: This change should be reverted (might be introduced by IDE auto-formatter).
Please check the diff everytime you make a commit to make sure all changes are intentional.

components/core/src/clp/ffi/ir_stream/DeserializerImpl.hpp
```diff
@@ -38,8 +38,8 @@ class DeserializerImpl {
     /**
      * Deserializes a UTC offset change IR unit from the given reader.
      * @param reader
-     * @return A result containing the deserialized UTC offset on success, or an error code
-     * indicating the failure:
+     * @return A result containing the deserialized UTC offset on success, or an
+     * error code indicating the failure:
```
LinZhihao-723: Revert.

components/core/src/clp/ffi/ir_stream/DeserializerImpl.hpp
```diff
@@ -48,7 +48,8 @@ class DeserializerImpl {
     /**
      * Deserializes the type of the next IR unit from the given reader.
      * @param reader
-     * @return A result containing a pair on success, or an error code indicating the failure:
+     * @return A result containing a pair on success, or an error code indicating
+     * the failure:
```
LinZhihao-723: Revert.

components/core/src/clp/ffi/ir_stream/DeserializerImpl.hpp
```diff
@@ -65,8 +66,8 @@ class DeserializerImpl {
      * @param auto_gen_keys_schema_tree
      * @param user_gen_keys_schema_tree
      * @param utc_offset
-     * @return A result containing the deserialized KV pair log event on success, or an error code
-     * indicating the failure defined by the derived class.
+     * @return A result containing the deserialized KV pair log event on success,
+     * or an error code indicating the failure defined by the derived class.
```
LinZhihao-723: Revert.

components/core/src/clp/ffi/ir_stream/DeserializerImpl.hpp
```diff
@@ -81,8 +82,10 @@ class DeserializerImpl {
      * Deserializes a schema tree node insertion IR unit from the given reader.
      * @param reader
      * @param tag
-     * @param key_name Returns the inserted node's key name.
-     * @return A result containing a pair on success, or an error code indicating the failure:
+     * @param key_name_buffer Returns the inserted node's key name. May be empty
+     * if the implementation does not read a key from the stream.
+     * @return A result containing a pair on success, or an error code indicating
+     * the failure:
```
LinZhihao-723:
```suggestion
     * @param key_name_buffer Returns the inserted node's key name. May be empty if the
     * implementation does not need to read the key from the stream.
     * @return A result containing a pair on success, or an error code indicating the failure:
```

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,197 @@
...
+#include "decoding_methods.hpp"
+#include "IrDeserializationError.hpp"
+#include "IrUnitType.hpp"
+#include "protocol_constants.hpp"
+
+namespace clp::ffi::ir_stream {
+namespace {
+constexpr SchemaTree::NodeLocator cRootTimestampLocator{
+        SchemaTree::cRootId,
+        std::string_view{"timestamp", sizeof("timestamp") - 1},
```
LinZhihao-723:
```suggestion
        std::string_view{"timestamp"},
```
You shouldn't need the size indicator. It's not a good practice to use `sizeof` on top of a string literal btw.

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,197 @@
...
+        SchemaTree::Node::Type::Int
+};
+constexpr SchemaTree::NodeLocator cRootMessageLocator{
+        SchemaTree::cRootId,
+        std::string_view{"message", sizeof("message") - 1},
```
LinZhihao-723:
```suggestion
        std::string_view{"message"},
```
Same.

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,197 @@
...
+    KeyValuePairLogEvent::NodeIdValuePairs user_gen_pairs;
+    user_gen_pairs.emplace(message_node_id, Value{std::move(encoded_text_ast)});
+
+    return KeyValuePairLogEvent::create(
+            auto_gen_keys_schema_tree,
+            user_gen_keys_schema_tree,
+            std::move(auto_gen_pairs),
+            std::move(user_gen_pairs),
+            utc_offset
+    );
```
LinZhihao-723:
```suggestion
    return KeyValuePairLogEvent::create(
            auto_gen_keys_schema_tree,
            user_gen_keys_schema_tree,
            KeyValuePairLogEvent::NodeIdValuePairs{
                    {timestamp_node_id, Value{static_cast<value_int_t>(absolute_timestamp)}}
            },
            KeyValuePairLogEvent::NodeIdValuePairs{
                    {message_node_id, Value{std::move(encoded_text_ast)}}
            },
            utc_offset
    );
```
- You forgot to update the eight-byte-encoded implementation.
- This is an updated version. I think it's the most efficient way to avoid an explicit move since in C++ move is actually not 0-cost (unlike Rust).

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,197 @@
...
+UnstructuredIrDeserializerImpl<encoded_variable_t>::deserialize_ir_unit_schema_tree_node_insertion(
+        [[maybe_unused]] ReaderInterface& reader,
+        [[maybe_unused]] encoded_tag_t tag,
+        std::string& key_name_buffer
+) -> ystdlib::error_handling::Result<std::pair<bool, SchemaTree::NodeLocator>> {
+    auto const [is_auto_gen, node_locator]{m_pending_schema_insertions.back()};
+    m_pending_schema_insertions.pop_back();
+
+    key_name_buffer.clear();
+    return {is_auto_gen, node_locator};
```
LinZhihao-723:
```suggestion
    key_name_buffer.clear();
    auto const pending_schema_insertion{m_pending_schema_insertions.back()};
    m_pending_schema_insertions.pop_back();
    return pending_schema_insertion;
```
You should no longer need the pair dispatching.

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.cpp
```diff
@@ -0,0 +1,197 @@
...
+        return IrDeserializationError{IrDeserializationErrorEnum::MissingRequiredSchemaNodes};
+    }
+
+    m_optional_message_node_id.emplace(message_node_id.value());
+    m_optional_timestamp_node_id.emplace(timestamp_node_id.value());
+    return {message_node_id.value(), timestamp_node_id.value()};
+}
+
+// Explicitly declare specializations to avoid having to validate that the
+// template parameters are supported
```
LinZhihao-723:
```suggestion
// Explicitly declare specializations to avoid having to validate that the template parameters are
// supported.
```

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.hpp
```diff
@@ -0,0 +1,146 @@
...
+#include "IrUnitType.hpp"
+
+namespace clp::ffi::ir_stream {
+/**
+ * Deserializer implementation for backward-compatible unstructured IR streams.
+ *
+ * Simulates KV-IR deserialization behavior so `Deserializer` can reuse the same
+ * flow across formats. It emits synthetic schema tree node insertion IR units
+ * for `message` and `timestamp` before reading stream tags and converts
+ * unstructured log events into `KeyValuePairLogEvent` values.
```
LinZhihao-723:
```suggestion
 * Simulates KV-IR deserialization behavior so `Deserializer` can reuse the same flow across
 * formats. It emits synthetic schema tree node insertion IR units for `message` and `timestamp`
 * before reading stream tags and converts unstructured log events into `KeyValuePairLogEvent`
 * values.
```

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.hpp
```diff
@@ -0,0 +1,146 @@
...
+public:
+    // Factory function
+    /**
+     * @param metadata
+     * @return A result containing the deserializer on success, or an error code
+     * indicating the failure:
+     * - IrDeserializationErrorEnum::InvalidReferenceTimestampMetadata if the
+     * reference timestamp is missing or not a string (four-byte only).
+     * - IrDeserializationErrorEnum::InvalidReferenceTimestampValue if the
+     * reference timestamp string is not an integer (four-byte only).
```
LinZhihao-723:
```suggestion
     * - IrDeserializationErrorEnum::InvalidReferenceTimestampMetadata if the reference timestamp is
     *   missing or not a string (four-byte only).
     * - IrDeserializationErrorEnum::InvalidReferenceTimestampValue if the reference timestamp
     *   string is not an integer (four-byte only).
```

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.hpp
```diff
@@ -0,0 +1,146 @@
...
+     * The possible error codes:
+     * - Forwards `deserialize_tag`'s return value on failure.
+     */
+    [[nodiscard]] auto get_next_ir_unit_type(ReaderInterface& reader)
+            -> ystdlib::error_handling::Result<std::pair<IrUnitType, encoded_tag_t>> override;
+
+    /**
+     * The possible error codes:
+     * - IrDeserializationErrorEnum::MissingRequiredSchemaNodes if required
+     *   schema nodes are missing.
```
LinZhihao-723:
```suggestion
     * - IrDeserializationErrorEnum::MissingRequiredSchemaNodes if required schema nodes are
     *   missing.
```

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.hpp
```diff
@@ -0,0 +1,146 @@
...
+     * - Forwards `deserialize_encoded_text_ast`'s return value on failure.
+     * - Forwards `deserialize_tag`'s return value on failure.
+     * - Forwards `deserialize_timestamp`'s return value on failure.
+     * - Forwards `KeyValuePairLogEvent::create`'s return value on failure.
```
LinZhihao-723:
```suggestion
     * - Forwards `resolve_required_node_ids`'s return value on failure.
```

components/core/src/clp/ffi/ir_stream/UnstructuredIrDeserializerImpl.hpp
```diff
@@ -0,0 +1,146 @@
...
+     * `timestamp` is resolved from the auto-generated schema tree.
+     * @param auto_gen_keys_schema_tree
+     * @param user_gen_keys_schema_tree
+     * @return A result containing a pair on success, or an error code indicating
+     * the failure:
+     * - The pair:
+     *   - `message` node ID
+     *   - `timestamp` node ID
+     * - IrDeserializationErrorEnum::MissingRequiredSchemaNodes if either required
+     * schema node is missing.
```
LinZhihao-723:
```suggestion
     * Resolves the node IDs for `message` and `timestamp` in the schema tree. If the node IDs are
     * already resolved, returns them immediately.
     *
     * NOTE: `message` is resolved from the user-generated schema tree, while `timestamp` is
     * resolved from the auto-generated schema tree.
     *
     * @param auto_gen_keys_schema_tree
     * @param user_gen_keys_schema_tree
     * @return A result containing a pair on success, or an error code indicating the failure:
     * - The pair:
     *   - `message` node ID
     *   - `timestamp` node ID
     * - IrDeserializationErrorEnum::MissingRequiredSchemaNodes if either required schema node is
     *   missing.
```

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
LinZhihao-723, line 248: Can we drop the use of `std::errc::xxx` with our own error code in this method?

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -261,21 +259,40 @@ auto Deserializer<IrUnitHandlerType, QueryHandlerType>::create_generic(
     if (metadata_json.end() == version_iter || false == version_iter->is_string()) {
         return std::errc::protocol_error;
     }
-    auto const version = version_iter->get_ref<nlohmann::json::string_t&>();
-    if (ffi::ir_stream::IRProtocolErrorCode::Supported
-        != ffi::ir_stream::validate_protocol_version(version))
+    auto const version{version_iter->get_ref<nlohmann::json::string_t const&>()};
+    auto const protocol_version_status{ffi::ir_stream::validate_protocol_version(version)};
+    if (ffi::ir_stream::IRProtocolErrorCode::Supported != protocol_version_status
+        && ffi::ir_stream::IRProtocolErrorCode::BackwardCompatible != protocol_version_status)
```
LinZhihao-723:
```suggestion
    if (IRProtocolErrorCode::Supported != protocol_version_status
        && IRProtocolErrorCode::BackwardCompatible != protocol_version_status)
```

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -261,21 +259,40 @@ auto Deserializer<IrUnitHandlerType, QueryHandlerType>::create_generic(
...
-        && false == metadata_json.at(cProtocol::Metadata::UserDefinedMetadataKey).is_object())
-    {
-        return std::errc::protocol_not_supported;
+    std::unique_ptr<DeserializerImpl> deserializer_impl;
+    if (ffi::ir_stream::IRProtocolErrorCode::Supported == protocol_version_status) {
+        if (metadata_json.contains(cProtocol::Metadata::UserDefinedMetadataKey)
+            && false == metadata_json.at(cProtocol::Metadata::UserDefinedMetadataKey).is_object())
+        {
+            return std::errc::protocol_not_supported;
+        }
```
LinZhihao-723: This is the metadata check. It should be supported for both kv-ir and the old IR format. Let's move it outside of this if-else cluster.

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -261,21 +259,40 @@ auto Deserializer<IrUnitHandlerType, QueryHandlerType>::create_generic(
...
     {
         return std::errc::protocol_not_supported;
     }
 
-    if (metadata_json.contains(cProtocol::Metadata::UserDefinedMetadataKey)
-        && false == metadata_json.at(cProtocol::Metadata::UserDefinedMetadataKey).is_object())
-    {
-        return std::errc::protocol_not_supported;
+    std::unique_ptr<DeserializerImpl> deserializer_impl;
+    if (ffi::ir_stream::IRProtocolErrorCode::Supported == protocol_version_status) {
```
LinZhihao-723:
```suggestion
    if (IRProtocolErrorCode::Supported == protocol_version_status) {
```

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -329,10 +346,13 @@ auto Deserializer<IrUnitHandler, QueryHandlerType>::deserialize_next_ir_unit(
         }
 
         case IrUnitType::SchemaTreeNodeInsertion: {
-            std::string key_name;
+            std::string key_name_buffer;
```
LinZhihao-723, line 356: I think this buffer should be made as a member variable of `Deserializer` to avoid memory reallocation. Can you create an issue to track it? @coderabbitai

components/core/src/clp/ffi/ir_stream/Deserializer.hpp
```diff
@@ -261,21 +259,40 @@ auto Deserializer<IrUnitHandlerType, QueryHandlerType>::create_generic(
...
+        if (metadata_json.contains(cProtocol::Metadata::UserDefinedMetadataKey)
+            && false == metadata_json.at(cProtocol::Metadata::UserDefinedMetadataKey).is_object())
+        {
+            return std::errc::protocol_not_supported;
+        }
+        deserializer_impl = std::make_unique<KvIrDeserializerImpl>();
+    } else {
+        if (EncodingType::FourByte == encoding_type) {
+            deserializer_impl = YSTDLIB_ERROR_HANDLING_TRYX(
+                    UnstructuredIrDeserializerImpl<ir::four_byte_encoded_variable_t>::create(
```
LinZhihao-723, line 288: Need to update the docstring to reflect that it forwards `UnstructuredIrDeserializerImpl::create`'s return values on failure.

LinZhihao-723 (changes requested):
Functionality wise looks good to me. Validated that:
- Unstructured IR logs can be ingested into clp-s using `timestamp-key="@timestamp"`.
- Unstructured IR logs can be searched by clp-s cli.

LinZhihao-723 (approved):
Verified that the latest code works well in `clp-s` in terms of:
* Compress unstructured IR into an archive
* Search the compressed archive with timestamp filtering
* Search the IR file directly

PR #2229, opened 2026-04-28
refactor(clp-s)!: Rework command line argument parsing for output destinations to improve extensibility.

gibber9809:
# Description

This PR reworks how we parse arguments for output destinations during search in order to make it easy to support having multiple output destinations in a follow-up PR.

Each output destination is treated like a subcommand, e.g. `<output-destination-type> <output-destination-arguments>`. Previously, when supporting only one output destination, we would simply collect the unrecognized options after parsing the main search options and hand those options to an options_description specific to the relevant output destination. With multiple output destinations though we need to do a little bit of extra work to figure out which unrecognized options are related to each subcommand.

This PR solves that problem by parsing the unrecognized options left to right, using both the known output destination names and the `options_description` object for each subcommand to attempt to find the options for each output destination. After this step the options can be parsed with each destination-specific `options_description` like normal. We can't simply iteratively use each `options_description` with `allow_unregistered` to parse the full list of options, because different output destinations share options with the same name (e.g. `--host` is used both by `reducer` and `network`), so some combinations of output destinations could result in throwing errors because of repeated options. As well, we can't simply rely on splitting up the options for each subcommand based solely on the ouptut destination names, because the output destination names can also appear as arguments (e.g. in the package, because of the way our DNS is set up, a reducer query will look like `clp-s s <archive> <query> reducer --host reducer --port <port> --job-id <id> --count`.

This PR also refactors the `CommandLineArguments` class to group together options for each output destination into structs.

This PR includes a small functional change, in that aggregation options are now part of the reducer output destination options. This changes how aggregation options must be specified on the command line (i.e. they explicitly need to be part of the reducer output destination arguments), but doesn't represent any change in functionality since aggregation options are only allowed for the reducer output destination anyway.

# Checklist

  breaking change.

# Validation performed

* Confirmed that package search through webui and cli work as expected
* Manually validated that options are split as expected when passing multiple output destinations  (before the command is rejected for having multiple output destinations)

## Summary by CodeRabbit

* **Bug Fixes**
  * Stronger validation and clearer errors for output-handler arguments; exactly one handler must be specified and stdout is treated as a no-options handler.
  * Misuse of reducer/aggregation options is rejected earlier with explicit validation.

* **Refactor**
  * Command-line parsing consolidated to per-handler option groups with typed handler configurations.
  * Aggregation flags are now scoped to reducer handlers and appended after reducer parameters in constructed commands.

* **Documentation**
  * Help output for available handlers reformatted for clarity.

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/clp/pull/2229?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -113,6 +157,99 @@ void validate_archive_paths(
...
+            throw std::invalid_argument(fmt::format("unrecognized option \"{}\"", option));
+        }
+
+        current_subcommand_arguments.value().emplace_back(option);
+        remaining_tokens_in_current_argument
+                = get_max_tokens_for_option(current_subcommand_options_description, option)
+                          .value_or(0);
+    }
+    collect_subcommand();
+    return collected_subcommands;
```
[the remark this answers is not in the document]
gibber9809: Yeah the last bit is why I'm fine with this edge case -- this is making a best-effort attempt to split up the arguments for different subcommands. If we fail to split them up correctly in this scenario, it's because the user has passed some invalid argument, which we'll catch when trying to parse the subcommands.

components/core/src/clp_s/clp-s.cpp
LinZhihao-723, line 1: From the API design perspective (to make it future-proof), I think it might make sense if we:
* Make the output handler config a variant of the four different types.
* The master config only contains one variant variable, and it must be one of the four.
* Replace the switch statement with a variant visitor.

components/core/src/clp_s/CommandLineArguments.cpp
LinZhihao-723, line 1: Didn't review the coding-style thing... the functionality looks good to me.

LinZhihao-723 (changes requested):
I don't have comments on the parsing logic. I think how we store the parsed options is the most important decision to make in this PR, since it affects how C-APIs will be written.
One way is to do the variant thing I mentioned in another comment.
The other way is we expose different entry points for search, each takes a particular config option type and initializes the output handler accordingly.
I think for now, it makes more sense to use the variant solution as it keeps the C++ version cleaner: one config variant variable instead of one variant enum + four config objects stored inside the master config.

components/core/src/clp_s/clp-s.cpp
gibber9809: I think this generally makes sense to me, and it should be pretty straightforward to change this to a vector of variants in the follow-up when we start supporting multiple output destinations at a time.

gibber9809: Or alternatively add another option to this variant that represents a vector of output handlers I guess.

LinZhihao-723 (approved):
lgtm.

PR #2231, opened 2026-04-29
feat(integration-tests): Add package start/stop action and verification infra and shared fixtures for package tests.

quinntaylormitchell:
# Description

This PR introduces session- and module-scoped fixtures for spinning up a `ClpPackage` instance for a given mode, along with start/stop helpers and verification utilities.

Note: The `clp_package` fixture combines the former `fixt_package_test_config` and `fixt_package_instance` fixtures.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **Tests**
  * Made centralized integration-test fixtures available globally for all integration runs.
  * Added session- and module-scoped fixtures to manage per-mode test instances with guaranteed startup/teardown verification.

* **Refactor**
  * Marked older fixture modules as deprecated in favour of the centralized fixtures.
  * Added start/stop/verification utilities, structured verification results, and a safe YAML writer for test configs.


integration-tests/tests/utils/utils.py
```diff
@@ -155,6 +155,20 @@ def validate_file_exists(file_path: Path) -> None:
...
 
 
+def write_dict_to_yaml(payload: dict[str, Any], file_path: Path) -> None:
+    """
+    Write a dictionary to a YAML file.
+
+    :param payload: The dictionary to write.
+    :param file_path: The path to the YAML file.
+    """
+    file_path.parent.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
quinntaylormitchell: you told me to add that line in the first place

integration-tests/tests/package_tests/fixtures.py
```diff
@@ -0,0 +1,93 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope="session")
+def clp_package_test_path_config() -> ClpPackageTestPathConfig:
+    """Provides paths relevant to all package integration tests."""
+    return ClpPackageTestPathConfig(
+        clp_build_dir=resolve_path_env_var("CLP_BUILD_DIR"),
+        integration_tests_project_root=resolve_path_env_var("INTEGRATION_TESTS_PROJECT_ROOT"),
```
Bill-hbrhbr, line 34: These two fields belong to the base `IntegrationTestPathConfig`. Do you think we should resolve those env vars once in an `IntegrationTestPathConfig` fixture, then build category-specific path config fixtures from that? 

Or we can resolve all env vars directly here.

integration-tests/tests/package_tests/fixtures.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # Assign ports to the `ClpConfig` pydantic object.
+    logger.info("Assigning ports to the '%s' package.", mode_name)
+    base_port_string = request.config.getoption("--base-port")
+    try:
+        base_port = int(base_port_string)
+    except ValueError as err:
+        err_msg = f"Invalid value '{base_port_string}' for '--base-port'; expected an integer."
+        raise ValueError(err_msg) from err
+    assign_ports_from_base(base_port, clp_config)
```
Bill-hbrhbr, line 73: Should we move `port_utils.py` into `package_tests`, given that it is currently tightly coupled to the `clp_config` object?

integration-tests/tests/package_tests/fixtures.py
```diff
@@ -0,0 +1,93 @@
...
+        temp_config_file_path,
+    )
+
+    # Construct `ClpPackage` object.
+    clp_package = ClpPackage(
+        path_config=clp_package_test_path_config,
+        mode_name=mode_name,
+        clp_config=clp_config,
+        component_list=component_list,
+    )
```
Bill-hbrhbr:
```suggestion
    # Construct `ClpPackage` object.
    clp_package = ClpPackage(
        path_config=clp_package_test_path_config,
        mode_name=mode_name,
        clp_config=clp_config,
        component_list=component_list,
    )

    # Write the temporary config file.
    logger.info("Writing the temporary config file for the '%s' package.", mode_name)
    write_dict_to_yaml(
        clp_config.dump_to_primitive_dict(),  # type: ignore[no-untyped-call]
        clp_package.temp_config_file_path,
    )
```
I don't think `ClpPackage` initializes anything using `temp_config_file_path` property. Can we move the yaml dumping after constructing the `ClpPackage` object so that we can remove duplicate logic in constructing the  `temp_config_file_path`?

integration-tests/tests/utils/classes.py
```diff
@@ -131,13 +132,24 @@ def logs_path(self) -> Path:
...
+
 @dataclass
 class ExternalAction:
     """Metadata for an external action executed during an integration test."""
 
     #: Command to pass to `subprocess.run()`.
     cmd: list[str]
 
+    #: Optional structured arguments for verification purposes. Not used by `ExternalAction` itself.
+    args: CmdArgs | None = None
```
Bill-hbrhbr: is there a way that we accept mutually exclusive constructor input?
- if `args` is not None, `cmd` is automatically `args.to_cmd()`
- if `args` is None, `cmd` list is required.

```
    @classmethod
    def from_args(cls, args: CmdArgs) -> "ExternalAction":
        return cls(cmd=args.to_cmd(), args=args)

    @classmethod
    def from_cmd(cls, cmd: list[str]) -> "ExternalAction":
        return cls(cmd=cmd, args=None)
```

Bill-hbrhbr: This way we don't have to repeat `cmd=args.to_cmd(), args=args` everywhere

integration-tests/tests/package_tests/utils/start_stop.py
```diff
@@ -0,0 +1,184 @@
...
+    args: StartStopArgs = _construct_start_clp_args(clp_package)
+    return ExternalAction(cmd=args.to_cmd(), args=args)
+
+
+def _construct_start_clp_args(clp_package: ClpPackage) -> StartStopArgs:
+    """Construct the `StartStopArgs` object that will start the package."""
+    path_config = clp_package.path_config
+    return StartStopArgs(
+        script_path=path_config.start_clp_path, config=clp_package.temp_config_file_path
+    )
```
Bill-hbrhbr, line 41:
```suggestion
def start_clp_package(
    clp_package: ClpPackage,
) -> ExternalAction:
    """
    Starts the CLP package.

    :param clp_package:
    :return: The `ExternalAction` instance that starts the package.
    """
    logger.info("Starting up the '%s' package.", clp_package.mode_name)
    args = StartStopArgs(
        script_path=clp_package.path_config.start_clp_path,
        config=clp_package.temp_config_file_path,
    )
    return ExternalAction(cmd=args.to_cmd(), args=args)
```
How about we keep this simple and remove the helper?

integration-tests/tests/package_tests/utils/start_stop.py
```diff
@@ -0,0 +1,184 @@
...
+
+    :param clp_package:
+    :return: The `ExternalAction` instance that stops the package.
+    """
+    logger.info("Stopping the '%s' package.", clp_package.mode_name)
+    args: StartStopArgs = _construct_stop_clp_args(clp_package)
+    return ExternalAction(cmd=args.to_cmd(), args=args)
+
+
+def _construct_stop_clp_args(clp_package: ClpPackage) -> StartStopArgs:
```
Bill-hbrhbr: ditto. remove this helper since I feel like it only has one-time usage.

integration-tests/tests/package_tests/fixtures.py
```diff
@@ -0,0 +1,93 @@
...
+
+    try:
+        start_clp_action = start_clp_package(clp_package)
```
Bill-hbrhbr:
```suggestion
        start_clp_action: ExternalAction = start_clp_package(clp_package)
```
somehow I feel like specifying the type here makes things clearer

integration-tests/tests/package_tests/utils/start_stop.py
```diff
@@ -0,0 +1,184 @@
...
+) -> VerificationResult:
+    """
+    Verify the spindown of the CLP package.
+
+    :param stop_clp_action:
+    :param clp_package:
+    :return: A `VerificationResult` indicating the success or failure of the verification.
+    """
+    logger.info("Verifying the spindown of the '%s' package.", clp_package.mode_name)
+    if stop_clp_action.completed_proc.returncode != 0:
```
Bill-hbrhbr: this returncode check against 0 is what I feel like could be a part of the `ExternalAction`.

Also I'm wondering why we are using `VerificationResult` here instead of directly writing `assert`s or `pytest.fail()`.

Bill-hbrhbr (changes requested):

integration-tests/tests/package_tests/utils/start_stop.py
quinntaylormitchell: Fine by me!

integration-tests/tests/package_tests/utils/start_stop.py
quinntaylormitchell: Agreed

integration-tests/tests/package_tests/utils/start_stop.py
quinntaylormitchell: 1. I thought a lot about putting the `returncode != 0` check in `ExternalAction` but I eventually decided against it for a couple reasons:
    a. In the future, we don't always want to check that `returncode != 0`. As part of our testing, we may want to test failure behaviour, i.e., we may want to intentionally cause failure and ensure that certain things happen. If we bury the retcode check in the object, it'll be less flexible.
    b. Even when we do want to check the retcode, we won't always be checking for `0`; a great example of this is `grep`, which (iirc) returns `0` when it finds matches,  `1` when it doesn't find any matches, and `2` (I think) when there's an error.

2. I want to use `VerificationResult` because I want the verification functions to refrain from making a judgement about what a failure should mean for the flow of a specific test (for the reasons described in 1. a. above). For example, a test might want to ensure that compression of text logs fails for `clp-json` or something like that. `VerificationResult` allows the tests to decide their own flow instead of being pulled in a certain direction by the verification functions that they use.

integration-tests/tests/package_tests/fixtures.py
quinntaylormitchell: I'm not sure if I follow you; the `env` vars are only being resolved once per session because this is a `session`-scoped fixture

integration-tests/tests/package_tests/fixtures.py
quinntaylormitchell: Sure, I like that idea. This PR is long enough, so I think I'll put it up in a later PR.

integration-tests/tests/package_tests/fixtures.py
quinntaylormitchell: Agreed

integration-tests/tests/package_tests/fixtures.py
quinntaylormitchell: Fine with me

integration-tests/tests/utils/classes.py
quinntaylormitchell: I like that, makes it cleaner

integration-tests/tests/package_tests/fixtures.py
Bill-hbrhbr: I meant creating a fixture for `IntegrationTestPathConfig` and then
```
return ClpPackageTestPathConfig(
        clp_build_dir=base_path_config.clp_build_dir,
        integration_tests_project_root=base_path_config.integration_tests_project_root,
        clp_package_dir=resolve_path_env_var("CLP_PACKAGE_DIR"),
    )
```
but I guess this may be more complicated in some sense. It does remove the need to reference env vars `CLP_BUILD_DIR` and `INTEGRATION_TESTS_PROJECT_ROOT` multiple times.

integration-tests/tests/package_tests/fixtures.py
Bill-hbrhbr: @coderabbitai Open an issue to move the `port_utils.py` inside `package_tests`.

integration-tests/tests/package_tests/fixtures.py
Bill-hbrhbr: Actually I found out that you didn't remove `integration_test_path_config`. Maybe we should just pass the values inside this fixture:
- integration_test_path_config.clp_build_dir
- integration_test_path_config.integration_tests_project_root

integration-tests/tests/package_tests/utils/start_stop.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+    logger.info("Verifying the startup of the '%s' package.", clp_package.mode_name)
+
+    returncode_result = start_clp_action.verify_returncode()
+    if not returncode_result:
+        return returncode_result
+
+    package_running_err_msg = _validate_clp_package_running(clp_package)
+    if package_running_err_msg is None:
+        return start_clp_action.pass_verification()
```
Bill-hbrhbr, line 84: When reading this, I still feel like success is just success. The `ClpVerificationResult` returned by `start_clp_action.pass_verification()` has nothing to do with `start_clp_action`.

I agree that failure needs context for error logging and reporting.

I feel the need to ask again if there's a strong reason against doing this:
```suggestion
        return ClpVerificationResult(True)
```

integration-tests/tests/package_tests/utils/start_stop.py
```diff
@@ -0,0 +1,147 @@
...
+        return returncode_result
+
+    package_not_running_err_msg = _validate_clp_package_not_running(clp_package)
+    if package_not_running_err_msg is None:
+        return stop_clp_action.pass_verification()
+
+    return stop_clp_action.fail_verification(package_not_running_err_msg)
+
+
+def _validate_clp_package_running(clp_package: ClpPackage) -> str | None:
```
Bill-hbrhbr:
```suggestion
def _validate_clp_package_running(clp_package: ClpPackage) -> ClpVerificationResult:
```
The return type `str | None` screams "verification result" to me. It seems cleaner to return `ClpVerificationResult` directly instead of encoding success/failure through `None` semantics.

From another perspective, are these two helpers:
- `_validate_clp_package_running`
- `_validate_clp_package_not_running`

actually necessary? They appear to be single-use wrappers around fairly small checks, so inlining the logic at start/stop functions feels more natural.

integration-tests/tests/package_tests/fixtures.py
```diff
@@ -0,0 +1,94 @@
...
+        start_clp_action: ClpAction = start_clp_package(clp_package)
+        start_result = verify_start_clp_action(start_clp_action, clp_package)
+        assert start_result, start_result.failure_message
+        start_successful = True
+        yield clp_package
+    finally:
+        stop_clp_action: ClpAction = stop_clp_package(clp_package)
+        if start_successful:
+            stop_result = verify_stop_clp_action(stop_clp_action, clp_package)
+            assert stop_result, stop_result.failure_message
```
Bill-hbrhbr:
```suggestion
        stop_result = verify_stop_clp_action(stop_clp_action, clp_package)
        assert stop_result, stop_result.failure_message
```
Let's always verify this. I believe the CLP package stop script can handle a partially started package. We don't really want leftover Docker containers.

integration-tests/tests/package_tests/fixtures.py
```diff
@@ -0,0 +1,94 @@
...
+        clp_config.dump_to_primitive_dict(),  # type: ignore[no-untyped-call]
+        clp_package.temp_config_file_path,
+    )
+
+    start_successful = False
+    try:
+        start_clp_action: ClpAction = start_clp_package(clp_package)
+        start_result = verify_start_clp_action(start_clp_action, clp_package)
+        assert start_result, start_result.failure_message
+        start_successful = True
```
Bill-hbrhbr:
```suggestion
```

integration-tests/tests/package_tests/fixtures.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+    # Write the temporary config file.
+    logger.info("Writing the temporary config file for the '%s' package.", mode_name)
+    write_dict_to_yaml(
+        clp_config.dump_to_primitive_dict(),  # type: ignore[no-untyped-call]
+        clp_package.temp_config_file_path,
+    )
+
+    start_successful = False
```
Bill-hbrhbr:
```suggestion
```

integration-tests/tests/package_tests/fixtures.py
```diff
@@ -0,0 +1,94 @@
...
+        assert start_result, start_result.failure_message
+        start_successful = True
+        yield clp_package
+    finally:
+        stop_clp_action: ClpAction = stop_clp_package(clp_package)
+        if start_successful:
+            stop_result = verify_stop_clp_action(stop_clp_action, clp_package)
+            assert stop_result, stop_result.failure_message
+
+        clp_package.temp_config_file_path.unlink(missing_ok=True)
```
Bill-hbrhbr: Let's make sure this always run even if stop package fails.

integration-tests/tests/package_tests/clp_json/test_clp_json.py
```diff
@@ -84,23 +81,21 @@ def test_clp_json_compression_json_multifile(
...
-    validate_package_running(fixt_package_instance)
-
     # TODO: compress a dataset
 
     # TODO: check the correctness of the compression
 
     # TODO: search through that dataset and check the correctness of the search results.
 
-    assert True
+    assert clp_package
```
Bill-hbrhbr, line 98: The fixture should already be guaranteed to yield successfully once we reach this point. Writing `assert clp_package` feels a bit confusing because it suggests that `clp_package` may have meaningful `__bool__` overrides.

That said, this is probably not a big issue if the code is going to be replaced by the actual search implementation soon anyway.

integration-tests/tests/package_tests/utils/start_stop.py
```diff
@@ -0,0 +1,147 @@
...
+        config=clp_package.temp_config_file_path,
+    )
+    return ClpAction.from_args(args)
+
+
+def verify_start_clp_action(
+    start_clp_action: ClpAction, clp_package: ClpPackage
+) -> ClpVerificationResult:
+    """
+    Verify the startup of the CLP package.
```
Bill-hbrhbr: can you improve docstring by mentioning you moved components checking here?

integration-tests/tests/package_tests/utils/start_stop.py
```diff
@@ -0,0 +1,147 @@
...
+        return start_clp_action.pass_verification()
+
+    return start_clp_action.fail_verification(package_running_err_msg)
+
+
+def verify_stop_clp_action(
+    stop_clp_action: ClpAction, clp_package: ClpPackage
+) -> ClpVerificationResult:
+    """
+    Verify the spindown of the CLP package.
```
Bill-hbrhbr: can you improve docstring by mentioning you moved components checking here?

integration-tests/tests/package_tests/utils/start_stop.py
```diff
@@ -0,0 +1,147 @@
...
+
+    package_not_running_err_msg = _validate_clp_package_not_running(clp_package)
+    if package_not_running_err_msg is None:
+        return stop_clp_action.pass_verification()
+
+    return stop_clp_action.fail_verification(package_not_running_err_msg)
+
+
+def _validate_clp_package_running(clp_package: ClpPackage) -> str | None:
+    """:return: `None` on success; otherwise a string describing the failure."""
```
Bill-hbrhbr: This docstring is definitely lacking. I feel like let's just improve the docstring of `verify_start_clp_action` and inline this component checking utility.

integration-tests/tests/utils/classes.py
```diff
@@ -208,9 +208,13 @@ def _run_subprocess(self) -> subprocess.CompletedProcess[str]:
...
         except subprocess.TimeoutExpired:
-            pytest.fail(f"Subprocess '{exe_name}' timed out after {DEFAULT_CMD_TIMEOUT_SECONDS}s.")
+            err_msg = f"Subprocess '{exe_name}' timed out after {DEFAULT_CMD_TIMEOUT_SECONDS}s."
+            logger.exception(err_msg)
+            pytest.fail(err_msg)
         except OSError as e:
-            pytest.fail(f"Subprocess '{exe_name}' failed to start: {e}")
+            err_msg = f"Subprocess '{exe_name}' failed to start: {e}"
+            logger.exception(err_msg)
+            pytest.fail(err_msg)
```
Bill-hbrhbr, line 221: can you explain the logger.exception? for traceback?

Bill-hbrhbr (changes requested):

integration-tests/tests/package_tests/clp_json/test_clp_json.py
quinntaylormitchell: It will be replaced by the actual search implementation, not to worry. If the parameter doesn't get used in the test function, the linter gets mad.

integration-tests/tests/package_tests/utils/start_stop.py
quinntaylormitchell: >The ClpVerificationResult returned by start_clp_action.pass_verification() has nothing to do with start_clp_action.

Not sure what you mean by this; `pass_verification()` returns a passed `ClpVerificationResult`  for `start_clp_action`.

My "strong" reason against using `ClpVerificationResult(True)` is articulated in [this subcomment](https://github.com/y-scope/clp/pull/2242#discussion_r3255710172) from #2242; it's for the sake of symmetricity and also to avoid callers constructing `ClpVerificationResult` instances directly.

integration-tests/tests/package_tests/utils/start_stop.py
quinntaylormitchell: Yeah good point, I will inline them.

integration-tests/tests/package_tests/utils/start_stop.py
quinntaylormitchell: Agreed.

integration-tests/tests/package_tests/utils/start_stop.py
quinntaylormitchell: agreed

integration-tests/tests/package_tests/utils/start_stop.py
quinntaylormitchell: agreed

integration-tests/tests/utils/classes.py
quinntaylormitchell: If I don't use `exception` then the linter gets mad (because it's in an `except` statement)

integration-tests/tests/package_tests/fixtures.py
quinntaylormitchell: Agreed, good catch. I'll modify `clp_package_test_path_config()` fixture to clear the directories at the end of the session

integration-tests/tests/package_tests/fixtures.py
quinntaylormitchell: Ah ok good point; agreed.

(I was afraid that on startup failure, there would have been a possibility that the `instance-id` file may not have been created (depending on how the package startup failed), and if that file isn't there, then `verify_stop_clp_action()` will fail, BUT I looked more at `start_clp.py` and `stop_clp.py` and it looks like the instance-id file will get created during spindown if it wasn't created during startup (plus, if it doesn't get created during startup, then that means that no Docker containers will have been started, so we are all good))

integration-tests/tests/package_tests/fixtures.py
quinntaylormitchell: agreed

integration-tests/tests/package_tests/fixtures.py
quinntaylormitchell: agreed

integration-tests/tests/package_tests/utils/start_stop.py
Bill-hbrhbr: I get the symmetry part, and I also saw that Junhao agreed. Here’s my argument:

Successful verification is basically just success. There is not much `ExternalAction` specific information encoded in the passed case. You could even `return random_action.pass_verification()` and the code would still function identically.

If you look at how **BoostOutcome** handles error reporting:
https://github.com/y-scope/ystdlib-cpp/blob/c03806a35641add635cba4f0235591fa79cafdad/src/ystdlib/error_handling/Result.hpp#L30
The success case is not tied to any user-defined error category. It is effectively a constant owned by the **Boost** library itself.

Only when reporting errors that we involve user-defined systems:
- an error [enum](https://github.com/y-scope/clp/blob/85eaa70302a933dc71c3806c9975179df5486ce3/components/core/src/clp/ffi/ir_stream/IrSerializationError.hpp#L12)
- an error [message](https://github.com/y-scope/clp/blob/85eaa70302a933dc71c3806c9975179df5486ce3/components/core/src/clp/ffi/ir_stream/IrSerializationError.cpp#L17)

Another way to think about it is the Tolstoy line: “All happy families are alike; each unhappy family is unhappy in its own way.”

Bill-hbrhbr: But let's not make the change in this PR. Just need @junhaoliao to be aware

integration-tests/tests/utils/classes.py
Bill-hbrhbr: can you provide the error? I cannot seem to reproduce it with `task lint:check-py`.

integration-tests/tests/utils/classes.py
Bill-hbrhbr, line 196: Let's add return type and two pytest.fail exceptions here.

Bill-hbrhbr:
```
:return: The completed subprocess result, encapsulating the return code, stdout, and stderr.
:raises pytest.fail.Exception: If the subprocess times out or fails to start.
```

Bill-hbrhbr, line 174:
```suggestion
         """
         Execute the external action and log its output.

         :raises TypeError: If this abstract base class is instantiated directly.
         :raises ValueError: If the command list is empty.
         :raises: Propagates `_run_subprocess()`'s exceptions.
         """
```

Bill-hbrhbr (dismissed):
Minor docstring comments.

integration-tests/tests/utils/classes.py
quinntaylormitchell: when I run `task lint:check-py`:

```text
...

TRY400 Use `logging.exception` instead of `logging.error`
   --> tests/utils/classes.py:212:13
    |
210 |         except subprocess.TimeoutExpired:
211 |             err_msg = f"Subprocess '{exe_name}' timed out after {DEFAULT_CMD_TIMEOUT_SECONDS}s."
212 |             logger.error(err_msg)
    |             ^^^^^^^^^^^^^^^^^^^^^
213 |             pytest.fail(err_msg)
214 |         except OSError as e:
    |
help: Replace with `exception`
```

integration-tests/tests/utils/classes.py
quinntaylormitchell: I agree, but as none of the other `__post_init__` method docstrings use this verbosity, we should change all of them, and that's out of scope for this PR. I've made a note to put up another PR to fix this. (fwiw, there are some other problems with the docstrings in the testing suite as well, the biggest of them being that many of them are in imperative form and not present simple; that can be fixed in future PR too.)

quinntaylormitchell: agreed

integration-tests/tests/utils/classes.py
Bill-hbrhbr: ok I guess the question becomes why we need `logging.error/exception()` if we have `pytest.fail()`? Does the latter not print out the error message?

quinntaylormitchell: So that it shows up in the test log file and also is explained to the user via command line

integration-tests/tests/package_tests/utils/start_stop.py
junhaoliao: i spent an hour staring at this and still can't figure out why we might prefer one way over the other lol... which probably should not be the focus of this PR anyways

> The ClpVerificationResult returned by start_clp_action.pass_verification() has nothing to do with start_clp_action.

i disagree. that we are getting the set of desired running services should be a success metric of this `start_clp_action`?

> You could even return random_action.pass_verification() and the code would still function identically.

i agree. it doesn't seem we can easily enforce any test writers to use an accurate `*_action`'s pass_verification(), not to mention that sometimes a validation success is a result of multiple actions - then to which action shall we attribute the success? then i agree that does sound a valid argument to avoid couple a successful ClpVerificationResult with a specific action.

though today, we are not doing any extra checks here, and the successful validation result does seem a direct result of the start action as reasoned above? then let's keep things unchanged, as you guys proposed and agreed

integration-tests/tests/package_tests/fixtures.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+
+    try:
+        start_clp_action: ClpAction = start_clp_package(clp_package)
+        start_result = verify_start_clp_action(start_clp_action, clp_package)
+        assert start_result, start_result.failure_message
+        yield clp_package
+    finally:
+        stop_clp_action: ClpAction = stop_clp_package(clp_package)
+        stop_result = verify_stop_clp_action(stop_clp_action, clp_package)
```
junhaoliao, line 97: it seems like we have an exception masking issue (throwing-and-catching near the same place)

for whatever reason if the package isn't started successfully, the `assert` in L93 brings us to L95. if the package fails to start before the instance id file is generated, it seems verify_stop_clp_action() -> clp_package.get_running_services() -> get_clp_instance_id() would fail with an instance id not found error, which may bury the L93 thrown error. 

to tackle this, one way i can think of is to add a boolean flag `started` and skip the verify_stop_clp_action() optionally.

e.g.,
```
started = False
try:
    ...
    assert start_result, start_result.failure_message
    started = True
    yield clp_package
finally:
    stop_clp_action: ClpAction = stop_clp_package(clp_package)

    if started:
        stop_result = verify_stop_clp_action(stop_clp_action, clp_package)
    else:
        stop_result = stop_clp_action.verify_returncode()
    ...
```

feel free to object if you think this may complicate the handling

junhaoliao (approved):

Bill-hbrhbr (approved):

integration-tests/tests/package_tests/fixtures.py
quinntaylormitchell: I originally had something like this but we addressed it in [this previous comment](https://github.com/y-scope/clp/pull/2231#discussion_r3285295904); plus I'm fairly sure that the `pytest` failure reporting is good enough that nothing will get buried (at least, when I've done runs where start and stop fail, nothing has been buried). Will keep an eye on it though.

PR #2232, opened 2026-04-29
feat(integration-tests): Add CLP package compression action helpers and verification scaffolding.

quinntaylormitchell:
# Description

This PR adds the action and verification infrastructure for testing package compression and decompression.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end package lifecycle helpers (start/stop, compress/decompress) and verification flows, including content checks for unstructured data and service-run validation; tests refactored to use the new package fixture and helpers.
  * Removed several legacy compression fixtures and helpers; some tests updated with TODOs/placeholders.

* **Chores**
  * Added utility to write YAML atomically and streamlined test config/path fixtures for package-focused integration tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/clp/pull/2232?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


integration-tests/tests/package_tests/clp_text/test_clp_text.py
Bill-hbrhbr: I have two issues with this test:
1. 
If we want `starting test` and `test complete` messages for every test, we can use pytest hooks, logging configs, or fixtures.
One option is to use `pytest_runtest_logstart` and `pytest_runtest_logfinish` inside `conftest.py`. However, after trying it myself, I noticed that `pytest_runtest_logfinish` leaks log statements to stdout, which conflicts with the current design where everything is written to the test output log.
```
bingranhu@DESKTOP-9ER04S9:~/clp/integration-tests$ uv run pytest tests/package_tests/clp_text/test_clp_text.py
======================================================== test session starts ========================================================
platform linux -- Python 3.10.12, pytest-9.0.2, pluggy-1.6.0 -- /home/bingranhu/clp/integration-tests/.venv/bin/python3
cachedir: .pytest_cache
rootdir: /home/bingranhu/clp/integration-tests
configfile: .pytest.ini
plugins: anyio-4.12.1, env-1.2.0
Log directory for this test run: /home/bingranhu/clp/build/integration_tests/test_logs/testrun_2026-05-28-02-13-31
collected 2 items
The following tests will run:
        tests/package_tests/clp_text/test_clp_text.py::test_clp_text_startup[clp_package0]
        tests/package_tests/clp_text/test_clp_text.py::test_clp_text_compression_text_multifile[clp_package0]

Running tests now...

tests/package_tests/clp_text/test_clp_text.py::test_clp_text_startup[clp_package0] PASSED
tests/package_tests/clp_text/test_clp_text.py::test_clp_text_compression_text_multifile[clp_package0] PASSED

======================================================== 2 passed in 41.07s =========================================================
2026-05-28T02:14:15.050 INFO [conftest] Test complete: 'test_clp_text_compression_text_multifile[clp_package0]'
```
The same issue exists for `pytest_runtest_teardown`.
Ref: https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_runtest_protocol

What you can do is to introduce the following fixture:
```
@pytest.fixture(autouse=True)
def log_test_lifecycle(request: pytest.FixtureRequest) -> Iterator[None]:
    """
    Logs the start and completion of a test.

    :param request:
    """
    test_name = request.node.name
    logger.info("Starting test: '%s'", test_name)
    yield
    logger.info("Test complete: '%s'", test_name)
```

2. 
Since package startup is already required for other tests to run successfully, and now that we have compression tests in place, I think we can remove this startup smoke test.

I understand the argument for keeping these tests: provide a quick workflow to verify that the package can start and stop successfully. In that case, though, the pytest marker should probably be `startstop` rather than `startup`.

integration-tests/tests/package_tests/clp_json/test_clp_json.py
Bill-hbrhbr: ditto

quinntaylormitchell: agreed

integration-tests/tests/package_tests/clp_text/test_clp_text.py
quinntaylormitchell: 1. Agreed; added `log_test_lifecycle` fixture to new module `integration-tests/tests/fixtures/logging.py`.
2. I'd like to keep the smoketest, I think it's helpful. I've changed the name to `test_clp_*_startstop` and changed the pytest marker to `startstop`.

integration-tests/tests/package_tests/utils/compress.py
```diff
@@ -0,0 +1,153 @@
...
+    args = CompressArgs(
+        script_path=path_config.compress_path,
+        config=clp_package.temp_config_file_path,
+        paths=[dataset.logs_path],
+    )
+
+    if clp_package.clp_config.package.storage_engine == StorageEngine.CLP_S:
+        args.dataset = dataset.metadata.dataset_name
+        args.timestamp_key = dataset.metadata.timestamp_key
+        args.unstructured = dataset.metadata.unstructured
```
Bill-hbrhbr: Though it may not have an effect for clp-text packages, the `unstructured` argument is not limited to clp-json packages.

integration-tests/tests/package_tests/utils/compress.py
```diff
@@ -0,0 +1,153 @@
...
+
+    args: CompressArgs = _construct_compress_args(clp_package, dataset)
+    return ClpAction.from_args(args)
+
+
+def _construct_compress_args(clp_package: ClpPackage, dataset: SampleDataset) -> CompressArgs:
+    """Constructs the `CompressArgs` object for compressing the specified dataset."""
+    path_config = clp_package.path_config
+    args = CompressArgs(
+        script_path=path_config.compress_path,
```
Bill-hbrhbr:
```suggestion
    args = CompressArgs(
        script_path=clp_package.path_config.compress_path,
```

integration-tests/tests/package_tests/utils/compress.py
```diff
@@ -0,0 +1,153 @@
...
+    :param compress_action:
+    :param clp_package:
+    :param original_dataset:
+    :return: A `ClpVerificationResult` indicating the success or failure of the verification.
+    """
+    logger.info("Verifying '%s' package compression.", clp_package.mode_name)
+
+    returncode_result = compress_action.verify_returncode()
+    if not returncode_result:
+        return returncode_result
```
Bill-hbrhbr:
```suggestion
    result = compress_action.verify_returncode()
    if not result:
        return result
```
The name `verify_returncode` is explanatory enough that we don't need a qualifier for `result` object.

integration-tests/tests/package_tests/utils/compress.py
```diff
@@ -0,0 +1,153 @@
...
+    # Decompress the contents of `clp-package/var/data/archives`.
+    clear_directory(path_config.package_decompression_dir)
+    decompress_action = decompress_clp_package(clp_package, path_config.package_decompression_dir)
+    decompress_returncode_result = decompress_action.verify_returncode()
+    if not decompress_returncode_result:
+        return compress_action.fail_verification(
+            "During compress action verification, supporting call to 'decompress.sh' returned a"
+            " non-zero exit code.",
+            supporting_action=decompress_action,
+        )
```
Bill-hbrhbr:
```suggestion
    result = decompress_action.verify_returncode()
    if not result:
        return compress_action.fail_verification(
            "During compress action verification, supporting call to 'decompress.sh' returned a"
            " non-zero exit code.",
            supporting_action=decompress_action,
        )
```

integration-tests/tests/package_tests/utils/compress.py
```diff
@@ -0,0 +1,153 @@
...
+        args.unstructured = dataset.metadata.unstructured
+
+    return args
+
+
+def verify_compress_action(
+    compress_action: ClpAction,
+    clp_package: ClpPackage,
+    original_dataset: SampleDataset,
+) -> ClpVerificationResult:
```
Bill-hbrhbr: Looking at this function, it really just seems like a multiplexer between `_verify_compress_action_unstructured_logs` and `_verify_compress_action_structured_logs`. Wonder why we cannot abstract this layer away.

That being said, it is possible to have unstructured logs as input and structured logs as output (e,g, clp-s + log-converter). We need a much better differentiation between clp-text, clp-json package concepts and unstructured/structured input concepts.

At the very least, `_verify_compress_action_structured_logs` will need two branches too, one for structured inputs and one for unstructured inputs.

Bill-hbrhbr: Can we inline `verify_compress_action_unstructured_logs` in the clp-text test?

integration-tests/tests/package_tests/utils/compress.py
```diff
@@ -0,0 +1,153 @@
...
+def _construct_compress_args(clp_package: ClpPackage, dataset: SampleDataset) -> CompressArgs:
+    """Constructs the `CompressArgs` object for compressing the specified dataset."""
+    path_config = clp_package.path_config
+    args = CompressArgs(
+        script_path=path_config.compress_path,
+        config=clp_package.temp_config_file_path,
+        paths=[dataset.logs_path],
+    )
+
+    if clp_package.clp_config.package.storage_engine == StorageEngine.CLP_S:
```
Bill-hbrhbr: Can we branch on package flavour/mode here instead of `clp_config.package.storage_engine`? The tests are oriented around the flavour concept, so using that would make the intent clearer.

Bill-hbrhbr (changes requested):
I think this refactor may have gone a bit too far in consolidating `clp-text` and `clp-json` behavior behind shared helpers. The test bodies are now very terse, but they also hide the important differences between the package flavors. For example, `clp-json` needs dataset/timestamp handling for CLP-S, while `clp-text` follows the plain CLP path.

If all of that gets abstracted behind common utilities, I think there is less value in keeping separate `clp_json` and `clp_text` test modules in the first place. I’d prefer keeping more of the command construction and package-specific expectations inline within each test, while limiting shared helpers to smaller pieces that remove obvious duplication. That makes the tests read more like executable documentation for each package flavor.

integration-tests/tests/package_tests/utils/compress.py
quinntaylormitchell: agreed

integration-tests/tests/package_tests/utils/compress.py
quinntaylormitchell: branching no longer needed (see new changes)

integration-tests/tests/package_tests/utils/compress.py
quinntaylormitchell: Good point, thank you

integration-tests/tests/package_tests/utils/compress.py
quinntaylormitchell: See new changes

integration-tests/tests/package_tests/utils/compress.py
quinntaylormitchell: agreed

integration-tests/tests/package_tests/utils/compress.py
quinntaylormitchell: agreed

integration-tests/tests/package_tests/clp_json/verification/compress.py
```diff
@@ -0,0 +1,59 @@
...
+    Verifies that the clp-json `clp_package` has compressed `original_dataset` correctly by
+    decompressing the compressed logs and comparing the decompressed logs to the original dataset.
+
+    :param action:
+    :param clp_package:
+    :param original_dataset:
+    :return: A `ClpVerificationResult` indicating whether the round-trip matched the original logs.
+    """
+    if not isinstance(action.args, CompressArgs):
+        err_msg = "'verify_compress_structured_clp_json' requires a 'CompressArgs' action."
```
Bill-hbrhbr:
```suggestion
        err_msg = "Verification expects a 'CompressArgs' action."
```
Nit: I think if the test fails, it already prints a stack trace. There would be less dup and maintenance overhead if we remove the function name from the error log.

integration-tests/tests/package_tests/clp_json/verification/compress.py
```diff
@@ -0,0 +1,59 @@
...
+        raise TypeError(err_msg)
+    if action.args.unstructured or original_dataset.metadata.unstructured:
+        err_msg = "'verify_compress_structured_clp_json' cannot verify unstructured datasets."
+        raise ValueError(err_msg)
```
Bill-hbrhbr, line 27:
```suggestion
    if not action.args.structured or not original_dataset.metadata.structured:
        err_msg = "Verification expects a structured dataset workflow."
        raise ValueError(err_msg)
```
Nit: I know this is currently binary, but it is better to validate against the expected state rather than rejecting a specific alternative state.

integration-tests/tests/package_tests/clp_json/verification/compress.py
```diff
@@ -0,0 +1,59 @@
...
+    :param clp_package:
+    :param original_dataset:
+    :return: A `ClpVerificationResult` indicating whether the round-trip matched the original logs.
+    """
+    if not isinstance(action.args, CompressArgs):
+        err_msg = "'verify_compress_unstructured_clp_json' requires a 'CompressArgs' action."
+        raise TypeError(err_msg)
+    if not action.args.unstructured or not original_dataset.metadata.unstructured:
+        err_msg = "'verify_compress_unstructured_clp_json' can only verify unstructured datasets."
+        raise ValueError(err_msg)
```
Bill-hbrhbr, line 53:
```suggestion
    if not isinstance(action.args, CompressArgs):
        err_msg = "Verification expects a 'CompressArgs' action."
        raise TypeError(err_msg)
    if not action.args.unstructured or not original_dataset.metadata.unstructured:
        err_msg = "Verification expects an unstructured dataset workflow."
        raise ValueError(err_msg)
```

Bill-hbrhbr: ditto

integration-tests/tests/package_tests/clp_json/verification/compress.py
Bill-hbrhbr: Also wondering if these will be removed since I don't see the same checks inside `test_clp_text.py`.

integration-tests/tests/package_tests/clp_json/verification/compress.py
Bill-hbrhbr: Also wondering if these will be removed since I don't see the same checks inside `test_clp_text.py`.

integration-tests/tests/package_tests/clp_json/verification/compress.py
Bill-hbrhbr: Also wondering if these will be removed since I don't see the same checks inside `test_clp_text.py`.

Bill-hbrhbr (changes requested):
Some nit picks to get things started. If you see the notification you can address them now. Will have more comments on the actual test functions (but I'm not expecting a lot).

integration-tests/tests/package_tests/clp_text/verification/compress.py
```diff
@@ -0,0 +1,62 @@
...
+    decompressing the compressed logs and comparing the decompressed logs to the original dataset.
+
+    :param action:
+    :param clp_package:
+    :param original_dataset:
+    :return: A `ClpVerificationResult` indicating whether the round-trip matched the original logs.
+    """
+    result = action.verify_returncode()
+    if not result:
+        return result
```
Bill-hbrhbr: This part should be inlined into test_clp_text_compression_text_multifile. Once the action object has been created and the subprocess has finished running, we should verify the result immediately.

In other words, this function should assume the compression is successful, as its primary purpose is to verify the integrity of compression, not to propagate compression failures.

integration-tests/tests/package_tests/clp_text/verification/compress.py
```diff
@@ -0,0 +1,62 @@
...
+
+    path_config = clp_package.path_config
+    clear_directory(path_config.package_decompression_dir)
```
Bill-hbrhbr, line 30: Now that I look at the code, we probably should have made it clearer that `package_decompression_dir` is a test directory by using a more descriptive name, such as `package_test_decompression_dir`.
Alternatively, we could use temporary directory and file utilities, with `test_cache_dir` as the parent directory. It seems unlikely that the one-shot decompression output would be shared elsewhere in the test.

The same naming issue applies to `temp_config_dir`. Something like `package_test_config_dir` or `package_test_temp_config_dir` would be more clear.

@coderabbitai open an issue to track these naming clarifications.

integration-tests/tests/package_tests/clp_text/verification/compress.py
```diff
@@ -0,0 +1,62 @@
...
+            supporting_action=decompress_action,
+        )
+
+    # Verify equality between original logs and decompressed logs.
+    original_logs_path = original_dataset.logs_path
+    decompressed_logs_path = path_config.package_decompression_dir / original_logs_path.relative_to(
+        original_logs_path.anchor
+    )
+    equal = is_dir_tree_content_equal(original_logs_path, decompressed_logs_path)
+    clear_directory(path_config.package_decompression_dir)
```
Bill-hbrhbr: The cleanup behavior is a little inconsistent here:
- If decompression fails, the output directory is left untouched
- If decompression succeeds, but content comparison fails, the directory is cleared

It would be useful to keep the decompressed output on content mismatch as well, since that would facilitate debugging.

integration-tests/tests/package_tests/clp_text/verification/compress.py
```diff
@@ -0,0 +1,62 @@
...
+
+    if equal:
+        return action.pass_verification()
+
+    return action.fail_verification(
+        f"Compress verification failure: mismatch between original logs at"
+        f" '{original_logs_path}' and decompressed logs at '{decompressed_logs_path}'.",
+        supporting_action=decompress_action,
+    )
```
Bill-hbrhbr:
```suggestion
    if not is_dir_tree_content_equal(original_logs_path, decompressed_logs_path):
        return action.fail_verification(
            f"Compress verification failure: mismatch between original logs at"
            f" '{original_logs_path}' and decompressed logs at '{decompressed_logs_path}'.",
            supporting_action=decompress_action,
        )
        
    clear_directory(path_config.package_decompression_dir)
    return action.pass_verification()
```

integration-tests/tests/package_tests/utils/decompress.py
```diff
@@ -0,0 +1,32 @@
...
+from tests.utils.classes import CmdArgs
+
+
+class DecompressArgs(CmdArgs):
+    """Command argument model for decompressing with the CLP package."""
+
+    script_path: Path
+    config: Path
+    extraction_dir: Path
+    paths: list[Path] | None = None
```
Bill-hbrhbr:
```suggestion
    paths: list[Path] = Field(default_factory=list)
```
Lists inherently can represent "zero or more paths`. Don't think we need `None`. It's hard to justify the difference between `None` and an empty list anyway.

integration-tests/tests/package_tests/utils/decompress.py
```diff
@@ -0,0 +1,32 @@
...
+            str(self.script_path),
+            "--config",
+            str(self.config),
+            EXTRACT_FILE_CMD,
+            "--extraction-dir",
+            str(self.extraction_dir),
+        ]
+
+        if self.paths:
+            cmd.extend([str(path) for path in self.paths])
```
Bill-hbrhbr:
```suggestion
        cmd.extend([str(path) for path in self.paths])
```

Bill-hbrhbr (changes requested):
more comments

integration-tests/tests/package_tests/clp_json/verification/compress.py
quinntaylormitchell: >remove the function name from the error log

Agreed.

>wondering if these will be removed

No, I just forgot to add similar checks to the `text` version. Will add.

integration-tests/tests/package_tests/clp_json/verification/compress.py
quinntaylormitchell: >better to validate against the expected state rather than rejecting a specific alternative state

I think I know what you mean here; I use `unstructured` rather than `structured` for a couple reasons:

1. `clp-json` is the default operating mode for CLP, so in a sense, I think of `structured` as being the default data type
2. The flag specified via the command line that signifies `structured` or `unstructured` is `--unstructured`, so it feels a bit more intuitive.

That being said, if you'd like me to change it to `structured` in spite of those reasons, I will (I don't feel very strongly about it).

>wondering if these will be removed

I'm not gonna check for structured/unstructured in `clp_text` because if the dataset was compressed with `clp_text`, then it doesn't matter whether the original dataset was structured or unstructured, it will have been compressed as unstructured and should be validated as such.

integration-tests/tests/package_tests/clp_json/verification/compress.py
quinntaylormitchell: see above

integration-tests/tests/package_tests/clp_text/verification/compress.py
quinntaylormitchell: I'm on the fence, cause I do agree with you but I'm also wary that moving it will increase test function length.

I'll move it for now, and I'll make a note to reassess later when there are more tests.

integration-tests/tests/package_tests/clp_text/verification/compress.py
quinntaylormitchell: Do you want me to make those changes in this PR?

integration-tests/tests/package_tests/clp_text/verification/compress.py
quinntaylormitchell: Agreed.

integration-tests/tests/package_tests/clp_text/verification/compress.py
quinntaylormitchell: Agreed.

integration-tests/tests/package_tests/utils/decompress.py
quinntaylormitchell: Agreed.

integration-tests/tests/package_tests/utils/decompress.py
quinntaylormitchell: Agreed.

integration-tests/tests/package_tests/clp_json/verification/compress.py
Bill-hbrhbr: I agree with the unstructured check. It's more about checking the `action.args` against `CompressArgs` on line 22.

integration-tests/tests/package_tests/clp_text/verification/compress.py
Bill-hbrhbr: Would be best in a separate pure refactor PR (which there's no urgent need to put up right now).

Bill-hbrhbr (dismissed):
Two explanatory comments on existing discussions from the previous review.
Lgtm.

Bill-hbrhbr:
For title, how about:
```
feat(integration-tests): Add CLP package compression tests with action-based execution and verification.
```

integration-tests/tests/fixtures/logging.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+@pytest.fixture(autouse=True)
+def log_test_lifecycle(request: pytest.FixtureRequest) -> Iterator[None]:
+    """
+    Logs a message that identifies a test by name before it starts and after it finishes.
+
+    :param request:
+    """
+    test_name = request.node.originalname
```
junhaoliao: qq: why not

```suggestion
    test_name = request.node.nodeid
```

which shows the full test suite name too i believe?

junhaoliao (dismissed):
for the title, how about:
```
feat(integration-tests): Add CLP package compression action helpers and verification scaffolding (resolves #2232).
```

to hint the core logic will come in a future pr

integration-tests/tests/fixtures/logging.py
quinntaylormitchell: Very good point, I will make that change

junhaoliao (approved):

PR #2235, opened 2026-04-29
feat(integration-tests): Add package search testing verification infra.

quinntaylormitchell:
# Description

This PR adds the action and verification infrastructure for testing package search.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **Tests**
  * Added package search coverage for file targeting, case sensitivity, result counts, count-by-time, and time-range filtering.
  * Search verification now checks structured and text results, including timestamp boundaries and malformed or invalid inputs.
  * Added support for epoch-millisecond and formatted timestamp metadata.
  * Improved validation of search output and comparison against expected results.
* **Documentation**
  * Updated dataset metadata guidance with supported timestamp formats and time-range search requirements.


integration-tests/tests/package_tests/utils/search.py
```diff
@@ -0,0 +1,277 @@
...
+    :return: The `ClpAction` instance that runs the search.
+    """
+    logger.info(
+        "Performing '%s' search on the '%s' dataset.",
+        search_type.name,
+        dataset.dataset_name,
+    )
+
+    args: SearchArgs = _construct_args(clp_package, dataset, search_type, wildcard_query)
+    return ClpAction(cmd=args.to_cmd(), args=args)
```
Bill-hbrhbr:
```suggestion
    return ClpAction.from_args(args)
```

integration-tests/tests/package_tests/utils/search.py
```diff
@@ -0,0 +1,277 @@
...
+    :param wildcard_query:
+    :return: The `ClpAction` instance that runs the search.
+    """
+    logger.info(
+        "Performing '%s' search on the '%s' dataset.",
+        search_type.name,
+        dataset.dataset_name,
+    )
+
+    args: SearchArgs = _construct_args(clp_package, dataset, search_type, wildcard_query)
```
Bill-hbrhbr: again, feels like there's no need to create a helper for what feels like the meat of `search_clp_package`.

integration-tests/tests/package_tests/utils/search.py
```diff
@@ -0,0 +1,277 @@
...
+    """
+    logger.info(
+        "Verifying '%s' search on the '%s' dataset.",
+        search_type.name,
+        original_dataset.dataset_name,
+    )
+
+    returncode_result = action.verify_returncode()
+    if not returncode_result:
+        return returncode_result
```
Bill-hbrhbr:
```suggestion
    result = action.verify_returncode()
    if not result:
        return result
```

integration-tests/tests/data/json_multifile/metadata.json
```diff
@@ -12,5 +12,6 @@
         "sts-135-2011-07-19.jsonl",
         "sts-135-2011-07-21.jsonl"
     ],
-    "single_match_wildcard_query": "\"detail\":\"Roll program complete, heads down attitude achieved for ascent\""
+    "single_match_wildcard_query": "\"detail\":\"Roll program complete, heads down attitude achieved for ascent\"",
+    "single_match_file": "sts-135-2011-07-08.jsonl"
```
Bill-hbrhbr: This is starting to feel like these entries are more like test cases instead of metadata. Why don't we inline these queries and query results in test code directly?

Also each time you add something, you'd have to update the metadata format documentation.

integration-tests/tests/package_tests/utils/search.py
```diff
@@ -0,0 +1,277 @@
...
+
+class ClpPackageSearchType(Enum):
+    """Possible search types."""
+
+    BASIC = auto()
+    FILE_PATH = auto()
+    IGNORE_CASE = auto()
+    COUNT_RESULTS = auto()
+    COUNT_BY_TIME = auto()
+    TIME_RANGE = auto()
```
Bill-hbrhbr: I don't really like different test case flavours hiding away as an Enum class inside a utils folder. Without looking at your existing infra, the clearest design to me would look something like:
```
  @pytest.mark.parametrize(
      "clp_package",
      [CLP_TEXT_MODE, CLP_JSON_MODE],
      indirect=True,
      ids=[CLP_TEXT_MODE.mode_name, CLP_JSON_MODE.mode_name],
  )
  class TestPackageSearch:
      def test_basic_search(self, clp_package: ClpPackage) -> None:
          ...

      def test_file_path_search(self, clp_package: ClpPackage) -> None:
          ...

      def test_ignore_case_search(self, clp_package: ClpPackage) -> None:
          ...

      def test_count_results_search(self, clp_package: ClpPackage) -> None:
          ...

      def test_count_by_time_search(self, clp_package: ClpPackage) -> None:
          ...

      def test_time_range_search(self, clp_package: ClpPackage) -> None:
          ...
```

Bill-hbrhbr: However, now you already have a split file design. The split file design is sound if the package behaviors are meaningfully different and you want each file to document a specific package flavor.

But it becomes weak if the tests end up looking like:

- `test_clp_text.py` → calls a generic helper
- `test_clp_json.py` → calls the same generic helper

with all meaningful behavior hidden behind shared utilities.

A sound split file design with search tests would look more like:

- `test_clp_text.py`
    - Uses `text_multifile`
    - Shows CLP/text specific command expectations
    - Does not pass `--dataset`
    - Does not pass `--timestamp-key`
    - Verifies plain text search behavior

- `test_clp_json.py`
    - Uses `json_multifile`
    - Shows CLP-S/JSON specific command expectations
    - Passes `--dataset`
    - Passes `--timestamp-key`
    - Verifies structured search behavior and unstructured + log-converter.

There is an abundance of over-abstracting the package specific behavior that I'm seeing in these PRs.

A rule of thumb for this test suite could be:

- Keep package lifecycle fixtures shared.
- Keep low-level action, logging, and result helpers shared.
- Keep subprocess calling shared.

However, keep package-specific command construction and expectations visible in `test_clp_text.py` and `test_clp_json.py`. That preserves the value of having separate test modules.

integration-tests/tests/utils/classes.py
```diff
@@ -131,6 +132,13 @@ def __post_init__(self) -> None:
             file_path_abs = self.logs_path / file_path
             validate_file_exists(file_path_abs)
 
+        if self.metadata.single_match_file not in self.metadata.file_names:
+            err_msg = (
+                f"`single_match_file` '{self.metadata.single_match_file}' is not listed in"
+                " `file_names`."
+            )
+            raise ValueError(err_msg)
```
Bill-hbrhbr: Technically, there's no need to verify this, because if the `single_match_file` contains the wrong value, the search test will fail anyway.
This touches upon my previous comment on what should be **metadata** and what should merely be test case constants/inputs/outputs.

Bill-hbrhbr (changes requested):
This PR doesn't actually include any search tests right? Let's include them instead of simply introducing a portion of the search test infra.
Also, most of my concerns are elaborated in the comments.

integration-tests/tests/utils/classes.py
```diff
@@ -76,6 +77,22 @@ def _static_paths(self) -> list[Path]:
         return [self.test_data_dir]
 
 
+class EpochMsTimestampFormat(BaseModel):
+    """Indicates a timestamp encoded as an integer number of milliseconds since the UNIX epoch."""
+
+    kind: Literal["epoch_ms"]
```
```suggestion
    type: Literal["epoch"]
    unit: Literal["s", "ms", "us", "ns"]
```
I like the `type` key name better than the `kind`. Sounds more canonical.
Also I think it be beneficial to have `epoch` as a category and `ms` in a separate unit key.

integration-tests/tests/package_tests/clp_json/verification/search.py
```diff
@@ -0,0 +1,350 @@
+"""Search verification helpers specific to the clp-json package."""
```
Bill-hbrhbr, line 1: since there are no actual new additions of clp-json tests, can we defer this to the PR where we add the actual tests?

integration-tests/tests/package_tests/clp_json/test_clp_json.py
```diff
@@ -105,20 +105,3 @@ def test_clp_json_compression_text_multifile(
     logger.info("Verifying the compression of the 'text_multifile' dataset.")
     result = verify_compress_unstructured_clp_json(action, clp_package, text_multifile)
     assert result, result.failure_message
-
```
Bill-hbrhbr, line 108: ditto. let's not touch this part until we decide to include clp-json search tests.

integration-tests/tests/package_tests/utils/search.py
```diff
@@ -0,0 +1,83 @@
...
+import logging
+from datetime import datetime, timezone
+from pathlib import Path
+
+from tests.utils.classes import CmdArgs
+
+logger = logging.getLogger(__name__)
+
+
+DEFAULT_COUNT_BY_TIME_INTERVAL = 10
```
Bill-hbrhbr: this is currently not referenced anywhere. probably belonged to a removed clp-json search test.

integration-tests/tests/package_tests/utils/search.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def parse_timestamp_to_epoch_ms(timestamp: str, strptime_pattern: str) -> int:
```
Bill-hbrhbr: this should ideally be a function of `class StrptimeTimestampFormat(BaseModel):`

can be renamed to something like `to_epoch_ms(timestamp: str)`.

integration-tests/tests/utils/classes.py
```suggestion
    type: Literal["epoch"]
    unit: Literal["s", "ms"]
```
`Type` might be a better key name. Also while we are at it, I think we can support epoch timestamps in different units (but it's up to you).

integration-tests/tests/package_tests/clp_json/verification/search.py
```diff
@@ -0,0 +1,350 @@
...
+            if char == open_quote:
+                open_quote = None
+        elif char in ('"', "'"):
+            open_quote = char
+        elif char == KV_DELIMITER_COLON:
+            return index
+    return None
+
+
+def _strip_surrounding_quotes(text: str) -> str:
```
Bill-hbrhbr: We should use JSON libraries to handle stripping away quotes. There may be unhandled escape sequences inside.

Bill-hbrhbr: Also, this should be inside the general utils directory, not specific to clp-json test folder, even though for now only clp-json uses it.

integration-tests/tests/package_tests/clp_json/verification/search.py
```diff
@@ -0,0 +1,350 @@
...
+    if _normalize_entries(found_entries) == _normalize_entries(search_output_list):
+        return action.pass_verification()
+
+    return action.fail_verification(
+        "Search verification failure: mismatch between search output and expected output.\n"
+        f"Expected output: '{found_entries}'\nActual output: '{search_output_list}'",
+    )
+
+
+def _construct_kv_pair_from_query(query: str) -> tuple[str, Any]:
```
Bill-hbrhbr, line 111: what if the query has `AND` `OR` operators?
For example, for:
```
level: ERROR AND component: scheduler
```
you would get key = `level` and value = `ERROR AND component: scheduler`

integration-tests/tests/package_tests/clp_json/verification/search.py
```diff
@@ -0,0 +1,350 @@
...
+def verify_search_clp_json(
+    action: ClpAction,
+    clp_package: ClpPackage,
+    dataset: SampleDataset,
+) -> ClpVerificationResult:
+    """
+    Verifies that the search action performed on the `clp-json` package returns the expected
+    results. Verification is performed as follows:
+
+    1. The query string from the search action is parsed into a key-value pair.
```
Bill-hbrhbr: Is this statement saying that we don't allow composite queries? (like the one I listed as a counter-example to your query splitting utility).

integration-tests/tests/package_tests/clp_json/verification/search.py
```diff
@@ -0,0 +1,350 @@
...
+    :return: The processed text string.
+    """
+    if len(text) >= MIN_QUOTED_LENGTH and text[0] == text[-1] and text[0] in ('"', "'"):
+        return text[1:-1]
+    return text
+
+
+def _load_all_jsonl_logs_from_dataset(dataset: SampleDataset) -> list[dict[str, Any]]:
+    """
+    Loads every JSONL log from a dataset into a single list of dicts.
```
Bill-hbrhbr:
```suggestion
    Loads every JSON Lines log from a dataset into a single list of dicts.
```
This is not an exhaustive comment, but let's write `JSON Lines` instead of `JSONL`.
Suffixes can include `.jsonl` and `.ndjson`.

integration-tests/tests/package_tests/clp_json/verification/search.py
Bill-hbrhbr: I'm not sure why the tests from `test_clp_json.py` were removed, since that file seems to contain a fairly comprehensive set of utilities for validating clp-json query results. My two comments:

1. The JSON parsing utilities would probably be better placed in the general utilities folder so they can be reused by other tests.
2. As a starting point, I think we could simply verify the total match count when testing clp-json search results. That would let us add basic coverage without having to deal with parsing JSON strings right away.

integration-tests/tests/utils/classes.py
```diff
@@ -76,6 +77,22 @@ def _static_paths(self) -> list[Path]:
         return [self.test_data_dir]
 
 
+class EpochMsTimestampFormat(BaseModel):
```
Bill-hbrhbr: In general, I think these new classes would be better placed in a dedicated `timestamps.py` module under the same `utils` directory. The current `classes.py` seems to be accumulating too many unrelated classes, so separating the timestamp-related ones into their own module would keep things better organized.

integration-tests/tests/package_tests/clp_json/verification/search.py
```diff
@@ -0,0 +1,350 @@
...
+    for entry in entries:
+        if timestamp_key not in entry:
+            pytest.fail(f"Log entry '{entry}' is missing the timestamp key '{timestamp_key}'.")
+        timestamp = _resolve_timestamp_to_ms(entry[timestamp_key], timestamp_key, timestamp_format)
+        if (begin_ts is None or begin_ts <= timestamp) and (end_ts is None or timestamp <= end_ts):
+            filtered_entries.append(entry)
+    return filtered_entries
+
+
+def _resolve_timestamp_to_ms(
```
Bill-hbrhbr: This can be an API that's provided respectively by `EpochMsTimestampFormat` and `StrptimeTimestampFormat`. You can check if the format string is convertible for strptime and whether the value is integer for `epoch` timestamp.

Bill-hbrhbr: If CLP is strictly taking epoch ms timestamps, `EpochMsTimestampFormat` and `StrptimeTimestampFormat` can share a base `TimestampFormat` class that provides the `to_epoch_ms` virtual method. Then `TimestampFormat` can represent more than a mere type alias.

integration-tests/tests/package_tests/clp_json/verification/search.py
```diff
@@ -0,0 +1,350 @@
...
+
+def _construct_kv_pair_from_query(query: str) -> tuple[str, Any]:
+    """
+    Constructs a key-value pair from a query string. The query is split on the first
+    `KV_DELIMITER_COLON` that is not inside a quoted group (`"..."` or `'...'`).
+
+    :param query:
+    :return: `(KEY, VALUE)` if `KV_DELIMITER_COLON` is present, else
+        `(WILDCARD_MULTIMATCH_CHAR, VALUE)`.
+    """
```
Bill-hbrhbr, line 118: using a constant `KV_DELIMITER_COLON` inside the docstring makes it a bit harder to read.

integration-tests/tests/package_tests/clp_json/verification/search.py
```diff
@@ -0,0 +1,350 @@
...
+    :return: The index of the first unquoted colon, or `None` if there is none.
+    """
+    open_quote: str | None = None
+    for index, char in enumerate(query):
+        if open_quote is not None:
+            if char == open_quote:
+                open_quote = None
+        elif char in ('"', "'"):
+            open_quote = char
+        elif char == KV_DELIMITER_COLON:
```
Bill-hbrhbr: Not sure why we use the `KV_DELIMITER_COLON` constant in one `if` statement but raw quote characters in another. For simple characters that are unlikely to change, I'd just use the literal characters consistently in both the code and the docstrings. 
BTW, that's already the approach taken for the quote characters in `_strip_surrounding_quotes`'s docstring.

integration-tests/tests/package_tests/clp_text/verification/search.py
```diff
@@ -0,0 +1,179 @@
+"""Search verification helpers specific to the clp-text package."""
```
Bill-hbrhbr, line 1: Just realized that we don't have actual running search tests for `clp-text` either. I think it'd be beneficial to add count tests to verify that things are working.
Grep uses regex queries, while clp-text uses wildcard substring queries, so maybe we need regex and wildcard translation utilities? I know there is one inside `clp_json/verification/search.py`, so it's crucial that we put these general utilities under the general `utils` directory.

Bill-hbrhbr (changes requested):

integration-tests/tests/package_tests/clp_json/test_clp_json.py
quinntaylormitchell: Sure

integration-tests/tests/package_tests/clp_json/verification/search.py
quinntaylormitchell: Yes, this code doesn't support composite queries. That was by design, because I didn't want to introduce too much complexity all at once. I'll update the docstring to make that clearer.

integration-tests/tests/package_tests/clp_json/verification/search.py
quinntaylormitchell: Right; see above

integration-tests/tests/package_tests/clp_json/verification/search.py
quinntaylormitchell: The main reason is that a delimiter can be anything (hypothetically), but the term "quotation marks" refers to a specific set of characters (either `"` or `'`). To me, it seems like delimiter characters are much more likely to change than quotation marks.

That being said, I'll rename `KV_DELIMITER_COLON` to `KV_DELIMITER` to fully recognize that it may change.

Also: not sure what you mean re. the docstrings; fairly sure that all methods in this file refer to quotation marks as literals and to the delimiter character as `KV_DELIMITER_COLON`. Did I misunderstand you?

integration-tests/tests/package_tests/clp_json/verification/search.py
quinntaylormitchell: I mostly agree, but since `jsonl` stands for "JSON Lines", I think I'll use the term "Line-delimited JSON" instead.

integration-tests/tests/package_tests/utils/search.py
quinntaylormitchell: Right on, thank you

integration-tests/tests/utils/classes.py
quinntaylormitchell: >I like the `type` key name better than the `kind`.

Agreed; changed to `type`

> I think it be beneficial to have epoch as a category and ms in a separate unit key

I think I’m going to leave it as `epoch_ms` for now, but later if/when we add datasets with other units, I’m happy to update it.

integration-tests/tests/package_tests/clp_json/verification/search.py
quinntaylormitchell: Trying to align with the fact that `KV_DELIMITER_COLON` is what we use in the function; see comment below about changing to `KV_DELIMITER`; that might make it more readable.

integration-tests/tests/package_tests/clp_json/verification/search.py
quinntaylormitchell: >We should use JSON libraries to handle stripping away quotes

I'm not really sure how JSON libraries could accomplish this; or at least I wasn't able to see how it could be done after looking it up. Could you elaborate?

>unhandled escape sequences

I've updated the following to handle escapes:

1. Updated `_find_unquoted_kv_delimiter` to handle escaped quotation marks (i.e., `\"`)
2. Updated `_convert_wildcard_to_regex` to handle escaped literal star (i.e., `\*`)

Does that handle what you were talking about?

>this should be inside the general utils directory

Sure, will do

integration-tests/tests/utils/classes.py
quinntaylormitchell: >Type might be a better key name

Agreed; changed to `type`

> I think we can support epoch timestamps in different units (but it's up to you)

I think I’m going to leave it as `epoch_ms` for now, but later if/when we add datasets with other units, I’m happy to update it.

integration-tests/tests/package_tests/clp_json/verification/search.py
quinntaylormitchell: I’ve combined the two classes into a single class `TimestampFormat` and added a method called `to_epoch_ms` that handles the format conversion. If that’s not aligned with what you meant, let me know.

integration-tests/tests/package_tests/utils/search.py
quinntaylormitchell: I’ve combined the two classes into a single class `TimestampFormat` and added a method called `to_epoch_ms` that handles the format conversion. If that’s not aligned with what you meant, let me know.

integration-tests/tests/utils/classes.py
quinntaylormitchell: Sure, I’ve added it to a new file `integration-tests/tests/utils/timestamps.py`.

integration-tests/tests/package_tests/clp_json/verification/search.py
quinntaylormitchell: >can we defer this to the PR where we add the actual tests

Will discuss offline

>I'm not sure why the tests from test_clp_json.py were removed

Not sure what you are referring to; other than the removal of the placeholder test (which I have now restored) this PR doesn't remove any utilities for validating clp-json query results; it adds them

>The JSON parsing utilities would probably be better placed in the general utilities folder so they can be reused by other tests.

There are a lot of specialized methods that use (or are related to) JSON in this module, but other than moving `strip_surrounding_quotes` to the general utilities folder (which I have done) I don't see a lot of opportunity for reuse here. We can move any method(s) we choose at a later date if there ends up being concrete opportunity for reuse later on.

>I think we could simply verify the total match count when testing clp-json search results.

Will discuss offline

integration-tests/tests/package_tests/clp_text/verification/search.py
quinntaylormitchell: >I think it'd be beneficial to add count tests to verify that things are working.

Will discuss offline

>maybe we need regex and wildcard translation utilities?

This module already has these utilities; did I misunderstand you?

>it's crucial that we put these general utilities under the general utils directory.

Like the methods used in `integration-tests/tests/package_tests/clp_text/verification/search.py`, I don't see a lot of opportunity for reuse. For example, `clp-json` search verification does not use `grep` at all, and `clp-text` uses only grep.

PR #2240, opened 2026-05-01
feat(clp-s::log_converter): Add support for compressing converted KV-IR files; make compressing outputs the default behaviour.

gibber9809:
# Description

This PR adds support for compressing converted KV-IR files using zstd. This is now the default behaviour for log-converter, and can be disabled using the `--no-compress-converted-files` flag.

Implementation-wise this involves changing `LogSeriailzer` to accept a series of nested writers instead of a single `FileWriter`. To simplify the implementation we switch to using the `clp` version of a `FileWriter` and `zstd::Compressor`, since they implement `WriterInterface`, which makes them easier to work with in a more generic way. Aside: `close` should probably be made part of `clp::WriterInterface`, and without that I'm forced to test whether a given `clp::WriterInterface` implements `close` via `dynamic_cast` to the many derived classes that individually implement `close`. (Added [2291](https://github.com/y-scope/clp/issues/2291) to track this issue with close).

# Checklist

  breaking change.

# Validation performed

* Validated that integration tests pass
* Measured that converting and compressing the `hive-24hr` dataset is only ~0.5% slower (on an HDD)

## Summary by CodeRabbit

* **New Features**
  * Added a command-line option to disable compression of converted log files (compression enabled by default).
  * Conversion output can be written with optional Zstandard streaming compression.

* **Refactor**
  * Public conversion APIs updated to accept an explicit compression flag so callers can control output compression.

* **Chores**
  * Build configuration expanded to include streaming compression and related I/O components.

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/clp/pull/2240?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


components/core/src/clp_s/log_converter/LogSerializer.hpp
```diff
@@ -88,26 +104,26 @@ class LogSerializer {
...
+              m_nested_writers{std::move(nested_writers)} {}
 
     // Methods
     /**
      * Flushes the buffer from the serializer to the output file.
      */
     void flush_buffer() {
         auto const buffer{m_serializer.get_ir_buf_view()};
-        m_writer.write(
+        m_nested_writers.back()->write(
```
20001020ycx, line 120: I am a bit confused on the order of the component in the pipeline. Looks like the `m_nested_writers` vector is defined in the order of [`FileWriter`, `Compressor`], given that you did `emplace_back` at `create`. When zstd is on, `.back()` resolves to `Compressor`, which means `flush_buffer()` now writes to `Compressor` rather than `FileWriter`.  Does this change the semantic of this function?

Furthermore, could we define the order of the expected writer chain at the `m_nested_writers` declaration? I feel like this PR assumes some order of the `m_nested_writers` in coding, I think this would help in case we add more components in the future?

components/core/src/clp_s/log_converter/LogSerializer.cpp
```diff
@@ -36,16 +44,32 @@ auto LogSerializer::create(std::string_view output_dir, std::string_view origina
     )};
 
     boost::uuids::random_generator uuid_generator;
-    std::string const file_name{boost::uuids::to_string(uuid_generator()) + ".clp"};
+    auto file_extension{use_zstd ? clp::ir::cIrFileExtension : cDefaultIRFileExtension};
```
20001020ycx: Nit, feel like this import `clp:ir:cIrFileExtension` doesn't convey the compression we will perform, but probably this is outside the scopeof this PR to update this as I assume there would be many uses of this constant. Maybe  name `cDefaultIrFileExtension` to `cUncompressedFileExtension`. 

Emmmm,  I will leave it to you, I am okay with either, just noticed while reading through the code, lol.

20001020ycx (commented):
Overall LGTM, good work, just a minor nit and a design question from myself.

components/core/src/clp_s/log_converter/LogSerializer.cpp
gibber9809: Yeah `cUncompressedFileExtension` is more readable, will update.

components/core/src/clp_s/log_converter/LogSerializer.hpp
gibber9809: Yeah I can update the docs to be clear about the expectation for nested writers. The intention is that the writers are in an order like:
`output sink writer` <- `nested writer 1` <- `nested writer 2`, etc.

The semantic of the `flush_buffer` function  is that it's supposed to write the  the buffered output from the IR serializer to some output sink. Both the previous implementation and this implementation don't flush the output sink at this time, so there's no semantic change w.r.t. durability.

20001020ycx (dismissed):
LGTM!

components/core/src/clp_s/log_converter/CommandLineArguments.cpp
```diff
@@ -116,6 +116,7 @@ auto CommandLineArguments::parse_arguments(int argc, char const** argv)
         po::options_description conversion_options("Conversion options");
         std::string input_path_list_file_path;
         std::string auth{cNoAuth};
+        bool no_compress_converted_files{};
```
kirkrodrigues:
```suggestion
        bool no_compress_converted_files{false == m_compress_converted_files};
```

components/core/src/clp_s/log_converter/LogSerializer.hpp
```diff
@@ -24,13 +28,15 @@ class LogSerializer {
      * Creates an instance of `LogSerializer`.
      * @param output_dir The destination directory for generated KV-IR.
      * @param original_file_path The original path for the file being converted to KV-IR.
+     * @param use_zstd Whether the output KV-IR should be zstd-compressed.
```
kirkrodrigues: How about `compress_with_zstd`? Feels a bit more specific.

components/core/src/clp_s/log_converter/LogSerializer.hpp
```diff
@@ -74,8 +80,18 @@ class LogSerializer {
      */
     void close() {
         flush_buffer();
-        m_writer.write_numeric_value(clp::ffi::ir_stream::cProtocol::Eof);
-        m_writer.close();
+        m_nested_writers.back()->write_numeric_value(clp::ffi::ir_stream::cProtocol::Eof);
+        for (auto it{m_nested_writers.rbegin()}; it != m_nested_writers.rend(); ++it) {
+            if (auto compressor{dynamic_cast<clp::streaming_compression::Compressor*>(it->get())};
```
kirkrodrigues, line 87: Can we open an issue about this? If one already exists, can we link it in the PR description?

components/core/src/clp_s/log_converter/LogSerializer.hpp
```diff
@@ -88,26 +104,28 @@ class LogSerializer {
...
+        m_nested_writers.back()->write(
                 clp::size_checked_pointer_cast<char const>(buffer.data()),
                 buffer.size_bytes()
         );
         m_serializer.clear_ir_buf();
     }
 
     clp::ffi::ir_stream::Serializer<clp::ir::eight_byte_encoded_variable_t> m_serializer;
-    clp_s::FileWriter m_writer;
+    // Nested writers are ordered from closest to furthest from output sink. Typically this will
```
kirkrodrigues:
```suggestion
    // Nested writers are ordered from closest to furthest from output sink. Typically, this will
```

components/core/src/clp_s/log_converter/LogSerializer.hpp
```diff
@@ -88,26 +104,28 @@ class LogSerializer {
...
+    // look like `FileWriter` <- `Compressor`.
```
kirkrodrigues, line 129:
```suggestion
    // look like `FileWriter` <- `Compressor`.
    // NOTE: This class depends on there being at least one writer in `m_nested_writers` at all times.
```

kirkrodrigues (approved):

PR #2242, opened 2026-05-04
refactor(integration-tests): Split `ExternalAction` into `ClpAction` and `NonClpAction`; remove `subprocess_utils` and `logging_utils` modules.

quinntaylormitchell:
# Description

This PR migrates the testing suite with respect to the new designs implemented in #2182 (`ExternalAction`, `CmdMsg`). It further amends the subprocess interaction layer by splitting certain attributes of `ExternalAction` into two derived classes: `ClpAction` and `NonClpAction`. Call sites across binary tests, package tests, fixtures, and utilities are migrated to the new classes, and `verify_package_compression()` now returns a `VerificationResult` (new class) instead of failing internally.

The obsolete `subprocess_utils.py` and `logging_utils.py` modules are removed, and the filesystem comparison helpers move into a new module (`fs_validation.py`).

# Checklist

  breaking change.

# Validation performed

Ran `uv run pytest -m 'package'` and `uv run pytest -m 'core'`; all tests pass.

## Summary by CodeRabbit

* **Tests**
  * Standardized external command handling across integration tests with explicit return-code checks and clearer fail-fast messages for more reliable, diagnosable runs.
  * Improved extraction and validation of log/event outputs to strengthen test assertions.

* **Chores**
  * Reorganized and consolidated test utilities, removing a legacy subprocess helper.
  * Added filesystem and JSON structural comparison utilities to enhance verification.

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/clp/pull/2242)


integration-tests/tests/utils/classes.py
```diff
@@ -155,10 +170,15 @@ class ExternalAction:
     def __post_init__(self) -> None:
         """Execute the external action and log output."""
         if not self.cmd:
-            pytest.fail("Cannot create `ExternalAction` object: `cmd` list is empty.")
+            pytest.fail(f"Cannot create '{type(self).__name__}' object: `cmd` list is empty.")
+
         self.completed_proc = self._run_subprocess()
         self._log_action_summary_to_file()
 
+    def get_output(self) -> str:
```
Bill-hbrhbr, line 186: what's the use case for this one?

integration-tests/tests/utils/classes.py
```diff
@@ -155,10 +170,15 @@ class ExternalAction:
     def __post_init__(self) -> None:
         """Execute the external action and log output."""
         if not self.cmd:
-            pytest.fail("Cannot create `ExternalAction` object: `cmd` list is empty.")
+            pytest.fail(f"Cannot create '{type(self).__name__}' object: `cmd` list is empty.")
```
```suggestion
            err_msg = f"Cannot create '{type(self).__name__}' object: `cmd` list is empty."
            raise ValueError(err_msg)
```
If the cmd list is empty, it is a bug of the integration test system rather than a CLP correctness failure.
Raising a non pytest failure would be more appropriate.

integration-tests/tests/utils/classes.py
```diff
@@ -139,9 +142,21 @@ def logs_path(self) -> Path:
...
+class CmdArgs(BaseModel, ABC):
+    """Abstract base class for all CLP command argument models."""
+
+    @abstractmethod
+    def to_cmd(self) -> list[str]:
+        """:return: list of command arguments constructed from this instance's data members."""
+
+
 @dataclass
 class ExternalAction:
```
Bill-hbrhbr, line 157:
```suggestion
class ExternalAction(ABC):
```
If the `ExternalAction` should not be instantiated directly, we should just make it derive from `ABC`.

We would also need an abstract method. `format_failure_msg` seems like a good choice.

Also, could you unify the names `check_returncode` and `verify_returncode`? The return type can take a union type of `None` and `VerificationResult`

integration-tests/tests/utils/classes.py
```diff
@@ -139,9 +142,21 @@ def logs_path(self) -> Path:
         return self.dataset_root_dir / self.metadata.logs_subdir
 
 
+class CmdArgs(BaseModel, ABC):
+    """Abstract base class for all CLP command argument models."""
```
```suggestion
    """
    Abstract base class for structured CLP command argument models.
    
    Derived classes define typed command-line options and convert their state into a list of
    subprocess arguments through `to_cmd()`.
    """
```

integration-tests/tests/utils/classes.py
```diff
@@ -230,9 +250,161 @@ def _log_action_summary_to_file(self) -> None:
...
+        """:return: A `ClpAction` whose `cmd` is derived from `args.to_cmd()`."""
+        return cls(cmd=args.to_cmd(), args=args)
+
+    def __post_init__(self) -> None:
+        """
+        Validate `args`/`cmd` agreement when both are provided during construction. Then execute the
+        action.
+        """
+        if self.args is not None and self.cmd != self.args.to_cmd():
+            pytest.fail("Cannot create `ClpAction` object: `cmd` does not match `args.to_cmd()`.")
```
Bill-hbrhbr: Similary I feel like this should be an integration test bug instead of a pytest failure.

Bill-hbrhbr: Also, if we force users to use factory methods `from_args` and `from_cmd`, this shouldn't be a problem. There are various ways to implement this, but maybe checking that args and cmd match (what you currently have) is the easiest.

integration-tests/tests/utils/classes.py
```diff
@@ -230,9 +250,161 @@ def _log_action_summary_to_file(self) -> None:
...
+        :return: The formatted failure message.
+        """
+        msg = f"{reason} See subprocess log at: '{self.log_file_path}'."
+        if related_action is not None:
+            related_exe = Path(related_action.cmd[0]).name
+            msg += (
+                f" This failure occurred during the verification of '{related_exe}';"
+                f" see related subprocess log at: '{related_action.log_file_path}'."
+            )
+        return msg
```
Bill-hbrhbr: Move this to the base class.

integration-tests/tests/utils/classes.py
```diff
@@ -230,9 +250,161 @@ def _log_action_summary_to_file(self) -> None:
...
+        :return: The formatted failure message.
+        """
+        msg = f"{reason} See subprocess log at: '{self.log_file_path}'."
+        if supporting_action is not None:
+            supporting_exe = Path(supporting_action.cmd[0]).name
+            msg += (
+                f" See supporting subprocess ({supporting_exe}) log at:"
+                f" '{supporting_action.log_file_path}'."
+            )
+        return msg
```
Bill-hbrhbr: Move this to the base class.

Bill-hbrhbr (changes requested):
First round pass. Next pass will likely be my final one.

integration-tests/tests/utils/classes.py
quinntaylormitchell: I agree in concept; I'm gonna do this:


```suggestion
    """
    Abstract base class for all CLP command argument models. Derived classes define typed
    command-line options, and `to_cmd()` converts the options into a list of subprocess arguments.
    """
```

integration-tests/tests/utils/classes.py
quinntaylormitchell: Agreed

integration-tests/tests/utils/classes.py
quinntaylormitchell: Some clp commands output their results on `stdout` and some of them output results on `stderr`; analysing the full output during verification is less error-prone than looking at only one or the other, so this method makes it simple to get both.

integration-tests/tests/utils/classes.py
quinntaylormitchell: Agreed about the change to `ValueError`; I think that this check is the easiest way to force it.

integration-tests/tests/utils/classes.py
quinntaylormitchell: **re. `ExternalAction` abstract and `format_failure_msg` abstract method:** the two `format_failure_msg` methods are designed for different situations because of the `dependent_action` vs `supporting_action` parameters. `NonClpAction.format_failure_msg.dependent_action` is used when there is some other action whose verification depends on this action; `ClpAction.format_failure_msg.supporting_action` is used when there is some other action that caused the failure of this action. It’s a subtle difference but I think it’s an important difference, because otherwise there isn’t really any other way to allow nested actions. 

That being said, your comment made me realize that `format_failure_msg` is only called within this module. I’ve changed a couple things with regard to your comment and that realization:

1. added `if type(self) is ExternalAction: raise TypeError(...)` check in `ExternalAction.__post_init__()`  to make sure that no one instantiates `ExternalAction` directly
2. changed `format_failure_msg` to `_format_failure_msg`, because it wasn’t clear that those methods will only be called from within `classes.py`. This gave me the idea to remove the `ok()` and `fail()` constructors from `ClpVerificationResult` and fold them into `ClpAction` as `pass_verification()` and `fail_verification()`. I think this change makes the success/failure flow for the CLP verification functions more intuitive because everything is connected to the action itself. It has the added bonus of insulating `_format_failure_msg` from the rest of the test code, which is a good thing because the nuances are difficult to understand.

---

**re. unification of `check_returncode` and `verify_returncode`:** I want these to remain separate because they have different purposes; `check_returncode`  is a structural function that checks an assumption that the testing system makes (namely that everything other than CLP works perfectly); `verify_returncode` is a CLP verification function that doesn’t assume anything about what’s going on with CLP. Keeping them separate helps to encode the check-vs-verify flow.

integration-tests/tests/utils/classes.py
quinntaylormitchell: see above

integration-tests/tests/utils/classes.py
quinntaylormitchell: see above

integration-tests/tests/binary_tests/test_log_converter.py
```diff
@@ -7,12 +7,11 @@
 
 import pytest
 
-from tests.utils.classes import IntegrationTestPathConfig, SampleDataset
+from tests.utils.classes import ClpAction, IntegrationTestPathConfig, NonClpAction, SampleDataset
```
Bill-hbrhbr:
```suggestion
from tests.utils.classes import (
    ClpAction,
    IntegrationTestPathConfig,
    NonClpAction,
    SampleDataset,
)
```

integration-tests/tests/utils/classes.py
```diff
@@ -230,9 +258,160 @@ def _log_action_summary_to_file(self) -> None:
...
+
+        reason = (
+            f"The '{Path(self.cmd[0]).name}' subprocess returned a bad return code"
+            f" ({self.completed_proc.returncode})."
+        )
+        return self.fail_verification(reason)
+
+    def pass_verification(self) -> ClpVerificationResult:
+        """:return: A successful `ClpVerificationResult`."""
+        return ClpVerificationResult(success=True)
```
Bill-hbrhbr, line 369: I feel like this method can be abstracted away. Just one extra layer that ultimately returns a pass status.

integration-tests/tests/utils/classes.py
```diff
@@ -230,9 +258,160 @@ def _log_action_summary_to_file(self) -> None:
...
+        """
+        failure_message = self._format_failure_msg(reason, supporting_action=supporting_action)
+        logger.warning(failure_message)
+        return ClpVerificationResult(success=False, failure_message=failure_message)
+
+    def _format_failure_msg(
+        self,
+        reason: str,
+        supporting_action: ExternalAction | None = None,
+    ) -> str:
```
Bill-hbrhbr, line 394: sure let's keep distinct `_format_failure_msg` impl in `ClpAction` and `NonClpAction`. Can you add back the docstring for this function?

integration-tests/tests/binary_tests/test_identity_transformation.py
```diff
@@ -54,10 +53,16 @@ def test_clp_identity_transform(
         src_path,
     ]
     # fmt: on
-    run_and_log_subprocess(compression_cmd)
+    compression_action = ClpAction.from_cmd(compression_cmd)
+    compression_result = compression_action.verify_returncode()
+    if not compression_result:
+        pytest.fail(compression_result.failure_message)
```
Bill-hbrhbr, line 59:
```suggestion
    ClpAction.from_cmd(compression_cmd).verify_default_success()
```
This pattern appears frequently enough in this PR that I think we should add a small `ClpAction` helper for the common case where the command is expected to return 0.
```
  def verify_default_success(self) -> None:
      """
      Verify this CLP action using the default success criteria.

      The default criteria require the subprocess to exit with return code 0 and do not attach any
      supporting-action context to the failure message. Fails the current test if verification fails.
      """
      result = self.verify_returncode()
      if not result:
          pytest.fail(result.failure_message)
```

integration-tests/tests/binary_tests/test_log_converter.py
```diff
@@ -71,12 +74,18 @@ def _convert_and_compress(
...
     if test_paths.num_log_events is None:
         return
 
-    output = run_and_log_subprocess([clp_s_bin_path, "s", compression_path, "timestamp > 0"])
-    lines = output.stdout.splitlines() if output.stdout else []
+    search_action = ClpAction.from_cmd([clp_s_bin_path, "s", compression_path, "timestamp > 0"])
+    search_result = search_action.verify_returncode()
+    if not search_result:
+        pytest.fail(search_result.failure_message)
+    lines = search_action.completed_proc.stdout.splitlines()
```
Bill-hbrhbr, line 91:
```suggestion
    search_action = ClpAction.from_cmd([clp_s_bin_path, "s", compression_path, "timestamp > 0"])
    search_action.verify_default_success()
    lines = search_action.completed_proc.stdout.splitlines()
```

integration-tests/tests/utils/classes.py
```diff
@@ -230,9 +258,160 @@ def _log_action_summary_to_file(self) -> None:
...
+        supporting_action: ExternalAction | None = None,
+    ) -> ClpVerificationResult:
+        """
+        Build a failed `ClpVerificationResult` for this action and log a warning. When this action's
+        verification has failed as a direct result of some other `ExternalAction`, that action
+        should be passed into `supporting_action` so that the path to its log file is included in
+        the failure message.
+
+        :param reason: A description of the failure.
+        :param supporting_action: A previous action that caused this failure.
```
Bill-hbrhbr, line 383: Do you think `upstream_action` would be a better name?
```
:param upstream_action: A previous action that produced output or state consumed by this action.
```

Bill-hbrhbr (changes requested):
Should look good to me after this round.

Afterwards, Junhao can take a look at whether the docstrings make sense at a high level, since he has less implementation context.

integration-tests/tests/utils/classes.py
quinntaylormitchell: I put that method there for the sake of symmetricity, because it's less awkward to use in code; consider a common use case in some verification function, Option A using the method as written and Option B if the method gets abstracted away:

Option A:

```python
# Compare actual result with expected result.
if actual_result == expected_result:
    return action.pass_verification()

return action.fail_verification("Actual result didn't match expected result.")
```

Option B:

```python
# Compare actual result with expected result.
if actual_result == expected_result:
    return ClpVerificationResult(success=True)

return action.fail_verification("Actual result didn't match expected result.")
```

I like Option A better.

integration-tests/tests/utils/classes.py
quinntaylormitchell: My concern is that it'll be unclear whether `upstream_action` means that the action occurred before this action chronologically, or if it means that the `upstream_action` depends on this current action. I think that `supporting_action` and `dependent_action` are less ambiguous.

integration-tests/tests/utils/classes.py
quinntaylormitchell: Good catch thanks

integration-tests/tests/binary_tests/test_identity_transformation.py
quinntaylormitchell: A couple things:

1. To me, a `verify_default_success()` method would encroach on the principle that the `ClpAction` objects shouldn't assume anything about the result of an action. I don't think it's good testing practice to have a "default success".
2. The `if ... pytest.fail()` in this file should really be an `assert` statement, that was part of my intention in the `ClpVerificationResult` design. I forgot to change it over (same with other similar sides in the code); I've changed it now.

integration-tests/tests/binary_tests/test_log_converter.py
quinntaylormitchell: sure

integration-tests/tests/binary_tests/test_log_converter.py
quinntaylormitchell: see above

Bill-hbrhbr (dismissed):
Passing onto @junhaoliao

integration-tests/tests/utils/classes.py
```diff
@@ -230,9 +258,170 @@ def _log_action_summary_to_file(self) -> None:
...
+    assumes that all non-CLP programs are functioning properly, and so `check_returncode()` raises
+    `RuntimeError` if this is not the case.
+    """
 
-    @abstractmethod
-    def to_cmd(self) -> list[str]:
-        """:return: list of command arguments constructed from this instance's data members."""
+    def check_returncode(
+        self,
+        good_returncodes: tuple[int, ...] = (0,),
```
junhaoliao: arguably, "good" seems a vague term in defining some return code.

`expected_returncodes` might be vague too, as in we can expect things to fail. would allowed_returncodes or acceptable_returncodes or success_returncodes be more accurate?

integration-tests/tests/utils/classes.py
junhaoliao: it does sound a better idea to avoid callers constructing ClpVerificationResult instances directly

integration-tests/tests/utils/classes.py
```diff
@@ -139,9 +142,24 @@ def logs_path(self) -> Path:
         return self.dataset_root_dir / self.metadata.logs_subdir
 
 
+class CmdArgs(BaseModel, ABC):
+    """
+    Abstract base class for all CLP command argument models. Derived classes define typed
+    command-line options, and `to_cmd()` converts the options into a list of subprocess arguments.
```
junhaoliao: "Derived classes" -> "Subclasses"
```suggestion
    Abstract base class for all CLP command argument models. Subclasses define typed command-line
    options, and `to_cmd()` converts the options into a list of subprocess arguments.
```

integration-tests/tests/utils/classes.py
```diff
@@ -230,9 +258,170 @@ def _log_action_summary_to_file(self) -> None:
...
+    """Outcome returned from functions that verify CLP functionality."""
+
+    #: Whether or not the verification was successful.
+    success: bool
+
+    #: Message describing the failure, if the verification failed.
+    failure_message: str = ""
+
+    def __bool__(self) -> bool:
+        """Makes class truthy."""
```
junhaoliao: this might be more meaningful?
```suggestion
        """:return: Whether the verification succeeded."""
```

junhaoliao (changes requested):
lgtm

for the title, how about `deprecate` -> `Remove`, since we aren't really phasing anything out here

integration-tests/tests/utils/classes.py
quinntaylormitchell: agreed

integration-tests/tests/utils/classes.py
quinntaylormitchell: I like `success_returncodes`, agreed

integration-tests/tests/utils/classes.py
quinntaylormitchell: agreed

junhaoliao (approved):

PR #2246, opened 2026-05-05
feat(webui): Make max search results configurable via settings.json

goynam:
## Summary

Makes the maximum number of search results configurable end-to-end using the existing `settings.json` pattern (following the `MaxDatasetsPerQuery` precedent), addressing all feedback from the initial review.

### Configuration Flow

`clp-config.yaml` → `controller.py` → `settings.json` (client + server) → application code

| Layer | File | Change |
|-------|------|--------|
| Config definition | `clp_config.py` | Add `max_search_results: PositiveInt = 1000` to `WebUi` |
| Config template | `clp-config.template.json.yaml` | Add commented `max_search_results: 1000` |
| Propagation | `controller.py` | Add `MaxSearchResults` to both client & server `settings_json_updates` |
| Client settings template | `client/public/settings.json` | Add `"MaxSearchResults": 1000` |
| Client types | `client/src/settings.ts` | Add `MaxSearchResults: number` to `Settings` type |
| Client config export | `client/src/config/index.ts` | Export `SETTINGS_MAX_SEARCH_RESULTS` |
| Client native search | `SearchResultsTable/typings.ts` | Use `SETTINGS_MAX_SEARCH_RESULTS` instead of hardcoded 1000 |
| Client Presto guided | `presto-guided-search-requests.ts` | Use `SETTINGS_MAX_SEARCH_RESULTS` instead of hardcoded 1000 |
| Server settings template | `server/settings.json` | Add `"MaxSearchResults": 1000` |
| Server search | `search/typings.ts` | Read from `settings.MaxSearchResults` instead of hardcoded 1000 |
| Server Presto search | `presto-search/typings.ts` | Read from `settings.MaxSearchResults` instead of hardcoded 1000 |
| Helm values | `values.yaml` | Add `max_search_results: 1000` to `webui` section |
| Helm configmap | `configmap.yaml` | Add `MaxSearchResults` to both client and server settings JSON blocks |

### Design Decisions

- **`settings.json` over env vars**: Environment variables are reserved for credentials; general config uses `settings.json` (per existing pattern)
- **Single source of truth**: Both client and server read from the same configured value
- **No mutable global state**: Server reads directly from the imported settings JSON; client reads at startup via config module
- **All search paths covered**: Native search, Presto search (server-side cap), and Presto guided SQL limit all respect the setting

## Checklist

* [x] The PR satisfies the [contribution guidelines](https://docs.yscope.com/dev-guide/contrib-guides-overview.html).
* [x] This is not a breaking change (default remains 1000).
* [x] Necessary docs have been updated (config template updated).

## Validation performed

- Verified all hardcoded `1000` constants across search paths are replaced with the configurable value
- Verified backward compatibility: existing configs without `max_search_results` use the default of 1000
- Verified the same pattern as `MaxDatasetsPerQuery` is followed throughout

## Summary by CodeRabbit

* **New Features**
  * Search result limits are now configurable via settings (default 1000) and exposed in UI.
  * Added a Max Results input on the search page so users can set per-query result limits.
  * Server now respects per-query max results while enforcing the configured cap.

* **Chores**
  * Deployment/config templates updated to include the configurable max search results.

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/clp/pull/2246)


junhaoliao:
Thanks for working on this! To give some context: the max result count isn't really a proper mechanism for scalability - we'll likely drop it entirely in the future - but in the meantime, making it configurable is the right step, and it's great to see someone take it on.

I've been thinking through the same problem and wanted to share some design considerations for the approach. The core difference is around **how the configuration flows through the system** and **how complete the coverage is**.

## 1. Configuration mechanism: `settings.json` vs environment variable

In the webui, environment variables are used for credentials (e.g., database passwords, AWS keys), while `settings.json` handles general configuration (e.g., `ClpQueryEngine`, `MaxDatasetsPerQuery`, `MongoDbSearchResultsMetadataCollectionName`). Since max search results is a general config value rather than a credential, it fits naturally in `settings.json`.

This also matters for the next point — `settings.json` is the mechanism that delivers config to **both** the client and the server from a single source.

## 2. Client-side coverage

This PR only configures the server-side limit, but there's also a **client-side cursor limit** that independently caps displayed results at 1000:

- `client/src/pages/SearchPage/SearchResults/SearchResultsTable/typings.ts` — `SEARCH_MAX_NUM_RESULTS = 1000`
- Used in `useSearchResults.ts` (MongoDB cursor `limit`) and `usePrestoSearchResults.ts` (same)

Even if the server is configured to store more results, the client will only fetch and display 1000 because the cursor limit is still hardcoded. For the feature to work end-to-end, the client needs to receive the same configured value — which is where `settings.json` naturally delivers it to both sides.

## 3. Presto search coverage

There are two additional hardcoded limits in the Presto search path that should also be configurable:

- `server/src/routes/api/presto-search/typings.ts` — `MAX_PRESTO_SEARCH_RESULTS = 1000` (caps MongoDB storage)
- `client/src/pages/SearchPage/SearchControls/Presto/Guided/presto-guided-search-requests.ts` — `DEFAULT_SEARCH_LIMIT = 1000` (SQL `LIMIT` clause in guided mode)

If the goal is "max search results is configurable," all three search paths (CLP server, Presto server, Presto guided SQL) should respect the same setting.

## 4. Avoiding mutable global state

The PR introduces `let SEARCH_MAX_NUM_RESULTS` with a `setSearchMaxNumResults()` setter. Since `settings.json` is already imported at module load time in `search/index.ts`, the value can be read directly as `settings.MaxSearchResults` — no mutable state or setter function needed. This keeps the code simpler and consistent with how the server reads other settings.

## 5. Standalone/Docker deployment coverage

The PR adds the Helm deployment path, but the standalone/Docker deployment (via `clp-package`) uses `controller.py`'s `_set_up_env_for_webui()` to generate `settings.json` from `clp-config.yaml`. That path needs updating too:

- Add `max_search_results` to the `WebUi` class in `clp_config.py`
- Add `"MaxSearchResults"` to `client_settings_json_updates` and `server_settings_json_updates` in `controller.py`
- Add the key to both `settings.json` template files (required because `_read_and_update_settings_json` validates that keys exist before updating)
- Update `components/package-template/src/etc/clp-config.template.json.yaml`

## Suggested approach (summary)

Following the `MaxDatasetsPerQuery` precedent:

| Step | File | Change |
|------|------|--------|
| Config definition | `clp_config.py` | Add `max_search_results: PositiveInt = 1000` to `WebUi` |
| Templates | `client/public/settings.json`, `server/settings.json` | Add `"MaxSearchResults": 1000` |
| Propagation | `controller.py` | Add to both client & server `settings_json_updates` |
| Client types | `client/src/settings.ts` | Add `MaxSearchResults: number` to `Settings` |
| Client config | `client/src/config/index.ts` | Export `SETTINGS_MAX_SEARCH_RESULTS` |
| Client usage (3 sites) | `typings.ts`, `useSearchResults.ts`, `usePrestoSearchResults.ts` | Replace hardcoded constant with `SETTINGS_MAX_SEARCH_RESULTS` |
| Presto guided | `presto-guided-search-requests.ts` | Replace `DEFAULT_SEARCH_LIMIT` with `SETTINGS_MAX_SEARCH_RESULTS` |
| Server usage (3 sites) | `search/typings.ts`, `search/index.ts`, `search/utils.ts`, `presto-search/typings.ts`, `presto-search/index.ts` | Use `settings.MaxSearchResults` instead of hardcoded/mutable constants |
| Package template | `clp-config.template.json.yaml` | Add commented `max_search_results: 1000` |
| Helm | `values.yaml`, `configmap.yaml` | Add `max_search_results` to webui section and both settings JSON blocks |

This covers all search paths end-to-end, uses the existing config flow, and avoids introducing mutable global state.

Happy to help if you'd like to iterate on any of this!

components/webui/server/settings.json
```diff
@@ -28,6 +28,7 @@
 
     "ClpQueryEngine": "clp-s",
     "ClpStorageEngine": "clp-s",
+    "MaxSearchResults": 1000,
```
junhaoliao: should be placed in the `Mongo...` group. within the group, properties should be alphabetically sorted

components/webui/client/src/pages/SearchPage/SearchControls/MaxResultsInput/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+    const updateMaxNumResults = useSearchStore((state) => state.updateMaxNumResults);
+
+    return (
+        <Space.Compact>
+            <Typography.Text
+                style={{
+                    alignSelf: "center",
+                    marginRight: 4,
+                    whiteSpace: "nowrap",
+                }}
```
junhaoliao: extract into a css class

components/webui/client/src/pages/SearchPage/SearchControls/MaxResultsInput/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+            >
+                {"Limit"}
+            </Typography.Text>
+            <InputNumber
+                changeOnBlur={true}
+                max={SETTINGS_MAX_SEARCH_RESULTS}
+                min={1}
+                size={"middle"}
+                style={{width: 100}}
```
junhaoliao: ditto - extract into a css class

components/webui/server/src/routes/api/search/index.ts
```diff
@@ -65,21 +65,26 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
                 datasets,
+                maxNumResults,
                 timestampBegin,
                 timestampEnd,
                 ignoreCase,
                 timeRangeBucketSizeMillis,
                 queryString,
             } = request.body;
 
+            const effectiveMaxResults = "undefined" === typeof maxNumResults ?
```
junhaoliao: How about renaming this to `clampedMaxNumResults`? It makes the server-side behavior clearer: the client-supplied `maxNumResults` is used only up to `SEARCH_MAX_NUM_RESULTS`.

components/webui/client/src/pages/SearchPage/SearchControls/MaxResultsInput/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+
+/**
+ * Renders an input for setting the maximum number of search results per query.
+ *
+ * @return
+ */
+const MaxResultsInput = () => {
+    const maxNumResults = useSearchStore((state) => state.maxNumResults);
+    const updateMaxNumResults = useSearchStore((state) => state.updateMaxNumResults);
```
junhaoliao: this doesn't have to be a React state. i.e., we should write

```ts
const handleInputChange = useCallback((value) => {
    if (null !== value) {
        const {updateMaxNumResults} = useSearchStore.getState();
        updateMaxNumResults(value);
    }
}, []);

<InputNumber
    onChange={handleInputChange}
/>
```

to avoid treating `updateMaxNumResults` as a React state variable, as its definition practically never changes

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/typings.ts
```diff
@@ -6,7 +9,7 @@ const TABLE_BOTTOM_PADDING = 95;
 /**
  * The maximum number of results to retrieve for a search.
  */
-const SEARCH_MAX_NUM_RESULTS = 1000;
+const SEARCH_MAX_NUM_RESULTS = SETTINGS_MAX_SEARCH_RESULTS;
```
junhaoliao: this is super confusing and defeats the purpose of defaulting to a lower limit on the client side, and is potentially incorrect.

as a result of this modification, now:
1. when we submit a query, we use the MaxResultsInput-controlled Zustand state store SearchStore 's variable `maxNumResults` (A), which is correct.
2. as the search performs, different workers dump maximum `maxNumResults` (A) into the result cache. assume the query needs to scan results from 10 archives -> 10 workers can dump up to (A)*10 results into the result cache.
3. we always subscribe to the result cache with `limit: SEARCH_MAX_NUM_RESULTS`(B), pulling this `SEARCH_MAX_NUM_RESULTS`(B) number of results from the server and display in the client

now suppose A (client-session-only, requested) is 1,000 and B (server-enforced-maximum) is 10,000 , the effective max number of results to be displayed in the client is 10,000. which doesn't correctly match the client requested number.

i think we can define a default value here to be used in `SEARCH_STATE_DEFAULT`. 

```suggestion
/**
 * The default maximum number of results to retrieve for a search.
 */
const DEFAULT_SEARCH_MAX_NUM_RESULTS = Math.min(1000, SETTINGS_MAX_SEARCH_RESULTS);
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -763,6 +764,7 @@ def _set_up_env_for_webui(self, container_clp_config: ClpConfig) -> EnvVarsDict:
             "ArchiveOutputTargetSegmentSize": self._clp_config.archive_output.target_segment_size,
             "ClpQueryEngine": self._clp_config.package.query_engine,
             "ClpStorageEngine": self._clp_config.package.storage_engine,
+            "MaxSearchResults": self._clp_config.webui.max_search_results,
```
junhaoliao: this definition should follow the order in components/webui/server/settings.json , after the order is corrected there

components/webui/client/src/pages/SearchPage/SearchState/index.tsx
```diff
@@ -23,6 +24,7 @@ import {SEARCH_UI_STATE} from "./typings";
  */
 const SEARCH_STATE_DEFAULT = Object.freeze({
     aggregationJobId: null,
+    maxNumResults: SETTINGS_MAX_SEARCH_RESULTS,
```
junhaoliao: see the comment in https://github.com/y-scope/clp/pull/2246/changes#r3214638646

components/webui/common/src/schemas/search.ts
```diff
@@ -15,6 +15,7 @@ import {
 const QueryJobCreationSchema = Type.Object({
     datasets: Type.Array(Type.String()),
     ignoreCase: Type.Boolean(),
+    maxNumResults: Type.Optional(Type.Integer({minimum: 1})),
```
junhaoliao: since this API is not public, i think we should not make it optional.

components/webui/server/src/routes/api/search/index.ts
junhaoliao: following the comment at https://github.com/y-scope/clp/pull/2246/changes#r3214643549 , we don't have to do this undefined check here if we don't make maxNumResults Optional

junhaoliao (changes requested):

components/webui/server/settings.json
goynam: Done. Moved MaxSearchResults before MongoDbHost so it's alphabetically sorted within that group.

components/webui/client/src/pages/SearchPage/SearchControls/MaxResultsInput/index.tsx
goynam: Done. Created index.module.css with .label and .input classes, replaced both inline style props. Fixed in 975479e.

components/webui/server/src/routes/api/search/index.ts
goynam: Done. Renamed to clampedMaxNumResults.

components/webui/client/src/pages/SearchPage/SearchControls/MaxResultsInput/index.tsx
goynam: Done. Switched to useCallback with useSearchStore.getState() pattern — updateMaxNumResults is no longer subscribed as a React state variable.

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/typings.ts
goynam: Good point. The client cursor limit (SEARCH_MAX_NUM_RESULTS) and the per-query submit limit (maxNumResults in the store) serve different roles: the cursor limit caps what MongoDB returns to the UI, while the per-query limit caps what the server tells workers to produce. Both are set to
SETTINGS_MAX_SEARCH_RESULTS so the cursor can display all results the server produces. If we want a lower default for the per-query input, we can change the store default independently — happy to adjust if you have a preferred value.

components/clp-package-utils/clp_package_utils/controller.py
goynam: Done

components/webui/client/src/pages/SearchPage/SearchState/index.tsx
goynam: Acknowledged

components/webui/common/src/schemas/search.ts
goynam: Acknowledged

components/webui/server/src/routes/api/search/index.ts
goynam: Done

components/webui/client/src/pages/SearchPage/SearchResults/SearchResultsTable/typings.ts
junhaoliao: sorry i think my previous message has confused you - 

i mean to say this shall be renamed to `DEFAULT_SEARCH_MAX_NUM_RESULTS` to correctly reflect the purpose, as it's only used as an init value in `SEARCH_STATE_DEFAULT.maxNumResults`

then we also want to clamp this default value to a safer number at 1000, so that most queries won't blow up the server even when the ClpConfig.webui.max_search_results  has been (incorrectly) configured with an aggressive value.

for example, if ClpConfig.webui.max_search_results is configured as 10,000 in the cluster, we might not want every query to return 10,000 results. The above change proposes a default of 1,000; if an end user ever needs more than 1,000 results, they can always change the MaxResultsInput to a higher number in the browser for that specific query session, without affecting most other queries.

overall:
1. we propose a default of max 1000 results for all queries
2. the intention to make the max number results configurable is to give end users freedom to temporarily limit the number to a higher value, for specific investigations
3. ClpConfig.webui.max_search_results serves as a safety net to disallow crazy values from an end user

components/webui/client/src/pages/SearchPage/SearchState/index.tsx
junhaoliao:
```suggestion
    maxNumResults: DEFAULT_SEARCH_MAX_NUM_RESULTS,
```

components/webui/client/src/pages/SearchPage/SearchControls/MaxResultsInput/index.module.css
junhaoliao, line 1: @hoophalab please help confirm the interactions and visuals match your expectations

components/webui/client/src/pages/SearchPage/SearchControls/Presto/Guided/presto-guided-search-requests.ts
junhaoliao, line 1: following the suggestion at https://github.com/y-scope/clp/pull/2246/changes#r3214638646 , i believe we shall update this file too.

@hoophalab please help confirm the exact changes needed

junhaoliao: and whether the presto flow is correctly updated

components/webui/client/src/pages/SearchPage/SearchState/index.tsx
junhaoliao: sorry for missing this in the previous round of review - there is a naming mismatch "max num results" vs "max search results"

i believe running this should be sufficient to globally rename all variables / constants:
```bash
git grep -Ilz -e 'MaxSearchResults' -e 'max_search_results' -e 'SETTINGS_MAX_SEARCH_RESULTS' -e 'SEARCH_MAX_NUM_RESULTS' -e 'MAX_PRESTO_SEARCH_RESULTS' -e 'maxNumResults' -e 'updateMaxNumResults' -e 'clampedMaxNumResults' -- . \
  | xargs -0 -r perl -0pi -e 's/SETTINGS_MAX_SEARCH_RESULTS/SETTINGS_MAX_NUM_SEARCH_RESULTS/g; s/SEARCH_MAX_NUM_RESULTS/SEARCH_MAX_NUM_SEARCH_RESULTS/g; s/MAX_PRESTO_SEARCH_RESULTS/MAX_NUM_PRESTO_SEARCH_RESULTS/g; s/MaxSearchResults/MaxNumSearchResults/g; s/max_search_results/max_num_search_results/g; s/updateMaxNumResults/updateMaxNumSearchResults/g; s/clampedMaxNumResults/clampedMaxNumSearchResults/g; s/maxNumResults/maxNumSearchResults/g'
```

tools/deployment/package-helm/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: "v2"
 name: "clp"
-version: "0.3.2-dev.4"
```
junhaoliao: the base seems outdated. we need to merge origin/main into the current branch

PR #2263, opened 2026-05-08
feat(helm): Add configurable resource limits for components (resolves #2250).

### LINKED ISSUE #2250: Add configurable CPU and memory resource limits to Helm chart and Docker Compose
## Request

Add configurable CPU and memory resource limits to both the Helm chart and Docker Compose deployment files.

Currently, neither deployment method specifies resource limits for containers:
- **Docker Compose**: No `deploy.resources.limits`, `mem_limit`, or `cpus` directives exist in any docker-compose files
- **Helm Chart**: Only PVC storage requests are defined; no container-level `limits:` or `requests:` for CPU/memory exist in any deployment template

This is important because:
1. Without resource limits, containers can consume unbounded resources, leading to resource contention in production environments
2. In Kubernetes, pods without limits may be evicted or cause node instability
3. Proper resource limits are essential for production readiness and operational stability
4. Enables better capacity planning and cost management

## Possible implementation

**Config schema (shared for both Helm and Docker Compose):**
- Add optional `resources` field inside each component's config block in `clp_config.py`. When not specified, sensible default limits will be used.
- Example with resources specified (Kubernetes-style with requests/limits):
  ```yaml
  database:
    type: "mariadb"
    host: "localhost"
    port: 30306
    resources:
      requests:
        cpu: "1"
        memory: "1G"
      limits:
        cpu: "2"
        memory: "2G"
  ```
- For Docker Compose, only `limits` will be used since it doesn't support `requests` (no scheduler)
  ```yaml
  database:
    type: "mariadb"
    host: "localhost"
    port: 30306
    resources:
      requests:
        cpu: "1"
        memory: "1G"
      limits:
        cpu: "2"
        memory: "2G"

  queue:
    host: "localhost"
    port: 5672
    resources:
      requests:
        cpu: "0.5"
        memory: "512M"
      limits:
        cpu: "1"
        memory: "1G"
  ```
- Apply same pattern to: `redis`, `results_cache`, `compression_scheduler`, `query_scheduler`, `compression_worker`, `query_worker`, `reducer`, `webui`, `api_server`, `log_ingestor`, `mcp_server`, `garbage_collector`, `presto`

**Helm Chart:**
- Update `values.yaml` to include resource limits per component under `clpConfig.*.resources`
- Update deployment/statefulset templates to render both `requests` and `limits`:
  ```yaml
  {{- with .Values.clpConfig.database.resources }}
  resources:
    {{- with .requests }}
    requests:
      cpu: {{ .cpu }}
      memory: {{ .memory }}
    {{- end }}
    {{- with .limits }}
    limits:
      cpu: {{ .cpu }}
      memory: {{ .memory }}
    {{- end }}
  {{- end }}
  ```

**Docker Compose:**
- Update `controller.py` `set_up_env()` to read `resources.limits` from each component's config and generate env vars like `CLP_DATABASE_MEMORY_LIMIT`, `CLP_DATABASE_CPU_LIMIT` (only limits used; no requests since Docker Compose has no scheduler)
- Update `docker-compose-all.yaml` to use these env vars:
  ```yaml
  deploy:
    resources:
      limits:
        cpus: "${CLP_DATABASE_CPU_LIMIT:-1}"
        memory: "${CLP_DATABASE_MEMORY_LIMIT:-1G}"
  ```

hoophalab:
# Description

Add configurable CPU and memory resource limits to both the Helm chart and Docker Compose deployment files.

Currently, neither deployment method specifies resource limits for containers:

Docker Compose: No deploy.resources.limits, mem_limit, or cpus directives exist in any docker-compose files
Helm Chart: Only PVC storage requests are defined; no container-level limits: or requests: for CPU/memory exist in any deployment template

**Helm Chart:**
- Update `values.yaml` to include resource limits per component under `clpConfig.*.resources`
- Update deployment/statefulset templates to render both `requests` and `limits`:
  ```yaml
  {{- with .Values.clpConfig.database.resources }}
  resources:
    {{- with .requests }}
    requests:
      cpu: {{ .cpu }}
      memory: {{ .memory }}
    {{- end }}
    {{- with .limits }}
    limits:
      cpu: {{ .cpu }}
      memory: {{ .memory }}
    {{- end }}
  {{- end }}
  ```

**Docker Compose:**
- Update `controller.py` `set_up_env()` to read `resources.limits` from each component's config and generate env vars like `CLP_DATABASE_MEMORY_LIMIT`, `CLP_DATABASE_CPU_LIMIT` (only limits used; no requests since Docker Compose has no scheduler)
- Update `docker-compose-all.yaml` to use these env vars:
  ```yaml
  deploy:
    resources:
      limits:
        cpus: "${CLP_DATABASE_CPU_LIMIT:-1}"
        memory: "${CLP_DATABASE_MEMORY_LIMIT:-1G}"
  ```

# Checklist

  breaking change.

# Validation performed

kubectl and docker show the limits after configuration.

## Release Notes

* **New Features**
  * Added configurable Kubernetes resource limits (CPU and memory) for all deployment components via Helm values, enabling better resource management and allocation control in deployments.

* **Updates**
  * Helm chart version updated to `0.3.2-dev.13`.


components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -219,6 +219,14 @@ def _set_up_env_for_database(self) -> EnvVarsDict:
...
         }
 
+        # Resource limits
+        database_resources = self._clp_config.database.resources
+        if database_resources and database_resources.limits:
+            if database_resources.limits.cpu is not None:
+                env_vars["CLP_DATABASE_CPU_LIMIT"] = database_resources.limits.cpu
+            if database_resources.limits.memory is not None:
+                env_vars["CLP_DATABASE_MEMORY_LIMIT"] = database_resources.limits.memory
+
```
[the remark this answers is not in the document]
junhaoliao: re the missing env var issue - this seems to be valid

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -406,6 +430,14 @@ def _set_up_env_for_spider_scheduler(self) -> EnvVarsDict:
...
         }
 
+        # Resource limits
+        spider_resources = self._clp_config.spider_scheduler.resources
+        if spider_resources and spider_resources.limits:
+            if spider_resources.limits.cpu is not None:
+                env_vars["CLP_SPIDER_SCHEDULER_CPU_LIMIT"] = spider_resources.limits.cpu
+            if spider_resources.limits.memory is not None:
+                env_vars["CLP_SPIDER_SCHEDULER_MEMORY_LIMIT"] = spider_resources.limits.memory
+
```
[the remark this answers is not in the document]
junhaoliao: this seems to be valid

owners of spider @sitaowang1998 @LinZhihao-723 please help comment if we should maintain this flow, or leave it until the new spider integration

components/clp-package-utils/clp_package_utils/controller.py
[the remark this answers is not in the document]
junhaoliao: re the refactor comment - i agree

something like this should help?
```
    @staticmethod
    def _inject_resource_limit_env_vars(
        component_name: str,
        resources: Resources | None,
        env_vars: EnvVarsDict,
    ) -> None:
        """Injects CPU and memory limit env vars for a component if configured.
 
        :param component_name: Uppercase component name for env var prefix (e.g., "DATABASE").
        :param resources:
        :param env_vars:
        """
        if resources and resources.limits:
            if resources.limits.cpu is not None:
                env_vars[f"CLP_{component_name}_CPU_LIMIT"] = resources.limits.cpu
            if resources.limits.memory is not None:
                env_vars[f"CLP_{component_name}_MEMORY_LIMIT"] = resources.limits.memory
```

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -59,6 +59,10 @@ services:
     deploy:
       # Value must be either 0 or 1. Set to 0 to disable the database.
       replicas: "${CLP_DATABASE_ENABLED:-1}"
+      resources:
+        limits:
+          cpus: "${CLP_DATABASE_CPU_LIMIT:-0}"
+          memory: "${CLP_DATABASE_MEMORY_LIMIT:-0}"
```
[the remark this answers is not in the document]
junhaoliao: instead of letting services run with unbound cpu and memory, maybe we should set recommended defaults? then users only need to override the values in the config

(we should mark the PR as BREAKING if so

tools/deployment/package-helm/values.yaml
```diff
@@ -80,6 +80,47 @@ scheduling:
...
+# Example:
+#   resources:
+#     compressionWorker:
+#       requests:
+#         cpu: "1"
+#         memory: "1Gi"
+#       limits:
+#         cpu: "2"
+#         memory: "2Gi"
+resources:
```
junhaoliao, line 119: ditto - we should configure defaults as recommended values

(then the values can be mapped directly to the individual services without a helper

components/clp-package-utils/clp_package_utils/controller.py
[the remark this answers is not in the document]
junhaoliao: re "there is no way to set them via clp-config.yaml" -

we should update the clp-config.yaml templates in this PR too?

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -251,6 +261,7 @@ class Database(BaseModel):
     ssl_cert: NonEmptyStr | None = None
     auto_commit: bool = False
     compress: bool = True
+    resources: Resources | None = None
```
junhaoliao: collocating the `resources` property with others within individual component configs seem to make sense in clp-config.yaml , but the clp-config.yaml schema will differ from the clpConfig object at the Helm values.yaml, where a `resources` property is proposed at the root level?

@kirkrodrigues what's your take on this?

junhaoliao (changes requested):

components/clp-package-utils/clp_package_utils/controller.py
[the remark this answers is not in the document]
sitaowang1998: Current Spider integration has following concerns:
* Each worker & task executor combination is supposed to consume a single core. We use the [script](https://github.com/y-scope/clp/blob/main/components/job-orchestration/job_orchestration/executor/start-spider-worker.py) to start a number of workers in a container. We should probably updated the helm and docker compose to pass in the number of cores. The first version of Spider rewrite sticks to this single core model, so the transition should be smooth. In long term, the model will change to a single execution manager per container with a pool of task executors.
* Spider scheduler is a single core process, so limit on the cpu should set to 1, with a note that the increase of limit brings little benefits. The rewritten Spider scheduler is multi-threaded, so this would change.

components/clp-py-utils/clp_py_utils/clp_config.py
kirkrodrigues: Yeah, I feel like that's going to be confusing long term. For instance, if I set `database.resources` in `values.yaml`, does that take precedence over `resources.database`?

junhaoliao: right

in helm, setting clpConfig.database.resources will not have any impact because the charts read the needed values from resources.database

so we really shouldn't use a different config schema; otherwise unintended behaviours occur especially when users migrate from docker compose to helm k8s

then do you see we should always take the value from clpConfig(clp-config.yaml).database.resources? otherwise, if you think this is more of a resources allocation concern (just like the "scheduling" key at the root of values.yaml already) and should not sit in clpConfig(clp-config.yaml), shall we put it at the root of values.yaml for k8s; on the docker compose side, we create a new file erc/resources.yaml or scheduling.yaml and put the config inside?

goynam:
Hi @hoophalab

Just a suggestion:

Compression/query workers use emptyDir volumes for temp data and staged archives — these write to the node's root disk. Without ephemeral-storage in requests/limits, the scheduler has no visibility into per-pod disk usage and can over-pack nodes until the kubelet hits its eviction threshold, causing node-wide worker eviction storms. We're seeing 30%+ of compression-worker pods evicted this way in production. 

Would it make sense to add ephemeral-storage as an option in ResourceRequirements alongside cpu/memory?

tools/deployment/package-helm/values.yaml
```diff
@@ -110,6 +110,51 @@ scheduling:
...
+
+  # Core
+  compressionScheduler: {}
+  logIngestor: {}
+  queryScheduler: {}
+  prestoCoordinator: {}
+  garbageCollector: {}
+  apiServer: {}
+  webui: {}
+  mcpServer: {}
```
junhaoliao, line 140:
```suggestion
  mcpServer: {}
  otelCollector: {}
```

tools/deployment/package-helm/values.yaml
```diff
@@ -110,6 +110,51 @@ scheduling:
...
+
+  # Jobs
+  dbTableCreator: {}
+  resultsCacheIndicesCreator: {}
+
```
junhaoliao, line 148:
```suggestion
  topologyMetricsEmitter: {}

```
also modify the job to use the limit

tools/deployment/package-helm/values.yaml
junhaoliao: discussed offline - due to current lack of understanding of CLP Package services' resource consumption, we will only provide the configuration interface without defaults (recommendations).

tools/deployment/package-helm/values.yaml
```diff
@@ -110,6 +110,51 @@ scheduling:
...
+  queryScheduler: {}
+  prestoCoordinator: {}
+  garbageCollector: {}
+  apiServer: {}
+  webui: {}
+  mcpServer: {}
+
+  # Jobs
+  dbTableCreator: {}
+  resultsCacheIndicesCreator: {}
```
junhaoliao, line 146: this should be wired

tools/deployment/package-helm/values.yaml
```diff
@@ -110,6 +110,51 @@ scheduling:
...
+  logIngestor: {}
+  queryScheduler: {}
+  prestoCoordinator: {}
+  garbageCollector: {}
+  apiServer: {}
+  webui: {}
+  mcpServer: {}
+
+  # Jobs
+  dbTableCreator: {}
```
junhaoliao, line 144:
```suggestion
  dbTableCreator: {}
  prestoWorkerSetupConfigs: {}
```

then we should correct the name in `presto-worker-deployment.yaml`

junhaoliao (changes requested):
we should update the docs too. something like this right before the `## Verifying the deployment` should be sufficient

````
### Component resources

Set resource requests and limits for each chart component under the top-level `resources` key.
Values use standard Kubernetes resource quantities:

```{code-block} yaml
:caption: resources.yaml

resources:
  compressionWorker:
    requests:
      cpu: "1"
      memory: "1Gi"
      ephemeral-storage: "2Gi"
    limits:
      cpu: "2"
      memory: "2Gi"
      ephemeral-storage: "4Gi"
```

Then install with the values file:

```bash
helm install clp clp/clp DOCS_VAR_HELM_VERSION_FLAG -f resources.yaml
```
````

**Review by junhaoliao — APPROVED** [review]
for the title, let's change the scope to `helm` since the changes are now helm-only; also it doesn't now seem the changes are breaking so probably drop the `!` too

```
feat(helm): Add configurable resource limits for components (resolves #2250).
```

PR #2269, opened 2026-05-09
fix(deployment): Use unique Celery node names per worker replica.

### LINKED ISSUE #2259: Compression workers crash-loop due to Celery pidbox conflict when using shared node name
### Bug

When running multiple compression worker pods (scaled via Kubernetes Deployment), all workers register with the same Celery node name (-n compression-worker). This causes a pidbox queue conflict in RabbitMQ, leading to periodic worker crashes (exit code 1) approximately every 60-70 minutes.

Each crash orphans in-flight compression tasks — their results never get written to Redis, and the compression scheduler never marks the corresponding jobs as completed or failed. Over time, this accumulates thousands of stuck RUNNING jobs in the database.

Symptoms:

Workers log warnings.warn(W_PIDBOX_IN_USE.format(node=self)) before crashing
Workers exit with code 1 (no OOM, no EBS issues, no error in application logs)
compression_jobs table accumulates thousands of jobs in status=1 (RUNNING) that never complete
Restarts occur every ~60-70 minutes per pod
With 100 workers × 12 concurrency, ~3000+ orphaned jobs accumulate per week

Root Cause:

The Celery worker command in the Helm deployment template uses a static node name:

celery -A job_orchestration.executor.compress worker --concurrency 12 -Q compression -n compression-worker
When multiple pods use the same -n compression-worker name, they conflict on the RabbitMQ pidbox (control) queue. RabbitMQ eventually disconnects one of the conflicting workers, causing it to exit.

Impact without fix:

Worker restarts every ~70 minutes (100 pods × ~20 restarts/day = 2000 restarts/day)
~3000+ orphaned compression jobs per week requiring manual DB cleanup
Wasted compute: tasks that were 90% complete get lost and must be reprocessed from scratch

### CLP version

0.10.1-dev

### Environment

Deployment: 100 compression worker replicas, concurrency 12
Kubernetes: EKS
RabbitMQ: bundled (single pod)

### Reproduction steps

N/A

goynam:
## Summary
- Appends `$(HOSTNAME)` to the Celery `-n` (node name) flag in both compression-worker and query-worker Helm deployment templates, making each pod register with a unique node name (e.g., `compression-worker@clp-compression-worker-abc123`)
- Fixes RabbitMQ pidbox queue conflicts and periodic crashes when running multiple worker pod replicas
- Bumps Helm chart version to 0.3.2-dev.2

Fixes #2259

## Test plan
- [ ] Deploy with `compressionWorker.replicas > 1` and verify each pod registers a unique Celery node name in RabbitMQ
- [ ] Confirm no pidbox queue conflicts in RabbitMQ management UI
- [ ] Verify compression and query tasks still execute correctly
- [ ] Check `celery inspect active` shows distinct node names per pod

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

## Summary by CodeRabbit

* **Chores**
  * Bumped Helm chart version to 0.3.2-dev.5.
  * Updated worker startup to use hostname-based node names for better operational visibility and support for multiple instances across distributed deployments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/clp/pull/2269)

Fixes https://github.com/y-scope/clp/issues/2259


tools/deployment/package-helm/templates/query-worker-deployment.yaml
```diff
@@ -87,7 +87,7 @@ spec:
             "--concurrency", "{{ .Values.workerConcurrency }}",
             "--loglevel", "WARNING",
             "-Q", "query",
-            "-n", "query-worker"
+            "-n", "query-worker@$(HOSTNAME)"
```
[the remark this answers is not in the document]
junhaoliao: as pointed out by the rabbit, the proposed fix won't work. see https://github.com/kubernetes/kubernetes/issues/57726 for details

junhaoliao: according to https://docs.celeryq.dev/en/stable/userguide/workers.html#starting-the-worker
```suggestion
            "-n", "query-worker@%h"
```

tools/deployment/package-helm/templates/compression-worker-deployment.yaml
```diff
@@ -87,7 +87,7 @@ spec:
             "--concurrency", "{{ .Values.workerConcurrency }}",
             "--loglevel", "WARNING",
             "-Q", "compression",
-            "-n", "compression-worker"
+            "-n", "compression-worker@$(HOSTNAME)"
```
junhaoliao: according to https://docs.celeryq.dev/en/stable/userguide/workers.html#starting-the-worker
```suggestion
            "-n", "compression-worker@%h"
```

tools/deployment/package-helm/templates/results-cache-statefulset.yaml
junhaoliao, line 1: changes in this file seem not related to the PR?

tools/deployment/package-helm/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: "v2"
 name: "clp"
-version: "0.3.2-dev.4"
```
junhaoliao, line 3: the git branch's base seems to be out of date. please merge origin (y-scope)'s `main` branch into the PR source branch

junhaoliao (changes requested):

tools/deployment/package-helm/templates/compression-worker-deployment.yaml
```diff
@@ -87,7 +87,7 @@ spec:
             "--concurrency", "{{ .Values.workerConcurrency }}",
             "--loglevel", "WARNING",
             "-Q", "compression",
-            "-n", "compression-worker"
+            "-n", "compression-worker@%h"
```
junhaoliao, line 90: @gibber9809 and i discussed offline - let's update https://github.com/y-scope/clp/blob/aeb077ca4/tools/deployment/package/docker-compose-all.yaml#L329 to match this command argument

though, practically, that fix won't be enough to ensure uniqueness in the docker compose services:
1. we hardcode some service name at: https://github.com/y-scope/clp/blob/aeb077ca4234ac4714941a273a6a95f56a20709d/tools/deployment/package/docker-compose-all.yaml#L298 . then if we ever scale the compression worker to multiple replicas, each replica reads the same hostname which results in same worker names across all replicas
2. we should avoid hardcoding the hostname then. that way, docker compose will assign numeric post-fixes to the hostnames (e.g. `compression-worker-1`), ensuring uniqueness of pidbox names when used by Celery, while the DNS can still resolve `compression-worker` to different IPs of the replicas. 
3. however, that why we hardcoded hostname for every docker-compose service was to ensure their hostnames match the "component names", defined in Python CLP utils, e.g. https://github.com/y-scope/clp/blob/01448089d8a370d03716c307952887f8d5b2695a/components/clp-py-utils/clp_py_utils/clp_config.py#L42 . those "component names" use `_` in names, which differs from the docker-compose service names where `-` are used as separators. We should also calibrate the component names (as tracked by #1777 ; #1785 has propsoed a fix) to match the service names.

since satisfying 2 + 3 will involve changing a lot of code, i think simply aligning the command in `docker-compose-all.yaml` (e.g. change "compression-worker" -> "compression-worker@h") is sufficient for the scope of this PR. we shall fix 2 + 3 seperately.

tools/deployment/package-helm/templates/query-worker-deployment.yaml
```diff
@@ -87,7 +87,7 @@ spec:
             "--concurrency", "{{ .Values.workerConcurrency }}",
             "--loglevel", "WARNING",
             "-Q", "query",
-            "-n", "query-worker"
+            "-n", "query-worker@%h"
```
junhaoliao, line 90: ditto - let's update https://github.com/y-scope/clp/blob/aeb077ca4/tools/deployment/package/docker-compose-all.yaml#L516 to match this command argument

tools/deployment/package-helm/templates/compression-worker-deployment.yaml
junhaoliao: that said, as pointed out by @gibber9809 , the above proposal will only ensure uniqueness across different compression-worker replicas on the **same host** where the docker-compose project runs. (which is still valuable as we can debug multi worker issues without a multi host cluster)

however, if the docker compose project is run across multiple hosts to form a cluster, hostnames can still collide. as proposed by #2274 , maybe we should formally remove mulithost support in the docker compose flow

junhaoliao (changes requested):
got some valuable insights from @gibber9809 offline - let's update the docker-compose config to match the changes in the helm chart.

for the PR title, how about:
```
fix(deployment): Append hostname to Celery worker node names to prevent pidbox conflicts (fixes #2259).
```

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -326,7 +326,7 @@ services:
       "--loglevel", "WARNING",
       "-f", "/var/log/compression_worker/worker.log",
       "-Q", "compression",
-      "-n", "compression-worker"
+      "-n", "compression-worker@%h"
```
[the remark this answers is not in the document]
junhaoliao: this is known limitation - discussed at https://github.com/y-scope/clp/pull/2269/changes#r3227145236 already

we will not attempt to address the limitation in this PR. @coderabbitai please help create an issue to track

junhaoliao (approved):

PR #2270, opened 2026-05-09
fix(deployment): Auto-initialize MongoDB replica set on fresh volumes

### LINKED ISSUE #2257: Results cache (MongoDB) fails to recover after PVC recreation due to uninitialized replica set
### Bug

When the clp-results-cache-0 pod's PersistentVolumeClaim is deleted and recreated (e.g., due to AZ mismatch after node replacement), MongoDB starts with an empty data directory but the mongod.conf still has replication.replSetName: "rs0" configured. This puts MongoDB in RSGhost state — it knows it should be part of a replica set but has no replica set configuration.

All clients (garbage collector, WebUI, query scheduler) fail with:
ServerSelectionTimeoutError: No servers match selector "Primary()"
server_type: RSGhost

Recovery requires manual intervention: rs.initiate() + rs.reconfig() to set the correct hostname.

Root Cause:

The Helm chart configures MongoDB with a replica set (replSetName: "rs0") but has no mechanism to automatically initialize the replica set on a fresh volume. The clp-results-cache-indices-creator job only creates MongoDB indices — it doesn't handle replica set initialization.

Additionally, rs.initiate() registers the member using the pod hostname (clp-results-cache-0:27017) rather than the service name (clp-results-cache:27017), causing DNS resolution failures from other pods.

Proposed Fix:

Option A (Recommended): Remove replica set requirement entirely.

The results cache is a single-node MongoDB instance used as an ephemeral cache for search results. It doesn't benefit from replica set features (no secondaries, no failover). Running in standalone mode eliminates the initialization problem completely.

Remove replication.replSetName from the mongod.conf template in the ConfigMap
Update the MongoDB connection URI in all clients to remove ?replicaSet=rs0 if present
No rs.initiate() needed on fresh volumes — MongoDB starts immediately as a standalone primary
Option B: Add an init container for auto-initialization.

If replica set mode is required (e.g., for change streams), add an init container to the clp-results-cache StatefulSet that:

Waits for MongoDB to be ready
Checks if the replica set is already initialized (rs.status())
If not, runs rs.initiate() with the service DNS name as the member host
initContainers:
  - name: init-replica-set
    image: mongo:7.0.1
    command:
      - bash
      - -c
      - |
        until mongosh --host localhost:27017 --eval "db.runCommand('ping')" --quiet; do
          sleep 2
        done
        mongosh --host localhost:27017 --eval '
          try {
            rs.status();
          } catch(e) {
            rs.initiate({
              _id: "rs0",
              members: [{_id: 0, host: "clp-results-cache:27017"}]
            });
          }
        '
Note: This requires the init container to run after the main MongoDB container starts, which isn't standard. An alternative is a post-start lifecycle hook or a sidecar.

Additional Issue: AZ Mismatch

The underlying trigger for PVC recreation is an AZ mismatch between the EBS PersistentVolume and the replacement node. The Helm chart should either:
use volumeBindingMode: WaitForFirstConsumer in the StorageClass (already default in EKS, but the PV from the old node persists in the wrong AZ)
Or add a note in the chart's values.yaml recommending EFS for multi-AZ deployments

### CLP version

0.10.0

### Environment

CLP version: 0.10.1-dev (Helm chart 0.2.1-dev.3)
MongoDB: 7.0.1
Deployment: EKS with EBS gp3 volumes
Trigger: ASG replaced bundled-service node in different AZ

### Reproduction steps

N/A

goynam:
## Summary
- Adds a `postStart` lifecycle hook to the results-cache (MongoDB) StatefulSet that automatically initializes the replica set (`rs0`) when the PVC is fresh/recreated.
- The hook waits for mongod to accept connections, then runs `rs.initiate()` only if `rs.status()` fails (i.e., the replica set is not yet configured). This is idempotent and safe on normal restarts.
- Bumps Helm chart version to `0.3.2-dev.3`.

## Problem
When the results-cache pod's PVC is deleted and recreated (e.g., during disaster recovery or testing), MongoDB starts in `RSGhost` state because `mongod.conf` has `replication.replSetName: "rs0"` but no replica set has been initiated on the fresh data directory. All clients then fail with `"No servers match selector Primary()"`.

Fixes #2257

## Test plan
- [ ] Deploy the Helm chart with a fresh PVC (delete the existing results-cache PVC first)
- [ ] Verify the MongoDB pod starts successfully and transitions to Primary state
- [ ] Verify clients (results-cache-indices-creator job, query workers, web UI) can connect without errors
- [ ] Verify that a normal pod restart (without PVC deletion) still works correctly (idempotent hook)
- [ ] Run `kubectl exec` into the results-cache pod and confirm `rs.status()` shows a healthy primary

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Bumped deployment version to 0.3.2-dev.3

* **Infrastructure Improvements**
  * Worker replica counts now follow the scheduling configuration for better scaling control
  * Results cache now performs automatic MongoDB replica set initialization on startup for improved cache stability and reliability

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/clp/pull/2270)

Fixes https://github.com/y-scope/clp/issues/2257

### DIFF
--- a/tools/deployment/package-helm/templates/compression-worker-deployment.yaml
+++ b/tools/deployment/package-helm/templates/compression-worker-deployment.yaml
@@ -87,7 +87,7 @@ spec:
             "--concurrency", "{{ .Values.workerConcurrency }}",
             "--loglevel", "WARNING",
             "-Q", "compression",
-            "-n", "compression-worker"
+            "-n", "compression-worker@$(HOSTNAME)"
           ]
       volumes:
         - name: {{ include "clp.volumeName" (dict
--- a/tools/deployment/package-helm/templates/compression-worker-deployment.yaml
+++ b/tools/deployment/package-helm/templates/compression-worker-deployment.yaml
@@ -6,7 +6,7 @@ metadata:
     {{- include "clp.labels" . | nindent 4 }}
     app.kubernetes.io/component: "compression-worker"
 spec:
-  replicas: {{ .Values.compressionWorker.replicas }}
+  replicas: {{ .Values.scheduling.compressionWorker.replicas }}
   selector:
     matchLabels:
       {{- include "clp.selectorLabels" . | nindent 6 }}
@@ -87,7 +87,7 @@ spec:
             "--concurrency", "{{ .Values.workerConcurrency }}",
             "--loglevel", "WARNING",
             "-Q", "compression",
-            "-n", "compression-worker@$(HOSTNAME)"
+            "-n", "compression-worker"
           ]
       volumes:
         - name: {{ include "clp.volumeName" (dict
--- a/tools/deployment/package-helm/templates/query-worker-deployment.yaml
+++ b/tools/deployment/package-helm/templates/query-worker-deployment.yaml
@@ -87,7 +87,7 @@ spec:
             "--concurrency", "{{ .Values.workerConcurrency }}",
             "--loglevel", "WARNING",
             "-Q", "query",
-            "-n", "query-worker"
+            "-n", "query-worker@$(HOSTNAME)"
           ]
       volumes:
         - name: {{ include "clp.volumeName" (dict
--- a/tools/deployment/package-helm/templates/query-worker-deployment.yaml
+++ b/tools/deployment/package-helm/templates/query-worker-deployment.yaml
@@ -7,7 +7,7 @@ metadata:
     {{- include "clp.labels" . | nindent 4 }}
     app.kubernetes.io/component: "query-worker"
 spec:
-  replicas: {{ .Values.queryWorker.replicas }}
+  replicas: {{ .Values.scheduling.queryWorker.replicas }}
   selector:
     matchLabels:
       {{- include "clp.selectorLabels" . | nindent 6 }}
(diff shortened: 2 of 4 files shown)


tools/deployment/package-helm/templates/compression-worker-deployment.yaml
junhaoliao, line 1: changes seem not related to the current PR

tools/deployment/package-helm/templates/query-worker-deployment.yaml
junhaoliao, line 1: changes seem not related to the current PR

tools/deployment/package-helm/templates/results-cache-statefulset.yaml
junhaoliao, line 1: this will likely cause conflicts with the results-cache-indices-creator Job. we have been working on some fix locally over the past week. @hoophalab may coordinate the submission of the changes (either in this PR or a separate PR)

hoophalab:
Changes are captured in #2276.

PR #2271, opened 2026-05-09
fix(job-orchestration)!: Add defensive timeouts to prevent scheduler hang (fixes #2252).

### LINKED ISSUE #2252: Query scheduler hangs indefinitely when polling stale Celery task results from Redis
### Bug

Description:

The query scheduler becomes unresponsive after processing a query job when query worker pods are recycled (due to rolling restarts, node terminations, or EBS-related crashes). The scheduler process remains alive but stops polling for new jobs, requiring a manual restart of both Redis and the scheduler to recover.

Symptoms:

Scheduler logs show query_scheduler started and processes one job successfully, then produces no further output
New query jobs remain in PENDING status (status=0) indefinitely
The scheduler process is alive (sleeping state) with open Redis connections that have unread data accumulating in rx_queue
The UI shows "Running - search job found 0 results" because the WebUI's polling times out
Root Cause Analysis:

The handle_jobs function in query_scheduler.py uses asyncio.wait with return_when=asyncio.FIRST_COMPLETED to monitor both the handle_job_updates task and reducer acquisition tasks. When a query job is dispatched to workers via Celery, the scheduler stores a celery.result.GroupResult and polls Redis for task completion.

When query worker pods die (due to node termination, EBS throughput exceeded, rolling restarts, etc.), their in-flight Celery tasks never write results to Redis. The scheduler's check_job_status_and_update_db function then blocks indefinitely waiting for these results, which prevents the main event loop from polling for new jobs.

Additionally, if handle_job_updates encounters an unhandled exception or completes unexpectedly, the main loop exits silently:

if handle_updating_task in done:
    logger.error("handle_job_updates completed unexpectedly.")
    return
This causes the scheduler to stop processing entirely without any visible error in logs.

### CLP version

0.12.0

### Environment

CLP version: 0.12.0
Deployment: Kubernetes (EKS) with Helm
Workers: Celery with RabbitMQ broker and Redis result backend
Trigger: Worker/scheduler pod recycling due to resource issues.

### Reproduction steps

Deploy CLP with query scheduler and multiple query workers
Submit a search query that dispatches tasks to workers
Kill/restart query worker pods while tasks are in-flight (e.g., kubectl rollout restart deployment/clp-query-worker)
The scheduler hangs after the current job completes and never picks up new jobs
Recovery Workaround:

kubectl exec clp-redis-0 -- redis-cli -a <password> FLUSHALL
kubectl rollout restart deployment/clp-query-worker
kubectl rollout restart deployment/clp-query-scheduler
Proposed Fix:

Add a timeout when polling Celery task results. If a task result isn't received within a configurable duration (e.g., 5 minutes), mark the task as failed and move on:
# In check_job_status_and_update_db or handle_finished_search_job
try:
    result = await asyncio.wait_for(
        async_task_result.get(), 
        timeout=task_timeout_seconds
    )
except asyncio.TimeoutError:
    logger.warning("Task %s timed out, marking as failed", task_id)
    # Mark task as failed and continue
Add error handling in handle_job_updates to prevent silent exits:
async def handle_job_updates(...):
    while True:
        try:
            await handle_cancelling_search_jobs(db_conn_pool)
            await check_job_status_and_update_db(db_conn_pool, results_cache_uri)
        except Exception:
            logger.exception("Error in handle_job_updates, retrying...")
            await asyncio.sleep(jobs_poll_delay)
On startup, detect and fail any jobs that are in RUNNING state but have no active workers (stale jobs from before a restart).

goynam:
# Description

1. Add defensive timeouts to `compress`, `search`, and `extract_stream` tasks. This terminates the job when it takes too long for some reason.
2. Add defensive timeouts in `task_results.get()`

# Checklist

  breaking change.

# Validation performed

1. Set the timeout to 1s and attempted to compress a file. The compression failed and threw an ERROR log.
2. The second timeout cannot actually be triggered; it was added strictly as a defensive measure per @goynam 's request.

**Original PR description below**

## Summary

Comprehensive fix for query and compression schedulers hanging/freezing when workers die, Redis becomes stale, or infrastructure has transient failures.

## Problem

The query scheduler becomes unresponsive after a worker pod is killed mid-task (rolling restart, node termination, OOM). Production recovery currently requires `redis-cli FLUSHALL` plus rolling-restart of workers and scheduler. Fixes #2252.

### Two distinct Redis-key concerns (clarification)

When `task_group.apply_async()` is called, Celery writes two different kinds of keys:

| Key | Written by | When |
|---|---|---|
| **Group manifest** (list of child task IDs) | Scheduler | At dispatch, immediately |
| **Per-child task result** | Worker | When the task finishes |

These keys drive the two distinct symptoms below — they are not the same bug:

**1. Why the scheduler hangs (the bug from the issue page).**
When a worker is killed mid-task, *only the per-child result key is missing.* `GroupResult.ready()` checks every child; one missing → returns `False` forever. The scheduler keeps polling for a result that will never arrive. Before this PR `async_task_result.get(...)` was also unbounded, so on the rare path where `ready()` returned True but the backend stalled, the asyncio loop could block directly on the get call.

**2. Why Redis grows unboundedly (a separate, pre-existing problem).**
The group manifests from row 1 — plus any partial child results that *were* written before the worker died — never get garbage-collected. With `result_persistent = True` and no `result_expires` configured, every aborted job leaks Redis keys. Across many incidents and scheduler restarts these accumulate.

The PR fixes both, with separate mechanisms:

- For the hang → per-dispatch elapsed-time check + `revoke(terminate=True)` + bounded `get(timeout=10)`, layered behind Celery's own `task_time_limit` as the primary worker-level defense.
- For the growth → `result_expires = 7200` in both Celery configs (Celery itself GCs stale manifests).

### Other failure modes addressed

3. **Dispatch blocks the event loop** — `concurrent.futures.as_completed` in `handle_pending_query_jobs` is synchronous; if dispatch hangs (Redis/DB/OTEL), the entire async loop freezes.
4. **Silent exit** — Unhandled exceptions in `handle_job_updates` kill the scheduler with no log.
5. **Compression multi-batch stall** — The pre-existing zero-completion timeout fires only before the first batch completes; a job that succeeds on batch 1 and stalls on batch 2 is invisible.

### Root-cause prevention

| Change | File | Effect |
|---|---|---|
| `task_soft_time_limit` / `task_time_limit` | Both `celeryconfig.py` | Worker-level time limits — primary defense for "task ran too long" (**#1**) |
| `result_expires = 7200` | Both `celeryconfig.py` | Celery auto-deletes results after 2h — addresses **#2** (unbounded Redis growth) |

### Query scheduler (`query_scheduler.py`)

| Change | Effect |
|---|---|
| Per-dispatch `last_dispatch_time` timer | Staleness check uses sub-job dispatch time, not overall job start (so healthy multi-batch queries aren't falsely failed on later batches) |
| Configurable `task_timeout_seconds` (default 300s) | Backstop when worker pod is killed before honoring `task_time_limit` — addresses **#1** |
| `revoke(terminate=True)` on timeout (try/except, `logger.exception`) | Prevents zombie workers from writing results to already-failed jobs |
| `async_task_result.get(timeout=10, interval=0.005)` + catch `celery.exceptions.TimeoutError` | Bounds the *other* path into **#1** — when `ready()` returns True but the backend stalls |
| Mark `query_tasks` rows (PENDING/RUNNING → FAILED) on timeout | Mirrors the cancel path — no orphan task rows |
| Dispatch timeout (60s) on `as_completed` | Unblocks event loop if dispatch hangs (**#3**); except clause uses `concurrent.futures.TimeoutError` for clarity |
| `GroupResult.restore()` try/except | Handles corrupt Redis after flush/partial expiry |
| `handle_job_updates` retry loop with `CancelledError` propagation | Survives transient errors; allows graceful shutdown (**#4**) |
| Clamp sleep to `max(0.0, ...)` | Prevents negative sleep raising under load |

### Compression scheduler (`compression_scheduler.py`)

| Change | Effect |
|---|---|
| Progress-based timeout via `last_progress_time` | Detects stalls in *any* batch, not just zero-completion jobs (**#5**) |
| Configurable `task_timeout_seconds` (default 600s) | Configurable via `CompressionScheduler` config |
| `result_handle.revoke()` on timeout | Cancels in-flight worker tasks (Celery: `revoke(terminate=True)`; Spider: no-op) |
| Mark RUNNING task rows as FAILED on timeout | No orphan task rows |
| Retry on transient errors instead of `return -1` | Scheduler stays alive through DB blips |
| Consecutive-error limit (10) | Exits on persistent failures (DB down, config corruption) |

### TaskManager interface (`scheduler/compress/task_manager/`)

| Change | Effect |
|---|---|
| Abstract `revoke()` method on `ResultHandle` | Lets the scheduler cancel in-flight work without reaching into per-implementation internals |
| Celery impl: `revoke(terminate=True)` | Cancels and SIGTERMs running children |
| Spider impl: no-op | Spider has no equivalent; relies on its per-task time limit |

### Configuration

| File | Change |
|---|---|
| `clp_config.py` | `task_timeout_seconds` field on both `CompressionScheduler` and `QueryScheduler` |
| `clp-config.template.json.yaml` | Documented with defaults (300s query, 600s compression) |
| `scheduler_data.py` | `last_progress_time` on `CompressionJob`, `last_dispatch_time` on `QueryJob` |

## Failure mode coverage

| Scenario | Before | After |
|---|---|---|
| Task runs longer than allowed | Polls forever | Celery `task_time_limit` SIGKILLs the worker (**#1**) |
| Worker pod killed before honoring `task_time_limit` (**#1**) | `ready()` returns False forever; polls forever | Per-dispatch timeout → revoke + FAIL after 300s |
| `get()` stalls when `ready()` is True (**#1**) | Async loop blocks | `get(timeout=10)` → retried next poll |
| Group manifests pile up in Redis (**#2**) | Unbounded growth | `result_expires=7200` auto-cleanup |
| Dispatch hangs (broker/DB/OTEL) (**#3**) | Event loop frozen | 60s dispatch timeout → FAIL |
| `handle_job_updates` exception (**#4**) | Scheduler exits silently | Retry + log |
| Compression multi-batch stall (**#5**) | Never detected (only zero-completion checked) | `last_progress_time` detects any stall |
| `GroupResult.restore()` corrupt | Unhandled exception | try/except → FAIL |
| Persistent infra failure | Spins or crashes forever | 10 consecutive errors → exit |

## Test plan

- [x] `ruff format` passes
- [x] All dispatch paths set `last_dispatch_time`
- [x] Compression `last_progress_time` updates on every batch completion
- [x] Dict iteration uses snapshot pattern (safe to delete during iteration)
- [x] `kill_hanging_jobs` covers stale RUNNING jobs on startup
- [ ] Manual test: kill query worker mid-task → job marked FAILED after timeout
- [ ] Manual test: kill compression worker → job times out after configured timeout
- [ ] Manual test: restart scheduler with stale Redis → starts clean, picks up new jobs

Fixes https://github.com/y-scope/clp/issues/2252


components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1160,6 +1262,51 @@ async def main(argv: list[str]) -> int:
...
+                        f"   SELECT id FROM {QUERY_JOBS_TABLE_NAME}"
+                        f"   WHERE status={QueryJobStatus.CANCELLED}"
+                        f" )"
+                    )
+                db_conn.commit()
+                if num_cancelled > 0:
+                    logger.info(f"Resolved {num_cancelled} stale CANCELLING jobs on startup.")
+    except Exception:
+        logger.exception("Failed to clean up CANCELLING jobs.")
+
```
hoophalab: Can we delete this clean-up? The scheduler will clean-up jobs in CANCELLING state in `handle_cancelling_search_jobs`.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -1050,6 +1093,50 @@ async def check_job_status_and_update_db(db_conn_pool, results_cache_uri):
...
+                # Check for stale sub-jobs that have been running too long without results
+                if job.last_dispatch_time is not None:
+                    elapsed = (datetime.datetime.now() - job.last_dispatch_time).total_seconds()
+                    if elapsed > task_timeout_seconds:
+                        logger.error(
+                            f"Job `{job_id}` sub-job timed out after {elapsed:.0f}s "
+                            f"without results. Workers may have died."
+                        )
+                        # Revoke in-flight tasks (best-effort)
+                        if job.current_sub_job_async_task_result is not None:
```
hoophalab: This timeout only addresses the following situation: Celery receives a successful result from the message broker, but Redis is empty. I don't see how waiting longer will resolve this issue.

I suggest:
1. Instead of catching TimeoutError and returning None in try_getting_task_result, just raise it.
2. Remove the error handling here—if an exception occurs, the exception handling above will already deal with it.

You would need to remove task_timeout_seconds completely

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -810,34 +814,71 @@ def handle_pending_query_jobs(
...
+                        job = active_jobs.pop(job_id)
+                        set_job_or_task_status(
+                            db_conn,
+                            QUERY_JOBS_TABLE_NAME,
+                            job_id,
+                            QueryJobStatus.FAILED,
+                            QueryJobStatus.RUNNING,
+                            duration=(datetime.datetime.now() - job.start_time).total_seconds(),
+                        )
+                    future.cancel()
```
hoophalab: I cannot see why celery can raise a TimeoutError in the dispatching procedure

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
```diff
@@ -531,6 +532,49 @@ def poll_running_jobs(
...
+
+    for job_id in jobs_to_timeout:
+        # Mark in-flight task rows as failed
+        db_context.cursor.execute(
+            f"UPDATE {COMPRESSION_TASKS_TABLE_NAME}"  # noqa: S608
+            f" SET status = '{CompressionTaskStatus.FAILED}'"
+            f" WHERE job_id = %s AND status = '{CompressionTaskStatus.RUNNING}'",
+            (job_id,),
+        )
+        db_context.connection.commit()
```
hoophalab: 1. I cannot see why updating the database from RUNNING to FAILED revokes task on compression worker.
2. If a timeout on compression workers is strictly needed, a better approach might be using celery's `time_limit` in decorator, or `timeout` when `apply_async`.

hoophalab (commented):
@goynam Thank you for your contribution. Here are some comments:

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
goynam: Good catch — agreed and removed in 014600ef. handle_cancelling_search_jobs runs every poll iteration and kill_hanging_jobs (called above) already flips RUNNING→ABORTED, so this block was redundant.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
goynam: This is concurrent.futures.TimeoutError, not Celery — it's raised by concurrent.futures.as_completed(futures, timeout=60) when a ProcessPoolExecutor future doesn't complete in 60s. It guards against DispatchExecutor.dispatch_job_and_update_db itself wedging (broker auth blocked, MySQL stalled inside the executor process, etc.) — nothing to do with Celery's task lifecycle. I've made that explicit by changing except TimeoutError: → except concurrent.futures.TimeoutError.

components/job-orchestration/job_orchestration/scheduler/compress/compression_scheduler.py
goynam: 1. DB flip didn't actually cancel the worker. Right — flipping rows to FAILED only updates state, the worker keeps running. I've added a revoke() method to TaskManager.ResultHandle (Celery: revoke(terminate=True); Spider: no-op since it relies on its task time limit), and  _timeout_stale_jobs now calls it before flipping rows. This matches the pattern on the query side.
2. task_time_limit is the better primary defense. Agreed — added task_soft_time_limit / task_time_limit to both executor/compress/celeryconfig.py and executor/query/celeryconfig.py (compression: 540/600s; query: 270/300s, mirroring the existing task_timeout_seconds defaults). I've kept the scheduler-side staleness sweep as a backstop with a comment noting that — its job now is the case Celery's own time limit can't help with: the worker pod being killed before it can honor the limit.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
goynam: Mode A — broker reports success but Redis result key is missing/late. GroupResult.ready() returns True, then get() raises or hangs. The bounded get(timeout=10) + except celery.exceptions.TimeoutError in try_getting_task_result is what handles this path, and you're right that it's the  only thing needed there. I can change it to raise instead of returning None if you'd prefer — the outer try/except in handle_job_updates will catch and retry. Happy to make that change.
Mode B — worker is killed before pushing anything (the original bug from #2252). ready() returns False forever because at least one child result key was never written. Polling never even reaches get(), so no TimeoutError is ever raised. The only signal we have that the job is dead is wall-clock elapsed since dispatch — that's what task_timeout_seconds + last_dispatch_time cover.

components/job-orchestration/job_orchestration/executor/compress/celeryconfig.py
```diff
@@ -18,6 +19,11 @@
...
 # Results backend settings
 result_persistent = True
+result_expires = 7200
 
 broker_url = os.getenv("BROKER_URL")
 result_backend = os.getenv("RESULT_BACKEND")
+
+_clp_config = load_clp_config_from_config_path_env_var()
+task_soft_time_limit = _clp_config.compression_worker.task_soft_time_limit
+task_time_limit = _clp_config.compression_worker.task_time_limit
```
junhaoliao, line 29: celery's internal treats 0 or None as "no limit" and therefore direct assignment without None / 0 checking is fine

components/package-template/src/etc/clp-config.template.json.yaml
```diff
@@ -104,9 +104,13 @@ telemetry:
 #
 #compression_worker:
 #  logging_level: "INFO"
+#  task_soft_time_limit: 0  # A value of 0 disables the limit
```
junhaoliao: i updated this to align with the max_concurrent_tasks_per_job handling and docs

components/job-orchestration/job_orchestration/executor/query/celeryconfig.py
```diff
@@ -17,6 +18,11 @@
 result_backend = os.getenv("RESULT_BACKEND")
 
 result_persistent = True
+result_expires = 7200
+
+_clp_config = load_clp_config_from_config_path_env_var()
```
junhaoliao, line 23: an attempt to avoid bloating the env vars since we pass the config file anyways

components/job-orchestration/job_orchestration/executor/compress/celery_compress.py
```diff
@@ -23,11 +24,17 @@ def compress(
...
+        return compression_entry_point(
+            job_id,
+            task_id,
+            clp_io_config_json,
+            paths_to_compress_json,
+            clp_metadata_db_connection_config,
+            logger,
+        )
+    except SoftTimeLimitExceeded:
+        logger.error(
```
junhaoliao:
```suggestion
        logger.exception(
```

components/job-orchestration/job_orchestration/executor/query/extract_stream_task.py
```diff
@@ -298,3 +297,31 @@ def extract_stream(
...
+            job_id,
+            task_id,
+            job_config,
+            archive_id,
+            clp_metadata_db_conn_params,
+            results_cache_uri,
+            dataset,
+        )
+    except SoftTimeLimitExceeded:
+        logger.error(
```
junhaoliao:
```suggestion
        logger.exception(
```

components/job-orchestration/job_orchestration/executor/query/fs_search_task.py
```diff
@@ -286,3 +285,29 @@ def search(
...
+            job_id,
+            task_id,
+            job_config_blob,
+            archive_id,
+            clp_metadata_db_conn_params,
+            results_cache_uri,
+            dataset,
+        )
+    except SoftTimeLimitExceeded:
+        logger.error(f"Search task job_id={job_id} task_id={task_id} exceeded soft time limit.")
```
junhaoliao:
```suggestion
        logger.exception(f"Search task job_id={job_id} task_id={task_id} exceeded soft time limit.")
```

components/job-orchestration/job_orchestration/scheduler/compress/task_manager/celery_task_manager.py
```diff
@@ -20,6 +23,9 @@ def get_result(self, timeout: float = 0.1) -> list[CompressionTaskResult] | None
                 return [CompressionTaskResult.model_validate(res) for res in results]
             except celery.exceptions.TimeoutError:
                 return None
+            except celery.exceptions.SoftTimeLimitExceeded:
+                logger.error("Compression task exceeded soft time limit.")
```
junhaoliao:
```suggestion
                logger.exception("Compression task exceeded soft time limit.")
```

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -837,7 +840,13 @@ def handle_pending_query_jobs(
 def try_getting_task_result(async_task_result):
     if not async_task_result.ready():
         return None
-    return async_task_result.get(interval=0.005)
+    try:
+        return async_task_result.get(
+            timeout=TASK_RESULT_GET_TIMEOUT, interval=TASK_RESULT_GET_INTERVAL
+        )
+    except celery.exceptions.TimeoutError:
+        logger.error("Timed out waiting for task result.")
```
junhaoliao:
```suggestion
        logger.exception("Timed out waiting for task result.")
```

components/package-template/src/etc/clp-config.template.text.yaml
```diff
@@ -94,9 +94,13 @@ log_ingestor: null
 #
 #compression_worker:
 #  logging_level: "INFO"
+#  task_soft_time_limit: 0  # A value of 0 disables the limit
+#  task_time_limit: 0  # A value of 0 disables the limit
```
junhaoliao: i feel we should configure some value like task_soft_time_limit=600 (seconds, = 10 minutes) and task_time_limit=1200 (seconds, = 20 minutes) instead of leaving them unconfigured. what do you think? @gibber9809 

```suggestion
#  task_soft_time_limit: 600  # A value of 0 disables the limit
#  task_time_limit: 1200  # A value of 0 disables the limit
```

components/package-template/src/etc/clp-config.template.text.yaml
```diff
@@ -94,9 +94,13 @@ log_ingestor: null
...
 #
 #query_worker:
 #  logging_level: "INFO"
+#  task_soft_time_limit: 0  # A value of 0 disables the limit
+#  task_time_limit: 0  # A value of 0 disables the limit
```
junhaoliao: see https://github.com/y-scope/clp/pull/2271/changes#r3345689411


```suggestion
#  task_soft_time_limit: 600  # A value of 0 disables the limit
#  task_time_limit: 1200  # A value of 0 disables the limit
```

components/package-template/src/etc/clp-config.template.text.yaml
junhaoliao: if so, we should update the json template, clp-config.yaml and values.yaml too
```patch
Index: tools/deployment/package-helm/values.yaml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/tools/deployment/package-helm/values.yaml b/tools/deployment/package-helm/values.yaml
--- a/tools/deployment/package-helm/values.yaml	(revision 8956bf89b2d2afc246eccba46c950afebf0f1411)
+++ b/tools/deployment/package-helm/values.yaml	(date 1780456416882)
@@ -181,13 +181,13 @@
 
   compression_worker:
     logging_level: "INFO"
-    task_soft_time_limit: 0  # A value of 0 disables the limit
-    task_time_limit: 0  # A value of 0 disables the limit
+    task_soft_time_limit: 600  # A value of 0 disables the limit
+    task_time_limit: 1200  # A value of 0 disables the limit
 
   query_worker:
     logging_level: "INFO"
-    task_soft_time_limit: 0  # A value of 0 disables the limit
-    task_time_limit: 0  # A value of 0 disables the limit
+    task_soft_time_limit: 600  # A value of 0 disables the limit
+    task_time_limit: 1200  # A value of 0 disables the limit
 
   reducer:
     logging_level: "INFO"
Index: components/clp-py-utils/clp_py_utils/clp_config.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/components/clp-py-utils/clp_py_utils/clp_config.py b/components/clp-py-utils/clp_py_utils/clp_config.py
--- a/components/clp-py-utils/clp_py_utils/clp_config.py	(revision 8956bf89b2d2afc246eccba46c950afebf0f1411)
+++ b/components/clp-py-utils/clp_py_utils/clp_config.py	(date 1780456439061)
@@ -447,8 +447,8 @@
 
 class WorkerConfigBase(BaseModel):
     logging_level: LoggingLevel = "INFO"
-    task_soft_time_limit: NonNegativeInt = 0
-    task_time_limit: NonNegativeInt = 0
+    task_soft_time_limit: NonNegativeInt = 600
+    task_time_limit: NonNegativeInt = 1200
 
 
 class CompressionWorker(WorkerConfigBase):
Index: components/package-template/src/etc/clp-config.template.text.yaml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/components/package-template/src/etc/clp-config.template.text.yaml b/components/package-template/src/etc/clp-config.template.text.yaml
--- a/components/package-template/src/etc/clp-config.template.text.yaml	(revision 8956bf89b2d2afc246eccba46c950afebf0f1411)
+++ b/components/package-template/src/etc/clp-config.template.text.yaml	(date 1780456418261)
@@ -94,13 +94,13 @@
 #
 #compression_worker:
 #  logging_level: "INFO"
-#  task_soft_time_limit: 0  # A value of 0 disables the limit
-#  task_time_limit: 0  # A value of 0 disables the limit
+#  task_soft_time_limit: 600  # A value of 0 disables the limit
+#  task_time_limit: 1200  # A value of 0 disables the limit
 #
 #query_worker:
 #  logging_level: "INFO"
-#  task_soft_time_limit: 0  # A value of 0 disables the limit
-#  task_time_limit: 0  # A value of 0 disables the limit
+#  task_soft_time_limit: 600  # A value of 0 disables the limit
+#  task_time_limit: 1200  # A value of 0 disables the limit
 #
 #webui:
 #  host: "localhost"
Index: components/package-template/src/etc/clp-config.template.json.yaml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/components/package-template/src/etc/clp-config.template.json.yaml b/components/package-template/src/etc/clp-config.template.json.yaml
--- a/components/package-template/src/etc/clp-config.template.json.yaml	(revision 8956bf89b2d2afc246eccba46c950afebf0f1411)
+++ b/components/package-template/src/etc/clp-config.template.json.yaml	(date 1780456422296)
@@ -104,13 +104,13 @@
 #
 #compression_worker:
 #  logging_level: "INFO"
-#  task_soft_time_limit: 0  # A value of 0 disables the limit
-#  task_time_limit: 0  # A value of 0 disables the limit
+#  task_soft_time_limit: 600  # A value of 0 disables the limit
+#  task_time_limit: 1200  # A value of 0 disables the limit
 #
 #query_worker:
 #  logging_level: "INFO"
-#  task_soft_time_limit: 0  # A value of 0 disables the limit
-#  task_time_limit: 0  # A value of 0 disables the limit
+#  task_soft_time_limit: 600  # A value of 0 disables the limit
+#  task_time_limit: 1200  # A value of 0 disables the limit
 #
 #webui:
 #  host: "localhost"
```

junhaoliao (dismissed):
@hoophalab Could you review the peer coding changes I submitted? once reviewed, this should be good to go.

components/package-template/src/etc/clp-config.template.text.yaml
junhaoliao: if so, we should mark this PR as breaking

hoophalab (dismissed):
I pushed the diff and bumped helm

junhaoliao (dismissed):

junhaoliao:
Let's wait for @gibber9809 's comments on https://github.com/y-scope/clp/pull/2271#discussion_r3345689411 before merging

components/package-template/src/etc/clp-config.template.text.yaml
gibber9809: Sorry for the slow reply -- I'm fine with having these 10minute/20minute defaults. Those should be conservative enough that they'll only ever kick in when something has gone wrong.

gibber9809 (dismissed):

junhaoliao (approved):

PR #2273, opened 2026-05-11
feat(clp-package): Parameterize clp-package Docker image by Ubuntu codename; auto-detect host version.

junhaoliao:
# Description

Previously, the `clp-package` Docker image was hardcoded to `ubuntu:jammy`. Binaries built on an Ubuntu 24.04 (Noble) host (glibc 2.39) cannot run inside a Jammy container (glibc 2.35), causing `task` to fail on Noble hosts at the package-image stage. This affects both the C++ binaries (which require GLIBC_2.38) and Python packages with native extensions (e.g., `_mysql_connector.so`, `psutil`, `greenlet`) that are compiled on the host and linked against its glibc.

This PR makes two changes to `tools/docker-images/clp-package/`:

1. **Dockerfile** — The base image is now parameterized via `ARG UBUNTU_VERSION_CODENAME=jammy` so `FROM ubuntu:${UBUNTU_VERSION_CODENAME}` resolves to the host-matching Ubuntu version. Additionally, `userdel -r ubuntu` is called before `useradd clp-user` to handle the default `ubuntu` user present in Noble cloud images (UID 1000), which would otherwise conflict with the `clp-user` UID.

2. **build.sh** — Auto-detects the host's Ubuntu codename from `/etc/os-release` (`VERSION_CODENAME`) and passes it as `--build-arg UBUNTU_VERSION_CODENAME=<codename>`. Defaults to `jammy` when detection fails, preserving backward compatibility.

# Checklist

  breaking change.

# Validation performed

### Task
Run `task` on an Ubuntu 24.04 (Noble) host to verify the build succeeds end-to-end.

### Command
```bash
task
```

### Output
```
task: [docker-images:package] ./build.sh
...
#4 [internal] load metadata for docker.io/library/ubuntu:noble
...
#30 exporting to image
#30 exporting layers done
...
task: [package] echo '0.12.1-dev' > '/home/ubuntu/workspace/clp/build/clp-package/VERSION'
```

### Explanation
The Docker image build resolved the base image to `ubuntu:noble`, and the full `task` pipeline completed successfully on a Noble host.

### Task
Verify that `build.sh` reads the host's codename from `/etc/os-release` and passes it as a build-arg.

### Command
```bash
bash -x tools/docker-images/clp-package/build.sh 2>&1 | grep -E "ubuntu_version|host_codename|UBUNTU_VERSION"
```

### Output
```
+ ubuntu_version_codename=jammy
+ host_codename=noble
+ ubuntu_version_codename=noble
+ build_cmd+=(--build-arg "UBUNTU_VERSION_CODENAME=${ubuntu_version_codename}")
+ docker build --pull ... --build-arg UBUNTU_VERSION_CODENAME=noble ...
```

### Explanation
The script defaults to `jammy` but correctly overrides with the host's `VERSION_CODENAME=noble`, ensuring the runtime image's glibc matches the host.

### Task
Verify the Dockerfile accepts `--build-arg UBUNTU_VERSION_CODENAME=noble` and pulls `ubuntu:noble` as the base image.

### Command
```bash
docker build --check --build-arg UBUNTU_VERSION_CODENAME=noble -f tools/docker-images/clp-package/Dockerfile . 2>&1
```

### Output
```
Check complete, no warnings found.
```

### Explanation
Docker resolved the base image to `docker.io/library/ubuntu:noble`, confirming the build-arg is correctly substituted into the `FROM` directive.

### Task
Verify backward compatibility — the Dockerfile still works with the default `jammy` codename.

### Command
```bash
docker build --check --build-arg UBUNTU_VERSION_CODENAME=jammy -f tools/docker-images/clp-package/Dockerfile . 2>&1
```

### Output
```
Check complete, no warnings found.
```

### Explanation
The Dockerfile passes syntax validation with the default `jammy` value, confirming no breaking change for existing builds.

### Task
Verify the built container is based on Ubuntu Noble and the `ubuntu` user is removed while `clp-user` exists with UID 1000.

### Command
```bash
docker run --rm <image> cat /etc/os-release | head -5
docker run --rm <image> id clp-user
docker run --rm <image> id ubuntu
```

### Output
```
PRETTY_NAME="Ubuntu 24.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.4 LTS (Noble Numbat)"
VERSION_CODENAME=noble

uid=1000(clp-user) gid=1000(clp-user) groups=1000(clp-user)

id: 'ubuntu': no such user
```

### Explanation
The container runs Ubuntu Noble; the default `ubuntu` user (present in Noble cloud images, UID 1000) was removed, and `clp-user` was created with UID 1000, avoiding the UID conflict that would otherwise occur.

### Task
Verify `start-clp.sh` starts all CLP services successfully.

### Command
```bash
cd build/clp-package && ./sbin/start-clp.sh
```

### Output
```
2026-05-11T03:36:54.358 INFO [controller] Setting up environment for compression_scheduler...
...
Container clp-package-ec9d-webui-1 Healthy
...
2026-05-11T03:37:06.327 INFO [controller] Started CLP.
```

### Explanation
All services (database, queue, redis, results-cache, compression worker/scheduler, query worker/scheduler, api-server, webui, reducer, garbage-collector) started and became healthy.

### Task
Verify log compression works end-to-end in the Noble-based package.

### Command
```bash
cd build/clp-package && ./sbin/compress.sh --timestamp-key timestamp ~/samples/postgresql.jsonl
```

### Output
```
2026-05-11T03:55:03.410 INFO [compress] Compression job 5 submitted.
2026-05-11T03:55:05.415 INFO [compress] Compressed 389.02MB into 10.05MB (38.70x). Speed: 203.97MB/s.
2026-05-11T03:55:05.916 INFO [compress] Compression finished.
2026-05-11T03:55:05.916 INFO [compress] Compressed 389.02MB into 10.05MB (38.70x). Speed: 184.50MB/s.
```

### Explanation
Compression completed successfully with a 38.70x compression ratio, confirming that the Noble-based runtime image correctly executes the CLP binaries built on the Noble host.

### Task
Verify log search works end-to-end in the Noble-based package.

### Command
```bash
cd build/clp-package && ./sbin/search.sh --count "error"
```

### Output
```
(Container started, search completed, exit code 0)
```

### Explanation
Search completed successfully, confirming the query pipeline works correctly with the Noble-based package.

## Summary by CodeRabbit

* **Chores**
  * Docker image build now supports selecting the Ubuntu base image at build time for greater flexibility.
  * Build script auto-detects the host Ubuntu codename and forwards it to the image build to improve compatibility.
  * CI workflow and build action updated to accept and pass the base image codename.
  * Image user initialization improved to detect and handle existing user ID conflicts, failing safely if unresolved.

* **Documentation**
  * Build docs updated to clarify supported architectures and how host Ubuntu codename is used for runtime images.


tools/docker-images/clp-package/Dockerfile
```diff
@@ -1,6 +1,7 @@
 # syntax=docker/dockerfile:1
 
-FROM ubuntu:jammy AS base
+ARG UBUNTU_VERSION_CODENAME=jammy
```
Bill-hbrhbr, line 3: We should also add an argument inside [`./.github/actions/clp-build-runtime-image`](https://github.com/junhaoliao/clp/blob/351588bec9efc696760e1f0a97e3b4969315611e/.github/actions/clp-build-runtime-image/action.yaml) to specify the codename, then pass `jammy` explicitly for [clp-artifact-build](https://github.com/junhaoliao/clp/blob/351588bec9efc696760e1f0a97e3b4969315611e/.github/workflows/clp-artifact-build.yaml#L736) workflow

tools/docker-images/clp-package/Dockerfile
```diff
@@ -23,7 +24,8 @@ ENV LD_LIBRARY_PATH="${CLP_HOME}/lib" \
     PYTHONPATH="${CLP_HOME}/lib/python3/site-packages" \
     USER="clp-user"
 
-RUN useradd --uid ${UID} --shell /bin/bash --home-dir ${CLP_HOME} ${USER}
+RUN userdel -r ubuntu 2>/dev/null || true \
```
Bill-hbrhbr: Instead of hardcoding `ubuntu`, we should resolve the username from the `UID`, using something like `getent passwd "${UID}" | cut -d: -f1`.

Also, is it possible to raise an error when encountering a UID conflict?

tools/docker-images/clp-package/build.sh
```diff
@@ -36,6 +36,15 @@ build_cmd=(
     --file "${script_dir}/Dockerfile"
 )
 
+ubuntu_version_codename="jammy"
+if [[ -f /etc/os-release ]]; then
+    host_codename="$(. /etc/os-release && echo "$VERSION_CODENAME")"
```
Bill-hbrhbr, line 41: Right now, `task package` builds a `clp-package` image using the same Ubuntu release as the host invoking the task command. However, I believe a 22.04 host should still be able to build a 24.04 `clp-package` image.

Should we expose the Ubuntu version codename as a Task argument or environment variable so the package target is not tied to the host release?

Bill-hbrhbr (changes requested):

tools/docker-images/clp-package/build.sh
junhaoliao: dynamic ubuntu version detection on host should make more sense. the main reason why we want to avoid hardcoding a version of the ubuntu image is because whatever (Python wheels, glibc dependent binaries, etc.) a non-Jammy host builds do not run inside the Jammy-based image container.

the long-term plan is to migrate the base image from Ubuntu to Muslinux / Manylinux as @jackluo923 proposed. we (or whoever has bandwidth) can work on it in the next month. also, i think we shall explore setting up https://containers.dev/ too.

for now, this PR can unblock our developers to develop on different versions of Ubuntu hosts, without use of Ubuntu Jammy build containers

let me know what you think

tools/docker-images/clp-package/Dockerfile
junhaoliao: sg. i tried to address it in 5af1885 . let me know if i understand correctly

.github/actions/clp-build-runtime-image/action.yaml
```diff
@@ -14,17 +14,14 @@ inputs:
...
-    description: "Platform ID of the container (e.g. ubuntu)"
-    required: false
-  platform_version_id:
-    description: "Platform VERSION_ID / VERSION_CODENAME of the container
-      (e.g. jammy, focal, etc.)"
-    required: false
   arch:
     description: "Target architecture (amd64 or arm64)"
     default: "amd64"
     required: false
```
Bill-hbrhbr, line 20:
```suggestion
required: true
```
Maybe open an issue to make this required?

tools/docker-images/clp-package/build.sh
Bill-hbrhbr: just realized that `tools/docker-images/clp-package/Dockerfile` copies built artifacts instead of building the package itself. Yes I agree we should keep host and host-built container image consistent.

tools/docker-images/clp-package/Dockerfile
```diff
@@ -23,6 +24,9 @@ ENV LD_LIBRARY_PATH="${CLP_HOME}/lib" \
     PYTHONPATH="${CLP_HOME}/lib/python3/site-packages" \
     USER="clp-user"
 
+RUN existing_user="$(getent passwd "${UID}" | cut --delimiter=: --fields=1)" || true; \
+    [ -z "${existing_user}" ] || userdel --remove "${existing_user}" \
+    || { echo "ERROR: Cannot delete user '${existing_user}' that occupies UID ${UID}"; exit 1; }
 RUN useradd --uid ${UID} --shell /bin/bash --home-dir ${CLP_HOME} ${USER}
```
Bill-hbrhbr: To make the user handling more coherent, we can maybe merge these two inside a single `RUN`. Your choice.

Bill-hbrhbr (changes requested):
Need to update docs: https://github.com/y-scope/clp/blob/137af2ccb159ea0250ddf65d92f724b162bfadc1/docs/src/dev-docs/building-package.md?plain=1#L8

tools/docker-images/clp-package/Dockerfile
junhaoliao: makes sense

.github/actions/clp-build-runtime-image/action.yaml
junhaoliao: @coderabbitai open an issue to track making the `arch` input required for `.github/actions/clp-build-runtime-image`. The issue should cover updating all current and future callers to pass `arch` explicitly, removing the default value, and verifying behavior for both `amd64` and `arm64`. This is a follow-up cleanup and is out of scope for this PR.

Bill-hbrhbr (approved):

PR #2275, opened 2026-05-12
feat(clp-s)!: Adds CLP+ as an experimental feature.

davidlion:
# Description

Adds CLP+ (`clpp`), an experimental extension of `clp-s` that parses unstructured text logs into structured, query-able fields using [log-surgeon](https://github.com/y-scope/log-surgeon). Enabled via `--experimental`; existing `clp-s` behaviour is unchanged when the flag is absent.

## CLP+ storage & compression
- CLP+ uses a (log-surgeon parsing spec](https://github.com/y-scope/log-surgeon/blob/log-mechanic/rust/docs/parsing-spec-file.md) to parse the unstructured text during ingestion.
    - Note that for pure unstructured logs, CLP-S requires the log-converter to first change them to JSON.
- The parsing spec contains rules (named regex patterns) that may contain sub-rules (named regex captures). These patterns split each log message into:
    - Leaf rule matches: a rule match containing no sub-rule. These are stored in CLP-S columns.
    - A log shape: a string containing static text of a log message with the leaf rule matches replaced with `%log-surgeon.qualified.leaf_rule_name%` placeholders. (Similar to CLP log types.)
    - Parent rule matches: a rule match containing at least one sub-rule. These are stored as log shape metadata, where the parent rule matches for each log shape are "views" into the log shape.
- The parsing spec is stored in the archive, so extraction and search need only `--experimental`. Reading a CLP+ archive without it is an error.
- `int`/`float` leaves are encoded as native numeric columns, enabling range queries.
- Some additional statistics are stored to support special "stat" queries (see below).

## Search
- Extends the KQL grammar with function-call syntax for new filter and and projection functions.
- Adds the filter function `shape(txt)`  to match against the shape of `txt` where it is a log message or parent rule field (e.g. `shape(msg): "INFO * static log text * %qualified.leaf_rule_name% *"`).
- Leaf rules are queryable by their rule path (e.g. `message.blockID.genStamp`), including wildcards and numeric comparisons.
- Whole-message and parent-rule value filters depend on query decomposition, which is still under development and not yet reliable. It is gated behind `-DCLP_BUILD_CLPP_DECOMPOSITION=ON` (default `OFF`).
- Adds `stats.archives`, `stats.log_shapes`, and `stats.schema_tree` queries for inspecting archive metadata.

## Projection
- Reworks `Projection` to track a per-node mode mask (`Value`/`Shape`/`Decompose`) rather than a flat column set, so a single query can request raw text, the shape, and decomposed leaves at different nesting levels.
- Added two functions that can be run on log message or parent rule fields.
    - `shape(txt)`: output `txt`'s shape as `{"txt": {"shape": "..."}}`
    - `decompose(txt)`: output `txt`'s shape and all of `x`'s leaf rule fields as `{"txt": {"shape": "...", "leaf": [...]}}`.

## Documentation
- Adds user docs for CLP+ covering concepts, compression/extraction, search, and projection output.

## Build & tests
- Switches log-surgeon to its Rust implementation and adds a `fast_float` dependency.
- Adds `integration-tests/tests/binary_tests/test_clpp.py` covering round-trip, search, projection, and value encoding; decomposition-dependent tests are skipped unless the build flag is set.
- Extends the `clp-s` unit tests to run against both plain and CLP+ archives.

# Checklist

  breaking change.

# Validation performed

- Verified compression and extraction are lossless by round-tripping the (open-source Hive, Hadoop, OpenStack, and MongoDB datasets](https://docs.yscope.com/clp/main/user-docs/resources-datasets) and comparing the extracted output against clp-s and the original input (made using the log-converter for fully unstructured datasets).
- Verified search over a compressed archive returns the expected result counts across leaf, `shape()`, and non-wildcard message and parent rule queries.
    - Wildcard message and parent rule search has know issues and is disabled by default.
- Verified every projection mode produces the expected JSON structure by comparing against known-good output.
- Unit tests and the integration test suite pass.


components/core/src/clp_s/search/ast/StringLiteral.hpp
```diff
@@ -57,21 +57,23 @@ class StringLiteral : public Literal {
...
-            m_string_type = LiteralType::ClpStringT;
-        } else {
-            m_string_type = LiteralType::VarStringT;
-        }
-
-        if (has_unescaped_wildcards(m_v)) {
-            m_string_type |= LiteralType::ClpStringT;
-        }
+        // if (m_v.find(' ') != std::string::npos) {
+        m_string_type = LiteralType::ClpStringT | LiteralType::VarStringT;
```
gibber9809, line 69: Making a note that we'll need to resolve this properly before merging.

I think one thing we could do is just handle column resolution in a special way for clp+ that understands that the VarString/ClpString types are interchangeable for its purposes, and also just add a new `as_string(...)` method to `Literal.hpp` that gives you a string if the underlying is either VarString or ClpString.

The main complicating factors are:
1. Will have to watch out for any filtering code that uses as_var_string/as_clp_string
2. The `NarrowTypes` pass handles some special cases for not equals queries on predicates that can be interpreted as just ClpString or just VarString. E.g.:
`NOT msg<ClpString,VarString>: "varstring"` -> `msg<ClpString> exists OR (NOT msg<VarString>: "varstring")`
`NOT msg<ClpString,VarString>: "clp string"` -> `(NOT msg<ClpString>: "clp string") OR msg<VarString> exists`.

Need to think harder about whether that second point matters though.

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -303,13 +382,32 @@ void JsonParser::parse_obj_in_array(simdjson::ondemand::object line, int32_t par
...
+                                node_id_stack.top(),
+                                NodeType::ClpString,
+                                cur_key
+                        );
+                        m_current_parsed_message.add_unordered_value(value);
+                    }
                 } else {
                     node_id = m_archive_writer
                                       ->add_node(node_id_stack.top(), NodeType::VarString, cur_key);
+                    m_current_parsed_message.add_unordered_value(value);
```
gibber9809, line 435: Applies to all of these blocks that check if a string contains a space before deciding how to encode it, but did we ever do the experiment where we encode all strings in clp+ using the log-mechanic path?

Only reason I'm bringing this up is that if we decide to do this later on (e.g., because handling everything with log-mechanic might make it simpler to do things like PII removal across all string fields), then we'll probably need to add a lot of special cases in the code for archive versions before and after that change.

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -1363,5 +1497,182 @@ void JsonParser::split_archive() {
...
+        std::string_view lexeme,
+        SchemaNode::id_t parent_node_id,
+        std::string_view rule_name
+) -> std::optional<SchemaNode::id_t> {
+    double double_value{};
+    if (false == clp::string_utils::convert_string_to_double(lexeme, double_value)) {
+        return std::nullopt;
+    }
+
+    if (m_retain_float_format) {
```
gibber9809, line 1694: I think it might make sense to force users to retain float formats for floats inside of strings, since we're supposed to compress string fields losslessly.

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -1363,5 +1497,182 @@ void JsonParser::split_archive() {
...
+                        break;
+                    }
+                }
+            }
+        }
+        if (0 == node_id) {
+            node_id = m_archive_writer->add_node(parent_node_id, NodeType::VarString, rule_name);
+            m_current_parsed_message.add_unordered_value(lexeme);
+        }
+        m_current_schema.insert_unordered(node_id);
```
gibber9809: I think I kind of follow how the schema + MPT are being handled here, but I don't 100% get it.

MPT side I'm pretty sure I understand -- we create nodes in the MPT for parent rules with a special node type, which is more or less what I would expect.

Schema side it looks like we're adding nodes to the schema for each parent rule? If I'm interpreting the code correctly, it looks like for a rule `a.b.c` the schema tree becomes `a -> b -> c` and the schema becomes `node id<a>, node id<b>, node  id<c>`? And if `c` has a sibling `d` (as in they're both subrules of the subrule `a.b`, and appear as part of the same parent match as opposed to two different instances of `a.b` being matched) the schema becomes `node id<a>, node id<b>, node  id<c>, node id<a>, node id<b>, node  id<d>`? Not sure if I'm completely following the code though.

The main thing I don't understand is how we figure out that `c` and `d` are siblings both for decompression, and for supporting search on parent rules to make sure we turn it into leaf queries on `c` and `d` that are part of the same parent match. If we don't currently have a good way of figuring this out, I think we'll need one once we implement a version of parent search that ensures that the resulting leaf queries target leafs from the same parent match.

I think what I would have expected is for us to use `start_unordered_object/end_unordered_object` for each parent rule to create schemas like:
`<ParentRule, 3>, <ParentRule, 2>, node_id<c>, node_id<d>`
or
`<ParentRule, 5>, node_id<a>, <ParentRule, 3>, node_id<b>, node_id<c>, node_id<d>`
if including the parent node IDs in the schema helps make parent search simpler.

Either way though, it lets the schema record whether some arbitrary group of leaf nodes is part of the same parent match.

gibber9809 (commented):
Read through the `JsonParser` ingestion path and left some comments. Going to make sure I at least also take a look at `ArchiveWriter` changes before I leave. Should be able to catch any format changes I'm worried about that way, which is the only kind of change I'm actually worried about for release-blocking.

components/core/src/clp_s/search/ast/StringLiteral.hpp
gibber9809: Thought about it a bit more -- I think it might be worth adding an option to `NarrowTypes` to disable the transformation into separate `ClpString` and `VarString` predicates, and use that option in the clp+ path. Since clp+ currently uses `NodeType::VarString` for captured variables we can produce `VarString` column entries that contain a space, which means that the transformation isn't even necessarily correct unless we also change column resolution to be aware of that, and even if we do make column resolution account for that we lose the perf benefit we get in clp-s from this transformation, so there's really no point in doing it for clp+.

components/core/src/clp_s/ArchiveWriter.hpp
```diff
@@ -172,7 +180,16 @@ class ArchiveWriter {
...
+    auto add_node(SchemaNode::id_t parent_node_id, NodeType type, std::string_view key)
+            -> SchemaNode::id_t;
+
+    /**
+     * @return The root node ID of the LogType sub-tree.
+     * @return -1 if the sub-tree does not exist.
+     */
+    auto get_logtype_node() -> SchemaNode::id_t {
+        return m_schema_tree.get_subtree_node_id(constants::cDefaultNamespace, NodeType::LogType);
+    }
```
gibber9809: Nit: not used anymore? IIRC this is an artifact of an older version of the design.

components/core/src/clp_s/ArchiveWriter.cpp
```diff
@@ -85,6 +111,34 @@ auto ArchiveWriter::close(bool is_split) -> ArchiveStats {
...
+        if (compressed_size.has_error()) {
+            throw OperationFailed(ErrorCodeFailure, __FILENAME__, __LINE__);
+        }
+        files.emplace_back(
+                std::string(constants::cArchiveLogShapeStatsFile),
+                compressed_size.value()
+        );
+    }
+
+    if (false == m_parsing_spec_str.empty()) {
```
gibber9809: Potential nit, is an empty string technically a legal schema? If so, might make sense to use optional for the parsing spec as well.

components/core/src/clp_s/ArchiveWriter.cpp
```diff
@@ -297,7 +352,13 @@ bool ArchiveWriter::matches_timestamp(int parent_node_id, std::string_view key)
...
 
 size_t ArchiveWriter::get_data_size() {
-    return m_log_dict->get_data_size() + m_var_dict->get_data_size() + m_array_dict->get_data_size()
+    size_t log_dict_size{};
+    if (nullptr != m_log_shape_dict) {
+        log_dict_size = m_log_shape_dict->get_data_size();
+    } else {
+        log_dict_size = m_log_dict->get_data_size();
+    }
+    return log_dict_size + m_var_dict->get_data_size() + m_array_dict->get_data_size()
```
gibber9809, line 364: Is that log shape stats array a meaningful size? May want to include that here as well if so.

components/core/src/clp_s/ArchiveWriter.cpp
```diff
@@ -85,6 +111,34 @@ auto ArchiveWriter::close(bool is_split) -> ArchiveStats {
             {constants::cArchiveArrayDictFile, array_dict_compressed_size},
             {constants::cArchiveTablesFile, table_compressed_size}
     };
+
+    if (m_parent_rule_shapes) {
+        auto compressed_size{close_parent_rule_shapes()};
+        if (compressed_size.has_error()) {
+            throw OperationFailed(ErrorCodeFailure, __FILENAME__, __LINE__);
+        }
+        files.emplace_back(
```
gibber9809, line 118: This one is pretty important -- the order that different components are packed into the single-file archive is determined by the order of entries of this files array. That means that by doing `emplace_back` for this and the other new sections we're putting them at the end of the archive after the ERTs. For local search that's fine, but for search from S3 that would e.g. force us to buffer the entire archive in memory so that we can read the parsing spec before we can go back and actually do search.

Generally, we want this array to be sorted in the exact order that we would access these sections during search, so that we can do early termination when streaming the archive from S3. I think we'd probably want to move the parsing spec either before or around the schema tree/schema map, and move the parent rule shapes/log shape stats before or around the dictionaries? You have a better understanding of the requirements for clpp search, so I'll leave the exact order up to you.

components/core/src/clp_s/ArchiveReaderAdaptor.cpp
```diff
@@ -366,6 +367,19 @@ void ArchiveReaderAdaptor::checkin_reader_for_section(std::string_view section)
...
+    return m_archive_file_info.files.end()
+           != std::find_if(
+                   m_archive_file_info.files.begin(),
+                   m_archive_file_info.files.end(),
+                   [&](ArchiveFileInfo const& info) { return info.n == section; }
+           );
+}
+
+auto ArchiveReaderAdaptor::is_experimental_archive() const -> bool {
+    return has_section(constants::cArchiveParentRuleShapesFile);
```
gibber9809: This works fine for distinguishing clp-s and clpp archives, but do you think it would also make sense to use some of the unused header bytes to indicate as much, or record it in one of the metadata packets?

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -1363,5 +1497,182 @@ void JsonParser::split_archive() {
...
+        log_shape.append(log_msg.substr(log_msg_pos, match->range.start - log_msg_pos));
+        log_shape.append(
+                fmt::format("%{}%", match->ffi_pointers.fully_qualified_name.as_cpp_view())
+        );
+        log_msg_pos = match->range.end;
+    }
+    log_shape.append(log_msg.substr(log_msg_pos));
+
+    auto [log_shape_id, new_log_shape]{
+            YSTDLIB_ERROR_HANDLING_TRYX(m_archive_writer->update_log_shape_dict(log_shape))
```
gibber9809, line 1664: I could be missing something, but do we have to do escaping for these log shapes? I would have assumed so, but I don't see it happening in this code. If we do, it might be nice to have a class or something to represent an escaped log shape, since it's a bit hard to audit whether escaping rules are being respected when we're building up the log shape as an `std::string` directly.

gibber9809 (commented):
Did another round of review mostly looking through the ArchiveWriter. I think I've covered everything I'm worried about on the format side now.

components/core/src/clp_s/JsonParser.cpp
davidlion: You're right I just hadn't gotten to it till recently and I hadn't pushed yet. This should be addressed now.

components/core/src/clp_s/ArchiveWriter.hpp
davidlion: You're right. It should be removed now.
There might be a few of these sitting around. I still need to do a thorough pass of all changes in the branch.

components/core/src/clp_s/JsonParser.cpp
davidlion: You're analysis was right that the code couldn't handle it.

The recent update should resolve this by switching parent rules to use `start/end_unordered_object`.

It was a good point that although the syntax can't expose this limitation yet, we will benefit by having the format ready so that "old" archives are usable when we support it.

components/core/src/clp_s/ArchiveWriter.cpp
davidlion: It shouldn't be a valid parsing spec (technically it can run without crashing, but everything would be static text so it should be disallowed). I refactored the clpp/experimental fields and flags, so I think everything should be nicer now.

There is a more deliberate check/error for an empty parsing spec.
All classes should have a single optional field that is a struct containing all the clpp/experimental fields. I remember trying this before and reverting, but (imo) everything feels nicer after the change so whatever reason I had to revert was probably wrong.

components/core/src/clp_s/ArchiveWriter.cpp
davidlion: I hadn't realized this at all, it makes sense to me to re-order for sure.

In the end I put the spec just before the ERTs. The reasoning:

* We only need the spec for queries that need to generate interpretations (i.e. wildcard search on a parent shape or full log message), but ultimately a user's workload determines how common that actually is.
* If we need the spec, we need the parent shapes (since we'll need to check the interpretations against them).

With that said, the only real decision should be whether the log shape stats should be before or after the spec. Right now the largest parsing spec is ~5MiB, whereas the stats are roughly a few KiB (although they can theoretically be larger depending on the number of shapes in a single archive). Therefore, I put the log shape stats first to avoid buffering the spec in the case the user is just running the shape-count query.

That's my reasoning, but I'm open to changes. I somewhat feel as long as they come before the ERTs the order of the clp+ files shouldn't make a real difference, but I could be very wrong (don't have much exp with S3).

components/core/src/clp_s/JsonParser.hpp
```diff
@@ -213,8 +224,52 @@ class JsonParser {
...
+     * @param parent_node_id
+     * @param field_key
+     * @param field_value
+     * @return A result containing an error code indicating the failure:
+     * - ClppErrorCodeEnum::Failure if parsing fails.
+     * - Forwards `store_capture_groups`'s return values on failure.
+     * - Forwards `m_archive_writer->update_log_shape_dict`'s return values on failure.
+     * - Forwards `m_archive_writer->update_parent_rule_shapes`'s return values on failure.
+     */
+    auto parse_str_field(
```
gibber9809: Minor nit - I think the name of this function might be a little bit too generic, maybe something like `parse_log_surgeon_string_field` or `parse_log_message` might be a bit better? (Or at least `parse_string_field` if you disagree with those).

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -117,6 +134,39 @@ class IrUnitHandler {
...
+/**
+ * An open `ParentRule` unordered object scope during `parse_str_field`. `match` identifies the
+ * scope within a `parse_str_field` call. `schema_start` is the scopes starting position in the
+ * schema and used when closing the unordered object.
+ */
+struct ParentScope {
+    log_surgeon::Match const* match;
+    size_t schema_start;
+    SchemaNode::id_t tree_node_id;
+};
```
gibber9809, line 147: I think that the `schema_start` name is mechanically accurate, but maybe not the clearest terminology since it mixes mechanism with functionality.

Instead of `schema_start`, maybe something like `scope_handle`, `schema_handle`, or `rule_handle` would make the functionality more clear? `schema_start` is aligned more with how we've been naming this in the rest of the code, but I don't think our existing naming is that good for this feature.

components/core/src/clp/string_utils/string_utils.cpp
```diff
@@ -187,6 +190,13 @@ auto clean_up_wildcard_search_string(string_view str) -> string {
     return cleaned_str;
 }
 
+// std::stod is locale-dependent (respects the C locale's decimal point). In practice, we don't
+// modify this so the decimal point should be '.'.
+auto convert_string_to_double(std::string_view raw, double& converted) -> bool {
+    auto const res{fast_float::from_chars(raw.begin(), raw.end(), converted)};
+    return res.ptr == raw.end() && std::errc{} == res.ec;
+}
+
```
gibber9809, line 199: Beyond the scope of this PR, but I'm making a note to check if this fast_float library can help speed up the `restore_encoded_float` utility from clp-s.

components/core/src/clp_s/JsonParser.cpp
```diff
@@ -1402,14 +1543,158 @@ void JsonParser::parse_kv_log_event(KeyValuePairLogEvent const& kv) {
...
+        auto const rule_name{match.get_rule_name()};
+        auto const lexeme{match.get_lexeme()};
+        auto const parent_node_id{update_open_parent_scopes(
+                match,
+                open_scopes,
+                m_current_schema,
+                log_msg_node_id,
+                *m_archive_writer
+        )};
+        SchemaNode::id_t node_id{0};
```
gibber9809:
```suggestion
        SchemaNode::id_t node_id{-1};
```
Should probably use -1 or `std::optional<SchemaNode::id_t>`, just because 0 is a valid node ID, even if by construction the leaf node here should never be node 0.

components/core/src/clp_s/DictionaryWriter.hpp
```diff
@@ -54,6 +54,8 @@ class DictionaryWriter {
      */
     size_t get_data_size() const { return m_data_size; }
 
+    [[nodiscard]] auto get_next_id() const -> uint64_t { return m_next_id; }
+
```
gibber9809:
```suggestion
```
Not used, as far as I can tell?

components/core/CMakeLists.txt
```diff
@@ -148,6 +148,14 @@ if(BUILD_TESTING AND CLP_BUILD_TESTING)
...
 endif()
 
+# Always enable profiling if building tests.
+set(CLP_ENABLE_PROFILING_EFFECTIVE "${CLP_ENABLE_PROFILING}")
+if(CLP_ENABLE_TESTS AND CLP_ENABLE_PROFILING LESS "1")
+    set(CLP_ENABLE_PROFILING_EFFECTIVE "1")
+endif()
+
+add_definitions(-DCLP_ENABLE_PROFILING=${CLP_ENABLE_PROFILING_EFFECTIVE})
+
```
gibber9809: Think this might have gotten messed up in a merge to this branch at some point, since we decided to not automatically enable profiling based on the test target being enabled when we originally merged the new profiling code (per [discussion](https://github.com/y-scope/clp/pull/2473#discussion_r3759323860)).

Any problems with just taking this out, or is there a new reason we need to do this?

components/core/src/clp/DictionaryWriter.hpp
```diff
@@ -85,6 +85,12 @@ class DictionaryWriter {
      */
     size_t get_data_size() const { return m_data_size; }
 
+    /**
+     * Gets the number of entries in the dictionary
+     * @return Number of entries
+     */
+    size_t get_num_entries() const { return m_next_id; }
+
```
gibber9809:
```suggestion
```
Not used?

components/core/CMakeLists.txt
gibber9809: Also, not sure if it's related, but when I build the `clp-s` cmake target after configuring with `CLP_ENABLE_TESTS=0` I see hundreds of instances of:

```
<command-line>: warning: "CLP_ENABLE_PROFILING" redefined
<command-line>: note: this is the location of the previous definition
```

in the build output.

components/core/src/clp_s/log_converter/LogConverter.cpp
```diff
@@ -66,39 +94,51 @@ auto LogConverter::convert_file(
...
-                    m_num_bytes_buffered,
-                    m_parser_offset,
-                    reached_end_of_stream
-            )};
-            if (log_surgeon::ErrorCode::BufferOutOfBounds == err) {
-                break;
-            }
-            if (log_surgeon::ErrorCode::Success != err) {
-                return std::errc::no_message;
+            auto event{m_parser.next_event(buf, &m_parser_offset)};
```
gibber9809, line 107: When converting hive-24hr dataset with log-converter I saw a crash happen inside of this line.

The particular file I saw the crash on appears to be `hive-24hr/i-6fca0960/application_1427088391284_0027/container_1427088391284_0027_01_000083/syslog`.

Didn't look into the internals, but as far as I can tell it happened on the first call to `m_parser.next_event()` after the last time the buffer was refilled for that input file. Specifically, after succesfully parsing ~510 log events we do one final ~14KB read leaving the (I think) 64KB buffer not totally full, then the following `next_event` call crashed.

I did observe though that when we break out of this loop to refill the buffer for the last time `prev_event_end` has a value of `116270`, and `m_parser_offset` gets set to that value, which exceeds the `65536B` buffer size, so if I had to guess that could be causing a problem.

components/core/src/clp_s/ArchiveWriter.hpp
```diff
@@ -304,6 +340,31 @@ class ArchiveWriter {
      */
     [[nodiscard]] std::pair<size_t, size_t> store_tables();
 
+    /**
+     * Compresses, stores, and clear the log shape statistics. The stats are not cleared if the
```
gibber9809, line 344:
```suggestion
     * Compresses, stores, and clears the log shape statistics. The stats are not cleared if the
```

components/core/src/clp_s/ArchiveWriter.hpp
```diff
@@ -304,6 +340,31 @@ class ArchiveWriter {
...
+     * result is an error.
+     * @return A result containing the compressed size in bytes or an error code indicating the
+     * failure:
+     * - clpp::ClppErrorCodeEnum::Unsupported if not using CLP+.
+     */
+    [[nodiscard]] auto close_log_shape_stats() -> ystdlib::error_handling::Result<size_t>;
+
+    /**
+     * Compresses, stores, and clear the parent rule shapes. The stats are not cleared if the result
+     * is an error.
```
gibber9809, line 354:
```suggestion
     * Compresses, stores, and clears the parent rule shapes. The stats are not cleared if the
     * result is an error.
```

components/core/src/clp_s/ArchiveWriter.hpp
```diff
@@ -288,7 +302,29 @@ class ArchiveWriter {
...
     }
 
+    /**
+     * Update the log shape dictionary for the given log shape, adding it to the dictionary if
+     * necessary.
+     * @param log_shape
+     * @return The log shape ID.
+     * @return True if the log shape is a new entry in the dictionary, false otherwise.
+     * @return ClppErrorCodeEnum::Unsupported if experimental stats are not enabled.
+     */
```
gibber9809, line 312: Our coding guidelines are a little bit vague about this specfic case (result containing a pair), but generally the standard seems to be something like:

```
* @return A result containing a pair or the error code indicating the failure:
* The pair:
* - ...
* The possible error codes:
* - ...
```
Or for shorter descriptions, something like:
```
* @return A result containing a pair holding ...xyz..., or an error code indicating the failure:
* - ...error code descriptions..
```

components/core/src/clp_s/ArchiveWriter.hpp
```diff
@@ -288,7 +302,29 @@ class ArchiveWriter {
         return rc;
     }
 
+    /**
+     * Update the log shape dictionary for the given log shape, adding it to the dictionary if
```
gibber9809, line 306:
```suggestion
     * Updates the log shape dictionary for the given log shape, adding it to the dictionary if
```

components/core/src/clp_s/JsonParser.cpp
gibber9809: Thought about this a bit more, and I think there are potentially some correctness issues with search if we don't preserve these floats losslessly in clp+, though I'm not sure whether you can hit the problems without wildcard search.

For example, say that we parse a float according to some rule:
```my_float: \d\.\d\d\d\d\de\d```

And we parse the log message:
```text 1.23456e2``` into ```text %my_float%``` and the variable `1.23456e2` which we store as a raw float without preserving its format.

Now if the user issues a query like `text 1.23*` we should match that message, but if we marshal that float as `123.456` (since we didn't preserve its format), then we will incorrectly fail to match the log message.

File: components/core/src/clp_s/Schema.hpp [hunk 23]
```diff
@@ -1,16 +1,238 @@
...
+            [&](UnorderedObject const& obj) -> bool {
+                if (obj.log_shape_id.has_value()) {
+                    log_shape_id = obj.log_shape_id;
+                    return true;
+                }
+                return false;
+            }
+    );
+    return log_shape_id;
+}
```
gibber9809, line 204: I think this/the corresponding code in `ClppMatcher` has a bug? If a JSON log has multiple fields containing a log message then it looks like we would treat its log shape as only the log shape of the first log message found in the schema.

Haven't read `ClppMatcher` in detail yet, but I'm guessing that could lead to missing search results.

components/core/src/clp_s/Schema.hpp
```diff
@@ -1,16 +1,238 @@
...
+/**
+ * An unordered object is stored in a schema as a delimiter entry followed by any metadata entries
+ * and the object's sub-schema.
+ *
+ * Metadata for each node type:
+ * - LogMessage:
+ *  - root MST node ID
+ *  - log shape ID
+ * - ParentRule:
+ *  - root MST node ID
```
gibber9809, line 137: I guess having the root MST node ID is a workaround so that we can find the root of the subtree containing a parent rule/log message with no actual columns (e.g., because the log message is static text with no variables) + be able to actually marshal the log message into the correct field in this situation?

Also, on terminology, could we use "MPT subtree root node ID" here + elsewhere in this file to make it a little bit more clear that these are the roots of LogMessage and ParentRule subtrees respectively? On first read I got confused and thought that this was talking about the lowest level root(s) in the MPT (i.e. nodes with parent ID -1).

components/core/src/clp_s/Schema.hpp
```diff
@@ -1,16 +1,238 @@
...
+
+/**
+ * A read-only view over a sequence of schema entries.
+ *
+ * Each top-level entry is either an MST node ID or the delimiter of an unordered object.
+ * `visit_entries` iterates these entries and decodes unordered objects into `UnorderedObject`s.
+ * `any_node_id` and `for_each_node_id` visit every MST node ID, recursing into unordered-object
+ * sub-schemas.
+ */
+class SchemaView {
```
gibber9809, line 25: Looks like you've adapted what we talked about re: special cases for things in the schema that aren't node IDs by capturing all of the special cases here + guarding them behind a visitor pattern. I think that approach generally makes sense to me + should help clean up other parts of the code; I'll come back and review the changes here in more detail when I start looking at the decompression + search code.

components/core/src/clp_s/SchemaTree.cpp
```diff
@@ -120,14 +159,36 @@ auto SchemaTree::store(std::string const& archives_dir, int compression_level) -
...
-int32_t SchemaTree::find_matching_subtree_root_in_subtree(
-        int32_t const subtree_root_node,
-        int32_t node,
-        NodeType type
-) const {
-    int32_t earliest_match = -1;
+auto SchemaTree::build_ls_rule_name(int32_t node_id) const -> std::string {
+    std::vector<std::string_view> names;
+    auto cur_id{node_id};
+    while (-1 != cur_id) {
```
gibber9809, line 165:
```suggestion
    while (constants::cRootNodeId != cur_id) {
```

components/core/src/clp_s/ArchiveWriter.cpp
gibber9809: Leaving a note for myself to think about this again once I read through the search code.

components/core/src/clpp/Array.hpp
```diff
@@ -0,0 +1,82 @@
...
+    return ystdlib::error_handling::success();
+}
+
+template <typename Element, typename Index>
+auto Array<Element, Index>::at_or_create(Index id, Element e) -> Element& {
+    if (m_array.size() <= id) {
+        m_array.resize(id + 1);
+        m_array.at(id) = e;
+    }
+    return m_array.at(id);
```
gibber9809, line 79: This interface seems a little awkward, since it inserts element `e`, but _only_ if its called with a contiguous series of `id`, otherwise insertion is skipped -- I think that makes it a bit too easy for the caller to accidentally miss initializing some element. Might make sense to have the underlying array be `std::vector<std::optional<Element>>`, then always set uninitialized values in `at_or_create` based on both the ID being new or the value being an empty option + have the `compress` behaviour for uninitialized elements be `Element{}.compress(compressor)`?

components/core/src/clpp/Defs.hpp
```diff
@@ -0,0 +1,54 @@
...
+    EncodingType type;
+    std::string_view name;
+    std::string_view pattern;
+};
+
+inline constexpr std::array cEncodingPatterns{
+        EncodingPattern{
+                .type = EncodingType::Float,
+                .name = "float",
+                .pattern = R"(-?\d+\.\d+([eE]-?\d+)?)"
```
gibber9809, line 36: I think the `.` should be optional when there is an exponent, e.g. `1e3` is a float as well.

Probably need to express that as an alternation like `R"(-?\d+((\.\d+)|((\.\d+)?[eE]-?\d+))"`?

components/core/src/clpp/TextShape.hpp
```diff
@@ -0,0 +1,313 @@
...
+        std::string result;
+        result.reserve(literal_segment.size());
+        for (size_t i{0}; i < literal_segment.size();) {
+            if ('%' == literal_segment.at(i) && i + 1 < literal_segment.size()
+                && '%' == literal_segment.at(i + 1))
+            {
+                result += '%';
+                i += 2;
+            } else {
+                result += literal_segment.at(i);
```
gibber9809, line 93:
```suggestion
                result.push_back(literal_segment.at(i));
```

components/core/src/clpp/TextShape.hpp
```diff
@@ -0,0 +1,313 @@
...
+     */
+    [[nodiscard]] static auto unescape_literal_text(std::string_view literal_segment)
+            -> std::string {
+        std::string result;
+        result.reserve(literal_segment.size());
+        for (size_t i{0}; i < literal_segment.size();) {
+            if ('%' == literal_segment.at(i) && i + 1 < literal_segment.size()
+                && '%' == literal_segment.at(i + 1))
+            {
+                result += '%';
```
gibber9809, line 90:
```suggestion
                result.push_back('%');
```

components/core/src/clpp/TextShape.hpp
```diff
@@ -0,0 +1,313 @@
...
+     * Reverses the escaping applied to a literal text segment (replaces every '%%' with '%').
+     *
+     * This function is only safe when used with a single literal segment. It cannot distinguish
+     * whether '%%' is an escaped literal or the closing+opening delimiters of two adjacent
+     * placeholders.
+     *
+     * @param shape The literal segment to unescape.
+     * @return An unescaped copy of the literal segment.
+     */
+    [[nodiscard]] static auto unescape_literal_text(std::string_view literal_segment)
```
gibber9809, line 82: Not sure if this code is on the critical path or not, but if it is it might be worth copying the style of implementation used in `StringUtils::escape_json_string` at `clp_s/Utils.cpp:246` or `EncodedTextAst::decode` at `clp/ffi/EncodedTextAst.hpp:192` where for regular unescaped text you just keep track of the range of regular unescaped characters, then append them all in one chunk once you hit either the end of the string or an escape character. IIRC it is much faster since the compiler can turn your happy path into a scan + memcpy more or less.

Your call though, since I haven't profiled this stuff myself.

gibber9809: Actually looking at the usage of this + the `segments()` API in marshalling it should probably be fine since it's once per ERT + if we did want to increase overall performance here we'd probably need to also redesign `segments()` to be a non-allocating iterator-style API, so should be fine to leave this unless we have some datasets where it really shows up in profiling.

components/core/src/clpp/TextShape.hpp
```diff
@@ -0,0 +1,313 @@
...
+            auto const open{find_placeholder_opening(pos)};
+            if (std::string_view::npos == open) {
+                result.emplace_back(Segment::Type::Literal, shape.substr(pos));
+                return result;
+            }
+            if (open > pos) {
+                result.emplace_back(Segment::Type::Literal, shape.substr(pos, open - pos));
+            }
+            // The closing placeholder delimiter is always the next '%' after the opening one as a
+            // column name cannot contain a delimiter.
```
gibber9809, line 141: Is this (making sure rule names can't contain %) enforced when parsing the spec file in log-surgeon right now?

components/core/src/clpp/TextShape.hpp
```diff
@@ -0,0 +1,313 @@
...
+            auto const close{shape.find('%', open + 1)};
+            if (std::string_view::npos == close) {
+                result.emplace_back(Segment::Type::Literal, shape.substr(open));
+                return result;
+            }
```
gibber9809, line 146: Can we hit this case for a valid text shape?

components/core/src/clpp/TextShape.hpp
gibber9809: Same comment goes for a lot of the rest of the implementations in this class -- looks like there's a lot of optimization opportunities that could help overall performance for degenerate specs/datasets with a ton of log shapes, but should be fine to leave it as is for an initial release.

components/core/src/clpp/ParentRuleShapes.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace clpp {
+auto ParentRuleShapes::compress(clp_s::ZstdCompressor& compressor) const
+        -> ystdlib::error_handling::Result<void> {
+    compressor.write_numeric_value(m_parent_rule_shapes.size());
+    for (auto const& shape : m_parent_rule_shapes) {
+        compressor.write_numeric_value(shape.m_name.size());
+        compressor.write_string(shape.m_name);
+        compressor.write_numeric_value(shape.m_start);
+        compressor.write_numeric_value(shape.m_size);
```
gibber9809, line 20:
```suggestion
    compressor.write_numeric_value<uint64_t>(m_parent_rule_shapes.size());
    for (auto const& shape : m_parent_rule_shapes) {
        compressor.write_numeric_value<uint64_t>(shape.m_name.size());
        compressor.write_string(shape.m_name);
        compressor.write_numeric_value<uint64_t>(shape.m_start);
        compressor.write_numeric_value<uint64_t>(shape.m_size);
```
Just because if we compile this for a 32-bit target (like wasm) we can accidentally change the archive format if we're writing/reading `size_t` directly. Could also choose `uint32_t` or something here, just need to be explicit with whatever we choose here + on the read side.

components/core/src/clpp/LogShapeStat.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include <clp_s/ErrorCode.hpp>
+#include <clp_s/ZstdCompressor.hpp>
+#include <clp_s/ZstdDecompressor.hpp>
+#include <clpp/ErrorCode.hpp>
+
+namespace clpp {
+auto LogShapeStat::compress(clp_s::ZstdCompressor& compressor) const
+        -> ystdlib::error_handling::Result<void> {
+    compressor.write_numeric_value(m_count);
```
gibber9809, line 13:
```suggestion
    compressor.write_numeric_value<uint64_t>(m_count);
```
Same as my other comment -- should update the read and write side to use some explicit integer type besides `size_t`.

components/core/src/clpp/Array.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+private:
+    // Data members
+    std::vector<Element> m_array;
+};
+
+template <typename Element, typename Index>
+auto Array<Element, Index>::compress(clp_s::ZstdCompressor& compressor)
+        -> ystdlib::error_handling::Result<void> {
+    compressor.write_numeric_value(m_array.size());
```
gibber9809, line 52:
```suggestion
    compressor.write_numeric_value<uint64_t>(m_array.size());
```
Same as other comments -- need to use explicit integer type for read/write side to avoid issue with `size_t` changing size between 32/64-bit platforms.

gibber9809 (changes requested):
Partial review focusing mostly on the parsing/write side. Should be mostly small changes, though I did spot one potential bug with find_log_shape_id/ClpMatcher that may have slightly bigger scope.

Also took a look at `log-converter` since I ran into a crash trying to use this branch's version.

components/core/src/clp_s/JsonParser.hpp
davidlion: That is fair, I'm also not happy with that name. Iirc, I used to have something closer to `parse_log_message`, but I was thinking that really it can parse any string not just log messages. `parse_unstructured_string_field` (or `unstructured_text_field`) is probably might be more accurate, but I'm fine with `parse_log_message` since realistically that is the purpose.

If you have a preference to anything lmk and I'll switch to that.

components/core/CMakeLists.txt
davidlion: Sorry, is a mistake from merging the PR back into this branch. I worked backwards from the warning to remove this earlier today before reading this comment.

components/core/src/clp/string_utils/string_utils.cpp
davidlion: I think that is worthwhile. Fwiw, the reason I left fast_float in is because I did notice a non-trivial improvement when the encoding was still buggy (and trying to convert everything).

components/core/src/clp_s/JsonParser.cpp
davidlion: Ya I don't like the name either (arguably it isn't super accurate either because that index isn't the "start of a schema" it is the scope's start within the schema).

I feel like I always associate "handle" with something that wraps a resource/object. I don't think I've ever used it for an index. What do you think of `scope_schema_idx`, `scope_start_schema_idx`. I got pretty confused working with schema entries so I personally would prefer there be some sort of relationship. At one point I was trying to refactor this stuff even more, but it was taking too long.

If you feel strongly about any name, I'll go with what you prefer.

components/core/src/clp_s/Schema.hpp
davidlion: You're right that this case won't behave correctly. I was under the (probably completely wrong) impression we weren't handling it since it causes non-trivial issues. Although, thinking about it again, it might not be "that" bad to fix. Iirc mongodb has this, so it is probably necessary, but I wouldn't expect "good" results (compression or search) on data with this property.

I think the only change is that schemas need to be keyed on an ordered list of log shapes. Conceptually that isn't too bad, but I think it touches a bunch of code. That or we just manually walk the entire schema everywhere.

This is one of those cases where I wanted a non-trivial design fix as separating the JSON and unstructured components of the schema into separate tables would likely be an important improvement. Having the schema be unique on the JSON fields + the log shape already has a lot of potential to create many small schemas/tables in mixed data. Having multiple log shapes makes this increasingly bad as (you'd think) the likely hood of the two unstructured logs having the same shape is probably low.

components/core/src/clpp/Array.hpp
davidlion: Hmm I need to think about this a bit.

components/core/src/clpp/TextShape.hpp
davidlion: That's correct. Although, now I realize it isn't explicitly stated in the log-surgeon docs. Iirc the only symbol is `_`.

components/core/src/clpp/TextShape.hpp
davidlion: No, good point. Do you think throwing is the best option since this basically means a bug occurred or there is corruption? Or what do you think is the best option?

components/core/src/clpp/ParentRuleShapes.cpp
davidlion: Does this make sense to you:

If 32-bit targets are a legitimate case, we probably want to explicitly limit everything to `uint32_`. Otherwise, I think we'd need to explicit write the size of size_t of the archive writer (for the reader to use).

components/core/src/clp_s/JsonParser.cpp
gibber9809: I think `scope_schema_idx` sounds good to me.

components/core/src/clp_s/JsonParser.hpp
gibber9809: Yeah lets go with `parse_log_message` then.

components/core/src/clpp/TextShape.hpp
gibber9809: Yeah I think throw/error is probably the right thing to do.

components/core/src/clpp/ParentRuleShapes.cpp
gibber9809: Right now all of the field sizes are explicitly defined in the format (there's a lot of 64-bit fields that really don't need to be 64-bit, but we can't change them because of backwards compatibility). When we added support for wasm what we did was we went through all of the 64-bit fields in the archive format that need to get treated as a `size_t` in the code after decompression and added an extra check to ensure that the uint64 in the archive could be converted to `size_t` on the target platform (using the `ReaderUtils::try_uint64_to_size_t` utility). Basically the 32-bit wasm path handles this problem by explicitly error-ing out on some archives.

But yeah, since for legitimate use-cases these values should fit within a `uint32_t` anyway I'd be fine with just using `uint32_t` in the format for all the places I pointed out in my review and making our lives easier.

components/core/src/clp_s/CommandLineArguments.hpp
```diff
@@ -228,6 +242,13 @@ class CommandLineArguments {
 
     void print_search_usage() const;
 
+    /**
+     * Validate the use of experimental features. Requires the program options to have been parsed.
+     * @throws std::invalid_argument if any experimental feature is used without setting the
+     * experimetnal flag.
+     */
+    auto validate_experimental() const -> void;
+
```
gibber9809:
```suggestion
```
Not used & there doesn't seem to be an implementation?

components/core/src/clp_s/SchemaTree.hpp
```diff
@@ -64,78 +65,116 @@ auto node_to_literal_type(NodeType type) -> clp_s::search::ast::LiteralType;
...
+    // Static methods
+    /**
+     * Converts a node type to a literal type.
+     * @param type
+     * @return A literal type
+     */
+    static auto node_to_literal_type(NodeType type) -> clp_s::search::ast::LiteralType;
+
+    // Constructors
+    SchemaNode() : m_parent_id(-1), m_id(-1), m_type(NodeType::Integer) {}
```
gibber9809: Probably worth just directly default initializing `m_parent_id`, `m_id`, and `m_type` as done here, then make this constructor `= default`.

components/core/src/clp_s/SchemaTree.cpp
```diff
@@ -112,6 +150,7 @@ auto SchemaTree::store(std::string const& archives_dir, int compression_level) -
         schema_tree_compressor.write_numeric_value(static_cast<uint64_t>(key.size()));
         schema_tree_compressor.write(key.data(), key.size());
         schema_tree_compressor.write_numeric_value(node.get_type());
+        schema_tree_compressor.write_numeric_value(node.get_count());
```
gibber9809: Since this is a breaking format change that'll force us to maintain a branch on the read-side, I want to double-check whether you think that this is an important feature or just a nice-to-have?

I don't think this particular change will be a ton of maintenance burden, but I still want to make sure there's a good reason for us to maintain it.

components/core/src/clp_s/ReaderUtils.cpp
```diff
@@ -53,7 +55,13 @@ std::shared_ptr<SchemaTree> ReaderUtils::read_schema_tree(ArchiveReaderAdaptor&
...
         }
 
-        tree->add_node(parent_id, (NodeType)node_type, key);
+        uint32_t count{0};
+        error_code = schema_tree_decompressor.try_read_numeric_value(count);
+        if (ErrorCodeSuccess != error_code) {
+            throw OperationFailed(error_code, __FILENAME__, __LINE__);
+        }
+
+        tree->add_node(parent_id, static_cast<NodeType>(node_type), key, count);
```
gibber9809: Related to my other comment on the schema tree: if we want to keep this then we'll need to add a branch here to only attempt to load this `count` field for archive versions >= this change.

components/core/src/clp_s/SingleFileArchiveDefs.hpp
```diff
@@ -107,10 +107,10 @@ enum struct ArchiveMetadataPacketType : uint8_t {
 };
 
 struct ArchiveInfoPacket {
-    uint64_t num_segments;
-    // TODO: Add more fields in the future
+    uint64_t num_segments{};
+    bool experimental{false};
 
-    MSGPACK_DEFINE_MAP(num_segments);
+    MSGPACK_DEFINE_MAP(num_segments, experimental);
```
gibber9809: Minor nit: I think this is fine, but as is it will make it so that the format will always have this `experimental` flag going forwards. Do you think it would make sense to use an 8-bit enum here instead (e.g. `archive_type`, with clp_structured=0, clp_plus_experimental=1, etc.) to give us some more flexibility/be more specific in case we try experimenting with more flavours of the archive format over time and want to distinguish between them?

components/core/src/clp_s/search/kql/kql.cpp
```diff
@@ -52,6 +56,69 @@ using generated::KqlLexer;
...
+    if (auto const* value{std::any_cast<std::shared_ptr<ast::Value>>(&parsed)}) {
+        if (auto func_call{std::dynamic_pointer_cast<ast::FunctionCall>(*value)}) {
+            auto const& function_name{func_call->get_function_name()};
+            auto const& args{func_call->get_args()};
+            if (args.size() != 1) {
+                SPDLOG_ERROR("Function '{}' requires exactly one column argument", function_name);
+                return nullptr;
+            }
+            auto column{std::dynamic_pointer_cast<ColumnDescriptor>(args.at(0))};
+            if (!column) {
```
gibber9809:
```suggestion
            if (nullptr == column) {
```

components/core/src/clp_s/search/kql/kql.cpp
```diff
@@ -52,6 +56,69 @@ using generated::KqlLexer;
...
+                SPDLOG_ERROR("Function '{}' argument must be a column", function_name);
+                return nullptr;
+            }
+            if (function_name == clpp::cShapeFunction) {
```
gibber9809:
```suggestion
            if (clpp::cShapeFunction == function_name) {
```

components/core/src/clp_s/search/kql/kql.cpp
```diff
@@ -52,6 +56,69 @@ using generated::KqlLexer;
...
+                column->set_subtree_type(std::string{clpp::cShapeFunction});
+            } else {
+                SPDLOG_ERROR("Function '{}' is not supported in filter expressions", function_name);
+                return nullptr;
+            }
+            return column;
```
gibber9809:
```suggestion
            if (clpp::cShapeFunction == function_name) {
                column->set_subtree_type(std::string{clpp::cShapeFunction});
                return column;
            }
            SPDLOG_ERROR("Function '{}' is not supported in filter expressions", function_name);
            return nullptr;
```

components/core/src/clp_s/search/kql/kql.cpp
```diff
@@ -237,8 +329,13 @@ class ParseTreeVisitor : public KqlBaseVisitor {
         return ctx->q->accept(this);
     }
 
-    std::any visitColumn_value_expression(KqlParser::Column_value_expressionContext* ctx) override {
-        auto descriptor = std::any_cast<std::shared_ptr<ColumnDescriptor>>(ctx->col->accept(this));
+    auto visitColumn_value_expression(KqlParser::Column_value_expressionContext* ctx)
+            -> std::any override {
+        auto parsed = ctx->col->accept(this);
+        auto descriptor{extract_column_from_parsed_value(parsed)};
+        if (!descriptor) {
```
gibber9809:
```suggestion
        if (nullptr == descriptor) {
```

components/core/src/clp_s/search/kql/kql.cpp
```diff
@@ -258,8 +355,13 @@ class ParseTreeVisitor : public KqlBaseVisitor {
         }
     }
 
-    std::any visitColumn_range_expression(KqlParser::Column_range_expressionContext* ctx) override {
-        auto descriptor = std::any_cast<std::shared_ptr<ColumnDescriptor>>(ctx->col->accept(this));
+    auto visitColumn_range_expression(KqlParser::Column_range_expressionContext* ctx)
+            -> std::any override {
+        auto parsed = ctx->col->accept(this);
+        auto descriptor{extract_column_from_parsed_value(parsed)};
+        if (!descriptor) {
```
gibber9809:
```suggestion
        if (nullptr == descriptor) {
```

components/core/src/clp_s/search/ast/StringLiteral.hpp
gibber9809: Leaving myself a note to double-check this after reading the rest of the search code since it still seems to be in a temporary state.

components/core/src/clp_s/search/ast/SearchUtils.hpp
```diff
@@ -6,10 +6,19 @@
...
 #include "Literal.hpp"
 
 namespace clp_s::search::ast {
+/**
+ * Builds a dotted column name string from a column descriptor's tokens.
+ * @param column
+ * @return The dotted column name.
+ */
+[[nodiscard]] auto column_descriptor_to_string(ColumnDescriptor& column) -> std::string;
+
```
gibber9809: Looks like this doesn't handle escaping `.` or anything else in the tokens. Since it looks like this is only used for error logs that's fine, but could we add some detail to the docstring explaining that this function doesn't do escaping?

components/core/src/clp_s/search/ast/NarrowTypes.cpp
```diff
@@ -62,10 +62,14 @@ std::shared_ptr<Expression> NarrowTypes::narrow(std::shared_ptr<Expression> cur)
...
-                column->remove_matching_type(LiteralType::IntegerT);
+                if (false == literal->has_wildcards()) {
+                    column->remove_matching_type(LiteralType::IntegerT);
+                }
             }
             if (false == literal->as_float(tmpdouble, op)) {
-                column->remove_matching_type(LiteralType::FloatT);
+                if (false == literal->has_wildcards()) {
+                    column->remove_matching_type(LiteralType::FloatT);
+                }
```
gibber9809: EncodedVariableInterpreter has these methods `wildcard_string_could_be_representable_integer_var` and `wildcard_string_could_be_representable_float_var` that try to prove whether a wildcard string could possibly match an int/float. We can't use the float version directly (since it seems to not support scientific notation), but we could apply the concept pretty easily -- they just do a simple check to make sure that every character in the wildcard string could possibly appear somewhere in an int/float value.

I think it would be worth checking here and/or adding methods to `Literal` to do this kind of heuristic check, since we can pretty easily avoid trying wildcard search on int/float columns for cases that are obviously not going to find matches.

components/core/src/clp_s/search/ast/FunctionCall.cpp
```diff
@@ -0,0 +1,78 @@
...
+        m_operands.push_back(std::move(arg));
+    }
+}
+
+auto FunctionCall::print() const -> void {
+    auto& out = get_print_stream();
+    out << "FunctionCall(" << m_function_name << ", [";
+    bool first{true};
+    for (auto it = op_begin(); it != op_end(); ++it) {
+        if (!first) {
```
gibber9809:
```suggestion
        if (false == first) {
```

components/core/src/clp_s/search/ast/ColumnDescriptor.hpp
```diff
@@ -282,26 +293,64 @@ class ColumnDescriptor : public Literal {
...
+
 private:
+    // Static data members
+    static id_t m_next_id;
+
+    // Data members
     DescriptorList m_descriptors;  // list of descriptors describing the column
     DescriptorList m_unresolved_tokens;  // unresolved tokens used for array search
     std::string m_namespace;
     literal_type_bitmask_t m_flags;  // set of types this column can match
```
gibber9809: While we're here, can we default initialize `m_flags` to `cAllTypes`.

components/core/src/clp_s/search/ast/ColumnDescriptor.hpp
```diff
@@ -282,26 +293,64 @@ class ColumnDescriptor : public Literal {
...
-    bool m_unresolved_descriptors;  // true if contains wildcards
-    bool m_pure_wildcard;  // true if column is single wildcard
+    int32_t m_schema_col_id{-1};  // unambiguous CLJ column id this column represents. May be unset.
+    bool m_unresolved_descriptors{false};  // true if contains wildcards
+    bool m_pure_wildcard{false};  // true if column is single wildcard
     std::optional<std::string> m_subtree_type;  // optional subtree type to resolve against
 
+    bool m_is_clpp_resolved{false};
+    std::optional<size_t> m_leaf_position;
+    uint64_t m_id;
```
gibber9809:
```suggestion
    uint64_t m_id{};
```

components/core/src/clp_s/search/ast/ColumnDescriptor.hpp
```diff
@@ -282,26 +293,64 @@ class ColumnDescriptor : public Literal {
...
+
+    /**
+     * Pins this column to a single leaf placeholder of the enclosing LogMessage.
+     * @param position The LogMessage-relative leaf placeholder position.
+     */
+    auto set_leaf_position(size_t position) -> void { m_leaf_position = position; }
+
 private:
+    // Static data members
+    static id_t m_next_id;
```
gibber9809: Haven't read all of the schema match code yet, but what's the motivation for this `m_next_id`/the new `m_id` field/the new `copy_with_new_id` method? Do we sometimes need a unique identifier for a column descriptor that's stable across deep copies now? If we just need a unique identifier in general then the address of a `ColumnDescriptor` is sufficient.

gibber9809 (changes requested):
Leaving another partial view covering all of the AST and KQL changes. Also took a look through some parts of the read-side code, as well as the indexer changes, but I still have to look at all of the most complicated parts of schema/column resolution, ERT scanning, projection, and marshalling code.

Besides the code, I did notice one unexpected behaviour when experimenting with the cli -- when searching without using any projection options my results would be returned as `{"message": {"text": "actual log message"}}` whereas when going through decompression my results would be returned as `{"message": "actual log message"}`. I'm guessing this has something to do with cases where a user projects both some leaf columns as well as the full log message, but it seemed a bit awkward as default behaviour when no projection is being applied. I'd personally prefer just getting the original log message back as the default behaviour, but I'm open to keeping it as is if it needs to be this way for some reason or most people prefer the current behaviour.

components/core/src/clp_s/search/ast/ColumnDescriptor.hpp
gibber9809: That aside, it's also worth noting that this code can run in a multithreaded setting when we pull it into our presto connector. Seems like there would be a benign race on incrementing `m_next_id` at the same time from different threads in that setting.

PR #2276, opened 2026-05-13
fix(helm): Re-initialize results cache on pod restart (fixes #2257).

### LINKED ISSUE #2257: Results cache (MongoDB) fails to recover after PVC recreation due to uninitialized replica set

hoophalab:
# Description

This PR fixes #2257 by rerunning initialization script every time the bundled results cache restarts. The script is idempotent so safe to rerun.

# Checklist

  breaking change.

# Validation performed

Force a crash on bundled results cache, the initialization scripts container is recreated and rerun in events.

## Summary by CodeRabbit

* **Refactor**
  * Standardized results-cache readiness handling across services for consistent startup sequencing.

* **New Features**
  * Added an init-step and readiness gating for the results-cache to support in-cluster initialization and health checks.

* **Chores**
  * Made results-cache initialization conditional between bundled and non-bundled deployments and bumped chart version.

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/clp/pull/2276?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


tools/deployment/package-helm/templates/results-cache-statefulset.yaml
```diff
@@ -56,6 +56,32 @@ spec:
...
+            - "/bin/sh"
+            - "-c"
+            - |
+              until python3 -c "import socket; s=socket.socket(); s.settimeout(2); s.connect(('127.0.0.1', 27017)); s.close()" 2>/dev/null; do
+                sleep 1
+              done
+              python3 -u -m clp_py_utils.initialize-results-cache \
+                --uri "mongodb://127.0.0.1:27017/{{ .Values.clpConfig.results_cache.db_name }}" \
+                --member-host "{{ include "clp.fullname" . }}-results-cache:27017" \
+                --stream-collection {{ .Values.clpConfig.results_cache.stream_collection_name | quote }}
```
junhaoliao, line 75: `--member-host` is not a valid argument for `clp_py_utils.initialize-results-cache` (it only accepts `--uri` and `--stream-collection` — see `initialize-results-cache.py:117-123`). argparse will error with `unrecognized arguments: --member-host`, causing the sidecar to always fail.

additionally, the script derives the replica set member host from the URI's netloc. using `127.0.0.1` means `rs.initiate()` registers the member as `127.0.0.1:27017`, which is unreachable from other pods — the same `ServerSelectionTimeoutError` described in #2257.

can we use the service hostname in the URI instead? this way the replica set member is registered correctly and `--member-host` is unnecessary.

```suggestion
              python3 -u -m clp_py_utils.initialize-results-cache \
                --uri "mongodb://{{ include "clp.fullname" . }}-results-cache:27017/{{ .Values.clpConfig.results_cache.db_name }}" \
                --stream-collection {{ .Values.clpConfig.results_cache.stream_collection_name | quote }}
```

tools/deployment/package-helm/templates/results-cache-statefulset.yaml
```diff
@@ -56,6 +56,32 @@ spec:
...
+              touch /tmp/init-complete
+              sleep infinity
```
junhaoliao, line 77: there's no error handling between the python3 command and `touch /tmp/init-complete`. if the init script fails (e.g., the argparse error from the `--member-host` finding, or a transient MongoDB connection failure), the shell continues to `touch`, making the sidecar report as ready. the pod becomes ready, consumers start connecting, and queries fail because the replica set is uninitialized.

can we add `set -e` so the sidecar exits (and restarts) on failure instead of silently reporting ready?

```suggestion
            - |
              set -e
              until python3 -c "import socket; s=socket.socket(); s.settimeout(2); s.connect(('127.0.0.1', 27017)); s.close()" 2>/dev/null; do
                sleep 1
              done
              python3 -u -m clp_py_utils.initialize-results-cache \
                --uri "mongodb://{{ include "clp.fullname" . }}-results-cache:27017/{{ .Values.clpConfig.results_cache.db_name }}" \
                --stream-collection {{ .Values.clpConfig.results_cache.stream_collection_name | quote }}
              touch /tmp/init-complete
              sleep infinity
```

junhaoliao (changes requested):

junhaoliao (approved):

Issue #2048, opened 2026-02-27
Add resource requests and limits to all containers


junhaoliao:
## Request

No containers in the Helm chart or docker-compose define CPU/memory resource requests or
limits. This applies to all 14+ containers:

**Helm chart** (`tools/deployment/package-helm/templates/`): api-server, webui,
compression-scheduler, compression-worker, query-scheduler, query-worker, reducer,
log-ingestor, garbage-collector, mcp-server, database, queue, redis, results-cache,
and the `bitnami/kubectl` init containers.

**Docker Compose** (`tools/deployment/package/docker-compose-all.yaml`): All services
use `deploy:` only for `replicas`, with no `resources:` block.

This causes problems:
- **Kubernetes autoscaling is blind**: Karpenter and Cluster Autoscaler use resource
  requests to determine how many and what size nodes to provision. Without requests, they
  cannot right-size nodes, leading to over- or under-provisioning.
- **No OOM protection**: Without memory limits, a single runaway container (e.g., a
  compression worker processing a large file) can consume all node memory and trigger
  kernel OOM kills on unrelated pods.
- **No QoS guarantees**: Kubernetes assigns the `BestEffort` QoS class to pods without
  resource requests, meaning they are the first to be evicted under memory pressure.
- **Docker Compose resource waste**: Without limits, local development environments may
  consume excessive host resources.

## Possible implementation

**Helm chart**: Add a `resources` block to `values.yaml` for each component so users can
tune values for their workload. Include sensible defaults:

```yaml
compressionWorker:
  replicas: 1
  resources:
    requests:
      cpu: ""
      memory: ""
    limits:
      cpu: ""
      memory: ""

# Same structure for: queryWorker, reducer, apiServer, webui,
# compression-scheduler, query-scheduler, garbage-collector,
# log-ingestor, mcp-server, database, queue, redis, results-cache
```

Specific values should be determined through profiling representative workloads.

Then in each deployment/statefulset template, add:

```yaml
resources:
  {{- toYaml .Values.<component>.resources | nindent 12 }}
```

**Docker Compose**: Add resource limits under the existing `deploy:` blocks:

```yaml
database:
  deploy:
    replicas: "${CLP_DATABASE_ENABLED:-1}"
    resources:
      limits:
        cpus: "1.0"
        memory: "2G"
      reservations:
        cpus: "0.25"
        memory: "512M"
```

The exact values should be determined through profiling.

PR #2278, opened 2026-05-14
feat(deployment): Add OpenTelemetry Collector.

Nathan903:
# Description

* Adds an OpenTelemetry Collector sidecar to both Docker Compose and Helm deployments (only instantiated if there is `telemetry` profile/config flag)
* Injects `OTEL_SERVICE_NAME`, `OTEL_EXPORTER_OTLP_ENDPOINT`, `OTEL_RESOURCE_ATTRIBUTES`, and `CLP_DISABLE_TELEMETRY` env vars into every service/component.
* The `DockerComposeController` reads the CLP version, builds resource attributes (deployment ID, version, storage engine, OS, arch), and emits deployment topology gauges (worker replicas, concurrency) to the collector at startup.
* Telemetry is opt-out. Set `CLP_DISABLE_TELEMETRY=true` (Docker Compose) or `clpConfig.telemetry.disable: true` (Helm) to disable.

# Checklist

  breaking change.

# Validation performed

Started CLP with the local telemetry backend (from the `clp-telemetry-server` repo). Verified `otel-collector` logs showed incoming OTLP data.

test end-to-end with the telemetry server: 
1. 
```
cd ~/clp/tools/deployment/telemetry_server
docker compose up -d
```
2. change build/clp-package/etc/clp-config.yaml's endpoint to http://172.17.0.1:4318 
3. test docker compose with start-clp.sh.

4. check telemetry:

```
cd ~/clp/tools/deployment/telemetry_server
docker exec telemetry_server-clickhouse-1 clickhouse-client --user default --password clickhouse --query "SELECT * FROM clp_telemetry.otel_metrics_gauge ORDER BY TimeUnix DESC LIMIT 5 FORMAT Vertical"
```

5. test helm

```
cat <<EOF | kind create cluster --name clp --config=-
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  extraMounts:
  - hostPath: /home
    containerPath: /home
  extraPortMappings:
  - containerPort: 30000
    hostPort: 30000
  - containerPort: 30017
    hostPort: 30017
  - containerPort: 30301
    hostPort: 30301
  - containerPort: 30302
    hostPort: 30302
  - containerPort: 30306
    hostPort: 30306
  - containerPort: 30800
    hostPort: 30800
EOF

cd tools/deployment/package-helm
helm install clp . \
  --set clpConfig.telemetry.endpoint="http://172.17.0.1:4318" \
  --set clpConfig.logs_input.directory="/home" \
  --timeout 10m

kubectl wait --for=condition=ready pod --all --timeout=300s
```
6. check telemetry: 

```
cd ~/clp/tools/deployment/telemetry_server
docker exec telemetry_server-clickhouse-1 clickhouse-client --user default --password clickhouse --query "SELECT * FROM clp_telemetry.otel_metrics_gauge ORDER BY TimeUnix DESC LIMIT 5 FORMAT Vertical"
```

## test helm uuid generation

```
# Host port mappings
export CLP_WEBUI_PORT=30000
export CLP_RESULTS_CACHE_PORT=30017
export CLP_API_SERVER_PORT=30301
export CLP_LOG_INGESTOR_PORT=30302
export CLP_DATABASE_PORT=30306
export CLP_MCP_SERVER_PORT=30800

# Credentials
export CLP_DB_PASS=$(openssl rand -hex 16)
export CLP_DB_ROOT_PASS=$(openssl rand -hex 16)
export CLP_QUEUE_PASS=$(openssl rand -hex 16)
export CLP_REDIS_PASS=$(openssl rand -hex 16)

# Create the `kind` cluster
cat <<EOF | kind create cluster --name clp --config=-
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  extraMounts:
  - hostPath: /home
    containerPath: /home
  extraPortMappings:
  - containerPort: $CLP_WEBUI_PORT
    hostPort: $CLP_WEBUI_PORT
  - containerPort: $CLP_RESULTS_CACHE_PORT
    hostPort: $CLP_RESULTS_CACHE_PORT
  - containerPort: $CLP_API_SERVER_PORT
    hostPort: $CLP_API_SERVER_PORT
  - containerPort: $CLP_LOG_INGESTOR_PORT
    hostPort: $CLP_LOG_INGESTOR_PORT
  - containerPort: $CLP_DATABASE_PORT
    hostPort: $CLP_DATABASE_PORT
  - containerPort: $CLP_MCP_SERVER_PORT
    hostPort: $CLP_MCP_SERVER_PORT
EOF


cd /home/nathanhung/clp/tools/deployment/package-helm

helm install clp . \
  --set clpConfig.telemetry.disable=false \
  --set clpConfig.webui.port="$CLP_WEBUI_PORT" \
  --set clpConfig.results_cache.port="$CLP_RESULTS_CACHE_PORT" \
  --set clpConfig.api_server.port="$CLP_API_SERVER_PORT" \
  --set clpConfig.log_ingestor.port="$CLP_LOG_INGESTOR_PORT" \
  --set clpConfig.database.port="$CLP_DATABASE_PORT" \
  --set clpConfig.mcp_server.port="$CLP_MCP_SERVER_PORT" \
  --set credentials.database.password="$CLP_DB_PASS" \
  --set credentials.database.root_password="$CLP_DB_ROOT_PASS" \
  --set credentials.queue.password="$CLP_QUEUE_PASS" \
  --set credentials.redis.password="$CLP_REDIS_PASS"

kubectl wait pods --all --for=condition=Ready --timeout=300s

# get the generated UUID from the new ConfigMap:
export FIRST_UUID=$(kubectl get configmap clp-instance-id -o jsonpath='{.data.instanceId}')
echo "Generated UUID is: $FIRST_UUID"

# check that a Deployment (e.g., api-server) successfully received the exact same UUID in its OTEL_RESOURCE_ATTRIBUTES:
kubectl set env deploy/clp-api-server --list | grep OTEL_RESOURCE_ATTRIBUTES
# Look for clp.deployment.id=... and confirm it matches FIRST_UUID
 
# verify the Topology Metrics Job was also generated with the same UUID:
kubectl get job clp-topology-metrics-emitter -o yaml | grep "clp.deployment.id"
# Confirm the stringValue matches FIRST_UUID

# Perform a Helm Upgrade
helm upgrade clp . \
  --set clpConfig.telemetry.disable=false \
  --set clpConfig.webui.port="$CLP_WEBUI_PORT" \
  --set clpConfig.results_cache.port="$CLP_RESULTS_CACHE_PORT" \
  --set clpConfig.api_server.port="$CLP_API_SERVER_PORT" \
  --set clpConfig.log_ingestor.port="$CLP_LOG_INGESTOR_PORT" \
  --set clpConfig.database.port="$CLP_DATABASE_PORT" \
  --set clpConfig.mcp_server.port="$CLP_MCP_SERVER_PORT" \
  --set credentials.database.password="$CLP_DB_PASS" \
  --set credentials.database.root_password="$CLP_DB_ROOT_PASS" \
  --set credentials.queue.password="$CLP_QUEUE_PASS" \
  --set credentials.redis.password="$CLP_REDIS_PASS"

# check UUID is same
export UPGRADE_UUID=$(kubectl get configmap clp-instance-id -o jsonpath='{.data.instanceId}')
echo "Post-upgrade UUID is: $UPGRADE_UUID"

if [ "$FIRST_UUID" = "$UPGRADE_UUID" ]; then
    echo "SUCCESS: The UUID persisted correctly!"
else
    echo "FAILURE: The UUID changed during upgrade!"
fi

kubectl set env deploy/clp-api-server --list | grep OTEL_RESOURCE_ATTRIBUTES


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Optional telemetry via an OpenTelemetry collector service and a telemetry compose/profile for gathering/exporting metrics.
  * Services now receive telemetry environment configuration (OTLP endpoint, resource/service attributes, disable toggle) and a development override exposing the collector port.
  * When enabled, runtime/service/version attributes are attached and topology metrics are emitted to the collector.

* **Chores**
  * Chart version bumped and Helm templates updated to support telemetry resources and conditional collector deployment.

<!-- review_stack_entry_start -->

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/y-scope/clp/pull/2278)

<!-- review_stack_entry_end -->
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### CODE REVIEW

File: tools/deployment/package-helm/templates/_helpers.tpl [hunk 1]
```diff
@@ -50,6 +50,28 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end }}
 
+{{/*
+Provides the standard OpenTelemetry resource attributes.
+*/}}
+{{- define "clp.resourceAttributes" -}}
+clp.deployment.id={{ .Values.clpConfig.instanceId | default "00000000-0000-0000-0000-000000000000" }},service.version={{ .Chart.AppVersion }},clp.deployment.method=helm,clp.storage.engine={{ .Values.clpConfig.package.storageEngine }},os.type={{ .Values.hostOS | default "linux" }},host.arch={{ .Values.hostArch | default "amd64" }}
+{{- end -}}
+
```
junhaoliao, line 103: how about this then we don't need to ask users to override in values.yaml

```
{{/*
Provides the standard OpenTelemetry resource attributes.
*/}}
{{- define "clp.resourceAttributes" -}}
{{- $existing := lookup "v1" "ConfigMap" .Release.Namespace (printf "%s-instance-id" (include "clp.fullname" .)) }}
{{- $instanceId := uuidv4 -}}
{{- if $existing }}
{{- $instanceId = index $existing.data "instanceId" -}}
{{- end }}
clp.deployment.id={{ $instanceId }},service.version={{ .Chart.version }},clp.deployment.method=helm,clp.storage.engine={{ .Values.clpConfig.package.storageEngine }}
{{- end -}}
```

NOTE: i also proposed using .Chart.version instead of AppVersion . the chart version helps uniquely identify a helm chart which is better than just the clp package version, imo

---


also we need to add an `instance-id-configmap.yaml`
```
{{/*
Stores a unique instance ID for this Helm deployment. On first install, a UUID is
generated. On subsequent upgrades, the existing value is read via lookup so the ID
remains stable across the lifetime of this release.
*/}}
apiVersion: "v1"
kind: "ConfigMap"
metadata:
  name: {{ include "clp.fullname" . }}-instance-id
  labels:
data:
  {{- $existing := lookup "v1" "ConfigMap" .Release.Namespace (printf "%s-instance-id" (include "clp.fullname" .)) }}
  {{- if $existing }}
  instanceId: {{ index $existing.data "instanceId" | quote }}
  {{- else }}
  instanceId: {{ uuidv4 | quote }}
  {{- end }}

```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1140,6 +1173,70 @@ def _get_docker_file_name(self) -> str:
...
+            )
+
+        num_workers = self._get_num_workers()
+
+        add_gauge("clp.deployment.compression_worker_replicas", 1)
+        add_gauge("clp.deployment.compression_worker_concurrency", num_workers)
+        add_gauge("clp.deployment.query_worker_replicas", 1)
+        add_gauge("clp.deployment.query_worker_concurrency", num_workers)
+        add_gauge("clp.deployment.reducer_replicas", 1)
+        add_gauge("clp.deployment.reducer_concurrency", num_workers)
```
junhaoliao, line 1292: all three `*_replicas` gauges are hardcoded to `1`. when multi-container workers land (#1424), these will silently report wrong values. add a comment to make this explicit.

```suggestion
        # NOTE: Replicas are hardcoded to 1 until multi-container workers land (see #1424).
        add_gauge("clp.deployment.compression_worker_replicas", 1)
        add_gauge("clp.deployment.compression_worker_concurrency", num_workers)
        add_gauge("clp.deployment.query_worker_replicas", 1)
        add_gauge("clp.deployment.query_worker_concurrency", num_workers)
        add_gauge("clp.deployment.reducer_replicas", 1)
        add_gauge("clp.deployment.reducer_concurrency", num_workers)
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1140,6 +1173,70 @@ def _get_docker_file_name(self) -> str:
...
+        try:
+            payload_str = json.dumps(payload)
+            subprocess.run(
+                [
+                    "docker",
+                    "run",
+                    "--rm",
+                    "--network",
+                    f"{self._project_name}_default",
+                    "curlimages/curl",
```
junhaoliao: can we instead submit the query directly using python?

e.g. add util in general.py:
```
def http_request(
    url: str,
    data: bytes | None = None,
    headers: dict[str, str] | None = None,
    method: str = "GET",
    max_attempts: int = 2,
    retry_delay: float = 1.0,
    timeout: float = 5.0,
) -> http.client.HTTPResponse:
    """Sends an HTTP request with retry logic.

    :param url: The URL to send the request to.
    :param data: Request body as bytes.
    :param headers: Request headers.
    :param method: HTTP method.
    :param max_attempts: Number of attempts before giving up.
    :param retry_delay: Seconds to wait between retries.
    :param timeout: Seconds to wait for a response per attempt.
    :return: The HTTP response.
    :raise urllib.error.URLError: If all attempts fail due to a network-level error.
    :raise urllib.error.HTTPError: If all attempts fail with an HTTP error status.
    :raise OSError: If all attempts fail due to a socket-level error.
    """
    req = urllib.request.Request(url, data=data, headers=headers or {}, method=method)
    last_exception: Exception | None = None
    for attempt in range(max_attempts):
        try:
            return urllib.request.urlopen(req, timeout=timeout)
        except (urllib.error.URLError, urllib.error.HTTPError, OSError) as e:
            last_exception = e
            if attempt < max_attempts - 1:
                time.sleep(retry_delay)
    raise last_exception
```


then this becomes:
```
        try:
            payload_str = json.dumps(payload).encode("utf-8")
            http_request(
                "http://127.0.0.1:4318/v1/metrics",
                method="POST",
                data=payload_str,
                headers={"Content-Type": "application/json"},
            )
        except (urllib.error.URLError, urllib.error.HTTPError, OSError) as e:
            logger.warning("Failed to emit topology metrics: %s", e)

```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1140,6 +1173,70 @@ def _get_docker_file_name(self) -> str:
...
+                    "-d",
+                    payload_str,
+                    "http://otel-collector:4318/v1/metrics",
+                ],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                timeout=15,
+            )
+        except Exception as e:
+            logger.warning(f"Failed to emit topology metrics: {e}")
```
junhaoliao: f-strings in logger calls compute the string even when the log level is filtered out. use `%s`-style lazy formatting instead.

```suggestion
            logger.warning("Failed to emit topology metrics: %s", e)
```

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -655,3 +711,20 @@ services:
         "-f",
         "http://log_ingestor:3002/health"
       ]
+
+  otel-collector:
+    <<: *service_defaults
+    image: "otel/opentelemetry-collector-contrib:0.152.0"
```
junhaoliao: this uses Docker Hub (`otel/`) with version `0.152.0`, while the Helm chart (`values.yaml:254`) uses GHCR (`ghcr.io/open-telemetry/opentelemetry-collector-releases/`) with version `0.119.0`. two concerns:

1. **different registries** — Docker Hub vs GHCR (the OTel project recommends GHCR)
2. **different versions** — ~33 version gap; makes debugging harder across deployment methods

can we align on one registry and one version? if there's a reason for the divergence, consider documenting it.

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -655,3 +711,20 @@ services:
...
+    hostname: "otel_collector"
+    profiles: ["telemetry"]
+    deploy:
+      # Value must be either 0 or 1. Set to 0 to disable the otel collector.
+      replicas: "${CLP_OTEL_COLLECTOR_ENABLED:-1}"
+    user: "${CLP_THIRD_PARTY_SERVICE_UID_GID:-1000:1000}"
+    environment:
```
junhaoliao, line 681: the otel-collector has no Docker `healthcheck:` block, so `docker compose up --wait` considers it "ready" as soon as the container starts. this means `_emit_topology_metrics` (which runs after `--wait`) may race against the collector's startup. the collector already exposes `health_check` on port 13133 — adding a Docker healthcheck ensures `--wait` actually waits for the collector to be ready.

```
    healthcheck:
      <<: *healthcheck_defaults
      test: [
        "CMD",
        "curl",
        "-f",
        "http://localhost:13133/"
      ]
```

tools/deployment/package-helm/templates/otel-collector-configmap.yaml
```diff
@@ -0,0 +1,29 @@
...
+kind: ConfigMap
+metadata:
+  name: {{ include "clp.fullname" . }}-otel-collector-config
+data:
+  config.yaml: |
+    receivers:
+      otlp:
+        protocols:
+          http:
+            endpoint: 0.0.0.0:4318
```
junhaoliao: the Docker Compose otel-collector config includes a `health_check` extension on port 13133, but the Helm config omits it. without it, there's no health check endpoint for liveness probes on the collector.

```suggestion
    extensions:
      health_check:
        endpoint: "0.0.0.0:13133"

    receivers:
      otlp:
        protocols:
          http:
            endpoint: 0.0.0.0:4318
```

tools/deployment/package-helm/templates/otel-collector-configmap.yaml
```diff
@@ -0,0 +1,29 @@
...
+    exporters:
+      otlphttp:
+        endpoint: {{ .Values.clpConfig.telemetry.endpoint | quote }}
+
+    service:
+      pipelines:
+        metrics:
+          receivers: [otlp]
+          processors: [batch]
+          exporters: [otlphttp]
```
junhaoliao: the Docker Compose config includes `traces` and `logs` pipelines plus `service.extensions: ["health_check"]`, but the Helm config only has `metrics`. this means traces and logs are silently dropped if any service emits them, and the health_check extension (if added) won't be activated.

```suggestion
    service:
      extensions: ["health_check"]
      pipelines:
        metrics:
          receivers: [otlp]
          processors: [batch]
          exporters: [otlphttp]
        traces:
          receivers: [otlp]
          processors: [batch]
          exporters: [otlphttp]
        logs:
          receivers: [otlp]
          processors: [batch]
          exporters: [otlphttp]
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1014,6 +1019,30 @@ def set_up_env(self) -> None:
...
+                version_file_path.read_text().strip() if version_file_path.exists() else "unknown"
+            )
+
+            self._resource_attrs = {
+                "clp.deployment.id": self._instance_id,
+                "service.version": self._clp_version,
+                "clp.deployment.method": "docker-compose",
+                "clp.storage.engine": self._clp_config.package.storage_engine,
+                "os.type": platform.system().lower(),
+                "host.arch": platform.machine().lower(),
```
junhaoliao: these can be gotten rid of, if https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor is used

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1014,6 +1019,30 @@ def set_up_env(self) -> None:
...
+            }
+
+            resource_attrs_str = ",".join(f"{k}={v}" for k, v in self._resource_attrs.items())
+            env_vars["OTEL_RESOURCE_ATTRIBUTES"] = resource_attrs_str
+            env_vars["OTEL_EXPORTER_OTLP_ENDPOINT"] = "http://otel-collector:4318"
+
+        env_vars["TELEMETRY_ENDPOINT"] = self._clp_config.telemetry.endpoint
```
junhaoliao: how about `CLP_TELEMETRY_ENDPOINT`

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1086,12 +1115,16 @@ def start(self) -> None:
 
         cmd = ["docker", "compose", "--project-name", self._project_name]
         cmd += ["--file", self._get_docker_file_name()]
+        if not self._clp_config.telemetry.disable:
```
junhaoliao: let's stick with the replica count approach

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1014,6 +1019,30 @@ def set_up_env(self) -> None:
...
+                "os.type": platform.system().lower(),
+                "host.arch": platform.machine().lower(),
+            }
+
+            resource_attrs_str = ",".join(f"{k}={v}" for k, v in self._resource_attrs.items())
+            env_vars["OTEL_RESOURCE_ATTRIBUTES"] = resource_attrs_str
+            env_vars["OTEL_EXPORTER_OTLP_ENDPOINT"] = "http://otel-collector:4318"
+
+        env_vars["TELEMETRY_ENDPOINT"] = self._clp_config.telemetry.endpoint
+
```
junhaoliao, line 1139:
```suggestion
        env_vars["CLP_OTEL_COLLECTOR_CONF_FILE_HOST"] = str(
            self._conf_dir / "otel-collector" / "config.yaml"
        )
```

instead of hardcoding, let's consider changing the code in docker-compose-all.yaml to 
```
      - type: "bind"
        source: "${CLP_OTEL_COLLECTOR_CONF_FILE_HOST:-./etc/otel-collector/config.yaml}"
        target: "/etc/otelcol/config.yaml"
        read_only: true
```

tools/deployment/package/docker-compose.dev.yaml
junhaoliao, line 1: i think we can get rid of this and always expose the otel collector port on the host

tools/deployment/package/docker-compose-all.yaml
junhaoliao: also, maybe we dont need the contrib image? this might be sufficient https://hub.docker.com/r/otel/opentelemetry-collector

but let's make sure we can use this non contrib version with the [Health Check Extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.152.0/extension/healthcheckextension) 

---

update: in fact, if we want to use https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/README.md , we should use the contrib version

components/package-template/src/etc/otel-collector/config.yaml
```diff
@@ -0,0 +1,34 @@
...
+  health_check:
+    endpoint: "0.0.0.0:13133"
+
+receivers:
+  otlp:
+    protocols:
+      http:
+        endpoint: "0.0.0.0:4318"
+
+processors:
```
junhaoliao, line 11: let's consider adding
```
  resourcedetection:
    detectors: ["env", "docker", "system"]
    system:
      hostname_sources: ["os"]
      resource_attributes:
        host.arch:
          enabled: true
        host.cpu.family:
          enabled: true
        host.cpu.model.id:
          enabled: true
        host.cpu.vendor.id:
          enabled: true
    override: false
```

then we don't have to manually populate certain info


see https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/README.md#docker-metadata - we might need to configure docker permission to get things actually working. e.g., in docker-compose.yaml
```
  otel-collector:
    ...
    group_add:
      - "${CLP_DOCKER_GID:-999}"
    volumes:
      - type: "bind"
        source: "/var/run/docker.sock"
        target: "/var/run/docker.sock"
```

tools/deployment/package/otel-collector-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+        endpoint: "0.0.0.0:4318"
+
+processors:
+  batch:
+    timeout: "30s"
+    send_batch_size: 1024
+
+exporters:
+  otlphttp:
+    endpoint: "${TELEMETRY_ENDPOINT}"
```
junhaoliao: i can be wrong but the docs say we should use `env:` before the env var name.


```suggestion
    endpoint: "${env:TELEMETRY_ENDPOINT}"
```

if you agree with the `CLP_` prefix, then this should be 

```suggestion
    endpoint: "${env:CLP_TELEMETRY_ENDPOINT}"
```

components/package-template/src/etc/otel-collector/config.yaml
```diff
@@ -0,0 +1,34 @@
...
+  batch:
+    timeout: "30s"
+    send_batch_size: 1024
+
+exporters:
+  otlphttp:
+    endpoint: "${TELEMETRY_ENDPOINT}"
+
+service:
```
junhaoliao, line 41: if we add resourcedetection

```
service:
  extensions: ["health_check"]
  pipelines:
    metrics:
      receivers: ["otlp"]
      processors: ["resourcedetection", "batch"]
      exporters: ["otlphttp"]
    traces:
      receivers: ["otlp"]
      processors: ["resourcedetection", "batch"]
      exporters: ["otlphttp"]
    logs:
      receivers: ["otlp"]
      processors: ["resourcedetection", "batch"]
      exporters: ["otlphttp"]
```

tools/deployment/package-helm/templates/otel-collector-configmap.yaml
junhaoliao, line 1: can we move this to configmap.yaml ? 

e.g., adding this towards the end of configmap.yaml should be sufficient:
```
{{- if and .Values.clpConfig.telemetry (not .Values.clpConfig.telemetry.disable) }}
  otel-collector-config.yaml: |
    extensions:
      health_check:
        endpoint: "0.0.0.0:13133"

    receivers:
      otlp:
        protocols:
          http:
            endpoint: "0.0.0.0:4318"

    processors:
      # NOTE: For EKS deployments, add the "eks" detector to detect cloud-specific attributes (e.g.,
      #cloud.provider, cloud.platform, host.id). This requires additional IAM permissions and K8s
      # downward API config.
      resourcedetection:
        detectors: ["env", "system"]
        system:
          hostname_sources: ["os"]
          resource_attributes:
            host.arch:
              enabled: true
            host.cpu.family:
              enabled: true
            host.cpu.model.id:
              enabled: true
            host.cpu.vendor.id:
              enabled: true
        override: false
      batch:
        timeout: "30s"
        send_batch_size: 1024

    exporters:
      otlphttp:
        endpoint: {{ .Values.clpConfig.telemetry.endpoint | quote }}

    service:
      extensions: ["health_check"]
      pipelines:
        metrics:
          receivers: ["otlp"]
          processors: ["resourcedetection", "batch"]
          exporters: ["otlphttp"]
        traces:
          receivers: ["otlp"]
          processors: ["resourcedetection", "batch"]
          exporters: ["otlphttp"]
        logs:
          receivers: ["otlp"]
          processors: ["resourcedetection", "batch"]
          exporters: ["otlphttp"]
{{- end }}
```

components/package-template/src/etc/otel-collector/config.yaml
junhaoliao, line 1: can we move this guy to  `components/package-template/src/etc/otel-collector/config.yaml`

tools/deployment/package/docker-compose.dev.yaml
junhaoliao: e.g., add this to the service in docker-compose-all.yaml

```
      - host_ip: "127.0.0.1"
        published: "${CLP_OTEL_COLLECTOR_PORT:-4318}"
        target: 4318
```

tools/deployment/package-helm/templates/otel-collector-service.yaml
```diff
@@ -0,0 +1,19 @@
...
+  ports:
+    - name: "otlp-http"
+      port: 4318
+      protocol: TCP
+      targetPort: 4318
+  selector:
+    {{- include "clp.selectorLabels" . | nindent 4 }}
+    app.kubernetes.io/component: "otel-collector"
+  type: ClusterIP
+{{- end }}
```
junhaoliao:
```suggestion
{{- end }}

```

tools/deployment/package-helm/templates/otel-collector-deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+          resources:
+            {{- toYaml .Values.clpConfig.telemetry.collector.resources | nindent 12 }}
+          volumeMounts:
+            - name: otel-collector-config
+              mountPath: /etc/otelcol-contrib
+      volumes:
+        - name: otel-collector-config
+          configMap:
+            name: {{ include "clp.fullname" . }}-otel-collector-config
+{{- end }}
```
junhaoliao:
```suggestion
{{- end }}

```

tools/deployment/package-helm/templates/otel-collector-configmap.yaml
```diff
@@ -0,0 +1,29 @@
...
+{{- end }}
```
junhaoliao:
```suggestion
{{- end }}

```

tools/deployment/package-helm/templates/otel-collector-deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+      app.kubernetes.io/component: "otel-collector"
+  template:
+    metadata:
+      labels:
+        {{- include "clp.labels" . | nindent 8 }}
+        app.kubernetes.io/component: "otel-collector"
+    spec:
+      containers:
+        - name: otel-collector
+          image: {{ .Values.clpConfig.telemetry.collector.image | quote }}
```
junhaoliao: let's split and move the config key to .Values.image.otelCollector.repository and .Values.image.otelCollector.tag
```
          image: {{ printf "%s:%s" .Values.image.otelCollector.repository .Values.image.otelCollector.tag | quote }}
```

tools/deployment/package-helm/values.yaml
```diff
@@ -246,6 +246,20 @@ clpConfig:
...
+    endpoint: "http://localhost:4318"
+    collector:
+      image: "ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib:0.119.0" # yamllint disable-line
+      resources:
+        requests:
+          cpu: "100m"
+          memory: "128Mi"
+        limits:
+          cpu: "500m"
+          memory: "256Mi"
```
junhaoliao: very good mindset but we should align with the config interface in #2048 

cc @hoophalab to comment what's the decided config intf

tools/deployment/package-helm/templates/otel-collector-deployment.yaml
junhaoliao: let's add pull policy too. 
```
          imagePullPolicy: {{ .Values.image.otelCollector.pullPolicy | quote }}

```

see the proposal in values.yaml for details

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -964,8 +966,11 @@ def __init__(
         self, clp_config: ClpConfig, instance_id: str, restart_policy: str = "on-failure:3"
     ) -> None:
         """Initializes the DockerComposeController."""
+        self._instance_id = instance_id
         self._project_name = f"clp-package-{instance_id}"
         self._restart_policy = restart_policy
+        self._clp_version: str | None = None
+        self._resource_attrs: dict[str, Any] = {}
```
junhaoliao: possible to avoid storing `self._instance_id` and `self._clp_version`?

e.g.,
```
self._resource_attrs: dict[str, Any] = {
                "clp.deployment.id": instance_id,
                "service.version": clp_version,
}
```

tools/deployment/package-helm/templates/otel-collector-deployment.yaml
```diff
@@ -0,0 +1,39 @@
+{{- if and .Values.clpConfig.telemetry (not .Values.clpConfig.telemetry.disable) }}
+apiVersion: apps/v1
```
junhaoliao:
```suggestion
apiVersion: "apps/v1"
```

tools/deployment/package-helm/templates/otel-collector-deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+kind: Deployment
```
junhaoliao:
```suggestion
kind: "Deployment"
```

tools/deployment/package-helm/templates/otel-collector-deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+    matchLabels:
+      {{- include "clp.selectorLabels" . | nindent 6 }}
+      app.kubernetes.io/component: "otel-collector"
+  template:
+    metadata:
+      labels:
+        {{- include "clp.labels" . | nindent 8 }}
+        app.kubernetes.io/component: "otel-collector"
+    spec:
+      containers:
```
junhaoliao, line 31: let's refer to the other deployment.yaml and make sure the props (name, image, imagePullPolicy, env, command, ports, volumeMounts) ordering follow theirs

junhaoliao (changes requested):
the implementation is very solid, especially given the tight timeline to put this up. let's see the suggestions to see if we can make this even better

tools/deployment/package-helm/values.yaml
Nathan903: do you think we should just remove it? or wait for a decision on the config interface?

components/clp-package-utils/clp_package_utils/controller.py
Nathan903: removed self._clp_version

however, i i think self._instance_id is needed because is a parameter of __init__, not of set_up_env

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -4,9 +4,11 @@
 import multiprocessing
 import os
 import pathlib
+
```
junhaoliao: if the empty line is not intentional
```suggestion
```

tools/deployment/package-helm/values.yaml
junhaoliao: it seems like #2048 might need some extra time to be merged. let's remove the resources limit in this PR 2278 first. #2048 should later be responsible for recommending the resource limits for the otel collector

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1014,6 +1019,33 @@ def set_up_env(self) -> None:
...
+            self._resource_attrs = {
+                "clp.deployment.id": self._instance_id,
+                "service.version": clp_version,
+                "clp.deployment.method": "docker-compose",
+                "clp.storage.engine": self._clp_config.package.storage_engine,
+            }
+
+            resource_attrs_str = ",".join(f"{k}={v}" for k, v in self._resource_attrs.items())
+            env_vars["OTEL_RESOURCE_ATTRIBUTES"] = resource_attrs_str
+            env_vars["OTEL_EXPORTER_OTLP_ENDPOINT"] = "http://otel-collector:4318"
```
junhaoliao: since the value of `OTEL_EXPORTER_OTLP_ENDPOINT` never changes (because services within the docker compose network can always just refer to other services using service names as the host name),

do you see that we can remove this env var
```suggestion
```

and hardcode the value in docker compose? e.g, in docker-compose-all.yaml, change all related references to:
```
OTEL_EXPORTER_OTLP_ENDPOINT: "http://otel-collector:4318"
```

junhaoliao: since the value of `OTEL_EXPORTER_OTLP_ENDPOINT` never changes (because services within the docker compose network can always just refer to other services using service names as the host name),

do you see that we can remove this env var
```suggestion
```

and hardcode the value in docker compose? e.g, in docker-compose-all.yaml, change all related references to:
```
OTEL_EXPORTER_OTLP_ENDPOINT: "http://otel-collector:4318"
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1014,6 +1019,33 @@ def set_up_env(self) -> None:
             "CLP_PACKAGE_STORAGE_ENGINE": self._clp_config.package.storage_engine,
         }
 
+        # Telemetry
```
junhaoliao, line 1137: for maintainability, can we extract a `_set_up_env_for_telemetry()` method. then below we can call
```py
env_vars |= self._set_up_env_for_telemetry()
```
you can refer to `_set_up_env_for_garbage_collector()` for the docstring, structure and logging. (note that with a dedicated method extraction, we can do early-return in the falsy if-branch, reducing the truthy code's indent by 1 level.)

tools/deployment/package-helm/templates/otel-collector-deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+        - name: otel-collector
```
junhaoliao:
```suggestion
        - name: "otel-collector"
```

tools/deployment/package-helm/templates/otel-collector-deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+          image: {{ printf "%s:%s" .Values.image.otelCollector.repository .Values.image.otelCollector.tag | quote }}
+          imagePullPolicy: {{ .Values.image.otelCollector.pullPolicy | quote }}
+          ports:
+            - containerPort: 4318
+              protocol: TCP
```
junhaoliao: 1. let's use named ports - then below health check can use the name instead of a magic number
2. "TCP" is the default so we can omit

```suggestion
            - name: "otel-collector"
              containerPort: 4318
```

tools/deployment/package-helm/templates/otel-collector-deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+          volumeMounts:
+            - name: otel-collector-config
+              mountPath: /etc/otelcol-contrib
+          args:
+            - "--config=/etc/otelcol-contrib/config.yaml"
+          readinessProbe:
+            {{- include "clp.readinessProbeTimings" . | nindent 12 }}
+            httpGet: &otel-health-check
+              path: "/"
+              port: 13133
```
junhaoliao:
```suggestion
              port: "otel-collector"
```

tools/deployment/package-helm/templates/otel-collector-service.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  name: {{ include "clp.fullname" . }}-otel-collector
+  labels:
+    {{- include "clp.labels" . | nindent 4 }}
+    app.kubernetes.io/component: "otel-collector"
+spec:
+  ports:
+    - name: "otlp-http"
+      port: 4318
+      protocol: TCP
```
junhaoliao:
```suggestion
```

tools/deployment/package-helm/templates/otel-collector-service.yaml
```diff
@@ -0,0 +1,19 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "clp.fullname" . }}-otel-collector
+  labels:
+    {{- include "clp.labels" . | nindent 4 }}
+    app.kubernetes.io/component: "otel-collector"
+spec:
+  ports:
+    - name: "otlp-http"
```
junhaoliao: though not used (by any k8s gateway / ingress) at the moment, let's be consistent with the proposed named port's name above
```suggestion
    - name: "otel-collector"
```

tools/deployment/package-helm/templates/otel-collector-service.yaml
```diff
@@ -0,0 +1,19 @@
...
+      port: 4318
+      protocol: TCP
+      targetPort: 4318
```
junhaoliao: use the named port proposed above
```suggestion
      targetPort: "otel-collector"
```

tools/deployment/package-helm/templates/otel-collector-deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+  selector:
+    matchLabels:
+      {{- include "clp.selectorLabels" . | nindent 6 }}
+      app.kubernetes.io/component: "otel-collector"
+  template:
+    metadata:
+      labels:
+        {{- include "clp.labels" . | nindent 8 }}
+        app.kubernetes.io/component: "otel-collector"
+    spec:
```
junhaoliao, line 26:
```suggestion
    spec:
      {{- include "clp.createSchedulingConfigs" (dict
          "root" .
          "component" "otelCollector"
        ) | nindent 6 }}
```

junhaoliao: then also update values.yaml 's `scheduling` key to include `"otelCollector": {}`

tools/deployment/package-helm/templates/otel-collector-service.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if and .Values.clpConfig.telemetry (not .Values.clpConfig.telemetry.disable) }}
+apiVersion: v1
+kind: Service
```
junhaoliao:
```suggestion
apiVersion: "v1"
kind: "Service"
```

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -655,3 +711,28 @@ services:
...
+    <<: *service_defaults
+    image: "ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib:0.152.0" # yamllint disable-line
+    deploy:
+      # Value must be either 0 or 1. Set to 0 to disable the otel collector.
+      replicas: "${CLP_OTEL_COLLECTOR_ENABLED:-1}"
+    user: "${CLP_THIRD_PARTY_SERVICE_UID_GID:-1000:1000}"
+    group_add:
+      - "${CLP_DOCKER_GID:-999}"
+    environment:
+      CLP_TELEMETRY_ENDPOINT: "${CLP_TELEMETRY_ENDPOINT:-https://telemetry.yscope.io}"
```
junhaoliao, line 682: (i haven't read their docs very carefully so please help confirm)

maybe we need to add `OTEL_RESOURCE_ATTRIBUTES` here too, so the resourcedetection 's "env" detector can actually read the attributes then merge with the other service's emitted metrics? that said, since other services use the same OTEL_RESOURCE_ATTRIBUTES, appending the same attributes again at the collector level (with "override": false) only add very slight robusty (in the case that individual services don't include the same OTEL_RESOURCE_ATTRIBUTES for whatever reason), but incurs no actual effect? in that case, would it be better to just remove the `env` detector in the config?

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -103,6 +103,10 @@ services:
       PYTHONPATH: "/opt/clp/lib/python3/site-packages"
       SPIDER_DB_PASS: "${SPIDER_DB_PASS:?Please set a value.}"
       SPIDER_DB_USER: "${SPIDER_DB_USER:-spider-user}"
+      OTEL_EXPORTER_OTLP_ENDPOINT: "${OTEL_EXPORTER_OTLP_ENDPOINT:-}"
```
junhaoliao: sorry for forgetting to post this in the last round of review

can we extract a dummy object at the top of the file, between `x-healthcheck-defaults` and `x-volume-definitions`:
```yaml
# Common telemetry environment defaults.
x-clp-telemetry-env-defaults: &clp_telemetry_env_defaults
  CLP_DISABLE_TELEMETRY: "${CLP_DISABLE_TELEMETRY:-false}"
  OTEL_EXPORTER_OTLP_ENDPOINT: "${OTEL_EXPORTER_OTLP_ENDPOINT:-http://otel-collector:4318}"
  OTEL_RESOURCE_ATTRIBUTES: "${OTEL_RESOURCE_ATTRIBUTES:-}"
```
(note i also propose default value fallbacks)


then here and in other services we can write
```yaml
    environment:
      <<: *clp_telemetry_env_defaults
      CLP_DB_PASS: "${CLP_DB_PASS:?Please set a value.}"
...
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1014,6 +1019,33 @@ def set_up_env(self) -> None:
...
+            }
+
+            resource_attrs_str = ",".join(f"{k}={v}" for k, v in self._resource_attrs.items())
+            env_vars["OTEL_RESOURCE_ATTRIBUTES"] = resource_attrs_str
+            env_vars["OTEL_EXPORTER_OTLP_ENDPOINT"] = "http://otel-collector:4318"
+
+        env_vars["CLP_TELEMETRY_ENDPOINT"] = self._clp_config.telemetry.endpoint
+        env_vars["CLP_OTEL_COLLECTOR_CONF_FILE_HOST"] = str(
+            self._conf_dir / "otel-collector" / "config.yaml"
+        )
```
junhaoliao: if i understand correctly, these two should be included in the above `else:` branch?

junhaoliao (changes requested):
very good. here are the very last batch of suggestions

for the title, how about simply:
```
feat(deployment): Add OpenTelemetry Collector.
```

tools/deployment/package/docker-compose-all.yaml
Nathan903: yes, the env is extra, ive removed them

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -655,3 +689,28 @@ services:
...
+        published: "${CLP_OTEL_COLLECTOR_PORT:-14318}"
+        target: 4318
+    volumes:
+      - type: "bind"
+        source: "${CLP_OTEL_COLLECTOR_CONF_FILE_HOST:-./etc/otel-collector/config.yaml}"
+        target: "/etc/otelcol-contrib/config.yaml"
+        read_only: true
+      - type: "bind"
+        source: "/var/run/docker.sock"
+        target: "/var/run/docker.sock"
```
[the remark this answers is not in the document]
Nathan903: removing this access would mean needing env vars to pass in the system info. Not sure if the security benefit is worth the complexity of changing. @junhaoliao can you please advise

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -50,6 +50,67 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
...
+{{- if and .Values.clpConfig.telemetry .Values.clpConfig.telemetry.disable }}
+- name: CLP_DISABLE_TELEMETRY
+  value: "true"
+{{- else }}
+- name: OTEL_EXPORTER_OTLP_ENDPOINT
+  value: "http://{{ include "clp.fullname" . }}-otel-collector:4318"
+- name: OTEL_RESOURCE_ATTRIBUTES
+  value: {{ include "clp.resourceAttributes" . | quote }}
+{{- end }}
+{{- end -}}
```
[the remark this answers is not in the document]
Nathan903: @junhaoliao  we dont want to include anything not yet used right?

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -50,6 +50,67 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
...
+  {{- $_ := set .Values "global" dict -}}
+{{- end }}
+{{- if not .Values.global.instanceId }}
+  {{- $existing := lookup "v1" "ConfigMap" .Release.Namespace (printf "%s-instance-id" (include "clp.fullname" .)) }}
+  {{- if $existing }}
+    {{- $_ := set .Values.global "instanceId" (index $existing.data "instanceId") -}}
+  {{- else }}
+    {{- $_ := set .Values.global "instanceId" uuidv4 -}}
+  {{- end }}
+{{- end }}
```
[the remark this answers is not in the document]
Nathan903: ignored. If configmap exist. the key is generated. The issue wont happen unless somebody manually deletes the uuid from config

components/package-template/src/etc/otel-collector/config.yaml
```diff
@@ -0,0 +1,48 @@
...
+        endpoint: "0.0.0.0:4318"
+
+processors:
+  batch:
+    timeout: "30s"
+    send_batch_size: 1024
+  resourcedetection:
+    detectors: ["docker", "system"]
+    system:
+      hostname_sources: ["os"]
```
junhaoliao: it appears that `host.name` detection is enabled by default: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/internal/system/documentation.md

as promised in the docs, we will not collect hostnames. can we remove this `hostname_sources` and set 

```yaml
    docker:
      resource_attributes:
        host.name:
          enabled: false
    system:
      resource_attributes:
        ...
        host.name:
          enabled: false
```

tools/deployment/package-helm/templates/compression-scheduler-deployment.yaml
junhaoliao, line 1: it seems like we are setting those telemetry env vars only for certain services in the Helm chart, while we set them for all docker compose services? What's the reason of the disparity? Also, I don't see any services consuming any of these env vars as of this PR. Would it be cleaner to add those later when we instrument different services later?

tools/deployment/package-helm/templates/topology-metrics-emitter-job.yaml
```diff
@@ -0,0 +1,47 @@
...
+      restartPolicy: "OnFailure"
+      initContainers:
+        - {{- include "clp.waitFor" (dict
+            "root" .
+            "type" "service"
+            "name" "otel-collector"
+          ) | nindent 10 }}
+      containers:
+        - name: "topology-metrics-emitter"
+          image: "curlimages/curl:8.14.1"
```
junhaoliao: can we move this under `values.yaml` like `image.clpPackage`, `image.kubectl`, and
`image.otelCollector`? also shall we make the pullPolicy configurable as well

e.g.,

.Values.image.curl.*

tools/deployment/package-helm/templates/configmap.yaml
```diff
@@ -435,3 +435,58 @@ data:
...
+
+    receivers:
+      otlp:
+        protocols:
+          http:
+            endpoint: "0.0.0.0:4318"
+
+    processors:
+      # NOTE: For EKS deployments, add the "eks" detector to detect cloud-specific attributes (e.g.,
+      #cloud.provider, cloud.platform, host.id). This requires additional IAM permissions and K8s
```
junhaoliao:
```suggestion
      # cloud.provider, cloud.platform, host.id). This requires additional IAM permissions and K8s
```

tools/deployment/package-helm/templates/_helpers.tpl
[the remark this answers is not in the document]
junhaoliao: Right. i hadn't seen this comment when i posted https://github.com/y-scope/clp/pull/2278/changes#r3318386877

Let's defer the addition of those env vars

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -166,6 +170,47 @@ def get_clp_home():
...
+    :raise ValueError: If max_attempts < 1.
+    :raise urllib.error.URLError: If all attempts fail due to a network-level error.
+    :raise urllib.error.HTTPError: If all attempts fail with an HTTP error status.
+    :raise OSError: If all attempts fail due to a socket-level error.
+    """
+    if max_attempts < 1:
+        raise ValueError(f"max_attempts must be >= 1, got {max_attempts}")
+
+    req = urllib.request.Request(url, data=data, headers=headers or {}, method=method)
+    last_exception: urllib.error.URLError | urllib.error.HTTPError | OSError = URLError(
```
junhaoliao: otherwise we see
```
2026-05-28T14:40:01.544 WARNING [controller] Failed to emit topology metrics: name 'URLError' is not defined
```

```suggestion
    last_exception: urllib.error.URLError | urllib.error.HTTPError | OSError = urllib.error.URLError(
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -166,6 +170,47 @@ def get_clp_home():
...
+def http_request(
+    url: str,
+    data: bytes | None = None,
+    headers: dict[str, str] | None = None,
+    method: str = "GET",
+    max_attempts: int = 2,
+    retry_delay: float = 1.0,
+    timeout: float = 5.0,
+) -> http.client.HTTPResponse:
+    """Sends an HTTP request with retry logic.
```
junhaoliao: need to run `task lint:fix-py` to auto fix various format issues
```suggestion
    """
    Sends an HTTP request with retry logic.
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -83,6 +85,8 @@
 THIRD_PARTY_SERVICE_GID = 999
 THIRD_PARTY_SERVICE_UID_GID = f"{THIRD_PARTY_SERVICE_UID}:{THIRD_PARTY_SERVICE_GID}"
 
+OTEL_COLLECTOR_HOST_PORT = 14318
```
junhaoliao: any reason not to use the same 4318 port on the host?

junhaoliao (changes requested):

components/clp-package-utils/clp_package_utils/controller.py
Nathan903: after we changed the _emit_topology_metrics function from using a docker container to send telemetry to using python's http_request, this script now opens the `OTEL_COLLECTOR_HOST_PORT` on the user's device (not in containers). So when testing my code, this causes port conflict with the clp side otel collector from talking to the server collector (otel default is 4318). In production this will not be an issue since clp side otel collector talks to the server collector by port 443. However, if users sets their own endpoint, the default would also be 4318, so if we use 4318 for the code above, users will not be able to just enter their own endpoint. Also vice versa: if users already have otel or something listening to 4318 on their device, the function would fail (causing telemetry to not start)

tools/deployment/package-helm/templates/compression-scheduler-deployment.yaml
Nathan903: would other people later add instrumentation code themselves? if so, it might be easier to keep the env vars (and add them for all, as well as for docker compose) so people can just change the python/rust code to add instrumentation. Or should I remove any unused ones for now?

components/clp-package-utils/clp_package_utils/controller.py
Nathan903: actually im considering reverting back this commit https://github.com/y-scope/clp/pull/2278/changes/c7bc4d5ea8c810149d9283a5747236f58fc20ed5

the old way is running `docker run curlimages/curl`, making a new container to make the HTTP POST. This is slow but avoids the need of opening any port. The current way uses python http_request but requires opening some port (currently 14318)

components/package-template/src/etc/otel-collector/config.yaml
```diff
@@ -0,0 +1,55 @@
...
+        host.cpu.model.id:
+          enabled: true
+        host.cpu.vendor.id:
+          enabled: true
+        host.name:
+          enabled: false
+    override: false
+
+exporters:
+  otlphttp:
```
junhaoliao: when `exporters.debug.verbosity: detailed` is set, this was found in the otel collector logs:
```
2026-05-29T02:04:27.921Z	warn	builders/builders.go:40	"otlphttp" alias is deprecated; use "otlp_http" instead	{"resource": {"service.instance.id": "34a0832b-7f88-4ca2-85ef-0b92b2f02d3c", "service.name": "otelcol-contrib", "service.version": "0.152.0"}, "otelcol.component.id": "otlphttp", "otelcol.component.kind": "exporter", "otelcol.signal": "metrics"}
2026-05-29T02:04:27.921596156Z 2026-05-29T02:04:27.921Z	warn	builders/builders.go:40	"otlphttp" alias is deprecated; use "otlp_http" instead	{"resource": {"service.instance.id": "34a0832b-7f88-4ca2-85ef-0b92b2f02d3c", "service.name": "otelcol-contrib", "service.version": "0.152.0"}, "otelcol.component.id": "otlphttp", "otelcol.component.kind": "exporter", "otelcol.signal": "logs"}
2026-05-29T02:04:27.921768599Z 2026-05-29T02:04:27.921Z	warn	builders/builders.go:40	"otlphttp" alias is deprecated; use "otlp_http" instead	{"resource": {"service.instance.id": "34a0832b-7f88-4ca2-85ef-0b92b2f02d3c", "service.name": "otelcol-contrib", "service.version": "0.152.0"}, "otelcol.component.id": "otlphttp", "otelcol.component.kind": "exporter", "otelcol.signal": "traces"}
```

according to their logs, it seems that exporter name has been deprecated and has come an alias

https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/otlphttpexporter#:~:text=The%20otlphttp%20deprecated%20alias%20exists%20for%20the%20component%20name.%20It%20will%20be%20removed%20in%20a%20future%20version.%20If%20you%20use%20the%20deprecated%20alias%20otlphttp%20in%20your%20configuration%2C%20change%20it%20to%20otlp_http.

> The otlphttp deprecated alias exists for the component name. It will be removed in a future version. If you use the deprecated alias otlphttp in your configuration, change it to otlp_http.

tools/deployment/package-helm/templates/topology-metrics-emitter-job.yaml
```diff
@@ -0,0 +1,50 @@
...
+apiVersion: "batch/v1"
+kind: "Job"
+metadata:
+  name: {{ include "clp.fullname" . }}-topology-metrics-emitter
+  labels:
+    {{- include "clp.labels" . | nindent 4 }}
+    app.kubernetes.io/component: "topology-metrics-emitter"
+  annotations:
+    "helm.sh/hook": "post-install,post-upgrade"
+    "helm.sh/hook-weight": "0"
```
junhaoliao: hook-weight 0 seems already the default and probably isn't needed?
```suggestion
```

tools/deployment/package-helm/templates/topology-metrics-emitter-job.yaml
```diff
@@ -0,0 +1,50 @@
...
+    "helm.sh/hook-delete-policy": "hook-succeeded"
```
junhaoliao: would it be better to just use the default "before-hook-creation" so that the job doesn't surprisingly got deleted after it's run?

```suggestion
```

junhaoliao (changes requested):

tools/deployment/package-helm/templates/topology-metrics-emitter-job.yaml
Nathan903: true, ive removed this

tools/deployment/package-helm/templates/topology-metrics-emitter-job.yaml
Nathan903: i think with hook-succeeded the job only gets deleted after it completes (which I assume is what we want?) or do we want to keep the process around after it runs? The only reason i can think of why we might need this is for looking at debug logs

components/package-template/src/etc/otel-collector/config.yaml
Nathan903: thanks for the catch, yes this fix looks correct

Nathan903:
ive reviewed the new changes and everything makes sense. also rerun the tests and everything works

tools/deployment/package-helm/templates/topology-metrics-emitter-job.yaml
```diff
@@ -0,0 +1,52 @@
...
+apiVersion: "batch/v1"
+kind: "Job"
+metadata:
+  name: {{ include "clp.fullname" . }}-topology-metrics-emitter
+  labels:
+    {{- include "clp.labels" . | nindent 4 }}
+    app.kubernetes.io/component: "topology-metrics-emitter"
+  annotations:
+    "helm.sh/hook": "post-install,post-upgrade"
+    "helm.sh/hook-delete-policy": "before-hook-creation"
```
junhaoliao:
```suggestion
```

was discuss this offline with @Nathan903 

> Nathan:
> Also, I tried deleting this line and the default behavior is actually no deletion, not before-hook-creation, so I changed the line to before-hook-creation

i believe "no deletion" after `helm install` is expected? [`before-hook-creation` is already the behaviour when the line isn't specified](https://helm.sh/docs/topics/charts_hooks/#:~:text=Delete%20the%20previous%20resource%20before%20a%20new%20hook%20is%20launched%20(default)). If i understand correctly, it means that:
1. there is no deletion of the job after it finishes.
2. when the hook is triggered again (on another `helm install` / `helm upgrade` of the same chart), the **previous** job is deleted, and a new job is run as a result of the hook trigger.

i tried removing the line and ran these on my end and the behaviours matched expectations:
```
set-up-test.sh # which sets up a kind cluster and `helm install` the chart

kubectl get jobs # which shows test-clp-topology-metrics-emitter

helm upgrade ... # which reinstall the chart and triggers the hool

kubectl get jobs # which again shows test-clp-topology-metrics-emitter but with a very recent AGE, which suggests it's a new job rather than the previous one

# no conflict of job creation is shown which also means the old job has been properly deleted before creation of the new one
```

Nathan903: I got the error below yesterday when testing. However, I think you're right because i just tested again and even without the line the error does not occur again. Maybe its some other issue or incomplete that caused this.
```console
nathanhung@baker18:/home/nathanhung/clp/tools/deployment/package-helm$ helm uninstall clp

release "clp" uninstalled
nathanhung@baker18:/home/nathanhung/clp/tools/deployment/package-helm$
nathanhung@baker18:/home/nathanhung/clp/tools/deployment/package-helm$ helm install clp .   --set clpConfig.telemetry.endpoint="http://172.17.0.1:4318"   --set clpConfig.logs_input
.directory="/home"   --timeout 10m
Error: INSTALLATION FAILED: failed post-install: warning: Hook post-install clp/templates/topology-metrics-emitter-job.yaml failed: server-side apply failed for object default/clp-
topology-metrics-emitter batch/v1, Kind=Job: Job.batch "clp-topology-metrics-emitter" is invalid: spec.template: Invalid value: {"labels":{"app.kubernetes.io/component":"topology-m
etrics-emitter","app.kubernetes.io/instance":"clp","app.kubernetes.io/managed-by":"Helm","app.kubernetes.io/name":"clp","app.kubernetes.io/version":"0.12.1-dev","batch.kubernetes.i
o/controller-uid":"f5fe65f0-b9a4-4038-b71e-39552f723209","batch.kubernetes.io/job-name":"clp-topology-metrics-emitter","controller-uid":"f5fe65f0-b9a4-4038-b71e-39552f723209","helm
.sh/chart":"clp-0.3.2-dev.8","job-name":"clp-topology-metrics-emitter"},"Spec":{"Volumes":null,"InitContainers":[{"Name":"wait-for-otel-collector","Image":"bitnami/kubectl@sha256:9
8736aabcecb8d3cbcdcd7b132d14b1d67ed99bac2f06d471f06235933103df3","Command":["kubectl","wait","--for=condition=ready","pod","--selector","app.kubernetes.io/component=otel-collector"
,"--timeout=300s"],"Args":null,"WorkingDir":"","Ports":null,"EnvFrom":null,"Env":null,"Resources":{"Limits":null,"Requests":null,"Claims":null},"ResizePolicy":null,"RestartPolicy":
null,"RestartPolicyRules":null,"VolumeMounts":null,"VolumeDevices":null,"LivenessProbe":null,"ReadinessProbe":null,"StartupProbe":null,"Lifecycle":null,"TerminationMessagePath":"/d
ev/termination-log","TerminationMessagePolicy":"File","ImagePullPolicy":"IfNotPresent","SecurityContext":null,"Stdin":false,"StdinOnce":false,"TTY":false}],"Containers":[{"Name":"t
opology-metrics-emitter","Image":"curlimages/curl:8.14.1","Command":["curl","--request","POST","--fail","--header","Content-Type: application/json","--max-time","5","--retry","1","
--retry-all-errors","--retry-delay","1","--show-error","--silent","--data","{\"resourceMetrics\":[{\"resource\":{\"attributes\":[{\"key\":\"service.name\",\"value\":{\"stringValue\
":\"controller\"}}]},\"scopeMetrics\":[{\"scope\":{\"name\":\"clp.controller\"},\"metrics\":[{\"name\":\"clp.deployment.compression_worker_replicas\",\"gauge\":{\"dataPoints\":[{\"
asInt\":\"1\",\"timeUnixNano\":\"1780068817802666866\"}]}},{\"name\":\"clp.deployment.compression_worker_concurrency\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"8\",\"timeUnixNano\":
\"1780068817802666866\"}]}},{\"name\":\"clp.deployment.query_worker_replicas\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"1\",\"timeUnixNano\":\"1780068817802666866\"}]}},{\"name\":\"
clp.deployment.query_worker_concurrency\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"8\",\"timeUnixNano\":\"1780068817802666866\"}]}},{\"name\":\"clp.deployment.reducer_replicas\",\"g
auge\":{\"dataPoints\":[{\"asInt\":\"1\",\"timeUnixNano\":\"1780068817802666866\"}]}},{\"name\":\"clp.deployment.reducer_concurrency\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"8\",\
"timeUnixNano\":\"1780068817802666866\"}]}}]}]}]}","http://clp-otel-collector:4318/v1/metrics"],"Args":null,"WorkingDir":"","Ports":null,"EnvFrom":null,"Env":null,"Resources":{"Lim
its":null,"Requests":null,"Claims":null},"ResizePolicy":null,"RestartPolicy":null,"RestartPolicyRules":null,"VolumeMounts":null,"VolumeDevices":null,"LivenessProbe":null,"Readiness
Probe":null,"StartupProbe":null,"Lifecycle":null,"TerminationMessagePath":"/dev/termination-log","TerminationMessagePolicy":"File","ImagePullPolicy":"IfNotPresent","SecurityContext
":null,"Stdin":false,"StdinOnce":false,"TTY":false}],"EphemeralContainers":null,"RestartPolicy":"OnFailure","TerminationGracePeriodSeconds":30,"ActiveDeadlineSeconds":null,"DNSPoli
cy":"ClusterFirst","NodeSelector":null,"ServiceAccountName":"clp-service-account","AutomountServiceAccountToken":null,"NodeName":"","SecurityContext":{"HostNetwork":false,"HostPID"
:false,"HostIPC":false,"ShareProcessNamespace":null,"HostUsers":null,"SELinuxOptions":null,"WindowsOptions":null,"RunAsUser":null,"RunAsGroup":null,"RunAsNonRoot":null,"Supplementa
lGroups":null,"SupplementalGroupsPolicy":null,"FSGroup":null,"FSGroupChangePolicy":null,"Sysctls":null,"SeccompProfile":null,"AppArmorProfile":null,"SELinuxChangePolicy":null},"Ima
gePullSecrets":null,"Hostname":"","Subdomain":"","SetHostnameAsFQDN":null,"Affinity":null,"SchedulerName":"default-scheduler","Tolerations":null,"HostAliases":null,"PriorityClassNa
me":"","Priority":null,"PreemptionPolicy":null,"DNSConfig":null,"ReadinessGates":null,"RuntimeClassName":null,"Overhead":null,"EnableServiceLinks":null,"TopologySpreadConstraints":
null,"OS":null,"SchedulingGates":null,"ResourceClaims":null,"Resources":null,"HostnameOverride":null,"WorkloadRef":null}}: field is immutable

```

junhaoliao: interesting... i guess the hook delete policy may have been in flux between hook-succeeded, removed, and before-hook-creation while you made the changes. 

anyways, it's good to know that without the explicit config things are now working as expected on your end

junhaoliao (approved):

junhaoliao:
it seems we forgot to update the integration test suite https://github.com/y-scope/clp/blob/120f63022c805e0c88eabf00d4f2cc2d31fe0aa3/integration-tests/tests/package_tests/utils/modes.py#L32 and now the integration tests are failing

```
E   AssertionError: 'clp-json' package start up verification failure: components could not be validated. Unexpected services: {'otel-collector'}. See subprocess log at: '/home/junhao/workspace/8-clp/build/integration_tests/test_logs/testrun_2026-06-23-13-08-26/subprocess_output/start-clp.sh_2026-06-23-13-08-36-372.log'.
```

PR #2281, opened 2026-05-14
feat(clp-s::search): Add a column-based filtering implementation for a subset of query shapes.

ShangDanLuXian:
# Description

This PR adds an initial column-scan filter execution path for `clp_s` search.

The implementation is intentionally kept small and focused. `QueryRunner` now prepares the schema-table filter and attempts to use `ColumnScan` for supported predicates. If column scan cannot handle the query expression, the existing row-scan path is used as a fallback.

Main changes:
- Add `ColumnScan` under `components/core/src/clp_s/search/`.
- Support building per-schema match bitmaps for supported filter ASTs.
- Support column scan for simple scalar/string filter predicates and `AND` / `OR` expression trees.
- Integrate column scan through `QueryRunner::prepare_filter`.
- Add CMake wiring for the new column-scan source files.

# Checklist

  breaking change.

# Validation performed

- Tested representative MongoDB search queries locally.
- For query `id: 22419`, observed total search time reduced from roughly 30s to roughly 15s on the test dataset, with table scan costs 1-2s.
- Sanity checked all queries for the MongoDB dataset in the paper.

## Summary by CodeRabbit

* **New Features**
  * Adds an optimized column-scan filter that precomputes per-message match bitmaps for search expressions, speeding up message filtering.
  * Exposes a schema-specific filter pathway that can be used automatically when supported.

* **Improvements**
  * Query execution now selects and prepares the schema-specific filter when available.
  * Output handling uses the prepared schema filter for message iteration, simplifying and improving filter application.


components/core/src/clp_s/search/ColumnScan.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+    [[nodiscard]] auto build_node(ast::Expression* expr) const -> Bitmap;
+    [[nodiscard]] auto build_filter(ast::FilterExpr* filter) const -> Bitmap;
+
+    std::shared_ptr<ast::Expression> m_expression;
+    BasicReaderMap const* m_basic_readers;
+    ClpStringReaderMap const* m_clp_string_readers;
+    VarStringReaderMap const* m_var_string_readers;
+    ClpQueryMap const* m_clp_queries;
+    VarMatchMap const* m_var_matches;
```
gibber9809: As far as I can tell, all of these members are only used to avoid passing these around during initialization.

Could we get rid of these members and just pass them around during initialization instead? I know it will make some of these function signatures longer, but I think it will make this class easier to read, since when a reader sees these pointers they have to start wondering about the lifetimes & ownership of these pointers after initialization.

components/core/src/clp_s/search/ColumnScan.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include <clp/Query.hpp>
+#include <clp_s/ColumnReader.hpp>
+#include <clp_s/SchemaReader.hpp>
+#include <clp_s/search/ast/Expression.hpp>
+#include <clp_s/search/ast/FilterExpr.hpp>
+
+namespace clp_s::search {
+class ColumnScan : public FilterClass {
+public:
+    using Bitmap = std::vector<uint8_t>;
```
gibber9809, line 20: I guess we're using this instead of a packed bitmap since there wasn't much of a performance difference during benchmarking? If so, that makes sense for now.

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,426 @@
...
+            return value < operand;
+        case FilterOperation::GT:
+            return value > operand;
+        case FilterOperation::LTE:
+            return value <= operand;
+        case FilterOperation::GTE:
+            return value >= operand;
+        case FilterOperation::EXISTS:
+        case FilterOperation::NEXISTS:
+            return false;
```
gibber9809:
```suggestion
            return true;
```
EXISTS/NEXISTS are considered always matching at scan time because schema matching has previously evaluated whether they match or not.

gibber9809 (changes requested):
Giving an initial set of comments -- sorry have been busy with other reviews. Will aim to give a full review tomorrow.

components/core/src/clp_s/search/ColumnScan.hpp
```diff
@@ -0,0 +1,76 @@
...
+            ClpQueryMap const& clp_queries,
+            VarMatchMap const& var_matches,
+            uint64_t num_messages
+    ) -> std::optional<ColumnScan>;
+
+    ColumnScan(ColumnScan&&) = default;
+    ColumnScan(ColumnScan const&) = delete;
+    auto operator=(ColumnScan&&) -> ColumnScan& = delete;
+    auto operator=(ColumnScan const&) -> ColumnScan& = delete;
+    virtual ~ColumnScan() = default;
```
gibber9809, line 61: nit: should group these together with the standard set of comments for defaulted/deleted constructors

components/core/src/clp_s/search/ColumnScan.hpp
```diff
@@ -0,0 +1,76 @@
...
+public:
+    using Bitmap = std::vector<uint8_t>;
+    using BasicReaderMap = std::unordered_map<int32_t, std::vector<BaseColumnReader*>>;
+    using ClpStringReaderMap = std::unordered_map<int32_t, std::vector<ClpStringColumnReader*>>;
+    using VarStringReaderMap
+            = std::unordered_map<int32_t, std::vector<VariableStringColumnReader*>>;
+    using ClpQueryMap = std::unordered_map<ast::Expression*, clp::Query*>;
+    using VarMatchMap = std::unordered_map<ast::Expression*, std::unordered_set<int64_t>*>;
+
+    [[nodiscard]] static auto try_create(
```
gibber9809, line 40: nit: missing docstring.

components/core/src/clp_s/search/ColumnScan.hpp
```diff
@@ -0,0 +1,76 @@
...
+            ClpQueryMap const& clp_queries,
+            VarMatchMap const& var_matches,
+            uint64_t num_messages
+    );
+
+    [[nodiscard]] auto can_build_node(ast::Expression* expr) const -> bool;
+    [[nodiscard]] auto can_build_filter(ast::FilterExpr* filter) const -> bool;
+
+    [[nodiscard]] auto build_node(ast::Expression* expr) const -> Bitmap;
+    [[nodiscard]] auto build_filter(ast::FilterExpr* filter) const -> Bitmap;
```
gibber9809: It's pretty straightforward to understand what these do in context, but I think it's still worth adding docstrings here.

components/core/src/clp_s/search/QueryRunner.hpp
```diff
@@ -86,6 +88,11 @@ class QueryRunner : public FilterClass {
      */
     auto schema_init(int32_t schema_id) -> EvaluatedValue;
 
+    /**
+     * Initializes readers for the schema table and returns the filter implementation to use.
```
gibber9809:
```suggestion
     * Selects a filtering implementation, and prepares a filter on a given ERT.
     *
     * Note: This method must be called after schema_init.
     *
     * @param reader A reader for an ERT.
     * @return The filtering implementation selected by QueryRunner.
```

gibber9809 (changes requested):
Leaving a few more comments -- only file I still need to review in my first pass after this is `ColumnScan.cpp`.

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,426 @@
...
+using clp_s::search::ast::AndExpr;
+using clp_s::search::ast::FilterExpr;
+using clp_s::search::ast::FilterOperation;
+using clp_s::search::ast::LiteralType;
+using clp_s::search::ast::OrExpr;
+
+namespace clp_s::search {
+namespace {
+template <typename T>
+[[nodiscard]] auto compare(FilterOperation operation, T value, T operand) -> bool {
```
gibber9809, line 119: For cases like this where we declare and define a function in a .cpp file, our guideline is to separate the declaration from the definition, and have all of the declarations (with docstrings) come before the definitions. It makes it easier to understand the code without having to scroll past all of the implementations.

In this instance that applies to this `compare` function, as well as `is_equality_operation`, `invert`, `build_basic_filter`, `clp_string_matches`, `build_clp_string_filter`, and `build_var_string_filter`.

You can take a look at `clp_s/timestamp_parser/TimestampParser.cpp` for a somewhat recent example of this in the codebase.

gibber9809 (changes requested):
Took a look through ColumnScan.cpp as well now -- left some initial comments, but will take a closer look after the first round of review comments have been addressed.

components/core/src/clp_s/search/ColumnScan.hpp
```diff
@@ -0,0 +1,117 @@
...
+    auto init(SchemaReader*, std::vector<BaseColumnReader*> const&) -> void override {}
+
+    [[nodiscard]] auto filter(uint64_t cur_message) -> bool override;
+
+private:
+    explicit ColumnScan(uint64_t num_messages);
+
+    /**
+     * Checks whether an expression tree can be evaluated by ColumnScan.
+     * @return true when every node is either a supported logical expression or a supported filter.
```
gibber9809:
```suggestion
     * Checks whether an expression can be evaluated by ColumnScan.
     * @param expr
     * @param clp_queries
     * @param var_matches
     * @return Whether every node including `expr` is a supported logical expression or a supported filter expression.
```

components/core/src/clp_s/search/ColumnScan.hpp
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * Checks whether a single filter can be evaluated by ColumnScan.
+     * @return true when the filter has a resolved column, a supported operation/literal type, and
+     * any required precomputed string query or variable-match data.
+     */
+    [[nodiscard]] static auto can_build_filter(
+            ast::FilterExpr* filter,
+            ClpQueryMap const& clp_queries,
+            VarMatchMap const& var_matches
+    ) -> bool;
```
gibber9809, line 110:
```suggestion
    /**
     * Checks whether a single filter can be evaluated by ColumnScan.
     * 
     * Supported filters have:
     * - A fully resolved column
     * - A supported operation and literal-type combination
     * - For string predicates, a valid corresponding entry in one of the precomputed string search
     * maps.  
     * 
     * @param filter
     * @param clp_queries A map of precomputed clp string searches.
     * @param var_matches A map of precomputed variable string searches.
     * @return Whether `filter` is a supported filter expression.
     */
    [[nodiscard]] static auto can_build_filter(
            ast::FilterExpr* filter,
            ClpQueryMap const& clp_queries,
            VarMatchMap const& var_matches
    ) -> bool;
```

components/core/src/clp_s/search/ColumnScan.hpp
```diff
@@ -0,0 +1,117 @@
...
+    using VarStringReaderMap
+            = std::unordered_map<int32_t, std::vector<VariableStringColumnReader*>>;
+    using ClpQueryMap = std::unordered_map<ast::Expression*, clp::Query*>;
+    using VarMatchMap = std::unordered_map<ast::Expression*, std::unordered_set<int64_t>*>;
+
+    /**
+     * Attempts to build a column scan for the given expression.
+     * @return A ColumnScan with a precomputed match bitmap when every node in the expression can
+     * be evaluated from the supplied reader/query maps, or std::nullopt otherwise.
+     */
```
gibber9809, line 39:
```suggestion
    /**
     * Attempts to build a column scan for the given expression over the given ERT.
     * @param expression
     * @param basic_readers A map of relevant primitive type readers for the given ERT.
     * @param clp_string_readers A map of relevant clp string readers for the given ERT.
     * @param var_string_readers A map of relevant variable string readers for the given ERT.
     * @param clp_queries A map of precomputed clp string searches.
     * @param var_matches A map of precomputed variable string searches.
     * @param num_messages The number of messages in the given ERT.
     * @return A ColumnScan with a precomputed bitmap recording every matching result on success,
     * or std::nullopt when ColumnScan does not support the query given by `expression`.
     */
```

components/core/src/clp_s/search/ColumnScan.hpp
```diff
@@ -0,0 +1,117 @@
...
+     * any required precomputed string query or variable-match data.
+     */
+    [[nodiscard]] static auto can_build_filter(
+            ast::FilterExpr* filter,
+            ClpQueryMap const& clp_queries,
+            VarMatchMap const& var_matches
+    ) -> bool;
+
+    /**
+     * Builds a bitmap for an expression tree using the supplied reader/query maps.
```
gibber9809, line 113:
```suggestion
     * Builds a bitmap for an expression tree using the supplied reader/query maps.
     * @param expr
     * @param basic_readers
     * @param clp_string_readers
     * @param var_string_readers
     * @param clp_queries
     * @param var_matches
```

components/core/src/clp_s/search/ColumnScan.hpp
```diff
@@ -0,0 +1,117 @@
...
+            ClpStringReaderMap const& clp_string_readers,
+            VarStringReaderMap const& var_string_readers,
+            ClpQueryMap const& clp_queries,
+            VarMatchMap const& var_matches
+    ) const -> Bitmap;
+
+    /**
+     * Builds a bitmap for a single filter using the supplied reader/query maps.
+     * @pre can_build_filter has returned true for filter with the same query and variable-match
+     * maps.
```
gibber9809, line 134:
```suggestion
     * maps.
     * @param filter
     * @param basic_readers
     * @param clp_string_readers
     * @param var_string_readers
     * @param clp_queries
     * @param var_matches
```

components/core/src/clp_s/search/ColumnScan.hpp
```diff
@@ -0,0 +1,117 @@
...
+    [[nodiscard]] auto build_filter(
+            ast::FilterExpr* filter,
+            BasicReaderMap const& basic_readers,
+            ClpStringReaderMap const& clp_string_readers,
+            VarStringReaderMap const& var_string_readers,
+            ClpQueryMap const& clp_queries,
+            VarMatchMap const& var_matches
+    ) const -> Bitmap;
+
+    uint64_t m_num_messages;
```
gibber9809:
```suggestion
    uint64_t m_num_messages{};
```
nit: default initialization.

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,539 @@
...
+            return true;
+        case LiteralType::BooleanT:
+            return is_equality_operation(operation);
+        case LiteralType::ClpStringT:
+            return is_equality_operation(operation) && clp_queries.contains(filter);
+        case LiteralType::VarStringT:
+            return is_equality_operation(operation) && var_matches.contains(filter);
+        case LiteralType::ArrayT:
+        case LiteralType::NullT:
+        case LiteralType::TimestampT:
```
gibber9809, line 377: Didn't notice initially that we don't support null or timestamp in this PR. We should probably add support for both of these in a small follow-up PR once this one is in, since null should be relatively simple, and timestamp will be part of pretty much every search through the package.

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,539 @@
...
+using clp_s::search::ast::OrExpr;
+
+namespace clp_s::search {
+namespace {
+/**
+ * Compares a value against an operand using the given filter operation.
+ * @param operation Comparison operation to apply.
+ * @param value Value read from a column.
+ * @param operand Operand from the filter expression.
+ * @return Whether the comparison succeeds.
```
gibber9809:
```suggestion
 * @return The result of the comparison.
```

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,539 @@
...
+                continue;
+            }
+            if (subquery.wildcard_match_required()) {
+                return clp::string_utils::wildcard_match_unsafe(
+                        std::get<std::string>(reader->extract_value(row)),
+                        query.get_search_string(),
+                        false == query.get_ignore_case()
+                );
+            }
+            return true;
```
gibber9809: Comparing this with the row-based implementation I think there's a subtle bug here. When wildcard match is required for a subquery and this wildcard match evaluates as false we should continue on to checking the next subquery interpretation, but the implementation here will immediately return false without considering the other subquery interpretations.

So something like the following is required
```
if (false == query.contains_sub_queries()) {
    return clp::string_utils::wildcard_match_unsafe(...);
}
for (subquery : subqueries) {
  if (false == matches_logtype() || false == matches_vars) { continue; }
    if (false == subquery.wildcard_match_required()) { return true; }
    if (clp::string_utils::wildcard_match_unsafe(...) { return true; }
}
return false;
```
Aside: probably worth reorganizing the function like I do in that pseudocode so that we can reduce the amount of nesting in this function a little bit, and make it more readable.

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,539 @@
...
+        ColumnScan::VarStringReaderMap const& reader_map,
+        int32_t column_id,
+        FilterOperation operation,
+        std::unordered_set<int64_t> const* matching_vars
+) -> ColumnScan::Bitmap {
+    ColumnScan::Bitmap bitmap(num_messages, 0);
+    auto const readers = reader_map.find(column_id);
+    if (nullptr == matching_vars || reader_map.end() == readers) {
+        return bitmap;
+    }
```
gibber9809: Could we pass matching_vars by reference here instead of by pointer? I think that it's _supposed_ to be always non-null. We can add an assertion in the can_build_x flow to verify that the entry is non-null as well.

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,539 @@
...
+auto ColumnScan::try_create(
+        std::shared_ptr<ast::Expression> const& expression,
+        BasicReaderMap const& basic_readers,
+        ClpStringReaderMap const& clp_string_readers,
+        VarStringReaderMap const& var_string_readers,
+        ClpQueryMap const& clp_queries,
+        VarMatchMap const& var_matches,
+        uint64_t num_messages
+) -> std::optional<ColumnScan> {
+    ColumnScan scan{num_messages};
```
gibber9809: nit: can we move this right before `scan.build_node`, since it isn't needed until then?

gibber9809: Or actually maybe it might make sense to just have the constructor call `build_node` internally as well as set `m_matches`. I think it looks a bit weird to set `scan.m_matches` from this function.

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,539 @@
...
+    while (false == pending.empty()) {
+        auto* cur_expr = pending.back();
+        pending.pop_back();
+
+        if (auto* and_expr = dynamic_cast<AndExpr*>(cur_expr); nullptr != and_expr) {
+            for (auto const& operand : and_expr->get_op_list()) {
+                auto* child = dynamic_cast<ast::Expression*>(operand.get());
+                if (nullptr == child) {
+                    return false;
+                }
```
gibber9809:
```suggestion
```
None of the children should be null by construction

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,539 @@
...
+                pending.push_back(child);
+            }
+            continue;
+        }
+        if (auto* or_expr = dynamic_cast<OrExpr*>(cur_expr); nullptr != or_expr) {
+            for (auto const& operand : or_expr->get_op_list()) {
+                auto* child = dynamic_cast<ast::Expression*>(operand.get());
+                if (nullptr == child) {
+                    return false;
+                }
```
gibber9809:
```suggestion
```

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,539 @@
...
+                pending.push_back(child);
+            }
+            continue;
+        }
+        auto* filter = dynamic_cast<FilterExpr*>(cur_expr);
+        if (false == can_build_filter(filter, clp_queries, var_matches)) {
+            return false;
+        }
```
gibber9809: I think we should also do if (auto* filter) {...} here. That way we can just put a return false afterwards, so if some new expression type gets added that we don't handle here we can correctly return false.

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,539 @@
...
+auto ColumnScan::build_node(
+        ast::Expression* expr,
+        BasicReaderMap const& basic_readers,
+        ClpStringReaderMap const& clp_string_readers,
+        VarStringReaderMap const& var_string_readers,
+        ClpQueryMap const& clp_queries,
+        VarMatchMap const& var_matches
+) const -> Bitmap {
+    assert(nullptr != expr);
+
```
gibber9809:
```suggestion
```
I think our codebase typically prefers returning/throwing errors over asserts. I don't think this assert is needed though, since we already check that this is non-null earlier.

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,539 @@
...
+auto ColumnScan::build_filter(
+        FilterExpr* filter,
+        BasicReaderMap const& basic_readers,
+        ClpStringReaderMap const& clp_string_readers,
+        VarStringReaderMap const& var_string_readers,
+        ClpQueryMap const& clp_queries,
+        VarMatchMap const& var_matches
+) const -> Bitmap {
+    assert(nullptr != filter);
+
```
gibber9809:
```suggestion
```
ditto

gibber9809 (changes requested):
Thanks for the revisions, left another round of comments. Should be pretty close after this round.

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,539 @@
...
+/**
+ * Checks whether a CLP string value matches a query.
+ * @param reader Column reader containing the value to check.
+ * @param query Query to match against.
+ * @param row Row to check.
+ * @return Whether the row's string value matches the query.
+ */
+[[nodiscard]] auto
+clp_string_matches(ClpStringColumnReader* reader, clp::Query const& query, uint64_t row) -> bool;
```
gibber9809: Sorry, forgot to leave this with my other comments.

Instead of `row` here could we use something like `message_index`, or another name that includes `message`? I think that would be more consistent with the `num_messages` terminology we're using elsewhere in this file.

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,537 @@
...
+            bitmap[message_index] |= compare(operation, value, operand) ? 1 : 0;
+        }
+    }
+    return bitmap;
+}
+
+[[nodiscard]] auto
+clp_string_matches(ClpStringColumnReader* reader, clp::Query const& query, uint64_t message_index)
+        -> bool {
+    auto const value = std::get<std::string>(reader->extract_value(message_index));
```
gibber9809: Looks like this rewrite makes it so that we now always marshal the string value up front, whereas the previous implementation would only marshal the string value for this row some of the time.

Since marshalling a clp string has non-trivial overhead, could we change this back to only marshalling the string when we need to perform wildcard matching?

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -0,0 +1,537 @@
...
+    switch (column->get_literal_type()) {
+        case LiteralType::IntegerT:
+        case LiteralType::FloatT:
+            return true;
+        case LiteralType::BooleanT:
+            return is_equality_operation(operation);
+        case LiteralType::ClpStringT:
+            return is_equality_operation(operation) && clp_queries.contains(filter);
+        case LiteralType::VarStringT:
+            return is_equality_operation(operation) && var_matches.contains(filter);
```
gibber9809: Should probably add one more check for `nullptr != var_matches.at(filter)` to be safe.

gibber9809 (changes requested):
Nice work! Just a few minor details, but otherwise looks good.

gibber9809 (approved):
LGTM. For PR title, how about:
```
feat(clp-s::search): Add a column-based filtering implementation for a subset of query shapes.
```

PR #2284, opened 2026-05-15
feat(clp-s::log_converter): Re-use the same `log_surgeon::BufferParser` for converting each file.

gibber9809:
# Description

This PR modifies the `LogConverter` class to have it own an instance of `log_surgeon::BufferParser` that is created during initialization, and re-used to convert each input file. The previous implementation created a new `log_surgeon::BufferParser` on each invocation of `convert_file`, which adds quite a bit of overhead when converting many small files.

Conversion of the `hive-24hr` dataset goes from about 10 minutes down to about 2 minutes on an old xeon server I tested this change on.

# Checklist

  breaking change.

# Validation performed

* Manually validated that conversion is significantly faster when ingesting many small files
* Validated that integration tests still pass

## Release Notes

* **Refactor**
  * Improved internal log converter initialization by centralizing object construction and streamlining the parsing workflow.

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/clp/pull/2284)


components/core/src/clp_s/log_converter/LogConverter.hpp
```diff
@@ -66,6 +75,7 @@ class LogConverter {
      */
     [[nodiscard]] auto grow_buffer_if_full() -> ystdlib::error_handling::Result<void>;
 
+    log_surgeon::BufferParser m_parser;
     ystdlib::containers::Array<char> m_buffer;
     size_t m_num_bytes_buffered{};
     size_t m_parser_offset{};
```
davidlion, line 81: I don't think clp follows the alphabetical rule from https://docs.yscope.com/dev-guide/contrib-guides-general.html very strictly, but technically this does violate it.
I'll leave it up to you if you want to fix it.

davidlion (approved):
lgtm other than the optional nit.

components/core/src/clp_s/log_converter/LogConverter.hpp
gibber9809: I think I'll leave this one as is since it leaves the members grouped together in a more logical way, but will try to keep this rule in mind.

PR #2286, opened 2026-05-16
feat(telemetry): Add `log-ingestor` and `api-server` counters.

Nathan903:
# Description

Adds telemetry metrics to CLP's Rust services (api-server, log-ingestor).
- New `telemetry` module in `clp-rust-utils` with `init_telemetry()`, `TelemetryGuard`, and `shutdown_telemetry()`
 - `api-server`: emits `clp.service.event` counter on startup
- `log-ingestor`: emits `clp.ingest.bytes_total` and `clp.ingest.records_total` counters per ingested object in `Buffer::add`
- Bug fix: `CLP_DISABLE_TELEMETRY` now checks the env var's value against `["1", "true", "yes", "y"]` instead of checking existence

# Checklist

  breaking change.

# Validation performed

1. run 
```
sbin/start-clp.sh
curl -X POST http://localhost:3002/s3_scanner   -H "Content-Type: application/json"   -d '{
    "bucket_name": "trash-nathan-test",
    "key_prefix": "hive-24hr-ts.jsonl",
    "region": "us-east-2",
    "dataset": "test_dataset_3",
    "scanning_interval_sec": 60
  }'
```
2. check clp-telemetry-server

image

## Summary by CodeRabbit

* **New Features**
  * Application-wide telemetry and metrics added: service startup event, ingest counters and service event counters.
  * Services initialize monitoring on startup and gracefully shut down telemetry on exit.
  * Telemetry can be disabled via configuration or the CLP_DISABLE_TELEMETRY environment variable.

* **Bug Fixes**
  * OTLP exporter build failures are surfaced as errors instead of causing panics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/clp/pull/2286?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


components/clp-rust-utils/src/telemetry.rs
```diff
@@ -0,0 +1,46 @@
...
+        .with_http()
+        .with_endpoint(endpoint)
+        .build()
+        .map_err(|e| Error::TelemetryExporterBuild(e.to_string()))?;
+
+    let reader = PeriodicReader::builder(exporter, opentelemetry_sdk::runtime::Tokio).build();
+    let provider = SdkMeterProvider::builder()
+        .with_reader(reader)
+        .with_resource(Resource::default())
+        .build();
```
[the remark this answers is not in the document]
Nathan903: unnecessary as callers are responsible for setting service name

components/clp-rust-utils/src/telemetry.rs
```diff
@@ -0,0 +1,83 @@
...
+    fn drop(&mut self) {
+        shutdown_telemetry(self.0.take());
+    }
+}
+
+pub fn shutdown_telemetry(provider: Option<SdkMeterProvider>) {
+    if let Some(p) = provider
+        && let Err(err) = p.shutdown()
+    {
+        eprintln!("Error shutting down telemetry: {err}");
```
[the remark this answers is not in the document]
Nathan903: ignored since this would need adding `tracing` dependency for just this one use case

components/clp-rust-utils/src/telemetry.rs
```diff
@@ -0,0 +1,73 @@
+use std::env;
+
+use opentelemetry::global;
+use opentelemetry_sdk::metrics::{PeriodicReader, SdkMeterProvider};
+
+use crate::{Error, clp_config::package::config::Telemetry};
+
+/// Values accepted by `CLP_DISABLE_TELEMETRY` and `DO_NOT_TRACK` to disable telemetry.
```
LinZhihao-723: Not familiar with the previous decisions made: I assume `CLP_DISABLE_TELEMETRY` and `DO_NOT_TRACK` are decided outside of this PR. Please reply to this comment to confirm.

components/clp-rust-utils/src/telemetry.rs
```diff
@@ -0,0 +1,73 @@
...
+
+/// Initializes OpenTelemetry metrics collection with the provided configuration.
+///
+/// Returns `Ok(None)` if telemetry is disabled, `Ok(Some(provider))` on success,
+/// or an `Err` if the metric exporter cannot be built.
+///
+/// # Errors
+///
+/// Returns an error if the OTLP metric exporter fails to build (e.g., invalid
+/// endpoint configuration or missing HTTP client support).
```
LinZhihao-723: The docstring doesn't match our coding standard.
- In terms of formatting, please make sure to wrap the line only if the next word can't fit in the 100-character/line limit.
- In terms of wording: please check [this](https://github.com/LinZhihao-723/claude-instruction/blob/main/rust/coding-style.md#docstrings) section. This is a WIP Rust coding guideline, it might be worth reading if you need to code more in Rust. Or it might be worth asking Claude or whatever agent you're using to review the code using the skills under this given repo.

components/clp-rust-utils/src/error.rs
```diff
@@ -13,4 +13,7 @@ pub enum Error {
 
     #[error("`sqlx::Error`: {0}")]
     Sqlx(#[from] sqlx::Error),
+
+    #[error("Failed to build OTLP metric exporter: {0}")]
+    TelemetryExporterBuild(String),
```
LinZhihao-723: Why not:

In this way, this particular enum would be a wrapper of the original telemetry error.
```suggestion
    #[error(transparent)]
    TelemetryExporterBuildError(#[from] opentelemetry_otlp::ExporterBuildError),
```

components/clp-rust-utils/src/telemetry.rs
```diff
@@ -0,0 +1,73 @@
...
+pub fn init_telemetry(telemetry_config: &Telemetry) -> Result<Option<SdkMeterProvider>, Error> {
```
LinZhihao-723: It might make more sense to return `Option<TelemetryGuard>`, otherwise the user code needs to construct `TelemetryGuard` manually, which is not ideal.

components/clp-rust-utils/src/telemetry.rs
```diff
@@ -0,0 +1,73 @@
...
+use opentelemetry::global;
+use opentelemetry_sdk::metrics::{PeriodicReader, SdkMeterProvider};
+
+use crate::{Error, clp_config::package::config::Telemetry};
+
+/// Values accepted by `CLP_DISABLE_TELEMETRY` and `DO_NOT_TRACK` to disable telemetry.
+///
+/// NOTE: This must be kept consistent with the Python implementation in
+/// `clp_package_utils/scripts/start_clp.py`.
+const TELEMETRY_DISABLE_VALUES: [&str; 4] = ["1", "true", "yes", "y"];
```
LinZhihao-723: Private symbols should be defined after public symbols (as required [here](https://github.com/LinZhihao-723/claude-instruction/blob/main/rust/coding-style.md#symbol-ordering)).

components/clp-rust-utils/src/telemetry.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    let provider = SdkMeterProvider::builder().with_reader(reader).build();
+
+    global::set_meter_provider(provider.clone());
+    Ok(Some(provider))
+}
+
+/// RAII guard that ensures [`shutdown_telemetry`] is called when the guard is
+/// dropped, even if the process unwinds or returns early.
+pub struct TelemetryGuard(Option<SdkMeterProvider>);
```
LinZhihao-723: We should not use tuple struct in Rust:

```suggestion
pub struct TelemetryGuard {
    provider: Option<SdkMeterProvider>
};
```

components/clp-rust-utils/Cargo.toml
```diff
@@ -21,6 +21,9 @@ thiserror = "2.0.18"
 tracing-appender = "0.2.4"
 tracing-subscriber = { version = "0.3.22", features = ["json", "env-filter", "fmt", "std"] }
 utoipa = { version = "5.4.0" }
+opentelemetry = { version = "0.32.0" }
+opentelemetry_sdk = { version = "0.32.0", features = ["metrics", "rt-tokio"] }
+opentelemetry-otlp = { version = "0.32.0", default-features = false, features = ["metrics", "http-proto", "reqwest-blocking-client"] }
```
LinZhihao-723: Make sure to follow alpha-beta-ordering.

components/clp-rust-utils/src/telemetry.rs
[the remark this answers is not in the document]
LinZhihao-723: I think this is a good concern though: all Rust components have tracing to be the single source of logging for now. Printing to stderr doesn't guarantee the error will be caught by our logging framework.
Let's use tracing.

LinZhihao-723 (changes requested):
First batch of comments: reviewed telemetry setup in `clp-rust-utils`.

components/api-server/src/error.rs
```diff
@@ -82,6 +82,9 @@ impl From<clp_rust_utils::Error> for ClientError {
             }
             clp_rust_utils::Error::Io(error) => error.into(),
             clp_rust_utils::Error::Sqlx(error) => error.into(),
+            clp_rust_utils::Error::TelemetryExporterBuild(msg) => {
+                Self::Io(std::io::Error::other(msg))
+            }
```
LinZhihao-723: Similarly, we should have a variant of the telemetry error.

components/api-server/src/bin/api_server.rs
```diff
@@ -71,10 +77,13 @@ async fn main() -> anyhow::Result<()> {
...
+    startup_counter.add(1, &[opentelemetry::KeyValue::new("type", "start")]);
 
     tracing::info!("Server started at {addr}");
-    axum::serve(listener, router)
+    let serve_result = axum::serve(listener, router)
         .with_graceful_shutdown(shutdown_signal())
-        .await?;
+        .await;
+
+    serve_result?;
```
LinZhihao-723: We shouldn't need this change, right?

components/log-ingestor/src/compression/buffer.rs
```diff
@@ -36,6 +37,8 @@ pub struct Buffer<Submitter: BufferSubmitter> {
     buf: Vec<S3ObjectMetadataId>,
     total_size: u64,
     size_threshold: u64,
+    bytes_total: opentelemetry::metrics::Counter<u64>,
+    records_total: opentelemetry::metrics::Counter<u64>,
```
LinZhihao-723: We should improve the naming here (including the ones registered in line 57&58):
- `bytes_total` -> `total_num_bytes`
- `records_total` -> `total_num_objects`

Reasons:
1. We usually use `num_xxx` to name counters; in this case, since the counter is global.
2. "record" is not a precise term. "Number of record" usually refers to the total number of log events, not the number of files/objects ingested. Since this buffer is only for S3 objects, it's probably better to use "num_objects".

In addition, I would suggest wrapping these two counters into a struct named `TelemetryMeter`, so that variable names don't confuse with other data members like `total_size`. We could expose a method `ingest(&self, num_bytes: u64)` to encapsulate the two counter operations in line 81&82.

Feel free to confirm with @junhaoliao

LinZhihao-723 (changes requested):
Reviewed metrics added in the API server and log-ingestor.

components/clp-rust-utils/src/telemetry.rs
Nathan903: yes, it was added in https://github.com/y-scope/clp/pull/2251

components/clp-rust-utils/src/telemetry.rs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        tracing::error!(err = ? err, "Error shutting down telemetry.");
+    }
+}
+
+/// Values accepted by `CLP_DISABLE_TELEMETRY` and `DO_NOT_TRACK` to disable telemetry.
+///
+/// NOTE: This must be kept consistent with the Python implementation in
+/// `clp_package_utils/scripts/start_clp.py`.
+const TELEMETRY_DISABLE_VALUES: [&str; 4] = ["1", "true", "yes", "y"];
```
LinZhihao-723, line 86: Nit:
```suggestion
const TELEMETRY_DISABLE_VALUES: [&'static str; 4] = ["1", "true", "yes", "y"];
```

components/clp-rust-utils/src/telemetry.rs
```diff
@@ -0,0 +1,81 @@
...
+
+/// Shuts down the given meter provider, flushing any pending metric exports.
+///
+/// If the shutdown fails, the error is logged and not propagated, since this is typically called
+/// during teardown (e.g., in a [`Drop`] implementation) where returning an error is not useful.
+pub fn shutdown_telemetry(provider: Option<SdkMeterProvider>) {
+    if let Some(p) = provider
+        && let Err(err) = p.shutdown()
+    {
+        tracing::error!(err = ? err, "Error shutting down telemetry.");
```
LinZhihao-723:
```suggestion
        tracing::error!(err = ? err, "Failed to shutdown OpenTelemetry meter provider.");
```

components/clp-rust-utils/src/telemetry.rs
```diff
@@ -0,0 +1,81 @@
...
+    global::set_meter_provider(provider.clone());
+    Ok(Some(TelemetryGuard::from(Some(provider))))
+}
+
+/// Shuts down the given meter provider, flushing any pending metric exports.
+///
+/// If the shutdown fails, the error is logged and not propagated, since this is typically called
+/// during teardown (e.g., in a [`Drop`] implementation) where returning an error is not useful.
+pub fn shutdown_telemetry(provider: Option<SdkMeterProvider>) {
+    if let Some(p) = provider
```
LinZhihao-723: This method shouldn't be made public. Let's drop `pub` marker and move it down to the const decl.

components/clp-rust-utils/src/telemetry.rs
```diff
@@ -0,0 +1,81 @@
...
+
+use opentelemetry::global;
+use opentelemetry_sdk::metrics::{PeriodicReader, SdkMeterProvider};
+
+use crate::{Error, clp_config::package::config::Telemetry};
+
+/// RAII guard that ensures [`shutdown_telemetry`] is called when the guard is dropped, even if the
+/// process unwinds or returns early.
+///
+/// Use [`init_telemetry`] to create an instance.
```
LinZhihao-723:
```suggestion
/// RAII guard that ensures [`shutdown_telemetry`] is called when the guard is dropped.
```
Should be sufficient.

components/log-ingestor/src/compression/buffer.rs
```diff
@@ -36,20 +36,22 @@ pub struct Buffer<Submitter: BufferSubmitter> {
     buf: Vec<S3ObjectMetadataId>,
     total_size: u64,
     size_threshold: u64,
+    telemetry: TelemetryMeter,
```
LinZhihao-723: `telemetry` -> `telemetry_meter`

LinZhihao-723 (changes requested):
Some more fixes before approval.

components/clp-rust-utils/src/telemetry.rs
Nathan903: I got this error after adding `'static`:
```console
error: constants have by default a `'static` lifetime
  --> components/clp-rust-utils/src/telemetry.rs:78:35
   |
78 | const TELEMETRY_DISABLE_VALUES: [&'static str; 4] = ["1", "true", "yes", "y"];
   |                                  -^^^^^^^---- help: consider removing `'static`: `&str`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_static_lifetimes
   = note: `-D clippy::redundant-static-lifetimes` implied by `-D clippy::all`
   = help: to override `-D clippy::all` add `#[allow(clippy::redundant_static_lifetimes)]`
```

components/clp-rust-utils/src/telemetry.rs
```diff
@@ -27,6 +24,18 @@ impl From<Option<SdkMeterProvider>> for TelemetryGuard {
...
+///
+/// If the shutdown fails, the error is logged and not propagated, since this is typically called
+/// during teardown (e.g., in a [`Drop`] implementation) where returning an error is not useful.
+fn shutdown_telemetry(provider: Option<SdkMeterProvider>) {
+    if let Some(p) = provider
+        && let Err(err) = p.shutdown()
+    {
+        tracing::error!(err = ? err, "Failed to shutdown OpenTelemetry meter provider.");
+    }
+}
```
LinZhihao-723: As indicated in the previous [comment](https://github.com/y-scope/clp/pull/2286#discussion_r3364928217), as well as the WIP guideline [here](https://github.com/LinZhihao-723/claude-instruction/blob/main/rust/coding-style.md#symbol-ordering): this method should be moved down to the const decl, not in the middle of pub symbols.

LinZhihao-723 (changes requested):

components/clp-rust-utils/src/telemetry.rs
Nathan903: mb, i accidentally moved it up not down. Now looking at this, i think maybe it can be made into a `TelemetryGuard` private method. what do you think about https://github.com/y-scope/clp/pull/2286/changes/338951706bf5e8e96304e4e1e1ece93217787ed9 ?

components/log-ingestor/src/compression/buffer.rs
```diff
@@ -36,20 +36,22 @@ pub struct Buffer<Submitter: BufferSubmitter> {
...
-    /// A newly created [`Buffer`] with the given submitter of type `T` and size threshold.
-    pub const fn new(submitter: Submitter, size_threshold: u64) -> Self {
+    /// A newly created [`Buffer`] with the given submitter and size threshold.
+    pub fn new(submitter: Submitter, size_threshold: u64) -> Self {
         Self {
             submitter,
             buf: Vec::new(),
             total_size: 0,
             size_threshold,
+            telemetry_meter: TelemetryMeter::new(),
```
junhaoliao: a TelemetryMeter creation might be cheap, but there doesn't seem to be any semantic isolation (all meters use name `"log-ingestor"`) anyways even now we're creating one per Buffer? what are the benefits of letting each buffer own its meter?

junhaoliao: if i understand correctly, we are not really recreating the meter but actually the counter handle here, so i guess this is fine

components/clp-rust-utils/src/telemetry.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    let exporter = opentelemetry_otlp::MetricExporter::builder()
+        .with_http()
+        .build()?;
+
+    let reader = PeriodicReader::builder(exporter).build();
+
+    let provider = SdkMeterProvider::builder().with_reader(reader).build();
+
+    global::set_meter_provider(provider.clone());
```
junhaoliao: [Libraries should NOT call this function.](https://docs.rs/opentelemetry/latest/opentelemetry/global/fn.set_meter_provider.html#:~:text=Libraries%20should%20NOT%20call%20this%20function.)

can we call this function in application code (log-ingestor, api-server) instead?

components/log-ingestor/src/compression/buffer.rs
```diff
@@ -115,3 +118,30 @@ impl<Submitter: BufferSubmitter> Buffer<Submitter> {
...
+    /// Factory function.
+    ///
+    /// # Returns
+    ///
+    /// A newly created [`TelemetryMeter`] with counters registered in the global meter provider.
+    fn new() -> Self {
+        let meter = opentelemetry::global::meter("log-ingestor");
+        Self {
+            total_num_bytes: meter.u64_counter("clp.ingest.total_num_bytes").build(),
+            total_num_objects: meter.u64_counter("clp.ingest.total_num_objects").build(),
```
junhaoliao: [    // It is recommended to reuse the same counter instance for the
    // lifetime of the application](https://docs.rs/opentelemetry/latest/opentelemetry/global/index.html#:~:text=//%20It%20is%20recommended%20to%20reuse%20the%20same%20counter%20instance%20for%20the%0A%20%20%20%20//%20lifetime%20of%20the%20application)

can we create this counter somewhere and let it have the same lifecycle as the application, and we use directly call `add()` on the instance whenever we need to?

junhaoliao: https://opentelemetry.io/docs/languages/rust/getting-started/ shows an example of creating global counters using [OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html). can we refer to that?

let's first get @LinZhihao-723 to confirm if it's fine to use static / global like that

junhaoliao: it might be cleaner to use LazyLock? not sure why otel doesn't use that in the example

junhaoliao (changes requested):

components/log-ingestor/src/compression/buffer.rs
junhaoliao: probably not fine, we shouldn't call `build()` repeatedly - see https://github.com/y-scope/clp/pull/2286#discussion_r3383151353 for the discussion

components/log-ingestor/src/compression/buffer.rs
```diff
@@ -71,6 +73,7 @@ impl<Submitter: BufferSubmitter> Buffer<Submitter> {
     ) -> Result<bool> {
         let mut submission_triggered = false;
         for entry in object_metadata_to_ingest {
+            self.telemetry_meter.ingest(entry.size);
```
junhaoliao: would it be too frequent to log this for every file added? would it be better to do this in `submit()` instead?

```rs
        let bytes = self.total_size;
        let records = self.buf.len() as u64;
        self.submitter.submit(&self.buf).await?;
        self.telemetry_meter.ingest(bytes, records);
```

components/log-ingestor/src/compression/buffer.rs
LinZhihao-723: I think using a singleton pattern should be fine. We should create a telemetry mod in log-ingestor then to hold these metrics counters.

components/log-ingestor/src/compression/buffer.rs
junhaoliao: cc @LinZhihao-723

components/log-ingestor/src/compression/buffer.rs
LinZhihao-723: If to use a global var, I will probably do the following:
- Create a mod telemetry.rs.
- Have the global counters private to this mod.
- Provide public methods to mutate the counters with semantics, for example, `fn ingest_from_s3(total_num_objects: usize, total_num_bytes: usize)`.

components/log-ingestor/src/compression/buffer.rs
LinZhihao-723: - Performance-wise, it seems fine to call on every ingested file: the underlying counter is an atomic variable, so it's just an atomic add operation.
- But I just realized this might not be the best location to collect the stats: we should probably move this inside [ClpIngestionState](https://github.com/y-scope/clp/blob/d8806847105bfee3b6164c8bc762dc9a09aa252f/components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs#L912). This is the method where the ingestion jobs are supposed to call on every file ingested.

components/clp-rust-utils/src/telemetry.rs
Nathan903: does the *libraries* refer to any utils file? i read it to mean just projects thats meant to be library crates. If we move `set_meter_provider` to application code it would cause code duplication. I assume the reason the opentelemetry docs says not to call this function in libraries is to prevent a user importing multiple libraries have each library call set_meter_provider and having conflicts. But its only 3-4 loc so i think moving `set_meter_provider` to application is also fine.

components/api-server/src/bin/api_server.rs
```diff
@@ -53,6 +53,16 @@ async fn main() -> anyhow::Result<()> {
     let (config, credentials) = read_config_and_credentials(&args)?;
     let _guard = clp_rust_utils::logging::set_up_logging("api_server.log");
 
+    let tel_guard = clp_rust_utils::telemetry::init_telemetry(&config.telemetry)?;
+    if let Some(guard) = &tel_guard
+        && let Some(provider) = guard.provider()
```
LinZhihao-723: Sorry for missing this in the previous round: since the provider is already an optional var and its destruction is handled by the guard's drop by checking if it's none, `init_telemetry` should probably not return an optional value; returning a guard with the provider set to none might be sufficient.

docs/src/user-docs/reference-telemetry.md
LinZhihao-723, line 1: Didn't review this section. Looks like it's also adding docs for things outside of this PR. Leave it to @junhaoliao.

LinZhihao-723 (changes requested):

components/clp-rust-utils/src/telemetry.rs
junhaoliao: that's fair. The important part is that this `init_telemetry` must be an explicit application-startup action, not something triggered implicitly by generic library/instrumentation code.

Although `clp-rust-utils` is technically a library crate, `init_telemetry()` is being used as a shared startup helper called directly from each binary's `main()`. If each application calls it exactly once and we don't support another provider being installed separately, I think it's reasonable for `init_telemetry()` to call `global::set_meter_provider(...)` internally to avoid duplicating the setup boilerplate.

Let's set the provider in `init_telemetry()` and document clearly about the expected usages. e.g.,
```
/// Initializes OpenTelemetry metrics collection with the provided configuration.
///
/// This function is intended for application startup code and should be called exactly once during process initialization.
```

docs/src/user-docs/reference-telemetry.md
junhaoliao: fine with me. let's run `npx markdownlint-cli2 --config ./tools/yscope-dev-utils/exports/lint-configs/.markdownlint-cli2.yaml --fix docs/src
/user-docs/reference-telemetry.md` and fix all lint errors then this is good to go

junhaoliao (changes requested):
good points raised and discussed at https://github.com/y-scope/clp/pull/2286#discussion_r3398460131 

once those are addressed, this PR is good to go

docs/src/user-docs/reference-telemetry.md
Nathan903: this npx command fixed a lot of files. I should only fix reference-telemetry.md right?
```
        modified:   docs/src/dev-docs/components-core/index.md
        modified:   docs/src/dev-docs/contributing-linting.md
        modified:   docs/src/dev-docs/design-parsing-wildcard-queries.md
        modified:   docs/src/index.md
        modified:   docs/src/user-docs/core-clp-s.md
        modified:   docs/src/user-docs/core-unstructured/glt.md
        modified:   docs/src/user-docs/guides-k8s-deployment.md
        modified:   docs/src/user-docs/quick-start/index.md
        modified:   docs/src/user-docs/reference-json-search-syntax.md
        modified:   docs/src/user-docs/reference-telemetry.md
        modified:   docs/src/user-docs/reference-text-search-syntax.md
        modified:   docs/src/user-docs/reference-unstructured-schema-file.md
```

junhaoliao: yes... only the reference-telemetry.md file

sorry I mistakenly added a space after /src in the command I shared

junhaoliao: please do fix the Markdown tables as well. not sure if there's an OSS tool but any Jetbrains IDE can format such

Nathan903: done

docs/src/user-docs/reference-telemetry.md
```diff
@@ -10,6 +10,64 @@ As an open-source project, we have limited visibility into how CLP is used in th
...
+
+#### Operational counters
+
+Emitted by long-running CLP services to track throughput:
+
+| Component | Metric | Type | Description |
+| --- | --- | --- | --- |
+| log-ingestor | `clp.ingest.total_num_bytes` | Counter | Total bytes ingested |
+| log-ingestor | `clp.ingest.total_num_objects` | Counter | Total objects (log events) ingested |
+| api-server | `clp.service.event` | Counter | Service lifecycle events (e.g., startup) |
```
junhaoliao:
```suggestion
| Component    | Metric                         | Type    | Description                              |
| ------------ | ------------------------------ | ------- | ---------------------------------------- |
| log-ingestor | `clp.ingest.total_num_bytes`   | Counter | Total bytes ingested                     |
| log-ingestor | `clp.ingest.total_num_objects` | Counter | Total objects (log events) ingested      |
| api-server   | `clp.service.event`            | Counter | Service lifecycle events (e.g., startup) |
```

docs/src/user-docs/reference-telemetry.md
```diff
@@ -10,6 +10,64 @@ As an open-source project, we have limited visibility into how CLP is used in th
...
+Emitted once at startup by the controller to record deployment sizing:
+
+| Metric | Type | Description |
+| --- | --- | ----------- |
+| `clp.deployment.compression_worker_replicas` | Gauge | Number of compression-worker replicas |
+| `clp.deployment.compression_worker_concurrency` | Gauge | Compression-worker concurrency |
+| `clp.deployment.query_worker_replicas` | Gauge | Number of query-worker replicas |
+| `clp.deployment.query_worker_concurrency` | Gauge | Query-worker concurrency |
+| `clp.deployment.reducer_replicas` | Gauge | Number of reducer replicas |
+| `clp.deployment.reducer_concurrency` | Gauge | Reducer concurrency |
```
junhaoliao:
```suggestion
| Metric                                          | Type  | Description                           |
| ----------------------------------------------- | ----- | ------------------------------------- |
| `clp.deployment.compression_worker_replicas`    | Gauge | Number of compression-worker replicas |
| `clp.deployment.compression_worker_concurrency` | Gauge | Compression-worker concurrency        |
| `clp.deployment.query_worker_replicas`          | Gauge | Number of query-worker replicas       |
| `clp.deployment.query_worker_concurrency`       | Gauge | Query-worker concurrency              |
| `clp.deployment.reducer_replicas`               | Gauge | Number of reducer replicas            |
| `clp.deployment.reducer_concurrency`            | Gauge | Reducer concurrency                   |
```

docs/src/user-docs/reference-telemetry.md
```diff
@@ -10,6 +10,64 @@ As an open-source project, we have limited visibility into how CLP is used in th
...
+| --- | --- | --- |
+| `clp.deployment.id` | `550e8400-e29b-41d4-a716-446655440000` | Deduplicate metrics from the same deployment |
+| `service.version` | `0.9.1` | Track version adoption |
+| `clp.deployment.method` | `docker-compose` or `helm` | Understand deployment preferences |
+| `clp.storage.engine` | `clp-s` or `clp` | Track feature adoption |
+| `service.name` | `log-ingestor`, `api-server`, `controller` | Identify the emitting component |
+| `host.arch` | `x86_64`, `aarch64` | Inform build target priorities |
+| `host.cpu.family` | `6` | Inform build target priorities |
+| `host.cpu.model.id` | `142` | Inform build target priorities |
+| `host.cpu.vendor.id` | `GenuineIntel` | Inform build target priorities |
```
junhaoliao:
```suggestion
| Resource Attribute      | Example                                    | Purpose                                      |
| ----------------------- | ------------------------------------------ | -------------------------------------------- |
| `clp.deployment.id`     | `550e8400-e29b-41d4-a716-446655440000`     | Deduplicate metrics from the same deployment |
| `service.version`       | `0.9.1`                                    | Track version adoption                       |
| `clp.deployment.method` | `docker-compose` or `helm`                 | Understand deployment preferences            |
| `clp.storage.engine`    | `clp-s` or `clp`                           | Track feature adoption                       |
| `service.name`          | `log-ingestor`, `api-server`, `controller` | Identify the emitting component              |
| `host.arch`             | `x86_64`, `aarch64`                        | Inform build target priorities               |
| `host.cpu.family`       | `6`                                        | Inform build target priorities               |
| `host.cpu.model.id`     | `142`                                      | Inform build target priorities               |
| `host.cpu.vendor.id`    | `GenuineIntel`                             | Inform build target priorities               |
```

junhaoliao (changes requested):

junhaoliao (changes requested):
for the title, how about
```
feat(telemetry): Add `log-ingestor` and `api-server` counters.
```

junhaoliao (approved):
note there's a suggestion for the title  https://github.com/y-scope/clp/pull/2286#pullrequestreview-4481102204

Nathan903:
> note there's a suggestion for the title [#2286 (review)](https://github.com/y-scope/clp/pull/2286#pullrequestreview-4481102204)

thanks, was outside, I just changed title now. I re-ran markdown lint and it worked. 
Somehow markdown linter originally says to make table more compact, and the previous table's formatting (with minimal space and dashes) also passes the linter. I will go with the full size formatted table in the future

Nathan903:
Ive re-ran the validations and everything works. 

just putting this here for the record to go with the PR 
I noticed that after an ingestion job, the ingestor will keep on ingesting again and again (about every 2min). if i run this:
```
sbin/start-clp.sh
curl -X POST http://localhost:3002/s3_scanner   -H "Content-Type: application/json"   -d '{
    "bucket_name": "trash-nathan-test",
    "key_prefix": "hive-24hr-ts.jsonl",
    "region": "us-east-2",
    "dataset": "test_dataset_3",
    "scanning_interval_sec": 60
  }'
  ```

I get this:

image

at first i thought it is a telemetry bug but then I find out its clp actually ingesting hundreds of GB of data
```
nathanhung@baker18:/home/nathanhung/clp/build/clp-package$ # see how many ingestion
docker exec -it $(docker ps --filter "name=database" --format "{{.ID}}" | head -1) bash -c 'mysql -u $MYSQL_USER -p$MYSQL_PASSWORD $MYSQL_DATABASE -e "SELECT \`key\`, count(*) as ingestion_count FROM ingested_s3_object_metadata WHERE \`key\` = '\''hive-24hr-ts.jsonl'\'' GROUP BY \`key\`;"'
+--------------------+-----------------+
| key                | ingestion_count |
+--------------------+-----------------+
| hive-24hr-ts.jsonl |             286 |
+--------------------+-----------------+
nathanhung@baker18:/home/nathanhung/clp/build/clp-package$ # total rows vs unique objects
docker exec -it $(docker ps --filter "name=database" --format "{{.ID}}" | head -1) \
  bash -c 'mysql -u $MYSQL_USER -p$MYSQL_PASSWORD $MYSQL_DATABASE -e "
  SELECT COUNT(*) AS total_rows,
         COUNT(DISTINCT CONCAT(bucket, '"'"'/'"'"', \`key\`)) AS unique_objects,
         SUM(size) AS total_bytes
  FROM ingested_s3_object_metadata;"'
+------------+----------------+--------------+
| total_rows | unique_objects | total_bytes  |
+------------+----------------+--------------+
|        286 |              1 | 673136435686 |
+------------+----------------+--------------+
nathanhung@baker18:/home/nathanhung/clp/build/clp-package$ # scanner state
docker exec -it $(docker ps --filter "name=database" --format "{{.ID}}" | head -1) \
  bash -c 'mysql -u $MYSQL_USER -p$MYSQL_PASSWORD $MYSQL_DATABASE -e "
  SELECT id, last_ingested_key FROM ingestion_job_s3_scanner_state;"'
+----+--------------------+
| id | last_ingested_key  |
+----+--------------------+
|  1 | hive-24hr-ts.jsonl |
+----+--------------------+
nathanhung@baker18:/home/nathanhung/clp/build/clp-package$
[1:12](https://yscope.slack.com/archives/C09G46K14JK/p1781197946419669?thread_ts=1781193759.687149&cid=C09G46K14JK)
```

LinZhihao-723 (approved):
The Rust code-level changes lgtm. Will look into the issue soon.

LinZhihao-723:
@coderabbitai Please create an issue for https://github.com/y-scope/clp/pull/2286#issuecomment-4688186954.

PR #2290, opened 2026-05-21
feat(yscope-clp-core): Emit single-file archives as files instead of directories; Improve `ClpArchiveWriter` lifecycle and API semantics.

Bill-hbrhbr:
# Description

Previously, `ClpArchiveWriter` emitted the CLP JSON single-file archives inside directories. This behavior differed from typical file-oriented compression tools such as `gzip` and `zstd`, where the provided output path corresponds directly to the generated file. This PR updates the writer APIs so single-file archives are emitted as regular files at the caller-provided output path.

The PR also refines the `ClpArchiveWriter` lifecycle and API semantics so that compression, cleanup, and closed-state behavior are clearer and more consistent for callers:
- `close()` now finalizes the archive and cleans up resources.
  -  Previously, `close()` only cleaned up temporary resources. It did not finalize or compress the archive.
- `__exit__()` follows the same lifecycle semantics as `close()`, but for context-manager usage.
- `_cleanup()` is now responsible only for temporary resource cleanup.
  - This is what `close()` used to do.
- Introduces `flush()`, which is a no-op since `ClpArchiveWriter` is not appendable.

Lastly, a small public helper is added for detecting CLP JSON single-file archives through magic-byte inspection.

# Checklist

  breaking change.

# Validation performed

* [X] Works with Tritonparse feature PR: https://github.com/y-scope/tritonparse/pull/2/changes#diff-3bc3c9c8048fe7f269d025f3c8a80453d882d5cd51081fea571206cd6a29d92fR365

### DIFF
--- a/python-wheels/yscope-clp-core/yscope_clp_core/__init__.py
+++ b/python-wheels/yscope-clp-core/yscope_clp_core/__init__.py
@@ -29,9 +29,9 @@ from yscope_clp_core._except import (
 )
 
 __all__ = [
+    "CLP_SFA_MAGIC_BYTES",
     "ArchiveClosedError",
     "BadArchiveSourceError",
-    "CLP_SFA_MAGIC_BYTES",
     "BadCompressionInputError",
     "ClpArchiveReader",
     "ClpArchiveWriter",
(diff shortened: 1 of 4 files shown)


Bill-hbrhbr:
@claude review

PR #2309, opened 2026-06-04
feat(clp-s::search): Instrument search with OpenTelemetry spans.

ShangDanLuXian:
# Description

Instruments `clp-s` archive search with OpenTelemetry. For each archive searched, `search_archive`
now records one span (`clp_s.search.archive`, instrumentation scope `clp_s.search`,
`service.name=clp-search`) describing the search:

* **Query shape:** column-type and predicate-type counts, number of predicates, whether an OR clause
  is present, and the search time range.
* **Results:** total archive records, candidate records after schema matching, records matching the
  query, the number of matched schemas, and how many of those produced at least one match.
* **Termination:** the stage at which the search stopped (`clp.search.termination_stage`).
* **Query identity:** the archive id; the scheduler's `query_id`/`task_id` (read from
  `CLP_QUERY_ID`/`CLP_TASK_ID`); a non-reversible xxHash fingerprint of the query
  (`clp.search.query_hash`); and the raw query string only when `CLP_TELEMETRY_INCLUDE_QUERY` is set
  (off by default, to honor the "no queries" telemetry policy).

This PR wires the instrumentation against the **OpenTelemetry API only** — `opentelemetry-cpp` is
built API-only, so the spans go to the API's default no-op provider and nothing is exported yet
(and no Protobuf dependency is added). Installing an SDK tracer provider + OTLP exporter so the
spans are actually exported is a stacked follow-up PR.

# Checklist

  breaking change.

# Validation performed

Here are a few dumped files to stderr from my debug branch:
https://github.com/ShangDanLuXian/clp/tree/telemetry-stderr-debug/telemetry-dumps

## Summary by CodeRabbit

* **New Features**
  * Search telemetry with an opt-in CLI flag to publish OpenTelemetry spans for searches.

* **Improvements**
  * Archive message-count lookup per schema exposed.
  * Search output reports richer per-archive/per-schema metrics and termination-stage info for telemetry and retrieval.

* **Documentation**
  * Core component docs list opentelemetry-cpp dependency/version.

* **Chores**
  * Build tooling and taskfile updated to fetch/install and wire opentelemetry-cpp into the core build.


components/core/src/clp_s/search/SearchTelemetry.cpp
```diff
@@ -0,0 +1,441 @@
...
+        uint64_t value
+) -> void;
+
+auto to_nostd_string_view(std::string_view sv) -> opentelemetry::nostd::string_view {
+    return opentelemetry::nostd::string_view{sv.data(), sv.size()};
+}
+
+auto get_env(char const* name) -> char const* {
+    // The environment is only read during single-threaded search setup.
+    // NOLINTNEXTLINE(concurrency-mt-unsafe)
```
ShangDanLuXian: I remember that there is no multi-threading for clp-s, so is this fine?

ShangDanLuXian:
Need to remove actual query

docs/src/dev-docs/components-core/index.md
```diff
@@ -50,6 +50,7 @@ The task will download, build, and install (within the build directory) the foll
 | [mongo-cxx-driver](https://github.com/mongodb/mongo-cxx-driver)       | r4.1.1         |
 | [msgpack-cxx](https://github.com/msgpack/msgpack-c/tree/cpp_master)   | v7.0.0         |
 | [nlohmann_json](https://github.com/nlohmann/json)                     | v3.11.3        |
+| [opentelemetry-cpp](https://github.com/open-telemetry/opentelemetry-cpp) | v1.27.0        |
```
gibber9809, line 53: nit: should reformat the rest of this table so that the columns line up again.

components/core/CMakeLists.txt
```diff
@@ -210,6 +210,13 @@ if(CLP_NEED_NLOHMANN_JSON)
     endif()
 endif()
 
+if(CLP_NEED_OPENTELEMETRY_CPP)
+    find_package(opentelemetry-cpp 1.27.0 CONFIG REQUIRED COMPONENTS api)
```
gibber9809: Do we need 1.27.0 specifically, or would other versions work as well? Usually I think we just use `find_package()` without a version when we can, just so that it's a bit easier to build the package for people that set up dependencies themselves.

components/core/cmake/Options/options.cmake
```diff
@@ -166,6 +166,7 @@ function(set_clp_binaries_dependencies)
         CLP_NEED_MONGOCXX
         CLP_NEED_MSGPACKCXX
         CLP_NEED_NLOHMANN_JSON
+        CLP_NEED_OPENTELEMETRY_CPP
```
gibber9809:
```suggestion
```
I think this `CLP_NEED_OPENTELEMETRY_CPP` flag should only be needed for the `clp_s::search` library, since it seems like the only code that directly uses open telemetry (`SearchTelemetry`) is part of the `clp_s::search` library?

components/core/cmake/Options/options.cmake
```diff
@@ -207,6 +208,7 @@ function(set_clp_tests_dependencies)
         CLP_NEED_MARIADB
         CLP_NEED_MONGOCXX
         CLP_NEED_NLOHMANN_JSON
+        CLP_NEED_OPENTELEMETRY_CPP
```
gibber9809:
```suggestion
```
ditto, same reasoning.

components/core/src/clp_s/ArchiveReader.hpp
```diff
@@ -166,6 +166,14 @@ class ArchiveReader {
      */
     [[nodiscard]] std::vector<int32_t> const& get_schema_ids() const { return m_schema_ids; }
 
+    /**
+     * @param schema_id
+     * @return The number of messages stored under the given schema in the archive.
+     */
+    [[nodiscard]] auto get_schema_num_messages(int32_t schema_id) const -> uint64_t {
```
gibber9809:
```suggestion
    [[nodiscard]] auto get_num_messages_for_schema(int32_t schema_id) const -> uint64_t {
```

components/core/src/clp_s/ArchiveReader.hpp
```diff
@@ -166,6 +166,14 @@ class ArchiveReader {
      */
     [[nodiscard]] std::vector<int32_t> const& get_schema_ids() const { return m_schema_ids; }
 
+    /**
+     * @param schema_id
+     * @return The number of messages stored under the given schema in the archive.
```
gibber9809, line 171:
```suggestion
     * @return The number of messages stored under the given schema in the archive.
     * @throw std::out_of_range if `schema_id` is not found in the schema metadata.
```

components/core/src/clp_s/search/Output.hpp
```diff
@@ -30,13 +31,15 @@ class Output {
            std::shared_ptr<ast::Expression> const& expr,
            std::shared_ptr<ArchiveReader> const& archive_reader,
            std::unique_ptr<OutputHandler> output_handler,
-           bool ignore_case)
+           bool ignore_case,
+           SearchTelemetry* telemetry = nullptr)
```
gibber9809: Could we make this an `std::shared_ptr`, just to make the ownership more clear?

components/core/src/clp_s/search/SearchTelemetry.hpp
```diff
@@ -0,0 +1,116 @@
...
+#include <memory>
+#include <optional>
+#include <string>
+#include <string_view>
+
+#include <clp_s/Defs.hpp>
+#include <clp_s/search/ast/Expression.hpp>
+
+namespace clp_s::search {
+constexpr std::string_view cTerminationStageRecordScan{"record_scan"};
```
gibber9809: Is `record_scan` used? Seems like we have it as a default, but we actually use `ert_scan` when recording termination after scanning records.

components/core/src/clp_s/search/SearchTelemetry.cpp
```diff
@@ -0,0 +1,405 @@
...
+    return value > cMaxInt64 ? std::numeric_limits<int64_t>::max() : static_cast<int64_t>(value);
+}
+
+auto descriptor_has_wildcard(ColumnDescriptor const& column) -> bool {
+    return std::ranges::any_of(
+            column.descriptor_begin(),
+            column.descriptor_end(),
+            [](auto const& descriptor) { return descriptor.wildcard(); }
+    );
+}
```
gibber9809: I think this is equivalent to `column.is_unresolved_descriptor()`, so let's use that directly instead.

components/core/src/clp_s/search/SearchTelemetry.cpp
gibber9809: Should be fine, yeah. Could we also use `std::get_env` directly instead of using this wrapper function?

components/core/src/clp_s/search/SearchTelemetry.cpp
```diff
@@ -0,0 +1,405 @@
...
+    return std::ranges::find(cTruthyEnvValues, value) != cTruthyEnvValues.end();
+}
+
+auto set_string_attribute(
+        opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> const& span,
+        opentelemetry::nostd::string_view key,
+        std::string_view value
+) -> void {
+    span->SetAttribute(key, to_nostd_string_view(value));
+}
```
gibber9809: I think `span->SetAttribute(key, to_nostd_string_view(value));` is already pretty readable. Could we get rid of this helper and just write that directly?

components/core/src/clp_s/search/SearchTelemetry.cpp
```diff
@@ -0,0 +1,405 @@
...
+    }
+}
+
+auto set_uint64_attribute(
+        opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> const& span,
+        opentelemetry::nostd::string_view key,
+        uint64_t value
+) -> void {
+    span->SetAttribute(key, to_int64_attribute(value));
+}
```
gibber9809: Same as with setting strings, could we use `span->SetAttribute` directly?

components/core/src/clp_s/search/SearchTelemetry.cpp
```diff
@@ -0,0 +1,405 @@
...
+            set_string_attribute(m_span, "clp.search.task_id", telemetry.task_id);
+        }
+        m_span->SetAttribute("clp.search.query_hash", static_cast<int64_t>(telemetry.query_hash));
+        if (telemetry.query.has_value()) {
+            set_string_attribute(m_span, "clp.search.query", *telemetry.query);
+        }
+
+        std::array<std::pair<opentelemetry::nostd::string_view, uint64_t>, 17> const
+                uint64_attributes{
+                        {{"clp.query_shape.column_types.pure_wildcard",
```
gibber9809: Would it be possible to switch to using `constexpr std::string_view` constants for all of these attribute names?

components/core/src/clp_s/search/SearchTelemetry.cpp
```diff
@@ -0,0 +1,405 @@
...
+            column.descriptor_begin(),
+            column.descriptor_end(),
+            [](auto const& descriptor) { return descriptor.wildcard(); }
+    );
+}
+
+auto add_column_shape(SearchTelemetry& telemetry, ColumnDescriptor const& column) -> void {
+    if (column.is_pure_wildcard()) {
+        ++telemetry.column_shape_metrics.pure_wildcard;
+    } else if (descriptor_has_wildcard(column)) {
```
gibber9809:
```suggestion
    } else if (column.is_unresolved_descriptor()) {
```

gibber9809 (changes requested):
Leaving a few initial comments. I still need to finish looking at `SearchTelemetry.{cpp,hpp}`, `Output.{cpp,hpp}`, and `clp-s.cpp` -- I plan on finishing the rest of this review tomorrow morning.

components/core/src/clp_s/search/SearchTelemetry.cpp
```diff
@@ -0,0 +1,405 @@
...
+    telemetry.query_hash = static_cast<uint64_t>(XXH3_64bits(query.data(), query.size()));
+    if (char const* const query_id{get_env("CLP_QUERY_ID")}; nullptr != query_id) {
+        telemetry.query_id = query_id;
+    }
+    if (char const* const task_id{get_env("CLP_TASK_ID")}; nullptr != task_id) {
+        telemetry.task_id = task_id;
+    }
+    if (is_env_var_enabled("CLP_TELEMETRY_INCLUDE_QUERY")) {
+        telemetry.query = std::string{query};
+    }
```
gibber9809: Per sidebar, let's get rid of telemetry.query and `CLP_TELEMETRY_INCLUDE_QUERY`

components/core/src/clp_s/search/SearchTelemetry.hpp
```diff
@@ -0,0 +1,116 @@
...
+        uint64_t string_with_wildcard{};
+        uint64_t integer{};
+        uint64_t floating_point{};
+        uint64_t null{};
+        uint64_t exact_match{};
+        uint64_t range{};
+        uint64_t exists{};
+    };
+
+    std::string archive_id;
```
gibber9809: I think it's a bit borderline whether `archive_id` counts as private information from a cluster, so not sure whether we should include it or not. What do you think @kirkrodrigues?

Also worth noting that someone that has access to the cluster could figure out the archive ID based on the query ID and task ID by looking at the query worker logs.

components/core/src/clp_s/search/SearchTelemetry.cpp
```diff
@@ -0,0 +1,405 @@
...
+    if (column.is_pure_wildcard()) {
+        ++telemetry.column_shape_metrics.pure_wildcard;
+    } else if (descriptor_has_wildcard(column)) {
+        ++telemetry.column_shape_metrics.some_wildcard;
+    } else {
+        ++telemetry.column_shape_metrics.no_wildcard;
+    }
+}
+
+auto add_predicate_type(SearchTelemetry& telemetry, FilterExpr const& filter) -> void {
```
gibber9809: It seems like we're recording the shape of the query before we run `preprocess_query`.

This means that `NarrowTypes` and `ConvertToExists` haven't been run on the query AST yet, so e.g. EXISTS operations won't appear in the query AST at this stage.

Do you think it would make sense to record the query shape after running that preprocessing? It would mean that some predicates get repeated (because of how we expand the query AST in `OrOfAndForm`), but then the switch block below should pick up EXISTS correctly, and the block below could rely on `ColumnDescriptor::matches_type` (except for checking whether a string predicate contains a wildcard).

components/core/src/clp_s/clp-s.cpp
```diff
@@ -132,6 +137,13 @@ bool search_archive(
         int reducer_socket_fd
 ) {
     auto const& query = command_line_arguments.get_query();
+    auto telemetry{collect_query_shape_metrics(
+            expr,
+            command_line_arguments.get_search_begin_ts(),
+            command_line_arguments.get_search_end_ts()
+    )};
+    populate_query_context(telemetry, query, archive_reader->get_archive_id());
+    SearchTelemetrySpan telemetry_span;
```
gibber9809: It seems like the package allows disabling telemetry using the `CLP_DISABLE_TELEMETRY` flag. As well, I'm not sure if we want telemetry enabled for cli users by default.

I think it might make sense to add a `--enable-telemetry` flag to the `CommandLineArguments` class that we set from the package side when `CLP_DISABLE_TELEMETRY` is false?

Either way, we should probably make all of this telemetry recording optional (e.g. by having `telemetry_span` be an `std::shared_ptr<SearchTelemetrySpan>` that is null when unused.

components/core/src/clp_s/search/SearchTelemetry.hpp
```diff
@@ -0,0 +1,116 @@
...
+    uint64_t num_schemas_with_matches{};
+
+    std::string_view termination_stage{cTerminationStageRecordScan};
+};
+
+/**
+ * Records the telemetry for one archive search as an OpenTelemetry span: the span starts on
+ * construction and ends on destruction, and is populated via `set_telemetry` or `set_error`.
+ */
+class SearchTelemetrySpan {
```
gibber9809, line 88: It seems like having `SearchTelemetry` by one huge struct not owned by `SearchTelemetrySpan` is making this interface a bit awkward -- properties in `SearchTelemetry` end up getting set all over the place, and every return path needs an explicit `set_telemetry` before exiting.

Could we refactor this code into something a bit more modular?

Specifically, I think we could make this code a lot simpler by breaking `SearchTelemetry` into several structs (e.g. `QueryShapeMetrics`, etc.), grouped together by when we can produce them in the code, have corresponding methods in `SearchTelemetry` span for setting those attributes as soon as we know them (e.g. `set_query_shape_metrics`, `set_termination_stage`, etc.). And then have utility functions (e.g. `create_query_shape_metrics`, etc.) as they make sense throughout the code?

That should eliminate the need to pass around both a `SearchTelemetry` struct as well as a `SearchTelemetrySpan` everywhere, eliminate the need to always call `set_telemetry` before a return (since we'll be setting properties as soon as we know them), and make it easier to follow in the code what properties are being set when.

components/core/src/clp_s/search/SearchTelemetry.hpp
```diff
@@ -0,0 +1,116 @@
...
+    SearchTelemetrySpan(SearchTelemetrySpan&&) = delete;
+    auto operator=(SearchTelemetrySpan const&) -> SearchTelemetrySpan& = delete;
+    auto operator=(SearchTelemetrySpan&&) -> SearchTelemetrySpan& = delete;
+
+    // Destructor
+    ~SearchTelemetrySpan();
+
+    // Methods
+    auto set_error(std::string_view message) -> void;
+    auto set_telemetry(SearchTelemetry const& telemetry) -> void;
```
gibber9809: Missing docstrings? Granted we should probably get rid of `set_telemetry` per my other comment -- `set_error` seems fine though.

gibber9809 (changes requested):
I think most of the pieces are here, but we need some refactoring in this PR. Probably the most important comment is the longer one I left in `SearchTelemetry.hpp` -- I think the change I'm proposing there should make the code a lot simpler/easier to follow.

Haven't reviewed Output.{cpp,hpp} in this pass, will leave that until after the high-level refactoring comments I left have been addressed.

components/core/src/clp_s/search/SearchTelemetry.cpp
ShangDanLuXian: I recorded both raw query and after preprocessing

components/core/src/clp_s/search/SearchTelemetry.hpp
```diff
@@ -0,0 +1,152 @@
...
+    // Destructor
+    ~SearchTelemetrySpan();
+
+    // Methods
+    /**
+     * Records the query-identity attributes: a non-reversible hash of the query and, when present
+     * in the environment, the scheduler's `query_id`/`task_id`.
+     * @param query The raw query string.
+     */
+    auto set_query_context(std::string_view query) -> void;
```
ShangDanLuXian, line 123: Do we need to collect this one? I remember we asked Nathan about this, but I forgot the detail lol

gibber9809: Yeah we still need this -- all that proxy will be adding is a deployment ID, but we'll still need the information this function adds to distinguish queries within a deployment.

components/core/src/clp_s/search/SearchTelemetry.hpp
```diff
@@ -0,0 +1,152 @@
...
+     * @param metrics
+     */
+    auto set_query_shape_metrics(QueryShapeMetrics const& metrics) -> void;
+
+    /**
+     * Records the query-shape metrics of the preprocessed form of the query (the form that gets
+     * evaluated, in which e.g. EXISTS predicates appear).
+     * @param metrics
+     */
+    auto set_preprocessed_query_shape_metrics(QueryShapeMetrics const& metrics) -> void;
```
gibber9809: I think it would be fine to just have the preprocessed version, since if the query terminates before preprocessing it should pretty much always just mean that the original KQL query was malformed or there's some other error.

components/core/src/clp_s/clp-s.cpp
```diff
@@ -133,6 +145,31 @@ bool search_archive(
...
+        if (nullptr != telemetry_span) {
+            telemetry_span->set_error(message);
+        }
+    };
+    auto const record_termination = [&](std::string_view termination_stage) {
+        if (nullptr == telemetry_span) {
+            return;
+        }
+        SearchResultMetrics metrics;
+        populate_total_archive_records(archive_reader, metrics);
```
gibber9809: I'm a tiny bit worried about whether this adds meaningful overhead on our early-return paths when reading a SFA (particularly over the network). Could you do a benchmark testing if there's any performance difference on a query that fails time range matching and a query that fails schema matching with telemetry on and off?

components/core/src/clp_s/clp-s.cpp
```diff
@@ -133,6 +145,31 @@ bool search_archive(
 ) {
     auto const& query = command_line_arguments.get_query();
 
+    std::shared_ptr<SearchTelemetrySpan> telemetry_span;
+    if (is_search_telemetry_enabled(command_line_arguments.get_enable_telemetry())) {
+        telemetry_span = std::make_shared<SearchTelemetrySpan>();
```
gibber9809: Since part of what the span is doing is tracking overall search time for this archive I think we should be creating it before before the `archive_reader->open` call and passing the span into `search_archive`. Otherwise we're not measuring the time it takes to load to open the archive, which can be significant for archives stored on S3.

components/core/src/clp_s/clp-s.cpp
```diff
@@ -133,6 +145,31 @@ bool search_archive(
...
+                command_line_arguments.get_search_begin_ts(),
+                command_line_arguments.get_search_end_ts()
+        ));
+    }
+    auto const record_error = [&](std::string_view message) {
+        if (nullptr != telemetry_span) {
+            telemetry_span->set_error(message);
+        }
+    };
+    auto const record_termination = [&](std::string_view termination_stage) {
```
gibber9809:
```suggestion
    auto const record_early_termination = [&](std::string_view termination_stage) {
```
Just to be a little bit more clear, since we only use this to record terminations before `output.filter()`.

components/core/src/clp_s/clp-s.cpp
```diff
@@ -296,7 +349,41 @@ bool search_archive(
...
+        telemetry_span->set_search_result_metrics(output.get_result_metrics());
+    }
+    return success;
+}
+
+auto populate_total_archive_records(
+        std::shared_ptr<clp_s::ArchiveReader> const& archive_reader,
+        SearchResultMetrics& metrics
+) -> void {
+    if (archive_reader->get_schema_ids().empty()) {
```
gibber9809: Since we're only using this for early termination I think we should just call `read_metadata` unconditionally.

Also, since this is only used for early termination, maybe it makes sense for this to just be part of that lambda we call in the early termination path instead of being a separate helper? If you do prefer keeping it as a separate helper though, maybe just rename it to `load_archive_metadata_and_set_total_records` then directly pass the span + call `set_search_result_metrics` in this function?

components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -742,6 +742,10 @@ CommandLineArguments::parse_arguments(int argc, char const** argv) {
                 "ignore-case,i",
                 po::bool_switch(&m_ignore_case),
                 "Ignore case distinctions between values in the query and the compressed data"
+            )(
+                "enable-telemetry",
+                po::bool_switch(&m_enable_telemetry),
+                "Record OpenTelemetry telemetry describing the query and how it was searched"
```
gibber9809:
```suggestion
                "Publish search telemetry to the OpenTelemetry endpoint specified in the"
                "CLP_TELEMETRY_ENDPOINT environment variable."
```

components/core/src/clp_s/clp-s.cpp
```diff
@@ -296,7 +349,41 @@ bool search_archive(
...
+                        error.category().name(),
+                        error.message()
+                );
+                return;
+            }
+        } catch (std::exception const& e) {
+            SPDLOG_WARN("Failed to read archive metadata for search telemetry: {}", e.what());
+            return;
+        }
+    }
```
gibber9809: Do you think it would make sense to set an error in the telemetry span as well when this fails?

components/core/src/clp_s/search/Output.cpp
```diff
@@ -90,10 +97,12 @@ bool Output::filter() {
 
     std::string message;
     auto const archive_id = m_archive_reader->get_archive_id();
+    bool searched_schema{false};
```
gibber9809:
```suggestion
    bool scanned_any_ert{false};
```
slightly more specific name?

components/core/src/clp_s/search/SearchTelemetry.hpp
```diff
@@ -0,0 +1,152 @@
...
+#include <cstdint>
+#include <memory>
+#include <optional>
+#include <string_view>
+
+#include <clp_s/Defs.hpp>
+#include <clp_s/search/ast/Expression.hpp>
+
+namespace clp_s::search {
+constexpr std::string_view cTerminationStageRecordScan{"record_scan"};
```
gibber9809:
```suggestion
```
Since I can't see the difference between this and `cTerminationStageErtScan`.

components/core/src/clp_s/search/Output.hpp
```diff
@@ -46,13 +48,29 @@ class Output {
...
+
 private:
     QueryRunner m_query_runner;
     std::shared_ptr<ArchiveReader> m_archive_reader;
     std::shared_ptr<ast::Expression> m_expr;
     std::shared_ptr<SchemaMatch> m_match;
     std::unique_ptr<OutputHandler> m_output_handler;
     bool m_should_marshal_records{true};
+    SearchResultMetrics m_result_metrics;
+    std::string_view m_termination_stage{cTerminationStageRecordScan};
```
gibber9809:
```suggestion
    std::string_view m_termination_stage{cTerminationStageErtScan};
```

components/core/src/clp_s/search/Output.cpp
```diff
@@ -102,18 +111,25 @@ bool Output::filter() {
         );
         auto& filter = m_query_runner.prepare_filter(reader);
 
+        uint64_t schema_matched_records{};
```
gibber9809: Since we seem to just be checking for this being non-zero to update metrics, maybe we can just use a boolean flag here?

components/core/src/clp_s/search/SearchTelemetry.cpp
```diff
@@ -0,0 +1,394 @@
...
+        }
+    }
+    return metrics;
+}
+
+auto is_search_telemetry_enabled(bool enable_telemetry) -> bool {
+    if (false == enable_telemetry) {
+        return false;
+    }
+    return false == is_env_var_enabled("CLP_DISABLE_TELEMETRY");
```
gibber9809: Do you think it would make sense to just rely on the explicit CLI flag instead of double-checking `CLP_DISABLE_TELEMETRY`?

gibber9809 (changes requested):
Nice work! Left another round of comments. Probably after this round it might make sense for @kirkrodrigues to take a look as well to see if he agrees with the set of metrics being collected.

Besides the other comments I left I'm a little bit borderline on gathering stats unconditionally inside of `Output.cpp`, but I think it's fine since what we're gathering there shouldn't add any real overhead + the alternatives probably make the code less readable and/or adds more overhead.

components/core/src/clp_s/search/SearchTelemetry.cpp
gibber9809: Actually just talked a bit more with Kirk, let's just rely on that command line flag.

I think we can get rid of a good chunk of code related to checking the truthyness of this env var in that case.

components/core/src/clp_s/search/SearchTelemetry.hpp
gibber9809: Talked a bit more with Kirk, could we just hash the archive ID like we do for the query? That way we can still correlate searches on the same archive if we want without worrying about leaking archive IDs. @ShangDanLuXian

components/core/src/clp_s/clp-s.cpp
ShangDanLuXian: It's a little bit tricky here - since the status is success after early termination, indicating a successful search, while set an error here may get later aggregation harder. 

I think it is fine here since total_archive_records = 0 if we cannot read the metadata - unless it is an empty archive, but is it possible?

gibber9809: It is possible to have an archive with zero records (though I think the read side actually doesn't support them correctly at the moment), but it's unlikely that we'll create any in a real package deployment.

I think I'm fine with not setting an error here though, since you're right that the search itself didn't hit any errors.

components/core/src/clp_s/clp-s.cpp
```diff
@@ -129,9 +133,43 @@ bool search_archive(
...
+                    SPDLOG_WARN(
+                            "Failed to read archive metadata for search telemetry - ({}) {}",
+                            error.category().name(),
+                            error.message()
+                    );
+                }
+            } catch (std::exception const& e) {
+                SPDLOG_WARN("Failed to read archive metadata for search telemetry - {}", e.what());
+            }
+        }
```
gibber9809:
```suggestion
        try {
            if (auto const result{archive_reader->read_metadata()}; result.has_error()) {
                auto const error{result.error()};
                SPDLOG_WARN(
                        "Failed to read archive metadata for search telemetry - ({}) {}",
                        error.category().name(),
                        error.message()
                );
            }
        } catch (std::exception const& e) {
            SPDLOG_WARN("Failed to read archive metadata for search telemetry - {}", e.what());
        }
        SearchResultMetrics metrics;
```
Since we're only using this for early termination, we won't ever have called `read_metadata` earlier, so we can go ahead and call `read_metadata` unconditionally here.

gibber9809 (changes requested):
One minor comment, but otherwise looks good!

components/core/src/clp_s/clp-s.cpp
ShangDanLuXian: There is no difference in terms of performance. 
query                   min   median     mean      p95      max  path               (errors)
early-timerange      0.0258   0.0269   0.0269   0.0278   0.0278  early:time-range   

Also this is not an important case I believe - as early termination means very fast rejection, which can be easily covered by network latency.

gibber9809:
Sorry also just realized it seems like we need to add the hash of the archive ID like [this comment](https://github.com/y-scope/clp/pull/2309#discussion_r3390804282) was asking for.

gibber9809 (dismissed):
LGTM. PR title is fine as well. Passing it off to a maintainer for final review.

components/core/src/clp_s/clp-s.cpp
```diff
@@ -148,21 +185,31 @@ bool search_archive(
...
         );
+        record_error("no authoritative timestamp column");
         return false;
     }
 
     if (expr = clp_s::search::ast::preprocess_query(expr);
         std::dynamic_pointer_cast<ast::EmptyExpr>(expr))
     {
         SPDLOG_ERROR("Query '{}' is logically false", query);
+        record_error("query is logically false");
```
davidlion: Since `record_error` is coupled with `SPDLOG_ERROR` in most places with almost identical messages (which makes sense to me fwiw), should we just have a `record_error_and_log` method that will call `SPDLOG_ERROR` internally with the same message?

components/core/src/clp_s/search/SearchTelemetry.hpp
davidlion, line 1: Somewhat minor, but we did formalize the category comments so it would be good if we followed them in new files: https://docs.yscope.com/dev-guide/contrib-guides-cpp.html#declaration-order.

Double check, but the main things I see are `private` missing the data member comment and no "top-level" operator comment. (Up to you and Devin if you want to keep each comment about deleting the operators, but personally I don't think they're necessary.)

components/core/src/clp_s/search/SearchTelemetry.hpp
```diff
@@ -0,0 +1,153 @@
...
+    auto operator=(SearchTelemetrySpan const&) -> SearchTelemetrySpan& = delete;
+
+    // Delete move constructor and assignment operator
+    SearchTelemetrySpan(SearchTelemetrySpan&&) = delete;
+    auto operator=(SearchTelemetrySpan&&) -> SearchTelemetrySpan& = delete;
+
+    // Destructor
+    ~SearchTelemetrySpan();
+
+    // Methods
```
davidlion, line 102: Very minor, but according to https://docs.yscope.com/dev-guide/contrib-guides-general.html# we should be using alphabetical order of declarations if there isn't a clear logical order. I don't think there is a logical grouping, we we might as well use alphabetical order. This applies to the cpp file as well.

components/core/src/clp_s/search/SearchTelemetry.cpp
```diff
@@ -0,0 +1,390 @@
...
+        for (auto it{node->op_begin()}; it != node->op_end(); ++it) {
+            if (auto const child{std::dynamic_pointer_cast<Expression>(*it)}; nullptr != child) {
+                to_visit.emplace_back(child);
+            }
+        }
+    }
+}
+}  // namespace
+
+class SearchTelemetrySpan::Impl {
```
davidlion, line 201: Similar to https://github.com/y-scope/clp/pull/2309/changes#r3405106171 double check the category comments. At a glance at least `// Constructors` and `// Methods` are missing.

davidlion (commented):
Mostly minor style stuff. I do think having a `record_error_and_log` method would be nice as I think that would follow what people want in the majoriy of cases.

components/core/src/clp_s/clp-s.cpp
ShangDanLuXian: Added a record_error_and_log helper that takes both the log text and the error recorded on the span. I couldn't make the two messages identical though — the original SPDLOG_ERROR messages include private user data (e.g. the raw query, projection column names, exception text) that we deliberately don't collect in telemetry, so the span gets a short generic message while the log keeps the full detail.

components/core/src/clp_s/search/Output.hpp
```diff
@@ -17,6 +18,7 @@
 #include "OutputHandler.hpp"
 #include "QueryRunner.hpp"
 #include "SchemaMatch.hpp"
+#include "SearchTelemetry.hpp"
```
davidlion:
```suggestion
#include <clp_s/search/SearchTelemetry.hpp>
```

Sorry, the coding guidelines are still a bit fragmented, but we're (slowly) migrating to this style of include statements so new includes should follow this (https://app.notion.com/p/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?source=copy_link#25304e4d9e6b801481fbe508b428af20).

components/core/src/clp_s/search/SearchTelemetry.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+    auto set_termination_stage(std::string_view termination_stage) -> void {
+        m_span->SetAttribute(
+                to_nostd_string_view(cAttrTerminationStage),
+                to_nostd_string_view(termination_stage)
+        );
+    }
+
+private:
+    // Variables
```
davidlion:
```suggestion
    // Data members
```
(https://docs.yscope.com/dev-guide/contrib-guides-cpp.html#declaration-order)

components/core/src/clp_s/search/SearchTelemetry.hpp
```diff
@@ -0,0 +1,154 @@
...
+     *
+     * @param termination_stage One of the `cTerminationStage*` constants.
+     */
+    auto set_termination_stage(std::string_view termination_stage) -> void;
+
+private:
+    // Types
+    class Impl;
+
+    // Variables
```
davidlion:
```suggestion
    // Data members
```

(https://docs.yscope.com/dev-guide/contrib-guides-cpp.html#declaration-order)

components/core/src/clp_s/search/SearchTelemetry.hpp
```diff
@@ -0,0 +1,154 @@
...
+ * @param expr The parsed query AST.
+ * @param search_begin_ts The lower bound of the search time range, if any.
+ * @param search_end_ts The upper bound of the search time range, if any.
+ * @return The query-shape metrics.
+ */
+[[nodiscard]] auto create_query_shape_metrics(
+        std::shared_ptr<ast::Expression> const& expr,
+        std::optional<epochtime_t> search_begin_ts,
+        std::optional<epochtime_t> search_end_ts
+) -> QueryShapeMetrics;
```
davidlion: This should be before `// Data members` (variables) and is also missing `// Methods`.

components/core/src/clp_s/clp-s.cpp
```diff
@@ -41,6 +42,7 @@
 #include "search/OutputHandler.hpp"
 #include "search/Projection.hpp"
 #include "search/SchemaMatch.hpp"
+#include "search/SearchTelemetry.hpp"
```
davidlion:
```suggestion
#include <clp_s/search/SearchTelemetry.hpp>
```

davidlion (changes requested):
Extremely minor nits I missed in the last review (sorry).

components/core/src/clp_s/search/SearchTelemetry.hpp
ShangDanLuXian: Just realized I forgot to use the factory method.

components/core/src/clp_s/search/SearchTelemetry.cpp
```diff
@@ -0,0 +1,388 @@
...
+using clp_s::search::ast::Expression;
+using clp_s::search::ast::FilterExpr;
+using clp_s::search::ast::FilterOperation;
+using clp_s::search::ast::OrExpr;
+using opentelemetry::trace::StatusCode;
+
+namespace clp_s::search {
+namespace {
+constexpr std::string_view cTracerName{"clp_s.query"};
+constexpr std::string_view cSearchArchiveSpanName{"clp_s.query.archive"};
```
gibber9809:
```suggestion
constexpr std::string_view cTracerName{"clp.query"};
constexpr std::string_view cSearchArchiveSpanName{"clp.query.archive"};
```
Per sidebar, for consistency with everything else.

components/core/src/clp_s/search/SearchTelemetry.cpp
```diff
@@ -0,0 +1,388 @@
...
+    auto const operand{filter.get_operand()};
+    if (nullptr == operand) {
+        return;
+    }
+
+    std::string string_value;
+    int64_t int_value{};
+    double float_value{};
+    if (operand->as_clp_string(string_value, op) || operand->as_var_string(string_value, op)) {
+        if (std::string::npos == string_value.find('*')) {
```
gibber9809: Sorry, should have caught this earlier, but we should be using `clp_s::search::ast::has_unescaped_wildcards(string_value)` from `<clp_s/search/ast/SearchUtils.hpp>` instead of `find('*')` here.

gibber9809 (changes requested):
One detail I should have caught earlier + one missing rename, but otherwise this looks ready to merge.

gibber9809 (approved):

davidlion (approved):

components/core/src/clp_s/search/SearchTelemetry.cpp
```diff
@@ -0,0 +1,389 @@
...
+}
+}  // namespace
+
+class SearchTelemetrySpan::Impl {
+public:
+    // Constructors
+    Impl()
+            : m_span{opentelemetry::trace::Provider::GetTracerProvider()
+                             ->GetTracer(to_nostd_string_view(cTracerName))
+                             ->StartSpan(to_nostd_string_view(cSearchArchiveSpanName))},
```
rishikeshdevsot, line 207: @gibber9809 Is there a reason why we chose to have a root span instead of having a trace parent for the span? The  `parent` argument of this function: https://opentelemetry-cpp.readthedocs.io/en/latest/otel_docs/structopentelemetry_1_1trace_1_1StartSpanOptions.html allows you to set a trace parent, so you can visualize all the spans nested under a single trace parent. I would assume we would want to place all the spans for a single query under a single parent (For example, when `clp s` is invoked with multiple paths, you want the queries for all the archives in a single traces instead of separate)

I am asking because I am adding OpenTelemetry traces to trace search/compression jobs in the package, and I want to keep the design consistent with the changes here.

PR #2315, opened 2026-06-09
feat(clp-s::search): Extend ColumnScan to support timestamp types.

ShangDanLuXian:
# Description

Add timestamp filter support for both `TimestampColumnReader` and the deprecated date-string reader.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **New Features**
  * Timestamp-aware filtering for searches to target timestamp values.
  * Optional legacy date-string support to preserve queries against older data.

* **Bug Fixes**
  * Improved handling of null literals in filters to prevent unintended evaluation and match documented behaviour.
  * Correct parsing/selection for timestamp literals across new and legacy date readers.


components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -533,9 +619,33 @@ auto ColumnScan::build_filter(
...
+            int64_t operand_value{};
+            if (false == operand->as_int(operand_value, operation)) {
+                return bitmap;
+            }
+            if (nullptr != deprecated_datestring_reader
+                && column_id == deprecated_datestring_reader->get_id())
+            {
+                return build_deprecated_datestring_filter(
+                        m_num_messages,
+                        deprecated_datestring_reader,
```
gibber9809:
```suggestion
                        *deprecated_datestring_reader,
```
May as well dereference and pass a `DeprecatedDateStringColumnReader&`, since we've already checked that it's non-null.

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -115,6 +115,38 @@ clp_string_matches(ClpStringColumnReader* reader, clp::Query const& query, uint6
...
+ * Builds a bitmap for a filter over a deprecated date-string column.
+ * @param num_messages Number of messages represented by the bitmap.
+ * @param reader Deprecated date-string column reader to scan.
+ * @param operation Filter operation to apply.
+ * @param operand Operand from the filter expression, encoded as epoch time.
+ * @return A bitmap indexed by message number, with nonzero entries for matching messages.
+ */
+[[nodiscard]] auto build_deprecated_datestring_filter(
+        uint64_t num_messages,
+        DeprecatedDateStringColumnReader* reader,
```
gibber9809:
```suggestion
        DeprecatedDateStringColumnReader& reader,
```

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -254,13 +286,49 @@ clp_string_matches(ClpStringColumnReader* reader, clp::Query const& query, uint6
...
+    for (uint64_t message_index{0}; message_index < num_messages; ++message_index) {
+        auto const value = reader->get_encoded_time(message_index);
+        bitmap[message_index] = compare(operation, value, operand) ? 1 : 0;
+    }
+    return bitmap;
+}
+
+[[nodiscard]] auto build_deprecated_datestring_filter(
+        uint64_t num_messages,
+        DeprecatedDateStringColumnReader* reader,
```
gibber9809:
```suggestion
        DeprecatedDateStringColumnReader& reader,
```

components/core/src/clp_s/search/ColumnScan.cpp
```diff
@@ -254,13 +286,49 @@ clp_string_matches(ClpStringColumnReader* reader, clp::Query const& query, uint6
...
+        FilterOperation operation,
+        int64_t operand
+) -> ColumnScan::Bitmap {
+    ColumnScan::Bitmap bitmap(num_messages, 0);
+    for (uint64_t message_index{0}; message_index < num_messages; ++message_index) {
+        auto const value = reader->get_encoded_time(message_index);
```
gibber9809:
```suggestion
        auto const value = reader.get_encoded_time(message_index);
```

gibber9809 (changes requested):
Some minor nits, but otherwise looks good. Speedup also seems to be pretty good -- some searches over a narrow time range on the mongodb dataset are about 3x faster on this branch. Validated that we get that speedup with both the new and old timestamp formats, and that returned results seem to be identical between this new implementation and the old row-scan implementation.

Also found what seems to be an existing bug with null filtering (#2322) while testing this PR, but the behaviour is consistent between this branch and the upstream, so no need to do anything about it here. Likely just a bug in `ConvertToExists`.

gibber9809 (approved):
LGTM! PR title is fine as well.

Issue #2026, opened 2026-02-24
docs: Temporarily remove non-ISO8601 datasets from resources documentation


junhaoliao:
## Background

After PR `#1788` was merged, CLP-S now uses stricter timestamp parsing that rejects non-ISO8601-compliant timestamp formats (e.g., RFC 2822 / RFC 822 formats). Attempting to compress logs with such timestamps now produces errors.

## Task

Temporarily remove datasets with non-ISO8601-compliant timestamps (e.g., the "postgresql" dataset) from the resources documentation at:
https://github.com/y-scope/clp/blob/a456dc914f26e85344e5ffe46471a45760fe804b/docs/src/user-docs/resources-datasets.md

These datasets can be restored once RFC 2822 / RFC 822 timestamp parsing support is added (see related feature request).

## References

- PR: https://github.com/y-scope/clp/pull/1788
- Requested by: `@junhaoliao`

PR #2325, opened 2026-06-11
feat(job-orchestration)!: Redirect worker logs to container stdout/stderr.

rishikeshdevsot:
# Description
This PR dumps all the celery worker logs directly into the container, so we can use log ingestion/aggregation tools to analyze our logs.

- Drop the `-f` flag when starting celery workers to log directly into the containers
- Dump subprocess stderr logs to container stderr after process completion in compression_task, query/utils, and reducer, so logs appear in `kubectl logs` / `docker compose logs`

> [!NOTE]
>BREAKING CHANGE: Celery worker logs are no longer written to CLP_WORKER_LOG_PATH on disk. Operators relying on worker.log files must now use container log drivers (docker compose logs/kubectl logs).

# Checklist

  breaking change.

# Validation performed
1. Ran a compression job on unstructured logs without specifying `--unstructured` and observed the following in the compression worker's logs
```
[2026-06-11 21:43:35,307: ERROR/ForkPoolWorker-16] job_orchestration.executor.compress.celery_compress.compress[65e05254-8702-4c57-85f2-5665935da167]: Contents of compression-job-1-task-1-stderr.log:
2026-06-11T21:43:35.120+00:00 [error] Direct ingestion of unstructured logtext is not supported from input /mnt/logs/tmp/test.log
2026-06-11T21:43:35.269+00:00 [error] Encountered error while parsing inpu
```
2. Ran a query `abc` and observed the following in the query worker's logs:
```
[2026-06-11 21:43:40,302: ERROR/ForkPoolWorker-16] job_orchestration.executor.query.fs_search_task.search[1b8becb6-4cde-42f5-b763-eddbc87a0b48]: Contents of 1-clo.log:
2026-06-11T21:43:40.298+00:00 [info] No matching schemas for query 'abc'
```
3. Forcefully killed the reducer process in the container and observed reducer logs dumped in the container logs:
```
2026-06-11 21:53:23,589 reducer [INFO] Contents of reducer-15.log:
2026-06-11T21:41:47.740+00:00 [info] Starting on host reducer port 14024
2026-06-11T21:41:47.740+00:00 [info] Acceptor socket listening successfully
2026-06-11T21:41:47.740+00:00 [info] Waiting for job...
```
4. Ran a compression job with `--unstructured` flag on structured logs and observed the log converter error:
```
[2026-06-15 20:32:24,110: ERROR/ForkPoolWorker-16] job_orchestration.executor.compress.celery_compress.compress[c6fc1f6f-dd92-46d8-b677-eea0ea6ec620]: Contents of compression-job-5-task-3-conversion-stderr.log:
2026-06-15T20:32:24.107+00:00 [error] Received input that was not unstructured logtext: /mnt/logs/home/ridevsot/YScope/Projects/clp-worktrees/clp-upstream-2/build/clp-package/var/log/api_server/api_server.log.2026-06-15-20.
```

## Summary by CodeRabbit

* **Improvements**
  * Stderr output from compression and query tasks is now captured and logged when non-empty.
  * Reducer shutdown now reports reducer log contents when available.
  * Worker services now send logs to standard output for simpler log collection.
  * Search and stream extraction failure messages now direct users to container logs for details.
* **Breaking Changes**
  * `QueryTaskResult` no longer includes `error_log_path`, and related error-log propagation was removed.

### DIFF
--- a/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
+++ b/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
@@ -559,7 +559,6 @@ def run_clp(
     finally:
         cleanup_temporary_files()
         stderr_log_file.close()
-        # Dump the stderr log so it's visible in container logs
         if stderr_log_path.stat().st_size > 0:
             logger.error(
                 f"Contents of {stderr_log_path.name}:\n"
--- a/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
+++ b/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
@@ -4,6 +4,7 @@ import os
 import pathlib
 import shutil
 import subprocess
+import sys
 from contextlib import closing
 from typing import Any
 
@@ -559,6 +560,13 @@ def run_clp(
     finally:
         cleanup_temporary_files()
         stderr_log_file.close()
+        # Dump the stderr log to sys.stderr so it's visible in container logs
+        if stderr_log_path.stat().st_size > 0:
+            sys.stderr.write(
+                f"--- Contents of {stderr_log_path.name} ---\n"
+                f"{stderr_log_path.read_text()}"
+                f"--- End of {stderr_log_path.name} ---\n"
+            )
 
     worker_output = {
         "total_uncompressed_size": total_uncompressed_size,
--- a/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
+++ b/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
@@ -446,7 +446,7 @@ def run_clp(
         )
         conversion_return_code = conversion_proc.wait()
     conversion_log_file.close()
-        
+
     try:
         if conversion_return_code != 0:
             logger.error(
@@ -458,14 +458,17 @@ def run_clp(
                 "error_message": f"Check logs in {conversion_log_path}",
             }
             log_file_contents(logger, conversion_log_path)
-            
+
             return CompressionTaskStatus.FAILED, worker_output
 
         # Start compression
         logger.debug("Compressing...")
         compression_successful = False
         proc = subprocess.Popen(
-            compression_cmd, stdout=subprocess.PIPE, stderr=compression_log_file, env=compression_env
+            compression_cmd,
+            stdout=subprocess.PIPE,
+            stderr=compression_log_file,
+            env=compression_env,
         )
 
         # Compute the total amount of data compressed
--- a/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
+++ b/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
@@ -429,18 +429,24 @@ def run_clp(
                     converted_inputs_dir,
                 )
 
-    # Open stderr log file
-    stderr_log_path = logs_dir / f"{instance_id_str}-stderr.log"
-    stderr_log_file = open(stderr_log_path, "w")
+    # Open log files
+    compression_log_path = logs_dir / f"{instance_id_str}-stderr.log"
+    compression_log_file = open(compression_log_path, "w")
+    conversion_log_path = logs_dir / f"{instance_id_str}-conversion-stderr.log"
+    conversion_log_file = open(conversion_log_path, "w")
 
     conversion_return_code = 0
     if conversion_cmd is not None:
         logger.debug("Execute log-converter with command: %s", conversion_cmd)
         conversion_proc = subprocess.Popen(
-            conversion_cmd, stdout=subprocess.DEVNULL, stderr=stderr_log_file, env=conversion_env
+            conversion_cmd,
+            stdout=subprocess.DEVNULL,
+            stderr=conversion_log_file,
+            env=conversion_env,
         )
         conversion_return_code = conversion_proc.wait()
-
+    conversion_log_file.close()
+        
     try:
         if conversion_return_code != 0:
             logger.error(
@@ -449,15 +455,17 @@ def run_clp(
             worker_output = {
                 "total_uncompressed_size": 0,
                 "total_compressed_size": 0,
-                "error_message": f"Check logs in {stderr_log_path}",
+                "error_message": f"Check logs in {conversion_log_path}",
             }
+            log_file_contents(logger, conversion_log_path)
+            
             return CompressionTaskStatus.FAILED, worker_output
 
         # Start compression
         logger.debug("Compressing...")
         compression_successful = False
         proc = subprocess.Popen(
-            compression_cmd, stdout=subprocess.PIPE, stderr=stderr_log_file, env=compression_env
+            compression_cmd, stdout=subprocess.PIPE, stderr=compression_log_file, env=compression_env
         )
 
         # Compute the total amount of data compressed
@@ -536,7 +544,7 @@ def run_clp(
                             subprocess.run(
                                 indexer_cmd,
                                 stdout=subprocess.DEVNULL,
-                                stderr=stderr_log_file,
+                                stderr=compression_log_file,
                                 check=True,
                                 env=indexer_env,
                             )
@@ -559,8 +567,8 @@ def run_clp(
         logger.debug("Compressed.")
     finally:
         cleanup_temporary_files()
-        stderr_log_file.close()
-        log_file_contents(logger, stderr_log_path)
+        compression_log_file.close()
+        log_file_contents(logger, compression_log_path)
 
     worker_output = {
         "total_uncompressed_size": total_uncompressed_size,
@@ -571,7 +579,7 @@ def run_clp(
         return CompressionTaskStatus.SUCCEEDED, worker_output
     error_msgs = []
     if compression_successful is False:
-        error_msgs.append(f"See logs {stderr_log_path}")
+        error_msgs.append(f"See logs {compression_log_path}")
     if s3_error is not None:
         error_msgs.append(s3_error)
     worker_output["error_message"] = "\n".join(error_msgs)
--- a/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
+++ b/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
@@ -4,7 +4,6 @@ import os
 import pathlib
 import shutil
 import subprocess
-import sys
 from contextlib import closing
 from typing import Any
 
@@ -560,12 +559,11 @@ def run_clp(
     finally:
         cleanup_temporary_files()
         stderr_log_file.close()
-        # Dump the stderr log to sys.stderr so it's visible in container logs
+        # Dump the stderr log so it's visible in container logs
         if stderr_log_path.stat().st_size > 0:
-            sys.stderr.write(
-                f"--- Contents of {stderr_log_path.name} ---\n"
+            logger.error(
+                f"Contents of {stderr_log_path.name}:\n"
                 f"{stderr_log_path.read_text()}"
-                f"--- End of {stderr_log_path.name} ---\n"
             )
 
     worker_output = {
--- a/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
+++ b/components/job-orchestration/job_orchestration/executor/compress/compression_task.py
@@ -32,6 +32,7 @@ from clp_py_utils.s3_utils import (
 )
 from clp_py_utils.sql_adapter import SqlAdapter
 
+from job_orchestration.executor.utils import log_file_contents
 from job_orchestration.scheduler.constants import CompressionTaskStatus
 from job_orchestration.scheduler.job_config import (
     ClpIoConfig,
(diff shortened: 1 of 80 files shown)


components/job-orchestration/job_orchestration/executor/query/utils.py
```diff
@@ -98,6 +99,12 @@ def sigterm_handler(_signo, _stack_frame):
         logger.info(f"{task_name} task {task_id} completed for job {job_id}")
 
     clo_log_file.close()
+    # Dump the stderr log so it's visible in container logs
+    if clo_log_path.stat().st_size > 0:
+        logger.error(
```
20001020ycx: log level across these components are not consistent, maybe all of them should be INFO as this is some general information that helps developer to locate the log file?

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -559,6 +559,12 @@ def cleanup_temporary_files():
     finally:
         cleanup_temporary_files()
         stderr_log_file.close()
+        # Dump the stderr log so it's visible in container logs
```
20001020ycx: Personally I dont think this comment is necessary, same comment goes for other components

components/job-orchestration/job_orchestration/executor/query/utils.py
```diff
@@ -88,6 +88,7 @@ def sigterm_handler(_signo, _stack_frame):
     # stdout/stderr.
     stdout_data, _ = task_proc.communicate()
     return_code = task_proc.returncode
+
```
20001020ycx: looks like an enter hit accidentally

20001020ycx (commented):
Overall LGTM, just some nits for you to address

tools/deployment/package/docker-compose-all.yaml
```diff
@@ -311,7 +311,7 @@ services:
       CLP_HOME: "/opt/clp"
       CLP_LOGGING_LEVEL: "${CLP_COMPRESSION_WORKER_LOGGING_LEVEL:-INFO}"
       CLP_LOGS_DIR: "/var/log/compression_worker"
-      CLP_WORKER_LOG_PATH: "/var/log/compression_worker/worker.log"
+      CLP_WORKER_LOG_PATH: "/dev/stdout"
```
junhaoliao: shall we get rid of this env var entirely?

junhaoliao (changes requested):
shall we update the helm chart too?

components/job-orchestration/job_orchestration/executor/query/utils.py
rishikeshdevsot: I made it logger.error when output log for sure only contains stderr logs. In the case of reducer, the log file contains both stdout and stderr so I made that logger.info

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -890,7 +890,7 @@ async def handle_finished_search_job(
             new_job_status = QueryJobStatus.FAILED
             logger.error(
                 f"Search task job-{job_id}-task-{task_id} failed. "
-                f"Check {task_result.error_log_path} for details."
+                "Check container logs for details."
```
junhaoliao: This seems redundant given this sentence will be printed right below the binary's error output in the same console


```suggestion
```

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -982,7 +982,7 @@ async def handle_finished_stream_extraction_job(
         if not QueryTaskStatus.SUCCEEDED == task_result.status:
             logger.error(
                 f"Extraction task job-{job_id}-task-{task_id} failed. "
-                f"Check {task_result.error_log_path} for details."
+                "Check container logs for details."
```
junhaoliao: ditto

```suggestion
```

junhaoliao (changes requested):

components/job-orchestration/job_orchestration/reducer/reducer.py
```diff
@@ -93,8 +93,15 @@ def main(argv: list[str]) -> int:
     for i, reducer in enumerate(reducers):
         reducer.communicate()
```
junhaoliao: reducer.communicate() is blocking so if the first process doesn't exit, any exited processes' logs won't really be printed as they exit?

consider using concurrent.futures or similar to spawn a thread pool to wait for any reducer to exit and as soon as they do, print the logs

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
junhaoliao: this guy's logs should go to console too?

components/job-orchestration/job_orchestration/executor/query/utils.py
```diff
@@ -98,6 +98,11 @@ def sigterm_handler(_signo, _stack_frame):
         logger.info(f"{task_name} task {task_id} completed for job {job_id}")
 
     clo_log_file.close()
+    if clo_log_path.stat().st_size > 0:
```
junhaoliao: instead of repeating the logics in many different places, shall we extract a util which accepts (logger: logging.Logger, log_path: Path, level: int) as parameters?

components/job-orchestration/job_orchestration/reducer/reducer.py
rishikeshdevsot: should we handle this in another PR? seems like a pre-existing issue unrelated to the current changes?

junhaoliao: The concern is a bit different here. Previously, we only wanted to prevent the parent process (`reducer.py`) from exiting before all child processes had exited. Now, the goal is to print logs for individual child processes as they exit, which is only useful if we can do it immediately when each child exits.

In practice, the parent process will not exit or crash until the container orchestrator, such as Docker Compose or Kubernetes, requests it. Even then, this part of the logic would not be executed. So effectively, the code being added here would be dead code.

That said, I'm fine with deferring the reducer changes to a separate PR.

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
rishikeshdevsot: both compression worker and converter use the same log file right? so we do dump this file as well?

components/job-orchestration/job_orchestration/executor/query/utils.py
rishikeshdevsot: done

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
junhaoliao: hmm... would there be any delay in reporting the logs if the `clp-s` binary is stuck / compression takes an excessively long time to finish?
do you think it might be better to use a different file and dump the logs separately?

i'm fine deferring this so long we have it tracked

components/job-orchestration/job_orchestration/reducer/reducer.py
rishikeshdevsot: Added changes you requested.
 
But I guess what's still not clear to me is that if `reducer.communicate()` is blocking, then the existing logging statement `logger.info(f"reducer-{i} exited with returncode={reducer.returncode}")` in that block would also not execute right, also blocking other reducer processes? that's what I meant by pre-existing bug?

junhaoliao: > the existing logging statement logger.info(f"reducer-{i} exited with returncode={reducer.returncode}") in that block would also not execute right

right. not all of these log statements would be printed until all of them exit. a common scenario that was seen was that the connection with the query scheduler dropped and all reducers had to exit

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
rishikeshdevsot: would the `indexer` also require similar treatment then (I don't know what the indexer does, so I can't tell)

rishikeshdevsot: Seems like a simple change, I added a different file for the converter. (kept the same stderr file for indexer and compressor)

tools/deployment/package/docker-compose-all.yaml
rishikeshdevsot: done, also removed from helm charts I think

components/job-orchestration/job_orchestration/executor/query/utils.py
rishikeshdevsot: fixed

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
rishikeshdevsot: removed

components/job-orchestration/job_orchestration/executor/utils.py
```diff
@@ -1,13 +1,23 @@
...
 
 
+def log_file_contents(
+    logger: logging.Logger,
+    log_path: Path,
+    level: int = logging.ERROR,
+) -> None:
+    """Dump the contents of a log file to the logger if it's non-empty."""
+    if log_path.stat().st_size > 0:
+        logger.log(level, f"Contents of {log_path.name}:\n{log_path.read_text()}")
```
junhaoliao, line 18: lint

```suggestion
        logger.log(level, "Contents of %s:\n%s", log_path.name, log_path.read_text())
```

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -428,18 +429,24 @@ def cleanup_temporary_files():
...
-            conversion_cmd, stdout=subprocess.DEVNULL, stderr=stderr_log_file, env=conversion_env
+            conversion_cmd,
+            stdout=subprocess.DEVNULL,
+            stderr=conversion_log_file,
+            env=conversion_env,
         )
         conversion_return_code = conversion_proc.wait()
-
+    conversion_log_file.close()
+
```
junhaoliao:
```suggestion
    conversion_log_file.close()

    compression_log_path = logs_dir / f"{instance_id_str}-stderr.log"
    compression_log_file = open(compression_log_path, "w")

    try:
```

components/job-orchestration/job_orchestration/executor/compress/compression_task.py
```diff
@@ -428,18 +429,24 @@ def cleanup_temporary_files():
...
                 )
 
-    # Open stderr log file
-    stderr_log_path = logs_dir / f"{instance_id_str}-stderr.log"
-    stderr_log_file = open(stderr_log_path, "w")
+    # Open log files
+    compression_log_path = logs_dir / f"{instance_id_str}-stderr.log"
+    compression_log_file = open(compression_log_path, "w")
+    conversion_log_path = logs_dir / f"{instance_id_str}-conversion-stderr.log"
+    conversion_log_file = open(conversion_log_path, "w")
```
junhaoliao, line 436: see below change about moving the declaration closer to the usages; also the comment seem redundant
```suggestion
    conversion_log_path = logs_dir / f"{instance_id_str}-conversion-stderr.log"
    conversion_log_file = open(conversion_log_path, "w")
```

junhaoliao (dismissed):
lgtm assuming things have been tested throughly

for the title, how about:
```
feat(job-orchestration)!: Redirect worker logs to container stdout/stderr.
```

rishikeshdevsot:
Changed the title.

@junhaoliao  Anything else you think I should test beyond what I mentioned in the PR description? 

(Integration tests seem to not work right now, there's a fix for them in another PR)

junhaoliao:
> Changed the title.
> 
> @junhaoliao Anything else you think I should test beyond what I mentioned in the PR description?
> 
> (Integration tests seem to not work right now, there's a fix for them in another PR)

we can trigger all those error conditions and make sure the error logs go to the services' console. e.g.,

1. compressing file with unsupported formats / unsupported timestamp keys (see #2026 )
2. (not sure if this still causes issues) compressing non log files with the --unstructured flag so the log convertor is used
3. deleting the archives on the disks and performing a search, to cause the search binaries to fail
4. using docker commands to kill the subprocesses in the reducer container

rishikeshdevsot:
>1. compressing file with unsupported formats / unsupported timestamp keys (see https://github.com/y-scope/clp/issues/2026 )
>2. (not sure if this still causes issues) compressing non log files with the --unstructured flag so the log convertor is used
>3. deleting the archives on the disks and performing a search, to cause the search binaries to fail
>4. using docker commands to kill the subprocesses in the reducer container

All tests pass. (for 2, I see `Failed to convert unstructured log text`)

20001020ycx (dismissed):

20001020ycx (approved):

junhaoliao (approved):

PR #2326, opened 2026-06-11
feat(clp-s): Support count and count-by-time aggregation output to the results cache.

davemarco:
# Description

Currently `--count` and `--count-by-time` can only output to the reducer. This PR lets clp-s write count results
directly to MongoDB (the results cache). This PR only makes changes to the core binary, and does not add
support to the package. Idea for this set of PRs is to add more aggregation support to the binary like min/max,
unique, etc., and deal with proper result output/package support (likely with spider) at a later date.

  Changes:
* `--count` and `--count-by-time` now work with the results-cache output handler.
* New output handlers write results to MongoDB as plain inserts: each clp-s archive writes its own result doc
* A consumer sums the docs itself.
  - Doc shapes:
    - Plain count: `{archive_id, count: N}` — one doc per archive.
    - Count-by-time: `{archive_id, timestamp: <bucket>, count: N}` — one doc per bucket per archive. 

# Checklist

  breaking change.

# Validation performed
Compressed the mongodb dataset:
- `--count` returns the exact `wc -l` count
-  `--count-by-time` returns buckets whose counts exactly match the reducer path
-  A filtered query matches `grep -c` and the reducer path.

## Release Notes

* **New Features**
  * Search results can now be aggregated and stored in the results cache for both simple counts and time-bucketed counts.
* **Bug Fixes**
  * Improved command-line validation for aggregation options, including clearer errors when required options are missing or incompatible.
  * Aggregation requests are now routed consistently between reducer output and the results cache.
* **Refactor**
  * Updated internal output handling to support count and time-bucketed aggregation workflows end-to-end.
* **Documentation**
  * Refined search usage/help text formatting for count-related examples.


components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -214,4 +243,92 @@ ErrorCode CountByTimeOutputHandler::finish() {
...
+        string const& collection
+)
+        : ::clp_s::search::OutputHandler(false, false) {
+    m_collection = connect_to_results_cache<OperationFailed>(uri, collection, m_client);
+}
+
+ErrorCode CountToResultsCacheOutputHandler::finish() {
+    if (0 == m_count) {
+        return ErrorCode::ErrorCodeSuccess;
+    }
```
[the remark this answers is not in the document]
davemarco: this matches the existing reducer path, which also publishes nothing for zero matches, so the established contract for count results is "no doc = 0". Whether the search ran is signaled by the exit code / job status, not by the results collection.

components/core/src/clp_s/OutputHandlerImpl.hpp
davidlion, line 1: Update to follow comment convention in https://docs.yscope.com/dev-guide/contrib-guides-cpp.html#declaration-order

components/core/src/clp_s/OutputHandlerImpl.hpp
```diff
@@ -279,6 +280,93 @@ class CountByTimeOutputHandler : public ::clp_s::search::OutputHandler {
...
+        // Constructors
+        OperationFailed(ErrorCode error_code, char const* const filename, int line_number)
+                : TraceableException(error_code, filename, line_number) {}
+    };
+
+    // Constructors
+    CountToResultsCacheOutputHandler(std::string const& uri, std::string const& collection);
+
+    // Methods inherited from OutputHandler
+    void write(
```
davidlion: Use trailing returns even for void functions. Applies below as well.

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -214,4 +232,86 @@ ErrorCode CountByTimeOutputHandler::finish() {
...
+    m_collection = connect_to_results_cache<OperationFailed>(uri, collection, m_client);
+}
+
+ErrorCode CountByTimeToResultsCacheOutputHandler::finish() {
+    if (m_bucket_counts.empty()) {
+        return ErrorCode::ErrorCodeSuccess;
+    }
+
+    // The bucket timestamp is also stored in a `timestamp` field so that the documents match the
+    // `{timestamp, count}` shape the reducer writes.
```
davidlion: Also long as the reducer and result cache classes are separate this is more than an implementation detail as it leaks outside the class (i.e. if someone changes the same of one class they should change the shape of the other). Let's move it to the class doc-string.

components/core/src/clp_s/kv_ir_search.cpp
```diff
@@ -246,10 +246,19 @@ auto search_kv_ir_stream(
...
+            = std::holds_alternative<CommandLineArguments::ReducerOutputHandlerOptions>(
+                    output_handler_options
+            );
+    auto const* results_cache_options
+            = std::get_if<CommandLineArguments::ResultsCacheOutputHandlerOptions>(
+                    &output_handler_options
+            );
+    bool const results_cache_aggregation_requested
+            = nullptr != results_cache_options
+              && results_cache_options->aggregation_type.has_value();
```
davidlion: Small nit use `auto` and `{}` when possible.

At minimum, either switch to auto or use `{}` for explicitly typed variables, see: https://app.notion.com/p/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?source=copy_link#3e50f81213d64d95bc20409a1a053e34.

components/core/src/clp_s/kv_ir_search.cpp
```diff
@@ -246,10 +246,19 @@ auto search_kv_ir_stream(
...
+    if (reducer_aggregation_requested || results_cache_aggregation_requested) {
         SPDLOG_ERROR("kv-ir search: Count support is not implemented.");
```
davidlion: Let's update this message since the check is generalized anyway.
```suggestion
        SPDLOG_ERROR("kv-ir search: Aggregation support is not implemented.");
```

components/core/src/clp_s/OutputHandlerImpl.hpp
davidlion, line 1: `CountReducerOutputHandler` and `CountByTimeReducerOutputHandler` match the current naming more I think (no `To`).

components/core/src/clp_s/CommandLineArguments.hpp
```diff
@@ -42,6 +42,8 @@ class CommandLineArguments {
         std::string dataset;
         uint64_t batch_size{1000};
         uint64_t max_num_results{1000};
+        std::optional<AggregationType> aggregation_type;
+        int64_t count_by_time_bucket_size{};  // Milliseconds
```
davidlion: Let's add the time unit to the name rather than in a comment. Iirc this is mimicking old/other code, but let's try to improve the new code.

(Alternatively, use std::chrono to have the unit in the type. Although, I don't think we really need the flexibility chrono gives for this. Up to you.)

I'll try to add this explicitly to our guidelines/docs when I have the time, sorry.

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -24,6 +26,28 @@ using std::string;
...
+/**
+ * Connects to the results cache and returns the requested collection.
+ * @tparam OperationFailedT The type of exception to throw on failure.
+ * @param uri
+ * @param collection
+ * @param client Returns the connected client.
+ * @return The collection.
+ */
+template <typename OperationFailedT>
+auto connect_to_results_cache(string const& uri, string const& collection, mongocxx::client& client)
```
davidlion: Unless there is a specific reason, let's use string_view instead of `string const&`. This applies all the way up the call chain.

davidlion (changes requested):

davidlion:
Sorry, I forgot to add:

After thinking about it, I think it is fine / makes sense, to match the stdout semantics of `{count: 12}, {count: 28}, ...` rather than using `upsert`.

components/core/src/clp_s/kv_ir_search.cpp
davemarco: switched both reducer_aggregation_requested and results_cache_aggregation_requested to auto const with brace-init

components/core/src/clp_s/OutputHandlerImpl.cpp
davemarco: I just removed this comment, its not really necessary

components/core/src/clp_s/OutputHandlerImpl.hpp
davemarco: updated the section title comments (implementing/overriding/Data members) per the convention

components/core/src/clp_s/OutputHandlerImpl.hpp
davemarco: applied trailing returns throughout the count handler classes

components/core/src/clp_s/OutputHandlerImpl.hpp
davemarco: dropped To from the reducer pair and the results-cache pair

components/core/src/clp_s/kv_ir_search.cpp
davemarco: done

components/core/src/clp_s/OutputHandlerImpl.cpp
davemarco: Kept string const& here — mongocxx's uri/collection lookup don't accept string_view

davemarco:
replaced the upserts with plain inserts and added archive_id to each doc - ({archive_id, count} / {archive_id, timestamp, count}).

components/core/src/clp_s/CommandLineArguments.hpp
davemarco: done

davemarco (commented):
addressed comments

components/core/src/clp_s/OutputHandlerImpl.cpp
davidlion: (edit: ignore I'm dumb)

Where do you see this? I'm probably missing something, but both https://mongocxx.org/api/mongocxx-4.1.1/classmongocxx_1_1v__noabi_1_1uri.html and https://mongocxx.org/api/mongocxx-4.1.1/classmongocxx_1_1v__noabi_1_1client.html#a10373137126703904f4f325a0624f445 appear to suggest they can take views.

(we should be on 4.1.1 based on deps/main.yaml)

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -214,4 +235,83 @@ ErrorCode CountByTimeOutputHandler::finish() {
     }
     return ErrorCode::ErrorCodeSuccess;
 }
+
+CountResultsCacheOutputHandler::CountResultsCacheOutputHandler(
+        string const& uri,
+        string const& collection,
+        string archive_id
+)
+        : ::clp_s::search::OutputHandler{false, false},
```
davidlion:
```suggestion
        : search::OutputHandler{false, false},
```
We're in the clp_s namespace already.

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -214,4 +235,83 @@ ErrorCode CountByTimeOutputHandler::finish() {
     }
     return ErrorCode::ErrorCodeSuccess;
 }
+
+CountResultsCacheOutputHandler::CountResultsCacheOutputHandler(
+        string const& uri,
+        string const& collection,
+        string archive_id
```
davidlion: Ignoring the uri/collection issues, archive_id can be a string_view here for increased flexibility.

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -214,4 +235,83 @@ ErrorCode CountByTimeOutputHandler::finish() {
...
+    } catch (mongocxx::exception const& e) {
+        return ErrorCode::ErrorCodeFailureDbBulkWrite;
+    }
+    return ErrorCode::ErrorCodeSuccess;
+}
+
+CountByTimeResultsCacheOutputHandler::CountByTimeResultsCacheOutputHandler(
+        string const& uri,
+        string const& collection,
+        string archive_id,
```
davidlion: Same as above comment.

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -214,4 +235,83 @@ ErrorCode CountByTimeOutputHandler::finish() {
...
+        int64_t count_by_time_bucket_size_ms
+)
+        : ::clp_s::search::OutputHandler{true, false},
```
davidlion: Same namespace comment.

components/core/src/clp_s/OutputHandlerImpl.hpp
```diff
@@ -245,38 +248,140 @@ class CountOutputHandler : public ::clp_s::search::OutputHandler {
...
+    auto write(
             std::string_view message,
             epochtime_t timestamp,
             std::string_view archive_id,
             int64_t log_event_idx
-    ) override {
-        int64_t bucket = (timestamp / m_count_by_time_bucket_size) * m_count_by_time_bucket_size;
+    ) -> void override {
+        int64_t bucket
+                = (timestamp / m_count_by_time_bucket_size_ms) * m_count_by_time_bucket_size_ms;
```
davidlion: I know this is technically from the existing code, but let's add _ms to timestamp (for both write functions) same as the bucket, since epochtime_t allows for different units of time.

components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -845,6 +846,15 @@ CommandLineArguments::parse_arguments(int argc, char const** argv) {
...
                     "The dataset name to include in each result document"
+            )(
+                    "count",
+                    "Count the number of results"
+            )(
+                    "count-by-time",
+                    po::value<int64_t>(
+                        &results_cache_options.count_by_time_bucket_size_ms
+                    )->value_name("SIZE"),
+                    "Count the number of results in each time span of the given size (ms)"
```
davidlion, line 862: Sorry, I should have thought of this before, but I'm thinking we might want to refactor the args to separate the actual aggregation operator args and the different output handler args. Otherwise there might be a lot of duplication.

Unless you want to add/start it in this PR, I think it is fine to leave this for the next PR that adds stdout output since that makes the problem even more clear.

davidlion (changes requested):
Very minor tweaks and then we're good to merge. I think the title is fine.

components/core/src/clp_s/OutputHandlerImpl.cpp
davidlion: Sorry, I read those docs more closely and understand now (imo the naming is a bit misleading lol).

I think it is still worth it to take string_view everywhere and then materialize it as a string when calling the mongodb functions to make everything more flexible.

davidlion (approved):

components/core/src/clp_s/CommandLineArguments.cpp
davidlion: (Delayed to later PR)

PR #2327, opened 2026-06-11
feat(helm)!: Replace global `workerConcurrency` with per-worker `slotsPerPod` concurrency settings (resolves #2297).

### LINKED ISSUE #2297: Allow per-workload worker concurrency settings
### Request

The Helm chart currently exposes a single `workerConcurrency` value for all worker workloads:

```yaml
workerConcurrency: 8
```

That value is applied to compression workers, query workers, and reducers. This makes deployments difficult to tune because each workload can have different CPU, memory, I/O, and latency characteristics. For example, a concurrency value that is appropriate for compression throughput may overcommit query workers or reducers, while a conservative setting for query workers may underutilize dedicated compression nodes.

Concurrency should be configurable independently for each worker type.

This would make Helm deployments easier to tune for:

- mixed workloads running in the same node pool;
- dedicated node pools with different CPU or memory sizes;
- query-heavy deployments where query workers should use lower concurrency than compression workers;
- reducer deployments where the number of reducer ports should match reducer concurrency rather than a global worker setting.

### Possible implementation

Replace the global `workerConcurrency` setting with per-worker execution-slot fields. The preferred new field name is `slotsPerPod`:

```yaml
scheduling:
  compressionWorker:
    replicas: 1
    slotsPerPod: 8
  queryWorker:
    replicas: 1
    slotsPerPod: 8
  reducer:
    replicas: 1
    slotsPerPod: 8
```

`slotsPerPod` distinguishes this setting from `replicas`, avoids exposing Celery’s `--concurrency` terminology as the public Helm API, and also fits reducers, where each slot maps to one reducer-server process and one exposed port.

Other possible names:

- `executionSlotsPerPod`: More explicit and self-documenting, but clunkier in values files.
- `processesPerPod`: Accurate for the current implementation, but exposes an implementation detail and is less framework/runtime agnostic.

The Helm templates should use the workload-specific concurrency value:

- `compression-worker-deployment.yaml`: pass `scheduling.compressionWorker.slotsPerPod` to Celery `--concurrency`.
- `query-worker-deployment.yaml`: pass `scheduling.queryWorker.slotsPerPod` to Celery `--concurrency`.
- `reducer-deployment.yaml`: pass `scheduling.reducer.slotsPerPod` to the reducer `--concurrency`.
- `reducer-service.yaml`: generate reducer service ports from `scheduling.reducer.slotsPerPod`.
- `_helpers.tpl`: report telemetry gauges using each workload's configured concurrency instead of reusing a single value for compression workers, query workers, and reducers.

This would align Helm behavior with the package/docker-compose deployment path, which already supports separate `CLP_COMPRESSION_WORKER_CONCURRENCY`, `CLP_QUERY_WORKER_CONCURRENCY`, and `CLP_REDUCER_CONCURRENCY` settings.

Nathan903:
# Description

Currently, the Helm chart uses one global `workerConcurrency` setting for all worker types. Since compression workers, query workers, and reducers have different CPU and memory requirements, this makes cluster scaling difficult.

This PR replaces `workerConcurrency` with `slotsPerPod` setting for each worker type under `scheduling`, allowing `compressionWorker`, `queryWorker`, and `reducer` pods to be tuned independently for better resource utilization.

**Breaking changes:** `workerConcurrency` is removed. Any existing `values.yaml` or --set command-line overrides using `workerConcurrency` will be ignored and must be migrated to `scheduling.<worker>.slotsPerPod`.

# Checklist

  breaking change.

# Validation performed

```console
nathanhung@baker18:~/clp$ helm template test-clp ./tools/deployment/package-helm | grep -E "concurrency|port: 140"
      scheduler_concurrency: 4
      base_port: 14009
      port: 14009
      port: 14010
      port: 14011
      port: 14012
      port: 14013
      port: 14014
      port: 14015
      port: 14016
            "--concurrency", "8",
            "--concurrency", "8",
            "--concurrency", "8",
            - "{\"resourceMetrics\":[{\"resource\":{\"attributes\":[{\"key\":\"service.name\",\"value\":{\"stringValue\":\"controller\"}}]},\"scopeMetrics\":[{\"scope\":{\"name\":\"clp.controller\"},\"metrics\":[{\"name\":\"clp.deployment.compression_worker_replicas\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"1\",\"timeUnixNano\":\"1781210113912693342\"}]}},{\"name\":\"clp.deployment.compression_worker_concurrency\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"8\",\"timeUnixNano\":\"1781210113912693342\"}]}},{\"name\":\"clp.deployment.query_worker_replicas\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"1\",\"timeUnixNano\":\"1781210113912693342\"}]}},{\"name\":\"clp.deployment.query_worker_concurrency\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"8\",\"timeUnixNano\":\"1781210113912693342\"}]}},{\"name\":\"clp.deployment.reducer_replicas\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"1\",\"timeUnixNano\":\"1781210113912693342\"}]}},{\"name\":\"clp.deployment.reducer_concurrency\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"8\",\"timeUnixNano\":\"1781210113912693342\"}]}}]}]}]}"
nathanhung@baker18:~/clp$ helm template test-clp ./tools/deployment/package-helm \
  --set scheduling.compressionWorker.slotsPerPod=4 \
  --set scheduling.queryWorker.slotsPerPod=2 \
  --set scheduling.reducer.slotsPerPod=1 | grep -E "concurrency|port: 140"
      scheduler_concurrency: 4
      base_port: 14009
      port: 14009
            "--concurrency", "4",
            "--concurrency", "2",
            "--concurrency", "1",
            - "{\"resourceMetrics\":[{\"resource\":{\"attributes\":[{\"key\":\"service.name\",\"value\":{\"stringValue\":\"controller\"}}]},\"scopeMetrics\":[{\"scope\":{\"name\":\"clp.controller\"},\"metrics\":[{\"name\":\"clp.deployment.compression_worker_replicas\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"1\",\"timeUnixNano\":\"1781210117375789593\"}]}},{\"name\":\"clp.deployment.compression_worker_concurrency\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"4\",\"timeUnixNano\":\"1781210117375789593\"}]}},{\"name\":\"clp.deployment.query_worker_replicas\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"1\",\"timeUnixNano\":\"1781210117375789593\"}]}},{\"name\":\"clp.deployment.query_worker_concurrency\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"2\",\"timeUnixNano\":\"1781210117375789593\"}]}},{\"name\":\"clp.deployment.reducer_replicas\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"1\",\"timeUnixNano\":\"1781210117375789593\"}]}},{\"name\":\"clp.deployment.reducer_concurrency\",\"gauge\":{\"dataPoints\":[{\"asInt\":\"1\",\"timeUnixNano\":\"1781210117375789593\"}]}}]}]}]}"
nathanhung@baker18:~/clp$ helm template test-clp ./tools/deployment/package-helm \
  --set workerConcurrency=8
Error: execution error at (clp/templates/service-account.yaml:5:4): The global 'workerConcurrency' setting has been removed. Please use 'slotsPerPod' under 'scheduling.compressionWorker', 'scheduling.queryWorker', and 'scheduling.reducer' instead.

Use --debug flag to render out invalid YAML
```

## Summary by CodeRabbit

* **Chores**
  * Deployment now supports separate concurrency (per-worker slots) for compression, query, and reducer workers for finer-grained resource allocation.
  * Runtime metrics, service port generation and deployed worker concurrency now reflect per-worker settings for more accurate behaviour and telemetry.
  * Chart version advanced to the next development release.

* **Documentation**
  * Kubernetes deployment guide example updated to show the new per-worker scheduling configuration.

### DIFF
--- a/docs/src/user-docs/guides-k8s-deployment.md
+++ b/docs/src/user-docs/guides-k8s-deployment.md
@@ -216,8 +216,14 @@ image:
     repository: "bitnami/kubectl"
     digest: "sha256:98736aabcecb8d3cbcdcd7b132d14b1d67ed99bac2f06d471f06235933103df3"  # v1.36.0
 
-# Adjust worker concurrency
-workerConcurrency: 16
+# Adjust worker concurrency per component
+scheduling:
+  compressionWorker:
+    slotsPerPod: 16
+  queryWorker:
+    slotsPerPod: 16
+  reducer:
+    slotsPerPod: 16
 
 # Configure CLP settings
 clpConfig:
--- a/docs/src/user-docs/guides-k8s-deployment.md
+++ b/docs/src/user-docs/guides-k8s-deployment.md
@@ -216,7 +216,7 @@ image:
     repository: "bitnami/kubectl"
     digest: "sha256:98736aabcecb8d3cbcdcd7b132d14b1d67ed99bac2f06d471f06235933103df3"  # v1.36.0
 
-# Adjust worker concurrency per component
+# Adjust worker concurrency
 scheduling:
   compressionWorker:
     slotsPerPod: 16
--- a/tools/deployment/package-helm/templates/_helpers.tpl
+++ b/tools/deployment/package-helm/templates/_helpers.tpl
@@ -97,9 +97,9 @@ in controller.py, ensuring feature parity between Docker Compose and Helm deploy
 {{- $compressionWorkerReplicas := $compressionWorkerScheduling.replicas | default 1 | int -}}
 {{- $queryWorkerReplicas := .Values.scheduling.queryWorker.replicas | default 1 | int -}}
 {{- $reducerReplicas := .Values.scheduling.reducer.replicas | default 1 | int -}}
-{{- $compressionWorkerConcurrency := include "clp.compressionWorkerConcurrency" . | default 8 | int -}}
-{{- $queryWorkerConcurrency := include "clp.queryWorkerConcurrency" . | default 8 | int -}}
-{{- $reducerConcurrency := include "clp.reducerConcurrency" . | default 8 | int -}}
+{{- $compressionWorkerConcurrency := include "clp.compressionWorkerConcurrency" . | int -}}
+{{- $queryWorkerConcurrency := include "clp.queryWorkerConcurrency" . | int -}}
+{{- $reducerConcurrency := include "clp.reducerConcurrency" . | int -}}
 {{- printf `{"resourceMetrics":[{"resource":{"attributes":[{"key":"service.name","value":{"stringValue":"controller"}}]},"scopeMetrics":[{"scope":{"name":"clp.controller"},"metrics":[{"name":"clp.deployment.compression_worker_replicas","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.compression_worker_concurrency","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.query_worker_replicas","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.query_worker_concurrency","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.reducer_replicas","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.reducer_concurrency","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}}]}]}]}` $compressionWorkerReplicas $timestampNs $compressionWorkerConcurrency $timestampNs $queryWorkerReplicas $timestampNs $queryWorkerConcurrency $timestampNs $reducerReplicas $timestampNs $reducerConcurrency $timestampNs -}}
 {{- end -}}
 
@@ -110,7 +110,12 @@ Gets the effective concurrency for the compression worker.
 @return {string} The effective concurrency
 */}}
 {{- define "clp.compressionWorkerConcurrency" -}}
-{{- coalesce .Values.scheduling.compressionWorker.slotsPerPod .Values.workerConcurrency -}}
+{{- $slots := .Values.scheduling.compressionWorker.slotsPerPod -}}
+{{- if eq (typeOf $slots) "<nil>" -}}
+{{- .Values.workerConcurrency | default 8 -}}
+{{- else -}}
+{{- $slots -}}
+{{- end -}}
 {{- end }}
 
 {{/*
@@ -120,7 +125,12 @@ Gets the effective concurrency for the query worker.
 @return {string} The effective concurrency
 */}}
 {{- define "clp.queryWorkerConcurrency" -}}
-{{- coalesce .Values.scheduling.queryWorker.slotsPerPod .Values.workerConcurrency -}}
+{{- $slots := .Values.scheduling.queryWorker.slotsPerPod -}}
+{{- if eq (typeOf $slots) "<nil>" -}}
+{{- .Values.workerConcurrency | default 8 -}}
+{{- else -}}
+{{- $slots -}}
+{{- end -}}
 {{- end }}
 
 {{/*
@@ -130,7 +140,12 @@ Gets the effective concurrency for the reducer.
 @return {string} The effective concurrency
 */}}
 {{- define "clp.reducerConcurrency" -}}
-{{- coalesce .Values.scheduling.reducer.slotsPerPod .Values.workerConcurrency -}}
+{{- $slots := .Values.scheduling.reducer.slotsPerPod -}}
+{{- if eq (typeOf $slots) "<nil>" -}}
+{{- .Values.workerConcurrency | default 8 -}}
+{{- else -}}
+{{- $slots -}}
+{{- end -}}
 {{- end }}
 
 {{/*
--- a/tools/deployment/package-helm/templates/_helpers.tpl
+++ b/tools/deployment/package-helm/templates/_helpers.tpl
@@ -97,9 +97,41 @@ in controller.py, ensuring feature parity between Docker Compose and Helm deploy
 {{- $compressionWorkerReplicas := $compressionWorkerScheduling.replicas | default 1 | int -}}
 {{- $queryWorkerReplicas := .Values.scheduling.queryWorker.replicas | default 1 | int -}}
 {{- $reducerReplicas := .Values.scheduling.reducer.replicas | default 1 | int -}}
-{{- $workerConcurrency := .Values.workerConcurrency | default 8 | int -}}
-{{- printf `{"resourceMetrics":[{"resource":{"attributes":[{"key":"service.name","value":{"stringValue":"controller"}}]},"scopeMetrics":[{"scope":{"name":"clp.controller"},"metrics":[{"name":"clp.deployment.compression_worker_replicas","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.compression_worker_concurrency","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.query_worker_replicas","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.query_worker_concurrency","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.reducer_replicas","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.reducer_concurrency","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}}]}]}]}` $compressionWorkerReplicas $timestampNs $workerConcurrency $timestampNs $queryWorkerReplicas $timestampNs $workerConcurrency $timestampNs $reducerReplicas $timestampNs $workerConcurrency $timestampNs -}}
+{{- $compressionWorkerConcurrency := include "clp.compressionWorkerConcurrency" . | default 8 | int -}}
+{{- $queryWorkerConcurrency := include "clp.queryWorkerConcurrency" . | default 8 | int -}}
+{{- $reducerConcurrency := include "clp.reducerConcurrency" . | default 8 | int -}}
+{{- printf `{"resourceMetrics":[{"resource":{"attributes":[{"key":"service.name","value":{"stringValue":"controller"}}]},"scopeMetrics":[{"scope":{"name":"clp.controller"},"metrics":[{"name":"clp.deployment.compression_worker_replicas","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.compression_worker_concurrency","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.query_worker_replicas","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.query_worker_concurrency","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.reducer_replicas","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.reducer_concurrency","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}}]}]}]}` $compressionWorkerReplicas $timestampNs $compressionWorkerConcurrency $timestampNs $queryWorkerReplicas $timestampNs $queryWorkerConcurrency $timestampNs $reducerReplicas $timestampNs $reducerConcurrency $timestampNs -}}
 {{- end -}}
+
+{{/*
+Gets the effective concurrency for the compression worker.
+
+@param {object} . Root template context
+@return {string} The effective concurrency
+*/}}
+{{- define "clp.compressionWorkerConcurrency" -}}
+{{- coalesce .Values.scheduling.compressionWorker.slotsPerPod .Values.workerConcurrency -}}
+{{- end }}
+
+{{/*
+Gets the effective concurrency for the query worker.
+
+@param {object} . Root template context
+@return {string} The effective concurrency
+*/}}
+{{- define "clp.queryWorkerConcurrency" -}}
+{{- coalesce .Values.scheduling.queryWorker.slotsPerPod .Values.workerConcurrency -}}
+{{- end }}
+
+{{/*
+Gets the effective concurrency for the reducer.
+
+@param {object} . Root template context
+@return {string} The effective concurrency
+*/}}
+{{- define "clp.reducerConcurrency" -}}
+{{- coalesce .Values.scheduling.reducer.slotsPerPod .Values.workerConcurrency -}}
+{{- end }}
 
 {{/*
 Provides environment variables for telemetry (except service.name).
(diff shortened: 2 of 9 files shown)


Nathan903:
@junhaoliao  should I consider the case of `slotsPerPod=0`? 
Is `workerConcurrency=0` suppose to work currently? From the current codebase, compression worker and query worker works with `--concurrency 0 `  and the reducer is limited to as least 1 in, but in the helm deployment would give it 0 ports in `tools\deployment\package-helm\templates\reducer-service.yaml` (which i think is a bug?)

junhaoliao:
> @junhaoliao should I consider the case of `slotsPerPod=0`? Is `workerConcurrency=0` suppose to work currently? From the current codebase, compression worker and query worker works with `--concurrency 0 ` and the reducer is limited to as least 1 in, but in the helm deployment would give it 0 ports in `tools\deployment\package-helm\templates\reducer-service.yaml` (which i think is a bug?)

We dont expect users to set a value lower than or equal to 0. that said, i think it's okay to defer the validation until #1941 is implemented

tools/deployment/package-helm/values.yaml
```diff
@@ -86,10 +86,13 @@ scheduling:
   # Worker
   compressionWorker:
     replicas: 1
+    # slotsPerPod: 8
   queryWorker:
     replicas: 1
+    # slotsPerPod: 2
   reducer:
     replicas: 1
+    # slotsPerPod: 2
```
junhaoliao: 1. let's explitcitly set them so we don't have to fallback to default values in _helpers.tpl
2. let's match the original count in the defaults

```suggestion
compressionWorker:
replicas: 1
    slotsPerPod: 8
queryWorker:
replicas: 1
    slotsPerPod: 8
reducer:
replicas: 1
    slotsPerPod: 8
```

tools/deployment/package-helm/values.yaml
junhaoliao: deep apology for my oversights in the writing of #2297 . i don't think we should keep the workerConcurrency key which should hopefully help clean up the code too. i've updated #2297

tools/deployment/package-helm/templates/reducer-service.yaml
```diff
@@ -12,7 +12,7 @@ spec:
     {{- include "clp.selectorLabels" . | nindent 4 }}
     app.kubernetes.io/component: "reducer"
   ports:
-    {{- range $i := until (int .Values.workerConcurrency) }}
+    {{- range $i := until (int .Values.scheduling.reducer.slotsPerPod) }}
```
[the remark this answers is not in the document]
Nathan903: We dont expect users to set a value lower than or equal to 0. that said, i think it's okay to defer the validation until https://github.com/y-scope/clp/issues/1941 is implemented

tools/deployment/package-helm/values.yaml
Nathan903: done. Ive removed `workerConcurrency`, and added a check to fail if there is `workerConcurrency` used.

Nathan903:
ive simplified the changes and re-run all test. 

> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > tools/deployment/package-helm/templates/_helpers.tpl (1)> `107-109`: _⚠️ Potential issue_ | _🟠 Major_
> > > **Defensively default `scheduling.*.slotsPerPod` across all worker/reducer templates (not just telemetry).**
> > > `values.yaml` sets `slotsPerPod: 8`, but templates currently use `scheduling.<worker>.slotsPerPod` with no defensive default. If a user overrides `scheduling.<worker>` without `slotsPerPod` (e.g., `compressionWorker: {}`), then:
> > > 
> > > * Telemetry: `_helpers.tpl` sets concurrency with `| int` and no `default`, so `clp.deployment.*_concurrency` becomes 0 when unset.
> > > * Worker start-up: `templates/compression-worker-deployment.yaml` / `templates/query-worker-deployment.yaml` pass `--concurrency` directly from `.Values.scheduling.<worker>.slotsPerPod`.
> > > * Reducer service: `templates/reducer-service.yaml` generates ports via `range ... until (int .Values.scheduling.reducer.slotsPerPod)`; if unset, the range is empty (no ports), breaking the headless service.
> > > 
> > > Apply the same defensive default/validation for missing `slotsPerPod` at every call site (ideally via shared helpers), not only inside `clp.topologyMetricsPayload`.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@tools/deployment/package-helm/templates/_helpers.tpl` around lines 107 - 109,
> > > The helper variables compressionWorkerConcurrency, queryWorkerConcurrency and
> > > reducerConcurrency currently convert .Values.scheduling.*.slotsPerPod with | int
> > > but no defensive default; update those expressions to apply a fallback (e.g. |
> > > default 8 | int) so an omitted slotsPerPod yields the intended default, and then
> > > propagate those helper variables everywhere the raw .Values are used (replace
> > > direct uses of .Values.scheduling.<worker>.slotsPerPod in places that render
> > > --concurrency and in the range(...) that generates reducer ports with the new
> > > helper variables or a shared helper function like schedulingSlotsPerPod that
> > > returns (.Values.scheduling.<name>.slotsPerPod | default 8 | int) so all call
> > > sites (including clp.topologyMetricsPayload, any --concurrency flags, and the
> > > reducer ports range) consistently use the defensive default.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Outside diff comments:
> In `@tools/deployment/package-helm/templates/_helpers.tpl`:
> - Around line 107-109: The helper variables compressionWorkerConcurrency,
> queryWorkerConcurrency and reducerConcurrency currently convert
> .Values.scheduling.*.slotsPerPod with | int but no defensive default; update
> those expressions to apply a fallback (e.g. | default 8 | int) so an omitted
> slotsPerPod yields the intended default, and then propagate those helper
> variables everywhere the raw .Values are used (replace direct uses of
> .Values.scheduling.<worker>.slotsPerPod in places that render --concurrency and
> in the range(...) that generates reducer ports with the new helper variables or
> a shared helper function like schedulingSlotsPerPod that returns
> (.Values.scheduling.<name>.slotsPerPod | default 8 | int) so all call sites
> (including clp.topologyMetricsPayload, any --concurrency flags, and the reducer
> ports range) consistently use the defensive default.
> ```
> 
> ℹ️ Review info

we already explicitly set slotsPerPod: 8 in the default values.yaml file.

Nathan903:
ive simplified the changes, and added documentation, and re-run the validation. PR should be good for merging now @junhaoliao

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -97,8 +97,10 @@ in controller.py, ensuring feature parity between Docker Compose and Helm deploy
...
+{{- $compressionWorkerConcurrency := .Values.scheduling.compressionWorker.slotsPerPod | int -}}
+{{- $queryWorkerConcurrency := .Values.scheduling.queryWorker.slotsPerPod | int -}}
+{{- $reducerConcurrency := .Values.scheduling.reducer.slotsPerPod | int -}}
```
junhaoliao: Use a shared helper so the topology metrics and runtime settings stay in sync, without adding an implicit default.

```suggestion
{{- $compressionWorkerConcurrency := include "clp.schedulingSlotsPerPod" (dict
  "root" .
  "component" "compressionWorker"
) | int -}}
{{- $queryWorkerConcurrency := include "clp.schedulingSlotsPerPod" (dict
  "root" .
  "component" "queryWorker"
) | int -}}
{{- $reducerConcurrency := include "clp.schedulingSlotsPerPod" (dict
  "root" .
  "component" "reducer"
) | int -}}
```

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -97,8 +97,10 @@ in controller.py, ensuring feature parity between Docker Compose and Helm deploy
...
+{{- printf `{"resourceMetrics":[{"resource":{"attributes":[{"key":"service.name","value":{"stringValue":"controller"}}]},"scopeMetrics":[{"scope":{"name":"clp.controller"},"metrics":[{"name":"clp.deployment.compression_worker_replicas","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.compression_worker_concurrency","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.query_worker_replicas","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.query_worker_concurrency","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.reducer_replicas","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}},{"name":"clp.deployment.reducer_concurrency","gauge":{"dataPoints":[{"asInt":"%d","timeUnixNano":"%d"}]}}]}]}]}` $compressionWorkerReplicas $timestampNs $compressionWorkerConcurrency $timestampNs $queryWorkerReplicas $timestampNs $queryWorkerConcurrency $timestampNs $reducerReplicas $timestampNs $reducerConcurrency $timestampNs -}}
 {{- end -}}
```
junhaoliao, line 113: Add the helper near the existing metric helper so each worker-like component uses the same required slot lookup.

```suggestion
{{- end -}}

{{/*
Gets the effective slot count for a worker-like component.

The `slotsPerPod` value is required for each component.

@param {object} root Template context
@param {string} component One of compressionWorker/queryWorker/reducer
@return {int} Slot count for the component
*/}}
{{- define "clp.schedulingSlotsPerPod" -}}
{{- $scheduling := index .root.Values.scheduling .component -}}
{{- $slotsPerPod := $scheduling.slotsPerPod -}}
{{- if not (hasKey $scheduling "slotsPerPod") -}}
{{- fail (printf "scheduling.%s.slotsPerPod is required" .component) -}}
{{- end -}}
{{- $slotsPerPod | int -}}
{{- end -}}
```

tools/deployment/package-helm/templates/compression-worker-deployment.yaml
```diff
@@ -84,7 +84,7 @@ spec:
             "/opt/clp/lib/python3/site-packages/bin/celery",
             "-A", "job_orchestration.executor.compress",
             "worker",
-            "--concurrency", "{{ .Values.workerConcurrency }}",
+            "--concurrency", "{{ .Values.scheduling.compressionWorker.slotsPerPod }}",
```
junhaoliao: Use the same required slot lookup here instead of reading the value directly.

```suggestion
            "--concurrency", "{{ include "clp.schedulingSlotsPerPod" (dict
              "root" .
              "component" "compressionWorker"
            ) }}",
```

tools/deployment/package-helm/templates/query-worker-deployment.yaml
```diff
@@ -84,7 +84,7 @@ spec:
             "/opt/clp/lib/python3/site-packages/bin/celery",
             "-A", "job_orchestration.executor.query",
             "worker",
-            "--concurrency", "{{ .Values.workerConcurrency }}",
+            "--concurrency", "{{ .Values.scheduling.queryWorker.slotsPerPod }}",
```
junhaoliao: Use the same required slot lookup here instead of reading the value directly.

```suggestion
            "--concurrency", "{{ include "clp.schedulingSlotsPerPod" (dict
              "root" .
              "component" "queryWorker"
            ) }}",
```

tools/deployment/package-helm/templates/reducer-deployment.yaml
```diff
@@ -53,7 +53,7 @@ spec:
             "python3", "-u",
             "-m", "job_orchestration.reducer.reducer",
             "--config", "/etc/clp-config.yaml",
-            "--concurrency", "{{ .Values.workerConcurrency }}",
+            "--concurrency", "{{ .Values.scheduling.reducer.slotsPerPod }}",
```
junhaoliao: Use the same required slot lookup here instead of reading the value directly.

```suggestion
            "--concurrency", "{{ include "clp.schedulingSlotsPerPod" (dict
              "root" .
              "component" "reducer"
            ) }}",
```

tools/deployment/package-helm/templates/reducer-service.yaml
junhaoliao: Keep the reducer service ports tied to the same slot-count helper used by the reducer process.

```suggestion
    {{- $reducerSlotsPerPod := include "clp.schedulingSlotsPerPod" (dict
      "root" .
      "component" "reducer"
    ) | int }}
    {{- range $i := until $reducerSlotsPerPod }}
```

tools/deployment/package-helm/templates/service-account.yaml
```diff
@@ -1,3 +1,9 @@
+{{/*
+Fail fast if the deprecated workerConcurrency setting is used.
+*/}}
+{{- if hasKey .Values "workerConcurrency" -}}
+{{- fail "The global 'workerConcurrency' setting has been removed. Please use 'slotsPerPod' under 'scheduling.compressionWorker', 'scheduling.queryWorker', and 'scheduling.reducer' instead." -}}
+{{- end -}}
```
junhaoliao: this guard is not needed as we discussed that validations can be added later

```suggestion
```

docs/src/user-docs/guides-k8s-deployment.md
```diff
@@ -216,8 +216,14 @@ image:
     repository: "bitnami/kubectl"
     digest: "sha256:98736aabcecb8d3cbcdcd7b132d14b1d67ed99bac2f06d471f06235933103df3"  # v1.36.0
 
-# Adjust worker concurrency
-workerConcurrency: 16
+# Adjust worker concurrency per component
```
junhaoliao: can be simplified as
```suggestion
# Adjust worker concurrency
```

junhaoliao (changes requested):
for the title

kinda feel "independent" is redudant when we have stated "per-worker". how about
```
feat(helm)!: Replace global `workerConcurrency` with per-worker `slotsPerPod` concurrency settings (resolves #2297).
```

junhaoliao (approved):

Nathan903:
just rerun validation and everything works (except the validation to catch legacy `workerConcurrency` which is removed). Should I edit or make a comment in https://github.com/y-scope/clp/issues/1941?

junhaoliao:
> just rerun validation and everything works (except the validation to catch legacy `workerConcurrency` which is removed). Should I edit or make a comment in https://github.com/y-scope/clp/issues/1941?

editing is fine. we can cross out the legacy workerConcurrency key validation part and state that we decided not to validate such, given (1) we have not declared a formal depreciation policy in general so there is no way to determinine how long we should keep such validation logic and (2) there is no clp-package-wide helm validation infra at the moment anyway 

thanks for being diligent

Nathan903:
> > just rerun validation and everything works (except the validation to catch legacy `workerConcurrency` which is removed). Should I edit or make a comment in #1941?
> 
> editing is fine. we can cross out the legacy workerConcurrency key validation part and state that we decided not to validate such, given (1) we have not declared a formal depreciation policy in general so there is no way to determinine how long we should keep such validation logic and (2) there is no clp-package-wide helm validation infra at the moment anyway
> 
> thanks for being diligent

so I add that all three slotsPerPod need to be >=1, and include the workerConcurrency key validation (but format as strike through) right?
for `slotsPerPod` of `compressionWorker` and `queryWorker` I think `--concurrency=0 ` means use number of cpu as the concurrency. Do we want to allow this? (so `compressionWorker` and `queryWorker` just need to be >=0)

Nathan903:
added validation task to https://github.com/y-scope/clp/issues/1941

PR #2288, opened 2026-05-19
feat(telemetry): Add metrics for query and compression schedulers/workers (fixes #2346).

Nathan903:
# Description

Adds OpenTelemetry metrics collection to the compression and query schedulers/workers, plus deployment config for the OTel collector.

Counters 
- `clp.compression.bytes_input_total`: Size of uncompressed data processed by compression, tagged by success/failure status
- `clp.compression.bytes_output_total`: Size of compressed output generated by the compression, tagged by success/failure status
- `clp.compression.tasks.completed` / `failed`: Task completion counts on the compression scheduler
- `clp.query.tasks.completed` / `failed`: Task completion counts on the query scheduler

Gauges (point-in-time snapshots)
- `clp.compression.active_jobs` / `outstanding_tasks`: In-flight job and task levels on the compression scheduler
- `clp.query.active_jobs` / `outstanding_tasks`: In-flight job and task levels on the query scheduler

Histograms
- `clp.compression.job.duration` / `task.duration`: Execution durations for compression jobs and tasks
- `clp.compression.input_rate` / `output_rate`: Speed of uncompressed bytes processed and compressed bytes output per task
- `clp.query.job.duration` / `task.duration`: Execution durations for query jobs and tasks

Also small fixes to fix integration test: `SO_REUSEADDR` on port checks and increased default command timeout.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

Integration test all passed:
```
task tests:integration
```

End-to-end test:
Run this on the client
```bash
sbin/start-clp.sh
sbin/compress.sh ~/hive-24hr-ts/hive-24hr-ts.jsonl
sbin/search.sh "ERROR"
```
Run this on the telemetry server:
```bash
docker exec clp-telemetry-server-clickhouse-1 clickhouse-client --user default --password clickhouse --query "SELECT * FROM clp_telemetry.otel_metrics_gauge ORDER BY TimeUnix DESC LIMIT 5 FORMAT Vertical"
docker exec clp-telemetry-server-clickhouse-1 clickhouse-client --user default --password clickhouse --query "SELECT * FROM clp_telemetry.otel_metrics_sum ORDER BY TimeUnix DESC LIMIT 5 FORMAT Vertical"
[...]

PR #2335, opened 2026-06-16
feat(clp-s::search): Add support for exporting search telemetry using an OTLP HTTP exporter.

gibber9809:
# Description

This PR is based on the changes in @ShangDanLuXian 's branch [here](https://github.com/ShangDanLuXian/clp/tree/telemetry-export). It adds a tracer provider so that the metrics gathered during search can be exported.

The changes comprise:
- Installing a tracer provider for the duration of the clp-s command when --enable-telemetry is passed
- Adding the TelemetryContext class, which is configured via the standard OTEL_*/CLP_TELEMETRY_ENDPOINT environment variables. This class creates and destroys the tracer provider using RAII.
- Changing the build system to build opentelemetry-cpp with the OTLP HTTP exporter, and adding the required protobuf dependency for the exporter.

Actually exporting these traces in the package requires a small amount of config change as well as a small change to `fs_search_task.py`. Since those changes overlap with #2288 they are being left for a small follow-up PR.

# Checklist

  breaking change.

# Validation performed

* @ShangDanLuXian previously validated that spans from search get exported to the collector successfully (when adding a few lines to `fs_search_task.py` to enable telemetry in clp-s.)

## Release Notes

* **New Features**
  * When telemetry is enabled, search tracing now initializes an OpenTelemetry telemetry context and cleanly shuts it down when the app exits.

* **Chores**
  * Improved OpenTelemetry C++ dependency discovery and extended search telemetry build/link support.
  * Updated the core build to include Protobuf v31.1 and configure OpenTelemetry for OTLP over HTTP (gRPC disabled).

* **Documentation**
  * Added Protobuf v31.1 to the core dependency table.


components/core/src/clp_s/search/TelemetryContext.cpp
davidlion, line 1: I still need to go over things in dept, but at a glance could you please update the category comments to https://docs.yscope.com/dev-guide/contrib-guides-cpp.html#declaration-order (same as last PR).

components/core/src/clp_s/search/TelemetryContext.cpp
```diff
@@ -0,0 +1,157 @@
...
+ * The standard `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`/`OTEL_EXPORTER_OTLP_ENDPOINT` variables are
+ * honoured natively by `OtlpHttpExporterOptions`, so this only supplies an override derived from
+ * CLP's `CLP_TELEMETRY_ENDPOINT` when none of the standard variables are set.
+ *
+ * @return The fully-qualified traces URL, or std::nullopt if opentelemetry defaults should be used.
+ */
+[[nodiscard]] auto resolve_clp_endpoint() -> std::optional<std::string>;
+
+/**
+ * @return A reResourcesource describing this process; a default `service.name` of `clp-search`
```
davidlion: typo?
```suggestion
 * @return A resource describing this process; a default `service.name` of `clp-search`
```

davidlion (dismissed):
Small typo otherwise lgtm

davidlion (approved):

PR #2336, opened 2026-06-17
fix(job-orchestration): Yield between cancelling search jobs (fixes #2296).

### LINKED ISSUE #2296: Query scheduler crash-loops when query_jobs table has large backlog of timed-out jobs
## Problem

The query scheduler enters a permanent CrashLoopBackOff when the `query_jobs` table accumulates a large number of rows (1.5M+ in our case). The scheduler becomes unresponsive to liveness probes and gets killed repeatedly.

## Root Cause

The `search-job-handler` in the query scheduler cancels timed-out jobs **one at a time**, with a full DB round-trip per job (~500ms each). When there is a large backlog of jobs to cancel (e.g., from a burst of query submissions or an external caller flooding the API), the scheduler spends its entire liveness window (180s) doing sequential cancellations and never serves the health probe.

```
2026-05-27 06:52:41,656 search-job-handler [INFO] Cancelled job 1889822.
2026-05-27 06:52:42,180 search-job-handler [INFO] Cancelled job 1889823.
2026-05-27 06:52:42,703 search-job-handler [INFO] Cancelled job 1889824.
... (~2 jobs/sec → would take days to clear a 600K+ backlog)
```

Exit code 137 confirms it is killed by the liveness probe (SIGKILL), not OOM.

### 1. Batch cancellation (high impact, low effort)

Replace the per-job cancellation loop with a bulk UPDATE:

```python
# Instead of:
for job in timed_out_jobs:
    cancel_job(job.id)  # individual DB round-trip

# Do:
job_ids = [j.id for j in timed_out_jobs[:BATCH_SIZE]]
cursor.execute("UPDATE query_jobs SET status=%s WHERE id IN (...)", (CANCELLED, *job_ids))
```

### 2. Decouple health endpoint from job processing

Run the health/liveness endpoint on a separate thread or async task so that slow business logic (cancellations, scheduling) never blocks health probe responses.

### 3. Add TTL / auto-cleanup for terminal query jobs

Terminal jobs (succeeded, failed, cancelled) serve no purpose after a short retention period. Either:
- Add a configurable cleanup that deletes jobs older than N days
- Or document that operators should set up a cron to purge them

### 4. Foreign key constraint makes manual cleanup difficult

`query_tasks` has a FK to `query_jobs(id)` with `ON DELETE RESTRICT`, making it impossible to delete old query_jobs without first deleting all associated query_tasks. Consider `ON DELETE CASCADE` or providing an admin API endpoint for cleanup.

## Environment

- CLP version: 0.12.0
- Deployment: Kubernetes (Helm chart clp-0.3.1)
- query_jobs table: ~1.89M rows
- Liveness probe: tcp-socket on port 7000, timeout 180s
- Pod exit code: 137 (killed by liveness probe)

## Workaround

Manually purge the table:
```sql
TRUNCATE TABLE query_tasks;
DELETE FROM query_jobs WHERE status IN (2, 3, 4, 5, 6);
-- then restart the query-scheduler deployment
```

hoophalab:
# Description

The issue can happen when handle_cancelling_search_jobs cancels a lot of search jobs in one coroutine. If there are many jobs to cancel, the loop keeps doing work without letting reducer handler respond to requests, failing the liveness probe.

Most of the time is spent on waiting for Celery to actually terminate the tasks in a job, and celery only provides a sync blocking wait.

We solve this issue by yield between cancelling search jobs.

# Checklist

  breaking change.

# Validation performed

On helm charts
1. Insert 1000 jobs and tasks to the database with PENDING state
2. After query scheduler start dispatching jobs
3. Update each job to CANCELLING state
4. `nc -vz` to check whether 7000 port of that pod is open
5. query-scheduler logs shows repeated `Cancelled job {}`.
6. `while true; do date "+%H:%M:%S"; nc -w 2 -vz 10.244.0.204 7000; sleep 0.1; done` to drain OS's TCP buffer quickly

Before the fix
1. After about 100 `nc -vz` (10 seconds), nc starts failing
2. Readiness probe fails

After the fix
1. Same operation, nc no longer fails.

## Release Notes

* **Bug Fixes**
  * Improved job cancellation responsiveness by ensuring the scheduler can process other tasks fairly when handling multiple concurrent job cancellations.


junhaoliao (approved):
lgtm

i validated the raw scheduling cost with this local script:

```python
import asyncio
import time

ITER = 100_000

async def bench(with_yield: bool) -> float:
    t0 = time.perf_counter_ns()
    for _ in range(ITER):
        if with_yield:
            await asyncio.sleep(0)
    dt = time.perf_counter_ns() - t0
    return dt / ITER

async def main() -> None:
    noop = await bench(False)
    with_yield = await bench(True)

    # repeat once for stability
    total = 0.0
    for _ in range(5):
        total += await bench(True)

    print(f"no-op loop ns/iter: {noop}")
    print(f"sleep0 loop ns/iter: {with_yield}")
    print(f"sleep0 avg ns/iter (5x): {total / 5}")
    print(f"overhead (sleep0 - noop) us: {(with_yield - noop) / 1000}")

asyncio.run(main())
```

Observed output (Python 3.10.12, Linux 6.8.0-124-generic):

```text
no-op loop ns/iter: 8.52423
sleep0 loop ns/iter: 1393.86432
sleep0 avg ns/iter (5x): 1393.16006
overhead (sleep0 - noop) us: 1.38534009
```

Interpretation:
- the measured overhead is ~**1.4 µs** per yielded cancellation (in this environment)
- this is a microbenchmark of the scheduler primitive only; real workload includes DB/reducer operations, so absolute impact is usually dominated by those heavier steps
- practically, it is a small per-iteration cost that trades off a small extra scheduling overhead for improved starvation prevention

PR #2337, opened 2026-06-17
feat(clp-s::filter): Add `BitmapView` class for bitmaps passed across FFI boundaries.

gibber9809:
# Description

As part of the extensible indexing interface, we expect to handle bitmaps passed into C++ through an FFI layer. That will generally mean that we'll be passed a bitmap as an array plus a field indicating the size of the bitmap, and have to interpret it according to some convention.

The `BitmapView` class added in this PR is designed to work with such bitmaps, and supports arbitrary element sizes in the backing array via templates. This class implements standard bitmap methods over this backing array (`test_bit`/`set_bit`), and also offers a templated convenience method that allows users to pass a visitor function that visits and optionally modifies the state of every bit set to `1` in the bitmap.

# Checklist

  breaking change.

# Validation performed

* Added unit tests for `BitmapView`.

## Summary by CodeRabbit

* **New Features**
  * Introduced a `BitmapView` wrapper for bit-level reads and writes, including visitor-based filtering of already-set bits.
* **Tests**
  * Added unit tests covering creation validation, `test_bit` results, `set_bit` out-of-range handling, and `filter_set_bits` behavior across component boundaries.
* **Chores**
  * Updated the build configuration to include the new bitmap view component and its tests when testing is enabled.


components/core/src/clp_s/filter/BitmapView.hpp
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * Bits are assumed to be ordered within each array element from least-significant bit to
+ * most-significant bit. Bits beyond the end of the bitmap that fit into the array are expected to
+ * be zero.
+ */
+template <clp::IntegerType BitmapComponentType>
+class BitmapView {
+public:
+    // Static constants
+    static constexpr size_t cNumBitsPerComponent{sizeof(BitmapComponentType) * 8};
```
LinZhihao-723: I feel like this can be a private member. Is it possible that we're gonna need this outside of the member methods?

components/core/src/clp_s/filter/BitmapView.hpp
LinZhihao-723, line 1: Shall we put it in a util namespace?

components/core/src/clp_s/filter/BitmapView.hpp
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * Creates a bitmap over the given array.
+     *
+     * @param bitmap_array A pointer to the underlying array backing the bitmap.
+     * @param num_bits The number of bits in the bitmap.
+     * @return A result containing the newly created BitmapView, or an error code indicating the
+     * failure:
+     * - std::errc::invalid_argument if either `bitmap_array` is null or `num_bits` is zero.
+     */
+    [[nodiscard]] static auto create(BitmapComponentType* bitmap_array, size_t num_bits)
```
LinZhihao-723: To be safer, I think we should use `std::span` instead of a raw pointer.
- Span should be the default representation for a non-owning array. It's also the way to avoid pointer arithmetic.
- When passing the array across the FFI layer, it's worth not just passing the pointer but also the length of the underlying array, along with the number of bits.

components/core/src/clp_s/filter/BitmapView.hpp
```diff
@@ -0,0 +1,173 @@
...
+            -> ystdlib::error_handling::Result<BitmapView> {
+        if (nullptr == bitmap_array || 0 == num_bits) {
+            return std::errc::invalid_argument;
```
LinZhihao-723, line 55: Shall we define our own error code? I prefer us to fully move away from the `std::errc` since when an error happens, the top-level caller that handles the error may lose context on exactly which component raises the error.

components/core/src/clp_s/filter/BitmapView.hpp
```diff
@@ -0,0 +1,173 @@
...
+     * - std::errc::result_out_of_range if the index is beyond the end of the bitmap.
+     */
+    [[nodiscard]] auto set_bit(size_t index, bool value) -> ystdlib::error_handling::Result<void> {
+        if (index >= m_num_bits) {
+            return std::errc::result_out_of_range;
+        }
+
+        auto const component_idx{index / cNumBitsPerComponent};
+        auto const bit_offset{index % cNumBitsPerComponent};
+        auto const bit_mask{static_cast<BitmapComponentType>(1) << bit_offset};
```
LinZhihao-723: Nit: maybe we can have a private method to return a tuple of these three + index validation?

components/core/src/clp_s/filter/BitmapView.hpp
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * Concept that defines a method that visits each set bit in a bitmap and determines whether it
+ * should remain set or become unset.
+ * @param bit_handler
+ * @param bit_idx The index of the set bit in the bitmap.
+ * @return A boolean result indicating whether the bit should remain set, or an error code
+ * indicating the failure.
+ */
+template <typename SetBitVisitor>
+concept SetBitVisitorConcept = requires(SetBitVisitor bit_handler, size_t bit_idx) {
```
LinZhihao-723: Forgot to add this in the convention doc...
I think we should rename it `SetBitVisitorReq` to meet our standard.

components/core/src/clp_s/filter/BitmapView.hpp
```diff
@@ -0,0 +1,173 @@
...
+                            : cNumBitsPerComponent
+            };
+            for (size_t bit_idx{}; bit_idx < max_bit; ++bit_idx) {
+                auto const current_bit{static_cast<BitmapComponentType>(1) << bit_idx};
+                if (0 != (current_component & current_bit)) {
+                    auto const current_idx{bit_idx + (cNumBitsPerComponent * component_idx)};
+                    if (YSTDLIB_ERROR_HANDLING_TRYX(set_bit_visitor(current_idx))) {
+                        new_component |= current_bit;
+                    }
+                }
```
LinZhihao-723: Nit:
```suggestion
                if (0 == (current_component & current_bit)) {
                    continue;
                }
                auto const current_idx{bit_idx + (cNumBitsPerComponent * component_idx)};
                if (YSTDLIB_ERROR_HANDLING_TRYX(set_bit_visitor(current_idx))) {
                    new_component |= current_bit;
                }
```

components/core/src/clp_s/filter/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 set(
         CLP_S_FILTER_SOURCES
+	BitmapView.hpp
```
LinZhihao-723: ops, catch a tab

LinZhihao-723 (changes requested):
Reviewed the implementation. Haven't checked the tests yet.

components/core/src/clp_s/filter/BitmapView.hpp
gibber9809: Kind of -- I think if we add utility methods to do bitwise combinations of multiple bitmaps we might want to make that work for `BitmapView` classes with differently sized underlying components. May have to use friend classes for that and/or expose more public APIs for that anyway though, so I'd be fine making this private for now.

components/core/src/clp_s/filter/BitmapView.hpp
gibber9809: Do you mean into a utils subdirectory + namespace here, just a utils namespace here, or moving it elsewhere in the codebase as a general utility?

components/core/src/clp_s/filter/BitmapView.hpp
gibber9809: Going to leave this until after I'm clear on whether/where you want me to move this code.

Not sure if I agree that this disambiguates anything in a useful way for error handling/debugging anyway though -- I think it would actually be harder to work with if every container-like class defined its own error codes for common things like out of range access; I could see an argument for creating a shared error code enum in ystdlib or something that provides common error codes for container-like classes though.

LinZhihao-723: > I could see an argument for creating a shared error code enum in ystdlib or something that provides common error codes for container-like classes though.

sgtm.

components/core/src/clp_s/filter/BitmapView.hpp
LinZhihao-723: I was thinking of moving it somewhere in the codebase as a general utility.
But I guess we should eventually move these things into ystdlib... so leaving it under `filter` namespace for now is probably ok.

components/core/src/clp_s/filter/BitmapView.hpp
```diff
@@ -0,0 +1,188 @@
...
+ */
+template <typename SetBitVisitor>
+concept SetBitVisitorReq = requires(SetBitVisitor bit_handler, size_t bit_idx) {
+    { bit_handler(bit_idx) } -> std::same_as<ystdlib::error_handling::Result<bool>>;
+};
+
+/**
+ * Non-owning view over a bitmap backed by an external array.
+ *
+ * Bits are assumed to be ordered within each array element from least-significant bit to
```
LinZhihao-723:
```suggestion
 * Bits are assumed to be ordered within each array element from the least-significant bit to the
```

components/core/src/clp_s/filter/BitmapView.hpp
```diff
@@ -0,0 +1,188 @@
...
+    // Factory methods
+    /**
+     * Creates a bitmap over the given array.
+     *
+     * @param bitmap_span A span of the underlying data backing the bitmap.
+     * @param num_bits The number of bits in the bitmap.
+     * @return A result containing the newly created BitmapView, or an error code indicating the
+     * failure:
+     * - std::errc::invalid_argument if either `bitmap_span` is too small for the number of bits,
+     * or `num_bits` is zero.
```
LinZhihao-723:
```suggestion
     *   or `num_bits` is zero.
```

components/core/src/clp_s/filter/BitmapView.hpp
```diff
@@ -0,0 +1,188 @@
...
+    /**
+     * @return The number of bits in the bitmap.
+     */
+    [[nodiscard]] auto get_num_bits() const -> size_t { return m_num_bits; }
+
+    /**
+     * @param bit_idx
+     * @return The value of the bit at the requested position, or an error code indicating the
+     * failure:
+     * - Forwards get_component_idx_and_bitmask's return values on failure.
```
LinZhihao-723:
```suggestion
     * - Forwards `get_component_idx_and_bitmask`'s return values on failure.
```

components/core/src/clp_s/filter/BitmapView.hpp
```diff
@@ -0,0 +1,188 @@
...
+                = YSTDLIB_ERROR_HANDLING_TRYX(get_component_idx_and_bitmask(bit_idx));
+        return 0 != (m_bitmap_span[component_idx] & bit_mask);
+    }
+
+    /**
+     * Sets a bit at a given index to the given value.
+     * @param bit_idx
+     * @param value
+     * @return A void result on success, or an error code indicating the failure:
+     * - Forwards get_component_idx_and_bitmask's return values on failure.
```
LinZhihao-723:
```suggestion
     * - Forwards `get_component_idx_and_bitmask`'s return values on failure.
```

components/core/src/clp_s/filter/BitmapView.hpp
```diff
@@ -0,0 +1,188 @@
...
+            m_bitmap_span[component_idx] &= ~bit_mask;
+        }
+        return ystdlib::error_handling::success();
+    }
+
+    /**
+     * Runs a visitor function over every set bit in the bitmap, modifying the state of each such
+     * bit according to the return value of the visitor.
+     *
+     * Note: on error, some updates for set bits may be lost.
```
LinZhihao-723:
```suggestion
     * Note: On error, some updates for set bits may be lost.
```

LinZhihao-723 (changes requested):
Some nit docstring fixes. Otherwise the implementation lgtm.
Will do a brief check on the test cases soon.

components/core/src/clp_s/filter/tests/test-clp_s-bitmap_view.cpp
LinZhihao-723, line 1: lgtm.

components/core/src/clp_s/filter/BitmapView.hpp
gibber9809: Created an issue to track #2368

LinZhihao-723 (approved):
For the PR title, how about:
```
feat(clp-s::filter): Add `BitmapView` class for bitmaps passed across FFI boundaries.
```

gibber9809:
CI failure is unrelated to this PR -- macos build is timing out right now since it's doing an uncached clang-tidy check over the whole repo for some reason.

PR #2342, opened 2026-06-19
feat(clp-s): Support count and count-by-time aggregation output to stdout.

davemarco:
# Description

Adds --count and --count-by-time aggregations to the clp-s search stdout output handler. Results are emitted per archive as newline-delimited JSON.

--count writes a single object per archive:

`clp-s s archives-dir 'level: ERROR' --count`
`{"archive_id":"<id>","count":42}`

--count-by-time writes one object per time bucket:

`clp-s s archives-dir 'level: ERROR' --count-by-time 1000`
`{"archive_id":"<id>","timestamp":1700000000000,"count":7}`

Since stdout is the default handler, the flags can be used without naming it (e.g. clp-s s --count). 

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed
On MongoDb dataset, --count matched grep across several queries, and --count-by-time bucket counts summed to the total as well 

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **New Features**
  * Added centralized search aggregation controls for counting results overall or by time bucket.
  * Standard output can now emit JSON-formatted aggregated counts (overall total or per time bucket).
* **Bug Fixes**
  * Improved validation to reject aggregation options for output handlers that don’t support them, and tightened aggregation parsing consistency.
* **Documentation**
  * Updated search help/visible options to reflect the new aggregation command format.

PR #2345, opened 2026-06-22
feat(clp-s): Support per-archive min and max aggregation output to stdout and the results cache.

davemarco:
# Description

Adds --min <field> and --max <field> search aggregations, computing the minimum/maximum value of a numeric field and writing the result to stdout or the results cache. Continues the aggregation work from #2326 and #2342. 

Outputs one document per archive, for both destinations:

{archive_id, field, min|max}

### Changes 
- Adds --min FIELD / --max FIELD, supported by both the stdout and results-cache destinations.
- New MinMaxAggregation reads the field from each matched record and tracks the running min/max
- Compares int64_t and double operands exactly
- Reworks aggregation output into a aggregator + sink design that scales as more aggregations/sinks are added

# Checklist

  breaking change.

# Validation performed

Validated a few queries on mongoDb dataset 

## Summary by CodeRabbit

* **New Features**
  * Added search result aggregations: count, count-by-time (bucketed), and min/max for a selected numeric field.
  * Aggregation output can now be routed through output handlers, including writing results to both standard output and the results cache.
  * Command-line aggregation options are centralized, with clear validation for supported combinations.
* **Bug Fixes**
  * Improved min/max evaluation for large integers and mixed numeric types, plus safer field descriptor handling.
  * Count and count-by-time now correctly handle zero/empty cases.
* **Other**
  * Aggregation-not-supported remains non-fatal, allowing IR-stream searching to continue when applicable.


components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -12,6 +13,7 @@
 #include "../clp/type_utils.hpp"
 #include "../reducer/types.hpp"
 #include "FileReader.hpp"
+#include "search/ast/SearchUtils.hpp"
```
davidlion:
```suggestion
#include <clp_s/search/ast/SearchUtils.hpp>
```

For new include statements let's follow the new guidelines to always use <>.

components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -199,7 +203,15 @@ auto collect_subcommands(
         if (false == current_subcommand.has_value()
             || false == current_subcommand_arguments.has_value())
         {
-            throw std::invalid_argument(fmt::format("unrecognized option \"{}\"", option));
+            // A leading option token (e.g. `--count`) with no named subcommand selects the default
+            // subcommand, letting its options be specified without naming the subcommand.
```
davidlion: Don't add inline comments unless they're necessary to explain some obscure behaviour we can't fix with refactoring/renaming.

Implementation details should go above the function definition and any important high level details should be in the doc string. This sounds like it should be in the doc string (plus default_subcommand wasn't added to the doc string).

components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -38,7 +40,8 @@ constexpr std::string_view cStdoutCacheOutputHandlerName{"stdout"};
  */
 auto collect_subcommands(
         std::map<std::string, po::options_description const*> const& subcommands,
-        std::vector<std::string> const& options
+        std::vector<std::string> const& options,
+        std::optional<std::string> const& default_subcommand = std::nullopt
```
davidlion: Update doc string with this param + update description.

components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -860,6 +886,31 @@ CommandLineArguments::parse_arguments(int argc, char const** argv) {
                     "File output path"
             );
 
+            StdoutOutputHandlerOptions stdout_options{};
+            po::options_description stdout_output_handler_options("Stdout Output Handler Options");
+            // clang-format off
```
davidlion: How ugly is it if we just let clang-format do it now? I feel like this might be unnecessary since it is likely a from very early on.

components/core/src/clp_s/CommandLineArguments.hpp
```diff
@@ -34,6 +34,8 @@ class CommandLineArguments {
     enum class AggregationType : uint8_t {
         Count,
         CountByTime,
+        Min,
+        Max,
```
davidlion:
```suggestion
        Max,
        Min,
```

super minor, but let's keep it alphabetical at least for now

components/core/src/clp_s/CommandLineArguments.hpp
```diff
@@ -42,6 +44,10 @@ class CommandLineArguments {
         std::string dataset;
         uint64_t batch_size{1000};
         uint64_t max_num_results{1000};
+        std::optional<AggregationType> aggregation_type;
+        int64_t count_by_time_bucket_size_ms{};
+        // Field whose value is aggregated by `--min`/`--max`.
+        std::string aggregation_field;
```
davidlion: I think we can improve this.

Two options I can think of are:
1. Make a variant where each aggregator is a struct with the relevant fields.
2. Make an interface where each implementation is an aggregator.

I think since the data is so simple, it is fine to just start off with the variant. Iirc this should replace `AggregationType` and allow us to use std::visit call the correct aggregator function with its respective data.

If you have better ideas, I'm open to anything.

components/core/src/clp_s/CommandLineArguments.cpp
davidlion: I think this shows up in a few other places. I know AI tends to write a lot of redundant comments.

davidlion (changes requested):
Quick pass to try and speed things up.

components/core/src/clp_s/CommandLineArguments.cpp
davemarco: removed

components/core/src/clp_s/CommandLineArguments.cpp
davemarco: see stdout pr

components/core/src/clp_s/CommandLineArguments.hpp
davemarco: PR is very different now, see description

components/core/src/clp_s/ResultsCacheUtils.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <mongocxx/client.hpp>
+#include <mongocxx/collection.hpp>
+
+#include <clp_s/ErrorCode.hpp>
+#include <clp_s/TraceableException.hpp>
+
+namespace clp_s {
+class ResultsCacheConnectionError : public TraceableException {
+public:
+    ResultsCacheConnectionError(ErrorCode error_code, char const* const filename, int line_number)
```
davidlion, line 16:
```suggestion
public:
    // Constructors
    ResultsCacheConnectionError(ErrorCode error_code, char const* const filename, int line_number)
```

components/core/src/clp_s/OutputHandlerImpl.hpp
```diff
@@ -7,18 +7,21 @@
...
 #include <mongocxx/client.hpp>
 #include <mongocxx/collection.hpp>
 
-#include <clp_s/CommandLineArguments.hpp>
-
 #include "../reducer/Pipeline.hpp"
 #include "../reducer/RecordGroupIterator.hpp"
+#include "Aggregation.hpp"
+#include "AggregationSink.hpp"
+#include "CommandLineArguments.hpp"
```
davidlion: Swap to `<clp_s/...`

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -5,53 +5,28 @@
...
 #include <msgpack.hpp>
-#include <nlohmann/json.hpp>
 #include <spdlog/spdlog.h>
 
 #include "../clp/networking/socket_utils.hpp"
 #include "../reducer/CountOperator.hpp"
 #include "../reducer/network_utils.hpp"
 #include "../reducer/Record.hpp"
 #include "archive_constants.hpp"
+#include "ResultsCacheUtils.hpp"
```
davidlion: `<clp_s/...`

components/core/src/clp_s/Aggregation.hpp
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Counts the number of matched records.
+ */
+class CountAggregation {
+public:
+    static constexpr bool cNeedsMetadata = false;
+    static constexpr bool cNeedsMarshalledRecord = false;
+
+    auto
+    add_record([[maybe_unused]] std::string_view message, [[maybe_unused]] epochtime_t timestamp_ms)
```
davidlion: I think for `add_record` it could take a variant of structs where each aggregator defines a struct of what args it needs (if an aggregator doesn't need args it doesn't need to define a struct and just leaves the variant param as maybe_unused).

This should be cleaner than passing in a superset of the possible params for all aggregators.

components/core/src/clp_s/aggregators.hpp
davidlion, line 1: Since this design has requirements on each aggregator class, is there a way to use templates/concepts to enforce the requirements?

components/core/src/clp_s/AggregationSink.hpp
```diff
@@ -0,0 +1,96 @@
...
+    AggregationSink() = default;
+
+    // Destructor
+    virtual ~AggregationSink() = default;
+
+    // Explicitly disable copy and move constructor/assignment
+    AggregationSink(AggregationSink const&) = delete;
+    auto operator=(AggregationSink const&) -> AggregationSink& = delete;
+    AggregationSink(AggregationSink&&) = delete;
+    auto operator=(AggregationSink&&) -> AggregationSink& = delete;
```
davidlion: Sorry, this particular change/rule is new and my fault for not getting around to updating the guidelines yet.

```suggestion
    // Delete copy constructor and assignment operator
    AggregationSink(AggregationSink const&) = delete;
    auto operator=(AggregationSink const&) -> AggregationSink& = delete;
    
    // Delete move constructor and assignment operator
    AggregationSink(AggregationSink&&) = delete;
    auto operator=(AggregationSink&&) -> AggregationSink& = delete;

    // Destructor
    virtual ~AggregationSink() = default;
```

davidlion (changes requested):

components/core/src/clp_s/aggregators.hpp
davemarco: switched to templated output handler

components/core/src/clp_s/Aggregation.hpp
davemarco: switched to templated output handler

components/core/src/clp_s/AggregationSink.hpp
davemarco: Applied the suggested ordering.

components/core/src/clp_s/OutputHandlerImpl.cpp
davemarco: switched the new project headers to <clp_s/…>.

components/core/src/clp_s/OutputHandlerImpl.hpp
davemarco: switched the new project headers to <clp_s/…>.

components/core/src/clp_s/CommandLineArguments.hpp
```diff
@@ -12,6 +12,7 @@
 #include <boost/program_options/variables_map.hpp>
 
 #include "../reducer/types.hpp"
+#include "Aggregation.hpp"
```
davidlion: style nit, use <clp_s/...

components/core/src/clp_s/aggregators.hpp
```diff
@@ -0,0 +1,142 @@
...
+ */
+using AggregationValue = std::variant<int64_t, double, std::string>;
+
+/**
+ * One aggregation result document: an ordered list of typed key-value pairs.
+ */
+using AggregationResult = std::vector<std::pair<std::string, AggregationValue>>;
+
+/**
+ * Requirements a type must satisfy to be used as an aggregator.
```
davidlion, line 29: I feel like some places (types/variables/etc) that say "Aggregation" should be "Aggregator", but my english is too bad to be certain. This comment essentially shows my feeling though as it uses "aggregator" when the actual types that adhere to the concept are named "CountAggregation" rather than "CountAggregator".

Potentially, the only changes are the classes that AggregatorReq applies to and anything related to them. I.e. maybe if something applies to all records it is an "aggregation", but if something applies to each record it is an "aggregator"?

Anyway, can you do a quick pass (or give me a quick explanation on why it is fine).

components/core/src/clp_s/Aggregation.hpp
```diff
@@ -0,0 +1,142 @@
...
+              { AggregatorType::cNeedsMarshalledRecord } -> std::convertible_to<bool>;
+          };
+
+/**
+ * Counts the number of matched records.
+ */
+class CountAggregation {
+public:
+    static constexpr bool cNeedsMetadata = false;
+    static constexpr bool cNeedsMarshalledRecord = false;
```
davidlion: style nit (shows up in other places too)
```suggestion
    static constexpr bool cNeedsMetadata{false};
    static constexpr bool cNeedsMarshalledRecord{false};
```

components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -1183,7 +1215,9 @@ void CommandLineArguments::parse_reducer_output_handler_options(
         throw std::invalid_argument("job-id cannot be negative.");
     }
 
-    if (false == m_aggregation_type.has_value()) {
+    if (false == m_aggregation.has_value()
+        || std::holds_alternative<MinMaxAggregation>(m_aggregation.value()))
+    {
```
davidlion: @coderabbitai Please make a issue to remove the old reducer code that shouldn't be used anymore.

components/core/src/clp_s/CommandLineArguments.cpp
davidlion: This was marked resolved, but the change was never applied...

components/core/src/clp_s/CommandLineArguments.cpp
```diff
@@ -772,15 +774,25 @@ CommandLineArguments::parse_arguments(int argc, char const** argv) {
             // clang-format on
             search_options.add(match_options);
 
+            int64_t count_by_time_bucket_size_ms{};
```
davidlion: My fault, recently we've been using `_millisecs` since ms is ambiguous (between milli and mega).
```suggestion
            int64_t count_by_time_bucket_size_millisecs{};
```

Since the fixes in other places should be small (and some code is touched in this PR anyway) I'm fine with doing all the renames in this PR. Otherwise, open an issue and do it in a separate very small PR.

Sorry about this.

components/core/src/clp_s/Aggregation.hpp
```diff
@@ -0,0 +1,142 @@
...
+#include <string_view>
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include <clp_s/Defs.hpp>
+
+namespace clp_s {
+/**
+ * A single typed value in an aggregation result document.
```
davidlion: I think with the way Aggregation is used in other places it should be possessive here (you also wrote it this way in another place too).

```suggestion
 * A single typed value in an aggregation's result document.
```

components/core/src/clp_s/Aggregation.hpp
```diff
@@ -0,0 +1,142 @@
...
+ */
+using AggregationValue = std::variant<int64_t, double, std::string>;
+
+/**
+ * One aggregation result document: an ordered list of typed key-value pairs.
```
davidlion:
```suggestion
 * One aggregation's result document: an ordered list of typed key-value pairs.
```

components/core/src/clp_s/Aggregation.cpp
```diff
@@ -0,0 +1,121 @@
...
+                string{field},
+                m_field_path,
+                descriptor_namespace
+        ))
+    {
+        throw std::invalid_argument("Invalid --min/--max field: " + string{field});
+    }
+    if (false == descriptor_namespace.empty()) {
+        // The tokenizer strips the namespace prefix out of the path, but namespaced fields live
+        // under a top-level object keyed by that namespace. Prepend it back so the path matches.
```
davidlion:
```suggestion
```

I think this counts as an implementation detail, so it goes above the function definition.

components/core/src/clp_s/int_float_compare.hpp
```diff
@@ -0,0 +1,94 @@
...
+    }
+    // `rhs` is in `int64_t` range, so its truncated value fits an `int64_t` with no rounding.
+    auto const truncated{std::trunc(rhs)};
+    auto const rhs_int{static_cast<int64_t>(truncated)};
+    if (lhs != rhs_int) {
+        return lhs < rhs_int;
+    }
+    // Equal integer parts: `lhs < rhs` only if `rhs` carries a fractional remainder.
+    return rhs > truncated;
+}
```
davidlion, line 64: * move fundamental comments to the doc-string (i.e. treating NaN as never less than anything)
* move implementation details above the function definition
* re-write the code to be more readable if it is not acceptable to remove any remaining comments

components/core/src/clp_s/int_float_compare.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Compares a double against an integer exactly.
+ * @param lhs
+ * @param rhs
+ * @return Whether `lhs < rhs`.
+ */
+[[nodiscard]] inline auto is_less(double lhs, int64_t rhs) -> bool {
```
davidlion, line 75: This doc-string should mirror the above. I think there is a doxygen command to refer to another function's doc-string, but I'm not sure we use it anywhere. Probably easier to just mirror the other doc-string once you update it.

davidlion (changes requested):
all minor naming and style changes

components/core/src/clp_s/Aggregation.hpp
davemarco: fixed

components/core/src/clp_s/Aggregation.hpp
davemarco: fixed

components/core/src/clp_s/Aggregation.hpp
davemarco: fixed

components/core/src/clp_s/CommandLineArguments.cpp
davemarco: sorry fixed

components/core/src/clp_s/CommandLineArguments.cpp
davemarco: fixed

components/core/src/clp_s/CommandLineArguments.hpp
davemarco: fixed

components/core/src/clp_s/int_float_compare.hpp
davemarco: done

components/core/src/clp_s/int_float_compare.hpp
davemarco: done

components/core/src/clp_s/aggregators.hpp
davemarco: i renamed to aggregator

components/core/src/clp_s/aggregators.hpp
davidlion, line 1: Sorry missed this before somehow, but the classes are missing the category comments.

components/core/src/clp_s/int_float_compare.hpp
davidlion, line 1: Rename to `int_float_compare.hpp` as files not dominated by a single "main" class should be `snake_case`.
(https://app.notion.com/p/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?source=copy_link#1a504e4d9e6b80aca8bdc1b2fbd1b1fa)

components/core/src/clp_s/aggregators.hpp
davidlion, line 1: Rename to `aggregators.hpp` as files not dominated by a single "main" class should be `snake_case`.
(https://app.notion.com/p/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?source=copy_link#1a504e4d9e6b80aca8bdc1b2fbd1b1fa)

This applies to `Aggregation.cpp -> aggregators.cpp` as well. I think "aggregators" feels better than "aggregation" as a file name after renaming the classes.

davidlion (changes requested):
very small changes and lgtm

components/core/src/clp_s/int_float_compare.hpp
```diff
@@ -0,0 +1,93 @@
...
+namespace clp_s {
+// Doubles span a far wider range than `int64_t`. These constants bracket the `int64_t` range,
+// letting the comparisons below detect out-of-range doubles before converting them to `int64_t`
+// (which would overflow).
+
+// `2^63`, one greater than `INT64_MAX`: any double `>=` this is larger than every `int64_t`. We use
+// `2^63` rather than `INT64_MAX` because `2^63` is exactly representable as a double and
+// `INT64_MAX` is not.
+constexpr double cInt64UpperBound{9223372036854775808.0};
+// `-2^63`, which equals `INT64_MIN` and is exactly representable as a double: any double `<` this
```
davidlion, line 27: Missed this. Comments should start after a newline (comments after namespace/class being exceptions).
```suggestion
constexpr double cInt64UpperBound{9223372036854775808.0};

// `-2^63`, which equals `INT64_MIN` and is exactly representable as a double: any double `<` this
```

davidlion:
I didn't think of this for the previous PRs, but what do you think of the title:
```
feat(clp-s): Support per-archive min and max aggregation output to stdout and the results cache.
```

davidlion (approved):

PR #2348, opened 2026-06-24
fix(build): Cap C++ compilation parallelism by memory limits.

jackluo923:
# Description

Compilation targets using `make -j` or `cmake --parallel` without a numeric cap would launch unlimited parallel jobs, exhausting RAM on machines with many cores but limited memory per core (e.g. 16-core/8 GB laptops, CI containers). This caused OOM kills during dependency builds.

Cap parallelism to `min(nproc, available_memory / 2 GB)` with a floor of 1, so each compilation job gets at least 2 GB. Detect available memory from: 1) cgroup v2 `memory.max` (Docker/Kubernetes containers), 2) cgroup v1 `memory.limit_in_bytes` (older containers), 3) `/proc/meminfo` `MemAvailable` (bare-metal Linux hosts), 4) Fall back to nproc with no memory cap (macOS, etc.)

Extract the computation into a shared shell helper (`tools/scripts/compute-cpp-max-parallelism.sh`) sourced by: `taskfile.yaml` (`G_CPP_MAX_PARALLELISM_PER_BUILD_TASK`), `components/core/tools/packaging/build.sh` (default `--cores`), 5 `lib_install` shell scripts (install-curl, libarchive, liblzma, lz4, zstandard).

Also update `build-and-run-unit-tests.py` with the same cgroup-aware logic, and update `build-in-container.sh` to let the taskfile compute parallelism when `CORES` is not explicitly set, so containers respect their own cgroup limits rather than inheriting a host-side value.

# Checklist

# Validation performed

* Tested in a 4 GB cgroup-limited Docker container: shared helper correctly returns 2 (4 GB / 2 GB per job = 2) instead of the host's 16 cores.
* Tested on host (16 cores, ~14 GB available): shared helper returns 6 (14 GB / 2 GB = 7, capped by `MemAvailable`).
* Tested `CLP_CPP_MAX_PARALLELISM_PER_BUILD_TASK` env var override: setting it to 8 correctly takes precedence over the computed value.
* Verified `task deps:core --dry --force` in Docker resolves `G_CPP_MAX_PARALLELISM_PER_BUILD_TASK` to the cgroup-aware value, and all `cmake --build` commands use `--parallel "2"` instead of unbounded `--parallel`.
* All modified shell scripts pass `bash -n` syntax checks. Python and YAML syntax checks pass.
* `MemAvailable: 0 kB` edge case correctly caps to 1 job (both shell and Python).

## Summary by CodeRabbit

* **New Features**
  * Build and unit test workflows now auto-select safer parallelism by combining CPU count with memory limits (including container/cgroup limits).
  * Added `--num-jobs` to override computed unit test build parallelism.
* **Bug Fixes**
  * Improved parallelism defaults when related environment variables are unset, and standardized CPU-core detection across install/build scripts.
  * Packaging/container builds now derive parallelism via a shared computation helper.
* **Documentation**
  * Updated build guidance to cap `make` parallelism when memory is constrained (now using `make -j$(nproc)` as the default example).


tools/scripts/compute-cpp-max-parallelism.sh
```diff
@@ -0,0 +1,80 @@
...
+        if [ -n "$cgroup_limit" ] && [ "$cgroup_limit" -lt 9223372036854771712 ] 2>/dev/null; then
+            mem_limit_kb=$((cgroup_limit / 1024))
+            mem_source="cgroup-v1"
+        fi
+    fi
+
+    # 3. Try /proc/meminfo (Linux hosts)
+    if [ -z "$mem_limit_kb" ]; then
+        local avail_kb
+        avail_kb=$(awk '/MemAvailable/ {print $2}' /proc/meminfo 2>/dev/null || true)
```
junhaoliao: to match the semantics of above mem limit, would it be better to use MemTotal instead of MemAvailable? otherwise the behaviour is also non deterministic as parallelism gets impacted by the load of the running machine, which varies due to a number of different factors

components/core/tools/packaging/build.sh
```diff
@@ -36,7 +36,9 @@ repo_root="$(cd "${script_dir}/../../../.." && pwd)"
 
 # Defaults
 format="all"
-cores="$(nproc 2>/dev/null || echo 4)"
+# Compute default parallelism based on CPU count and available memory (min 2 GB per core)
+source "${repo_root}/tools/scripts/compute-cpp-max-parallelism.sh"
+cores=$compute_cpp_max_parallelism_result
```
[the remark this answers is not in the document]
jackluo923: done

components/core/tools/scripts/lib_install/install-curl.sh
```diff
@@ -16,8 +16,10 @@ if [ ${EUID:-$(id -u)} -ne 0 ] ; then
   sudo echo "Script can elevate privileges."
 fi
 
-# Get number of cpu cores
-num_cpus=$(grep -c ^processor /proc/cpuinfo)
+# Get number of cpu cores, capped by available memory (min 2 GB per core)
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
+source "${script_dir}/../../../../../tools/scripts/compute-cpp-max-parallelism.sh"
+num_cpus=$compute_cpp_max_parallelism_result
```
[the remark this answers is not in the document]
jackluo923: done

components/core/tools/scripts/utils/build-and-run-unit-tests.py
```diff
@@ -85,15 +146,19 @@ def main(argv: list[str]) -> int:
...
+        f" / {MIN_MEMORY_PER_JOB_GB}) with a floor of 1.",
     )
     args_parser.add_argument("--test-spec", help="Catch2 test specification.")
 
     parsed_args = args_parser.parse_args(argv[1:])
     src_dir: Path = Path(parsed_args.source_dir)
     build_dir: Path = Path(parsed_args.build_dir)
     use_shared_libs: bool = parsed_args.use_shared_libs
-    num_jobs: int | None = parsed_args.num_jobs
+    num_jobs: int = parsed_args.num_jobs if parsed_args.num_jobs is not None else _compute_max_parallel_jobs()
```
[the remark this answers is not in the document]
jackluo923: done

tools/scripts/compute-cpp-max-parallelism.sh
jackluo923: done

tools/scripts/compute-cpp-max-parallelism.sh
```diff
@@ -0,0 +1,80 @@
...
+MIN_MEMORY_PER_JOB_GB=2
+# 2 GB in KB
+MIN_MEMORY_PER_JOB_KB=$((MIN_MEMORY_PER_JOB_GB * 1024 * 1024))
+
+compute_cpp_max_parallelism() {
+    local cpus
+    cpus=$(nproc 2>/dev/null || grep -c ^processor /proc/cpuinfo 2>/dev/null || echo 1)
+
+    local mem_limit_kb=""
+    local mem_source=""
```
junhaoliao: this is set but never read?

components/core/tools/packaging/build.sh
```diff
@@ -36,7 +36,10 @@ repo_root="$(cd "${script_dir}/../../../.." && pwd)"
 
 # Defaults
 format="all"
-cores="$(nproc 2>/dev/null || echo 4)"
+# Compute default parallelism based on CPU count and available memory (min 2 GB per core)
+source "${repo_root}/tools/scripts/compute-cpp-max-parallelism.sh"
```
junhaoliao: seems like we're calculating the host's parallelism and passing that into containers, which overrides any limit detections? then why do we want to read cgroup v2 limits in the containers?

components/core/tools/scripts/utils/build-and-run-unit-tests.py
```diff
@@ -27,6 +28,86 @@
...
 
+# Minimum memory (in GB) per compilation job.
+MIN_MEMORY_PER_JOB_GB = 2
+
+# cgroup v1 reports this sentinel value (2**63 - 4096) when memory is "unlimited", so any limit at
+# or above it is treated as unbounded.
+_CGROUP_V1_UNLIMITED_SENTINEL = 9223372036854771712
+
+
+def _get_cgroup_memory_limit_kb() -> int | None:
```
junhaoliao, line 39: have we considered calling the same bash script to compute parallelism to avoid any logic drifts?

components/core/tools/packaging/common/build-in-container.sh
```diff
@@ -4,22 +4,29 @@
...
 #
 # Required environment variables:
-#   CORES        Number of parallel build jobs
 #   FORMAT_DIR   Name of the format-specific packaging directory
 #   HOST_UID     UID to restore file ownership after build
 #   HOST_GID     GID to restore file ownership after build
+#
+# Optional environment variables:
+#   CORES        Number of parallel build jobs (default: computed based on CPU
+#                count and available memory, min 2 GB per core)
```
junhaoliao: shall we rename any "available" memory naming to "total memory" to match the implementation and avoid ambiguity?

components/core/tools/scripts/utils/build-and-run-unit-tests.py
```diff
@@ -27,6 +28,86 @@
...
+    memory. Each job gets at least MIN_MEMORY_PER_JOB_GB GB of memory, and the result is floored
+    at 1.
+
+    Memory detection order:
+      1. cgroup v2 memory.max (containers)
+      2. cgroup v1 memory.limit_in_bytes (containers)
+      3. /proc/meminfo MemTotal (Linux hosts)
+      4. Fall back to CPU count with no memory cap
+    """
+    cpu_count = os.cpu_count() or 1
```
junhaoliao: technically this is more accurate?

```suggestion
    cpu_count = os.process_cpu_count() or 1
```

components/core/tools/packaging/build.sh
junhaoliao:
```suggestion
#   --cores N       Parallel build jobs (default: auto, based on CPU count and memory limits)
```

components/core/tools/packaging/build.sh
junhaoliao: maybe a fix is to simply set cores="" here?

components/core/tools/scripts/utils/build-and-run-unit-tests.py
```diff
@@ -27,6 +28,86 @@
...
+    except (OSError, ValueError):
+        pass
+
+    return None
+
+
+def _compute_max_parallel_jobs() -> int:
+    """
+    Computes the maximum number of parallel compilation jobs based on CPU count and available
+    memory. Each job gets at least MIN_MEMORY_PER_JOB_GB GB of memory, and the result is floored
```
junhaoliao:
```suggestion
    Computes the maximum number of parallel compilation jobs based on CPU count and memory
    limits. Each job gets at least MIN_MEMORY_PER_JOB_GB GB of memory, and the result is floored
```

components/core/tools/packaging/common/build-in-container.sh
junhaoliao: or "memory limits"

docs/src/dev-docs/components-core/index.md
```diff
@@ -111,9 +111,13 @@ the relevant paths on your machine.
 
 * Build:
   ```shell
-  make -j
+  make -j$(nproc)
```
junhaoliao:
```suggestion
  cmake --build . --parallel "$(getconf _NPROCESSORS_ONLN)"
```

docs/src/dev-docs/components-core/index.md
```diff
@@ -111,9 +111,13 @@ the relevant paths on your machine.
...
   ```
 
+  > **NOTE:** If the build runs out of memory, reduce parallelism by capping based on available
+  > memory (at least 2 GB per core). For example, on a machine with 8 GB of available memory:
+  > `make -j4`
```
junhaoliao:
```suggestion
  > **NOTE:** If the build runs out of memory, reduce parallelism by capping based on total memory
  > (at least 2 GB per core). For example, on a machine with 8 GB of memory:
  > `cmake --build . --parallel 4`
```

junhaoliao (changes requested):
for the title, how about:
```
fix(build): Cap C++ compilation parallelism by memory limits.
```

tools/scripts/compute-cpp-max-parallelism.sh
jackluo923: You are right, removed mem_source varilable as well as the 3 places where the field is set.

components/core/tools/packaging/build.sh
jackluo923: sure, I configured it as core="" so container can compute its own.

components/core/tools/scripts/utils/build-and-run-unit-tests.py
jackluo923: This seems to be a Python 3.13+ only feature, so replicated the functionality manaully: 
` getattr(os, "process_cpu_count", os.cpu_count)`

components/core/tools/packaging/common/build-in-container.sh
jackluo923: changed it to memory limits

components/core/tools/scripts/utils/build-and-run-unit-tests.py
jackluo923: removed the bash script entirely so we don't have Python logic and bash logic.

components/core/tools/scripts/utils/build-and-run-unit-tests.py
```diff
@@ -27,6 +28,90 @@
...
+      1. cgroup v2 memory.max (containers)
+      2. cgroup v1 memory.limit_in_bytes (containers)
+      3. /proc/meminfo MemTotal (Linux hosts)
+      4. Fall back to CPU count with no memory cap
+    """
+    # os.process_cpu_count() (Python 3.13+) respects taskset/cgroup CPU limits, so it's more
+    # accurate than os.cpu_count() in containers; fall back to the latter on older Pythons (the
+    # project supports >=3.10).
+    get_cpu_count = getattr(os, "process_cpu_count", os.cpu_count)
+    cpu_count = get_cpu_count() or 1
```
junhaoliao:
```suggestion
    try:
        cpu_count = len(os.sched_getaffinity(0))
    except (AttributeError, OSError):
        cpu_count = os.cpu_count() or 1
```

taskfile.yaml
```diff
@@ -47,9 +47,18 @@ vars:
...
+  # NOTE: If CLP_CPP_MAX_PARALLELISM_PER_BUILD_TASK is set, it takes precedence. Otherwise, we cap
+  # parallelism at min(cpu_count, memory_limit_GB / 2) with a floor of 1, so that each compilation
+  # job gets at least 2 GB of memory. The computation is delegated to build-and-run-unit-tests.py
+  # (via --compute-max-parallel-jobs) to keep a single source of truth.
+  G_CPP_MAX_PARALLELISM_PER_BUILD_TASK:
+    sh: |
+      if [ -n "${CLP_CPP_MAX_PARALLELISM_PER_BUILD_TASK:-}" ]; then
+        echo "${CLP_CPP_MAX_PARALLELISM_PER_BUILD_TASK}"
+      else
+        python3 "{{.ROOT_DIR}}/components/core/tools/scripts/utils/build-and-run-unit-tests.py" \
```
junhaoliao, line 59: a cleaner fix might be to extract a python script for parallelism computation though i'm really fine to defer the refactoring into another pr... if you agree, maybe ask the rabbit to create an issue to track

junhaoliao (dismissed):

components/core/tools/scripts/utils/build-and-run-unit-tests.py
junhaoliao: AttributeError is for windows compatibility ("non-unix" as documented by https://docs.python.org/3/library/os.html#interface-to-the-scheduler )

OSError is for system errors in general

junhaoliao (approved):

taskfile.yaml
jackluo923: @coderabbitai create a new github issue to refactor parallelism computation into a dedicated script.

PR #2289, opened 2026-05-20
feat(webui): Add configurable search result limit to CLP search UI.

hoophalab:
# Description

1. Add a MaxResultsSelect dropdown allowing users to choose the number of search results (10-10,000) instead of using the hardcoded default of 1000.
2. The selected limit is passed through the search state, query submission and backend API.
3. Use flex on query row.

limit

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

1. search button and dragging on timeline submits query with limit
2. the number of rows in the tables are the same as the limit

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **New Features**
  * Added a dropdown to set the maximum number of search results per query; the chosen limit is applied to submissions, results listing and timeline zooming.

* **Style**
  * Improved search control layout and responsiveness (flex wrapping, min/max widths, input sizing).
  * Refined select control and input sizing for a more consistent UI.

* **Bug Fixes**
  * Ensured the configured results limit is consistently respected when submitting and finalizing searches.

PR #2350, opened 2026-06-24
feat(webui): Make Presto search result limit configurable and warn on truncation (resolves #2330).

### LINKED ISSUE #2330: WebUI: Presto guided search result limit (1000) should be user-configurable
## Problem

The Presto search results are capped at **1000 rows** regardless of the SQL `LIMIT` clause. This affects both guided and raw SQL query modes.

There are **two** hardcoded limits:

### 1. Client-side (guided mode only)
`components/webui/client/src/pages/SearchPage/SearchControls/Presto/Guided/presto-guided-search-requests.ts`
```typescript
const DEFAULT_SEARCH_LIMIT = 1000;
```
This sets `LIMIT 1000` in the generated SQL for guided mode.

### 2. Server-side (affects ALL modes including raw SQL)
`components/webui/server/src/routes/api/presto-search/typings.ts`
```typescript
export const MAX_PRESTO_SEARCH_RESULTS = 1000;
```
This caps the number of rows stored in MongoDB from Presto query results. Even if the user sets `LIMIT 10000` in raw SQL mode, only the first 1000 rows are persisted and displayed.

The enforcement logic in `routes/api/presto-search/index.ts`:
```typescript
if (storedResultsCount < MAX_PRESTO_SEARCH_RESULTS) {
    const remainingSlots = MAX_PRESTO_SEARCH_RESULTS - storedResultsCount;
    const dataToInsert = data.slice(0, remainingSlots);
    // ...
}
```

## Expected Behavior

- The server-side limit should respect the user's SQL `LIMIT` clause (or be configurable via settings.json / env var)
- The guided mode limit should be user-adjustable in the UI
- When results are truncated by the server cap, the UI should clearly indicate this

20001020ycx:
# Description

The number of Presto search results stored and shown in the WebUI was hardcoded to 1000, capping the raw SQL queries regardless of their `LIMIT`. This PR makes the cap configurable and surfaces a non-blocking warning when results are truncated.

# Checklist

  breaking change.

# Validation performed

In the current WebUI, when issuing a raw SQL query to Presto, only 1,000 results are rendered even though the query returned 10,000 rows in accordance with the user's LIMIT clause:
image

 In this PR, we make the rendered-row cap configurable via `webui.presto_max_num_search_results` and display a warning when results are truncated:
image

I also tested the case where the raw SQL query has no LIMIT clause: the warning appears once the configured limit is reached, while Presto continues fetching the remaining rows. I think this behaviour is good: the user still sees the total number of matched rows, while only the capped subset is rendered, avoiding overwhelming the UI-rendering when returning the full result set.

### Helm chart validation

Since this PR also adds the new setting through the Helm chart (`values.yaml`, `configmap.yaml`), I validated the chart rendering with `helm template`:

- `helm template clp tools/deployment/package-helm` renders `presto_max_num_search_results` into the generated `clp-config.yaml`, and `PrestoMaxNumSearchResults` into both the client and server `settings.json` (all defaulting to `1000`).
- Overriding the value (`--set clpConfig.webui.presto_max_num_search_results=5000`) propagates `5000` to all three rendered outputs, confirming it is configurable end-to-end through the chart.
- `helm lint` passes.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable limit for Presto search results in the Web UI.
  * Display a warning when search results are truncated, including the total matched count and current limit.

* **Bug Fixes**
  * Search result storage and display now respect the same configurable maximum across the app.

* **Chores**
  * Updated default configuration and deployment values to include the new setting.
  * Bumped the package chart version.


20001020ycx:
@junhaoliao should we support pagination on the rendered result in the future?

components/package-template/src/etc/clp-config.template.json.yaml
```diff
@@ -120,6 +120,7 @@ telemetry:
 #  port: 4000
 #  results_metadata_collection_name: "results-metadata"
 #  rate_limit: 1000
+#  default_max_num_query_results: 1000
```
hoophalab: Since it only modifies presto, how about default -> presto. Also how about query -> search to keep consistent with the other var names in code
```suggestion
#  presto_max_num_search_results: 1000
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -764,6 +764,7 @@ class WebUi(BaseModel):
     port: Port = DEFAULT_PORT
     results_metadata_collection_name: NonEmptyStr = "results-metadata"
     rate_limit: PositiveInt = 1000
+    default_max_num_query_results: PositiveInt = 1000
```
hoophalab: naman uses helm charts. So we probably need to modify `values.yaml` and `configmap.yaml` as well

components/webui/packages/client/src/config/index.ts
```diff
@@ -7,6 +7,7 @@ import {
 import {settings} from "../settings";
 
 
+const SETTINGS_DEFAULT_MAX_NUM_QUERY_RESULTS = settings.DefaultMaxNumQueryResults;
```
hoophalab: Can we also modify the sql builder to use this value?
https://github.com/y-scope/clp/blob/main/components/webui/packages/client/src/pages/SearchPage/SearchControls/Presto/Guided/presto-guided-search-requests.ts#L85-L94

hoophalab (commented):
some comments

components/clp-py-utils/clp_py_utils/clp_config.py
20001020ycx: Awesome catch, I've added the new config to `values.yaml` and `configmap.yaml`, and bumped the chart version.

I am not a K8s expert, do you think this needs a full K8s cluster deployment to validate, or is helm template (verifying the value renders and is overridable) enough?

components/webui/packages/client/src/config/index.ts
20001020ycx: I think this cap is reserved for the guided mode, where a `LIMIT` clause of 1000 is always appended when querying Presto. Given that guided mode never expose the `LIMIT` to user, we are not violating any semantics as what raw SQL mode did. 

However, I do a agree this is something that we should address for a better user experience, and I will handle it in a follow up PR:
1)  Add Limit selection in the guided mode, similar to what this PR did: https://github.com/y-scope/clp/pull/2289
2)  Wire the config into it and apply the same truncation-warning approach from this PR.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -764,6 +764,7 @@ class WebUi(BaseModel):
     port: Port = DEFAULT_PORT
     results_metadata_collection_name: NonEmptyStr = "results-metadata"
     rate_limit: PositiveInt = 1000
+    presto_max_num_search_results: PositiveInt = 1000
```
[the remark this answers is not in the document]
20001020ycx: `default_max_num_query_results` is for API server, `presto_max_num_search_results` is the new parameter I added in the PR for webui only. I dont understand how this would incur backward compatibility problem, can you double check?

components/webui/packages/client/src/config/index.ts
20001020ycx: @coderabbit, please create a issue to track this, and assign it to me

20001020ycx: @coderabbitai, please create a issue to track this, and assign it to me, thank you.

20001020ycx:
@hoophalab, I have resolved all conflicts caused by the recent refactor pushed by main, and the validation performed in the PR description still holds, please take a look again, thank you!

hoophalab (dismissed):
LGTM.

Validation: The WebUI + Presto deployment on Helm behaves the same as before, and the new warning is displayed correctly.

One remaining comment:
Using an old `values.yaml` will prevent the deployment from starting because the new config field is now required -- this essentially makes this PR "breaking" if we regard configuring charts by providing a standalone `values.yaml` is a normal user behavior. However, this issue isn't introduced by this PR -- none of the existing Helm values have default values either. Per @junhaoliao's decision, should we add default values in configmap.yaml in a follow-up?

tools/deployment/package-helm/templates/configmap.yaml
```diff
@@ -223,6 +223,8 @@ data:
       host: "localhost"
       port: 4000
       rate_limit: {{ .Values.clpConfig.webui.rate_limit | int }}
+      presto_max_num_search_results: {{ .Values.clpConfig.webui.presto_max_num_search_results
+        | int }}
       results_metadata_collection_name: {{ .Values.clpConfig.webui.results_metadata_collection_name
         | quote }}
```
junhaoliao: nit
```suggestion
      presto_max_num_search_results: {{ .Values.clpConfig.webui.presto_max_num_search_results
        | int }}
      rate_limit: {{ .Values.clpConfig.webui.rate_limit | int }}
      results_metadata_collection_name: {{ .Values.clpConfig.webui.results_metadata_collection_name
        | quote }}
```

tools/deployment/package-helm/values.yaml
```diff
@@ -259,6 +259,7 @@ clpConfig:
     port: 30000
     results_metadata_collection_name: "results-metadata"
     rate_limit: 1000
+    presto_max_num_search_results: 1000
```
junhaoliao:
```suggestion
    presto_max_num_search_results: 1000
    rate_limit: 1000
    results_metadata_collection_name: "results-metadata"
```

components/webui/packages/client/src/pages/SearchPage/index.tsx
```diff
@@ -25,6 +26,7 @@ const SearchPage = () => {
             {SETTINGS_QUERY_ENGINE === CLP_QUERY_ENGINES.PRESTO && <ProgressBar/>}
             <div className={styles["searchPageContainer"]}>
                 <SearchControls/>
+                <ResultsLimitWarning/>
```
junhaoliao: if we don't plan to extend this to the other engines
```suggestion
                {SETTINGS_QUERY_ENGINE === CLP_QUERY_ENGINES.PRESTO && <ResultsLimitWarning/>}
```

junhaoliao: actually i think we can reuse it. maybe we should add a note to the docstring of ResultsLimitWarning?

components/webui/packages/client/src/pages/SearchPage/ResultsLimitWarning/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+    SETTINGS_QUERY_ENGINE,
+} from "../../../config";
+import useSearchStore from "../SearchState";
+import {SEARCH_UI_STATE} from "../SearchState/typings";
+
+
+/**
+ * Displays a warning when the number of matched results exceeds the configured maximum, indicating
+ * that the displayed results have been truncated and how to raise the limit.
+ *
```
junhaoliao, line 15:
```suggestion
 *
 * NOTE: Currently this is only supported for Presto.
 *
```

junhaoliao (changes requested):
the rest lgtm

components/webui/packages/client/src/pages/SearchPage/index.tsx
junhaoliao: https://github.com/y-scope/clp/pull/2350#discussion_r3492943493

junhaoliao (approved):

components/webui/packages/client/src/pages/SearchPage/index.tsx
20001020ycx: Why don't we commit as what you have suggested above to be align with https://github.com/y-scope/clp/pull/2350#discussion_r3492943493, and keep track issue for non-presto route to use this warning component.

20001020ycx: @coderabbitai, please create an issue to track exporting this warning component to non-presto route.

junhaoliao (commented):
we should mark this PR as BREAKING because end users may observe less results than before. also, we should note the values.yaml schema change

1. should modify the description to explain the BREAKING change
2. for the title, how about
	```
	feat(webui)!: Make Presto search result limit configurable and warn on truncation (resolves #2330); Cap results cache fetching limit for Presto at 1000 by default.
	```

20001020ycx:
@junhaoliao As discussed offline, the title is reverted with no breaking change annotation.

junhaoliao (approved):

PR #2357, opened 2026-06-26
feat(docker): Consolidate per-arch manylinux/musllinux dependency images into multi-arch images.

jackluo923:
# Description

Consolidates the four architecture-specific base-image directories (`clp-env-base-{manylinux_2_28,musllinux_1_2}-{aarch64,x86_64}`) into two multi-arch directories that select the target architecture via Docker buildx `--platform` (locally through a `PLATFORM` env var in `build.sh`). Both architectures now share a single Dockerfile/build.sh per image family.

The dependency images are published as multi-arch manifests on `main`: each architecture is built natively on its matching runner and published under an arch-suffixed tag (e.g. `:main-amd64`), then a single `deps-image-merge` matrix job combines the per-arch tags into one multi-arch `:main` manifest. Consumers pull `:main` and Docker auto-selects the host architecture. On PRs the images are still consumed as local tar artifacts (no publish/merge), so the merge job is gated to push-to-`main` only; each leg is gated on its own deps-image matrix result so a failure or skip of one OS doesn't block the other.

Other changes:
- `clp-core-build-containers` and `run-on-image` composite actions gain an `arch` input to carry the architecture through arch-suffixed tags/artifacts.
- `packaging/build.sh` builds the consolidated base image via `PLATFORM` and the three downstream binary/wheel jobs now `need` the merged `deps-image-merge` matrix job.
- Dev docs updated (pull commands, paths, workflow diagram, job descriptions).
- Fixes a Python 3.6 regression in the shared deb/rpm builder image: `dnf install dpkg` pulls in AlmaLinux's python3-3.6, which wins the `/usr/bin/python3` alternative and is too old for the auto-parallelism script (`build-and-run-unit-tests.py --compute-max-parallel-jobs`) invoked by the packaging task. Re-registers the manylinux `python3.12` at a higher alternatives priority to restore it as the default `python3`.

Migration note: the published ghcr repository name changes from separate per-arch repos (`clp-core-dependencies-{aarch64,x86}-{manylinux_2_28,musllinux_1_2}`) to a single multi-arch repo per family (`clp-core-dependencies-{manylinux_2_28,musllinux_1_2}`). The old per-arch ghcr repos will no longer receive `:main` updates; any external consumers pulling the old per-arch repo names should switch to the new `:main` multi-arch tag. No in-repo consumers reference the old names.

Scope note: ubuntu-jammy and centos-stream-9 dependency images are intentionally left on the legacy per-arch model (out of scope for this PR).

# Checklist

# Validation performed

* Built both consolidated base images locally — `clp-env-base-manylinux_2_28` and `clp-env-base-musllinux_1_2` — for amd64 and arm64 (via `PLATFORM`).
* Built `clp-s` end-to-end inside the new manylinux image.
* Built deb, rpm, and apk packages for aarch64 using `packaging/build.sh --format all --arch aarch64`; verified each package installs and `clp-s --help` runs in emulated arm64 containers (debian:bookworm, almalinux:9, alpine:3.20).
* Verified the Python 3.6 fix: in the rebuilt deb/rpm builder image, `python3 --version` reports 3.12.13 (was 3.6.8) and the taskfile auto-parallelism command (`build-and-run-unit-tests.py --compute-max-parallel-jobs`) runs successfully (previously failed with a `SyntaxError`).
* Validated the workflow YAML parses and traced `needs`/`if` skip-propagation across event scenarios (PR image-only, PR clp-only, push-to-main, schedule).
* Swept the repo for stale references to the deleted per-arch directories / image names / job names; none remain (only two still-accurate usage comments in packaging Dockerfiles that cite the preserved local tag name).

## Summary by CodeRabbit

* **New Features**
  * Added optional single-architecture targeting for local dependency container image builds/loads, with arch-aware naming for the produced artifacts.
  * Updated Linux dependency container publishing to build OS-level multi-arch manylinux_2_28 and musllinux_1_2 images (amd64/arm64) and merge per-architecture tags into multi-arch manifests.
* **Documentation**
  * Refreshed tooling and workflow docs to use the new multi-arch image names/paths and the revised dependency-image build/merge job structure.


.github/workflows/clp-artifact-build.yaml
```diff
@@ -164,71 +166,72 @@ jobs:
...
-
-  musllinux_1_2-x86_64-deps-image:
-    name: "musllinux_1_2-x86_64-deps-image"
-    if: "needs.filter-relevant-changes.outputs.musllinux_1_2_x86_64_image_changed == 'true'"
+        run: |
+          image_base="ghcr.io/${{steps.sanitization.outputs.REPOSITORY}}/${{env.DEPS_IMAGE_NAME_PREFIX}}${{matrix.os}}"
+          docker manifest create "${image_base}:main" \
+            "${image_base}:main-amd64" \
+            "${image_base}:main-arm64"
+          docker manifest push "${image_base}:main"
```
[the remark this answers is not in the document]
jackluo923: sure, applied the suggestion

components/core/tools/docker-images/clp-env-base-manylinux_2_28/Dockerfile
```diff
@@ -1,6 +1,8 @@
 # syntax=docker/dockerfile:1
 
-FROM quay.io/pypa/manylinux_2_28_aarch64
+# Multi-arch base image (amd64, arm64, ...). Build with `--platform` to select the
+# target architecture; see build.sh (PLATFORM env var for local cross-builds).
+FROM quay.io/pypa/manylinux_2_28
```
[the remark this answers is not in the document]
jackluo923: out of scope, will apply this in another PR

components/core/tools/docker-images/clp-env-base-musllinux_1_2/Dockerfile
```diff
@@ -1,6 +1,8 @@
 # syntax=docker/dockerfile:1
 
-FROM quay.io/pypa/musllinux_1_2_aarch64
+# Multi-arch base image (amd64, arm64, ...). Build with `--platform` to select the
+# target architecture; see build.sh (PLATFORM env var for local cross-builds).
+FROM quay.io/pypa/musllinux_1_2
```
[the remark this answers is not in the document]
jackluo923: out of scope, will apply this in another PR

.github/workflows/clp-artifact-build.yaml
```diff
@@ -350,7 +384,8 @@ jobs:
       success()
       || (!cancelled() && !failure() && needs.filter-relevant-changes.outputs.clp_changed == 'true')
```
junhaoliao: For an image-only PR, `clp_changed` is also false; does this job skip instead of loading the freshly built tar artifact?

```suggestion
      success()
      || (!cancelled() && !failure() && (
        needs.filter-relevant-changes.outputs.clp_changed == 'true'
        || needs.filter-relevant-changes.outputs.manylinux_2_28_image_changed == 'true'
      ))
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -393,7 +429,8 @@ jobs:
       ))
```
junhaoliao, line 433: ditto

```suggestion
        || needs.filter-relevant-changes.outputs.manylinux_2_28_image_changed == 'true'
      ))
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -443,7 +481,8 @@ jobs:
       success()
       || (!cancelled() && !failure() && needs.filter-relevant-changes.outputs.clp_changed == 'true')
```
junhaoliao: ditto

```suggestion
      success()
      || (!cancelled() && !failure() && (
        needs.filter-relevant-changes.outputs.clp_changed == 'true'
        || needs.filter-relevant-changes.outputs.musllinux_1_2_image_changed == 'true'
      ))
```

docs/src/dev-docs/tooling-gh-workflows.md
```diff
@@ -30,18 +30,18 @@ shown below.
...
     filter-relevant-changes --> manylinux_2_28-x86_64-binaries
     filter-relevant-changes --> musllinux_1_2-x86_64-binaries
     filter-relevant-changes --> ubuntu-jammy-binaries
     centos-stream-9-deps-image --> centos-stream-9-binaries
-    manylinux_2_28-x86_64-deps-image --> manylinux_2_28-x86_64-binaries
-    musllinux_1_2-x86_64-deps-image --> musllinux_1_2-x86_64-binaries
+    manylinux_2_28-deps-image --> manylinux_2_28-deps-image-merge
+    manylinux_2_28-deps-image-merge --> manylinux_2_28-x86_64-binaries
+    musllinux_1_2-deps-image --> musllinux_1_2-deps-image-merge
+    musllinux_1_2-deps-image-merge --> musllinux_1_2-x86_64-binaries
```
junhaoliao, line 46: would this be more accurate? or overly specific for the purpose

```suggestion
    manylinux_2_28-deps-image --> manylinux_2_28-deps-image-merge
    manylinux_2_28-deps-image --> manylinux_2_28-x86_64-binaries
    manylinux_2_28-deps-image-merge --> manylinux_2_28-x86_64-binaries
    musllinux_1_2-deps-image --> musllinux_1_2-deps-image-merge
    musllinux_1_2-deps-image --> musllinux_1_2-x86_64-binaries
    musllinux_1_2-deps-image-merge --> musllinux_1_2-x86_64-binaries
```

docs/src/dev-docs/tooling-containers.md
```diff
@@ -19,75 +20,50 @@ distros using glibc 2.28+, including:
...
-  ```
-
-* Path:
-
-  ```text
-  components/core/tools/docker-images/clp-env-base-manylinux_2_28-x86_64
-  ```
+  To cross-build for a non-native architecture locally (using QEMU emulation), set the `PLATFORM`
+  env var when running `build.sh`, e.g. `PLATFORM=linux/arm64 ./build.sh`
+  (requires binfmt/QEMU support on the host).
```
junhaoliao, line 38: for dev container readiness

```suggestion
  To cross-build for a non-native architecture locally (using QEMU emulation), set the `PLATFORM`
  env var when running `build.sh`, e.g. `PLATFORM=linux/arm64 ./build.sh`
  (requires binfmt/QEMU support on the host).

  Local builds keep the architecture-specific `:dev` tags used by the packaging scripts:
  `clp-core-dependencies-x86-manylinux_2_28:dev` for amd64 and
  `clp-core-dependencies-aarch64-manylinux_2_28:dev` for arm64.
```

docs/src/dev-docs/tooling-containers.md
```diff
@@ -19,75 +20,50 @@ distros using glibc 2.28+, including:
...
-  ```
-
-* Path:
-
-  ```text
-  components/core/tools/docker-images/clp-env-base-musllinux_1_2-x86_64
-  ```
+  To cross-build for a non-native architecture locally (using QEMU emulation), set the `PLATFORM`
+  env var when running `build.sh`, e.g. `PLATFORM=linux/arm64 ./build.sh`
+  (requires binfmt/QEMU support on the host).
```
junhaoliao, line 70: ditto

```suggestion
  To cross-build for a non-native architecture locally (using QEMU emulation), set the `PLATFORM`
  env var when running `build.sh`, e.g. `PLATFORM=linux/arm64 ./build.sh`
  (requires binfmt/QEMU support on the host).

  Local builds keep the architecture-specific `:dev` tags used by the packaging scripts:
  `clp-core-dependencies-x86-musllinux_1_2:dev` for amd64 and
  `clp-core-dependencies-aarch64-musllinux_1_2:dev` for arm64.
```

junhaoliao (changes requested):
lgtm

.github/workflows/clp-artifact-build.yaml
jackluo923: good catch. This is a regerssion and we should compile CLP inside the iamge to prove the image change is good.

docs/src/dev-docs/tooling-gh-workflows.md
jackluo923: Let's make it more detailed and accurate. Seems LLM also suggest adding more details. 

```
What the diagram claims to show. The doc text right below it says "Arrows between jobs indicate a dependency" — i.e., the diagram is a rendering of the workflow's needs graph. The actual needs of manylinux_2_28-x86_64-binaries after this PR is three jobs:

  needs:
    - "manylinux_2_28-deps-image"        # <-- direct edge missing from your diagram
    - "manylinux_2_28-deps-image-merge"
    - "filter-relevant-changes"

  Your diagram draws only the chain deps-image → merge → binaries, dropping the direct deps-image → binaries edge. That's a "transitive reduction" — fine if you only care about execution ordering, since the ordering is implied through the merge job.

  Why the dropped edge actually matters here. The merge job is skipped on every PR. The direct deps-image → binaries edge is precisely the path PRs run on: binaries consumes the tar artifact that deps-image uploaded. Meanwhile merge → binaries is the path that matters on
  push-to-main (binaries pulls the :main manifest the merge job pushes). So the two edges represent two genuinely different data flows, not a redundant shortcut — and after the if: fix you just accepted, the direct edge is exactly what an image-only PR exercises. A reader of the
  current diagram would reasonably conclude binaries always waits on a merge, which is wrong for PRs.

  The consistency argument seals it. The diagram already follows the "draw the literal needs graph, redundant edges included" convention: it has filter-relevant-changes --> manylinux_2_28-x86_64-binaries even though that's equally implied transitively through the deps-image
  chain. So the PR's version is the one deviating from the diagram's own convention — "overly specific" doesn't apply.

  My recommendation: accept his suggestion as-is (it's the literal needs edges for both families, nothing more). If you want the doc to also convey when each edge is live (tar on PRs vs :main manifest on main), that's already covered in the job descriptions below the diagram, so
  no further change needed.
  ```

junhaoliao (approved):

PR #2363, opened 2026-07-02
feat(clp-s::timestamp_parser): Add support for parsing RFC 2822 / 822 timestamp strings; Bump archive version to 0.5.1 (fixes #2027).

### LINKED ISSUE #2027: feat: Add support for parsing RFC 2822 / RFC 822 timestamp strings
## Feature Request

Add support for parsing RFC 2822 and RFC 822 timestamp formats in CLP-S timestamp parser.

## Background

Currently, CLP-S only accepts ISO8601-compliant timestamp formats. After PR `#1788` introduced stricter timestamp parsing, non-compliant formats like RFC 2822 / RFC 822 are rejected, causing compression errors.

## Use Case

Some datasets (e.g., "postgresql" logs) use RFC 2822 / RFC 822 timestamp formats. Supporting these formats would allow CLP to compress a wider range of log datasets.

## Acceptance Criteria

- CLP-S timestamp parser can parse RFC 2822 and RFC 822 timestamp strings
- Affected datasets can be restored to the resources documentation

## References

- PR: https://github.com/y-scope/clp/pull/1788
- Requested by: `@junhaoliao`

Nathan903:
# Description

Adds support for parsing RFC 2822 / RFC 822 timestamp strings to the CLP-S timestamp parser.

* Added a new `\o{name,offset}` specifier to handle named timezones.
* Updated `\Z` to automatically recognize standard named timezones (like EST, PDT, etc).
* Hardcoded obsolete timezones (UT and GMT) as literals to map to a `+0000` offset.
* Added the RFC 2822 patterns to the default fallback patterns list.

Also added tests and caught & fixed a bug: the timestamp parser throws an `IncompatibleTimestampPattern` error when using `\Z` to parse a string ending in `"UTC"` without any trailing offset (e.g., parsing `"01 Jan 2000 00:00:00 UTC"` against pattern `\d \B{Jan,Feb,Mar,Apr,May,Jun,Jul,Aug,Sep,Oct,Nov,Dec} \Y \H:\M:\S \Z`). This is fixed by adding `optional_timezone_offset_in_minutes = 0;` to the `cUtc` handling block.

The archive version is being bumped since any change to timestamp patterns is a format change.

# Checklist

  breaking change.

# Validation performed

Added unit tests to test_TimestampParser.cpp verifying the \o specifier, \Z fallback logic, and full RFC 2822 formatting strings.
```bash
cmake --build build/core --target unitTest && ./build/core/unitTest "[clp-s][timestamp-parser]"
```

## Summary by CodeRabbit

* **New Features**
  * Added support for named time zones with explicit offsets in timestamp patterns and formatting via the `o` specifier (`\o{name,offset}`).
* **Bug Fixes**
  * Improved validation and parsing of timezone segments, including handling for common abbreviations (EST, UT, GMT) and correct generic timezone expansion.
* **Tests**
  * Expanded coverage for invalid timestamp templates and additional CAT-sequence transformations, plus new timestamp parsing cases for named time zones.


components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -77,8 +77,69 @@ constexpr std::array cPlusMinus
...
+        std::string_view{R"(\e \B{Jan,Feb,Mar,Apr,May,Jun,Jul,Aug,Sep,Oct,Nov,Dec} \Y \H:\M \Z)"},
+        std::string_view{
+                R"(\A{Sun,Mon,Tue,Wed,Thu,Fri,Sat}, \d \B{Jan,Feb,Mar,Apr,May,Jun,Jul,Aug,Sep,Oct,Nov,Dec} \y \H:\M \Z)"
+        },
+        std::string_view{R"(\d \B{Jan,Feb,Mar,Apr,May,Jun,Jul,Aug,Sep,Oct,Nov,Dec} \y \H:\M \Z)"},
+        std::string_view{
+                R"(\A{Sun,Mon,Tue,Wed,Thu,Fri,Sat}, \e \B{Jan,Feb,Mar,Apr,May,Jun,Jul,Aug,Sep,Oct,Nov,Dec} \y \H:\M \Z)"
+        },
+        std::string_view{R"(\e \B{Jan,Feb,Mar,Apr,May,Jun,Jul,Aug,Sep,Oct,Nov,Dec} \y \H:\M \Z)"},
+
```
gibber9809: Do you think we need these patterns in the default set? I think it's pretty unlikely that anyone is omitting seconds in their log messages.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -77,8 +77,69 @@ constexpr std::array cPlusMinus
...
+            {"CDT", "-0500", -300},
+            {"CST", "-0600", -360},
+            {"MDT", "-0600", -360},
+            {"MST", "-0700", -420},
+            {"PDT", "-0700", -420},
+            {"PST", "-0800", -480}}};
 
 constexpr std::array cDefaultDateTimePatterns{
+        // RFC 2822 / 822 patterns (16 variations: with or without day of week, with or without
+        // comma, with or without seconds, with or without timezone offset)
```
gibber9809:
```suggestion
```
Don't think we need the comment.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -77,8 +77,69 @@ constexpr std::array cPlusMinus
...
+        std::string_view{
+                R"(\A{Sun,Mon,Tue,Wed,Thu,Fri,Sat}, \d \B{Jan,Feb,Mar,Apr,May,Jun,Jul,Aug,Sep,Oct,Nov,Dec} \Y \H:\M:\s \Z)"
```
gibber9809:
```suggestion
        std::string_view{
                R"(\A{Sun,Mon,Tue,Wed,Thu,Fri,Sat}, \d \B{Jan,Feb,Mar,Apr,May,Jun,Jul,Aug,Sep,Oct,Nov,Dec} \Y \H:\M:\s\Z)"
```
Same goes for the other newly added formats. I don't think we need the space before `\Z`, since `\Z` already tries to consume a space if there is one, right?

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -77,8 +77,69 @@ constexpr std::array cPlusMinus
...
+            {"CST", "-0600", -360},
+            {"MDT", "-0600", -360},
+            {"MST", "-0700", -420},
+            {"PDT", "-0700", -420},
+            {"PST", "-0800", -480}}};
 
 constexpr std::array cDefaultDateTimePatterns{
+        // RFC 2822 / 822 patterns (16 variations: with or without day of week, with or without
+        // comma, with or without seconds, with or without timezone offset)
+        std::string_view{
```
gibber9809, line 133: Could we move all of these new patterns to the end of the array?

components/core/src/clp_s/timestamp_parser/test/test_TimestampParser.cpp
```diff
@@ -177,6 +177,31 @@ TEST_CASE("timestamp_parser_parse_timestamp", "[clp-s][timestamp-parser]") {
...
         }
+
+        std::vector<std::string> const invalid_timestamp_pattern_templates{
+                R"(\o{,-0500})",  // Empty timezone name
+                R"(\z{-0500}\o{EST,-0500})",  // Duplicate timezone
+                R"(\z{-0500}\z{-0400})",  // Duplicate timezone
+                R"(\o{EST,-0500}\o{UT,+0000})",  // Duplicate timezone
+                R"(\o{EST})",  // Missing comma
+                R"(\o{})",  // Empty brackets
+                R"(\o{EST,-0500,abc})",  // Extra text after offset
```
gibber9809:
```suggestion
                R"(\o{,-0500})",
                R"(\z{-0500}\o{EST,-0500})",
                R"(\z{-0500}\z{-0400})",
                R"(\o{EST,-0500}\o{UT,+0000})",
                R"(\o{EST})",
                R"(\o{})",
                R"(\o{EST,-0500,abc})",
```
Our style guide generally recommends against inline code comments for c++.

components/core/src/clp_s/timestamp_parser/test/test_TimestampParser.cpp
```diff
@@ -177,6 +177,31 @@ TEST_CASE("timestamp_parser_parse_timestamp", "[clp-s][timestamp-parser]") {
...
+        };
+        for (auto const& invalid_timestamp_pattern_template : invalid_timestamp_pattern_templates) {
+            auto const result{TimestampPattern::create(invalid_timestamp_pattern_template)};
+            REQUIRE(result.has_error());
+            REQUIRE(ErrorCode{ErrorCodeEnum::InvalidTimestampPattern} == result.error());
+        }
+
+        std::vector<std::string> const invalid_timezone_offset_templates{
+                R"(\o{EST,-abc})",  // Invalid offset format
+                R"(\o{EST,})",  // Missing offset
```
gibber9809:
```suggestion
                R"(\o{EST,-abc})",
                R"(\o{EST,})",
```
Same as other comment.

components/core/src/clp_s/timestamp_parser/test/test_TimestampParser.cpp
```diff
@@ -609,6 +652,103 @@ TEST_CASE("timestamp_parser_parse_timestamp", "[clp-s][timestamp-parser]") {
                 {"Jan 21 11:56:42 UTC+0130",
                  R"(\B{Jan,Feb,Mar,Apr,May,Jun,Jul,Aug,Sep,Oct,Nov,Dec} \d \H:\M:\S UTC\z{+0130})",
                  1'765'602'000'000'000},
+
+                // RFC 2822 / 822 Variations
+                {"Sat, 09 Mar 2024 15:04:05 -0500",
+                 R"(\A{Sun,Mon,Tue,Wed,Thu,Fri,Sat}, \d \B{Jan,Feb,Mar,Apr,May,Jun,Jul,Aug,Sep,Oct,Nov,Dec} \Y \H:\M:\S \z{-0500})",
+                 1'710'014'645'000'000'000},
+                {"Sat, 09 Mar 2024 15:04 -0500",
```
gibber9809, line 662: Related to my comment in the implementation, I think we can probably get rid of tests for these timestamps without seconds components as well.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -1118,6 +1196,38 @@ auto TimestampPattern::create(std::string_view pattern)
...
                 break;
             }
+            case 'o': {  // Named time-zone with specific offset.
+                if (optional_timezone_size_and_offset.has_value()) {
+                    return ErrorCode{ErrorCodeEnum::InvalidTimestampPattern};
+                }
+                auto const bracket_pattern{YSTDLIB_ERROR_HANDLING_TRYX(
+                        extract_bracket_pattern(pattern.substr(pattern_idx + 1ULL))
+                )};
+                auto const comma_idx{bracket_pattern.find(',')};
```
gibber9809: Instead of doing this and having a bunch of extra validation + logic for making sure these elements are non-empty etc. let's just use `auto const bracket_pattern_list = YSTDLIB_ERROR_HANDLING_TRYX(extract_bracket_pattern_list(bracket_pattern));`, since that function already validates that elements are non-empty etc. + it does some extra validation we aren't doing here about escape sequences.

Also would need to add a check that the returned list has length 2, and return `ErrorCodeEnum::InvalidTimestampPattern` otherwise.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -1691,6 +1801,32 @@ auto parse_timestamp(
...
+                }
+
+                optional_timezone_offset_in_minutes = extracted_timezone_offset;
+                timestamp_idx += name_size;
+
+                auto const closing_brace_idx{raw_pattern.find('}', pattern_idx)};
+                if (std::string_view::npos == closing_brace_idx) {
+                    return ErrorCode{ErrorCodeEnum::InvalidTimestampPattern};
+                }
+                pattern_idx = closing_brace_idx;
```
gibber9809: I don't like this way of advancing the pattern index, since it's fragile to changes in escaping rules, and it's effectively re-validating the timestamp pattern on every parse.

Could we change the optional data we store about timezone to an `std::optional<std::tuple<size_t, size_t, int>>`, with meaning "timezone size", "timezone pattern size", "offset", where "timezone size" is the length of the {...} sequence including the brackets? That way both this code and the `'z'` code can just add that "timezone pattern size" to `pattern_idx` without having to do anything special.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -1707,6 +1843,42 @@ auto parse_timestamp(
...
+                    timezone_pattern.append(cUt);
+                    timestamp_idx += cUt.size();
+                    optional_timezone_offset_in_minutes = 0;
+                    cat_sequence_replacements
+                            .emplace_back(pattern_idx - 1, 2ULL, std::move(timezone_pattern));
+                    break;
+                } else if (remaining_unparsed_content.starts_with(cGmt)) {
+                    timezone_pattern.append(cGmt);
+                    timestamp_idx += cGmt.size();
+                    optional_timezone_offset_in_minutes = 0;
```
gibber9809: I can see why you added `optional_timezone_offset_in_minutes = 0;` to the UTC path, but doesn't seem necessary for the UT or GMT paths?

Also, I guess we're not allowing timezones like UT-YYYY or GMT-ZZZZ for now? If so, fine by me.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
```diff
@@ -786,6 +847,20 @@ auto marshal_date_time_timestamp(
...
+                auto const timezone_offset{pattern.get_optional_timezone_size_and_offset()};
+                if (false == timezone_offset.has_value()) {
+                    return ErrorCode{ErrorCodeEnum::InvalidTimestampPattern};
+                }
+                auto const name_size{timezone_offset.value().first};
+                buffer.append(raw_pattern.substr(pattern_idx + 2ULL, name_size));
+                auto const closing_brace_idx{raw_pattern.find('}', pattern_idx)};
+                if (std::string_view::npos == closing_brace_idx) {
+                    return ErrorCode{ErrorCodeEnum::InvalidTimestampPattern};
+                }
```
gibber9809: Same as other comment -- should use the same fix as in the other instance to advance pattern_idx without re-validating/re-parsing the timestamp pattern.

components/core/src/clp_s/timestamp_parser/test/test_TimestampParser.cpp
```diff
@@ -662,6 +802,9 @@ TEST_CASE("timestamp_parser_parse_timestamp", "[clp-s][timestamp-parser]") {
                     quoted_content,
                     generated_pattern
             )};
+            if (!searched_result.has_value()) {
```
gibber9809:
```suggestion
            if (false == searched_result.has_value()) {
```
Nit: coding guidelines.

gibber9809 (changes requested):
Left an initial review for the whole PR.

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
Nathan903: The RFC standard allows timestamp to include or omit seconds. I previously just added these to cover the RFC standard. I've removed these (as well as tests for timestamps without seconds). I guess we dont need to cover the standard but the goal is just to cover the different formats that would appear in log messages right

Nathan903: However, I just found that the repo already had timestamp formats with no seconds in 
`components/core/src/clp_s/timestamp_parser/TimestampParser.cpp` 
```cpp
        std::string_view{
                R"(\B{January,February,March,April,May,June,July,August,September,October,)"
                R"(November,December} \d, \Y \H:\M)"
        },
```

and `components/core/src/clp_s/timestamp_parser/test/test_TimestampParser.cpp`
```cpp
                {"February 01, 2015 01:02",
                 R"(\B{January,February,March,April,May,June,July,August,September,October,November,December} \d, \Y \H:\M)",
                 1'422'752'520'000'000'000},

```

do you think i should keep the patterns without seconds, or remove them?

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
Nathan903: thanks for the catch, yes those lines are useless. 
Yes, we are not allowing timezones like UT-YYYY or GMT-ZZZZ, since 
> In the obsolete time zone, "UT" and "GMT" are indications of "Universal Time" and "Greenwich Mean Time" respectively and are both semantically identical to "+0000". 
[https://datatracker.ietf.org/doc/html/rfc2822](url)

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
gibber9809: Sure let's keep them and just move them to the end of the array then.

components/core/src/clp_s/timestamp_parser/test/test_TimestampParser.cpp
Nathan903: kept per https://github.com/y-scope/clp/pull/2363#discussion_r3531038038

components/core/src/clp_s/timestamp_parser/TimestampParser.cpp
Nathan903: done. Instead of `std::optional<std::tuple<size_t, size_t, int>>` I added a `TimezoneInfo` struct (per claude's suggestion).

components/core/src/clp_s/timestamp_parser/TimestampParser.hpp
```diff
@@ -14,6 +14,14 @@
 #include "../Defs.hpp"
 
 namespace clp_s::timestamp_parser {
+struct TimezoneInfo {
+    // Length of the timezone content consumed/emitted in the timestamp.
+    size_t timestamp_length;
+    // Length of the timezone specifier's `{...}` sequence in the pattern.
+    size_t pattern_length;
+    int offset;
```
gibber9809:
```suggestion
    size_t timestamp_length{};
    // Length of the timezone specifier's `{...}` sequence in the pattern.
    size_t pattern_length{};
    int offset{};
```
Nit: default initialization with {}.

gibber9809 (changes requested):
Besides that one nit, I see some clang-tidy errors when I run your changes through clang-tidy -- mostly seem to be related to the new `TimezoneInfo` struct, which we could probably fix by adding a constructor. Could you go through and fix all of the clang-tidy errors in all of the files in this PR?

After that, we should be good to merge.

Nathan903:
> Besides that one nit, I see some clang-tidy errors when I run your changes through clang-tidy -- mostly seem to be related to the new `TimezoneInfo` struct, which we could probably fix by adding a constructor. Could you go through and fix all of the clang-tidy errors in all of the files in this PR?

thanks, somehow `task lint:check-cpp-full` didnt caught them (i think because it only lints then changed lines). I've ran clang-tidy and fixed everything. @gibber9809

gibber9809 (approved):
LGTM. For PR title how about:

```
feat(clp-s::timestamp_parser): Add support for parsing RFC 2822 / 822 timestamp strings; Bump archive version to 0.5.1 (fixes #2027).
```

And maybe also edit the PR description to mention that the archive version is being bumped since any change to timestamp patterns is a format change.

PR #2364, opened 2026-07-03
perf(clp-rust-utils): Switch Rust Services' logging to lossless non-blocking writer (fixes #2362).

### LINKED ISSUE #2362: Switch Rust services' logging to lossless non-blocking writer
## Request

Rust services in `clp-rust-utils` should use a non-blocking logger when writing to stdout.

The current logging setup writes directly to stdout when `CLP_LOGS_DIR` is not set:

https://github.com/y-scope/clp/blob/a16d9f5acf8db6a9759c489d24f94ec97619faff/components/clp-rust-utils/src/logging.rs#L36

This makes log writes blocking on the calling thread, which puts logging I/O on the critical path. That is undesirable for async Rust services, especially because `tracing` was originally chosen for its async-friendly non-blocking logging support.

PR #2023 appears to have inadvertently switched the stdout writer to blocking without documenting the tradeoff or design decision. From team discussion, the consensus is that this was likely an oversight rather than an intentional change.

The logging behavior should be made consistent and explicit: either always blocking, always non-blocking, or controlled by a documented switch. It should not silently depend on whether `CLP_LOGS_DIR` is set.

## Possible implementation

Switch the stdout logger to use a lossless non-blocking writer via `WorkerGuard`.

This would ensure that:

1. Log writes are offloaded to a background worker thread instead of blocking the calling thread, as long as the channel is not full.
2. Log records are preserved by using lossless mode.
3. The logging behavior is consistent across output targets, or at least explicitly configured and documented.

This should likely be a small change, possibly only 1–2 lines.

The implementation should also document the logging behavior and tradeoffs in code or developer docs so that the decision is discoverable by future developers.

Discussion context:

* Blocking logging slows down application code on the critical path.
* Non-blocking logging moves I/O off the critical path.
* The channel is unlikely to become full given the current log volume.
* Memory and buffering cost are not expected to be concerns.
* stdout already has system-level buffering outside of the logger configuration.
* We want lossless mode so log records are not dropped.
* Logging behavior should not implicitly change based only on whether `CLP_LOGS_DIR` is set.

Nathan903:
# Description

Switches the Rust logger to use lossless, non-blocking writers for both stdout and file outputs. This moves I/O to a background thread. Also cleans up the API by returning a `LoggerGuards` struct instead of an `Option`, and updates the docs.

# Checklist

  breaking change.

# Validation performed

```
task tests:rust-all
task clean-core
task tests:integration
```

Performance test with a test file shows the new method is indeed almost 2x faster: 

```rust
use std::env;
use std::time::Instant;
use tracing_appender::non_blocking::NonBlockingBuilder;

fn main() {
    let args: Vec<String> = env::args().collect();
    let mode = args.get(1).map(|s| s.as_str()).unwrap_or("non-blocking");
    let num_logs = 100_000;

    let mut _guards = vec![];
    
    let subscriber = tracing_subscriber::fmt()
        .with_ansi(false)
        .with_target(false)
        .with_level(true);

    if mode == "blocking" {
        subscriber.with_writer(std::io::stdout).init();
    } else {
        let (non_blocking_stdout, stdout_guard) = NonBlockingBuilder::default()
            .lossy(false)
            .finish(std::io::stdout());
        _guards.push(stdout_guard);
        subscriber.with_writer(non_blocking_stdout).init();
    }

    let start = Instant::now();
    for i in 0..num_logs {
        tracing::info!("This is test log message number {} to test the logging overhead.", i);
    }
    let duration = start.elapsed();
    
    eprintln!("Mode: {}, Num logs: {}, Time on main thread: {:?}", mode, num_logs, duration);
}
```

```console 
nathan@c4a-standard-8vcpu-32gb-ubuntu2204:~/clp/scratch_logging$ cargo run --release -- blocking > out_blocking.log 2> result_blocking.txt
nathan@c4a-standard-8vcpu-32gb-ubuntu2204:~/clp/scratch_logging$ cat result_blocking.txt
    Finished `release` profile [optimized] target(s) in 0.01s
     Running `/home/nathan/clp/build/rust-targets/release/scratch_logging blocking`
Mode: blocking, Num logs: 100000, Time on main thread: 136.758603ms
nathan@c4a-standard-8vcpu-32gb-ubuntu2204:~/clp/scratch_logging$ cargo run --release -- non-blocking > out_non_blocking.log 2> result_non_blocking.txt
nathan@c4a-standard-8vcpu-32gb-ubuntu2204:~/clp/scratch_logging$ cat result_non_blocking.txt
    Finished `release` profile [optimized] target(s) in 0.01s
     Running `/home/nathan/clp/build/rust-targets/release/scratch_logging non-blocking`
Mode: non-blocking, Num logs: 100000, Time on main thread: 71.02241ms
nathan@c4a-standard-8vcpu-32gb-ubuntu2204:~/clp/scratch_logging$ ls -l |grep .log
-rw-rw-r-- 1 nathan nathan 10188890 Jul  3 22:49 out_blocking.log
-rw-rw-r-- 1 nathan nathan 10188890 Jul  3 22:49 out_non_blocking.log
```

## Summary by CodeRabbit

* **New Features**
  * Improved logging setup so both console and optional file logging are kept active for the full process lifetime.
  * When log files are enabled, logs are rotated hourly with background handling to reduce dropped messages.

* **Documentation**
  * Clarified developer and operator guidance for Rust logging and log file behaviour.
  * Updated instructions to emphasise keeping the logging guard alive.

### DIFF
--- a/components/clp-rust-utils/src/logging.rs
+++ b/components/clp-rust-utils/src/logging.rs
@@ -1,5 +1,5 @@
 use tracing_appender::{
-    non_blocking::WorkerGuard,
+    non_blocking::{NonBlockingBuilder, WorkerGuard},
     rolling::{RollingFileAppender, Rotation},
 };
 use tracing_subscriber::{self, fmt::writer::MakeWriterExt};
@@ -7,9 +7,9 @@ use tracing_subscriber::{self, fmt::writer::MakeWriterExt};
 /// Initializes the global tracing subscriber with JSON-formatted output to stdout.
 ///
 /// If the `CLP_LOGS_DIR` environment variable is set, logs are also written to a
-/// rolling file (`{log_filename}`) in that directory. The returned [`WorkerGuard`]
-/// must be held for the lifetime of the program to ensure file logs are flushed.
-pub fn set_up_logging(log_filename: &str) -> Option<WorkerGuard> {
+/// rolling file (`{log_filename}`) in that directory. The returned [`WorkerGuard`]s
+/// must be held for the lifetime of the program to ensure logs are flushed.
+pub fn set_up_logging(log_filename: &str) -> Vec<WorkerGuard> {
     let subscriber = tracing_subscriber::fmt()
         .event_format(
             tracing_subscriber::fmt::format()
@@ -22,18 +22,28 @@ pub fn set_up_logging(log_filename: &str) -> Option<WorkerGuard> {
         .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
         .with_ansi(false);
 
+    let (stdout_writer, stdout_guard) = NonBlockingBuilder::default()
+        .lossy(false)
+        .finish(std::io::stdout());
+
+    let mut guards = vec![stdout_guard];
+
     if let Ok(logs_directory) = std::env::var("CLP_LOGS_DIR") {
         let logs_directory = std::path::Path::new(logs_directory.as_str());
         let file_appender =
             RollingFileAppender::new(Rotation::HOURLY, logs_directory, log_filename);
-        let (non_blocking_writer, guard) = tracing_appender::non_blocking(file_appender);
+        let (file_writer, file_guard) = NonBlockingBuilder::default()
+            .lossy(false)
+            .finish(file_appender);
+        
+        guards.push(file_guard);
 
         subscriber
-            .with_writer(std::io::stdout.and(non_blocking_writer))
+            .with_writer(stdout_writer.and(file_writer))
             .init();
-        Some(guard)
     } else {
-        subscriber.with_writer(std::io::stdout).init();
-        None
+        subscriber.with_writer(stdout_writer).init();
     }
+
+    guards
 }
--- a/components/clp-rust-utils/src/logging.rs
+++ b/components/clp-rust-utils/src/logging.rs
@@ -4,12 +4,20 @@ use tracing_appender::{
 };
 use tracing_subscriber::{self, fmt::writer::MakeWriterExt};
 
+/// Opaque struct to hold the worker guards for the non-blocking loggers.
+/// These guards must be held for the lifetime of the program to ensure logs are flushed.
+pub struct LoggerGuards {
+    _stdout_guard: WorkerGuard,
+    _file_guard: Option<WorkerGuard>,
+}
+
 /// Initializes the global tracing subscriber with JSON-formatted output to stdout.
 ///
 /// If the `CLP_LOGS_DIR` environment variable is set, logs are also written to a
-/// rolling file (`{log_filename}`) in that directory. The returned [`WorkerGuard`]s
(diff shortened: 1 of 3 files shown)


components/clp-rust-utils/src/logging.rs
```diff
@@ -1,15 +1,23 @@
 use tracing_appender::{
-    non_blocking::WorkerGuard,
+    non_blocking::{NonBlockingBuilder, WorkerGuard},
     rolling::{RollingFileAppender, Rotation},
 };
 use tracing_subscriber::{self, fmt::writer::MakeWriterExt};
 
+/// Opaque struct to hold the worker guards for the non-blocking loggers.
```
junhaoliao: with lossy=false, the crate applies backpressure if the bounded queue fills, so it's not truly non-blocking, would this be a more accurate wording?

```suggestion
/// Opaque struct to hold the worker guards for the background log writers.
```

docs/src/dev-docs/logging-operator-guide.md
```diff
@@ -45,8 +45,8 @@ For continuously running deployed services such as Python orchestration services
 * **Docker Compose**: Logs are available using `docker compose logs`. If `CLP_LOGS_DIR` is set,
 logs are additionally written to `<CLP_LOGS_DIR>/<component_name>.log`. This directory mounts to the host via `CLP_LOGS_DIR_HOST` (defaults to `./var/log`).
   :::{note}
-  For Rust services, setting `CLP_LOGS_DIR` enables file logging with hourly log rotation and a
-  non-blocking file writer.
+  For Rust services, `stdout` logging is always non-blocking. Setting `CLP_LOGS_DIR` additionally
+  enables file logging with hourly log rotation and a non-blocking file writer.
```
junhaoliao:
```suggestion
  For Rust services, `stdout` logging is sent through a background writer configured to avoid
  dropped log records. Setting `CLP_LOGS_DIR` additionally enables file logging with hourly log
  rotation and the same lossless background writer behavior.
```

junhaoliao (changes requested):

LinZhihao-723 (approved):
lgtm.

components/clp-rust-utils/src/logging.rs
```diff
@@ -22,18 +30,32 @@ pub fn set_up_logging(log_filename: &str) -> Option<WorkerGuard> {
...
+        }
     } else {
-        subscriber.with_writer(std::io::stdout).init();
-        None
+        subscriber.with_writer(stdout_writer).init();
+
+        LoggerGuards {
+            _stdout_guard: stdout_guard,
+            _file_guard: None,
+        }
```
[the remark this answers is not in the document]
Nathan903: _Dearest coderabbit,_ please read the issue carefully:  https://github.com/y-scope/clp/issues/2362

> We want lossless mode so log records are not dropped.
> The channel is unlikely to become full given the current log volume.

junhaoliao (approved):

PR #2369, opened 2026-07-06
perf(log-ingestor): Reduce compression-submission lock contention with `READ COMMITTED` and smaller update batches (fixes #2358).

### LINKED ISSUE #2358: log-ingestor: large single-transaction status UPDATEs on ingested_s3_object_metadata cause lock contention / Lock wait timeout (1205) under concurrent listeners
### Summary

When many SQS listeners run concurrently and catch up on a large backlog, the compression-submission path issues a single large `UPDATE` (the entire job's worth of rows) inside one transaction. Under concurrency this produces a pile-up of connections in MySQL state `waiting for handler commit`, individual UPDATEs taking 30–50s, and eventually `ERROR 1205 (HY000): Lock wait timeout exceeded`. The compression scheduler is starved of DB throughput while this happens. The larger the backlog, the larger each per-job transaction, the worse the contention.

### Where in the code

`components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs`, in the compression-submission path:

- Transaction begins at **L1100** (`self.db_pool.begin()`).
- The status update loops over `object_metadata_ids.chunks(10000)` at **L1117**, building `UPDATE ingested_s3_object_metadata SET compression_job_id = ?, status = 'Submitted' WHERE id IN (...) AND status = 'Buffered'`.
- A **single** `tx.commit()` runs *after* the whole loop at **L1148** — so every chunk's write-set is held until one final commit, regardless of how many chunks (i.e. how large the job) it took.
- The 10000 chunk size is explicitly chosen only to stay under MySQL's 65,535 bound-parameter limit (comment at L1110–1116) — it is **not** a lock-footprint or commit-frequency control.
- There is **no index on `status`** (`ingested_s3_object_metadata` has PK `id` and indexes on `ingestion_job_id` / `compression_job_id`), so the `AND status = 'Buffered'` predicate is not index-assisted.
- The completion path (**L1180**, `UPDATE ... SET status = ? WHERE compression_job_id = ?`) is index-assisted and not the hot spot.

### Why it collapses under concurrency

Each listener holds a large set of row locks for the full duration of its (potentially multi-chunk) transaction, releasing them only at the single commit. With dozens of listeners committing large transactions simultaneously, commit-stage serialization (`waiting for handler commit`) and lock waits compound. Because backlog size drives job size drives transaction size, a bigger backlog makes the problem strictly worse — exactly when you most need throughput.

### Repro sketch

1. Accumulate a large `buffered` backlog in `ingested_s3_object_metadata` (e.g. pause consumers, let a queue build).
2. Start many listeners concurrently against the same DB so they all enter the submission path at once.
3. Observe `SHOW PROCESSLIST`: many connections in `waiting for handler commit`, submission UPDATEs at tens of seconds, and `1205` lock-wait-timeout errors; scheduler dispatch stalls.

### Proposed fixes (any subset)

1. **Commit per chunk** — move `tx.commit()` inside the `chunks()` loop (or use a fresh transaction per chunk) so lock hold time is bounded by one chunk, not one job.
2. **Smaller, configurable batch size** (e.g. 500–2000) decoupled from the placeholder-limit ceiling, to bound lock footprint and commit size.
3. **`READ COMMITTED` isolation** for this path to reduce gap/next-key locking under the `IN (...) AND status = ...` pattern.
4. **Backpressure / concurrency cap** on the number of listeners simultaneously in the submission/commit path.

### Workaround

Reduce per-listener buffer size and run fewer listeners concurrently during multi-hour catch-up (smaller per-job transactions → shorter lock holds).

### Environment

CLP `log-ingestor` (Rust, sqlx + MySQL/InnoDB, `REPEATABLE READ`), dozens of concurrent SQS listeners during a multi-hour backlog catch-up.

LinZhihao-723:
# Description

When an ingestion job flushes its buffered objects, `ClpCompressionState::submit_for_compression` runs a single transaction that inserts a `compression_jobs` row and then stamps `compression_job_id` and `status = 'submitted'` onto every buffered `ingested_s3_object_metadata` row via a batched `UPDATE ... WHERE id IN (...) AND status = 'buffered'`.

Under many concurrent ingestion jobs these submission transactions serialize heavily: the InnoDB default isolation level (REPEATABLE READ) takes gap / next-key locks over the index ranges scanned for the `IN (...)` set and the `status` predicate, so concurrent submissions block one another. This makes submission the dominant DB cost and the ingestion bottleneck (#2358).

This PR relieves the contention on the submission path with two changes:

- Run the submission transaction at **READ COMMITTED** isolation. READ COMMITTED does not take gap locks, which is what drives the cross-transaction serialization here; only the matching rows are locked, for a shorter duration.
- Reduce the metadata-update batch size from 10000 to 1000, keeping each `UPDATE`'s lock footprint small under concurrency.

## Microbenchmark result

We built a microbenchmark that drives the real DB persistence path with **50 concurrent ingestion jobs**, each ingesting object-metadata entries at a throttled rate (~15,000 entries/min/job), and measured the average DB time per `submit_for_compression` call.

Switching the submission transaction to `READ COMMITTED` substantially reduced that cost:

- **REPEATABLE READ (before):** ~30 s per submission call at comparable scale, and climbing as the metadata table grew.
- **READ COMMITTED (after):** ~12.5 s per submission call, roughly flat as the table grew — about a **2.4x** reduction — with end-to-end ingestion throughput improving from ~5,800 to ~8,400 rows/s.

# Checklist

  breaking change.

# Validation performed

* [x] Ensure log-ingestor can submit compression jobs as expected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compression job submission reliability by running related updates inside a dedicated `READ COMMITTED` transaction context.
  * Reduced ingested-object metadata update batch size (from larger batches to smaller batches of 1,000) and added validation after each batch to detect unexpected row counts.
  * Updated transaction handling so isolation-level settings won’t affect subsequent database operations from the same connection.

### DIFF
--- a/components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
+++ b/components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
@@ -22,7 +22,7 @@ use clp_rust_utils::{
 };
 use const_format::formatcp;
 use non_empty_string::NonEmptyString;
-use sqlx::{Acquire as _, MySqlPool};
+use sqlx::{Connection, MySqlPool};
 use strum_macros::{AsRefStr, Display, EnumIter, EnumString};
 use tokio::sync::mpsc;
 
@@ -1071,9 +1071,9 @@ impl ClpCompressionState {
     /// * [`anyhow::Error`] if one or more object metadata rows fail to be updated in the DB.
     /// * Forwards [`clp_rust_utils::serde::BrotliMsgpack::serialize`]'s return values on failure.
     /// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
-    /// * Forwards [`sqlx::Pool::acquire`]'s return values on failure.
-    /// * Forwards [`sqlx::Acquire::begin`]'s return values on failure.
+    /// * Forwards [`sqlx::Connection::begin`]'s return values on failure.
     /// * Forwards [`sqlx::Transaction::commit`]'s return values on failure.
+    /// * Forwards [`run_read_committed_tx`]'s return values on failure.
     ///
     /// # Panics
     ///
@@ -1082,7 +1082,6 @@ impl ClpCompressionState {
         &self,
         io_config: ClpIoConfig,
     ) -> anyhow::Result<CompressionJobId> {
-        const SET_READ_COMMITTED: &str = "SET TRANSACTION ISOLATION LEVEL READ COMMITTED";
         const COMPRESSION_JOB_SUBMISSION_QUERY: &str = formatcp!(
             r"INSERT INTO {table} (`clp_config`) VALUES (?)",
             table = CLP_COMPRESSION_JOB_TABLE_NAME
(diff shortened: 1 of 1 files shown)


components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -1093,56 +1094,61 @@ impl ClpCompressionState {
             }
         };
 
-        let mut tx = self.db_pool.begin().await?;
+        // Run the submission at `READ COMMITTED` to avoid the gap/next-key locks that serialize
+        // concurrent submissions under the default `REPEATABLE READ`.
```
sitaowang1998: This is confusing. I think the `READ COMMITED` is on transaction level. This comment read like it is on query level.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
```diff
@@ -1509,6 +1515,53 @@ async fn update_job_status(
...
+/// The value returned by `tx` on success.
+///
+/// # Errors
+///
+/// Returns an error if:
+///
+/// * Forwards [`sqlx::Pool::acquire`]'s return values on failure.
+/// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
+/// * Forwards `tx`'s return values on failure.
+async fn run_read_committed_tx<ReturnType, TransactionType>(
```
sitaowang1998, line 1547: I am ok with leaving it here just for this PR, but we probably need to move it to some db helper file later.

LinZhihao-723: Yeah, there's no good place to move it around, but I agree eventually we should have our own namespace for these helpers.

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
LinZhihao-723: Hmmm, sure I can explicitly mention it's on tx level, but just curious why do you think this makes it sound like it's in query level?

sitaowang1998: The word `submission` is ambiguous here. It can mean the whole function, the transaction, or the query `COMPRESSION_JOB_SUBMISSION_QUERY`. The first impression I get is the last one.

sitaowang1998 (approved):

components/log-ingestor/src/ingestion_job_manager/clp_ingestion.rs
sitaowang1998: I think what adds to the confusion is that the body of the transaction is an unnamed lambda function, and the next appearance of the word `submission` is in the name of the query.

LinZhihao-723: Ok, I've added `transaction` to this comment besides `submission`.

jackluo923 (approved):

PR #2371, opened 2026-07-07
fix(deps): Pin scikit-build-core build backend (fixes #2370).

### LINKED ISSUE #2370: clp-lint fails after scikit-build-core 1.0.0 rejects deprecated cmake.verbose
### Bug

`clp-lint` can fail in a fresh environment while linting `python-wheels/yscope-clp-core` because the package used the deprecated `scikit-build-core` setting `cmake.verbose` and left the PEP 517 build backend requirement unpinned.

After `scikit-build-core` 1.0.0 was published, editable builds started failing before `ruff` or `mypy` could run:

```text
ERROR: Use build.verbose instead of cmake.verbose for scikit-build-core >= 0.10
```

Expected behavior: Python lint should be reproducible and should not depend on the latest unpinned `scikit-build-core` build backend behavior.

Observed examples:

- Passing scheduled `clp-lint` on `main`: https://github.com/y-scope/clp/actions/runs/28768186187
- Failing scheduled `clp-lint` on the same `main` commit: https://github.com/y-scope/clp/actions/runs/28841112795
- Failing PR job with the same root cause: https://github.com/y-scope/clp/actions/runs/28843605221/job/85542564743?pr=2369

### CLP version

`732fea6c9d11685c36c78f0cda3e811290b8c81e`

### Environment

GitHub Actions `ubuntu-24.04` runner in the `clp-lint` workflow, using Python 3.10 and `uv run --directory python-wheels/yscope-clp-core ...`.

Locally reproduced on Ubuntu with Python 3.10.12 and `uv 0.11.16`.

### Reproduction steps

1. Check out `732fea6c9d11685c36c78f0cda3e811290b8c81e` or another commit before the fix.
2. Run:

   ```shell
   uv run --directory python-wheels/yscope-clp-core ruff check .
   ```

3. Observe that `uv` creates the editable build environment and fails while building `yscope-clp-core`:

   ```text
   ERROR: Use build.verbose instead of cmake.verbose for scikit-build-core >= 0.10
   ```

4. The failure happens before Ruff checks the package.

junhaoliao:
# Description

Fixes https://github.com/y-scope/clp/issues/2370.

`python-wheels/yscope-clp-core` used the deprecated `scikit-build-core` setting `cmake.verbose` and left the build backend requirement unpinned. After `scikit-build-core` 1.0.0 was published, fresh `uv run --directory python-wheels/yscope-clp-core ...` invocations started resolving the new backend in the isolated editable-build environment and failing before Ruff or mypy could run.

This PR updates the setting to `build.verbose` and pins `scikit-build-core==1.0.0` in `[build-system].requires`, which is the PEP 517 control point for the isolated build backend environment. It intentionally does not add `scikit-build-core` to the package's `build` dependency group because that would only affect uv's project environment and lockfile, not the isolated build backend resolution that caused this failure.

# Checklist

  breaking change.

# Validation performed

## Scenario 1: Lockfile is current and package lint/type checks pass

**Task:** Verify the `yscope-clp-core` lockfile is current and the package no longer fails during the editable build setup used by Python lint.

**Commands and output:**

```text
$ uv lock --directory python-wheels/yscope-clp-core --check
Using CPython 3.10.12 interpreter at: /usr/bin/python3.10
Resolved 17 packages in 0.55ms

$ UV_NO_CACHE=1 uv run --directory python-wheels/yscope-clp-core ruff check --fix .
Using CPython 3.10.12 interpreter at: /usr/bin/python3.10
Creating virtual environment at: .venv
   Building yscope-clp-core @ file:///home/junhao/workspace/5-clp/python-wheels/yscope-clp-core
Downloading mypy (12.9MiB)
Downloading ruff (10.7MiB)
 Downloaded ruff
 Downloaded mypy
      Built yscope-clp-core @ file:///home/junhao/workspace/5-clp/python-wheels/yscope-clp-core
Installed 14 packages in 7ms
All checks passed!

$ uv run --directory python-wheels/yscope-clp-core ruff format .
2 files left unchanged

$ uv run --directory python-wheels/yscope-clp-core ruff check .
All checks passed!

$ uv run --directory python-wheels/yscope-clp-core ruff format --diff .
2 files already formatted

$ uv run --directory python-wheels/yscope-clp-core mypy .
Success: no issues found in 2 source files

```

**Explanation:** `uv lock --check` confirms the lockfile matches the manifest. The package-level Ruff and mypy commands pass after `uv` builds `yscope-clp-core` with the pinned `scikit-build-core` backend from `[build-system].requires`.

## Summary by CodeRabbit

* **Chores**
  * Updated the package build setup to use a fixed tool version for more consistent wheel builds.
  * Adjusted verbose build output handling so install/build logs are more reliable and easier to inspect.

### DIFF
--- a/python-wheels/yscope-clp-core/pyproject.toml
+++ b/python-wheels/yscope-clp-core/pyproject.toml
@@ -33,7 +33,6 @@ build = [
   "auditwheel>=6.6.0",
   "build>=1.4.0",
   "patchelf>=0.17.2.4",
-  "scikit-build-core==1.0.0",
 ]
 dev = [
   "mypy>=1.16.0",
--- a/python-wheels/yscope-clp-core/pyproject.toml
+++ b/python-wheels/yscope-clp-core/pyproject.toml
@@ -17,7 +17,7 @@ requires-python = ">=3.10"
 "Issue Tracker" = "https://github.com/y-scope/clp/issues"
 
 [build-system]
-requires = ["scikit-build-core"]
+requires = ["scikit-build-core==1.0.0"]
 build-backend = "scikit_build_core.build"
 
 [tool.scikit-build]
@@ -25,7 +25,7 @@ wheel.packages = ["yscope_clp_core"]
 wheel.py-api = "py3"
 cmake.version = ">=3.31.0"
 ninja.version = ">=1.10.0"
-cmake.verbose = true
+build.verbose = true
 logging.level = "DEBUG"
 
 [dependency-groups]
@@ -33,6 +33,7 @@ build = [
   "auditwheel>=6.6.0",
   "build>=1.4.0",
   "patchelf>=0.17.2.4",
+  "scikit-build-core==1.0.0",
 ]
 dev = [
   "mypy>=1.16.0",
--- a/python-wheels/yscope-clp-core/uv.lock
+++ b/python-wheels/yscope-clp-core/uv.lock
@@ -41,18 +41,6 @@ wheels = [
 ]
 
 [[package]]
-name = "exceptiongroup"
-version = "1.3.1"
-source = { registry = "https://pypi.org/simple" }
-dependencies = [
-    { name = "typing-extensions", marker = "python_full_version < '3.13'" },
-]
-sdist = { url = "https://files.pythonhosted.org/packages/50/79/66800aadf48771f6b62f7eb014e352e5d06856655206165d775e675a02c9/exceptiongroup-1.3.1.tar.gz", hash = "sha256:8b412432c6055b0b7d14c310000ae93352ed6754f70fa8f7c34141f91c4e3219", size = 30371, upload-time = "2025-11-21T23:01:54.787Z" }
-wheels = [
-    { url = "https://files.pythonhosted.org/packages/8a/0e/97c33bf5009bdbac74fd2beace167cab3f978feb69cc36f1ef79360d6c4e/exceptiongroup-1.3.1-py3-none-any.whl", hash = "sha256:a7a39a3bd276781e98394987d3a5701d0c4edffb633bb7a5144577f82c773598", size = 16740, upload-time = "2025-11-21T23:01:53.443Z" },
-]
-
-[[package]]
 name = "importlib-metadata"
 version = "8.7.1"
 source = { registry = "https://pypi.org/simple" }
@@ -271,22 +259,6 @@ wheels = [
 ]
 
 [[package]]
-name = "scikit-build-core"
-version = "1.0.0"
-source = { registry = "https://pypi.org/simple" }
-dependencies = [
-    { name = "exceptiongroup", marker = "python_full_version < '3.11'" },
-    { name = "packaging" },
-    { name = "pathspec" },
-    { name = "tomli", marker = "python_full_version < '3.11'" },
-    { name = "typing-extensions", marker = "python_full_version < '3.11'" },
-]
-sdist = { url = "https://files.pythonhosted.org/packages/29/e2/4c0431fe84f9d8c24c1fc77b27264f568c71724ca888f514766986f270b0/scikit_build_core-1.0.0.tar.gz", hash = "sha256:b82a8b41dd66926b96096a61e8fc8df22214bbec437d251c0fda1bfb9d7df558", size = 466325, upload-time = "2026-07-06T16:58:35.859Z" }
-wheels = [
-    { url = "https://files.pythonhosted.org/packages/98/47/471ffd83e97c221bedb0c16ab8bdf6087683686c2bdbecda743b24c01002/scikit_build_core-1.0.0-py3-none-any.whl", hash = "sha256:3d52323489b3d83c87922e7bd5bd781cc4bcf0d43961eb565a78d28e74b98fef", size = 275217, upload-time = "2026-07-06T16:58:33.971Z" },
-]
-
-[[package]]
 name = "tomli"
 version = "2.4.0"
 source = { registry = "https://pypi.org/simple" }
@@ -359,7 +331,6 @@ build = [
     { name = "auditwheel" },
     { name = "build" },
     { name = "patchelf" },
-    { name = "scikit-build-core" },
 ]
 dev = [
     { name = "mypy" },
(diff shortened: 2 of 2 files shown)


jackluo923 (dismissed):

python-wheels/yscope-clp-core/pyproject.toml
```diff
@@ -17,22 +17,23 @@ requires-python = ">=3.10"
...
-cmake.verbose = true
+build.verbose = true
 logging.level = "DEBUG"
 
 [dependency-groups]
 build = [
   "auditwheel>=6.6.0",
   "build>=1.4.0",
   "patchelf>=0.17.2.4",
+  "scikit-build-core==1.0.0",
```
kirkrodrigues: Hm, is there a way to avoid repeating the version in both places?
Also, why pin scikit-build-core but not the dependencies?

junhaoliao: 1. i verified with local uv 0.11.16: if `scikit-build-core==1.0.0` exists only in `[build-system].requires`, `uv lock` does not include `scikit-build-core` in `uv.lock`, so the duplicate is what makes the backend version appear in the lockfile
2. the other deps' versions, together with the `scikit-build-core`'s transitive deps' version, are locked in `uv.lock` already. technically we can write `scikit-build-core>=1.0.0,<2` in pyproject.toml too, assuming their maintainers can bump versions for any breaking changes correctly, but it's still less locked. what do you think?

kirkrodrigues: 1. Hm, then what if we don't lock it in `build.system.requires`?
2. Well I think we need to stick to either locking explicitly or not, so to keep with current practice, I would actually do `scikit-build-core>=1.0.0` and let the lock file take care of the explicit lock.

junhaoliao: i ran an experiment - result proves the dependency-group lock does not constrain the isolated PEP 517 build env

```sh
tmp=$(mktemp -d)

cat >"$tmp/pyproject.toml" <<'EOF'
[project]
name = "pep517-lock-proof"
version = "0.0.0"
requires-python = ">=3.10"

[build-system]
requires = ["packaging>=24"]
build-backend = "probe_backend"
backend-path = ["backend"]

[dependency-groups]
build = [
  "build>=1.4.0",
  "packaging==24.0",
]
EOF

mkdir -p "$tmp/backend"

cat >"$tmp/backend/probe_backend.py" <<'EOF'
from pathlib import Path
from zipfile import ZIP_DEFLATED, ZipFile

from packaging import __version__ as packaging_version

NAME = "pep517_lock_proof"
VERSION = "0.0.0"
DIST_INFO = f"{NAME}-{VERSION}.dist-info"


def get_requires_for_build_wheel(config_settings=None):
    return []


def prepare_metadata_for_build_wheel(metadata_directory, config_settings=None):
    dist = Path(metadata_directory) / DIST_INFO
    dist.mkdir(parents=True, exist_ok=True)
    (dist / "METADATA").write_text(
        "Metadata-Version: 2.1\nName: pep517-lock-proof\nVersion: 0.0.0\n",
        encoding="utf-8",
    )
    (dist / "WHEEL").write_text(
        "Wheel-Version: 1.0\nGenerator: probe\nRoot-Is-Purelib: true\nTag: py3-none-any\n",
        encoding="utf-8",
    )
    return DIST_INFO


def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
    Path("build-backend-packaging-version.txt").write_text(
        packaging_version,
        encoding="utf-8",
    )

    wheel_name = f"{NAME}-{VERSION}-py3-none-any.whl"
    wheel_path = Path(wheel_directory) / wheel_name

    with ZipFile(wheel_path, "w", ZIP_DEFLATED) as zf:
        zf.writestr(
            f"{DIST_INFO}/METADATA",
            "Metadata-Version: 2.1\nName: pep517-lock-proof\nVersion: 0.0.0\n",
        )
        zf.writestr(
            f"{DIST_INFO}/WHEEL",
            "Wheel-Version: 1.0\nGenerator: probe\nRoot-Is-Purelib: true\nTag: py3-none-any\n",
        )
        zf.writestr(f"{DIST_INFO}/RECORD", "")

    return wheel_name
EOF

printf '\n# Lock + sync uv project env, but do not install the local project\n'
uv lock --directory "$tmp"
uv sync --directory "$tmp" --group build --no-install-project

printf '\n# uv-managed project env packaging version\n'
"$tmp/.venv/bin/python" -c 'import packaging; print(packaging.__version__)'

printf '\n# Isolated PEP 517 build using [build-system].requires = ["packaging>=24"]\n'
(cd "$tmp" && "$tmp/.venv/bin/python" -m build --wheel --outdir "$tmp/dist")

printf '\n# packaging version imported inside isolated build backend\n'
cat "$tmp/build-backend-packaging-version.txt"

printf '\n\n# uv.lock pinned the project env package to 24.0\n'
rg -n 'name = "packaging"|version = "24\.0"|specifier = "==24\.0"' "$tmp/uv.lock"

rm -rf "$tmp"
```

Example observed output:

```text
# Lock + sync uv project env, but do not install the local project
Using CPython 3.14.5
Resolved 8 packages in 1ms
Using CPython 3.14.5
Creating virtual environment at: .venv
Resolved 8 packages in 0.45ms
Prepared 2 packages in 105ms
Installed 3 packages in 2ms
 + build==1.5.0
 + packaging==24.0
 + pyproject-hooks==1.2.0

# uv-managed project env packaging version
24.0

# Isolated PEP 517 build using [build-system].requires = ["packaging>=24"]
* Creating isolated environment: venv+pip...
* Installing packages in isolated environment:
  - packaging>=24
* Getting build dependencies for wheel...
* Building wheel...
Successfully built pep517_lock_proof-0.0.0-py3-none-any.whl

# packaging version imported inside isolated build backend
26.2

# uv.lock pinned the project env package to 24.0
43:name = "packaging"
44:version = "24.0"
67:    { name = "packaging", specifier = "==24.0" }
```

so even with a working uv.lock and project env pinned to packaging==24.0, the isolated build backend env installed packaging>=24 separately and got 26.2

i think we should remove the locking in the `build` deps group then. the version locking in `[build-system].requires` is what truly matters for locking the build backend version

junhaoliao: do note that there's no good way to pin the transitive deps versions of what we specify in `[build-system].requires` - we can specify [`[tool.uv].build-constraint-dependencies`](https://docs.astral.sh/uv/reference/settings/#build-constraint-dependencies), but it is specific to uv tooling (probably fine) and can be hard to be kept in sync (not good).

kirkrodrigues (approved):

PR #2377, opened 2026-07-09
fix(build): Prevent concurrent executions of core-build (fixes #2376).

### LINKED ISSUE #2376: bug(build): core-build runs twice concurrently and fails when running integration test
### Bug

`task tests:integration` sporadically fails with a linker error such as `undefined reference to 'main'` 

This is because `tests:integration` invokes multiple targets in parallel (`tests:integration:core` and `tests:integration:package`) that both depend on the `core` task. The `core` task in `taskfile.yaml` uses imperative `cmds` instead of declarative `deps`, so task's built-in dependency deduplication fails to recognize the overlapping executions. Thus, two parallel `cmake --build` processes are spawned in the same directory. 

```text
28066: task: [utils:cmake:build:/home/nathan/clp/build/core-[clg clo clp clp-s indexer log-converter reducer-server]-[]] cmake --build "/home/nathan/clp/build/core" --parallel "8" ...
28109: task: [utils:cmake:build:/home/nathan/clp/build/core-[clg clo clp clp-s indexer log-converter reducer-server]-[]] cmake --build "/home/nathan/clp/build/core" --parallel "8" ...
```

These processes race to compile the same source files, resulting in linker errors.

```text
28800: [ 22%] Building CXX object src/clp_s/indexer/CMakeFiles/indexer.dir/__/__/clp/GlobalMetadataDBConfig.cpp.o
...
28804: [ 22%] Building CXX object src/clp_s/indexer/CMakeFiles/indexer.dir/__/__/clp/GlobalMetadataDBConfig.cpp.o
...
28872: [ 77%] Linking CXX executable ../../../indexer
28873: [ 77%] Linking CXX executable ../../../indexer
28874: /usr/bin/ld: /usr/lib/gcc/aarch64-linux-gnu/11/../../../aarch64-linux-gnu/Scrt1.o: in function `_start':
28875: (.text+0x1c): undefined reference to `main'
28877: collect2: error: ld returned 1 exit status
```

### CLP version

Current HEAD (version 0.12.1-dev)

### Environment

tested on baker21 and GCP Arm linux server

### Reproduction steps

1. run `task clean; task tests:integration` a few times some bad luck. 
2. see that `cmake --build` is launched twice concurrently for the `core` target.
3. Observe the build sporadically failing during the linking phase with a `collect2` or `undefined reference` error.

Nathan903:
# Description
Fixes race condition of `task tests:integration` sporadically fails with linker errors (like `undefined reference to 'main'`). 

The `core` task previously invoked `core-build` using `cmds` instead of `deps`. Because of this, `go-task` could not deduplicate the build when multiple parallel test targets triggered the `core` task at the same time. This caused two parallel `cmake --build` processes to run simultaneously and corrupt each other's object files.

This PR
* Converts the `core` task to use `deps` instead of `cmds` so the task runner can track it natively.
* Adds `core-generate` as a dependency of `core-build` to maintain strict ordering.
* Adds `run: once` to `core-generate` and `core-build` in `taskfile.yaml`, and to the `core` dependency task in `taskfiles/deps/main.yaml`, to guarantee they evaluate only once globally, regardless of differing environment configurations passed down by the parallel test wrappers.

# Checklist

  breaking change.

# Validation performed

*Task now deduplicates the build step.*

Previously: 
```bash
task clean
task --dry tests:integration > /tmp/task-dry-run-check.log 2>&1
grep -c "utils:cmake:build.*/home/nathan/clp/build/core" /tmp/task-dry-run-check.log
```
gives
```
2
```

Now 
```bash
task clean
task --dry tests:integration > /tmp/task-dry-run-check.log 2>&1
grep -c "utils:cmake:build.*/home/nathan/clp/build/core" /tmp/task-dry-run-check.log
```
gives
```
1
```

*End to end test:*
```
task tests:integration > /tmp/clp-build-verify.log 2>&1
grep -c "cmake --build .*/home/nathan/clp/build/core" /tmp/clp-build-verify.log
```
gives `1`  (`cmake --build` only invoked once)

## Summary by CodeRabbit

* **Refactor**
  * Updated task execution to use explicit dependencies, ensuring generation runs before build.
  * Added single-run safeguards so related tasks execute only once per invocation.
* **Chores**
  * Updated task configuration to run the core workflow in a more controlled, deterministic way.

### DIFF
--- a/taskfile.yaml
+++ b/taskfile.yaml
@@ -173,9 +173,8 @@ tasks:
           INCLUDE_PATTERNS: ["{{.OUTPUT_DIR}}"]
 
   core:
-    cmds:
-      - task: "core-generate"
-      - task: "core-build"
+    deps:
+      - "core-build"
 
   core-generate:
     internal: true
@@ -198,6 +197,8 @@ tasks:
   core-build:
     internal: true
     sources: *core_source_files
+    deps:
+      - "core-generate"
     generates:
       - "{{.G_CORE_COMPONENT_BUILD_DIR}}/clg"
       - "{{.G_CORE_COMPONENT_BUILD_DIR}}/clo"
--- a/taskfile.yaml
+++ b/taskfile.yaml
@@ -178,7 +178,7 @@ tasks:
 
   core-generate:
     internal: true
-    run: once
+    run: "once"
     sources: &core_source_files
       - "{{.G_DEPS_CPP_CHECKSUM_FILE}}"
       - "{{.G_CORE_COMPONENT_DIR}}/cmake/**/*"
@@ -197,7 +197,7 @@ tasks:
 
   core-build:
     internal: true
-    run: once
+    run: "once"
     sources: *core_source_files
     deps:
       - "core-generate"
--- a/taskfile.yaml
+++ b/taskfile.yaml
@@ -178,6 +178,7 @@ tasks:
 
   core-generate:
     internal: true
+    run: once
     sources: &core_source_files
       - "{{.G_DEPS_CPP_CHECKSUM_FILE}}"
       - "{{.G_CORE_COMPONENT_DIR}}/cmake/**/*"
@@ -196,6 +197,7 @@ tasks:
 
   core-build:
     internal: true
+    run: once
     sources: *core_source_files
     deps:
       - "core-generate"
--- a/taskfile.yaml
+++ b/taskfile.yaml
@@ -173,12 +173,12 @@ tasks:
           INCLUDE_PATTERNS: ["{{.OUTPUT_DIR}}"]
 
   core:
-    cmds:
-      - task: "core-generate"
-      - task: "core-build"
+    deps:
+      - "core-build"
 
   core-generate:
     internal: true
+    run: "once"
     sources: &core_source_files
       - "{{.G_DEPS_CPP_CHECKSUM_FILE}}"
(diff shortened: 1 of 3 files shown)


quinntaylormitchell (changes requested):
I think that a simpler way to fix this is to change might be to change [line 8 of `taskfiles/tests/integration.yaml`](https://github.com/y-scope/clp/blob/251510ab3b41ef4f64ac2435879dcc11e4b49e5b/taskfiles/tests/integration.yaml#L8) from `deps:` to `cmds:`. iirc, dependencies run concurrently, and commands run sequentially.

@junhaoliao you will probably know which approach is best!

Nathan903:
> The `core` task previously invoked `core-build` using `cmds` instead of `deps`. Because of this, `go-task` could not deduplicate the build when multiple parallel test targets triggered the `core` task at the same time. This caused two parallel `cmake --build` processes to run simultaneously and corrupt each other's object files.

thx for the review. Yes, I just tested this much simpler fix (just 1 line change) 
```diff
 tasks:
   default:
-    deps:
+    cmds:
       - task: "clp-py-project-imports"
       - task: "core"
       - task: "package"

```
and this fix works. However, removing parallelism makes building slower. I tested both version of the fixes and this simpler fix adds 35 seconds to the build time of `task tests:integration`.

What do you guys think about this tradeoff of speed and simplicity

PR #2380, opened 2026-07-15
feat(telemetry): Add bytes-scanned metrics for query jobs.

rishikeshdevsot:
# Description

* Adds OpenTelemetry metrics to the query scheduler that track how much archive data is selected for each search/aggregation job, in both uncompressed (original log volume) and compressed (on-disk) bytes.

* Two new histogram metrics, emitted once per query job: `clp.query.uncompressed_bytes_scanned`, `clp.query.compressed_bytes_scanned`. Two new cumulative metrics: `clp.query.uncompressed_bytes_scanned_total`, `clp.query.compressed_bytes_scanned_total`

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **New Features**
  * Added telemetry for compressed and uncompressed archive bytes scanned during successful query tasks and jobs.
  * Query jobs now track cumulative scanned-byte totals and per-task archive sizes.
  * Added consistent bucket ranges for detailed scanned-byte measurements.
  * Improved visibility into archive scanning volume and query performance through success-based metrics.

* **Documentation**
  * Updated the telemetry reference with definitions for the new query-scheduling counters and histograms.


Nathan903:
When there is a `max_num_results`, a query can stop early if it has found enough results (equal to `max_num_results`). Do we want to count:
1. archive tasks that was dispatched, ran, and suceeded 
2. archive tasks that was dispatched, ran (ran and used cpu but might have failed or somehow stopped)
3. archive tasks that was dispatched.
4. all initially selected archive (disregarding `max_num_results`) which might not run because of early termination with `max_num_results`. If the scheduler selected 100 archives, and after scanning the first 10 archives we already got all the results we need, this measurement will still report the size of the 100 archives (not just the 10).

I think the current implementation is # 4. I think last time we discussed about cpu-time billing, which we want to measure the exact computation resource used. In that case, we already have `clp.query.task.duration` which is a proxy for cpu-time. # 2 might be a good supplement to it so we can correlate the bytes scanned with cpu-time. For data-based (usage) billing, I think either # 1, # 2, # 3 might make sense. Which one (ones) do we want? Or maybe just record all of them (as 4 separate metrics)?

rishikeshdevsot:
> I think the current implementation is # 4. I think last time we discussed about cpu-time billing, which we want to measure the exact computation resource used. In that case, we already have clp.query.task.duration which is a proxy for cpu-time. # 2 might be a good supplement to it so we can correlate the bytes scanned with cpu-time. For data-based (usage) billing, I think either # 1, # 2, # 3 might make sense. Which one (ones) do we want? Or maybe just record all of them (as 4 separate metrics)?

`#1` should be the way to go for data-based billing. `#2` would be a nice to have. I don't think `#3` or `#4` are needed.

Nathan903:
Got it, the PR will need a major change then, I think I will start from
scratch. I already got a draft for # 1 working, and will try to have # 2 as
well if I can get it in today

components/clp-py-utils/clp_py_utils/telemetry.py
```diff
@@ -5,12 +5,30 @@
...
+    1 << 24,  # 16 MiB
+    1 << 26,  # 64 MiB
+    1 << 28,  # 256 MiB
+    1 << 30,  # 1 GiB
+    1 << 32,  # 4 GiB
+    1 << 34,  # 16 GiB
+    1 << 36,  # 64 GiB
+    1 << 38,  # 256 GiB
+    1 << 40,  # 1 TiB
+    1 << 42,  # 4 TiB
```
rishikeshdevsot, line 28: any reason you stopped at 4TiB?

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -186,6 +186,59 @@ def _observe_outstanding_tasks(_options: metrics.CallbackOptions):
...
+    unit="By",
+    description=(
+        "Distribution of logical uncompressed archive bytes scanned per successful search job"
+    ),
+)
+compressed_bytes_scanned_histogram = meter.create_histogram(
+    "clp.query.compressed_bytes_scanned",
+    unit="By",
+    description=(
+        "Distribution of logical compressed archive bytes scanned per successful search job"
```
rishikeshdevsot, line 200: why do you say "logical" compressed archive bytes?

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -186,6 +186,59 @@ def _observe_outstanding_tasks(_options: metrics.CallbackOptions):
...
+)
+
+
+def _record_search_bytes_scanned(
+    job: SearchJob,
+    archive_sizes: tuple[int, int] | None,
+) -> None:
+    if archive_sizes is None:
+        logger.error(
+            "Archive-size metadata is missing for the completed search task; scan byte metrics "
```
rishikeshdevsot: "scan byte metrics" here should be reworded to "bytes scanned metrics"

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
```diff
@@ -186,6 +186,59 @@ def _observe_outstanding_tasks(_options: metrics.CallbackOptions):
...
+            "were not emitted."
+        )
+        return
+    uncompressed_size, compressed_size = archive_sizes
+    if uncompressed_size < 0 or compressed_size < 0:
+        logger.error(
+            "Search task result contains negative archive-size metadata; scan byte metrics were not"
```
rishikeshdevsot: same as above

rishikeshdevsot: Q2: are there cases when `uncompressed_size < 0`? asking to understand why this case is being handled.

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
Nathan903: its because the metric is not the actual bytes scanned, which could be less than the metric we are reporting here if clp didn't scan the entire archive during search. This metric is just summing up the size of archives according to the metadata DB.

components/clp-py-utils/clp_py_utils/telemetry.py
Nathan903: good point, my bad I didnt really thought about this when reviewing the code generated by AI. These are bin sizes so everything bigger than 4TB will go into this bin. Now thinking about this I think scanned uncompressed sizes probably will often go over 4TB. Do you know whats the typical scan sizes of clp deployments out there? Maybe I extend this to 1PB (assuming that anything above PB is an outlier so we just group them into the 1PB to infinite bin in the histogram)

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
Nathan903: done

components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py
Nathan903: No, I think unless theres a bug, it couldn't be negative. AI added this when I asked it to make the PR more robust with failure and edge cases (it added many other things, most I removed), and I thought I would keep it just to be safe since `uncompressed_size` is a BIGINT. lmk if u think its a good idea to just remove this check

PR #2406, opened 2026-07-21
feat(clp-rust-utils): Add the config and S3 helpers the Spider compression tasks need:

LinZhihao-723:
* Define `SpiderTaskExecutorConfig` and resolve its relative paths against `CLP_HOME`.
* Add `ArchiveOutputStorage` for both `S3` and `Fs`.
* Add `Database`'s fixed `table_prefix`.
* Add `resolve_dataset_name` to encapsulate the default dataset name resolution.
* Add `generate_s3_url`.

# Description

This PR depends on #2404, which adds the shared config types and helpers that the Spider compression tasks depend on. Split out of the compression-task implementation so that the changes to CLP's config mirror can be reviewed on their own; nothing in this PR has a caller yet.

**Archive output storage (`clp_config::package::config`)**

`ArchiveOutput` gains a `storage` field, typed as a new `ArchiveOutputStorage` enum with `Fs { directory }` and `S3 { staging_directory, s3_config }` variants. This mirrors the Python config, where archive output is either filesystem- or S3-backed and the two carry different fields.

`dataset_archive_storage_directory` derives a dataset's storage base — `s3_config.key_prefix` for S3, `directory` for `Fs` — joined with the dataset name. It mirrors Python's `add_dataset`, which computes the same value when registering a dataset, and it defaults a `None` dataset to `default` so the `CLP_S` naming rule lives in one place rather than at each call site.

**`Database::table_prefix`**

Mirrors Python's `CLP_METADATA_TABLE_PREFIX` (`"clp_"`). It is a fixed constant rather than a configurable value, so it is `#[serde(skip)]`: a value supplied in the YAML is ignored rather than honoured. The metadata table names are built from it, so allowing it to be overridden per-deployment would let a config file silently point a component at tables that don't exist.

**`SpiderTaskExecutorConfig`**

Previously an empty placeholder; now carries the four sections the task executor actually reads (`package`, `archive_output`, `tmp_directory`, `database`).

Its two accessors, `abs_tmp_directory` and `abs_archive_output_staging`, resolve the config's relative paths against `CLP_HOME` through a new `make_config_path_absolute` helper. That helper mirrors `clp_py_utils.core.make_config_path_absolute` exactly, **including its early return for paths that are already absolute**. This distinction is load-bearing: CLP's deployment tooling rewrites these paths to absolute container paths before writing the config a worker reads, and in that case `CLP_HOME` must not be prepended. Note the config *file* itself is located directly from `CLP_CONFIG_PATH` and is never joined with `CLP_HOME` — only relative path *values* inside it are resolved.

**Dataset naming (`dataset`)**

Adds `CLP_DEFAULT_DATASET_NAME` (`"default"`, mirroring Python) and `resolve_dataset_name`, which maps a `None` dataset onto it. `CLP_S` never omits the dataset — the CLI defaults it — so callers that receive an optional dataset resolve it through this one function instead of each deciding what a missing dataset means.

**S3 URLs (`s3`)**

Adds `generate_s3_url`, which builds an object URL from an optional endpoint, optional region, bucket, and key. A custom endpoint takes precedence (for MinIO and similar); otherwise it produces the virtual-hosted AWS form, with or without a region. `clp-s` consumes these URLs through its `--files-from` list.

# Checklist

  breaking change.

# Validation performed

* [x] E2e-tested in a dev branch.

## Summary by CodeRabbit

* **New Features**
  * Added helpers to resolve dataset archive storage locations across local and S3-backed storage.
  * Added automatic resolution of temporary and archive staging paths relative to CLP home.
  * Added S3 URL generation supporting default, regional, and custom endpoints.
  * Added a consistent default dataset name when none is provided.
* **Bug Fixes**
  * Default S3 staging directories are applied when omitted.
  * Database table prefixes are enforced to `clp_` during configuration handling.
  * API now reports unsupported S3 endpoint details for invalid inputs.


components/clp-rust-utils/src/clp_config/package/config.rs
```diff
@@ -251,16 +315,38 @@ impl Default for LogIngestor {
...
+    pub fn dataset_archive_storage_directory(&self, dataset: Option<&str>) -> String {
+        let base = match &self.storage {
+            ArchiveOutputStorage::Fs { directory } => directory.as_str(),
+            ArchiveOutputStorage::S3 { s3_config, .. } => s3_config.key_prefix.as_str(),
+        };
+        Path::new(base)
+            .join(resolve_dataset_name(dataset))
+            .to_string_lossy()
+            .into_owned()
+    }
```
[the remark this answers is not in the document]
LinZhihao-723: This is to mirror the existing logic in the code base (in Python utils).

components/clp-rust-utils/src/s3/url.rs
```diff
@@ -0,0 +1,77 @@
...
+    #[test]
+    fn custom_endpoint_trailing_slash_trimmed() {
+        assert_eq!(
+            generate_s3_url(Some("http://minio:9000/"), None, "logs", "a/b.json"),
+            "http://minio:9000/logs/a/b.json"
+        );
+    }
+
+    #[test]
+    fn custom_endpoint_ignores_region() {
```
Bill-hbrhbr: @coderabbitai open an issue.
Python's `generate_s3_url` (https://github.com/y-scope/clp/blob/ed9742a63b5d86248b2ce55ba03d4070bc73e244/components/clp-py-utils/clp_py_utils/s3_utils.py#L257) never ignores the region code even when a custom endpoint is already present. It contradicts the test case provided here. Fix either the Python or the Rust version to keep behaviours consistent across the codebase.

components/clp-rust-utils/src/clp_config/package/config.rs
```diff
@@ -279,6 +356,29 @@ impl Default for ArchiveOutput {
...
+#[serde(tag = "type")]
+pub enum ArchiveOutputStorage {
+    #[serde(rename = "fs")]
+    Fs { directory: String },
+
+    #[serde(rename = "s3")]
+    S3 {
+        #[serde(default = "default_archive_staging_directory")]
+        staging_directory: String,
+        s3_config: S3Config,
```
Bill-hbrhbr, line 370: @coderabbitai Open an issue:
https://github.com/y-scope/clp/blob/ed9742a63b5d86248b2ce55ba03d4070bc73e244/components/clp-rust-utils/src/clp_config/s3_config.rs#L17
Support credential-free workload in Rust.
Long-term speaking, AWS credentials should never be serialized into `ClpIoConfig`, persisted in compression job records, or forwarded through Spider task inputs. Instead, the coordinator and scheduler should pass only non-sensitive S3 location metadata, such as the endpoint, region, bucket, and object key. Each task executor should retrieve credentials locally using the AWS default credential provider chain, a mounted AWS profile, workload identity, or environment variables injected at runtime.

Bill-hbrhbr (dismissed):
opened 2 issues for future improvements; the rest looks good.

components/clp-rust-utils/src/clp_config/package/config.rs
LinZhihao-723: To be fair: I think this is out of the scope for "Rust" components. This is a package-level design problem, which we already have a solution for through a new component called credential-manager. This component has been completed in a private branch, but we haven't had time to open-source it.

components/clp-rust-utils/src/s3/url.rs
LinZhihao-723: Since this may affect some of our users, I've updated the implementation to match the feature parity with the Python implementation. Please check.

components/clp-rust-utils/src/s3/url.rs
```diff
@@ -0,0 +1,191 @@
...
+        .expect("scheme should be unconditionally captured")
+        .as_str();
+    let endpoint = captures
+        .name(ENDPOINT_GROUP)
+        .expect("endpoint should be unconditionally captured")
+        .as_str();
+
+    Ok(region_code.map_or_else(
+        || format!("{scheme}://{s3_prefix}{endpoint}/{bucket}/{object_key}"),
+        |region| format!("{scheme}://{s3_prefix}{region}.{endpoint}/{bucket}/{object_key}"),
```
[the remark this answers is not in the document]
Bill-hbrhbr: python side doesn't have these validations either

Bill-hbrhbr (approved):

PR #2411, opened 2026-07-22
feat(clp-tdl-package): Implement the clp-s S3 compression task.

LinZhihao-723:
# Description

This PR depends on #2406.

This PR fills in the compression task registered by `clp-tdl-package`: it compresses one partition of S3 objects into single-file archives with `clp-s`, indexes each archive, and uploads it to S3, returning the archives' metadata for the commit task. This is the worker behind the `compression::clp_s_s3_compress` TDL task; the commit task remains unimplemented.

**Compression flow**

`compress` mirrors the Python compression task:

* Writes the S3 object URLs to a `--files-from` list file in the tmp directory.
* Runs `clp-s` with `--print-archive-stats`, streaming each archive's stats line off stdout as it is produced.
* For each reported archive, spawns a finisher that concurrently uploads it to S3 (`put_object`) and runs the `indexer` over it, then the staged copy is cleaned up.

Archives are streamed rather than collected up front so uploads and indexing overlap `clp-s` still running, and the finishers are drained before the task returns so no work is abandoned. The first finisher error (or a `clp-s` failure) fails the task.

**Unstructured (log-converter) input**

Structured (JSON) input is compressed directly from S3. Unstructured text input is first converted by `log-converter` into a local directory that `clp-s` then compresses, selected by the `unstructured` flag on the task's `clp-s` options. Following the Python task, the unstructured path drops S3 auth from the `clp-s` invocation (it reads the local converted files, not S3) and pins the timestamp key to `timestamp`; only `log-converter` carries the S3 credentials in that case.

**Subprocess I/O**

Both child processes capture stderr so it can be surfaced on failure. `clp-s` pipes both stdout and stderr, so its stderr is drained on a dedicated thread to avoid a pipe-buffer deadlock while stdout is streamed, and it is killed and reaped before an error is returned. `log-converter` discards stdout (`/dev/null`) and has only a single stderr pipe, so a straight read-to-EOF-then-wait is deadlock-free there; the wait is issued unconditionally so the child is always reaped even if reading its stderr fails.

**Temp-file cleanup**

`TmpFileDeleter` registers the list file, the converted directory, and each archive's staging path, and deletes them all when `compress` returns — on success or on any early error. Deletion is sequential and blocking (a handful of local paths, `unlink` has no async form, and a destructor cannot be async), and a path that cannot be deleted is logged as a warning rather than propagated.

**Observability**

Each stage logs structured `tracing` events; every subprocess or S3 failure is logged with its captured context (exit status, stderr, archive id, S3 key) at the point it occurs.

# Checklist

  breaking change.

# Validation performed

* [ ] Ensure all workflows pass
* [x] Exercised `compress` end-to-end (outside Spider) against `clp-s`, `indexer`, and `log-converter` from `build/core`, a local MariaDB, and a real S3 bucket, with a JSON `tracing` subscriber capturing logs. Each scenario asserts the return value, the S3 side effect, and that the tmp directory is left clean:

| Scenario | Input | Output | Logging | Tmp dir |
| --- | --- | --- | --- | --- |
| Structured JSON | S3 object of JSON records, `unstructured: false` | `Ok` with the produced archive's metadata; the archive object is present in S3 under `<key_prefix>/<dataset>/<archive_id>`; the dataset's `column_metadata` table gains rows (indexer ran) | `started` + `completed successfully` info lines | list file removed |
| Unstructured text | S3 object of plaintext logs, `unstructured: true` | `Ok` with archive metadata; archive object present in S3 (log-converter ran before clp-s, which compressed the local converted directory) | `started` + `completed successfully` info lines | list file **and** the `-converted` directory removed |
| Metadata DB unreachable | valid input, DB port pointed at a closed port | `Err` (indexer exits non-zero) | `error` line `indexer exited on failure.` with `status` + `archive_path` | cleaned |
| Missing S3 object | `object_keys` referencing a non-existent S3 key | `Err` (clp-s exits non-zero) | `error` line `clp-s exited on failure.` with the captured clp-s stderr (S3 read, 404) | cleaned |
| Invalid S3 credentials | valid input, wrong secret access key | `Err` (clp-s exits non-zero on the credential path forwarded to the child) | `error` line `clp-s exited on failure.` with the captured clp-s stderr (S3 read, 403) | cleaned |
| Malformed input content | readable S3 object of invalid JSONL, valid credentials | `Err` (clp-s exits non-zero) | `error` line `clp-s exited on failure.` with the captured clp-s stderr (JSON parse error, after the object was read) | cleaned |

Note: surfaced by the DB-unreachable and malformed-content cases: because a finisher uploads to S3 and runs the indexer concurrently, once an archive is produced its S3 upload is not rolled back if the indexer (or the overall task) then fails. A failed task can thus leave an orphan archive object in S3 (local staging is still cleaned up); the object has no metadata row, so it is invisible to the commit task and left to retention/GC. This is a known behavior and will be addressed in the future.

## Summary by CodeRabbit

- **New Features**
  - Added task support for compressing data from Amazon S3.
  - Supports both structured and unstructured log inputs, including automatic conversion where required.
  - Archives are uploaded to S3 and indexed as they are produced.
  - Compression results now include archive statistics and improved error reporting.
- **Known Limitations**
  - The compression commit task is not yet available.


components/clp-tdl-package/src/task/compression/compress.rs
```diff
@@ -1 +1,1071 @@
...
+/// How clp-s reads its input.
+enum ClpSInput {
+    /// A `--files-from` list file of S3 object URLs; clp-s reads the objects directly from S3.
+    FilesFrom(PathBuf),
+
+    /// A local directory of converted logs files passed positionally. clp-s reads the local files.
+    Directory(PathBuf),
+}
+
+/// The owned inputs one spawned finisher needs to publish a single archive.
```
sitaowang1998:
```suggestion
/// The owned inputs needed by a spawned finisher to publish a single archive.
```

components/clp-tdl-package/src/task/compression/compress.rs
```diff
@@ -1 +1,1071 @@
...
+/// [`AWS_DEFAULT_REGION`] when the config omits it.
+///
+/// # Returns
+///
+/// The S3 client the archives are uploaded through.
+fn build_s3_client(runtime: &tokio::runtime::Handle, s3_config: &S3Config) -> aws_sdk_s3::Client {
+    let region = s3_config
+        .region_code
+        .as_ref()
+        .map_or(AWS_DEFAULT_REGION, NonEmptyString::as_str);
```
sitaowang1998, line 299: Should we log when a user provided region code is invalid and need to use default value?

components/clp-tdl-package/src/task/compression/compress.rs
```diff
@@ -1 +1,1071 @@
...
+fn prepare_clp_s_input(
+    ctx: &spider_tdl::TaskContext,
+    clp_home: &Path,
+    config: &SpiderTaskExecutorConfig,
+    clp_s_option: &ClpSCompressionOption,
+    list_path: &Path,
+    credential_env: Vec<(&'static str, String)>,
+    tmp_file_deleter: &mut TmpFileDeleter,
+) -> anyhow::Result<(ClpSInput, Vec<(&'static str, String)>)> {
+    if !clp_s_option.unstructured {
```
sitaowang1998, line 445: Not really in scope of this review, but we should avoid using boolean flag here. I understand that for serialization a boolean is needed, but that should be a serialization details, hidden from user-facing code.

components/clp-tdl-package/src/task/compression/compress.rs
```diff
@@ -1 +1,1071 @@
...
+/// Uploads a local file to S3 through `PutObject`.
+///
+/// # Errors
+///
+/// Returns an error if:
+///
+/// * Forwards [`aws_sdk_s3::primitives::ByteStream::from_path`]'s return values on failure.
+/// * Forwards [`aws_sdk_s3::operation::put_object::builders::PutObjectFluentBuilder::send`]'s
+///   return values on failure.
+async fn put_file(
```
sitaowang1998: How about `upload_s3_file` or `upload_file_to_s3`?

components/clp-tdl-package/src/task/compression/compress.rs
```diff
@@ -1 +1,1071 @@
...
+            format!(
+                "failed to spawn log-converter at {}",
+                log_converter_bin.display()
+            )
+        })?;
+
+    let mut stderr = child
+        .stderr
+        .take()
+        .context("log-converter stderr was not captured")?;
```
sitaowang1998: This early return might leave the child unwaited on and becomes a zombie.

sitaowang1998 (changes requested):
I am not familiar with clp s3 details, so I assume that the overall algorithm is correct during my review.

components/clp-tdl-package/src/task/compression/compress.rs
LinZhihao-723: This map is not for error; it's for cases where the region code is omitted.

components/clp-tdl-package/src/task/compression/compress.rs
LinZhihao-723: Hmmm, I'm not sure I fully understand. Can you elaborate on the desired interface?

components/clp-tdl-package/src/task/compression/compress.rs
LinZhihao-723: This should be unreachable in reality. Let me switch to panic for these cases.

components/clp-tdl-package/src/task/compression/compress.rs
sitaowang1998: If should be an enum.

LinZhihao-723: Hmmm, I know this is out of the scope for this PR. But I'm not sure if we really need an enum for this. If we use a binary flag to determine whether the input source is from s3, for example, `is_from_s3`, which differs from "from filesystem", then I think it would be a good practice to define an enum for it, even it only has two options. However, if an option is purely a binary flag of two opposite concepts like "structured" and "unstructured", I don't really see the benefit of using an enum. You can even match this flag like `match unstructured { true => xxx, false => xxx, }` with the readability preserved. So I'd argue that using a binary flag for this case should be sufficient.

components/clp-tdl-package/src/task/compression/compress.rs
LinZhihao-723: Actually this is also stale since I later made the spawned finisher as a member method. I will refactor it.

LinZhihao-723:
> I am not familiar with clp s3 details, so I assume that the overall algorithm is correct during my review.

Added the scenarios I've tested in the PR description.

sitaowang1998 (dismissed):

sitaowang1998 (approved):

PR #2415, opened 2026-07-22
feat(clp-tdl-package): Implement the clp-s compression commit task.

LinZhihao-723:
# Description

This PR depends on #2411.

This PR fills in the commit task registered by `clp-tdl-package`: once every compression task of a job has succeeded, it publishes the produced archives to CLP's metadata DB and marks the job succeeded. This is the worker behind the `compression::commit` TDL task; with it, the compress + commit pair is complete.

**Commit flow**

`commit` runs the whole finalization in a single DB transaction, mirroring the Python commit path:

* Looks up the CLP compression job by reverse-lookup on `spider_id`, selecting it `FOR UPDATE`.
* Idempotently registers the dataset (a `None` dataset resolves to the `clp-s` default) in the `datasets` table.
* Inserts every archive's metadata into `<prefix><dataset>_archives`.
* CAS-transitions the job from `RUNNING` to `SUCCEEDED`, recording total uncompressed/compressed sizes and a duration derived from the DB clock, then commits.

The transition is a compare-and-set: the job must be `RUNNING`. A job already `SUCCEEDED` is a no-op (so a retried commit is safe), and any other state is refused. Because everything runs in one transaction, a failure at any step (e.g. an archive insert) rolls back the dataset registration too, leaving no partial state.

**Wrapper**

The `commit_task` wrapper deserializes the msgpack `CompressionTaskOutput`s from the task-graph outputs, **validates that every output belongs to the same dataset** (a job spans exactly one dataset), flattens their archives, and drives `commit`.

**Status enum encoding**

`CompressionJobStatus` gains `#[derive(sqlx::Type)]` (over its existing `#[repr(i32)]`), so it binds and decodes directly as the integer status column rather than through a hand-rolled conversion.

**Observability**

Following the task's logging decision, the async `commit` never logs errors — it returns them with context — while the (non-async) `commit_task` wrapper logs a failure once, with the full error chain. `commit` emits info lines for start, completion, and the already-committed no-op.

# Checklist

  breaking change.

# Validation performed

* [x] Exercised `commit` and the `commit_task` wrapper (outside Spider) against a local MariaDB, with a JSON `tracing` subscriber capturing logs. Each scenario asserts the return value, the DB side effects, and the logging; the failure cases additionally confirm that the async worker logs no error (only the wrapper does):

| Scenario | Input | Output | Logging | DB side effects |
| --- | --- | --- | --- | --- |
| Happy path | RUNNING job, `Some("default")`, 3 archives | `Ok` | `started` + `completed successfully` info lines | job → SUCCEEDED with total sizes and a non-null duration; `datasets` row written; 3 archive rows (`creator_id=''`, `creation_ix=0`) |
| Default dataset | RUNNING job, `dataset = None` | `Ok` | `started` + `completed successfully` | `None` resolves to `default`; datasets row `name=default`; archives in `<prefix>default_archives` |
| Idempotent re-commit | job already SUCCEEDED | `Ok`, no-op | `started` + `already committed` info lines | none — no duplicate rows, job untouched |
| Wrong state | job PENDING/FAILED | `Err` (refuses to commit) | `started` info only; no error line | none; job unchanged |
| No job for `spider_id` | no matching row | `Err` | `started` info only; no error line | none |
| Transaction atomicity | RUNNING job, archives table absent | `Err` (archive insert fails) | `started` info only | **rolled back** — the earlier `datasets` insert is gone; job stays RUNNING |
| DB unreachable | port pointed at a closed port | `Err` (pool creation) | `started` info only; no error line | none |
| Missing DB credentials | `CLP_DB_USER`/`CLP_DB_PASS` unset | `Err` | `started` info only; no error line | none (bails before the pool) |
| Empty archives | `archives = []` | `Err` | `started` info only | none (bails before any DB call) |
| Invalid dataset name | `Some("bad name!")` | `Err` | `started` info only | none (bails before any DB call) |
| Multi-dataset rejection (wrapper) | two outputs with different datasets | `Err` (more than one dataset) | wrapper `started` only | none — rejected before `commit` |
| Multi-output aggregation (wrapper) | several outputs, same dataset, RUNNING job | `Ok` | wrapper + worker `started`, `completed successfully` | every output's archives flattened and inserted; job SUCCEEDED |
| Non-commit context (wrapper) | a non-commit task with no outputs | `Err` (commit must run as a commit task) | wrapper `started` only | none |
| Logging split | a worker failure driven through `commit_task` | `Err` | wrapper emits `CLP commit task failed.` with the full error chain, while the same failure called on `commit` directly emits no error line | none |

## Summary by CodeRabbit

* **New Features**
  * Added compression tasks for processing log data from S3, including support for structured and unstructured logs.
  * Added automatic archive creation, upload, and metadata registration.
  * Added optional dataset support with dataset-specific archive storage.
  * Added commit processing to record completed compression jobs and archive details.
  * Added default staging and temporary storage locations for task execution.

* **Bug Fixes**
  * Improved handling of repeated or already-completed compression jobs.
  * Added validation for dataset names and compression task outputs.


components/clp-tdl-package/src/task/compression/mod.rs
```diff
@@ -13,22 +13,45 @@ mod compress;
...
+    if outputs.iter().any(|output| output.dataset != dataset) {
+        return Err(TdlError::ExecutionError(
+            "compression task outputs belong to more than one dataset".to_owned(),
+        ));
+    }
+    let archives = outputs
+        .into_iter()
+        .flat_map(|output| output.archives)
+        .collect();
+    crate::common::runtime()
```
sitaowang1998, line 51: Do we allow this import style?

components/clp-tdl-package/src/task/compression/commit.rs
```diff
@@ -1 +1,234 @@
...
+/// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
+async fn insert_archives(
+    tx: &mut sqlx::Transaction<'_, sqlx::MySql>,
+    archives_table: &str,
+    archives: &[ArchiveMetadata],
+) -> anyhow::Result<()> {
+    for archives in archives.chunks(1000) {
+        let mut builder = sqlx::QueryBuilder::<sqlx::MySql>::new(format!(
+            "INSERT INTO `{archives_table}` (id, begin_timestamp, end_timestamp, \
+             uncompressed_size, size, creator_id, creation_ix);"
```
sitaowang1998:
```suggestion
             uncompressed_size, size, creator_id, creation_ix)"
```

sitaowang1998 (changes requested):
Just curious. How do we handle empty table if the commit fail right after table creation?

components/clp-tdl-package/src/task/compression/mod.rs
LinZhihao-723: sorry which import style?

components/clp-tdl-package/src/task/compression/commit.rs
LinZhihao-723: I tried to remove it, but the formatter will append an empty space after `)`.

LinZhihao-723:
> Just curious. How do we handle empty table if the commit fail right after table creation?

- To clairfy: The table is created in compression coordinator.
- The desired behavior for now is: if the compression job fails, we just leave the table empty there. The proper solution is like what Kirk mentioned today: we should enforce the dataset creation as a separate step executed before compression jobs.

components/clp-tdl-package/src/task/compression/commit.rs
LinZhihao-723: Actually it should be a space. My bad. The query builder will append values into this query.

components/clp-tdl-package/src/task/compression/mod.rs
sitaowang1998: I have the impression that we do not allow `crate::` and should use import instead.

LinZhihao-723: hmmm I didn't explicitly add such rules in the standard.

sitaowang1998 (approved):

PR #2416, opened 2026-07-22
feat(compression-coordinator): Add the compression coordinator executable scaffolding.

LinZhihao-723:
# Description

This PR adds the compression coordinator executable target.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* [x] Ensure all workflows pass.
* [x] Ensure the compiled target is executable:
  ```
  ☁  clp-dev [compression-coordinator-bin] ⚡  ./build/rust-targets/release/compression-coordinator --help
  Run the compression coordinator.

  Usage: compression-coordinator --config <PATH>

  Options:
    -c, --config <PATH>  Path to the configuration file
    -h, --help           Print help
  ```

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **New Features**
  * Added a new command-line tool for the compression coordinator.
  * Includes a required `--config` / `-c` option to specify the configuration file path.
  * Starts with an asynchronous runtime and parses command-line arguments.
* **Chores**
  * Updated the container image to include the new compression coordinator executable.

PR #2402, opened 2026-07-20
feat(compression-coordinator): Implement the compression job-submission API for Spider.

LinZhihao-723:
# Description

This PR depends on #2401.

This PR implements `S3CompressionJobSubmitter` for [`SpiderClient`][spider], filling in the trait added by the preceding scaffolding PR.

**Submitting a job**

`submit_s3_compression_job` builds a task graph with one compression task per `S3InputSource` and a commit task as the graph's termination task, so the commit runs exactly once after every compression task has succeeded. The task inputs (`ClpSCompressionOption`, the dataset, and the input source) are msgpack-encoded into opaque `bytes` payloads, matching how the tasks decode them on the Spider side.

Execution policies are applied as given: each compression task receives the policy paired with its input source, and the commit task receives `commit_task_execution_policy`. The implementation makes no decisions about timeouts, retries, or instance counts — that is the caller's, since only the caller knows how much data a partition holds and how long its jobs typically run.

The TDL package and task-function names are duplicated here as constants rather than imported: the package that defines them depends on this crate, so importing them would form a dependency cycle. They must be kept in sync with the TDL package's definitions, which the constants are annotated to say.

**Running a job to completion**

`run_s3_compression_job_to_completion` starts the job idempotently — an "already started" rejection from the cluster is treated as success, so the call is safe whether the job is not-yet-started, running, or already terminal — then polls its state with an exponential backoff bounded by the caller-supplied `initial_poll_backoff` and `max_poll_backoff`. The terminal state maps onto `CompressionJobOutcome`; for a failed job, the cluster's error message is fetched and attached, falling back to a placeholder if that fetch itself fails, so a reporting failure can't mask the underlying job failure.

[spider]: https://github.com/y-scope/spider

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* [x] Ensure all workflows pass.
* [x] E2e-tested in a dev-branch.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **New Features**
  * Added support for submitting S3 compression jobs with per-input and execution-policy task setup.
  * Added job monitoring with exponential backoff until a terminal outcome is reached.
[...]

PR #2417, opened 2026-07-23
feat(compression-coordinator): Add the compression coordinator implementation.

LinZhihao-723:
# Description

This PR is my birthday gift, and it depends on #2416.

This PR adds the `Coordinator` that drives CLP's compression pipeline: it recovers jobs left in-flight by a previous instance, polls the metadata DB for pending compression jobs, and dispatches each one to Spider — plus the configuration and `clp-rust-utils` support the coordinator needs. This is the layer above the job handle (#2405) and the Spider submitter (#2402); with it, the coordinator binary can discover work, hand it off, and survive a restart.

**WARNING**: Compression coordinator will not work until the methods marked as `TODO` in #2405 are implemented.

**Poll loop**

`Coordinator::run` loops until cancelled:

* Fetches the pending jobs to dispatch. The first fetch after startup returns every pending job (`WHERE status = PENDING ORDER BY id ASC`) so that jobs a previous instance had already picked up are re-dispatched on a restart; every subsequent fetch returns only the jobs this instance hasn't dispatched yet (`WHERE status = PENDING AND dispatch_time IS NULL ORDER BY id ASC`).
* For each job, deserializes its Brotli-`MessagePack` `clp_config`, builds an `S3CompressionJobHandle` via `create_job_handle`, and spawns a detached task to drive it to completion.
* Sleeps for the configured polling interval, waking early when the cancellation token fires.
* Once the sleep elapses, stamps the jobs dispatched in this iteration with the database's current time (`UPDATE ... SET dispatch_time = CURRENT_TIMESTAMP() WHERE id IN (...)`, chunked at 1000 IDs and run in a single transaction) so the next fetch skips them. Deferring this update until after the sleep keeps it from contending with concurrent job submissions during the poll interval.

**De-duplication via `dispatch_time`**

Each dispatched row is marked with a dedicated `dispatch_time` column (initialized to `NULL`) rather than tracked by a monotonic in-memory ID cursor. A job with ID `N` can be committed and become visible before job `N - 1`, so an ID-ordered cursor that advances past `N` would skip the later-committed `N - 1` permanently. Keying off a per-row `dispatch_time` marker makes the fetch independent of ID ordering, while re-fetching all pending jobs on the first poll after startup ensures nothing a crashed instance left behind is lost.

**Startup recovery**

`Coordinator::new` recovers jobs that a previous coordinator instance had already submitted to Spider — those still `RUNNING` with a non-null `spider_id`. For each, it rebuilds the handle and spawns a detached task that resumes waiting on the existing Spider job (`recover`) rather than resubmitting it, so a coordinator restart doesn't duplicate or abandon work. A recovered job whose stored config can't be deserialized is marked `FAILED` and skipped.

`new` also connects to the Spider cluster and resolves the coordinator's Spider resource group, and hands back a `CancellationToken` the binary uses to request a graceful shutdown.

**Resource-group registration**

`get_or_create_resource_group_id` idempotently registers the configured resource group with Spider and caches the assigned ID in a `spider_resource_groups` table, so a restart reuses the same group rather than creating a new one.

**Per-job failure handling**

Handle construction is centralized in `create_job_handle`, shared by both scheduling and recovery. A job whose config can't be deserialized, or whose handle can't be constructed, is marked `FAILED` with a status message rather than left stranded in `PENDING`/`RUNNING` — except an unsupported input config, which is only warned and left for another handler. Failures that occur once a handle is running are owned by the handle itself.

**Configuration**

Adds `compression_coordinator` and `spider` sections to the package config. The coordinator section is `#[serde(default)]` with sensible defaults (polling interval, result-poll backoff, task retry limits, DB pool size, termination and commit-task timeouts); intervals and sizes use `NonZero*` types so a zero can't be configured.

**`clp-rust-utils` support**

* `BrotliMsgpack::deserialize` — the inverse of the existing serializer, so the coordinator can read the `clp_config` the ingestor wrote.
* `ClpIoConfig` and its input/output config structs now derive `Deserialize`.
* `CompressionJobStatus` derives `sqlx::Type`, so it binds directly as the integer status column.
* A `MsgpackDecode` error variant (mapped to `MalformedData` in the API server).

**Binary entry point**

The `compression-coordinator` binary takes a single `--config <PATH>` argument. On startup it initializes logging, loads the YAML package config, and reads the metadata-DB credentials from the `CLP_DB_USER`/`CLP_DB_PASS` environment variables. It requires both the `compression_coordinator` and `spider` config sections, erroring out if either is missing. It then builds the metadata-DB connection pool (sized by `database_connection_pool_size`), constructs the `Coordinator` — which performs startup recovery — and spawns its poll loop before handing off to the shutdown handling below.

**Shutdown**

The binary spawns the coordinator, then waits on `SIGTERM`/`ctrl-c` or the coordinator returning on its own. On a signal it cancels the token and joins the coordinator within a configurable termination timeout, aborting if it overruns.

# Checklist

  breaking change.

# Validation performed

* [x] E2e-tested in a dev branch with a real Spider cluster.

## Summary by CodeRabbit

- **New Features**
  - Added a compression coordinator service that dispatches, monitors, and recovers compression jobs.
  - Added configuration options for Spider connectivity, resource groups, polling, retries, timeouts, and database capacity.
  - Added command-line configuration loading and graceful shutdown handling.
  - Added support for reading compressed MessagePack data and deserializing job configurations.

- **Bug Fixes**
  - MessagePack decoding failures are now reported as malformed data.
  - Invalid job configurations are marked as failed with an explanatory status.


components/compression-coordinator/src/bin/compression_coordinator.rs
```diff
@@ -0,0 +1,141 @@
...
+        .await?;
+
+    let mut coordinator_handle = tokio::spawn(coordinator.run());
+
+    let mut sigterm = tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())
+        .expect("failed to listen for SIGTERM");
+
+    // `None` if a shutdown signal arrived while the coordinator is still running; `Some` if the
+    // coordinator returned on its own (an early exit, possibly on error).
+    let early_exit = tokio::select! {
```
sitaowang1998: `early_exit` is a confusing name since it holds coordinator result.

sitaowang1998 (commented):
LGTM

components/compression-coordinator/src/bin/compression_coordinator.rs
LinZhihao-723: Changed to `early_exit_result`.

sitaowang1998 (dismissed):

20001020ycx:
**Environment**

  * `clp-package`: built locally against the current main (Hash: 326f6c62fc6d5c7389cfa515472e51032c0bf509)
  * Spider subchart 0.1.4 from the official y-scope/spider Helm repo.
  * Worker: built locally which includes the change from #2434, official image is still on flight
  * Inputs: 4 JSON log objects in S3 (MinIO), on kind.

**Procedure and results**
Ingest logs with log-ingestor from S3, compress with Spider and search with Celery. Results returned as expected.

20001020ycx (approved):
Confirmed e2e working with the methodology explained in https://github.com/y-scope/clp/pull/2417#issuecomment-5119566360

PR #2408, opened 2026-07-21
build(spider): Add Dockerfile and GitHub CI for CLP-tailored Spider Huntsman worker image.

20001020ycx:
# Description

This PR adds a Dockerfile that builds on top of the Spider Huntsman's worker for integration with the compression coordinator, plus a CI workflow to publish them to GHCR. 

This PR also refactors `build.sh`, and addresses a bug discovered in https://github.com/y-scope/spider/pull/381.

After this PR, the clp-artifact-build workflow publishes one new image to GHCR:
```
ghcr.io/y-scope/clp/clp-spider-worker:main (amd64-only)
```

 > [!NOTE]
  > This PR depends on #2404. Because the image rebuilds `libclp_tdl_package.so` on every commit, it is **not** blocked by the upcoming implementation PRs for `libclp_tdl_package.so` (Note: #2404 only writes the scaffolding of `libclp_tdl_package.so`). With that being said, the resulting  `libclp_tdl_package.so` won't be functional until those implementation PRs are merged.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

All CI flow passes, the new workflow `spider-worker-image` builds successfully, the artifact will be pushed once this PR lands to main.

We build the CLP Spider worker image locally and run the steps below to verify it carries a loadable `libclp.so` and a working `clp-s`/`indexer`/`log-converter`. The goal: confirm the image gives the Spider task executor everything it needs to load the CLP TDL package and run a compression job.

1. Build the image:
```
task docker-images:spider-worker
```
Produces `clp-spider-worker:dev-<user>-<short>` (referenced as `$IMG` below).

2. Confirm `libclp.so` is staged where the executor resolves it (`${SPIDER_TDL_PACKAGE_DIR}/clp/libclp.so`) and exports the four FFI symbols the executor calls:
```
docker run --rm --entrypoint bash "$IMG" -c \
  'nm -D "$SPIDER_TDL_PACKAGE_DIR/clp/libclp.so" | grep -o "__spider_tdl_package_[a-z_]*" | sort -u'
```
Expected:
```
__spider_tdl_package_execute
__spider_tdl_package_get_name
__spider_tdl_package_get_version
__spider_tdl_package_init
```
[...]

PR #2418, opened 2026-07-23
feat(helm)!: Add the Spider subchart and compression-coordinator wiring for Spider-orchestrated compression.

20001020ycx:
# Description

See PR title. At a high level, this PR integrates Spider into the CLP Helm chart as an optional subchart that, when enabled, orchestrates compression jobs in place of the Celery-based scheduler and workers. Note, compression coordinator, compression scheduler, and compression worker are on regardless of Spider's toggle; their internal implementation handles the routing of the jobs with respect to the types.

Note: the scope of the integration is only limited to compression jobs, search jobs routing shall fall back to the Celery's query scheduler + worker.

> [!IMPORTANT]
> **Blocked on y-scope/spider#413，y-scope/spider#415, y-scope/clp#2408, complete compression coordinator implementation (y-scope/clp#2417, y-scope/clp#2420, and y-scope/clp#2421 ), and complete TDL package implementation (y-scope/clp#2406, y-scope/clp#2411, y-scope/clp#2415).**

> [!NOTE]
> This PR is breaking because we set `fullnameOverride` to global. A user now needs to run `helm install --set global.fullnameOverride` rather than `helm install --set fullnameOverride`.

# Validation performed

> [!NOTE]
 >The validation below uses only official artifacts:
 > - ghcr.io/y-scope/clp/clp-package and ghcr.io/y-scope/clp/clp-spider-worker, CI-published from main
 > - the Spider subchart (0.1.8) resolved from the Spider repo's gh-pages
 > - an in-cluster MinIO as the S3 endpoint (via s3_config.endpoint_url), standing in for AWS S3

## Full-Spider compression E2E: the job executes in Spider and the result is searchable

0. Environment: a single-node kind cluster; an S3 bucket `clp-logs` with 4 JSONL log objects under `demo-logs/`, e.g. `{"ts":1700000001,"level":"INFO","msg":"demo log line 1"}`; an empty S3 bucket `clp-archives` for archive output.

1. Overlay runtime values for `values.yaml` during deployment, you may review the complete `values.yaml` [here](https://github.com/20001020ycx/clp/blob/docs/2026-07-28-spider-demo-values/notes/spider-demo-values.yaml).

```yaml
spider:
  enabled: true

clpConfig:
  logs_input: # S3 ingestion config (type s3 + bucket credentials).
  archive_output: # S3 archive-output config; `region_code` must be omitted with a custom endpoint_url (the query path would build http://<region>.<endpoint-host>/..., which doesn't resolve).
```

```
helm dependency update tools/deployment/package-helm
helm install test tools/deployment/package-helm -f <overlay above>
kubectl get pods | grep -E "compression|ingestor|spider"
```
Expected (no `compression-scheduler`/`compression-worker` pods exist):
```
spider-database-0                                   1/1     Running
spider-scheduler-95fd9ddf-s29qm                     1/1     Running
spider-storage-76fd7d95b5-dtwp8                     1/1     Running
spider-worker-6996fd4b8f-5kzm9                      1/1     Running
test-clp-compression-coordinator-5765bbb96b-vssl5   1/1     Running
test-clp-log-ingestor-7bbfd478d7-k9dmk              1/1     Running
```

2. Start ingestion through the log-ingestor's REST API:
```
curl -X POST http://test-clp-log-ingestor:3002/s3_scanner -H 'Content-Type: application/json' \
  -d '{"bucket_name": "clp-logs", "key_prefix": "demo-logs/", "endpoint_url": "http://minio:9000",
       "region": "us-east-1", "scanning_interval_sec": 5, "buffer_config": {"timeout_sec": 10}}'
```
Expected: `{"id":1}`. ~30 s later, in the CLP DB:
```
kubectl exec test-clp-database-0 -- mariadb --table -u clp-user -ppass clp-db \
  -e "SELECT id, \`key\`, status, compression_job_id FROM ingested_s3_object_metadata;"
```

| id | key | status | compression_job_id |
|---|---|---|---|
| 1 | demo-logs/app-1.jsonl | compressed | 1 |
| 2 | demo-logs/app-2.jsonl | compressed | 1 |
| 3 | demo-logs/app-3.jsonl | compressed | 1 |
| 4 | demo-logs/app-4.jsonl | compressed | 1 |

3. The compression coordinator drove the full submit→execute→result loop:
```
kubectl logs deploy/test-clp-compression-coordinator | grep -oE '"message":"[^"]*"|"outcome":"[^"]*"'
```
Expected:
```
"message":"Scheduling new job."
"message":"Starting compression job."
"message":"Submitted compression job to Spider."
"message":"Compression job submitted."
"message":"Compression job submission persisted."
"message":"Compression job reached a terminal state."
"outcome":"Succeeded"
```
and in the CLP DB (`status=2` is `SUCCEEDED`):
```
kubectl exec test-clp-database-0 -- mariadb --table -u clp-user -ppass clp-db \
  -e "SELECT id, status, num_tasks, spider_id FROM compression_jobs;"
```

| id | status | num_tasks | spider_id |
|---|---|---|---|
| 1 | 2 | 1 | 1 |

4. Spider's own database confirms the execution:
```
kubectl exec spider-database-0 -- mariadb --table -u spider-user -pspider-password spider-db \
  -e "SELECT id, resource_group_id, state FROM jobs; SELECT external_id FROM resource_groups;
      SELECT id, state FROM execution_managers;"
```
Expected:

`jobs`
| id | resource_group_id | state |
|---|---|---|
| 1 | 1 | Succeeded |

`resource_groups`
| external_id |
|---|
| compression-coordinator |

`execution_managers`
| id | state |
|---|---|
| 1 | Alive |

5. The archive the job produced is traceable end to end, at spider's job table:
```
kubectl exec spider-database-0 -- mariadb -u spider-user -pspider-password spider-db \
  -e "SELECT serialized_job_outputs FROM jobs\G"
```
```
serialized_job_outputs:  ...default...$8ae188fb-8775-414a-8bdb-9b135ec91292...
```

The commit task registered the same archive in CLP's metadata table:
```
kubectl exec test-clp-database-0 -- mariadb --table -u clp-user -ppass clp-db \
  -e "SELECT id, uncompressed_size, size FROM clp_default_archives;"
```

| id | uncompressed_size | size |
|---|---|---|
| 8ae188fb-8775-414a-8bdb-9b135ec91292 | 228 | 648 |

The compress task uploaded the same archive to the output bucket:
```
mc ls -r s3/clp-archives/
```
```
[2026-07-24 21:08:22 UTC]   648B STANDARD archives/default/8ae188fb-8775-414a-8bdb-9b135ec91292
```

6. Search returns the original log events from the Spider-compressed archive; KQL: `msg: "demo log line*"`:
```
curl -X POST http://test-clp-api-server:3001/query -H 'Content-Type: application/json' \
  -d '{"query_string": "msg: \"demo log line*\"", "datasets": ["default"], "ignore_case": true,
       "max_num_results": 10, "buffer_results_in_mongodb": true}'   # -> {"query_results_uri":"/query_results/1"}
curl http://test-clp-api-server:3001/query_results/1
```
Expected:
```
data: {"ts":1700000001,"level":"INFO","msg":"demo log line 1"}
data: {"ts":1700000002,"level":"INFO","msg":"demo log line 2"}
data: {"ts":1700000003,"level":"INFO","msg":"demo log line 3"}
data: {"ts":1700000004,"level":"INFO","msg":"demo log line 4"}
```

## Summary by CodeRabbit

* **New Features**
  * Added optional Spider storage integration for Helm deployments, including storage host/port helpers.
  * Introduced a configurable compression coordinator component.

* **Bug Fixes**
  * Improved Helm packaging and test install flows by updating chart dependencies immediately before packaging/installation.
  * Enhanced Helm linting to resolve the Spider chart repository during lint checks.

* **Chores**
  * Bumped the Helm chart version, wired Spider chart dependency, and excluded downloaded subcharts from source control.
  * Updated Helm templates/config so the coordinator renders when Spider is enabled, and related compression components are omitted accordingly.


tools/deployment/package-helm/set-up-multi-dedicated-test.sh
```diff
@@ -138,6 +138,7 @@ helm uninstall test --ignore-not-found
 sleep 2
 # Word splitting is intentional: helper functions return multiple --set flags.
 # shellcheck disable=SC2046
+helm dependency update "${script_dir}"
```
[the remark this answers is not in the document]
20001020ycx: its hosted on github, should be fine

tools/deployment/package-helm/templates/compression-coordinator-deployment.yaml
```diff
@@ -0,0 +1,75 @@
...
+        - {{- include "clp.waitFor" (dict
+            "root" .
+            "type" "job"
+            "name" "db-table-creator"
+          ) | nindent 10 }}
+        - {{- include "clp.waitFor" (dict
+            "root" .
+            "type" "service"
+            "name" "storage"
+          ) | nindent 10 }}
```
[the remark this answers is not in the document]
20001020ycx: The Spider subchart's storage pods carry app.kubernetes.io/component: "storage" (upstream chart 0.1.4, storage-deployment.yaml:18), spider-storage is just a DNS name, can you double check?

tools/deployment/package-helm/values.yaml
```diff
@@ -359,3 +376,12 @@ serviceAccount:
...
   #   eks.amazonaws.com/role-arn: "arn:aws:iam::<ACCOUNT_ID>:role/<role-name>"
+
+# Overrides for the Spider subchart's `values.yaml`
+spider:
+  enabled: false
+  fullnameOverride: "spider"
+  image:
+    worker:
+      repository: "ghcr.io/y-scope/clp/clp-spider-worker"
+      tag: "main"
```
junhaoliao, line 394: shall we move this into the `image` object starting at Line 7?

tools/deployment/package-helm/values.yaml
```diff
@@ -359,3 +376,12 @@ serviceAccount:
   annotations:
   # For IRSA on EKS, set:
   #   eks.amazonaws.com/role-arn: "arn:aws:iam::<ACCOUNT_ID>:role/<role-name>"
+
+# Overrides for the Spider subchart's `values.yaml`
+spider:
+  enabled: false
```
hoophalab, line 390: ~~Shall we move this `enabled: false` field under `clpConfig.package.scheduler: "spider" | "celery"` https://github.com/20001020ycx/clp/blob/c201a472e881bab0358c305931770a81896bdc36/tools/deployment/package-helm/values.yaml#L164~~

ignore this.

tools/deployment/package-helm/values.yaml
```diff
@@ -359,3 +376,40 @@ serviceAccount:
...
+  fullnameOverride: "spider"
+  image:
+    worker:
+      repository: "ghcr.io/y-scope/clp/clp-spider-worker"
+      tag: "main"
+  spiderConfig:
```
20001020ycx, line 395: Not sure if we should put these into the default as part of the clp helm or ask user to provide at run time. I think we should put them here to relieve user experience

tools/deployment/package-helm/values.yaml
```diff
@@ -359,3 +376,40 @@ serviceAccount:
...
+          valueFrom:
+            secretKeyRef:
+              name: "{{ include \"clp.fullnameFromRelease\" . }}-database"
+              key: "username"
+        - name: "CLP_DB_PASS"
+          valueFrom:
+            secretKeyRef:
+              name: "{{ include \"clp.fullnameFromRelease\" . }}-database"
+              key: "password"
+      extra_volume_mounts:
```
hoophalab, line 410: aws_config field https://github.com/y-scope/clp/blob/main/tools/deployment/package-helm/values.yaml#L344
isn't supported, we should either add a TODO issue or fix it

tools/deployment/package-helm/values.yaml
hoophalab, line 395: Spider worker doesn't have a serviceAccountName https://github.com/y-scope/clp/blob/main/tools/deployment/package-helm/templates/compression-worker-deployment.yaml#L24. As a result, this won't work on cloud. We should add a TODO in spider helm and expose here later.

tools/deployment/package-helm/values.yaml
```diff
@@ -359,3 +376,40 @@ serviceAccount:
   annotations:
   # For IRSA on EKS, set:
   #   eks.amazonaws.com/role-arn: "arn:aws:iam::<ACCOUNT_ID>:role/<role-name>"
+
+# Overrides for the Spider subchart's `values.yaml`
+spider:
+  enabled: false
+  fullnameOverride: "spider"
```
hoophalab: it's best to remove this, fullname is usually auto derived.

```suggestion

```

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -27,6 +27,19 @@ used as a full name.
...
+
+@return {string} The fully qualified app name (truncated to 63 characters)
+*/}}
+{{- define "clp.fullnameFromRelease" -}}
+{{- if contains "clp" .Release.Name }}
+{{- .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s-clp" .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
```
hoophalab: I guess this will break if the user passes a clp fullnameOverride https://github.com/y-scope/clp/blob/main/tools/deployment/package-helm/values.yaml#L2

It's best to move clp's fullnameOverride under globals in values, modify clp.fullname function, and use clp.fullname here

tools/deployment/package-helm/values.yaml
20001020ycx: This is for spider's worker image override. We cant move to the `image` object.

We can, however, add a nested object `spider.image.worker` after the `image` object. But I feel like its not as organized as this version.

tools/deployment/package-helm/values.yaml
hoophalab: We can maybe eval `{{ include "spider.componentFullname" (dict "root" (index .Subcharts "spider") "component" "storage") }}` in spider's context in `{{- define "clp.spiderStorageHost" -}}` directly

tools/deployment/package-helm/values.yaml
20001020ycx: Reviewer agrees with the approach

tools/deployment/package-helm/templates/_helpers.tpl
20001020ycx: done, making it global does ensure spider sees the naming as well, good catch

tools/deployment/package-helm/values.yaml
20001020ycx: applied the solution, thanks

tools/deployment/package-helm/values.yaml
20001020ycx: Done in y-scope/spider#420

tools/deployment/package-helm/values.yaml
20001020ycx: @coderabbitai, can you add a TODO for this, since this is non-blocking for cloud

hoophalab (dismissed):
LGTM

Validation: the charts are deployed successfully.

20001020ycx:
We have deployed the current state on the helm chart on the staging namespace of YScope Cloud; all pods are functioning, compression and search jobs complete successfully, provided that the credential-fix PR lands (y-scope/clp#2438).

Greatly appreciate for the help from @hoophalab for the Cloud deployment testing.

tools/deployment/package-helm/templates/_helpers.tpl
```diff
@@ -442,6 +444,26 @@ Gets the port for the Presto service.
...
+
+{{/*
+Gets the port for the Spider storage service.
+
+@param {object} . Root template context
+@return {string} The Spider storage port
+*/}}
+{{- define "clp.spiderStoragePort" -}}
+50051
+{{- end }}
```
junhaoliao, line 466: i could be missing the context

shall we read the effective subchart value? if not, what's the reason?

```suggestion
{{- define "clp.spiderStoragePort" -}}
{{- $spider := index .Subcharts "spider" -}}
{{- $spider.Values.spiderConfig.storage.port -}}
{{- end }}
```

tools/deployment/package-helm/values.yaml
```diff
@@ -1,5 +1,6 @@
-nameOverride: ""
-fullnameOverride: ""
+global:
+  fullnameOverride: ""
+  nameOverride: ""
```
junhaoliao, line 3: mind quickly explaining why we have to move those two keys to `global`? do we need this to also override spider's fullnameOverride and nameOverride (but spider's Helm chart would need modifications in the first place?)

tools/deployment/package-helm/set-up-multi-dedicated-test.sh
junhaoliao: the line-specific suppression now applies to `helm dependency update`; ShellCheck reports the intentional substitutions below. keep the new command order while restoring the directive to `helm install`:

```suggestion
helm dependency update "${script_dir}"
# Word splitting is intentional: helper functions return multiple --set flags.
# shellcheck disable=SC2046
```

tools/deployment/package-helm/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Subcharts are downloaded by `helm dependency update`.
+charts/
```
junhaoliao: this also ignores nested `charts` directories; Helm only generates the chart-root path:

```suggestion
/charts/
```

tools/deployment/package-helm/values.yaml
junhaoliao: i was thinking that we could somehow map clp's values.yaml's `image.spider` object into spider's values.yaml, though now i realize that it might not be possible. please ignore this comment then

tools/deployment/package-helm/set-up-multi-shared-test.sh
```diff
@@ -44,6 +44,7 @@ helm uninstall test --ignore-not-found
 sleep 2
 # Word splitting is intentional: helper functions return multiple --set flags.
 # shellcheck disable=SC2046
+helm dependency update "${script_dir}"
```
junhaoliao:
```suggestion
helm dependency update "${script_dir}"
# Word splitting is intentional: helper functions return multiple --set flags.
# shellcheck disable=SC2046
```

tools/deployment/package-helm/set-up-test.sh
```diff
@@ -28,6 +28,7 @@ helm uninstall test --ignore-not-found
 sleep 2
 # Word splitting is intentional: helper functions return multiple --set flags.
 # shellcheck disable=SC2046
+helm dependency update "${script_dir}"
```
junhaoliao:
```suggestion
helm dependency update "${script_dir}"
# Word splitting is intentional: helper functions return multiple --set flags.
# shellcheck disable=SC2046
```

tools/deployment/package-helm/templates/_helpers.tpl
junhaoliao: not an issue today but just in case some day both CLP and Spider will have some services that share the same names:

```suggestion
  "pod", "--selector", "app.kubernetes.io/instance={{ .root.Release.Name }},app.kubernetes.io/component={{ .name }}",
```

junhaoliao (changes requested):
the rest lgtm. everything looks quite clean and the changes are idiomatic

20001020ycx:
I also performed regression test to ensure the old compression path using Celery does not fail. Below steps are performed:

0. Set up kind context
```
CTX="--context kind-clp-test"
```

1. Log file on the node
```
docker exec clp-test-control-plane sh -c \
  'mkdir -p /demo-logs && printf "{\"ts\":1700000011,\"level\":\"INFO\",\"msg\":\"webui fs line 11\"}\n" > /demo-logs/app.jsonl'
```

2. Install with defaults (fs mode) — from tools/deployment/package-helm
```
helm $CTX install test . \
  --set clpConfig.logs_input.directory=/demo-logs \
  --set image.clpPackage.tag=main-amd64 --set image.clpPackage.pullPolicy=IfNotPresent
```

3. Compress via webui (or use the Ingest page: kubectl port-forward svc/test-clp-webui 4000:4000)
```
kubectl $CTX run curl-compress --rm -i --restart=Never --image=curlimages/curl:8.14.1 -- \
  -sS -X POST http://test-clp-webui:4000/api/compress \
  -H 'Content-Type: application/json' -d '{"paths": ["/app.jsonl"], "dataset": "default"}'
-> {"jobId":1}
```

4. Verify: Celery took it (spider_id NULL)
```
DBPASS=$(kubectl $CTX get secret test-clp-database -o jsonpath='{.data.password}' | base64 -d)
kubectl $CTX exec test-clp-database-0 -- mariadb --table -u clp-user -p$DBPASS clp-db \
  -e "SELECT id, status, spider_id, duration FROM compression_jobs;"
# -> id=1, status=2, spider_id=NULL
```

tools/deployment/package-helm/values.yaml
20001020ycx: This is a great question which had haunted me for a while, and I am happy to explain as this is the problem with subchart.

Consider the below config we must set at Spider side such that worker can talk to CLP database (for commit task).
```
- name: "CLP_DB_USER"
     valueFrom:
       secretKeyRef:
       name: "{{ include \"clp.fullname\" . }}-database"
```

However, `clp.fullname` is only recognized by CLP's helm chart. Without declaring it as global, Spider subchart resolves it to the literal string `{{ include \"clp.fullname\" . }}` rather the actual full name of the clp subchart.

junhaoliao: I see

shall we mark this as a breaking change? as in, those who need to override those values should be aware they need to change different keys now

junhaoliao (approved):
lgtm. the title lgtm too but let's consider marking this as a breaking change before merging

PR #2419, opened 2026-07-23
fix(compression-coordinator): Restrict Zstandard detection to RFC-compliant `.zst` suffixes (fixes #2414); match suffixes case-insensitively.

### LINKED ISSUE #2414: Restrict Zstandard extension handling to .zst only
### Bug
Per [RFC 8878 Section 7.1](https://datatracker.ietf.org/doc/html/rfc8878#section-7.1), `.zst` is the designated file extension for Zstandard compressed files.

CLP currently treats additional suffixes as Zstandard (or gzip) in code-paths that estimate uncompressed size:

- `components/compression-coordinator/src/partition.rs` (`estimate_uncompressed_size`) treats `.zstd`, `.zstandard`, `.tar.zstd`, `.tar.zstandard` as ZSTD suffixes.
  - https://github.com/y-scope/clp/blob/d1248bec557f1de0010e369afdb451003ec139b9/components/compression-coordinator/src/partition.rs#L235
- `components/clp-py-utils/clp_py_utils/core.py` (`FileMetadata`) applies the same non-standard extension set when setting `estimated_uncompressed_size`.
  - https://github.com/y-scope/clp/blob/d1248bec557f1de0010e369afdb451003ec139b9/components/clp-py-utils/clp_py_utils/core.py#L23

Given the RFC guidance, these should be narrowed to `.zst` only to avoid handling non-standard extensions as Zstandard.

### CLP version
d1248bec557f1de0010e369afdb451003ec139b9

### Environment
Not environment-specific; issue identified via source inspection.

### Reproduction steps
1. Open the two files linked above in the `main` branch at SHA `d1248bec557f1de0010e369afdb451003ec139b9`.
2. Observe that `.zstd`, `.zstandard`, `.tar.zstd`, `.tar.zstandard` are still included in Zstandard extension detection in both locations.
3. Compare with [RFC 8878 Section 7.1](https://datatracker.ietf.org/doc/html/rfc8878#section-7.1), which indicates `.zst` as the valid extension.
4. Confirm that this causes estimation logic to treat those non-standard extensions as compressed with Zstandard scaling.

### Additional context
- This issue intentionally updates `.zstd`/`.zstandard` handling only, not behavioral changes to true `.zst` compression workflows.

junhaoliao:
# Description

The [Rust compression coordinator](https://github.com/y-scope/clp/blob/c0c86763a6300afe1690e1d64fb311d233ee6cd3/components/compression-coordinator/src/partition.rs#L231-L243) and [Python `FileMetadata`](https://github.com/y-scope/clp/blob/c0c86763a6300afe1690e1d64fb311d233ee6cd3/components/clp-py-utils/clp_py_utils/core.py#L10-L25) estimated Zstandard files using non-standard `.zstd` and `.zstandard` suffixes while overlooking the standard `.zst` suffix.

This change updates both implementations to recognize the standard `.zst` extension case-insensitively and applies the same case-insensitive matching to the existing gzip suffixes. The Rust implementation uses a shared allocation-free suffix helper and does not suppress Clippy checks. Focused regression tests cover `.zst`, `.clp.zst`, `.tar.zst`, uppercase `.ZST`, and lower- and uppercase forms of every supported gzip suffix.

# Checklist

  breaking change.

# Validation performed

* `task lint:fix-py` (passed; the task reported existing non-blocking mypy diagnostics in unrelated Python components).
* `task lint:fix-rust` (passed).
* `task tests:integration:clp-py-project-imports`:

```text
============================== 3 passed in 0.06s ===============================
```

* `task tests:rust-all`:

```text
     Summary [  12.606s] 36 tests run: 36 passed, 0 skipped
```

* `task`:

```text
#33 DONE 7.5s
```

* The package smoke test could not reach compression because `./sbin/start-clp.sh` failed while starting the filesystem-input package:

```text
clp-package-0df21562-8d02-496e-b831-b1d77395659d is missing dependency log-ingestor
2026-07-23T17:59:05.831 ERROR [start_clp] Failed to start CLP.
```

  `./sbin/stop-clp.sh` then removed the partially started stack successfully (`stop_status=0`).

## Summary by CodeRabbit

- **Bug Fixes**
  - Compression size estimates now recognize supported gzip and Zstandard archive extensions regardless of letter casing.
  - Gzip formats including `.gz`, `.gzip`, `.tgz`, and `.tar.gz` receive accurate uncompressed-size estimates.
  - Zstandard estimates now consistently apply to files ending in `.zst`, including nested archive names.

- **Tests**
  - Added coverage for mixed-case extensions and supported gzip and Zstandard formats to help ensure reliable size estimates.


components/compression-coordinator/src/partition.rs
```diff
@@ -384,6 +392,38 @@ mod tests {
...
+
+        for key in [
+            "logs/app.log.zst",
+            "logs/app.log.clp.zst",
+            "logs/app.log.tar.zst",
+            "logs/app.log.ZST",
+        ] {
+            assert_eq!(FILE_SIZE * 8, estimate_uncompressed_size(key, FILE_SIZE));
+        }
+    }
```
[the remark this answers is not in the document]
junhaoliao: > omit the suffixes that must no longer be recognised

i'm not sure if that's a must. seems like adding such tests would also only add maintenance debt rather than actual value. what do you think?

components/compression-coordinator/src/partition.rs
```diff
@@ -238,29 +238,21 @@ impl Iterator for RoundRobinIterator {
 /// # Returns
 ///
 /// The estimated uncompressed size.
-fn estimate_uncompressed_size(key: &str, size: u64) -> u64 {
+fn estimate_uncompressed_size(path: &str, size: u64) -> u64 {
```
LinZhihao-723: I'm not sure if using `path` is a good option here since the path is bucket + key.

components/compression-coordinator/src/partition.rs
```diff
@@ -238,29 +238,21 @@ impl Iterator for RoundRobinIterator {
...
-    const GZIP_SUFFIXES: &[&str] = &[".gz", ".gzip", ".tgz", ".tar.gz"];
+    const GZIP_SUFFIXES: &[&str] = &[".gz", ".gzip", ".tgz"];
     const ZSTD_COMPRESSION_RATIO_ESTIMATE: u64 = 8;
 
-    if GZIP_SUFFIXES
-        .iter()
-        .any(|suffix| ends_with_ignore_ascii_case(key, suffix))
-    {
+    let path = path.to_ascii_lowercase();
+    if GZIP_SUFFIXES.iter().any(|suffix| path.ends_with(suffix)) {
```
LinZhihao-723: Please explain why you need `let path = path.to_ascii_lowercase();`.
Similar questions asked by https://github.com/y-scope/clp/pull/2419/changes/3c66a96dfa0b0dfe091adfdba1248ba2e54f44b3#r3684813671.

LinZhihao-723 (commented):
The issue reported in #2414 is to restrict the suffix for zstdandard to `.zst`. It doesn't mention case sensitivity.
@junhaoliao Why do we need case-insensitive match?

components/clp-py-utils/clp_py_utils/core.py
```diff
@@ -15,13 +15,10 @@ def __init__(self, path: pathlib.Path, size: int):
         self.size = size
         self.estimated_uncompressed_size = size
 
-        filename = path.name
-        if any(filename.endswith(extension) for extension in [".gz", ".gzip", ".tgz", ".tar.gz"]):
+        filename = path.name.lower()
+        if any(filename.endswith(extension) for extension in [".gz", ".gzip", ".tgz"]):
```
junhaoliao, line 19: i think ` ".tar.gz"` is worth keeping - 

https://man7.org/linux/man-pages/man1/tar.1.html#:~:text=%2Da%2C%20%2D%2Dauto%2Dcompress%20Use%20archive%20suffix%20to%20determine%20the%20compression%20program%2E

```
-a, --auto-compress
              Use archive suffix to determine the compression program.
```

the `.gz` suffix is directly appended after `.tar` when using the `tar` tool with compression

junhaoliao: please ignore this comment cuz i think an ending string match is sufficient here - i forgot that we removed the .tar.zst match for the same rationale

Bill-hbrhbr (dismissed):
I think the lowercase feature is useful. It is not tracked in issue 2414 but I suppose we can update the PR title to better reflect what's being fixed and what's being added.

Bill-hbrhbr:
For title, how about:
```
fix(compression-coordinator): Restrict Zstandard detection to RFC-compliant `.zst` suffixes (fixes #2414); match suffixes case-insensitively.
```

components/compression-coordinator/src/partition.rs
```diff
@@ -238,21 +238,35 @@ impl Iterator for RoundRobinIterator {
...
     } else {
         size
     }
 }
 
+/// Checks whether `object_key` ends with `suffix`, ignoring ASCII case.
+///
+/// # Returns
+///
+/// `true` if `object_key` ends with `suffix`, ignoring ASCII case; `false` otherwise.
```
LinZhihao-723:
```suggestion
/// # Returns
///
/// Whether `object_key` ends with `suffix` using ASCII case-insensitive match.
```
- This function is simple. I don't think we need both the description and the return section.
- We should pay attention to the return section for using "Whether ..." instead of "`true` if ..., `false` otherwise". This guideline should always be applied regardless of the language we use.

components/compression-coordinator/src/partition.rs
LinZhihao-723, line 1: You should revert the renaming of the following:
- `path` -> `object_key`
- `key` -> `object_key`

We may need a separate PR to clean this up before we add local filesystem compression input/output support. But before that, I prefer to keep it untouched as they're irrelevant to the changes involved in this PR. We should try to make the diff clean, especially for these cases where renaming doesn't make things clearer.

LinZhihao-723 (changes requested):

components/compression-coordinator/src/partition.rs
Bill-hbrhbr: @coderabbitai check for possible remaining issues.

LinZhihao-723 (approved):
lgtm

PR #2435, opened 2026-07-29
feat(compression-coordinator): Introduce `max_concurrent_tasks` config to limit job-handler concurrency.

Bill-hbrhbr:
# Description

This PR limits the number of compression jobs processed concurrently to prevent the coordinator from creating an unbounded number of job-handler tasks and consuming excessive memory when a large backlog accumulates.

The coordinator now only fetches as many new jobs as it has capacity to process. Once the configured concurrency limit is reached, no additional jobs are fetched until existing jobs finish and capacity becomes available.

On startup, the coordinator recovers work from the previous instance before scheduling new jobs. This includes both jobs that were already submitted to Spider and jobs that were dispatched but did not progress far enough to be marked as running. The former resume tracking their existing Spider jobs, while the latter are re-dispatched.

--------------
**NOTE**

The `is_first_fetch` flag and its two-query fetch logic have been removed because they do not work well with bounded concurrency. Previously, the first fetch after startup returned all pending jobs, including those with an existing `dispatch_time`, to recover jobs that may not have completed dispatch before the previous coordinator exited.

With bounded concurrency, recovery can no longer rely on the first fetch returning everything, since the coordinator should only fetch as many new jobs as it has capacity to process. Recovery is therefore handled explicitly by `recover_previous_jobs`, which identifies **all** pending jobs with an existing `dispatch_time` and re-dispatches them before normal scheduling begins.

This leaves the steady-state dispatcher with a single, clear eligibility rule: only pending jobs without a `dispatch_time` are considered new work.

--------------------------------

The concurrency limit is configured through a new `max_concurrent_tasks` field in `ClpConfig`, with a default of 1000.

-------------------------------

This PR opens https://github.com/y-scope/clp/issues/2472 to be resolved in a future PR.

# Checklist

  breaking change.

# Validation performed

## New Features
* Added a configurable limit for concurrent compression tasks, defaulting to 1,000.
* Scheduling now adapts to available capacity and avoids dispatching work when capacity is unavailable.
* Pending and interrupted compression jobs are recovered and resumed more reliably.

## Bug Fixes
* Invalid concurrency settings, including zero or unsupported values, now produce clear configuration errors.
* Improved startup handling for previously dispatched compression jobs.


components/compression-coordinator/src/coordination.rs
```diff
@@ -239,21 +259,34 @@ impl Coordinator {
...
-        })?;
-        let dispatched_job_ids: Vec<CompressionJobId> =
-            new_job_rows.iter().map(|row| row.id).collect();
-        for job_row in new_job_rows {
+        if self.pending_job_queue.is_empty() && self.job_handler_sem.available_permits() > 0 {
+            let new_job_rows = self.fetch_new_job_rows().await.inspect_err(|e| {
+                tracing::error!(error = % e, "Failed to fetch new jobs from database.");
+            })?;
+            self.pending_job_queue.extend(new_job_rows);
+        }
```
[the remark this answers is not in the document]
Bill-hbrhbr: We currently only fetch new jobs when the pending queue has been exhausted. so it's not like we will accumulate a backlog or get any duplicated pending jobs.
also this PR is about the limit for tokio spawns.
if you want to do paged fetches that's another story.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -795,6 +795,7 @@ class PollingBackoff(BaseModel):
 class CompressionCoordinator(BaseModel):
     resource_group: SpiderResourceGroup = SpiderResourceGroup(name="compression-coordinator")
     job_polling_interval_millisecs: PositiveInt = 100
+    max_concurrent_tasks: PositiveInt = 1000
```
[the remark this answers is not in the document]
Bill-hbrhbr: I updated the rust side. python side doesn't need such strict checking

components/compression-coordinator/src/coordination.rs
```diff
@@ -37,14 +39,17 @@ pub struct Coordinator {
...
 impl Coordinator {
     /// Factory function.
     ///
-    /// On construction, this recovers compression jobs that a previous coordinator instance had
-    /// already submitted to Spider (those still [`CompressionJobStatus::Running`] with a Spider job
-    /// ID) by spawning a detached handle to drive each one to completion.
+    /// On construction, this begins recovering all compression jobs that a previous coordinator
+    /// instance had already submitted to Spider (those still [`CompressionJobStatus::Running`] with
+    /// a Spider job ID). During the restart phase, no concurrency limit is imposed, so all
+    /// recovered jobs are resumed immediately.
```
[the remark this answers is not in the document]
Bill-hbrhbr: check the docstring.

components/package-template/src/etc/clp-config.template.json.yaml
```diff
@@ -64,6 +64,27 @@ telemetry:
 #  port: 6000
 #  logging_level: "INFO"
 #
+## Compression coordinator config. When set, the `spider` config below must also be set.
```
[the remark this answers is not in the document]
Bill-hbrhbr: we will implement fs output in the near future. check docstring for other wording updates.

components/package-template/src/etc/clp-config.template.json.yaml
```diff
@@ -64,6 +64,27 @@ telemetry:
...
+#compression_coordinator:
+#  resource_group:
+#    name: "compression-coordinator"
+#  job_polling_interval_millisecs: 100
+#  max_concurrent_tasks: 1000
```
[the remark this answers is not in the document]
Bill-hbrhbr: check update

components/compression-coordinator/src/coordination.rs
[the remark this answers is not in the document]
Bill-hbrhbr: @LinZhihao-723 do we need bounded job fetching

components/package-template/src/etc/clp-config.template.json.yaml
```diff
@@ -64,6 +64,27 @@ telemetry:
 #  port: 6000
 #  logging_level: "INFO"
 #
+## Compression coordinator config. Requires `logs_input.type` to be "s3" and `spider` to be set.
```
sitaowang1998: We are removing Spider config from this file, as it is only used for docker compose.

components/compression-coordinator/src/coordination.rs
```diff
@@ -239,21 +270,34 @@ impl Coordinator {
...
+                tracing::error!(error = % e, "Failed to fetch new jobs from database.");
+            })?;
+            self.pending_job_queue.extend(new_job_rows);
+        }
+
+        let mut dispatched_job_ids = Vec::new();
+        while !self.pending_job_queue.is_empty() {
+            let Ok(permit) = self.job_handler_sem.clone().try_acquire_owned() else {
+                break;
+            };
```
sitaowang1998: Getting a semaphore in a busy loop is not a good design.

sitaowang1998 (changes requested):

components/compression-coordinator/src/coordination.rs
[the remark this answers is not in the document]
LinZhihao-723: Yeah, I thought it's a part of the plan.

components/package-template/src/etc/clp-config.template.json.yaml
LinZhihao-723, line 1: I don't think we should update this config. The current compression coordinator only supports helm (k8s) deployment which doesn't use this template. We should update it when we have docker compose support ready,

components/compression-coordinator/src/coordination.rs
```diff
@@ -37,14 +39,17 @@ pub struct Coordinator {
     is_first_fetch: bool,
     job_polling_interval: Duration,
     cancellation_token: CancellationToken,
+    job_handler_sem: Arc<Semaphore>,
+    pending_job_queue: VecDeque<PendingJobRowProjection>,
```
LinZhihao-723: I'm not sure if we need this. Can you explain why we can't do the following instead:
* On the main loop, fetch compression jobs with a limit and ID ordering (sth like `SELECT * FROM t ORDER BY id ASC LIMIT 100;`)
* Get a permit before spawning the coroutine. The permit automatially drops itself if the coroutine exits, aborts, or crashed.
* Let the semaphore to block coroutine creation implicitly. The main loop may be blocked, which is fine because it shouldn't push more jobs into Spider.

This should lead to the behavior we expect iiuc.

LinZhihao-723 (changes requested):

components/compression-coordinator/src/coordination.rs
LinZhihao-723: As discussed offline, I think it should be fine: the performance overhead isn't that large though especially in this use case.

Bill-hbrhbr:
@coderabbitai review

components/compression-coordinator/src/coordination.rs
```diff
@@ -123,53 +135,38 @@ impl Coordinator {
...
-    /// token is triggered. The jobs dispatched in the iteration are marked once the sleep elapses,
-    /// so their update does not contend with concurrent job submissions during the poll interval.
+    /// 1. Schedule pending compression jobs up to the available concurrency limit.
+    /// 2. Wait until the next polling interval or until cancellation.
+    /// 3. Mark the scheduled jobs as dispatched.
+    ///
+    /// `dispatch_time` marks jobs that have already been dispatched by the current coordinator,
+    /// preventing them from being dispatched again before their handlers persist the `Running`
+    /// state. These updates are batched and applied after the polling interval to reduce
+    /// contention with the handlers' `Running` state updates.
```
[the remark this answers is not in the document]
Bill-hbrhbr: @LinZhihao-723 @sitaowang1998 This does seem like a real issue, although I think addressing it is outside the scope of this PR.

One possible solution would be to make claiming a job atomic by setting `dispatch_time` as part of the fetch operation. For example, `fetch_new_job_rows` could first claim a bounded set of pending jobs with:

```sql
UPDATE compression_jobs
SET dispatch_time = CURRENT_TIMESTAMP()
WHERE status = 'PENDING' AND dispatch_time IS NULL
ORDER BY id ASC
LIMIT ?
```

Once a row is claimed, it can no longer be picked up as new work, regardless of whether the coordinator is cancelled or crashes between claiming and spawning the handler.

The remaining failure case is when a job is claimed but its handler never starts. In that case, the job remains `Pending` with `dispatch_time` set, so `recover_previous_jobs` can identify and re-dispatch it after restart.

components/compression-coordinator/src/coordination.rs
```diff
@@ -391,106 +456,120 @@ impl Coordinator {
...
+        Ok(rows)
+    }
+
+    /// Deserializes `serialized_config` as a [`ClpIoConfig`].
+    ///
+    /// On failure, logs the error, marks the compression job as [`CompressionJobStatus::Failed`].
+    ///
+    /// # Returns
+    ///
+    /// The deserialized CLP io config.
```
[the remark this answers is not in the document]
Bill-hbrhbr: @coderabbitai check if the issue still persists

components/compression-coordinator/src/coordination.rs
```diff
@@ -229,13 +287,9 @@ impl Coordinator {
...
-    /// Fetches the pending compression jobs and spawns a detached handle to drive each one.
-    ///
-    ///
-    /// A job whose config cannot be deserialized is marked [`CompressionJobStatus::Failed`] and
-    /// skipped; a job whose handle cannot be constructed is skipped as well (and marked
-    /// [`CompressionJobStatus::Failed`] unless its input config is unsupported, in which case it is
-    /// left for the legacy Celery-based compression scheduler).
+    /// Fetches up to the configured concurrency limit of pending jobs and spawns a detached
+    /// handler for each. Jobs with invalid configurations or whose handles cannot be constructed
+    /// are marked with failure and skipped.
```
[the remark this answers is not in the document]
Bill-hbrhbr: this is too detailed for `schedule_new_jobs` as it pertains to the operations inside the helper `create_job_handle`

Bill-hbrhbr:
@coderabbitai review

components/compression-coordinator/src/coordination.rs
```diff
@@ -391,106 +444,119 @@ impl Coordinator {
...
+                );
+                self.mark_job_failed(
+                    job_id,
+                    &format!("Failed to deserialize CLP I/O config: {e}"),
+                )
+                .await;
+                None
+            }
+        }
+    }
```
LinZhihao-723: It's a really bad idea to return an option instead of a result, and log inside the helper. The best practice for designing this type of helper is:
- Return a result. Let the caller decide whether it wants to inspect the error or just propagate.
- Log on the caller side.
  - You may see `create_job_handle` logs inside the helper. That is because it needs to print different logs based on the type of the error and execute different reaction calls accordingly. This makes it an exception since logging it inside would actually make things cleaner. In this helper, errors are printed unconditionally so it should be logged outside.

components/compression-coordinator/src/coordination.rs
```diff
@@ -246,36 +300,42 @@ impl Coordinator {
...
+            let Some(clp_io_config) = self
+                .try_deserialize_clp_io_config(job_id, &job_row.serialized_clp_io_config)
+                .await
+            else {
+                continue;
+            };
+
             tracing::info!(job_id = % job_id, "Scheduling new job.");
-            let Ok(job_handle) = self.create_job_handle(job_id, clp_io_config).await else {
+            let Some(job_handle) = self.create_job_handle(job_id, clp_io_config).await else {
```
LinZhihao-723: Similar to https://github.com/y-scope/clp/pull/2435/changes#r3736535562: you didn't change the implementation of `create_job_handle`, then why would you need to update the return type? This is polluting the diff in some sense and making the PR harder to review.

components/compression-coordinator/src/coordination.rs
```diff
@@ -204,6 +198,70 @@ impl Coordinator {
...
+                continue;
+            };
+
+            let permit = self.job_handler_sem.clone().try_acquire_owned().ok();
+            tokio::spawn(async move {
+                let _permit = permit;
+                let result = match spider_job_id {
+                    Some(id) => job_handle.recover(id).await,
+                    None => job_handle.run().await,
```
LinZhihao-723: I think this is actually a bug from my previous PR: at this stage, we should not handle any job that is already dispatched but doesn't have a spider ID.
The correct thing to do is (which is also what we actually expected), at the beginning of the factory, we should reset dispatch to NULL for all jobs that have a dispatch time set but no spider ID. These jobs should fall into the normal `run` loop instead of being handled in this recovery stage.

components/compression-coordinator/src/coordination.rs
```diff
@@ -391,106 +444,119 @@ impl Coordinator {
...
+    }
 }
 
-/// A projection of the columns read from a [`CompressionJobStatus::Running`] compression job row.
+const COMPRESSION_JOB_TABLE_NAME: &str = "compression_jobs";
+
+/// A projection of the columns read from a compression job row.
 #[derive(Debug, sqlx::FromRow)]
-struct RunningJobRowProjection {
+struct JobRowProjection {
```
LinZhihao-723: It's a bad idea to merge `PendingJobRowProjection` and `RunningJobRowProjection` into one type: the user needs to check whether `spider_job_id` is NULL now. Since we will have https://github.com/y-scope/clp/pull/2435/changes#r3736798212, let roll back to two separate types instead of one.
In Rust, if you really want a type to cover both, a better idea is probably to make an enum containing both types.

components/compression-coordinator/src/coordination.rs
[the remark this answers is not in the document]
LinZhihao-723: As explain https://github.com/y-scope/clp/pull/2435/changes#r3736798212: I think it should be fine if we fix the dispatch time cleaning.
`dispatch_time` is just a marker and it should be reset if `spider_id` is not set yet. It is possible to see cases where dispatch time is not set but it has a spider ID, which is fine and it can be properly handled by the recovery path.

components/compression-coordinator/src/coordination.rs
```diff
@@ -204,6 +198,70 @@ impl Coordinator {
...
+            tracing::info!(
+                job_id = % job_id,
+                spider_job_id = ? spider_job_id,
+                "Recovering a previously submitted job."
+            );
+            let Some(job_handle) = self.create_job_handle(job_id, clp_io_config).await else {
+                continue;
+            };
+
+            let permit = self.job_handler_sem.clone().try_acquire_owned().ok();
```
LinZhihao-723: I'm not sure if we really need to enforce the permit here: the recovery may not necessarily happen from a failure recovery, but a restart with clp config changed. If the new `max_concurrent_jobs` is changed to a smaller number, this recovery path will be blocked for an improper reason: eventually, a number of compression jobs are already submitted to Spider that exceeds the configured concurrency.
My proposal:
- For now, it's probably easier if we don't do anything to the permit: let the already-submitted jobs run in background. As we would assume the recovery is a rare case, the number of jobs should be roughly bounded by `old_max_concurrent_jobs` + `new_max_concurrent_jobs`. There are extreme cases where this bound can be broken by a series of frequent restarts, but it should be safe to ignore this case for now.
- To improve the naive implementation, we could compare the number of jobs to recover vs. the currently configured `max_concurrent_jobs`:
  - If `num_jobs_to_recover` > `max_concurrent_jobs`: print a warning, make a join of all the recovered job handlers, so the main loop only starts when all previously submitted jobs are finished.
  - Otherwise, do the current implementation to recover jobs while each job holds an acquired permit.

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -761,6 +761,7 @@ class PollingBackoff(BaseModel):
...
```
LinZhihao-723: The concurrency control is in the job level not the task level. Shouldn't it be called `max_concurrent_jobs`?

LinZhihao-723 (changes requested):
There's a risk that we won't be able to merge this PR before the release since it's not in an expected shape yet. Let's fix the recovery path bug first.

components/compression-coordinator/src/coordination.rs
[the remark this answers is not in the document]
LinZhihao-723: @coderabbitai Can you create an issue based on your investigation and my reply on  https://github.com/y-scope/clp/pull/2435/changes#r3736798212?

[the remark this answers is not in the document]
Bill-hbrhbr: @coderabbit The issue is closed right now because we deemed that the architecture would not cause the problem you described. The main reason is that spawned job handlers' lifetime cannot out live that of the coordinator itself.

components/compression-coordinator/src/coordination.rs
```diff
@@ -245,14 +257,33 @@ impl Coordinator {
...
         let dispatched_job_ids: Vec<CompressionJobId> =
             new_job_rows.iter().map(|row| row.id).collect();
         for job_row in new_job_rows {
+            let permit = self
+                .job_handler_sem
+                .clone()
+                .acquire_owned()
+                .await
+                .map_err(|e| {
+                    Error::Semaphore(format!("Failed to acquire a job handler permit: {e}"))
```
LinZhihao-723:
```suggestion
                    Error::Semaphore(format!("failed to acquire a job handler permit: {e}"))
```
This message is formatted as a part of the error. We don't capitalize error messages in such cases. Check https://github.com/LinZhihao-723/claude-instruction/blob/main/rust/coding-style.md#error-messages.

components/compression-coordinator/src/coordination.rs
```diff
@@ -245,14 +257,33 @@ impl Coordinator {
...
+                })?;
```
LinZhihao-723: Since this is mainly for spider job submission rate control, we could do it after job IO dispatching. Let's move this after job handle creation.

components/compression-coordinator/src/coordination.rs
```diff
@@ -392,27 +425,39 @@ impl Coordinator {
     /// Returns an error if:
     ///
     /// * Forwards [`sqlx::query::QueryAs::fetch_all`]'s return values on failure.
-    async fn fetch_new_job_rows(&mut self) -> Result<Vec<PendingJobRowProjection>, Error> {
+    async fn fetch_new_job_rows(
+        &mut self,
+        limit: usize,
+    ) -> Result<Vec<PendingJobRowProjection>, Error> {
```
LinZhihao-723: This API design is a bit error-prone: `limit` is actually not used in the first fetch, which is not documented in the docstring.
Took a closer look, it makes more sense to move `self.job_handler_sem.available_permits` inside this method and drop `limit` parameter. In this way, this method is self-contained where it's behavior is easier to document (we might need to rename it to `fetch_new_job_rows_with_concurrency_limit`).

components/compression-coordinator/src/coordination.rs
```diff
@@ -392,27 +425,39 @@ impl Coordinator {
...
-            .await?;
+
+        let mut query_builder =
+            sqlx::query_as::<_, PendingJobRowProjection>(query).bind(CompressionJobStatus::Pending);
+        if !is_first_fetch {
+            query_builder = query_builder.bind(
+                i64::try_from(limit)
+                    .expect("limit is bounded by Semaphore::MAX_PERMITS, which fits in i64"),
+            );
+        }
```
LinZhihao-723:
```rust
        let query = if self.is_first_fetch {
            self.is_first_fetch = false;
            sqlx::query_as::<_, PendingJobRowProjection>(FIRST_FETCH_QUERY)
                .bind(CompressionJobStatus::Pending)
        } else {
            sqlx::query_as::<_, PendingJobRowProjection>(SUBSEQUENT_FETCH_QUERY)
                .bind(CompressionJobStatus::Pending)
                .bind(
                    i64::try_from(limit)
                        .expect("limit is bounded by Semaphore::MAX_PERMITS, which fits in i64"),
                )
        };
```
Rewrite in a cleaner way.
- Avoid making a variable mutable in Rust whenever possible.

LinZhihao-723 (changes requested):
Overall it's in a good shape now. Some more comments.

components/compression-coordinator/src/coordination.rs
LinZhihao-723: @coderabbitai we decide to stick to the first option for this PR. Can you create an issue to keep track of it, documenting that on restart, recovered jobs may disobey the configured rate limit? You should also document my section option as the proper way to implement it.

LinZhihao-723 (changes requested):
Please add a comment in the recover to mention that we don't bound recovered jobs, referring to this issue: https://github.com/y-scope/clp/issues/2472

Bill-hbrhbr:
@coderabbitai review

components/compression-coordinator/src/coordination.rs
```diff
@@ -399,20 +433,24 @@ impl Coordinator {
...
+                    i64::try_from(self.job_handler_sem.available_permits())
+                        .expect("limit is bounded by Semaphore::MAX_PERMITS, which fits in i64"),
+                )
         };
-        let rows = sqlx::query_as::<_, PendingJobRowProjection>(query)
-            .bind(CompressionJobStatus::Pending)
-            .fetch_all(&self.db_pool)
-            .await?;
+
+        let rows = query.fetch_all(&self.db_pool).await?;
```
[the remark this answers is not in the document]
Bill-hbrhbr: @LinZhihao-723 last question. initial fetch should we include both dispatch_time != NULL and == NULL, or just dispatch_time != NULL, which is guaranteed to be bounded by the previous coordinator's concurrency limit?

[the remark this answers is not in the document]
LinZhihao-723: I think you're right: we should only fetch `dispatch_time != NULL`, which is bounded by the previous concurrency limit.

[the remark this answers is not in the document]
Bill-hbrhbr: @coderabbitai review the current code to see if it aligns with discussion result:

```
we should only fetch dispatch_time != NULL, which is bounded by the previous concurrency limit.

```

components/compression-coordinator/src/coordination.rs
```diff
@@ -147,6 +157,8 @@ impl Coordinator {
             job_handler_sem: Arc::new(Semaphore::new(max_concurrent_jobs)),
         };
 
+        // NOTE: The current implementation does not enforce concurrency limits for recovered jobs
+        // since they were already submitted to Spider. See #2472.
```
LinZhihao-723, line 161: Downgrade this to a code-level comment: this should be more like a TODO instead of a formal behavior to documenet in the method-level docstring.

components/compression-coordinator/src/coordination.rs
```diff
@@ -415,9 +427,9 @@ impl Coordinator {
     /// `dispatch_time` is set, so that jobs dispatched but not started by the previous coordinator
     /// instance can be re-dispatched. No explicit limit is imposed because:
     ///
-    /// 1. This query runs only once, so limiting it could leave previously dispatched jobs
-    ///    unfetched.
-    /// 2. The recovery set is bounded by the previous coordinator's concurrency limit.
+    /// * This query runs only once, so limiting it could leave previously dispatched jobs
+    ///   unfetched.
+    /// * The recovery set is bounded by the previous coordinator's concurrency limit.
```
LinZhihao-723, line 432: We should use bullet point since they are unordered.

components/compression-coordinator/src/coordination.rs
```diff
@@ -1,5 +1,20 @@
...
+//! |----------|-------------|-----------------|--------------------------------------------------|
+//! | PENDING  | NULL        | NULL            | New jobs awaiting dispatch.                      |
+//! | PENDING  | NULL        | NOT NULL        | Jobs dispatched but not yet submitted to Spider. |
+//! | RUNNING  | NOT NULL    | NOT NULL        | Jobs submitted to Spider.                        |
+//!
+//! NOTE:
+//!
+//! * These are the only legal states for a job that hasn't terminated.
+//! * A non-NULL `dispatch_time` indicates that the coordinator has picked up the job and granted it
+//!   permission to run under the concurrency limit.
```
LinZhihao-723, line 17: Adding a section to show all legal states.

LinZhihao-723 (approved):
Approved with some final updates.
For the PR title, how about:
```
feat(compression-coordinator): Add configurable limit for concurrent compression jobs handled by the coordinator.
```

sitaowang1998 (approved):

PR #2444, opened 2026-08-01
feat(search-coordinator): Add initial scaffolding.

Bill-hbrhbr:
# Description

# Checklist

  breaking change.

# Validation performed


components/clp-py-utils/clp_py_utils/initialize-orchestration-db.py
LinZhihao-723, line 1: I'm not sure whether we need all of the schema changes in this file. In the formal PR, we should document why we need to add a new column.

components/compression-coordinator/src/coordination.rs
```diff
@@ -229,7 +229,7 @@ impl Coordinator {
             self.mark_jobs_dispatched(&dispatched_job_ids).await?;
         }
 
-        tracing::info!("Coordinator shutting down.");
+        tracing::info!("Compression coordinator shutting down.");
```
LinZhihao-723, line 232: Why?

LinZhihao-723 (changes requested):
As discussed offline, this PR is too large and we should break it into small ones to tight the scope for better decision making. I would propose the following:
- `search_job_submitter` with only the trait. The implementation depends on TDL. We can implement it for Spider client, but leave it with `todo!`.
- `job_handle` implemention on top of `search_job_submitter`.
  - We probably need to split this mod even further (splitting into smaller PRs). Use `todo!` to define the interface first.
- `coordination` implemention on top of `job_handle`.
- `bin` implementation.

Assume `tdl` becomes available in parallel, the binary itself should be able to serve search jobs. We then plan to add the Python side and deployment integration.

PR #2449, opened 2026-08-01
fix(ci): Run LocalStack-backed Rust tests on self-hosted runners.

jackluo923:
# Description

`tests:rust-all` starts LocalStack with `--publish <port>:4566` and connects to it over loopback. Our self-hosted runners mount the host's Docker socket, so that container is a sibling on the host: the port lands in the host's network namespace rather than the runner's, and the job can't reach it. This is why the job was pinned to `GH_RUNNER_LINUX_LIGHT_X64` (`["ubuntu-24.04"]`, GitHub-hosted).

`start.py` now checks whether it's running inside a container the Docker daemon can resolve. If so, it starts LocalStack with `--network container:<id>` and `GATEWAY_LISTEN=0.0.0.0:<port>` instead of publishing a port, so LocalStack shares the job's network namespace and loopback reaches it. No extra privileges are needed, and because each runner has its own namespace, concurrent runners on one machine can't contend over ports.

GitHub-hosted runners and dev machines fail the check and keep today's publishing behaviour, so `AWS_ENDPOINT_URL` and the steps in `components/log-ingestor/README.md` are unchanged.

The job then moves to `GH_RUNNER_LINUX_HEAVY_X64` (self-hosted), which suits it anyway since `tests:rust-all` builds the workspace in release mode.

Two smaller changes:

* Address LocalStack as `127.0.0.1` rather than `localhost`. Inside a container `localhost` can resolve to `::1` alone, which LocalStack's IPv4 gateway doesn't bind, and it fails identically to the bug above. `tests/aws_config.rs` already used `127.0.0.1`.
* `start.py` waits for `/_localstack/health` before returning. Nothing waited for readiness before; boto3 retries happened to cover it.

# Checklist

  breaking change.

# Validation performed

Simulated a runner with the host Docker socket mounted, matching the `y-scope/ci` setup:

* LocalStack was reachable over loopback, with no port published on the host.
* Two such runners ran LocalStack on the same port without interfering.
* The runner kept DNS and egress with LocalStack in its namespace.

Also re-ran publishing mode on the host end to end through `create-bucket.py`. `ruff check`, `ruff format`, and `yamllint --strict` are clean.

### DIFF
--- a/.github/workflows/clp-rust-checks.yaml
+++ b/.github/workflows/clp-rust-checks.yaml
@@ -28,10 +28,8 @@ concurrency:
 
 jobs:
   rust-checks:
-    # Stays on GH_RUNNER_LINUX_LIGHT_X64: tests:rust-all uses LocalStack on
-    # 127.0.0.1, broken on self-hosted.
     runs-on: >-
-      ${{ fromJSON(vars.GH_RUNNER_LINUX_LIGHT_X64 || '["ubuntu-24.04"]') }}
+      ${{ fromJSON(vars.GH_RUNNER_LINUX_HEAVY_X64 || '["ubuntu-24.04"]') }}
     steps:
       - uses: "actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd"  # v6.0.2
         with:
--- a/taskfiles/tests/main.yaml
+++ b/taskfiles/tests/main.yaml
@@ -31,7 +31,7 @@ tasks:
         --port "{{.LOCALSTACK_PORT}}"
       - |-
         . "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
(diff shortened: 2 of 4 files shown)


jackluo923:
Follow-ups noted while working on this, not addressed here:

* A hard-cancelled job leaks a sibling LocalStack container on the host, since the `defer:` stop never runs. A `--label` plus a periodic prune on the runner machines would cover it.
* `get-free-port.py` closes its probe socket before LocalStack binds, leaving a narrow TOCTOU window. Retrying `start.py` on bind failure would close it.
* `docs/src/dev-docs/tooling-gh-workflows.md:99-110` still documents the owner-based runner selection and `["self-hosted", "x64", "ubuntu-noble"]` tags removed in d12530ee3.

Issue #2440, opened 2026-07-31
Remove legacy Spider base deployment setup


LinZhihao-723:
### Request

We've deprecated the legacy Spider integration (Wolf) from CLP's user/dev doc. However, the development script is not removed yet. These should be removed before the next release.

### Possible implementation

The following list should be removed:
- Deployment yaml for Spider docker compose integration.
- Legacy Spider's option in clp-config.yaml.
- Legacy Spider's option in CLP package controller.
- Spider build flow and package installation.

LinZhihao-723:
#2452 removed legacy Spider's Docker Compose deployment, its `clp-config.yaml` options, and its wiring in the CLP package controller — the first three bullets of #2440. The fourth bullet, **"Spider build flow and package installation"**, was deliberately deferred; this issue tracks it, plus the code that became unreachable once #2452 deleted its callers.

After #2452, legacy Spider is still downloaded, compiled, and shipped in every package image — it simply can no longer be launched. Concretely:

- `task package-build-deps` still builds the pinned `y-scope/spider@32711f9` C++ tarball on every package build.
- The `clp-package` image still ships `spider_scheduler` (2,116,648 B) and `spider_worker` (2,417,256 B). Nothing can start them: `spider_scheduler` was launched only by the deleted `spider-scheduler` Compose service, and `spider_worker` only via `job_orchestration.executor.start-spider-worker`, whose Compose service was also deleted.
- `mariadb-connector-cpp` is built and `libmariadbcpp.so` (2,329,080 B) is copied into the image solely for those two binaries, per #1414.
- `deps:spider` carries `platforms: ["linux"]` plus an Ubuntu-only precondition, so `task package` still hard-fails on non-Ubuntu Linux — purely for dead code.
- Three legacy Spider Python modules and the `yscope-spider-py` runtime dependency remain with zero callers.

None of this breaks the build today; it is pure dead weight, an unnecessary platform restriction, and a live supply-chain dependency on a frozen legacy commit.

> [!Note]
> None of the below refers to the **new** Rust Spider. `tools/docker-images/clp-spider-worker/`, `taskfiles/docker-images.yaml`'s `spider-worker` target, `taskfiles/lint.yaml`'s `--chart-repos="spider=..."`, the Helm `spider` subchart, `components/clp-tdl-package`, and `components/compression-coordinator` must all stay.

PR #2452, opened 2026-08-02
feat(spider)!: Remove legacy Spider support from Docker Compose deployment.

sitaowang1998:
# Description

This PR removes the user-facing configuration and docker compose deployment for the legacy Spider orchestration, as tracked by #2440:

- Remove legacy Spider services and configuration from Docker Compose.
- Make Celery the only compression orchestration option.
- Remove the legacy Spider scheduler and database configuration.
- Remove the obsolete Spider database initialization script.
- Update Python and Rust configuration models accordingly.
- Update package and Helm configuration templates.

> [!Note]
> Note to reviewer: Make sure the changes do not conflict with #2418, which adds helm support for Spider huntsman.

# Checklist

  breaking change.

# Validation performed

- [ ] GitHub workflows pass.

## Summary by CodeRabbit

* **Removed Features**
  * Removed Spider orchestration, scheduling, worker services, database credentials, and related deployment configuration.
  * Removed the Spider-specific Docker Compose configuration.
  * Docker Compose deployments now use the standard configuration.

* **Configuration**
  * Removed Spider database and scheduler settings from configuration templates, Helm values, and generated credentials.
  * Compression scheduling now uses Celery exclusively.

* **Maintenance**
  * Updated the Helm chart version.


components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -179,21 +168,18 @@ class ClpDbUserType(KebabCaseStrEnum):
 
     CLP = auto()
     ROOT = auto()
-    SPIDER = auto()
 
 
 class ClpDbNameType(KebabCaseStrEnum):
```
LinZhihao-723, line 173: Rust mirrors this config in `clp-rust-utils` which is not updated: https://github.com/y-scope/clp/blob/8710828181a418ddaa4303f29221e88a9ec8da24/components/clp-rust-utils/src/clp_config/package/config.rs#L119

LinZhihao-723 (changes requested):
TODO and future work are listed as following:

1. In the user doc [here](https://github.com/y-scope/clp/blob/main/docs/src/user-docs/guides-docker-compose-deployment.md#starting-clp), we still have `(if using Celery)` labels which should be removed (since Celery is the only option now).
2. When cross-referencing PRs for Spider Wolf integration, the coding agent found a bug that was missed from the very beginning in [this](https://github.com/y-scope/clp/blob/8710828181a418ddaa4303f29221e88a9ec8da24/components/clp-py-utils/clp_py_utils/sql_adapter.py#L142) line: `self.database_config.name` will be an unresolved key; it has been renamed to `names`. I'm ok for either fixing it in this PR or creating a separate PR for it.
3. The current `clp_config.py` explicitly rejects `database.names.spider`, but it will not complain if `compression_scheduler.type` or `spider_scheduler` is given as they will be silently ignored. Consider we probably don't have real users for the old Spider scheduler, I'm ok with this behavior. Just need a sanity check with @junhaoliao.
    - The proper fix for this is to add `@model_validator(mode="before")` validator to reject any removed keys. But I'm not sure if we ever need to do it.
4. This PR doesn't remove Spider Wolf's build and installation flow. I've added a followup comment in https://github.com/y-scope/clp/issues/2440#issuecomment-5199506152 to keep track of this. I think it's worth to cut Spider from our build flow for this release, while we can keep the Python side implementation untouched since we may want to remove `compression-scheduler` all at once.

components/clp-py-utils/clp_py_utils/clp_config.py
sitaowang1998: The rust side has already removed the `spider` option.

sitaowang1998:
> TODO and future work are listed as following:
> 
> 1. In the user doc [here](https://github.com/y-scope/clp/blob/main/docs/src/user-docs/guides-docker-compose-deployment.md#starting-clp), we still have `(if using Celery)` labels which should be removed (since Celery is the only option now).
> 
> 2. When cross-referencing PRs for Spider Wolf integration, the coding agent found a bug that was missed from the very beginning in [this](https://github.com/y-scope/clp/blob/8710828181a418ddaa4303f29221e88a9ec8da24/components/clp-py-utils/clp_py_utils/sql_adapter.py#L142) line: `self.database_config.name` will be an unresolved key; it has been renamed to `names`. I'm ok for either fixing it in this PR or creating a separate PR for it.

1 and 2 fixed.

LinZhihao-723 (approved):
- Directly modified the PR title. Notice that it should be "Docker Compose" not "Docker compose".
- Make sure CI passes before you merge.

PR #2453, opened 2026-08-02
fix(build): fix task core on Ubuntu 26.04

4ertus2:
# Description

Fix build errorors related to cmake version < 3.5 suppport and more strict g++ headers includes.
yaml-cpp-0.7.0  has cpp errors for `uint16_t` and `uint32_t` undefined.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

- **Documentation**
  - Updated developer documentation to reflect the current YAML processing library version.

- **Chores**
  - Improved build configuration for more consistent compatibility across supported CMake environments.
  - Disabled unnecessary third-party C++ test builds, helping streamline build setup and execution.

### DIFF
--- a/docs/src/dev-docs/components-core/index.md
+++ b/docs/src/dev-docs/components-core/index.md
@@ -57,7 +57,7 @@ The task will download, build, and install (within the build directory) the foll
 | [SQLite3](https://www.sqlite.org/download.html)                          | v3.36.0        |
 | [utfcpp](https://github.com/nemtrif/utfcpp)                              | v4.0.6         |
 | [xxHash](https://github.com/Cyan4973/xxHash)                             | v0.8.3         |
-| [yaml-cpp](https://github.com/jbeder/yaml-cpp)                           | v0.7.0         |
+| [yaml-cpp](https://github.com/jbeder/yaml-cpp)                           | v0.9.0         |
 | [ystdlib-cpp](https://github.com/y-scope/ystdlib-cpp)                    | 9ed78cd        |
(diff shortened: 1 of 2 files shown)


jackluo923:
Thanks @4ertus2 — good timing on this. CMake v4 support has been open since #795 (April 2025), but it sat below our priority line and, more to the point, the upstream pieces weren't there yet: the first attempt (#794) had to point yaml-cpp at an unreleased commit, and the fixes we depend on for mongo-cxx-driver and mariadb-connector-cpp were only released this July — r4.4.1 about two weeks ago. Now that they exist the whole thing is doable, and #2454 is that effort — a superset of this PR.

Your two fixes are carried into it: the yaml-cpp v0.9.0 upgrade (same tarball and checksum) and `-DANTLR_BUILD_CPP_TESTS=OFF`, moved into the block's alphabetical order. GCC 15 resolves there the same way it does here — a side effect of that same yaml-cpp upgrade rather than a separate fix, since v0.9.0 both declares a v4-compatible minimum and picks up the missing `<cstdint>` include. I dropped `-DCMAKE_POLICY_VERSION_MINIMUM=3.5`, which becomes a no-op once yaml-cpp is on 0.9.0 and declares `cmake_minimum_required(VERSION 3.5...3.30)`.

The extra scope is the remaining v4 blockers: libarchive 3.5.1, the mongo-c-driver bundled inside mongo-cxx-driver (which only fails from v4.4), and mariadb-connector-cpp on the Spider path — plus removing the `<4` pin in `install-cmake.sh`, without which none of the rest takes effect.

#2454 closes #795. Mind closing this one in its favour?

PR #2456, opened 2026-08-04
docs(compression-coordinator): Add a user guide for using `compression-coordinator`.

LinZhihao-723:
# Description

CLP's user docs describe how to deploy Spider (see #2457), but there's no user-facing page for `compression-coordinator` itself. As a result, a user who enables Spider has no documentation explaining what `compression-coordinator` does, why they'd want the Spider-based compression path over the existing Celery-based `compression-scheduler`, which configuration knobs affect it, or what it can't do yet. The Spider guide even had to refer to `compression-coordinator`'s limitations as an unlinked, plain-text mention because the target page didn't exist.

This PR adds that page and wires it into the docs:

## New guide: `guides-using-compression-coordinator.md`

* Introduces `compression-coordinator` as the component that coordinates compression jobs scheduled via Spider, and states that it's intended to replace the Celery-based `compression-scheduler`.
* Notes that the two are meant to run side-by-side until feature parity is reached, with `compression-scheduler` handling any jobs `compression-coordinator` doesn't yet support.
* Points users to the Spider scheduler guide for enabling the Spider scheduling framework, and notes that Kubernetes is currently the only supported deployment method.
* Compares the `compression-coordinator`-based (Spider) architecture against the `compression-scheduler`-based (Celery) architecture across the improvements it provides.
* Attaches the relevant configuration knob to each improvement it applies to, so the guide doubles as a pointer into the settings that govern this behaviour.
* Adds a **Limitations** section recording `compression-coordinator`'s current limitations.

## Wiring

* `guides-using-spider-scheduler.md`: replaces the plain-text `See "Using compression-coordinator"` reference with a real link to the new guide's limitations section.
* `user-docs/index.md`: adds a grid card for the new guide and registers it in the toctree.

# Checklist

  breaking change.

# Validation performed

* [x] Manually checked the rendered doc page.

## Documentation

- Added a guide for using the compression coordinator, including deployment prerequisites, benefits, and current limitations.
- Added the compression coordinator to the Input & storage documentation navigation.
- Updated references to compression coordinator limitations in the Spider scheduler guide.


docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+  two levels of concurrency control to improve fairness among compression jobs running concurrently:
+  * Coordinator-side rate limit: TODO, depends on #2435.
+  * Spider-side active job limit: Controls the maximum number of jobs that can make progress
+    concurrently while sharing compute resources in Spider. Jobs are admitted on a first-come,
+    first-served basis.
+* **Data integrity**: Compression jobs coordinated by `compression-coordinator` publish their
+  results to the rest of the system through a dedicated commit stage. The commit operation is both
+  transactional and idempotent.
+  * A job-level failure does not result in partial updates.
+  * Internal retries do not result in duplicate updates
```
LinZhihao-723: @junhaoliao Can you review this section to check if it makes sense (since it compares to the existing scheduling)?

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
+# Using `compression-coordinator`
+
+`compression-coordinator` is a new component that coordinates compression jobs on [Spider][spider].
+It is designed to replace the existing `compression-scheduler`, which schedules compression jobs
+using Celery.
```
kirkrodrigues:
```suggestion
`compression-coordinator` is a new component that coordinates compression jobs scheduled via
[Spider][spider]. It's designed to replace the existing `compression-scheduler`, which schedules
compression jobs using Celery.
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+
+To use `compression-coordinator`, enable the Spider scheduling framework when starting the CLP
+package.
+
+TODO: Link to Chenxing's doc.
+
+:::{note}
+Currently, `compression-coordinator` can be deployed only on Kubernetes using the CLP Helm chart.
```
kirkrodrigues:
```suggestion
Currently, `compression-coordinator` can only be deployed on Kubernetes using the CLP Helm chart.
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+Support for Docker Compose deployments is planned for a future release.
+:::
+
+Compared with `compression-scheduler`, `compression-coordinator` provides the following improvements
+to the user experience, reliability, and performance:
```
kirkrodrigues:
```suggestion
Compared with `compression-scheduler`, `compression-coordinator` provides the following
user-experience, reliability, and performance improvements:
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+
+* **Automatic failure recovery**: If a service fails or restarts, `compression-coordinator`
```
kirkrodrigues: Before we edit---

1. What service?

2. > fails or restarts
    
    Does that imply a service could fail but not restart and `compression-coordinator` would still resume the job?

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+  automatically resumes previously submitted jobs. Users do not need to restart them manually.
```
kirkrodrigues:
```suggestion
  automatically resumes previously submitted jobs (users don't need to restart them manually).
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+* **Improved resource utilization**: Compression jobs run in a Spider-managed execution environment,
```
kirkrodrigues: 1. Maybe swap this and the next point so the failure-recovery points are grouped together?
2. Before we edit, what's a Spider-managed execution environment and why does that allow resources to be shared more effectively?

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+  allowing computational resources to be shared more effectively across all configured workers.
+* **Configurable retries for compression failures**: `compression-coordinator` allows users to
```
kirkrodrigues: Before we edit, what's configurable specifically?

kirkrodrigues:
```suggestion
* **Easier failure handling via a configurable retry policy**: `compression-coordinator` allows
  users to configure a simple retry policy for failed compression tasks. Automatic retries can help
  recover from transient issues, such as temporary network interruptions, without requiring user
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+  configure a simple retry policy for failed compression tasks. Automatic retries can help recover
+  from transient issues, such as temporary network interruptions, without requiring user
+  intervention.
+  * TODO: Configure retry policy through <LINK>
+* **Improved fairness across concurrent compression jobs**: `compression-coordinator` provides
+  two levels of concurrency control to improve fairness among compression jobs running concurrently:
```
kirkrodrigues:
```suggestion
* **Improved fairness between concurrent compression jobs**: `compression-coordinator` provides
  two levels of concurrency control to improve fairness among compression jobs that run
  concurrently:
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+  * Coordinator-side rate limit: TODO, depends on #2435.
```
kirkrodrigues: Before we edit, how does this improve fairness?

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+  * Spider-side active job limit: Controls the maximum number of jobs that can make progress
```
kirkrodrigues: Before we edit, how does this improve fairness compared to the current policy?

docs/src/user-docs/guides-using-compression-coordinator.md
kirkrodrigues:
```suggestion
* **Improved all-or-nothing semantics**: Compression jobs coordinated by `compression-coordinator`
  publish their results all at once via a dedicated commit operation. The commit operation is both
  transactional and idempotent, so:
  * a job-level failure doesn't result in partial updates.
  * internal task retries don't result in duplicate updates.
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+    concurrently while sharing compute resources in Spider. Jobs are admitted on a first-come,
+    first-served basis.
+* **Data integrity**: Compression jobs coordinated by `compression-coordinator` publish their
+  results to the rest of the system through a dedicated commit stage. The commit operation is both
+  transactional and idempotent.
+  * A job-level failure does not result in partial updates.
```
kirkrodrigues: Although, the indexer table still gets updated, right? Maybe we want to warn about that so that no one gets surprised?

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+  * Internal retries do not result in duplicate updates
+
+:::{note}
+In this release, `compression-coordinator` has the following functional limitations:
```
kirkrodrigues:
```suggestion
`compression-coordinator` currently has the following functional limitations:
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+
+* It supports only the `clp-json` package.
+* It handles only compression jobs created by `log-ingestor`.
+* It supports only S3 archive output.
```
kirkrodrigues:
```suggestion
* It's only available when using `clp-json`.
* It only handles compression jobs created by `log-ingestor`.
* It only supports writing archives to S3.
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+
+We are actively expanding the functionality of `compression-coordinator`. Additional capabilities
+will be introduced in future releases as it moves toward feature parity with
```
kirkrodrigues:
```suggestion
Additional capabilities will be introduced in future releases as it moves toward feature parity with
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,60 @@
...
+`compression-scheduler`.
+:::
+
+:::{note}
+`compression-scheduler` is planned for deprecation and will eventually be fully replaced by
+`compression-coordinator`. In this release, however, `compression-scheduler` continues to run
+alongside `compression-coordinator` to handle compression jobs that `compression-coordinator` does
+not yet support.
```
kirkrodrigues:
```suggestion
`compression-coordinator`. Currently, `compression-scheduler` continues to run alongside
`compression-coordinator` to handle compression jobs that `compression-coordinator` doesn't yet
support.
```

docs/src/user-docs/guides-using-compression-coordinator.md
LinZhihao-723: As discussed offline,  "only on Kubernetes" is more precise since it constrains "only"'s scope to "Kubernetes", instead of both "Kubernetes" and "deployed".

docs/src/user-docs/guides-using-compression-coordinator.md
LinZhihao-723: - For 1: My bad. The service should only refer to `compression-coordinator`.
- For 2: If the service is `compression-coordinator` in specific, this sentence only makes sense by assuming `compression-coordinator` restarts.
- Shall we restructure this sentence to indicate that Spider's service failure is automatically handled as well? But I'm not sure if we need to assume a Spider service fails infinitely without getting recovered.

docs/src/user-docs/guides-using-compression-coordinator.md
LinZhihao-723: 1. Sure.
2. Spider schedules and runs jobs back to back without batching. In the current Celery-based implementation, a straggler inside the batch would cause other available workers to be idle until the next batch is dispatched.

docs/src/user-docs/guides-using-compression-coordinator.md
LinZhihao-723: Was planning to refer to #2457: `compression-coordinator` has a config option to set the number of retries permitted to each compression task. This is simply a counter: no conditional retry is supported.

docs/src/user-docs/guides-using-compression-coordinator.md
LinZhihao-723: That is correct. This is mainly talking about archive metadata and omitting indexer metadata intentionally. Indexer metadata consistency is already misleading in some sense:
- In the current implementation, a compression task that fails to update the indexer table is still considered successful.
- When an archive is removed by retention-based GC, the indexer table is not updated accordingly.

This means the indexer table can:
- Be incomplete about all possible schemas of existing archives.
- Contain schemas that don't exist in any of the archives.

`compression-coordinator` doesn't solve any of these. And to document this behavior, it probably needs a separate section... not sure if we want to cover it here.

Back to this bullet point, how about we be explicit that this guarantee is only for archive metadata?

docs/src/user-docs/guides-using-compression-coordinator.md
LinZhihao-723: Will answer https://github.com/y-scope/clp/pull/2456#discussion_r3752004089 together in this response.

In my understanding of the current `compression-scheduler`, the fairness is only guaranteed by `max_concurrent_tasks_per_job`:

- If `max_concurrent_tasks_per_job` is set to **N**, concurrent jobs are interleaving with a granularity of **N** since up to **N** tasks can be submitted in a batch. But these **N** tasks are executed sequentially without further interleaving.
- If `max_concurrent_tasks_per_job` is not set, all tasks are submitted in a batch. This means job B's first task can't start before job A finishes.

Consider the following extreme case: 4 compression jobs arrived with 1000 tasks per job at T_0, and get processed by `compression-scheduler` at T_1. 2 compression jobs arrived with 2 tasks after at T_2.

- If `max_concurrent_tasks_per_job` is not set, the later submitted 2 jobs need to wait for all 4000 tasks to complete before they can make progress.
- If `max_concurrent_tasks_per_job` is set but larger than 2, the later 2 jobs still need to wait for 4 * **N** tasks to complete before they can make progress.

In Spider, jobs are running in round-robin task by task within an active job queue. The configured number guarantees that at most **M** jobs can make progress concurrently. So in the above case, Spider can make sure all 6 jobs are making progress together even if the two small jobs are submitted later, as long as the configured **M** is larger than 6. (NOTE: Since the dispatch queue implementation would introduce some scheduling -> dispatching delay, we can't guarantee this is 100% fair in real time. But I doubt we need to explain that in this user doc)

Besides, even with `max_concurrent_tasks_per_job` configured, the later submitted jobs still need to wait for the current batch to complete before they can be scheduled (and the straggler becomes a broker again in this case). This is not strictly related to what's documented here, but it also hurts fairness.

For the coordinator-side rate limit, I'm not sure if it's the best place to document it since it's mainly to avoid submitting too many jobs to Spider and thus blowing up the memory usage inside storage/scheduler. I put it here with a TODO since they're related and in case I forgot. But strictly speaking, this should be an improvement as well. `compression-scheduler` doesn't have any rate limit control, and it becomes extremely slow if the number of compression jobs blows up to thousands (from what I've experienced in testing env). We might still want to mention it somewhere else, maybe in another bullet point.

As mentioned https://github.com/y-scope/clp/pull/2456#discussion_r3721560492, it might be better to confirm with @junhaoliao to make sure what I state above matches `compression-scheduler` implementation since I wasn't actively involved in `compression-scheduler`'s development.

docs/src/user-docs/guides-using-compression-coordinator.md
kirkrodrigues: > Shall we restructure this sentence to indicate that Spider's service failure is automatically handled as well?

Perhaps we can add a few sub-bullets under "Automatic failure recovery:". One about `compression-coordinator` and one about Spider being fault-tolerant itself?

> But I'm not sure if we need to assume a Spider service fails infinitely without getting recovered.

I think that's outside the scope of reasonable failure recovery, right? Like failure recovery also has its limits.

docs/src/user-docs/guides-using-compression-coordinator.md
kirkrodrigues: Just to clarify. In the current compression scheduler, tasks are submitted in batches, but it's possible for one batch to be from job A and another to be from job B, right? In other words, each job is making progress at a granularity of one batch.

Do you want to say that in Spider, each job is making progress at a granularity of one task?

docs/src/user-docs/guides-using-compression-coordinator.md
kirkrodrigues: Gotcha. Yeah, maybe simplify the language to that.

docs/src/user-docs/guides-using-compression-coordinator.md
LinZhihao-723: - Sure. We can explain `compression-coordinator` and Spider separately.
- Fair. I have the same intuition but just to confirm.

docs/src/user-docs/guides-using-compression-coordinator.md
LinZhihao-723: Yeah, that's true. The key point is the task granularity execution.

docs/src/user-docs/guides-using-compression-coordinator.md
kirkrodrigues: > If max_concurrent_tasks_per_job is set but larger than 2, the later 2 jobs still need to wait for 4 * N tasks to complete before they can make progress.

This is assuming max_concurrent_tasks_per_job > the number of compression workers, right?

> Besides, even with max_concurrent_tasks_per_job configured, the later submitted jobs still need to wait for the current batch to complete before they can be scheduled (and the straggler becomes a broker again in this case). This is not strictly related to what's documented here, but it also hurts fairness.

As far as I understand (maybe I'm misunderstanding the code), that's not exactly true. The code seems to add `max_concurrent_tasks_per_job` tasks per job to Celery's queue, so Celery processes the batches at task granularity. In other words, job B doesn't need to wait for the job A's batch to complete; job B just needs to wait for a single available compression worker, which would happen when there are less than num_compression_workers tasks for job A left to complete.

Overall, I agree that round-robin of a task from each job is more fair than round-robin of N tasks from each job, so perhaps we can include that detail in the docs.

> We might still want to mention it somewhere else, maybe in another bullet point.

Yeah, I think it's worth mentioning.

docs/src/user-docs/guides-using-compression-coordinator.md
kirkrodrigues: Sure, we can be explicit, but I think it's still worth adding a short warning to say this doesn't apply to data in the `<dataset>_column_metadata` tables. Actually, maybe we should open a GitHub issue and then just link to that.

LinZhihao-723: @coderabbitai Create an issue to track this.
- Document the inconsistency introduced by the discussed behavior.
- Document the proposed solution for the all-or-nothing update: make `indexer` pass its results to the compression jobs' commit stage.
- Document we should come up with a solution to keep track of the removed column metadata in archive GC.

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,80 @@
...
+:::
+
+Compared with `compression-scheduler`, `compression-coordinator` provides the following
+user-experience, reliability, and performance improvements:
+
+* **Automatic failure recovery**:
+  * `compression-coordinator` automatically resumes previously submitted jobs (users don't need to
+    restart them manually).
+  * Spider's fault-tolerance model allows it to recover transparently from internal failures without
+    requiring user intervention.
```
LinZhihao-723: Split into two bulllet points for `compression-coordinator` and Spider.

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,80 @@
...
+* **Easier failure handling via a configurable retry policy**: `compression-coordinator` allows
+  users to configure the maximum number of retries for each compression task (retries are
+  unconditional). This can help tasks recover automatically from transient failures, such as
+  temporary network connectivity issues.
+  * TODO: Link to #2457.
```
LinZhihao-723: * Simplified the writing a bit.
* Mentioned that the retry is unconditional.

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,80 @@
...
+* **Improved resource utilization**: `compression-scheduler` processes tasks in batches, where each
+  batch must wait for its slowest task to finish. `compression-coordinator` instead schedules
+  individual tasks through Spider, allowing resources to be reassigned as soon as individual tasks
+  finish.
```
LinZhihao-723: Be more specific about batch-granularity vs. task-granularity.

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,80 @@
...
+* **Bounded job concurrency**: `compression-coordinator` limits the number of compression jobs that
+  can be submitted to Spider concurrently. In contrast, `compression-scheduler` does not bound the
+  number of concurrent jobs, which can cause scheduling overhead to grow significantly at high
+  levels of job concurrency.
```
LinZhihao-723: Split this from "fairness".

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,80 @@
...
+* **Improved fairness between concurrent compression jobs**: Compression jobs submitted by
+  `compression-coordinator` are admitted into a bounded set of active jobs. Tasks from these active
+  jobs are scheduled in round-robin order, while additional submitted jobs remain pending until an
+  active slot becomes available.
+  * `compression-scheduler` achieves similar job-level fairness only when
+    `max_concurrent_tasks_per_job` is set to 1, effectively scheduling one task per job in
+    round-robin order. However, this round-robin scheduling spans an unbounded number of concurrent
+    jobs and can incur significant overhead at high concurrency. Spider provides similar fairness
+    with much lower scheduling overhead through a more efficient architecture and a bounded set of
+    active jobs.
```
LinZhihao-723: iiuc, this is the most precise description of `compression-scheduler`'s behavior: only if `max_concurrent_tasks_per_job` is 1, it can do a task-level round-robin. But that means jobs are reported at the task level in the metadata DB, whose overhead should be as bad as Spider's C++ implementation.
I'm not sure if we need this sub-bullet point.

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,80 @@
...
+* **Improved all-or-nothing semantics**: Compression jobs coordinated by `compression-coordinator`
+  publish all archive metadata to the `clp_<dataset>_archives` table in a single dedicated commit
+  operation. The commit operation is both transactional and idempotent, so:
+  * a job-level failure doesn't result in partial updates.
+  * internal task retries don't result in duplicate updates.
```
LinZhihao-723, line 27: Be specific that the atomic updates is on `clp_<dataset>_archives` table.

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,80 @@
...
+
+:::{warning}
+The **all-or-nothing semantics** do not apply to the column metadata table
+(`<dataset>_column_metadata`), because it is updated during compression job execution. If a
+compression job fails, this table may contain partial updates. This known limitation is tracked in
+[this GitHub issue][column-metadata-issue].
```
LinZhihao-723: A small section warning the column metadata updates is not "all-or-nothing".

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,80 @@
+# Using `compression-coordinator`
+
+`compression-coordinator` is a new component that coordinates compression jobs scheduled via
+[Spider][spider]. It's designed to replace the existing `compression-scheduler`, which schedules
+compression jobs using Celery.
```
kirkrodrigues:
```suggestion
[Spider][spider]. `compression-scheduler` is designed to replace the existing
`compression-scheduler` which schedules compression jobs using Celery.
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,80 @@
...
+
+To use `compression-coordinator`, enable the Spider scheduling framework when starting the CLP
+package.
+
+TODO: Link to Chenxing's doc.
+
+:::{note}
+Currently, `compression-coordinator` can be deployed only on Kubernetes using the CLP Helm chart.
+Support for Docker Compose deployments is planned for a future release.
```
kirkrodrigues:
```suggestion
Currently, `compression-coordinator` can be deployed via Kubernetes and not via Docker Compose.
Support for Docker Compose is planned for a future release.
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,80 @@
...
+:::
+
+Compared with `compression-scheduler`, `compression-coordinator` provides the following
+user-experience, reliability, and performance improvements:
```
kirkrodrigues:
```suggestion
Compared with the `compression-scheduler`-based architecture (which uses Celery),
the `compression-coordinator`-based architecture (which uses Spider) provides the following
```

docs/src/user-docs/guides-using-compression-coordinator.md
kirkrodrigues:
```suggestion
* **Improved failure recovery**:
  * After a failure and restart, `compression-coordinator` can resume in-progress jobs, whereas
    `compression-scheduler` assumes that in-progress jobs have hung and so it kills them.
  * After a failure and restart, Spider can resume in-progress jobs, whereas Celery cannot since
    the compression tasks aren't idempotent.
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,80 @@
...
+
+* **Automatic failure recovery**:
+  * `compression-coordinator` automatically resumes previously submitted jobs (users don't need to
+    restart them manually).
+  * Spider's fault-tolerance model allows it to recover transparently from internal failures without
+    requiring user intervention.
+* **Easier failure handling via a configurable retry policy**: `compression-coordinator` allows
+  users to configure the maximum number of retries for each compression task (retries are
+  unconditional). This can help tasks recover automatically from transient failures, such as
+  temporary network connectivity issues.
```
kirkrodrigues:
```suggestion
* **Easier failure handling**: With `compression-coordinator` and Spider, if a compression task
  fails, it can automatically be retried, allowing tasks to automatically recover from transient
  failures.
```

docs/src/user-docs/guides-using-compression-coordinator.md
kirkrodrigues: I think we can drop this point because the improvement is only visible in certain scenarios; specifically, when there are multiple active jobs and the batch size is much smaller than the number of workers, or there's a single active job.

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,80 @@
...
+  * TODO: Link to #2457.
+* **Improved resource utilization**: `compression-scheduler` processes tasks in batches, where each
+  batch must wait for its slowest task to finish. `compression-coordinator` instead schedules
+  individual tasks through Spider, allowing resources to be reassigned as soon as individual tasks
+  finish.
+* **Bounded job concurrency**: `compression-coordinator` limits the number of compression jobs that
```
kirkrodrigues: Can we swap this and the next point?

docs/src/user-docs/guides-using-compression-coordinator.md
kirkrodrigues:
```suggestion
* **Bounded job concurrency**: `compression-coordinator` limits the number of compression jobs that
  can be submitted to Spider concurrently, whereas `compression-scheduler` doesn't; the latter can
  cause significantly high scheduling overheads when there are many concurrent jobs.
```

docs/src/user-docs/guides-using-compression-coordinator.md
kirkrodrigues:
```suggestion
* **Improved fairness between concurrent compression jobs**: Both architectures process jobs in
  round-robin order, but the `compression-coordinator`-based architecture does so at the granularity
  of tasks whereas the `compression-scheduler`-based architecture does so at the granularity of
  batches of tasks. If `max_concurrent_tasks_per_job` is set to 1, the
  `compression-scheduler`-based architecture can achieve the same level of fairness as the
  `compression-coordinator`-based architecture, but with higher overhead since Spider is more
  efficient than `compression-scheduler`.
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,80 @@
...
+  `compression-coordinator` are admitted into a bounded set of active jobs. Tasks from these active
+  jobs are scheduled in round-robin order, while additional submitted jobs remain pending until an
+  active slot becomes available.
+  * `compression-scheduler` achieves similar job-level fairness only when
+    `max_concurrent_tasks_per_job` is set to 1, effectively scheduling one task per job in
+    round-robin order. However, this round-robin scheduling spans an unbounded number of concurrent
+    jobs and can incur significant overhead at high concurrency. Spider provides similar fairness
+    with much lower scheduling overhead through a more efficient architecture and a bounded set of
+    active jobs.
+* **Improved all-or-nothing semantics**: Compression jobs coordinated by `compression-coordinator`
```
kirkrodrigues, line 23: Can we make this the first point so that we can refer to it in later points?

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,80 @@
...
+  publish all archive metadata to the `clp_<dataset>_archives` table in a single dedicated commit
+  operation. The commit operation is both transactional and idempotent, so:
+  * a job-level failure doesn't result in partial updates.
+  * internal task retries don't result in duplicate updates.
+
+:::{warning}
```
kirkrodrigues: Can we add this warning under the bullet rather than after the entire bullet list? (Essentially, indent this warning.)

kirkrodrigues (commented):
Might need a few more touch-ups after this round of review.

docs/src/user-docs/guides-using-compression-coordinator.md
LinZhihao-723: As discussed offline, we decided to discuss scalability in this bullet point.

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,77 @@
+# Using `compression-coordinator`
+
+`compression-coordinator` is a new component that coordinates compression jobs scheduled via
+[Spider][spider]. `compression-scheduler` is designed to replace the existing
+`compression-scheduler` which schedules compression jobs using Celery.
+
```
kirkrodrigues, line 6: This is to move the last note from below up here and to change the second-last note into a "Limitations" section instead.

```suggestion

:::{note}
`compression-coordinator` has not yet reached feature-parity with `compression-scheduler` (see
the [limitations](#limitations) section for details). Until then, the two are meant to run
side-by-side, with `compression-scheduler` handling any jobs that `compression-coordinator` doesn't
yet support.
:::
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,77 @@
...
+TODO: Link to Chenxing's doc.
+
+:::{note}
+Currently, `compression-coordinator` can be deployed via Kubernetes and not via Docker Compose.
+Support for Docker Compose is planned for a future release.
+:::
+
+Compared with the `compression-scheduler`-based architecture (which uses Celery),
+the `compression-coordinator`-based architecture (which uses Spider) provides the following
```
kirkrodrigues, line 20:
```suggestion
the `compression-coordinator`-based architecture (which uses Spider) provides the following
user-experience, reliability, and performance improvements:
```

Sorry, I accidentally deleted this in my last suggestion.

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,77 @@
...
+
+* **Improved all-or-nothing semantics**: Compression jobs coordinated by `compression-coordinator`
+  publish all archive metadata to the `clp_<dataset>_archives` table in a single dedicated commit
+  operation. The commit operation is both transactional and idempotent, so:
```
kirkrodrigues:
```suggestion
  publish all archive metadata to the `clp_<dataset>_archives` table in a single commit operation.
  The commit operation is both transactional and idempotent, so:
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,77 @@
...
+  * a job-level failure doesn't result in partial updates.
+  * internal task retries don't result in duplicate updates.
+
+  :::{warning}
+  The **all-or-nothing semantics** do not apply to the column metadata table
+  (`<dataset>_column_metadata`), because it is updated during compression job execution. If a
+  compression job fails, this table may contain partial updates. This known limitation is tracked in
+  [this GitHub issue][column-metadata-issue].
```
kirkrodrigues:
```suggestion
  (`clp_<dataset>_column_metadata`), since it's still updated during compression job execution. So
  if a compression job fails, this table may contain partial updates. This limitation is tracked in
  [y-scope/clp#2480][column-metadata-issue].
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,77 @@
...
+    the compression tasks aren't idempotent.
+* **Easier failure handling**: With `compression-coordinator` and Spider, if a compression task
+  fails, it can automatically be retried, allowing tasks to automatically recover from transient
+  failures.
+  * TODO: Link to #2457.
+* **Improved horizontal scalability**: Celery serializes task dispatching, so dispatch overhead
+  grows linearly with the number of tasks submitted by `compression-scheduler`, limiting the
+  benefits of adding Celery workers beyond a certain scale. Spider dispatches tasks concurrently,
+  allowing `compression-coordinator` to scale more effectively as additional Spider workers are
+  added.
```
kirkrodrigues:
```suggestion
* **Improved horizontal scalability**: Within Celery, tasks are dispatched to the work queue
  serially, whereas within Spider, tasks are dispatched in parallel. When there are many more tasks
  than workers, Celery's task-dispatching overhead is hidden since task execution is the bottleneck.
  However, Celery's horizontal scalability is limited since as the number of workers is increased,
  the bottleneck shifts to Celery's task-dispatching overhead.
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,77 @@
...
+* **Improved fairness between concurrent compression jobs**: Both architectures process jobs in
+  round-robin order, but the `compression-coordinator`-based architecture does so at the granularity
+  of tasks whereas the `compression-scheduler`-based architecture does so at the granularity of
+  batches of tasks. If `max_concurrent_tasks_per_job` is set to 1, the
```
kirkrodrigues:
```suggestion
  of *tasks* whereas the `compression-scheduler`-based architecture does so at the granularity of
  *batches of tasks*. If `max_concurrent_tasks_per_job` is set to `1`, the
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,77 @@
...
+:::{note}
+`compression-coordinator` currently has the following functional limitations:
+
+* It's only available when using `clp-json`.
+* It only handles compression jobs created by `log-ingestor`.
+* It only supports writing archives to S3.
+
+Additional capabilities will be introduced in future releases as it moves toward feature parity with
+`compression-scheduler`.
+:::
```
kirkrodrigues:
```suggestion
# Limitations

`compression-coordinator` currently has the following functional limitations:

* It's only available when using `clp-json`.
* It only handles compression jobs created by `log-ingestor`.
* It only supports writing archives to S3.

Additional capabilities will be introduced in future releases as it moves toward feature parity with
`compression-scheduler`.
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,77 @@
...
+
+:::{note}
+`compression-scheduler` is planned for deprecation and will eventually be fully replaced by
+`compression-coordinator`. Currently, `compression-scheduler` continues to run alongside
+`compression-coordinator` to handle compression jobs that `compression-coordinator` doesn't yet
+support.
+:::
```
kirkrodrigues:
```suggestion
```

docs/src/user-docs/guides-using-compression-coordinator.md
LinZhihao-723: - Is "work queue" a specific term used by Celery?
- Can we simplify it to "Celery's horizontal scalability is limited because the serialized task-dispatching overhead becomes the bottleneck as the number of workers increases.". "When there are many more tasks than workers, Celery's task-dispatching overhead is hidden since task execution is the bottleneck." This is not 100% correct. This depends on the task execution time vs. the number of workers. If the number of worker is small while the average task execution time is also short, the bottleneck would still be the serialized task dispatching. Assume each task takes 3ms, task dispatching is 1 ms, and there're 8 workers. For a single task execution, 1ms vs. 3ms, the execution is still the bottleneck. But since 8 workers are serialized, the first worker will wait for another 5ms to serve the next task. Made a drawing for visualization purposes. This varies w.r.t. the number of workers and the average execution time per task.
  - At a given time, only 3 workers are making actual task progress in parallel. This is also evidence of the resource utilization argument (but might be irrelevant for the compression coordinator; worth discussing for the search coordinator).
  - Compression would still face this issue when the number of workers grows to a certain threshold, so I think the second half of the explanation still makes sense.

image

kirkrodrigues: How about:

```suggestion
* **Improved horizontal scalability**: With Celery, tasks are dispatched to the task queue serially,
  whereas with Spider, tasks are dispatched in parallel. Thus, if we add more workers, at some
  point, Spider scales better whereas Celery gets bottlenecked by the task-dispatching overhead.
```

docs/src/user-docs/guides-using-compression-coordinator.md
LinZhihao-723: My bad. Didn't do a second check on this paragraph.

docs/src/user-docs/guides-using-compression-coordinator.md
LinZhihao-723, line 1: Adding sub-bullet points to explain how to configure it. Not sure if we actually need this. If not, we can remove them.

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,81 @@
...
+  efficient than `compression-scheduler`.
+  * To configure the maximum number of jobs Spider schedules concurrently, set
+    {confval}`spider.spiderConfig.scheduler.config.active_job_queue_size`.
+* **Bounded job concurrency**: `compression-coordinator` limits the number of compression jobs that
+  can be submitted to Spider concurrently, whereas `compression-scheduler` doesn't; the latter can
+  cause significantly high scheduling overheads when there are many concurrent jobs.
+  * To configure the maximum number of compression jobs that `compression-coordinator` submits to
+    Spider concurrently, set `max_concurrent_jobs` in the `compression-coordinator` configuration.
+
+## Limitations
```
LinZhihao-723, line 65: The suggestion made this an H1 title. I think it's a typo so I switched to H2. Let me know if it's suggested on purpose.

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,81 @@
...
+side-by-side, with `compression-scheduler` handling any jobs that `compression-coordinator` doesn't
+yet support.
+:::
+
+To use `compression-coordinator`, enable the Spider scheduling framework when starting the CLP
+package. See [Using Spider scheduler][using-spider-scheduler] for details.
+
+:::{note}
+Currently, `compression-coordinator` can be deployed via Kubernetes and not via Docker Compose.
+Support for Docker Compose is planned for a future release.
```
kirkrodrigues: Can we move this to the limitations section?

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,81 @@
...
+:::
+
+Compared with the `compression-scheduler`-based architecture (which uses Celery),
```
kirkrodrigues, line 19:
```suggestion
## Benefits

Compared with the `compression-scheduler`-based architecture (which uses Celery),
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,81 @@
...
+
+  :::{warning}
+  The **all-or-nothing semantics** do not apply to the column metadata table
+  (`clp_<dataset>_column_metadata`), since it's still updated during compression job execution. So
+  if a compression job fails, this table may contain partial updates. This limitation is tracked in
+  [y-scope/clp#2480][column-metadata-issue].
+  :::
+
+* **Improved failure recovery**:
+  * After a failure and restart, `compression-coordinator` can resume in-progress jobs, whereas
```
kirkrodrigues:
```suggestion
  * After `compression-coordinator` fails and restarts, it can resume in-progress jobs, whereas
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,81 @@
...
+    `compression-scheduler` assumes that in-progress jobs have hung and so it kills them.
+  * After a failure and restart, Spider can resume in-progress jobs, whereas Celery cannot since
+    the compression tasks aren't idempotent.
```
kirkrodrigues:
```suggestion
  * After Spider fails and restarts, it can resume in-progress jobs, whereas Celery cannot since the
    compression tasks aren't idempotent.
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,81 @@
...
+* **Easier failure handling**: With `compression-coordinator` and Spider, if a compression task
+  fails, it can automatically be retried, allowing tasks to automatically recover from transient
+  failures.
+  * To configure the maximum number of retries for compression and commit tasks, set
+    `compression_task_max_retry` and `commit_task_max_retry`, respectively, in the
+    `compression-coordinator` configuration.
+* **Improved horizontal scalability**: With Celery, tasks are dispatched to the task queue serially,
+  whereas with Spider, tasks are dispatched in parallel. Thus, if we add more workers, at some
+  point, Spider scales better whereas Celery gets bottlenecked by the task-dispatching overhead.
```
kirkrodrigues:
```suggestion
  point, Spider scales better whereas Celery gets bottlenecked by its task-dispatching overhead.
```

docs/src/user-docs/guides-using-compression-coordinator.md
```diff
@@ -0,0 +1,81 @@
...
+* **Improved fairness between concurrent compression jobs**: Both architectures process jobs in
+  round-robin order, but the `compression-coordinator`-based architecture does so at the granularity
+  of *tasks* whereas the `compression-scheduler`-based architecture does so at the granularity of
+  *batches of tasks*. If `max_concurrent_tasks_per_job` is set to `1`, the
+  `compression-scheduler`-based architecture can achieve the same level of fairness as the
+  `compression-coordinator`-based architecture, but with higher overhead since Spider is more
+  efficient than `compression-scheduler`.
+  * To configure the maximum number of jobs Spider schedules concurrently, set
+    {confval}`spider.spiderConfig.scheduler.config.active_job_queue_size`.
```
kirkrodrigues, line 58: In a future PR, can add a configuration-list for `compression-coordinator` like we have in the Spider guide linked here?

LinZhihao-723: Ideally we should make a configuration-list for all components. We should probably have a dedicated `Configuration` section for all components.

docs/src/user-docs/guides-using-spider-scheduler.md
```diff
@@ -41,7 +41,8 @@ The CLP Helm chart will then automatically deploy the following components:
 
   :::{note}
   In the current release, `compression-coordinator` supports only compression jobs that use certain
-  CLP package configurations. See "Using compression-coordinator" for the current limitations.
+  CLP package configurations. See [this section][using-compression-coordinator-limitations] for the
+  current limitations.
```
kirkrodrigues:
```suggestion
  CLP package configurations. Other notable limitations are listed
  [here][using-compression-coordinator-limitations].
```

docs/src/user-docs/guides-using-compression-coordinator.md
LinZhihao-723: @coderabbitai Can you create an issue to track this feature request on CLP doc?

docs/src/user-docs/guides-using-compression-coordinator.md
LinZhihao-723: Done. Split functional and deployment limitations into subsections.

kirkrodrigues (approved):

PR #2462, opened 2026-08-08
fix(compression-coordinator): Ensure running Spider jobs always have a `dispatch_time`.

Bill-hbrhbr:
# Description

The compression coordinator can leave a `Running` job without a `dispatch_time` if it restarts after `persist_spider_job_id` changes the job from `Pending` to `Running`, but before the batched `mark_jobs_dispatched` update at the end of the polling interval.

On the next startup, the recovery process picks up `Running` jobs with a `spider_id` and resumes them via `to_completion()`, but does not populate the missing `dispatch_time`.

To keep `dispatch_time` representative of when the job was actually dispatched, `persist_spider_job_id` now also sets it if it has not already been set by the coordinator. This ensures the timestamp is recorded close to the actual dispatch rather than being reconstructed after a restart.

# Checklist

  breaking change.

# Validation performed

* [X] Benign SQL query string update that doesn't affect coordinator's functional correctness.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved existing job dispatch timestamps instead of overwriting them.
  * Ensured running compression jobs receive a dispatch timestamp when one is not already set.
  * Improved timestamp accuracy for jobs whose dispatch time is recorded by the job handler.

### DIFF
--- a/components/compression-coordinator/src/coordination.rs
+++ b/components/compression-coordinator/src/coordination.rs
@@ -290,6 +290,9 @@ impl Coordinator {
 
     /// Marks the compression jobs identified by `job_ids` with the current dispatch time.
     ///
+    /// If the `dispatch_time` has already been set by the `job_handler`, we preserve the value and
+    /// skip the update. See [`S3CompressionJobHandle::persist_spider_job_id`] for details.
+    ///
     /// # Errors
     ///
     /// Returns an error if:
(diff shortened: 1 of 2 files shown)


components/compression-coordinator/src/coordination.rs
```diff
@@ -290,6 +290,9 @@ impl Coordinator {
 
     /// Marks the compression jobs identified by `job_ids` with the current dispatch time.
     ///
+    /// If the `dispatch_time` has already been set by the `job_handler`, we preserve the value and
+    /// skip the update. See [`S3CompressionJobHandle::persist_spider_job_id`] for details.
+    ///
```
LinZhihao-723:
```suggestion
```
- I don't think this is needed.
- The tone in this docstring is problematic: it shoudn't be "we xxx".

LinZhihao-723 (changes requested):
One docstring update otherwise lgtm.

LinZhihao-723 (approved):

PR #2467, opened 2026-08-10
feat(telemetry): Add archive storage metrics.

Nathan903:
# Description

This PR adds two new OpenTelemetry metrics (`clp.storage.archive.bytes_compressed` and `clp.storage.archive.bytes_uncompressed`) to track the total logical size of all archives across the deployment. 

To prevent this from slowing down the main compression scheduler, a background thread is used, which periodically polls the metadata database to sum up the archive sizes. 

* What is included/excluded: The metrics include structures and indexes embedded in each archive, but exclude database storage overhead, logs, staging files, query results, and deleted archives. They do not contain archive-specific attributes.
* Empty deployments: If a deployment is valid but has no archives, it will correctly report zero for both gauges. 
* Failure handling: If the database query fails, produces partial results, or finds invalid data sizes, the metrics are simply not emitted for that cycle until a subsequent poll succeeds.  
* Polling Frequency: `telemetry_update_interval_ms` now controls both how often metrics are exported and how often the database is checked for new storage totals.
* Assumptions: This implementation assumes there is only a single compression scheduler running, which owns these deployment-wide metrics. Running multiple schedulers could result in double-counted metrics.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

- **New Features**
  - Added optional archive storage telemetry gauges for logical compressed and uncompressed archive sizes.
  - Metrics collection now handles incomplete or invalid results safely and shuts down cleanly.
  - Added configurable polling interval, defaulting to the existing telemetry update interval when unset.
- **Documentation**
  - Documented archive storage gauge units, scope, exclusions, attributes, and unavailable-data behaviour.
- **Deployment**
  - Added Helm configuration support for the archive metrics polling interval.
  - Updated the Helm chart version.


components/compression-coordinator/src/job_handle.rs
LinZhihao-723, line 1: Didn't review this file carefully yet, but the changes are far from the coding guideline. Future reviewers (possibly myself) should pay attention to it.

LinZhihao-723 (changes requested):

PR #2473, opened 2026-08-11
feat(profiling): Adds the `utils/profiler` library to simplify performance benchmarking.

davidlion:
# Description

## Overview
This PR adds a convenient compile-time-enabled profiling library to help perform performance benchmarking. It is largely adapted from https://github.com/y-scope/clp/pull/2080.

## Usage

Currently the main intended usage has 3 parts:
1. Enable profiling through the compilation flag `CLP_ENABLE_PROFILING`:
    - `unset/0`: Disable profiling with most code reduced to empty stubs.
    - `1`: Enable base profiling through `PROFILE_SCOPE` macros.
    - `2`: Enable fine grained profiling through `PROFILE_SCOPE_DEBUG` macros.
    - Larger values imply lower ones as well.
2. Create a `utils::profiling::Reporter` that will emit all measurements taken within its lifetime/scope.
    - Each profiler is named and takes a callback used to emit the measurements.
    - When reporters are nested any measurements taken within a child will only be emit by the child and the name of the child reporter is prefixed by its parent's name (e.g. `<parent>.<child>`).
3. Add `PROFILE_SCOPE("scope")` calls under the `Reporter` to measure scopes and report the results.
    - The scope-name is appended to the current scope path to create the full name (e.g. `<reporter>.<scope>`). The scope path is built up of both parent `Reporter` and `PROFILE_SCOPE` names.
    - `PROFILE_SCOPE_DEBUG("scope")` was added as another level for more fine grained profiling. It is not used in this PR, but is currently used in more internal CLP+ code outside this PR.

### Reporting

A `set_profiler_measurement` function was added to `SearchTelemetry.hpp` to enable reporting measurements through the existing telemetry infrastructure. Otherwise, `Reporter` contains a convenience function to write to Spdlog that is used to replace the old profiling in CLP/CLG/CLO/GLT.

## Details

The top level class is `ScopedProfiler` which utilizes 3 underlying classes `Profiler`, `Reporter`, and `Stopwatch`.

- `Stopwatch`: Measure an interval duration and count using start/stop calls.
    - `start` is a no-op if already running.
    - `stop` is a no-op if not running.
    - The duration is accumulated across all intervals with the count increased on each start+stop pair.
    - Can be used directly for manual measurement without the profiler/reporter.
    - Not disabled at compile time and can be used directly at any time if useful (otherwise not referenced).
- `Profiler`: Thread-local registry of named Stopwatch objects.
    - Owns a map of fully named `Stopwatch`es and keeps thread-local stacks of nested scope path names and `Profiler`s.
    - Exposes static methods to start/stop a named Stopwatch without the caller needing to hold a reference to the current `Profiler` (the one at the top of the stack).
    - It is functional on its own, if there was a reason to manage measurements without a `Reporter`.
- `Reporter`: Thread-local measurement reporting using RAII.
    - Owns a `Profiler` whose measurements are reported through a call back function (emitter).
- `ScopedProfiler`: Thread-local measurement using RAII.
    - Starts a named `Stopwatch` on construction and stops it on destruction.
    - Utilizes `Profiler` to write the measurement to the active `Profiler`.
    - If the same "full name" (same scope name and scope path) are measured it is accumulated in the same `Stopwatch`/measurement.
        - The scope path is a string so it will treat two different code paths that use identical naming as the same measurement IFF the active profiler is the same. If you have multiple codes paths with different `Reporter`s (and therefore different `Profiler`s) and the exact same names, you'll end up with two distinct measurements with the same full name.
    - Uses a thread-local cached values at each call site to avoid re-building the full name in hot paths.
    - Recursion is guarded against and re-entrant calls are a no-op.

All of the components are thread-safe as they use thread local storage and delete move+copy constructor+assignment operators to ensure objects are destructed on the same thread they were constructed on. However, there is no way to ensure the emitters are thread-safe (fwiw, the current ones aren't).

Unit tests are always built and directly undefine and define `CLP_ENABLE_PROFILING` to ensure it is enabled. Any issues when profiling is disabled hopefully (admittedly not 100% sure) appear at compile time as everything other than `Stopwatch` either ceases to exist or becomes empty stub types.

## Other

Besides adding the new profiling utility this PR:
- Adds unit tests.
- Updated the profiling in CLP/CLG/CLO to use the new classes.
- Removes the old profiler utility.
- Updates tasks to ensure linting occurs with profiling enabled.
- Adds `utils` as an external include for clang-format.
    - Ideally, `regex_utils` and `string_utils` should be moved under `utils/regex` and `utils/string`, at least until they're moved to ystdlib.

Current formatting indents nested preprocessor directives, but not "normal" code. This makes macros look nice, but causes include statements to be indented unlike the normal code. After talking to Kirk, include statements get their own `#if` block to make things less confusing.

# Checklist

  breaking change.

# Validation performed

- CI passing.
- New unit tests added.
- Currently used with CLP+ development.

## Summary by CodeRabbit

- **New Features**
  - Added configurable profiling support with hierarchical scopes and measurement reporting.
  - Added profiling data emission through application logs and search telemetry.
  - Added profiling instrumentation for searches, compression, and main application workflows.

- **Refactor**
  - Replaced legacy profiling and timing components with a shared profiling utility.
  - Profiling can now be enabled or disabled at build time.

- **Tests**
  - Added coverage for profiling scopes, reporters, nested measurements, and duration tracking.


components/core/src/utils/profiling/ScopedProfiler.hpp
```diff
@@ -0,0 +1,129 @@
...
+// NOLINTBEGIN(cppcoreguidelines-macro-usage)
+#if defined(CLP_ENABLE_PROFILING) && CLP_ENABLE_PROFILING > 0
+    #define PROFILE_SCOPE_IMPL(counter, name) \
+        static thread_local ::std::string _prof_scope_full_name_##counter; \
+        static thread_local ::std::string _prof_scope_path_##counter; \
+        ::utils::profiling::ScopedProfiler const _prof_scope_profiler_##counter { \
+            name, _prof_scope_full_name_##counter, _prof_scope_path_##counter \
+        }
+
+    #define PROFILE_SCOPE(name) PROFILE_SCOPE_IMPL(__COUNTER__, name)
```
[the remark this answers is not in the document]
davidlion: Isn't `__COUNTER__` expanded in line 117 by `PROFILE_SCOPE` and `counter` is already the expanded version?

components/core/CMakeLists.txt
```diff
@@ -143,6 +148,14 @@ if(BUILD_TESTING AND CLP_BUILD_TESTING)
...
 endif()
 
+# Always enable profiling if building tests.
+set(CLP_ENABLE_PROFILING_EFFECTIVE "${CLP_ENABLE_PROFILING}")
+if(CLP_ENABLE_TESTS AND CLP_ENABLE_PROFILING LESS "1")
+    set(CLP_ENABLE_PROFILING_EFFECTIVE "1")
+endif()
+
+add_definitions(-DCLP_ENABLE_PROFILING=${CLP_ENABLE_PROFILING_EFFECTIVE})
+
```
gibber9809: I think without reading this cmake file I wouldn't expect enabling tests to mean that profiling is always enabled.

Is the reason we do this just so we can run the unit tests that test the profiler directly? If so, is there any way we could make it so that profiling is only enabled by default for those unit tests? E.g., I think we could turn it into a header-only library, then do some dirty hacks inside of the profiler unit tests to undefine all of our include guards for the profiler classes, include the profiler classes inside of a namespace unique to those unit tests, then undefine the include guards again.

Alternatively, we can just try to document this well internally, and make sure e.g., we aren't accidentally doing releases that accidentally have profiling enabled as a side effect of building tests.

gibber9809: Also, if we did go with the approach of making it header only then doing some hacks for the tests, it seems like we can get around any problems with defining the thread local static variables using `thread_local inline static` to avoid needing a separate definition, since we're using c++20.

components/core/src/utils/profiling/test/test_Profiler.cpp
```diff
@@ -0,0 +1,75 @@
...
+    Profiler::start_measurement(full_name);
+    std::this_thread::sleep_for(std::chrono::milliseconds(cSleepMsMedium));
+    Profiler::stop_measurement(full_name);
+}
+
+TEST_CASE("profiler_noop_without_active_reporter", "[Profiler]") {
+    Profiler::start_measurement("no_reporter");
+    std::this_thread::sleep_for(std::chrono::milliseconds(cSleepMsShort));
+    Profiler::stop_measurement("no_reporter");
+}
```
gibber9809: Do we need this test? I guess it tells us that starting/stopping a measurement without a reporter doesn't crash, but it doesn't have any asserts, so it doesn't seem like it's doing much for us.

components/core/src/utils/profiling/test/emitters.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <string>
+#include <string_view>
+
+#include <catch2/catch_test_macros.hpp>
+#include <utils/profiling/Stopwatch.hpp>
+
+namespace utils::profiling::test {
+inline constexpr int cSleepMsShort{10};
+inline constexpr int cSleepMsMedium{50};
+inline constexpr int cSleepMsLong{100};
```
gibber9809: Could we get rid of the medium and long options? As far as I can tell they're not really adding anything to these tests.

gibber9809 (changes requested):
Code pretty much looks good, just had a few comments. Mostly I want understand what the reasoning is for tests being built enabling profiling for all targets as a side-effect, and see if there's anything we can do.

components/core/CMakeLists.txt
davidlion: > Mostly I want understand what the reasoning is for tests being built enabling profiling for all targets as a side-effect

Honestly, I just didn't realize how awkward of a situation it creates since our `task`s don't build `unitTest` by default.

I like the header-only solution so I implemented that. Although, I'm open to anything else too.

For the unit tests, I still think it is better to always test profiling rather than have multiple unit test builds (although conditionally building the profiling tests would make a lot of sense). To avoid macro redefinition warnings (and since it seems kinda fragile), I added undef+define to force the profiling test files to enable profiling before including the headers. Again, open to any other ideas too.

components/core/src/utils/profiling/test/test_Profiler.cpp
davidlion: Good point, removed.

components/core/src/utils/profiling/test/emitters.hpp
davidlion: Good point, removed.

gibber9809 (approved):
LGTM. PR title is fine as well. Also messed around with the `CLP_ENABLE_PROFILING` flag on my own, and everything seems to work as expected.

PR #2474, opened 2026-08-11
build(clp_s::ffi::sfa): Add search dependencies and make OpenTelemetry and Libarchive optional for downstream WebAssembly.

Bill-hbrhbr:
# Description

This change prepares `clp_s::ffi::sfa` for downstream WebAssembly search functionalities by providing flags to disable OpenTelemetry and Libarchive support. Downstream `clp-ffi-js` will link against `clp_s::search` without the need to cross-compile these two libraries.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **New Features**
  * Added an option to enable OpenTelemetry-based search telemetry.
  * Added support for configuring search dependencies through the build system.
  * Added no-op telemetry behavior when OpenTelemetry is disabled, preserving search functionality without telemetry overhead.
* **Build Improvements**
  * Improved dependency linking for search-related functionality.
  * Updated feature configuration so FFI SFA can use search capabilities.

### DIFF
--- a/components/core/cmake/Options/options.cmake
+++ b/components/core/cmake/Options/options.cmake
@@ -96,6 +96,12 @@ option(
 )
 
 option(
+    CLP_BUILD_CLP_S_SEARCH_ENABLE_OPENTELEMETRY_CPP
+    "Include open telemetry support for clp_s::search."
+    ON
+)
+
+option(
     CLP_BUILD_CLP_S_SEARCH_KQL
     "Build clp_s::search::kql."
     ON
@@ -311,6 +317,8 @@ endfunction()
 function(validate_clp_s_ffi_sfa_dependencies)
     validate_clp_dependencies_for_target(CLP_BUILD_CLP_S_FFI_SFA
         CLP_BUILD_CLP_S_ARCHIVEREADER
+        CLP_BUILD_CLP_S_SEARCH
+        CLP_BUILD_CLP_S_SEARCH_KQL
     )
 endfunction()
 
@@ -392,16 +400,21 @@ function(set_clp_s_search_dependencies)
     set_clp_need_flags(
         CLP_NEED_ABSL
         CLP_NEED_LOG_SURGEON
-        CLP_NEED_OPENTELEMETRY_CPP
         CLP_NEED_SIMDJSON
         CLP_NEED_SPDLOG
-        CLP_NEED_XXHASH
     )
 endfunction()
 
 function(set_clp_s_search_ast_dependencies)
     set_clp_need_flags(
         CLP_NEED_SIMDJSON
+    )
+endfunction()
+
+function(set_clp_s_search_enable_opentelemetry_cpp_dependencies)
+    set_clp_need_flags(
+        CLP_NEED_OPENTELEMETRY_CPP
+        CLP_NEED_XXHASH
     )
 endfunction()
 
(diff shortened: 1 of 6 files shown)


components/core/cmake/Options/options.cmake
```diff
@@ -405,6 +411,13 @@ function(set_clp_s_search_ast_dependencies)
     )
 endfunction()
 
+function(set_clp_s_search_enable_opentelemetry_cpp_dependencies)
+    set_clp_need_flags(
+        CLP_NEED_OPENTELEMETRY_CPP
+        CLP_NEED_XXHASH
+    )
+endfunction()
```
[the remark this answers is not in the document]
Bill-hbrhbr: users can manually turn off both if they want. having `CLP_BUILD_CLP_S_SEARCH = OFF` and `CLP_BUILD_CLP_S_SEARCH_ENABLE_OPENTELEMETRY_CPP=ON` is harmless. We don't want to couple build options tightly together as shown by other build options code inside this file.

components/core/cmake/Options/options.cmake
```diff
@@ -95,6 +95,12 @@ option(
     ON
 )
 
+option(
+    CLP_BUILD_CLP_S_SEARCH_ENABLE_OPENTELEMETRY_CPP
+    "Include open telemetry support for clp_s::search."
```
gibber9809, line 100:
```suggestion
    CLP_BUILD_CLP_S_ENABLE_OPENTELEMETRY
    "Include open telemetry support for clp-s."
```
Could we rename the flag (and other corresponding functions, etc.) to this? If we did add opentelemetry support to other parts of clp-s we'd probably want to share the same flag, instead of relying on a different flag for each sub-component, so we might as well just create a flag at the clp-s level now.

components/core/src/clp_s/ffi/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ if(CLP_BUILD_CLP_S_FFI_SFA)
                 clp_s_ffi_sfa
                 PUBLIC
                 clp_s::archive_reader
+                clp_s::search
```
gibber9809, line 21:
```suggestion
```
Not used in this PR? Should probably be part of the PR where you add search to the ffi.

components/core/src/clp_s/search/CMakeLists.txt
```diff
@@ -47,11 +45,30 @@ if(CLP_BUILD_CLP_S_SEARCH)
...
+                        opentelemetry-cpp::resources
+                        opentelemetry-cpp::trace
+                        xxHash::xxhash
+                )
+        else()
+                target_sources(
+                        clp_s_search
+                        PRIVATE
+                        SearchTelemetryNoop.cpp
+                        TelemetryContextNoop.cpp
```
gibber9809, line 71: Instead of adding new cpp files let's use the more standard approach of doing:
```
target_compile_definitions(
        clp_s_search
        PRIVATE
        CLP_BUILD_CLP_S_ENABLE_OPENTELEMETRY=$<BOOL:${CLP_BUILD_CLP_S_ENABLE_OPENTELEMETRY}>
)
```
Then use `#if ...` macros to stub out implementations in the existing cpp files.

gibber9809 (changes requested):
Could we also use the `CLP_BUILD_CLP_S_ENABLE_OPENTELEMETRY` flag in `clp_s/CommandLineArguments.cpp` to remove the `enable-telemetry` flag using some `#if` macros when open telemetry is disabled in the build?

PR #2484, opened 2026-08-20
refactor(webui)!: Stream search and aggregation results via the API server SSE endpoint.

hoophalab:
# Description

Replace WebUI server search routes and MongoDB socket cursors with SSE
streaming from the API server, which now owns job submission, cancellation,
and result delivery. The WebUI reverse-proxies `/api/v1/*` to it.

BREAKING CHANGE: The API server is now required to run the WebUI; the old
`/api/search/*` routes are removed.

# Checklist

  breaking change.

# Validation performed

1. clp-json on docker compose behaves correctly.
2. clp-text on docker compose behaves correctly.
3. clp-json on helm behaves correctly.

## Summary by CodeRabbit

* **New Features**
  * Search results and timeline aggregations now stream through the API server.
  * Added optional raw-document results for query-result streams.
  * Added count-by-time aggregation support for search queries.
  * Web UI now connects to the API server through a configured proxy.

* **Bug Fixes**
  * Improved Web UI configuration validation and API server defaults.
  * Search and aggregation jobs now report completion and failures more consistently.

* **Documentation**
  * Updated API server, buffering, deployment, and configuration guidance.


docs/src/user-docs/guides-using-presto.md
```diff
@@ -45,12 +45,10 @@ the clp-s native query components can optionally be disabled to save resources.
...
-     # Optional: Disable the clp-s native query pipeline to save resources.
-     # NOTE: The API server depends on the clp-s native query pipeline.
-     api_server: null
-     query_scheduler: null
-     query_worker: null
-     reducer: null
+      # Optional: Disable the clp-s native query pipeline to save resources.
+      query_scheduler: null
+      query_worker: null
+      reducer: null
```
junhaoliao:
```suggestion
     # Optional: Disable the clp-s native query pipeline to save resources.
     query_scheduler: null
     query_worker: null
     reducer: null
```

components/api-server/src/client.rs
```diff
@@ -649,8 +683,11 @@ impl Client {
             .with_options(find_options)
```
junhaoliao, line 700: discussed offline - to get the latest N results

```rust
let find_options = if max_num_results > 0 {
    mongodb::options::FindOptions::builder()
        .sort(mongodb::bson::doc! {"timestamp": -1, "_id": -1})
        .limit(i64::from(max_num_results))
        .build()
} else {
    mongodb::options::FindOptions::default()
};
```

components/webui/packages/api-client/package.json
```diff
@@ -22,10 +22,13 @@
   "license": "Apache-2.0",
   "type": "module",
   "dependencies": {
-    "openapi-fetch": "^0.17.0"
+    "eventsource-parser": "^4.0.0",
+    "openapi-fetch": "^0.17.0",
+    "react": "^19.2.3"
```
junhaoliao: the build is broken at this head, on a cold build, so CI will fail. Moving `react` out of `peerDependencies` was the right direction, but `dependencies` still triggers injection while `syncInjectedDepsAfterScripts` is now gone, so nothing populates the injected copy's `dist/`.

Verified at `7923ab5` from a clean install:

```shell
rm -rf node_modules packages/*/node_modules packages/*/dist
pnpm install --frozen-lockfile
pnpm run build
```

`packages/client/node_modules/@webui/api-client` still resolves to `node_modules/.pnpm/@webui+api-client@file+packages+api-client_react@19.2.7/...`, both builds run as cache misses, and `@webui/client:build` fails with `TS2307: Cannot find module '@webui/api-client'`. `pnpm install` reports no lockfile drift, so the lockfile is not the problem — a runtime `react` dependency is enough to keep the package injected under `injectWorkspacePackages: true`.

`devDependencies` is what restores the plain symlink:

```json
  "dependencies": {
    "eventsource-parser": "^4.0.0",
    "openapi-fetch": "^0.17.0"
  },
  "devDependencies": {
    "@types/node": "24.13.3",
    "@types/react": "^19.2.9",
    "eslint": "^9.39.4",
    "eslint-config-yscope": "1.1.2",
    "openapi-typescript": "^7.9.1",
    "react": "^19.2.3",
    "typescript": "^5.9.3"
  }
```

Verified: `packages/client/node_modules/@webui/api-client -> ../../../api-client` and all five packages build. `dist/` is then shared by construction, so no sync hook is needed and the result does not depend on turbo's cache state. Safe here because the consumer bundles React and the workspace pins a single version; not offered as an applyable suggestion because `pnpm-lock.yaml` needs regenerating alongside it.

components/webui/packages/server/src/plugins/app/ApiServerProxy.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Reverse proxy that forwards `/api/v1/*` requests to the API server. The `/api/v1` prefix is
+ * stripped before forwarding since the API server's routes live at its root.
+ */
+export default fp(
+    async (fastify) => {
+        await fastify.register(fastifyHttpProxy, {
+            prefix: "/api/v1",
+            upstream: serverSettings.ApiServerUrl,
+        });
```
junhaoliao: no timeout is set here, so queries running longer than five minutes are cut off by the proxy and surface in the UI as failures while the job succeeds server-side.

`fetch_results` polls `get_status` until the job reaches `Succeeded` before returning the SSE stream, so the upstream sends no response headers for the entire duration of the query. `@fastify/reply-from` selects undici unless `opts.undici === false` or `opts.http`/`opts.http2` is set, and falls back to `undiciOpts.headersTimeout`/`bodyTimeout`, which undici's `Client` defaults to `300e3` ms. The change-stream transport this replaces had no such ceiling.

```ts
await fastify.register(fastifyHttpProxy, {
    prefix: "/api/v1",
    undici: {headersTimeout: 0, bodyTimeout: 0},
    upstream: serverSettings.ApiServerUrl,
});
```

0 disables; a deliberate upper bound works too. Left as an example rather than a suggestion since the ceiling is a policy call. Worth recording either way, because the same value will be needed as `proxy_read_timeout` when nginx replaces this server.

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -652,6 +652,9 @@ def validate_webui_config(
             f"{WEBUI_COMPONENT_NAME} {settings_json_path} is not a valid path to settings.json"
         )
 
+    if clp_config.api_server is None:
+        raise ValueError("The API server must be enabled when the Web UI is enabled.")
```
junhaoliao: updating the Presto guide resolved the contradiction here, but this still couples every Presto deployment to the API server. A Presto Web UI never calls `/api/v1` — `useResultsMetadata` and the Presto tables use `/api/presto-search` and the Mongo socket — and `package-helm/.set-up-common.sh:147` still passes `--set clpConfig.api_server=null` for its Presto path, so the two deployment routes now disagree.

Gating on the query engine keeps the useful part of the check (a clp-s user who sets `api_server: null` otherwise gets a search page that 502s with no startup signal) without forcing a dead API server onto Presto users. `QueryEngine` is already imported in this module.

```suggestion
    if QueryEngine.PRESTO != clp_config.webui.query_engine and clp_config.api_server is None:
        raise ValueError(
            "The API server must be enabled when the Web UI uses the"
            f" `{clp_config.webui.query_engine}` query engine."
        )
```

components/api-server/src/client.rs
```diff
@@ -205,6 +206,12 @@ pub struct QueryConfig {
     /// will be stored in `MongoDB` instead.
     #[serde(default)]
     pub buffer_results_in_mongodb: bool,
+
+    /// The size of each time bucket (in epoch milliseconds) for count-by-time aggregation.
```
junhaoliao:
```suggestion
    /// The size of each time bucket (in milliseconds) for count-by-time aggregation.
```

components/api-server/src/client.rs
```diff
@@ -279,6 +288,7 @@ impl Client {
     /// * Forwards [`rmp_serde::to_vec_named`]'s return values on failure.
     /// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
     pub async fn submit_query(&self, query_config: QueryConfig) -> Result<u64, ClientError> {
+        let count_by_time_bucket_size = query_config.count_by_time_bucket_size_millisecs;
```
junhaoliao, line 293: this value reaches the job config unvalidated. `clo` rejects it at execution time — `CommandLineArguments.cpp:557-560` raises "Value for count-by-time must be greater than zero." — so a client posting `0` or a negative bucket size gets a 200 with a job ID, then a job that fails in the worker.

The Web UI always sends a positive value from `computeTimelineConfig`, so this only affects direct API consumers, but it is a new public field and a 400 here is cheaper to debug than a failed job.

components/webui/packages/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Native/SearchResultsVirtualTable/useSearchResults.ts
```diff
@@ -1,54 +1,51 @@
...
-            if (searchJobId === SEARCH_STATE_DEFAULT.searchJobId) {
-                return null;
+    return useQueryResults<SearchResult>(apiClient, searchJobId, {
+        // Sort by timestamp (desc) to match the previous MongoDB cursor ordering.
+        compareFn: (a, b) => b.timestamp - a.timestamp,
+        onDone: () => {
+            const {searchUiState, updateSearchUiState} = useSearchStore.getState();
+            if (searchUiState === SEARCH_UI_STATE.QUERYING) {
+                updateSearchUiState(SEARCH_UI_STATE.DONE);
             }
```
junhaoliao: this flips to `DONE` when the search stream ends, without regard for the aggregation stream. The deleted `updateSearchSignalWhenJobsFinish` awaited both `searchJobId` and `aggregationJobId` before writing `RESP_DONE`, and `useAggregationResults` has no `onDone` at all.

The two jobs finish independently and the aggregation job goes through the reducer, so it can lag. When it does, the controls re-enable and the cancel button disappears while the timeline is still filling — and the aggregation job becomes uncancellable at that point, because `handleQueryCancel` early-returns unless the state is `QUERYING`.

Having `useAggregationResults` report completion and transitioning only once both streams have ended would restore the previous semantics.

tools/deployment/package-helm/templates/configmap.yaml
```diff
@@ -325,6 +325,8 @@ data:
           {{ .Values.clpConfig.webui.results_metadata_collection_name | quote }}
       },
       "server": {
+        "ApiServerUrl": "http://{{ include "clp.fullname" . }}-api-server:3001",
```
junhaoliao: this line is unconditional, but `api-server-deployment.yaml` and `api-server-service.yaml` are both gated on `{{- if .Values.clpConfig.api_server }}`. `helm template clp tools/deployment/package-helm --set clpConfig.api_server=null` renders `"ApiServerUrl": "http://clp-api-server:3001"` with no matching Service or Deployment.

Latent rather than breaking today — that combination is only used for the Presto path, which never calls `/api/v1`, and `@fastify/http-proxy` resolves the upstream lazily. But it is the Helm counterpart of the invariant now enforced in `validate_webui_config`, so the chart permits a combination the package rejects. Gating this line the same way would need `ApiServerUrl` to become nullable in `WebuiServerSettingsSchema` and `ApiServerProxy` to skip registration when it is null; a chart-level `fail` when the query engine is native and `api_server` is null is the cheaper option.

components/webui/packages/api-client/src/useQueryResults.ts
```diff
@@ -0,0 +1,163 @@
...
+    apiClient: ApiClient,
+    jobId: string | null,
+    options: UseQueryResultsOptions<T>
+): T[] | null => {
+    const [results, setResults] = useState<T[] | null>(null);
+
+    // Keep the latest options in a ref so that the stream isn't restarted when callers pass
+    // inline callbacks.
+    const optionsRef = useRef<UseQueryResultsOptions<T>>(options);
+    optionsRef.current = options;
```
junhaoliao: assigning a ref during render is unsafe under concurrent rendering, since a render can be discarded after the ref has been mutated. Low impact here because the callbacks are idempotent, but it is the kind of thing that surfaces once Strict Mode double-rendering or Suspense is in play.

```suggestion
    useEffect(() => {
        optionsRef.current = options;
    });
```

`useEffect` is already imported. Deliberately no dependency array, so it runs after every render and keeps the same "latest options" behavior.

components/webui/packages/api-client/src/useQueryResults.ts
```diff
@@ -0,0 +1,163 @@
...
+    // eslint-disable-next-line max-lines-per-function
+    useEffect(() => {
+        setResults(null);
+
+        if (null === jobId) {
+            return () => {
+            };
+        }
+
+        console.log(`Streaming query results for job ID: ${jobId}`);
```
junhaoliao: `@webui/api-client` has no other logging. The `client` package logs liberally, but a shared transport package writing to the console on every job change is a different ownership level — consumers cannot turn it off. Worth either dropping or exposing through a hook option alongside `onDone`/`onError`.

components/webui/packages/client/src/pages/SearchPage/SearchControls/Native/search-requests.ts
```diff
@@ -89,24 +80,28 @@ const handleQuerySubmit = (payload: QueryJobCreation) => {
...
+            ...queryConfig,
+            count_by_time_bucket_size_millisecs: payload.timeRangeBucketSizeMillis,
+        }),
+    ])
+        .then(([searchJobId, aggregationJobId]) => {
             store.updateSearchJobId(searchJobId.toString());
             store.updateAggregationJobId(aggregationJobId.toString());
             store.updateSearchUiState(SEARCH_UI_STATE.QUERYING);
             store.updateSubmittedMaxNumResults(
                 payload.maxNumResults ?? SEARCH_STATE_DEFAULT.submittedMaxNumResults
```
junhaoliao: `submittedMaxNumResults` is now write-only. `useSearchResults` was its only reader and this PR removed that read, so the store field, its updater, and this call are all dead — the result limit is owned by the API server's `max_num_results` now.

Removing it means deleting the field, its interface entry, and `updateSubmittedMaxNumResults` from `SearchState/index.tsx` as well, which is why this is not an applyable suggestion.

components/webui/packages/server/src/plugins/app/ApiServerProxy.ts
```diff
@@ -0,0 +1,18 @@
+import fastifyHttpProxy from "@fastify/http-proxy";
+import fp from "fastify-plugin";
+
+import {serverSettings} from "../../settings.js";
+
+
+/**
+ * Reverse proxy that forwards `/api/v1/*` requests to the API server. The `/api/v1` prefix is
```
junhaoliao, line 8: shall we add a TODO? to say this reverse proxy is temporary and will be removed together with the fasify server once we port everything over to the API server

junhaoliao (changes requested):
Suggest recreating the search results index in _handle_new_search_job

```python
def create_search_results_index(results_cache_uri: str, job_id: str) -> None:
    try:
        with pymongo.MongoClient(results_cache_uri) as results_cache_client:
            results_cache_client.get_default_database()[job_id].create_index(
                [("timestamp", pymongo.ASCENDING), ("_id", pymongo.ASCENDING)],
                name="timestamp-ascending",
            )
    except pymongo.errors.PyMongoError:
        logger.exception("Failed to create the results index for job %s.", job_id)
```

discussed offline -  **Results are no longer incremental.** `fetch_results` blocks on `get_status` until `Succeeded`, so nothing reaches the browser until the job finishes. This is intentional, and it is what makes a correct "latest N" possible at all. If delaying the result retrieval of a long running job is a concern, we can add polling later.

**Docs parity.** Two gaps remain: `guides-using-the-api-server.md` still says the example submits "a query to clp-json" and passes `"datasets": ["default"]`, which is clp-s-only, now that the flavour restriction is lifted; and `quick-start/clp-text.md` has no API-server section while `clp-json.md` has "Searching via the API server" - i'm fine deferring the docs updates for these into a future PR - since we have not officially declared API server support for CLP-text

components/webui/packages/server/src/plugins/app/ApiServerProxy.ts
hoophalab: @coderabbitai Can you create an issue on this?

junhaoliao (approved):

PR #2487, opened 2026-08-20
ci: Cache `task deps:core` artifacts for the Linux and macOS builds.

jackluo923:
# Description

Every binary job rebuilds ~25 C++ dependencies from source on each run. This caches the `task deps:core` tree across the four Linux binary jobs and the macOS job.

**`run-on-image` across the four Linux static legs: 111 min → 47 min (−58%).**

| variant | cold | warm | restore |
|---|---|---|---|
| centos-stream-9 | [36m05s][cold-centos] | [**9m41s**][warm-centos] | [40s][warm-centos] |
| musllinux_1_2 | [38m15s][cold-musl] | [**10m34s**][warm-musl] | [23s][warm-musl] |
| manylinux_2_28 | [26m14s][cold-many] | [**17m41s**][warm-many] | [29s][warm-many] |
| ubuntu-jammy | [10m20s][cold-jammy] | [**8m59s**][warm-jammy] | [25s][warm-jammy] |

Entries are 367–380 MB. Measured on an [earlier revision][cold-run] of this branch ([warm run][warm-run]); the mechanism is unchanged but the key has since moved to the image ID. The spread reflects contention on a shared runner pool, and `run-on-image` also covering the image pull and the core build, which the cache doesn't touch.

# How it works

`build/deps/cpp` and `.task` are restored before `task deps:core` and saved after. The CMake build directories are part of the tree, or a restore would recompile everything, and `.task` is required or `deps:utils:clean-outdated-cpp-checksum-files` wipes the restored checksums. A step deletes `mongo-c-driver`'s `.git` directory before the save, taking ~200 MB off the entry.

### Key

`deps-core-<arch>-<os>-<environment>-<taskfiles>`.

`<environment>` identifies the thing that compiles the dependencies. On Linux that's the dependency image's **image ID**: a new `prepare-image` action pulls or loads the image and reports it, so the identity exists before the key is computed, whether the image came from the registry or a workflow artifact. `run-on-image` reuses the prepared image rather than fetching it again. macOS has no image, so it hashes its installed toolchain and passes that through the same input.

Hashing the image's *sources* instead would only describe the recipe — `FROM` isn't pinned by digest and packages aren't pinned to versions, so the same recipe can yield a different compiler. The image ID can't drift that way, and being the digest of a config that carries the architecture, a multi-arch tag can't resolve to a shared identity.

There is deliberately no `restore-keys`. A prefix fallback would, after a dependency bump, build on a tree from the previous versions; `cmake --install` never removes what the old one left, so deleted or renamed headers survive in the install prefixes and the core build still finds them. Matching exactly costs a full rebuild on a dependency change — which every run pays today — so the worst case is a slow run, not a wrong binary.

### Save conditions

The ref is `main` and the event is not a `pull_request` (entries are branch-scoped but share one 10 GB quota; reads are unaffected, since every ref can restore from `main`); the restore was not an exact hit (`saveCache` archives *before* reserving the key); `deps:core` succeeded, though the unit tests need not have; and it is the `use_shared_libs: false` leg, since that flag never reaches `deps:core` so both legs share a key.

# Checklist

  breaking change.

# Validation performed

`yamllint --strict` and `actionlint` are clean.

Saves are verified: on a companion branch with the `main` gate removed, all four Linux jobs saved, and each key carried exactly the image ID `prepare-image` resolved.

Restores can't be shown before merge. Any branch carrying this change rebuilds the dependency image locally, and a rebuilt image gets a new ID, so its key never matches. That's intended — a rebuilt image can't be shown to match the one that filled the cache. It only affects the ~6% of PRs that change the image recipe, which pay a cold build anyway; the rest pull the identical published image and will hit from `main`'s second run onward.

[cold-run]: https://github.com/y-scope/clp/actions/runs/32337661306
[warm-run]: https://github.com/y-scope/clp/actions/runs/32341627754
[cold-centos]: https://github.com/y-scope/clp/actions/runs/32337661306/job/96331675977
[cold-musl]: https://github.com/y-scope/clp/actions/runs/32337661306/job/96331865420
[cold-many]: https://github.com/y-scope/clp/actions/runs/32337661306/job/96331883658
[cold-jammy]: https://github.com/y-scope/clp/actions/runs/32337661306/job/96332122243
[warm-centos]: https://github.com/y-scope/clp/actions/runs/32341627754/job/96343585236
[warm-musl]: https://github.com/y-scope/clp/actions/runs/32341627754/job/96343347002
[warm-many]: https://github.com/y-scope/clp/actions/runs/32341627754/job/96343998682
[warm-jammy]: https://github.com/y-scope/clp/actions/runs/32341627754/job/96343788603


.github/actions/deps-cache-key/action.yaml
```diff
@@ -0,0 +1,82 @@
...
+      # it. That only ever adds to the tree: the new version is downloaded, rebuilt, and installed
+      # over the old one, but `cmake --install` doesn't remove what the old one left behind, so
+      # headers and libraries that the bump deleted or renamed survive in the install prefixes and
+      # the core build can still find them. The result is a tree that matches neither version, and
+      # nothing downstream would flag it.
+      #
+      # The cost of matching exactly is that a dependency bump pays for a full rebuild -- which is
+      # what every run pays today -- so the worst case is a slow run rather than a wrong binary.
+      env:
+        IMAGE_HASH: >-
```
junhaoliao: technically, the actual docker image can change because we are not pinning the `FROM` image by digest in the docker files? also we are not specifying the apt / dnf / apk dep versions in the install-prebuilt-packages scripts so they can also change without us detecting a hash change.

anyways, i think it's more of an issue to be solved in the longer run. if there isn't already an issue tracking this, shall we ask the rabbit to open an issue?

.github/workflows/clp-core-build-macos.yaml
```diff
@@ -68,9 +68,64 @@ jobs:
...
+      # rather than the install script. Must run after `install-all.sh`.
+      - id: "toolchain"
+        name: "Capture the macOS toolchain identity"
+        run: |
+          hash=$(
+            {
+              sw_vers
+              clang --version
+              xcodebuild -version 2>/dev/null || true
+              brew list --versions
```
junhaoliao, line 84: `install-all.sh` installs CMake and Go Task from open version ranges, and both produce state included in this cache. Include their versions so upgrades invalidate the key.

```suggestion
              xcodebuild -version 2>/dev/null || true
              cmake --version
              task --version
              brew list --versions
```

.github/workflows/clp-core-build-macos.yaml
```diff
@@ -68,9 +68,64 @@ jobs:
...
+            } | shasum -a 256 | cut -d " " -f 1
+          )
+          echo "hash=${hash}" >> "$GITHUB_OUTPUT"
+          echo "macOS toolchain hash: ${hash}"
         shell: "bash"
 
+      - id: "deps_cache_key"
+        uses: "./.github/actions/deps-cache-key"
```
junhaoliao, line 92: let's add this to `monitored_paths`

```yaml
      - ".github/actions/deps-cache-key/**"
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -368,6 +368,24 @@ jobs:
...
 
+      - id: "deps_cache_key"
+        uses: "./.github/actions/deps-cache-key"
+        with:
+          os_name: "centos-stream-9"
+
+      # NOTE: Restoring after the `chown` above avoids chowning the restored tree (~2.3GB).
+      - name: "Restore deps:core cache"
+        id: "deps_cache_restore"
+        uses: "actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684"
```
junhaoliao: pin this cache use to the latest stable release and keep the readable tag beside the immutable SHA.

```suggestion
        uses: "actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9"  # v6.1.0
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -378,12 +396,34 @@ jobs:
...
+          && '' != hashFiles('build/.deps-core-ok')
+          && 'true' != steps.deps_cache_restore.outputs.cache-hit
+          && 'pull_request' != github.event_name
+          && 'refs/heads/main' == github.ref
+        run: "rm -rf build/deps/cpp/mongocxx-build/_deps/mongo-c-driver-src/.git"
+        shell: "bash"
+
+      - name: "Update deps:core cache"
+        if: *deps_cache_save_condition
+        uses: "actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684"
```
junhaoliao: pin this cache use to the latest stable release and keep the readable tag beside the immutable SHA.

```suggestion
        uses: "actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9"  # v6.1.0
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -412,6 +452,18 @@ jobs:
...
         shell: "bash"
 
+      - id: "deps_cache_key"
+        uses: "./.github/actions/deps-cache-key"
+        with:
+          os_name: "manylinux_2_28"
+
+      - name: "Restore deps:core cache"
+        id: "deps_cache_restore"
+        uses: "actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684"
```
junhaoliao: pin this cache use to the latest stable release and keep the readable tag beside the immutable SHA.

```suggestion
        uses: "actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9"  # v6.1.0
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -423,12 +475,25 @@ jobs:
...
             --num-jobs $(getconf _NPROCESSORS_ONLN)
 
+      - name: "Drop mongo-c-driver's .git directory before caching"
+        if: *deps_cache_save_condition
+        run: "rm -rf build/deps/cpp/mongocxx-build/_deps/mongo-c-driver-src/.git"
+        shell: "bash"
+
+      - name: "Update deps:core cache"
+        if: *deps_cache_save_condition
+        uses: "actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684"
```
junhaoliao: pin this cache use to the latest stable release and keep the readable tag beside the immutable SHA.

```suggestion
        uses: "actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9"  # v6.1.0
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -513,6 +578,18 @@ jobs:
...
         shell: "bash"
 
+      - id: "deps_cache_key"
+        uses: "./.github/actions/deps-cache-key"
+        with:
+          os_name: "musllinux_1_2"
+
+      - name: "Restore deps:core cache"
+        id: "deps_cache_restore"
+        uses: "actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684"
```
junhaoliao: pin this cache use to the latest stable release and keep the readable tag beside the immutable SHA.

```suggestion
        uses: "actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9"  # v6.1.0
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -524,12 +601,25 @@ jobs:
...
             --num-jobs $(getconf _NPROCESSORS_ONLN)
 
+      - name: "Drop mongo-c-driver's .git directory before caching"
+        if: *deps_cache_save_condition
+        run: "rm -rf build/deps/cpp/mongocxx-build/_deps/mongo-c-driver-src/.git"
+        shell: "bash"
+
+      - name: "Update deps:core cache"
+        if: *deps_cache_save_condition
+        uses: "actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684"
```
junhaoliao: pin this cache use to the latest stable release and keep the readable tag beside the immutable SHA.

```suggestion
        uses: "actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9"  # v6.1.0
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -563,6 +653,18 @@ jobs:
...
         shell: "bash"
 
+      - id: "deps_cache_key"
+        uses: "./.github/actions/deps-cache-key"
+        with:
+          os_name: "ubuntu-jammy"
+
+      - name: "Restore deps:core cache"
+        id: "deps_cache_restore"
+        uses: "actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684"
```
junhaoliao: pin this cache use to the latest stable release and keep the readable tag beside the immutable SHA.

```suggestion
        uses: "actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9"  # v6.1.0
```

.github/workflows/clp-artifact-build.yaml
```diff
@@ -571,12 +673,25 @@ jobs:
...
             --num-jobs $(getconf _NPROCESSORS_ONLN)
 
+      - name: "Drop mongo-c-driver's .git directory before caching"
+        if: *deps_cache_save_condition
+        run: "rm -rf build/deps/cpp/mongocxx-build/_deps/mongo-c-driver-src/.git"
+        shell: "bash"
+
+      - name: "Update deps:core cache"
+        if: *deps_cache_save_condition
+        uses: "actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684"
```
junhaoliao: pin this cache use to the latest stable release and keep the readable tag beside the immutable SHA.

```suggestion
        uses: "actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9"  # v6.1.0
```

.github/workflows/clp-core-build-macos.yaml
```diff
@@ -68,9 +68,64 @@ jobs:
...
+        with:
+          os_name: "${{matrix.os}}"
+          arch: "${{runner.arch}}"
+          env_hash: "${{steps.toolchain.outputs.hash}}"
+
+      - name: "Restore deps:core cache"
+        id: "deps_cache_restore"
+        uses: "actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684"
```
junhaoliao: pin this cache use to the latest stable release and keep the readable tag beside the immutable SHA.

```suggestion
        uses: "actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9"  # v6.1.0
```

.github/workflows/clp-core-build-macos.yaml
```diff
@@ -68,9 +68,64 @@ jobs:
...
+          && 'success' == steps.deps_core.outcome
+          && 'true' != steps.deps_cache_restore.outputs.cache-hit
+          && 'pull_request' != github.event_name
+          && 'refs/heads/main' == github.ref
+        run: "rm -rf build/deps/cpp/mongocxx-build/_deps/mongo-c-driver-src/.git"
+        shell: "bash"
+
+      - name: "Update deps:core cache"
+        if: *deps_cache_save_condition
+        uses: "actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684"
```
junhaoliao: pin this cache use to the latest stable release and keep the readable tag beside the immutable SHA.

```suggestion
        uses: "actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9"  # v6.1.0
```

junhaoliao (changes requested):
lgtm

.github/actions/deps-cache-key/action.yaml
jackluo923: cache key includes the build docker image hash.

.github/workflows/clp-core-build-macos.yaml
jackluo923: done

junhaoliao (approved):
lgtm

the CI isn't passing though - https://github.com/y-scope/clp/actions/runs/33122688976/job/98693245069?pr=2487 any chance the CI is flaky?

jackluo923:
> lgtm
> 
> the CI isn't passing though - https://github.com/y-scope/clp/actions/runs/33122688976/job/98693245069?pr=2487 any chance the CI is flaky?

Mac is very flaky and have sporadic issue downloading assets from the internet. I will re-run the mac CI until it is green.

PR #2479, opened 2026-08-17
feat(clp-package): Add a Spider-based Docker Compose deployment for Spider-orchestrated compression.

20001020ycx:
# Description

See PR title.

The goal of this PR is to show that the downloaded Spider Compose deployment can be integrated with the CLP package.
The wiring that lets a user select and configure Spider through `clp-config.yaml` is deferred to a future PR.

This PR contains a dedicated Compose entry point for Spider (`docker-compose-spider.yaml`) and the CLP runtime override for Spider's worker (`compose.clp-spider.yaml`). We also add the `compression-coordinator` service to the CLP package's Compose deployment, following the convention in the Helm chart deployment.

## Some Implementation Details for Reviewer

**`docker-compose-spider.yaml`:**
Includes all of Spider's components from `usr/share/spider-compose/compose.yaml`, where the `compose:download-spider` task places Spider's Compose deployment in the package, plus all CLP package components from `docker-compose-all.yaml`. This means `compression-scheduler` and `compression-worker` are still present, to support the ingestion paths that don't come from S3 and that `compression-coordinator` doesn't currently handle.

When this entry point gets invoked is again subject to user configuration and is deferred to a future PR; here we invoke it manually with the required env, which `sbin/start-clp.sh --setup-only` already generates into `.env`.

**`compose.clp-spider.yaml`:**
As in the Helm deployment, the CLP worker image overrides Spider's worker image so the worker carries the CLP compression TDL package. It also supplies the env vars the CLP tasks need (`CLP_DB_PASS`, `CLP_DB_USER`) and mounts the generated `clp-config.yaml` for worker to access.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

In what follows, we show a compression job submitted to the Spider Compose deployment is executed by Spider and its archive is searchable:

0. Environment: a MinIO deployment serving an input bucket `clp-input` seeded with one JSONL object under `logs/`, plus an empty archive bucket `clp-archives`. Ingestion goes through `log-ingestor`'s S3 scanner, which is the input type `compression-coordinator` consumes.

The following `clp-config.yaml` is required:

```yaml
logs_input: # type s3 + MinIO access key/secret.
archive_output:
  storage: # type s3 + staging_directory, MinIO endpoint_url/bucket/key_prefix, and the same credentials.
[...]

PR #2491, opened 2026-08-20
feat(clp-package): Select and configure Spider as the scheduler through `clp-config.yaml`.

20001020ycx:
# Description

After this PR, a CLP package user can now select the Spider-orchestrated Docker Compose deployment from `clp-config.yaml`, and configure Spider through the same file.

This PR added field `package.scheduler` to switch between Spider and Celery in `clp-config.yaml`, defaulting to "celery". The default mode preserves existing behavior, while setting it to "spider" selects docker-compose-spider.yaml and spins up the Spider and `compression_coordinator` services.

Unlike CLP-native components who read configuration directly from `clp-config.yaml` rendered by clp-config.py's pydantic model, Spider's components read theirs from downloaded Docker Compose, with env vars for interpolating the values. This PR wires values from `clp-config.yaml` to env vars.

## Remarks/implementation details for reviewers

package.scheduler is backward compatible, making this PR a non-breaking change—when omitted, the default value is used.
- Note that Spider is for clp-s only. We only expose this configuration explicitly to users in the clp-s template; we intentionally omit this field from the clp-text template to avoid confusing user with an unsupported option.
- It is also worth discussing that such field is Docker Compose only. Helm's switch for Spider is toggled by `spider.enabled` because of the subchart deployment. Therefore, in Helm, we have the freedom to neglect this field entirely.

To dive deeper on how env vars are propagated to Docker Compose, we declare every field under `Spider` as optional in `clp_config.py`: when unset, it fallback to Spider's own default value. On the other hand, for fields explicitly set by user, after translating to env vars, the rendered `clp-config.yaml` shall only retain the fields CLP itself consumes, e.g spider.host/port for compression coordinator to read. The rest is excluded through `Spider.dump_to_primitive_dict()`.
- This implementation is not entirely new, there's one precedent, `aws_config_directory`, and this optional string helper (`_optional_str()`) follows that convention to convert these types to str for env vars.
- Note that we shall rely on the fallback for the default value, rather than maintain another sets of default value in the `clp-config.yaml` to avoid maintaining duplicated copy.

There's some env vars that we deliberately don't wire for user to configure, I will briefly discuss what they are and the reason why:
| Not emitted | Why |
  |---|---|
  | `SPIDER_DATABASE_{IMAGE_REF,NAME,PORT,ROOT_PASSWORD}`, `SPIDER_STORAGE_DB_{USERNAME,PASSWORD}` | Docker compose only supports bundled database, shall not let user configures |
  | `SPIDER_STORAGE_PORT`, `SPIDER_SCHEDULER_PORT` | In-container port, not something a user should set. See y-scope/spider#457. |
  | `SPIDER_WORKER_IMAGE_REF` | CLP package related setting, we override the worker's `image:` at `compose.clp-spider.yaml`, shall not expose to user |
  | `SPIDER_WORKER_{INHERITED_ENV,PACKAGE_DIR,TASK_EXECUTOR_BIN_PATH}` | Determined by the CLP worker image's layout, e.g. its Dockerfile installs `libclp.so` under Spider's default `package_dir`, shall not expose to user |

# Checklist

# Validation performed

Environment: A MinIO S3 `clp-input` bucket seeded with one JSONL object under `logs/`, e.g. `{"timestamp":"2026-08-20T00:01:01Z","service":"pr6-e2e","message":"PR6_SPIDER_COMPOSE_PROBE_000001"}`, and an empty `clp-archives` bucket.

`clp-config.yaml:`

```yaml
package:
  storage_engine: "clp-s"
  scheduler: "spider"
logs_input: # type s3 + MinIO access key/secret
archive_output:
  storage: # type s3 + staging_directory, MinIO endpoint_url/bucket/key_prefix, and the same credentials
spider:
  port: 24517
  worker:
    replicas: 2
    log_level: "DEBUG"
  storage:
    inbound_queue:
      task_capacity: 4096
  scheduler:
    round_robin:
      tick_interval_ms: 7
compression_coordinator:
  logging_level: "DEBUG"
```

The steps below verify that the wiring takes effect. For a quick check with the clp-config.yaml above, save it to `components/package-template/src/etc/clp-config.yaml` and run `task package && build/clp-package/sbin/start-clp.sh`.

1. `.env` carries the fields the config sets, and only those

```bash
build/clp-package/sbin/start-clp.sh --config build/clp-package/etc/clp-config.yaml --setup-only
grep "^SPIDER_" build/clp-package/.env | sort
```

Expected — one line per configured field, the ~40 fields left unset produce no line, so Spider's `:-` defaults apply to them:

```
SPIDER_RESTART_POLICY=on-failure:3
SPIDER_SCHEDULER_ROUND_ROBIN_TICK_INTERVAL_MS=7
SPIDER_STORAGE_INBOUND_QUEUE_TASK_CAPACITY=4096
SPIDER_STORAGE_PORT=24517
SPIDER_STORAGE_PUBLISHED_IP=127.0.0.1
SPIDER_STORAGE_PUBLISHED_PORT=24517
SPIDER_WORKER_LOG_LEVEL=DEBUG
SPIDER_WORKER_REPLICAS=2
```

The coordinator's log level reaches the variable its service reads for `RUST_LOG`:

```bash
grep "^CLP_COMPRESSION_COORDINATOR_LOGGING_LEVEL" build/clp-package/.env
```

Expected: `CLP_COMPRESSION_COORDINATOR_LOGGING_LEVEL=DEBUG`

The config mounted into the containers resolves Spider's endpoint to the in-deployment service name, so no hand-written hostname is needed:

```bash
python3 -c "import yaml;d=yaml.safe_load(open('build/clp-package/var/log/.clp-config.yaml'));print(d['package'], d['spider']['host'], d['spider']['port'])"
```

Expected:

```
{'scheduler': 'spider', 'storage_engine': 'clp-s'} spider-storage 24517
```

2. `package.scheduler` selects the Compose entry point

```bash
build/clp-package/sbin/start-clp.sh --config build/clp-package/etc/clp-config.yaml
PROJ="clp-package-$(cat build/clp-package/var/log/instance-id)"
docker inspect "${PROJ}-compression-coordinator-1" --format '{{index .Config.Labels "com.docker.compose.project.config_files"}}'
```

Expected — the label Compose records is the file the controller chose:

```
/home/ycx/clp/clp-spider-docker-compose/build/clp-package/docker-compose-spider.yaml
```

Spin up the docker compose in Spider mode, the Celery compression services run alongside Spider's, and the worker count is the configured `spider.worker.replicas: 2` rather than Spider's default of 4:

```bash
docker compose --project-name "$PROJ" ps --format 'table {{.Service}}\t{{.Status}}' | grep -E "SERVICE|spider|compression"
```

Expected:

```
SERVICE                   STATUS
compression-coordinator   Up 11 seconds
compression-scheduler     Up 11 seconds
compression-worker        Up 46 seconds
spider-database           Up 47 seconds (healthy)
spider-scheduler          Up 16 seconds (healthy)
spider-storage            Up 19 seconds (healthy)
spider-worker             Up 13 seconds
spider-worker             Up 13 seconds
```

3. The emitted variables land in Spider's shipped configs

```bash
docker exec "${PROJ}-spider-storage-1" cat /etc/spider/storage.yaml
```

Expected — `port` and `task_capacity` come from `clp-config.yaml`, while `max_connections`, the other queue capacities and the GC intervals keep the defaults from Spider's own config, which is the reason no config file is generated:

```
host: "0.0.0.0"
port: 24517
runtime:
  db:
    host: "spider-database"
    max_connections: 64
    name: "spider-db"
    port: 3306
  inbound_queue:
    cleanup_capacity: 256
    commit_capacity: 256
    task_capacity: 4096
  job_cache_gc:
    gc_interval_sec: 30
    terminated_job_retention_sec:
      300
  task_instance_pool:
    execution_manager_stale_cutoff_sec:
      60
    gc_interval_sec: 30
    message_channel_capacity:
      128
```

Similar grep-test is performed on scheduler and worker.

4. Same E2E testing with #2479 to confirm no functionality regression

## New Features
- Added support for Spider-based compression orchestration alongside Celery.
- Added configuration options for Spider storage, scheduling, workers, health checks, queues, caching, and round-robin processing.
- Added compression coordinator logging configuration and validation for Spider connectivity.
- Docker Compose operations now automatically use the appropriate orchestration setup.
- Added example configuration settings for Spider and compression coordinator deployments.

### DIFF
--- a/components/clp-package-utils/clp_package_utils/controller.py
+++ b/components/clp-package-utils/clp_package_utils/controller.py
@@ -1372,6 +1372,7 @@ class DockerComposeController(BaseController):
                 "--file",
                 self._get_docker_file_name(),
                 "down",
+                "--remove-orphans",
             ],
             cwd=self._clp_home,
             check=True,
--- a/components/clp-package-utils/clp_package_utils/controller.py
+++ b/components/clp-package-utils/clp_package_utils/controller.py
@@ -631,7 +631,10 @@ class BaseController(ABC):
         """
         component_name = COMPRESSION_COORDINATOR_COMPONENT_NAME
         coordinator_config = self._clp_config.compression_coordinator
-        if coordinator_config is None:
+        if (
+            CompressionOrchestration.SPIDER != self._clp_config.package.scheduler
+            or coordinator_config is None
+        ):
             logger.info("%s is not configured, skipping environment setup...", component_name)
             return EnvVarsDict()
 
--- a/components/clp-package-utils/clp_package_utils/controller.py
+++ b/components/clp-package-utils/clp_package_utils/controller.py
@@ -10,6 +10,7 @@ import subprocess
 import time
 import uuid
 from abc import ABC, abstractmethod
+from collections.abc import Callable
 from typing import Any
 
 from clp_py_utils.clp_config import (
@@ -26,9 +27,11 @@ from clp_py_utils.clp_config import (
     ClpConfig,
     ClpDbNameType,
     ClpDbUserType,
+    COMPRESSION_COORDINATOR_COMPONENT_NAME,
     COMPRESSION_JOBS_TABLE_NAME,
     COMPRESSION_SCHEDULER_COMPONENT_NAME,
     COMPRESSION_WORKER_COMPONENT_NAME,
+    CompressionOrchestration,
     CONTAINER_INPUT_LOGS_ROOT_DIR,
     DatabaseEngine,
     DB_COMPONENT_NAME,
@@ -44,6 +47,9 @@ from clp_py_utils.clp_config import (
     REDIS_COMPONENT_NAME,
     REDUCER_COMPONENT_NAME,
     RESULTS_CACHE_COMPONENT_NAME,
+    SPIDER_COMPONENT_NAME,
+    SpiderScheduler,
+    SpiderSchedulerPolicy,
     StorageEngine,
     StorageType,
     WEBUI_COMPONENT_NAME,
@@ -70,6 +76,7 @@ from clp_package_utils.general import (
     validate_queue_config,
     validate_redis_config,
     validate_results_cache_config,
+    validate_spider_config,
     validate_webui_config,
 )
 
@@ -519,6 +526,122 @@ class BaseController(ABC):
             "CLP_QUERY_SCHEDULER_METRIC_EXPORT_INTERVAL": str(
                 self._clp_config.query_scheduler.telemetry_update_interval_ms
             ),
+        }
+
+        return env_vars
+
+    def _set_up_env_for_spider(self) -> EnvVarsDict:
+        """
+        Sets up environment variables for Spider's components.
+
+        :return: Dictionary of environment variables necessary to launch the components.
+        """
+        component_name = SPIDER_COMPONENT_NAME
+        spider_config = self._clp_config.spider
+        if (
+            CompressionOrchestration.SPIDER != self._clp_config.package.scheduler
+            or spider_config is None
+        ):
+            logger.info("%s is not configured, skipping environment setup...", component_name)
+            return EnvVarsDict()
+
+        logger.info("Setting up environment for %s...", component_name)
+
+        validate_spider_config(self._clp_config)
+
+        storage = spider_config.storage
+        scheduler = spider_config.scheduler
+        worker = spider_config.worker
+
+        env_vars = EnvVarsDict()
+
+        # Connection config
+        env_vars |= {
+            "SPIDER_STORAGE_PUBLISHED_IP": _get_ip_from_hostname(spider_config.host),
+            "SPIDER_STORAGE_PUBLISHED_PORT": str(spider_config.port),
+        }
+
+        # Storage config
+        env_vars |= {
+            "SPIDER_STORAGE_LOG_LEVEL": storage.log_level,
+            "SPIDER_STORAGE_DB_MAX_CONNECTIONS": _optional_str(storage.db_max_connections),
+            "SPIDER_STORAGE_INBOUND_QUEUE_CLEANUP_CAPACITY": _optional_str(
+                storage.inbound_queue.cleanup_capacity
+            ),
+            "SPIDER_STORAGE_INBOUND_QUEUE_COMMIT_CAPACITY": _optional_str(
+                storage.inbound_queue.commit_capacity
+            ),
+            "SPIDER_STORAGE_INBOUND_QUEUE_TASK_CAPACITY": _optional_str(
+                storage.inbound_queue.task_capacity
+            ),
+            "SPIDER_STORAGE_JOB_CACHE_GC_INTERVAL_SEC": _optional_str(
+                storage.job_cache_gc.gc_interval_sec
+            ),
+            "SPIDER_STORAGE_JOB_CACHE_TERMINATED_JOB_RETENTION_SEC": _optional_str(
+                storage.job_cache_gc.terminated_job_retention_sec
+            ),
+            "SPIDER_STORAGE_TASK_INSTANCE_POOL_EXECUTION_MANAGER_STALE_CUTOFF_SEC": _optional_str(
+                storage.task_instance_pool.execution_manager_stale_cutoff_sec
+            ),
+            "SPIDER_STORAGE_TASK_INSTANCE_POOL_GC_INTERVAL_SEC": _optional_str(
+                storage.task_instance_pool.gc_interval_sec
+            ),
+            "SPIDER_STORAGE_TASK_INSTANCE_POOL_MESSAGE_CHANNEL_CAPACITY": _optional_str(
+                storage.task_instance_pool.message_channel_capacity
+            ),
+        }
+
+        # Scheduler config
+        env_vars |= {
+            "SPIDER_SCHEDULER_LOG_LEVEL": scheduler.log_level,
+            "SPIDER_SCHEDULER_POLICY": scheduler.policy,
+            "SPIDER_SCHEDULER_CONNECTION_POOL_SIZE": _optional_str(scheduler.connection_pool_size),
+            "SPIDER_SCHEDULER_STOP_TIMEOUT_SEC": _optional_str(scheduler.stop_timeout_sec),
+            "SPIDER_SCHEDULER_EM_REGISTRY_DEAD_EM_CUTOFF_SEC": _optional_str(
+                scheduler.em_registry.dead_em_cutoff_sec
+            ),
+            "SPIDER_SCHEDULER_EM_REGISTRY_LIVENESS_TRACKING_INTERVAL_MS": _optional_str(
+                scheduler.em_registry.liveness_tracking_interval_ms
+            ),
+        }
+        env_vars |= _get_env_for_spider_scheduler_policy(scheduler)
+
+        # Worker config
+        env_vars |= {
+            "SPIDER_WORKER_REPLICAS": _optional_str(worker.replicas),
+            "SPIDER_WORKER_LOG_LEVEL": worker.log_level,
+            "SPIDER_WORKER_CONNECTION_POOL_SIZE": _optional_str(worker.connection_pool_size),
+            "SPIDER_WORKER_SCHEDULER_POLL_WAIT_MS": _optional_str(worker.scheduler_poll_wait_ms),
+            "SPIDER_WORKER_MAX_LOG_LINE_BYTES": _optional_str(worker.max_log_line_bytes),
+            "SPIDER_WORKER_LIVENESS_SCHEDULER_HEARTBEAT_INTERVAL_SEC": _optional_str(
+                worker.liveness.scheduler_heartbeat_interval_sec
+            ),
+            "SPIDER_WORKER_LIVENESS_STORAGE_HEARTBEAT_INTERVAL_SEC": _optional_str(
+                worker.liveness.storage_heartbeat_interval_sec
+            ),
+        }
+
+        return env_vars
+
+    def _set_up_env_for_compression_coordinator(self) -> EnvVarsDict:
+        """
+        Sets up environment variables for the compression coordinator component.
+
+        :return: Dictionary of environment variables necessary to launch the component.
+        """
+        component_name = COMPRESSION_COORDINATOR_COMPONENT_NAME
+        coordinator_config = self._clp_config.compression_coordinator
+        if coordinator_config is None:
+            logger.info("%s is not configured, skipping environment setup...", component_name)
+            return EnvVarsDict()
+
+        logger.info("Setting up environment for %s...", component_name)
+
+        env_vars = EnvVarsDict()
+
+        # Runtime config
+        env_vars |= {
+            "CLP_COMPRESSION_COORDINATOR_LOGGING_LEVEL": coordinator_config.logging_level,
         }
 
         return env_vars
@@ -1097,6 +1220,7 @@ class DockerComposeController(BaseController):
         # Restart Policy
         env_vars |= {
             "CLP_RESTART_POLICY": self._restart_policy,
+            "SPIDER_RESTART_POLICY": self._restart_policy,
         }
 
         # Credentials
@@ -1132,9 +1256,8 @@ class DockerComposeController(BaseController):
         env_vars |= self._set_up_env_for_telemetry()
 
         # Paths
-        aws_config_dir = self._clp_config.aws_config_directory
         env_vars |= {
-            "CLP_AWS_CONFIG_DIR_HOST": (None if aws_config_dir is None else str(aws_config_dir)),
+            "CLP_AWS_CONFIG_DIR_HOST": _optional_str(self._clp_config.aws_config_directory),
             "CLP_DATA_DIR_HOST": str(self._clp_config.data_directory),
             "CLP_LOGS_DIR_HOST": str(self._clp_config.logs_directory),
             "CLP_TMP_DIR_HOST": str(self._clp_config.tmp_directory),
@@ -1171,6 +1294,8 @@ class DockerComposeController(BaseController):
         env_vars |= self._set_up_env_for_results_cache()
         env_vars |= self._set_up_env_for_otel_collector()
         env_vars |= self._set_up_env_for_compression_scheduler()
+        env_vars |= self._set_up_env_for_compression_coordinator()
+        env_vars |= self._set_up_env_for_spider()
         env_vars |= self._set_up_env_for_query_scheduler()
         env_vars |= self._set_up_env_for_compression_worker(num_workers)
         env_vars |= self._set_up_env_for_query_worker(num_workers)
@@ -1202,7 +1327,7 @@ class DockerComposeController(BaseController):
         logger.info("Starting CLP using Docker Compose...")
 
         cmd = ["docker", "compose", "--project-name", self._project_name]
-        cmd += ["--file", "docker-compose.yaml"]
+        cmd += ["--file", self._get_docker_file_name()]
         cmd += ["up", "--detach", "--wait"]
         subprocess.run(
             cmd,
@@ -1239,11 +1364,27 @@ class DockerComposeController(BaseController):
             logger.info("Stopping all CLP containers using Docker Compose...")
 
         subprocess.run(
-            ["docker", "compose", "--project-name", self._project_name, "down"],
+            [
+                "docker",
+                "compose",
+                "--project-name",
+                self._project_name,
+                "--file",
+                self._get_docker_file_name(),
+                "down",
+            ],
             cwd=self._clp_home,
             check=True,
         )
         logger.info("Stopped CLP.")
+
+    def _get_docker_file_name(self) -> str:
+        """
+        :return: The Docker Compose file name to use based on the config.
+        """
+        if CompressionOrchestration.SPIDER == self._clp_config.package.scheduler:
+            return "docker-compose-spider.yaml"
+        return "docker-compose.yaml"
 
     @staticmethod
     def _get_num_workers() -> int:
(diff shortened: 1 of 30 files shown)


components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -1025,6 +1126,24 @@ def validate_log_ingestor_config(self):
...
+                f" `{CompressionOrchestration.SPIDER}`."
+            )
+            raise ValueError(msg)
+        if self.compression_coordinator is None:
+            msg = (
+                "`compression_coordinator` must be configured when `package.scheduler` is"
+                f" `{CompressionOrchestration.SPIDER}`."
+            )
+            raise ValueError(msg)
+        return self
```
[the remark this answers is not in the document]
20001020ycx: docker compose wont deploy these service anyway when celery mode is on, add documentation to clarify this behaviour

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -135,6 +138,21 @@ class DatabaseEngine(KebabCaseStrEnum):
 DatabaseEngineStr = Annotated[DatabaseEngine, StrEnumSerializer]
 
 
+class CompressionOrchestration(KebabCaseStrEnum):
```
junhaoliao: non-blocking: the new public configuration types and methods add fresh Ruff `D101`/`ANN201` failures to an ignored component baseline. Add concise docstrings to every new public enum/model and annotate the new public method/validator returns. This spans several ranges, so an example is safer than one suggestion:

```python
class CompressionOrchestration(KebabCaseStrEnum):
    """Supported compression orchestration backends."""

def dump_to_primitive_dict(self) -> dict[str, Any]:
    ...

def transform_for_container(self) -> None:
    ...

def validate_compression_orchestration_config(self) -> "ClpConfig":
    ...
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -750,9 +769,89 @@ class LogIngestor(BaseModel):
     logging_level: LoggingLevelRust = "INFO"
 
 
+class SpiderInboundQueue(BaseModel):
+    cleanup_capacity: PositiveInt | None = None
```
junhaoliao: blocking: these values are interpolated into configs deserialized by pinned Rust services, but `PositiveInt` accepts values beyond `u32`, `u64`, and 64-bit `usize`. Bound `db_max_connections` as `u32`; bound every other Spider numeric override except Compose-owned `replicas` as `u64`/64-bit `usize`:

```python
NonNegativeUInt32 = Annotated[int, Field(ge=0, le=2**32 - 1)]
PositiveUInt32 = Annotated[int, Field(gt=0, le=2**32 - 1)]
PositiveUInt64 = Annotated[int, Field(gt=0, le=2**64 - 1)]

class SpiderStorage(BaseModel):
    db_max_connections: PositiveUInt32 | None = None

class SpiderInboundQueue(BaseModel):
    cleanup_capacity: PositiveUInt64 | None = None
    commit_capacity: PositiveUInt64 | None = None
    task_capacity: PositiveUInt64 | None = None
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -765,6 +864,7 @@ class PollingBackoff(BaseModel):
```
junhaoliao: blocking: the coordinator currently accepts values Rust cannot deserialize, values above Tokio’s semaphore limit, and timeout policies Spider rejects. The affected fields span unchanged context, so this is an example rather than a misleading one-line suggestion:

```python
class PollingBackoff(BaseModel):
    init_backoff_millisecs: Annotated[int, Field(gt=0, le=2**64 - 1)]
    max_backoff_millisecs: Annotated[int, Field(gt=0, le=2**64 - 1)]


class CompressionCoordinator(BaseModel):
    # Tokio reserves the top three bits of `usize` for semaphore state.
    MAX_CONCURRENT_JOBS: ClassVar[int] = (2**64 - 1) >> 3
    MAX_COMMIT_TASK_TIMEOUT_SECS: ClassVar[int] = 24 * 60 * 60

    logging_level: LoggingLevelRust = "INFO"
    resource_group: SpiderResourceGroup = SpiderResourceGroup(name="compression-coordinator")
    job_polling_interval_millisecs: Annotated[int, Field(gt=0, le=2**64 - 1)] = 100
    max_concurrent_jobs: Annotated[int, Field(gt=0, le=MAX_CONCURRENT_JOBS)] = 1000
    result_polling: PollingBackoff = PollingBackoff(
        init_backoff_millisecs=100, max_backoff_millisecs=1000
    )
    compression_task_max_retry: Annotated[int, Field(ge=0, le=2**32 - 1)] = 1
    commit_task_max_retry: Annotated[int, Field(ge=0, le=2**32 - 1)] = 1
    database_connection_pool_size: Annotated[int, Field(gt=0, le=2**32 - 1)] = 10
    termination_timeout_secs: Annotated[int, Field(gt=0, le=2**64 - 1)] = 30
    commit_task_soft_timeout_secs: Annotated[
        int, Field(gt=0, le=MAX_COMMIT_TASK_TIMEOUT_SECS)
    ] = 45
    commit_task_hard_timeout_secs: Annotated[
        int, Field(gt=0, le=MAX_COMMIT_TASK_TIMEOUT_SECS)
    ] = 60

    @model_validator(mode="after")
    def validate_commit_task_timeouts(self) -> "CompressionCoordinator":
        if self.commit_task_hard_timeout_secs <= self.commit_task_soft_timeout_secs:
            msg = (
                "`commit_task_hard_timeout_secs` must be greater than"
                " `commit_task_soft_timeout_secs`."
            )
            raise ValueError(msg)
        return self
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -1025,6 +1125,26 @@ def validate_log_ingestor_config(self):
...
+            msg = (
+                "`compression_coordinator` must be configured when `package.scheduler` is"
+                f" `{CompressionOrchestration.SPIDER}`."
+            )
+            raise ValueError(msg)
+        return self
+
     @model_validator(mode="after")
     def validate_compression_coordinator_config(self):
         if self.compression_coordinator is None:
```
junhaoliao: blocking: the preceding validator says coordinator config is ignored outside Spider mode, but this validator still rejects `scheduler: celery` with a coordinator-only block. Gate the compatibility checks on actual deployment:

```suggestion
        if (
            CompressionOrchestration.SPIDER != self.package.scheduler
            or self.compression_coordinator is None
        ):
```

components/clp-package-utils/clp_package_utils/general.py
```diff
@@ -658,6 +659,10 @@ def validate_webui_config(
     validate_port(f"{WEBUI_COMPONENT_NAME}.port", clp_config.webui.host, clp_config.webui.port)
 
 
+def validate_spider_config(clp_config: ClpConfig):
+    validate_port(f"{SPIDER_COMPONENT_NAME}.port", clp_config.spider.host, clp_config.spider.port)
```
junhaoliao: non-blocking: this public validator dereferences an optional model and adds new annotation/docstring lint failures. Make standalone calls fail deliberately rather than with `AttributeError`:

```suggestion
def validate_spider_config(clp_config: ClpConfig) -> None:
    """Validate Spider's host-published storage endpoint."""
    spider_config = clp_config.spider
    if spider_config is None:
        msg = "`spider` must be configured before its endpoint can be validated."
        raise ValueError(msg)
    validate_port(f"{SPIDER_COMPONENT_NAME}.port", spider_config.host, spider_config.port)
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -523,6 +530,122 @@ def _set_up_env_for_query_scheduler(self) -> EnvVarsDict:
...
+
+    def _set_up_env_for_compression_coordinator(self) -> EnvVarsDict:
+        """
+        Sets up environment variables for the compression coordinator component.
+
+        :return: Dictionary of environment variables necessary to launch the component.
+        """
+        component_name = COMPRESSION_COORDINATOR_COMPONENT_NAME
+        coordinator_config = self._clp_config.compression_coordinator
+        if coordinator_config is None:
```
junhaoliao: blocking: coordinator settings are documented as ignored in Celery mode, but this helper still emits coordinator environment variables. Keep setup deployment-inert under the same discriminator:

```suggestion
        if (
            CompressionOrchestration.SPIDER != self._clp_config.package.scheduler
            or coordinator_config is None
        ):
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1202,7 +1327,7 @@ def start(self) -> None:
         logger.info("Starting CLP using Docker Compose...")
 
         cmd = ["docker", "compose", "--project-name", self._project_name]
-        cmd += ["--file", "docker-compose.yaml"]
+        cmd += ["--file", self._get_docker_file_name()]
```
junhaoliao: blocking: selecting `docker-compose-spider.yaml` currently launches both compression orchestrators because that file still extends `compression-scheduler` and `compression-worker`. Both schedulers fetch `PENDING` jobs, so they can race. The YAML is outside this PR diff; remove those services and gate the two Celery env helpers:

```diff
-  compression-scheduler:
-    extends:
-      file: "docker-compose-all.yaml"
-      service: "compression-scheduler"
-  compression-worker:
-    extends:
-      file: "docker-compose-all.yaml"
-      service: "compression-worker"
```

```python
if CompressionOrchestration.CELERY != self._clp_config.package.scheduler:
    return EnvVarsDict()
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1239,12 +1364,28 @@ def stop(self) -> None:
...
         subprocess.run(
-            ["docker", "compose", "--project-name", self._project_name, "down"],
+            [
+                "docker",
+                "compose",
+                "--project-name",
+                self._project_name,
+                "--file",
+                self._get_docker_file_name(),
+                "down",
```
junhaoliao: blocking: after starting Spider, changing the config to Celery before `stop` selects the default file; Compose then leaves the undefined Spider services running. Remove project orphans during teardown:

```suggestion
                "down",
                "--remove-orphans",
```

components/package-template/src/etc/clp-config.template.json.yaml
```diff
@@ -57,6 +58,34 @@ telemetry:
 #  logging_level: "INFO"
 #  telemetry_update_interval_ms: 60000
 #
+#spider:
```
junhaoliao: non-blocking: make scheduler ownership explicit where users copy these optional blocks; otherwise ignored settings look active. Add analogous Celery-only comments above `compression_scheduler` and `compression_worker` (outside this hunk):

```suggestion
## The Spider and compression-coordinator settings below are used only when
## package.scheduler is "spider".
#spider:
```

```yaml
## The compression-scheduler settings below are used only when package.scheduler is "celery".
## The compression-worker settings below are used only when package.scheduler is "celery".
```

sitaowang1998 (dismissed):
Reviewed the config entries. All config entries are available.

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: non-blocking: Ruff reports `D101` and `ANN201` for the public types and methods added here. This component already ignores these lint categories, but this PR should not add more violations. Please add short docstrings and return types to the new public API. The changes span several sections, so this example shows the expected form:

```python
class CompressionOrchestration(KebabCaseStrEnum):
    """Supported compression orchestration backends."""

def dump_to_primitive_dict(self) -> dict[str, Any]:
    ...

def transform_for_container(self) -> None:
    ...

def validate_compression_orchestration_config(self) -> "ClpConfig":
    ...
```

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao, line 773: shall we consider checking the sizes of the inputs before passing them into the Rust services? e.g.,

```python
NonNegativeUInt32 = Annotated[int, Field(ge=0, le=2**32 - 1)]
PositiveUInt32 = Annotated[int, Field(gt=0, le=2**32 - 1)]
PositiveUInt64 = Annotated[int, Field(gt=0, le=2**64 - 1)]

class SpiderStorage(BaseModel):
    db_max_connections: PositiveUInt32 | None = None

class SpiderInboundQueue(BaseModel):
    cleanup_capacity: PositiveUInt64 | None = None
    commit_capacity: PositiveUInt64 | None = None
    task_capacity: PositiveUInt64 | None = None
```

components/clp-py-utils/clp_py_utils/clp_config.py
```diff
@@ -765,6 +864,7 @@ class PollingBackoff(BaseModel):
 
 
 class CompressionCoordinator(BaseModel):
+    logging_level: LoggingLevelRust = "INFO"
```
junhaoliao, line 867: ditto - consider bound checking for the numbers

components/clp-py-utils/clp_py_utils/clp_config.py
junhaoliao: blocking: `validate_compression_orchestration_config()` says the coordinator config is ignored unless the scheduler is Spider. This validator still runs in Celery mode. A Celery config with only `compression_coordinator` then fails because `spider` is missing. Return before these checks unless Spider is selected:

```suggestion
        if (
            CompressionOrchestration.SPIDER != self.package.scheduler
            or self.compression_coordinator is None
        ):
```

components/clp-package-utils/clp_package_utils/general.py
junhaoliao: non-blocking: the only caller already narrows `spider_config`, so another runtime `None` check would duplicate that invariant. Encode it in the parameter type instead. This also addresses the missing Ruff docstring and return annotation; the import and caller need coordinated changes:

```python
from clp_py_utils.clp_config import Spider


def validate_spider_config(spider_config: Spider) -> None:
    """Validate the host-published Spider storage endpoint."""
    validate_port(
        f"{SPIDER_COMPONENT_NAME}.port",
        spider_config.host,
        spider_config.port,
    )

# controller.py, after the existing None check
validate_spider_config(spider_config)
```

components/clp-package-utils/clp_package_utils/controller.py
junhaoliao: i guess there is no harm, but technically we don't need to populate any compression_coordinator env vars when spider isn't enabled?

```suggestion
        if (
            CompressionOrchestration.SPIDER != self._clp_config.package.scheduler
            or coordinator_config is None
        ):
```

components/clp-package-utils/clp_package_utils/controller.py
junhaoliao, line 1377: i think this sequence is possible when an end user doesn't read our docs before they switch the orchestrator type:
1. they start a CLP package with spider compression coordinator (`package.scheduler: spider`)
2. they change `package.scheduler` back to `celery`
3. run `stop-clp.sh`
4. because the default docker compose file is now used, which doesn't include the spider services, spider related service would still be running after

not to say we should officially support changing config in between startups / teardowns, but i think we should make our helper scripts more fault tolerant -

i think simply specifying `--remove-orphans` will address the concern

```suggestion
                "down",
                "--remove-orphans",
```

components/package-template/src/etc/clp-config.template.json.yaml
```diff
@@ -57,6 +58,34 @@ telemetry:
...
+#  result_polling:
+#    init_backoff_millisecs: 100
+#    max_backoff_millisecs: 1000
+#  compression_task_max_retry: 1
+#  commit_task_max_retry: 1
+#  database_connection_pool_size: 10
+#  termination_timeout_secs: 30
+#  commit_task_soft_timeout_secs: 45
+#  commit_task_hard_timeout_secs: 60
+#
```
junhaoliao: these fields default to `null` (`None` in the Pydantic definitions in `ClpConfig`), but the current examples look like mappings with implicit defaults. Let's show the actual defaults, keep the nested examples easy to enable, and state the required scheduler change:

```suggestion
## `spider` and `compression_coordinator` default to `null`. To use them, set `package.scheduler`
## to "spider", remove `null` from both, and uncomment their fields below.
#spider: null
##  host: "localhost"
##  port: 50051
##  storage:
##    log_level: "INFO"
##  scheduler:
##    log_level: "INFO"
##    policy: "round_robin"
##  worker:
##    replicas: 4
##    log_level: "INFO"
#
#compression_coordinator: null
##  logging_level: "INFO"
##  resource_group:
##    name: "compression-coordinator"
##  job_polling_interval_millisecs: 100
##  max_concurrent_jobs: 1000
##  result_polling:
##    init_backoff_millisecs: 100
##    max_backoff_millisecs: 1000
##  compression_task_max_retry: 1
##  commit_task_max_retry: 1
##  database_connection_pool_size: 10
##  termination_timeout_secs: 30
##  commit_task_soft_timeout_secs: 45
##  commit_task_hard_timeout_secs: 60
#
```

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1362,6 +1503,73 @@ def _chown_recursively(
     subprocess.run(chown_cmd, stdout=subprocess.DEVNULL, check=True)
 
 
+def _get_env_for_spider_round_robin_scheduler(scheduler: SpiderScheduler) -> EnvVarsDict:
```
junhaoliao: how about renaming as  `_get_spider_round_robin_scheduler_env_vars`

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1362,6 +1503,73 @@ def _chown_recursively(
...
+
+
+_SPIDER_SCHEDULER_POLICY_ENV_GETTERS: dict[
+    SpiderSchedulerPolicy, Callable[[SpiderScheduler], EnvVarsDict]
+] = {
+    SpiderSchedulerPolicy.ROUND_ROBIN: _get_env_for_spider_round_robin_scheduler,
+}
+
+
+def _get_env_for_spider_scheduler_policy(scheduler: SpiderScheduler) -> EnvVarsDict:
```
junhaoliao: how about renaming as `_get_spider_scheduler_policy_env_vars`

components/package-template/src/etc/clp-config.template.json.yaml
junhaoliao: we should also update `components/package-template/src/etc/clp-config.template.text.yaml` to write 

```yaml
#spider: null
#compression_coordinator: null
```

and state they are unsupported with CLP Text.

components/clp-package-utils/clp_package_utils/controller.py
```diff
@@ -1362,6 +1503,73 @@ def _chown_recursively(
...
+    policy = SpiderScheduler.DEFAULT_POLICY if scheduler.policy is None else scheduler.policy
+    get_env = _SPIDER_SCHEDULER_POLICY_ENV_GETTERS.get(policy)
+    if get_env is None:
+        msg = f"No environment getter is registered for Spider scheduler policy `{policy}`."
+        raise NotImplementedError(msg)
+
+    return get_env(scheduler)
+
+
+def _optional_str(value: int | pathlib.Path | None) -> str | None:
```
junhaoliao: the name `_optional_str` sounds like it accepts an optional string, rather than converting some optional value to an optional string. how about renaming it as `_str_or_none`?

junhaoliao (changes requested):
the title lgtm

components/clp-package-utils/clp_package_utils/controller.py
20001020ycx: Correct, good catch.

components/clp-py-utils/clp_py_utils/clp_config.py
20001020ycx: Emmm, bound checking is already handled by Rust. If we were to do bound checking, shouldn't we refer back to Spider's type, which makes this code bound to Spider code, which is not good for maintainence? I think as long as we are not mixing up the type here should be fine, let me know if there's more concern that I didnt catch

junhaoliao: right, it could be too much maintenance overhead now i think about it again. fine to defer this

junhaoliao (approved):

components/clp-py-utils/clp_py_utils/clp_config.py
20001020ycx: Ditto as discussed in https://github.com/y-scope/clp/pull/2491#discussion_r3896727244

PR #2497, opened 2026-08-24
perf(clp-package): Reduce overhead of retrieving Celery compression task results (fixes #2387).

### LINKED ISSUE #2387: Compression scheduler: poll_running_jobs sleeps ~0.5s per in-flight job, delaying dispatch of new jobs
## Summary

The compression scheduler's main loop is single threaded and alternates between `poll_running_jobs()` and `search_and_schedule_new_tasks()`. `poll_running_jobs()` calls `ResultHandle.get_result()` serially for every in-flight job. For any job whose tasks are not complete yet, the call sleeps 0.5s (the celery default poll interval) before returning `None`, so one scheduler cycle takes roughly 0.5s x N for N in-flight jobs, and new PENDING jobs are not dispatched until the current cycle finishes.

On our production deployment with 150-190 jobs in flight, scheduler cycles take 75-95s and dispatch latency (`creation_time` -> `start_time`) is 140-330s. Jobs sit undispatched for 2-5 minutes while the celery queue is empty and workers are idle. The delay scales linearly with the number of concurrent jobs, so deployments with many datasets are hit hardest.

## Environment

- clp-package 0.12.0 lineage, CLP_S storage engine
- CeleryTaskManager with Redis result backend, RabbitMQ broker
- celery 5.6.3, kombu 5.6.2, redis-py 6.4.0
- ~50 datasets ingesting concurrently, 300 celery workers
- `jobs_poll_delay: 0.1`, `max_concurrent_tasks_per_job: 0`

## Root cause

`components/job-orchestration/job_orchestration/scheduler/compress/task_manager/celery_task_manager.py`:

```python
def get_result(self, timeout: float = 0.1) -> list[CompressionTaskResult] | None:
    try:
        results = self._celery_result.get(timeout=timeout)
        return [CompressionTaskResult.model_validate(res) for res in results]
    except celery.exceptions.TimeoutError:
        return None
```

`timeout=0.1` does not bound the wall time. With celery 5.6.3 and the Redis result backend the call path is:

`GroupResult.get(timeout=0.1)` -> `ResultSet.join_native` (Redis has `supports_native_join = True`) -> `iter_native` -> `AsyncBackendMixin._wait_for_pending` -> `BaseResultConsumer._wait_for_pending` -> Redis `on_wait_for_pending` -> `result._iter_meta()` -> `BaseKeyValueStoreBackend.get_many(ids, timeout=0.1, interval=0.5, max_iterations=1)`.

In `celery/backends/base.py::get_many()`, when any task in the group is not READY after the `mget`:

```python
if timeout and iterations * interval >= timeout:   # 0 * 0.5 >= 0.1 is False on the first pass
    raise TimeoutError(f'Operation timed out ({timeout})')
if on_interval:
    on_interval()
time.sleep(interval)  # don't busy loop.
iterations += 1
if max_iterations and iterations >= max_iterations:
    break
```

The timeout check compares `iterations * interval` rather than elapsed time and runs before the sleep with `iterations == 0`, so it never fires on the first pass; the 0.5s sleep always happens. `interval` is the default from `ResultSet.get()` and is never overridden by the caller. After `get_many` returns, `Drainer.drain_events_until` raises `socket.timeout`, which `_wait_for_pending` converts to the `celery.exceptions.TimeoutError` that `get_result()` catches. Total cost per incomplete job is the 0.5s sleep plus one Redis `mget` plus a short pubsub drain, which matches the ~0.504s spacing in our scheduler logs.

`compression_scheduler.py::poll_running_jobs()` then does this serially:

```python
for job_id, job in scheduled_jobs.items():
    returned_results = job.result_handle.get_result()
    if returned_results is None:
        continue
```

Cycle cost is about 0.5s per incomplete job. `search_and_schedule_new_tasks()` only runs between sweeps, so dispatch latency for a new job is 1-3 full cycles.

## Evidence

- Scheduler INFO logs show job completion processing spaced at almost exactly the poll interval: `05:21:43.446 -> 05:21:43.950 -> 05:21:45.214 -> 05:21:45.719 -> 05:21:46.224` (~0.50s per still-running job in between).
- `compression_jobs` rows show `TIMESTAMPDIFF(SECOND, creation_time, start_time)` of 141/234/328s, stepping in ~95s increments, i.e. quantized by scheduler cycle length (190 in-flight jobs x 0.5s).
- The RabbitMQ compression queue is empty and consumers are idle during these waits, so this is not worker capacity. No scheduler DB statement exceeded 1s in sampled processlist output.

## Impact

- Adds 2-5 minutes of ingestion latency to every job, regardless of job size.
- Completion bookkeeping (SUCCEEDED status, `duration`) lags by the same amount, which skews job latency metrics.
- The effect compounds during backlog catch-up: more jobs in flight means longer cycles means longer dispatch waits.
- Configurations that produce smaller, more frequent jobs increase N and make the cycle proportionally worse.

## Suggested fix

Check readiness before collecting results:

```python
def get_result(self, timeout: float = 0.1) -> list[CompressionTaskResult] | None:
    if not self._celery_result.ready():
        return None
    results = self._celery_result.get(timeout=timeout)
    return [CompressionTaskResult.model_validate(res) for res in results]
```

`GroupResult.ready()` is a plain backend read with no sleep (sub-ms per task on Redis). The poll sweep over ~190 jobs drops from ~95s to milliseconds and dispatch latency collapses to roughly `jobs_poll_delay`. The `get()` after a successful `ready()` returns immediately. Passing `interval=0.01` to `.get()` would also work, but `ready()` avoids raising and catching a `TimeoutError` per job on every cycle.

Running polling and dispatch on separate threads would remove the coupling entirely, but it needs a second DB connection (`DbContext` wraps a connection that is not thread safe), locking around the shared `scheduled_jobs` dict, and changes to the SIGTERM drain logic. With the `ready()` check the sweep is milliseconds, so I don't think that's warranted.

Note that when `max_concurrent_tasks_per_job > 0`, `_dispatch_next_task_batch()` is gated on the same sweep, so intra-job batch turnaround inherits the same latency. The `ready()` check fixes that path as well.

## Repro

1. Run enough concurrent compression jobs that 100+ are in RUNNING simultaneously (many datasets or long-running jobs).
2. Observe `creation_time -> start_time` on newly inserted jobs grow linearly with the in-flight job count (~0.5s each) despite an empty broker queue and idle workers.
3. Add the `ready()` check; dispatch latency drops to ~`jobs_poll_delay`.

gibber9809:
# Description

This PR significantly reduces the overhead of retrieving celery compression task results by copying the trick we used for mitigating a similar problem on the search side in #1899; by checking `result.ready()` before calling `result.get()` we significantly reduce overhead in polling cycles where tasks are not yet complete, and by calling `result.get()` with a low `interval` argument we reduce the time that celery spends polling for the result inside of `get()` (compared to the default 0.5 second polling interval).

Based on some small-scale benchmarks on old xeon servers, this reduces the overhead of checking/retrieving task results from ~1 second when the result is not ready yet and ~1 second when the result is ready, to <1ms when the result is not yet ready and ~5ms when the result is ready.

This PR introduces some minor behaviour changes in `CeleryTaskManager::ResultHandle::get_result` to make it line up with the behaviour of `try_getting_task_result` in `query_scheduler.py`. Specifically, since we're now checking `ready()` before calling `get()`, we now treat a `celery.exceptions.TimeoutError` as an actual exception and re-raise it; previously we were effectively using that exception to check for result readiness, but now that we check `ready()` this exception indicates something is actually wrong with the system.

# Checklist

  breaking change.

# Validation performed

* Added some ad-hoc instrumentation to validate that the overhead of polling for/retrieving results is _significantly_ reduced by this change
* Validated that package compression flow through the cli seems to still work as expected

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task result handling when asynchronous work is still in progress.
  * Increased the default wait time for retrieving task results to 10 seconds.
  * Added more responsive polling while waiting for results.
  * Timeout errors are now logged and reported instead of being silently treated as unavailable results.

### DIFF
--- a/components/job-orchestration/job_orchestration/scheduler/compress/task_manager/celery_task_manager.py
+++ b/components/job-orchestration/job_orchestration/scheduler/compress/task_manager/celery_task_manager.py
@@ -6,13 +6,15 @@ from typing import Any
 import celery
 
 from job_orchestration.executor.compress.celery_compress import compress
-from job_orchestration.scheduler.compress.task_manager.task_manager import TaskManager
+from job_orchestration.scheduler.compress.task_manager.task_manager import (
+    TASK_GET_RESULT_DEFAULT_TIMEOUT_SECONDS,
+    TaskManager,
+)
 from job_orchestration.scheduler.task_result import CompressionTaskResult
 
 logger = logging.getLogger(__name__)
 
-TASK_RESULT_GET_INTERVAL_SECONDS = 0.005
-TASK_RESULT_DEFAULT_GET_TIMEOUT_SECONDS = 10
+TASK_GET_RESULT_INTERVAL_SECONDS = 0.005
 
 
 class CeleryTaskManager(TaskManager):
@@ -21,13 +23,13 @@ class CeleryTaskManager(TaskManager):
             self._celery_result: celery.result.GroupResult = celery_result
 
         def get_result(
-            self, timeout: float = TASK_RESULT_DEFAULT_GET_TIMEOUT_SECONDS
+            self, timeout: float = TASK_GET_RESULT_DEFAULT_TIMEOUT_SECONDS
         ) -> list[CompressionTaskResult] | None:
             if not self._celery_result.ready():
                 return None
             try:
                 results = self._celery_result.get(
-                    timeout=timeout, interval=TASK_RESULT_GET_INTERVAL_SECONDS
+                    timeout=timeout, interval=TASK_GET_RESULT_INTERVAL_SECONDS
                 )
                 return [CompressionTaskResult.model_validate(res) for res in results]
             except celery.exceptions.TimeoutError:
--- a/components/job-orchestration/job_orchestration/scheduler/compress/task_manager/celery_task_manager.py
+++ b/components/job-orchestration/job_orchestration/scheduler/compress/task_manager/celery_task_manager.py
@@ -11,18 +11,25 @@ from job_orchestration.scheduler.task_result import CompressionTaskResult
 
 logger = logging.getLogger(__name__)
 
+TASK_RESULT_GET_INTERVAL_SECONDS = 0.005
+TASK_RESULT_DEFAULT_GET_TIMEOUT_SECONDS = 10
 
 class CeleryTaskManager(TaskManager):
     class ResultHandle(TaskManager.ResultHandle):
         def __init__(self, celery_result: celery.result.GroupResult) -> None:
             self._celery_result: celery.result.GroupResult = celery_result
 
-        def get_result(self, timeout: float = 0.1) -> list[CompressionTaskResult] | None:
+        def get_result(self, timeout: float = TASK_RESULT_DEFAULT_GET_TIMEOUT_SECONDS) -> list[CompressionTaskResult] | None:
+            if not self._celery_result.ready():
+                return None
             try:
-                results = self._celery_result.get(timeout=timeout)
+                results = self._celery_result.get(
+                    timeout=timeout, interval=TASK_RESULT_GET_INTERVAL_SECONDS
+                )
                 return [CompressionTaskResult.model_validate(res) for res in results]
             except celery.exceptions.TimeoutError:
-                return None
+                logger.exception("Timed out waiting for task result.")
+                raise
             except celery.exceptions.SoftTimeLimitExceeded:
                 logger.exception("Compression task exceeded soft time limit.")
                 raise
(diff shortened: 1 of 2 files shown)


components/job-orchestration/job_orchestration/scheduler/compress/task_manager/celery_task_manager.py
```diff
@@ -6,23 +6,35 @@
...
+                return None
             try:
-                results = self._celery_result.get(timeout=timeout)
+                results = self._celery_result.get(
+                    timeout=timeout, interval=TASK_GET_RESULT_INTERVAL_SECONDS
+                )
                 return [CompressionTaskResult.model_validate(res) for res in results]
             except celery.exceptions.TimeoutError:
-                return None
+                logger.exception("Timed out waiting for task result.")
```
kirkrodrigues:
```suggestion
                logger.exception("Compression task timed out waiting for result.")
```

For consistency.

sitaowang1998 (dismissed):
LGTM

sitaowang1998 (approved):

kirkrodrigues (approved):
Capitalized "celery" in the title.

PR #2503, opened 2026-08-27
feat(clp-tdl-package): Add task signatures and shared I/O types for non-aggregation queries.

Bill-hbrhbr:
# Description

This PR establishes the shared TDL contract for aggregation-less CLP-S queries. It adds the serializable `ClpSQueryOption` and `QueryTaskOutput` types, defines and registers the `search::clp_s_search_to_results_cache` task signature, and leaves the task body as a placeholder for a subsequent implementation. 

Query execution, result-cache upserts, and coordinator integration are planned for subsequent PRs.

# Checklist

  breaking change.

# Validation performed

No behavioral validation was performed because this PR defines an interface-only contract and does not include an executable query-task implementation. The contract serves as the integration boundary for subsequent worker-side TDL implementation and coordinator-side graph construction.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring CLP-S queries with query text, result limits, optional timestamp ranges, and case-sensitivity settings.
  * Added query job task registration for CLP-S archive searches.
  * Added query output details identifying the relevant dataset and archive.
  * Exposed query-related task options and results through the task package.


components/clp-rust-utils/src/task_io/query.rs
```diff
@@ -0,0 +1,95 @@
+//! Protocol types exchanged with the Spider (Huntsman) tasks that run CLP query jobs.
+
+use std::num::NonZeroU32;
+
+use serde::Deserialize;
+use serde::Serialize;
+
+/// `clp-s` tuning and engine options for a query job.
```
LinZhihao-723:
```suggestion
/// `clp-s` options for a query job.
```
Unlike compression, these are not engine tuning options.

components/clp-rust-utils/src/task_io/query.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// The query string passed positionally to `clp-s`.
+    pub query_string: String,
+
+    /// The maximum number of results retained by one archive-query invocation.
+    pub max_num_results: NonZeroU32,
+
+    /// The inclusive lower timestamp bound (`--tge`), in Unix epoch microseconds.
+    pub begin_timestamp: Option<i64>,
+
+    /// The inclusive upper timestamp bound (`--tle`), in Unix epoch microseconds.
```
LinZhihao-723:
```suggestion
    /// The inclusive lower timestamp bound (`--tge`), in Unix epoch milliseconds.
    pub begin_timestamp: Option<i64>,

    /// The inclusive upper timestamp bound (`--tle`), in Unix epoch milliseconds.
```
Double-checked that they're actually milliseconds.

components/clp-rust-utils/src/task_io/query.rs
```diff
@@ -0,0 +1,95 @@
...
+/// This output only identifies the archive whose query invocation completed; it does not
+/// finalize the query job.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct QueryTaskOutput {
+    /// The resolved dataset containing the queried archive.
+    pub dataset: String,
+
+    /// The identifier of the queried archive.
+    pub archive_id: String,
+}
```
LinZhihao-723: Why do we need this? From `query-coordinator`'s perspective, it only needs to care whether the job completes or not, right?
I prefer not to store any outputs in the spider for optimal performance.

components/clp-rust-utils/src/task_io/query.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#[cfg(test)]
+mod tests {
```
LinZhihao-723: I don't think we need the round-trip serde tests: realistically speaking u can't assert this for all the types. As long as they can derive `Serialize` and `Deserialize` it should be good.

components/clp-tdl-package/src/task/query/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+
+/// Queries one CLP-S archive and writes its matches directly to the results cache.
+#[task(name = "query::clp_s_query_to_results_cache")]
+pub(crate) fn clp_s_query_to_results_cache_task(
+    ctx: TaskContext,
+    query_job_id: i32,
+    clp_s_query_option: ClpSQueryOption,
+    dataset: String,
+    archive_id: String,
+) -> Result<QueryTaskOutput, TdlError> {
```
LinZhihao-723: As discussed, the output is meaningless.
```suggestion
) -> Result<(), TdlError> {
```

components/clp-tdl-package/src/task/query/mod.rs
```diff
@@ -0,0 +1,20 @@
+//! The query-task signatures registered with Spider.
+
+use clp_rust_utils::task_io::query::ClpSQueryOption;
+use clp_rust_utils::task_io::query::QueryTaskOutput;
+use spider_tdl::TaskContext;
+use spider_tdl::TdlError;
+use spider_tdl::task;
+
+/// Queries one CLP-S archive and writes its matches directly to the results cache.
```
LinZhihao-723:
```suggestion
```
Please check the existing task signatures in compression coordinator: we don't document the task definition. Also, this looks like a random-added comment that doesn't match the formal docstring format.

components/clp-rust-utils/src/task_io/query.rs
```diff
@@ -0,0 +1,95 @@
...
+            archive_id: "archive-id".to_owned(),
+        };
+
+        let serialized = rmp_serde::to_vec(&expected).expect("task output should serialize");
+        let actual: QueryTaskOutput =
+            rmp_serde::from_slice(&serialized).expect("task output should deserialize");
+
+        assert_eq!(expected, actual);
+    }
+}
```
LinZhihao-723: I don't think we need the round-trip serde tests: realistically speaking u can't assert this for all the types. We usually add round-trip assertions to check certian serde options, for example, field renaming. This type leaves everything by default. As long as they can derive `Serialize` and `Deserialize` and get compiled we should be good.

```suggestion
```

components/clp-tdl-package/src/task/query/mod.rs
LinZhihao-723: Also, `CLP` is a general term for the system. `CLP-S` is not. You should either use `CLP` to refer to any CLP package in general, or `clp-s` for the structured engine.

components/clp-tdl-package/src/task/query/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+#[task(name = "query::clp_s_query_to_results_cache")]
+pub(crate) fn clp_s_query_to_results_cache_task(
+    ctx: TaskContext,
+    query_job_id: i32,
+    clp_s_query_option: ClpSQueryOption,
+    dataset: String,
+    archive_id: String,
+) -> Result<QueryTaskOutput, TdlError> {
+    let _ = (ctx, query_job_id, clp_s_query_option, dataset, archive_id);
```
LinZhihao-723:
```suggestion
    _ctx: TaskContext,
    _query_job_id: i32,
    _clp_s_query_option: ClpSQueryOption,
    _dataset: String,
    _archive_id: String,
) -> Result<(), TdlError> {
```
In Rust, add the `_` prefix to explicitly mark a symbol as `unused`.

LinZhihao-723 (changes requested):

components/clp-tdl-package/src/task/query/mod.rs
```diff
@@ -0,0 +1,20 @@
+//! The query-task signatures registered with Spider.
+
+use clp_rust_utils::task_io::query::ClpSQueryOption;
+use clp_rust_utils::task_io::query::QueryTaskOutput;
+use spider_tdl::TaskContext;
+use spider_tdl::TdlError;
+use spider_tdl::task;
+
+/// Queries one CLP-S archive and writes its matches directly to the results cache.
+#[task(name = "query::clp_s_query_to_results_cache")]
```
LinZhihao-723: Sorry for missing this in the previous pass:
```suggestion
#[task(name = "query::clp_s_search")]
```
- Let's be explicit that this is a search task.
- We don't need to explain where this archive is coming from. Instead, we will add an enum to allow the task to select the output handle.

LinZhihao-723 (changes requested):

components/clp-tdl-package/src/task/query/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+pub(crate) fn clp_s_query_to_results_cache_task(
+    ctx: TaskContext,
+    query_job_id: i32,
```
LinZhihao-723: Like compression coordinator, we should define a dedicated type for the query job ID in clp-rust-utils.

LinZhihao-723 (changes requested):

components/clp-tdl-package/src/task/query/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+    clp_s_query_option: ClpSQueryOption,
+    dataset: String,
```
LinZhihao-723:
```suggestion
    dataset: Option<String>,
```
The dataset should be optional, no? See https://github.com/y-scope/clp/blob/af0c9276478139a30f7c6c8dfd681d883e5c33bb/components/clp-tdl-package/src/task/compression/mod.rs#L17

LinZhihao-723 (changes requested):

components/clp-rust-utils/src/task_io/query.rs
```diff
@@ -0,0 +1,40 @@
...
+use serde::Serialize;
+
+/// `clp-s` options for a query job.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct ClpSQueryOption {
+    /// The query string passed positionally to `clp-s`.
+    pub query_string: String,
+
+    /// The maximum number of results retained by one archive-query invocation.
+    pub max_num_results: NonZeroU32,
```
LinZhihao-723: This should be optional; otherwise, we can't express an unlimited case.

```suggestion
    pub max_num_results: Option<NonZeroU32>,
```

LinZhihao-723 (changes requested):

components/clp-rust-utils/src/task_io/query.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    /// The inclusive lower timestamp bound (`--tge`), in Unix epoch microseconds.
+    pub begin_timestamp: Option<i64>,
+
```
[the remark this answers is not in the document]
Bill-hbrhbr: Keep todo instead of returning error.

components/clp-rust-utils/src/clp_config/package/config.rs
```diff
@@ -196,6 +196,12 @@ impl Database {
     }
 }
 
+/// The default maximum number of results retained for a query.
+///
+/// Mirror of `clp_py_utils.clp_config.ApiServer.default_max_num_query_results`. Must be kept in
+/// sync.
+pub const DEFAULT_MAX_NUM_QUERY_RESULTS: u32 = 1000;
+
```
LinZhihao-723: According to the symbol ordering rule (https://github.com/LinZhihao-723/claude-instruction/blob/main/rust/coding-style.md#symbol-ordering), constants should come before all types.

components/clp-rust-utils/src/job_config/search.rs
```diff
@@ -3,6 +3,8 @@ use num_enum::TryFromPrimitive;
 use serde::Deserialize;
 use serde::Serialize;
 
+pub type QueryJobId = i32;
+
 pub const QUERY_JOBS_TABLE_NAME: &str = "query_jobs";
```
LinZhihao-723: Similarly, constants come first. https://github.com/LinZhihao-723/claude-instruction/blob/main/rust/coding-style.md#symbol-ordering

components/clp-rust-utils/src/task_io/query.rs
```diff
@@ -0,0 +1,34 @@
...
+use serde::Serialize;
+
+/// `clp-s` options for a query job.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct ClpSQueryOption {
+    /// The query string passed positionally to `clp-s`.
+    pub query_string: NonEmptyString,
+
+    /// The per-archive result limit. When absent,
+    /// [`crate::clp_config::package::config::DEFAULT_MAX_NUM_QUERY_RESULTS`] is used.
```
LinZhihao-723: This is actually not what we should expect. Nowadays, both celery side and clp-s side set the default value. I don't think this is a good design. In my search task implementation, I default this to clp-s' default (which is also 1000), which makes the default behavior a single-source decision inside clp-s.
You should rewrite it to say this defaults to clp-s' setting.

LinZhihao-723 (changes requested):

components/clp-tdl-package/src/task/query/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+#[task(name = "query::clp_s_search")]
+pub(crate) fn clp_s_search_task(
+    _ctx: TaskContext,
+    _query_job_id: QueryJobId,
+    _clp_s_query_option: ClpSQueryOption,
+    _dataset: Option<NonEmptyString>,
+    _archive_id: NonEmptyString,
+    _query_output_handle: QueryOutputHandle,
+) -> Result<(), spider_tdl::TdlError> {
+    todo!("query task is not implemented")
```
LinZhihao-723:
```suggestion
    todo!("clp-s search task is not implemented")
```

components/clp-rust-utils/src/task_io/query.rs
```diff
@@ -0,0 +1,34 @@
...
+    /// Whether `clp-s` performs a case-insensitive search.
+    pub ignore_case: bool,
+}
+
+/// The output handler that `clp-s` writes a query task's results to.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub enum QueryOutputHandle {
+    /// Writes results to the results-cache collection named after the query job's ID.
+    ResultsCache,
+}
```
LinZhihao-723: 1. Can we rename to `OutputHandle`? "Query" is redundant in this type name.
2. Let's empty the variant definition first. I will do it in the search task PR.

```suggestion
/// The output handler that `clp-s` writes a query task's results to.
#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
pub enum QueryOutputHandle {
}
```

LinZhihao-723 (changes requested):

LinZhihao-723 (approved):
Please merge after v0.14.0 release is made.

PR #2509, opened 2026-08-31
feat(clp-s)!: Deduplicate cached search results using compound MongoDB IDs.

sitaowang1998:
# Description

This PR removes duplicated entries in MongoDB by:
- Using `{archive_id, log_event_idx}` as the MongoDB result document `_id` for `clp-s`.
- Using `{origin_file_id, log_event_idx}` as the MongoDB result document `_id` for `clo`.
- Preventing duplicate result documents when searches are retried and treating a query as success if all insertions either succeeds or fails with duplicate key error.
- Removing redundant top-level `archive_id`/`origin_file_id` and `log_event_idx` fields.
- Expiring results based on query completion time, calculated from MariaDB’s existing `start_time` and `duration`. When the `start_time` and `duration` are missing, removed any complete job in terminated state and `creation_timestamp` before expire cutoff.
- Updating WebUI and MCP result parsing for the new `_id` structure. No backward compatibility support is provided. WebUI supports both `clo` and `clp-s` result, while MCP supports only `clp-s` result.

> [!Note]
> This PR is a breaking change because the MongoDB schema changes.

# Checklist

  breaking change.

# Validation performed

* [ ] GitHub workflows pass.
* [x] Repeated searches do not insert duplicate MongoDB documents for both `clp-s` and `clp-o`.
* [x] Runs end-to-end compression and search job on `clp-s` to confirm that:
  * Repeated search do not insert duplicate MongDB documents.
  * Result cache is deleted by garbage collection based on `start_time + duration`.
  * WebUI raw-results endpoint streamed expected log events.

I am unable to run end-to-end search with MCP server. After I ran `search_by_kql`, which created a query that was verified to complete successfully, the response was "Please call `get_instructions()` first", which I did before calling `search_by_kql`.

# Performance evaluation

This is reported by @LinZhihao-723. The evaluation confirmed that the compound `_id` costs nothing measurable at any concurrency tested.

Measured 2025-09-03 (commit [3165049](https://github.com/y-scope/clp/pull/2509/commits/31650498ec064355c011ccb255bac709ab5ca683)). Two stock builds replay 128 pre-built archives into a fresh MongoDB collection from a pool of exactly N OS threads: `BASELINE` = merge-base `2a5fbee5`, server-assigned ObjectId, `PR` = the evaluated commit. Query `level: "FATAL"`, `--max-num-results 2500 --batch-size 1000`, 320,000 documents per pass; 1 discarded warmup + 5 measured reps per (arm, N); **no secondary unique index anywhere**; MongoDB 8.0.21 standalone; i9-14900K / 32 threads / 47 GiB / WSL2. All measured cells held exactly 320,000 documents.

## PR versus baseline

| N | BASELINE wall (s) | PR wall (s) | Δ wall | BASELINE e2e mean (s) | PR e2e mean (s) | Δ e2e | per-rep ranges |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 1 | 16.3142 | 16.3512 | **+0.2%** | 0.1280 | 0.1278 | −0.2% | overlap |
| 2 | 8.1691 | 8.2075 | **+0.5%** | 0.1286 | 0.1284 | −0.2% | overlap |
| 4 | 4.3009 | 4.2349 | **−1.5%** | 0.1337 | 0.1328 | −0.7% | overlap |
| 8 | 2.4526 | 2.4679 | **+0.6%** | 0.1526 | 0.1532 | +0.4% | overlap |
| 16 | 1.7886 | 1.8101 | **+1.2%** | 0.2181 | 0.2319 | +6.3% | overlap |

The five measured walls per cell overlap in **all five cells**. The largest delta (−1.5% at N=4, in the PR's favour) is smaller than the spread of either arm's own five reps at that N, and the sign is not consistent across N. Scalability curves are identical (BASELINE 1.00/2.00/3.79/6.65/9.12×, PR 1.00/1.99/3.86/6.63/9.03×), with the same knee between N=8 and N=16: the ceiling is the shared MongoDB write path, not an `_id` effect.

## `insert` vs. `upsert`

See https://github.com/y-scope/clp/pull/2509#discussion_r3960936653.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search-result handling across current and legacy result formats.
  * Corrected event-index mapping so result links and displayed messages open the intended log events.
  * Improved reliability when storing duplicate search results during retries.
  * Enhanced error handling and logging for failed result-storage operations.

* **Maintenance**
  * Search-result cleanup now uses job completion times and retention settings to remove expired data more accurately.
  * Standardized search-result identifiers for more consistent display and navigation.

### DIFF
--- a/components/clp-mcp-server/clp_mcp_server/clp_connector.py
+++ b/components/clp-mcp-server/clp_mcp_server/clp_connector.py
@@ -160,8 +160,7 @@ class ClpConnector:
         collection = self._results_cache[str(query_id)]
         results = []
 
-        async for raw_doc in collection.find({}, limit=SEARCH_MAX_NUM_RESULTS):
-            doc = dict(raw_doc)
+        async for doc in collection.find({}, limit=SEARCH_MAX_NUM_RESULTS):
             result_id = doc["_id"]
             doc["archive_id"] = result_id["archive_id"]
             doc["log_event_idx"] = result_id["log_event_idx"]
--- a/components/clp-mcp-server/clp_mcp_server/clp_connector.py
+++ b/components/clp-mcp-server/clp_mcp_server/clp_connector.py
@@ -160,12 +160,16 @@ class ClpConnector:
         collection = self._results_cache[str(query_id)]
         results = []
 
-        async for doc in collection.find({}, limit=SEARCH_MAX_NUM_RESULTS):
+        async for raw_doc in collection.find({}, limit=SEARCH_MAX_NUM_RESULTS):
+            doc = dict(raw_doc)
+            result_id = doc["_id"]
+            doc["archive_id"] = result_id["archive_id"]
+            doc["log_event_idx"] = result_id["log_event_idx"]
             doc["link"] = (
                 f"{self._webui_addr}/streamFile?type=json"
                 f"&streamId={doc['archive_id']}"
                 f"&dataset={CLP_DEFAULT_DATASET_NAME}"
-                f"&logEventIdx={doc['log_event_ix']}"
+                f"&logEventIdx={doc['log_event_idx']}"
             )
             doc["_id"] = None
             results.append(doc)
--- a/components/clp-mcp-server/clp_mcp_server/clp_connector.py
+++ b/components/clp-mcp-server/clp_mcp_server/clp_connector.py
@@ -163,11 +163,8 @@ class ClpConnector:
         async for raw_doc in collection.find({}, limit=SEARCH_MAX_NUM_RESULTS):
             doc = dict(raw_doc)
             result_id = doc["_id"]
-            if isinstance(result_id, dict):
-                doc["archive_id"] = result_id["archive_id"]
-                doc["log_event_idx"] = result_id["log_event_idx"]
-            else:
-                doc["log_event_idx"] = doc.pop("log_event_ix")
+            doc["archive_id"] = result_id["archive_id"]
+            doc["log_event_idx"] = result_id["log_event_idx"]
             doc["link"] = (
                 f"{self._webui_addr}/streamFile?type=json"
                 f"&streamId={doc['archive_id']}"
--- a/components/clp-mcp-server/clp_mcp_server/clp_connector.py
+++ b/components/clp-mcp-server/clp_mcp_server/clp_connector.py
@@ -163,8 +163,11 @@ class ClpConnector:
         async for raw_doc in collection.find({}, limit=SEARCH_MAX_NUM_RESULTS):
             doc = dict(raw_doc)
             result_id = doc["_id"]
-            doc["archive_id"] = result_id["archive_id"]
-            doc["log_event_idx"] = result_id["log_event_idx"]
+            if isinstance(result_id, dict):
+                doc["archive_id"] = result_id["archive_id"]
+                doc["log_event_idx"] = result_id["log_event_idx"]
+            else:
+                doc["log_event_idx"] = doc.pop("log_event_ix")
             doc["link"] = (
                 f"{self._webui_addr}/streamFile?type=json"
                 f"&streamId={doc['archive_id']}"
--- a/components/core/src/clp_s/MongoDBUtils.cpp
+++ b/components/core/src/clp_s/MongoDBUtils.cpp
@@ -1,6 +1,5 @@
 #include "MongoDBUtils.hpp"
 
-#include <algorithm>
 #include <cstdint>
 
 #include <bsoncxx/types.hpp>
@@ -79,8 +78,10 @@ constexpr int32_t cDuplicateKeyErrorCode{11'000};
 }
 }  // namespace
 
-auto contains_only_duplicate_key_write_errors(mongocxx::bulk_write_exception const& exception)
-        -> bool {
+auto contains_only_duplicate_key_write_errors(
+        mongocxx::bulk_write_exception const& exception,
+        size_t num_documents
+) -> bool {
     auto const& raw_server_error = exception.raw_server_error();
     if (false == raw_server_error.has_value()) {
         return false;
@@ -102,6 +103,26 @@ auto contains_only_duplicate_key_write_errors(mongocxx::bulk_write_exception con
     if (write_errors.empty()) {
         return false;
     }
-    return std::all_of(write_errors.begin(), write_errors.end(), is_duplicate_key_write_error);
+
+    size_t num_write_errors{0};
+    for (auto const& write_error : write_errors) {
+        if (false == is_duplicate_key_write_error(write_error)) {
+            return false;
+        }
+        ++num_write_errors;
+    }
+
+    auto const num_inserted_element = reply["nInserted"];
+    if (false == static_cast<bool>(num_inserted_element)
+        || bsoncxx::type::k_int32 != num_inserted_element.type())
+    {
+        return false;
+    }
+    auto const num_inserted = num_inserted_element.get_int32().value;
+    if (num_inserted < 0) {
+        return false;
+    }
+
+    return static_cast<size_t>(num_inserted) + num_write_errors == num_documents;
 }
 }  // namespace clp_s
--- a/components/core/src/clp_s/MongoDBUtils.cpp
+++ b/components/core/src/clp_s/MongoDBUtils.cpp
@@ -1,0 +1,106 @@
+#include "MongoDBUtils.hpp"
+
+#include <algorithm>
+
+#include <bsoncxx/types.hpp>
+#include <mongocxx/exception/bulk_write_exception.hpp>
+
+namespace clp_s {
+namespace {
+constexpr int32_t cDuplicateKeyErrorCode{11'000};
+
+/**
+ * Checks whether an aggregated bulk-write reply contains any command errors.
+ * @param reply The raw MongoDB bulk-write reply.
+ * @return true if the reply contains a command error, false otherwise.
+ */
+[[nodiscard]] auto has_command_errors(bsoncxx::document::view const& reply) -> bool;
+
+/**
+ * Checks whether a bulk-write reply contains any write-concern errors.
+ * @param reply The raw MongoDB bulk-write reply.
+ * @return true if the reply contains a write-concern error, false otherwise.
+ */
+[[nodiscard]] auto has_write_concern_errors(bsoncxx::document::view const& reply) -> bool;
+
+/**
+ * Checks whether an entry from a bulk-write reply's `writeErrors` array is a duplicate-key error.
+ * @param write_error The write-error entry to inspect.
+ * @return true if the entry has MongoDB's duplicate-key error code, false otherwise.
+ */
+[[nodiscard]] auto is_duplicate_key_write_error(bsoncxx::array::element const& write_error) -> bool;
+
+[[nodiscard]] auto has_command_errors(bsoncxx::document::view const& reply) -> bool {
+    auto const errors_element = reply["errorReplies"];
+    if (false == static_cast<bool>(errors_element)) {
+        return false;
+    }
+    if (bsoncxx::type::k_array != errors_element.type()) {
+        return true;
+    }
+    auto const errors = errors_element.get_array().value;
+    return errors.begin() != errors.end();
+}
+
+[[nodiscard]] auto has_write_concern_errors(bsoncxx::document::view const& reply) -> bool {
+    if (static_cast<bool>(reply["writeConcernError"])) {
+        return true;
+    }
+
+    auto const errors_element = reply["writeConcernErrors"];
+    if (false == static_cast<bool>(errors_element)) {
+        return false;
+    }
+    if (bsoncxx::type::k_array != errors_element.type()) {
+        return true;
+    }
+    auto const errors = errors_element.get_array().value;
+    return errors.begin() != errors.end();
+}
+
+[[nodiscard]] auto is_duplicate_key_write_error(bsoncxx::array::element const& write_error)
+        -> bool {
+    if (bsoncxx::type::k_document != write_error.type()) {
+        return false;
+    }
+
+    auto const code = write_error.get_document().value["code"];
+    if (false == static_cast<bool>(code)) {
+        return false;
+    }
+    if (bsoncxx::type::k_int32 == code.type()) {
+        return cDuplicateKeyErrorCode == code.get_int32().value;
+    }
+    if (bsoncxx::type::k_int64 == code.type()) {
+        return cDuplicateKeyErrorCode == code.get_int64().value;
+    }
+    return false;
+}
+}  // namespace
+
+auto contains_only_duplicate_key_write_errors(mongocxx::bulk_write_exception const& exception)
+        -> bool {
+    auto const& raw_server_error = exception.raw_server_error();
+    if (false == raw_server_error.has_value()) {
+        return false;
+    }
+
+    auto const reply = raw_server_error->view();
+    if (has_command_errors(reply) || has_write_concern_errors(reply)) {
+        return false;
+    }
+
+    auto const write_errors_element = reply["writeErrors"];
+    if (false == static_cast<bool>(write_errors_element)
+        || bsoncxx::type::k_array != write_errors_element.type())
+    {
+        return false;
+    }
+
+    auto const write_errors = write_errors_element.get_array().value;
+    if (write_errors.begin() == write_errors.end()) {
+        return false;
+    }
+    return std::all_of(write_errors.begin(), write_errors.end(), is_duplicate_key_write_error);
+}
+}  // namespace clp_s
--- a/components/core/src/clp_s/MongoDBUtils.cpp
+++ b/components/core/src/clp_s/MongoDBUtils.cpp
@@ -39,7 +39,7 @@ constexpr int32_t cDuplicateKeyErrorCode{11'000};
         return true;
     }
     auto const errors = errors_element.get_array().value;
-    return errors.begin() != errors.end();
+    return false == errors.empty();
 }
 
 [[nodiscard]] auto has_write_concern_errors(bsoncxx::document::view const& reply) -> bool {
@@ -55,7 +55,7 @@ constexpr int32_t cDuplicateKeyErrorCode{11'000};
         return true;
     }
     auto const errors = errors_element.get_array().value;
-    return errors.begin() != errors.end();
+    return false == errors.empty();
 }
 
 [[nodiscard]] auto is_duplicate_key_write_error(bsoncxx::array::element const& write_error)
@@ -98,7 +98,7 @@ auto contains_only_duplicate_key_write_errors(mongocxx::bulk_write_exception con
     }
 
     auto const write_errors = write_errors_element.get_array().value;
-    if (write_errors.begin() == write_errors.end()) {
+    if (write_errors.empty()) {
         return false;
     }
     return std::all_of(write_errors.begin(), write_errors.end(), is_duplicate_key_write_error);
--- a/components/core/src/clp_s/MongoDBUtils.cpp
+++ b/components/core/src/clp_s/MongoDBUtils.cpp
@@ -1,6 +1,7 @@
 #include "MongoDBUtils.hpp"
 
 #include <algorithm>
+#include <cstdint>
 
 #include <bsoncxx/types.hpp>
 #include <mongocxx/exception/bulk_write_exception.hpp>
--- a/components/core/src/clp_s/MongoDBUtils.cpp
+++ b/components/core/src/clp_s/MongoDBUtils.cpp
@@ -12,21 +12,21 @@ constexpr int32_t cDuplicateKeyErrorCode{11'000};
 /**
  * Checks whether an aggregated bulk-write reply contains any command errors.
  * @param reply The raw MongoDB bulk-write reply.
- * @return true if the reply contains a command error, false otherwise.
+ * @return Whether the reply contains a command error.
  */
 [[nodiscard]] auto has_command_errors(bsoncxx::document::view const& reply) -> bool;
 
 /**
  * Checks whether a bulk-write reply contains any write-concern errors.
  * @param reply The raw MongoDB bulk-write reply.
- * @return true if the reply contains a write-concern error, false otherwise.
+ * @return Whether the reply contains a write-concern error.
  */
 [[nodiscard]] auto has_write_concern_errors(bsoncxx::document::view const& reply) -> bool;
 
 /**
  * Checks whether an entry from a bulk-write reply's `writeErrors` array is a duplicate-key error.
  * @param write_error The write-error entry to inspect.
- * @return true if the entry has MongoDB's duplicate-key error code, false otherwise.
+ * @return Whether the entry has MongoDB's duplicate-key error code.
  */
 [[nodiscard]] auto is_duplicate_key_write_error(bsoncxx::array::element const& write_error) -> bool;
 
--- a/components/core/src/clp_s/MongoDBUtils.hpp
+++ b/components/core/src/clp_s/MongoDBUtils.hpp
@@ -1,6 +1,8 @@
 #ifndef CLP_S_MONGODBUTILS_HPP
 #define CLP_S_MONGODBUTILS_HPP
 
+#include <cstddef>
+
 #include <mongocxx/exception/bulk_write_exception.hpp>
 
 namespace clp_s {
@@ -8,13 +10,16 @@ namespace clp_s {
  * Returns whether the bulk write failed only because some documents already exist.
  *
  * Command and write-concern errors are rejected since they mean MongoDB did not confirm the outcome
- * of the entire batch. At least one write error must be present, and every write error must be a
- * duplicate-key error.
+ * of the entire batch. The number of inserted documents and write errors must account for every
+ * submitted document, and every write error must be a duplicate-key error.
  * @param exception The exception containing the raw MongoDB bulk-write reply.
- * @return Whether the reply contains only duplicate-key write errors.
+ * @param num_documents The number of documents submitted in the bulk write.
+ * @return Whether the reply accounts for every document using successful inserts and duplicate-key
+ * errors only.
  */
 [[nodiscard]] auto contains_only_duplicate_key_write_errors(
-        mongocxx::bulk_write_exception const& exception
+        mongocxx::bulk_write_exception const& exception,
+        size_t num_documents
 ) -> bool;
 }  // namespace clp_s
 
--- a/components/core/src/clp_s/MongoDBUtils.hpp
+++ b/components/core/src/clp_s/MongoDBUtils.hpp
@@ -11,7 +11,7 @@ namespace clp_s {
  * outcome of the entire batch. At least one write error must be present, and every write error
  * must be a duplicate-key error.
  * @param exception The exception containing the raw MongoDB bulk-write reply.
- * @return true if the reply contains only duplicate-key write errors, false otherwise.
+ * @return Whether the reply contains only duplicate-key write errors.
  */
 [[nodiscard]] auto contains_only_duplicate_key_write_errors(
         mongocxx::bulk_write_exception const& exception
--- a/components/core/src/clp_s/MongoDBUtils.hpp
+++ b/components/core/src/clp_s/MongoDBUtils.hpp
@@ -1,0 +1,21 @@
+#ifndef CLP_S_MONGODBUTILS_HPP
+#define CLP_S_MONGODBUTILS_HPP
+
+#include <mongocxx/exception/bulk_write_exception.hpp>
+
+namespace clp_s {
+/**
+ * Returns whether the bulk write failed only because some documents already exist.
+ *
+ * Command and write-concern errors are rejected since they mean MongoDB did not confirm the
+ * outcome of the entire batch. At least one write error must be present, and every write error
+ * must be a duplicate-key error.
+ * @param exception The exception containing the raw MongoDB bulk-write reply.
+ * @return true if the reply contains only duplicate-key write errors, false otherwise.
+ */
+[[nodiscard]] auto contains_only_duplicate_key_write_errors(
+        mongocxx::bulk_write_exception const& exception
+) -> bool;
+}  // namespace clp_s
+
+#endif  // CLP_S_MONGODBUTILS_HPP
--- a/components/core/src/clp_s/MongoDBUtils.hpp
+++ b/components/core/src/clp_s/MongoDBUtils.hpp
@@ -7,9 +7,9 @@ namespace clp_s {
 /**
  * Returns whether the bulk write failed only because some documents already exist.
  *
- * Command and write-concern errors are rejected since they mean MongoDB did not confirm the
- * outcome of the entire batch. At least one write error must be present, and every write error
- * must be a duplicate-key error.
+ * Command and write-concern errors are rejected since they mean MongoDB did not confirm the outcome
+ * of the entire batch. At least one write error must be present, and every write error must be a
+ * duplicate-key error.
  * @param exception The exception containing the raw MongoDB bulk-write reply.
  * @return Whether the reply contains only duplicate-key write errors.
  */
--- a/components/core/src/clp_s/OutputHandlerImpl.cpp
+++ b/components/core/src/clp_s/OutputHandlerImpl.cpp
@@ -153,7 +153,7 @@ ErrorCode ResultsCacheOutputHandler::finish() {
                                                     ),
                                                     bsoncxx::builder::basic::kvp(
                                                             constants::results_cache::search::
-                                                                    cLogEventIx,
+                                                                    cLogEventIdx,
                                                             result.log_event_idx
                                                     )
                                             )
@@ -169,14 +169,6 @@ ErrorCode ResultsCacheOutputHandler::finish() {
                                     bsoncxx::builder::basic::kvp(
                                             constants::results_cache::search::cTimestamp,
                                             result.timestamp
-                                    ),
-                                    bsoncxx::builder::basic::kvp(
-                                            constants::results_cache::search::cArchiveId,
-                                            std::move(result.archive_id)
-                                    ),
-                                    bsoncxx::builder::basic::kvp(
-                                            constants::results_cache::search::cLogEventIx,
-                                            result.log_event_idx
                                     ),
                                     bsoncxx::builder::basic::kvp(
                                             std::string{constants::results_cache::search::cDataset},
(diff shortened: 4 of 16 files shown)


components/core/src/clp_s/archive_constants.hpp
```diff
@@ -53,8 +53,9 @@ constexpr char cIsLastChunk[]{"is_last_chunk"};
 }  // namespace results_cache::decompression
 
 namespace results_cache::search {
+constexpr char cId[]{"_id"};
 constexpr char cOrigFilePath[]{"orig_file_path"};
-constexpr char cLogEventIx[]{"log_event_ix"};
+constexpr char cLogEventIdx[]{"log_event_idx"};
```
gibber9809, line 58: I think there are a few more places in the codebase that still use `log_event_ix` that I've found -- in particular there are a few instances in `clp-mcp-server/tests/server/test_utils.py`, and usage in clo in `core/src/clp/clo/constants.hpp` + corresponding output handler code.

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -178,6 +304,24 @@ void ResultsCacheOutputHandler::write(
     }
 }
 
+auto ResultsCacheOutputHandler::insert_results() -> bool {
+    try {
+        mongocxx::options::insert options;
+        options.ordered(false);
```
gibber9809: Just to make sure my own understanding is correct, this is what allows us to reliably insert all records in the batch even if some of them have duplicate key errors, right?

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -126,21 +248,25 @@ ErrorCode ResultsCacheOutputHandler::finish() {
...
-            m_collection.insert_many(m_results);
-            m_results.clear();
+            if (false == insert_results()) {
+                return ErrorCode::ErrorCodeFailureDbBulkWrite;
+            }
         }
     } catch (mongocxx::exception const& e) {
+        SPDLOG_ERROR("Failed to insert final search-results batch - {}", e.what());
         return ErrorCode::ErrorCodeFailureDbBulkWrite;
     }
```
gibber9809: I think that this try/catch is redundant, since the insert_results helper also does a try/catch. We should be able to remove this one to simplify the code.

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -126,21 +248,25 @@ ErrorCode ResultsCacheOutputHandler::finish() {
...
+                if (false == insert_results()) {
+                    return ErrorCode::ErrorCodeFailureDbBulkWrite;
+                }
                 count = 0;
             }
         } catch (mongocxx::exception const& e) {
+            SPDLOG_ERROR("Failed to build or insert search results - {}", e.what());
             return ErrorCode::ErrorCodeFailureDbBulkWrite;
         }
     }
```
gibber9809: Ditto, should be able to move `insert_results` out of this try/catch.

I think we might still need it for the bsoncxx builder code though.

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -28,6 +32,117 @@ using std::string;
...
 namespace clp_s {
+namespace {
+constexpr int32_t cDuplicateKeyErrorCode{11'000};
+
+/**
+ * Checks whether a bulk-write reply reports a successful command.
+ * @param reply The raw MongoDB bulk-write reply.
+ * @return true if the reply contains no command error, false otherwise.
+ */
+[[nodiscard]] auto is_successful_command_reply(bsoncxx::document::view const& reply) -> bool {
```
gibber9809: Per our style guide we should separate the prototypes/docstrings for these functions from their implementations.

Within this `namespace{}` the order should be:
1. All prototypes for helpers with their docstrings
2. All implementations

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -28,6 +32,117 @@ using std::string;
...
+    }
+    if (bsoncxx::type::k_double == command_status.type()) {
+        return 1.0 == command_status.get_double().value;
+    }
+    if (bsoncxx::type::k_int32 == command_status.type()) {
+        return 1 == command_status.get_int32().value;
+    }
+    if (bsoncxx::type::k_int64 == command_status.type()) {
+        return 1 == command_status.get_int64().value;
+    }
```
gibber9809: Are these really all possible/is this the canonical way to validate these mongodb responses in c++? Same comment goes for the other validation we're doing in these helpers.

I'm fine with this if this is canonical, it's just that this is a bit more involved/verbose than what I'd expect us to have to do to deal with an API response.

gibber9809 (changes requested):
Leaving an initial review focused mostly on the c++ changes. Will take a look at the python code in another round.

I think we also need to replicate the changes in clp-s into clo so that clp-text can work with the new results cache schema.

components/core/src/clp_s/archive_constants.hpp
sitaowang1998: That's outside the scope of this PR.

components/core/src/clp_s/OutputHandlerImpl.cpp
sitaowang1998: Yes.

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
```diff
@@ -1,33 +1,76 @@
...
+            job_id_placeholders = ",".join(["%s"] * len(job_ids_batch))
+            query = f"""
+                SELECT id
+                FROM `{QUERY_JOBS_TABLE_NAME}`
+                WHERE id IN ({job_id_placeholders})
+                AND TIMESTAMPADD(
+                    MICROSECOND,
+                    CAST(duration * 1000000 AS SIGNED),
+                    start_time
+                ) < TIMESTAMPADD(MINUTE, %s, CURRENT_TIMESTAMP(3))
```
LinZhihao-723: It looks like it's possible that a job has already terminated but the duration is null. For example, if the start time is not set but the job is canceled, duration will never be set, meaning that this query would never return that job: https://github.com/y-scope/clp/blob/8bbec00d1b0a4c4e779376ee36f736959f42b567/components/job-orchestration/job_orchestration/scheduler/query/query_scheduler.py#L540.
Arguably, when the start time is not set, the collection might not be collected either. But for safety purposes, how about we add an `OR` logic after the timestamp check to return any job that:
- is already in the terminated state (failed, succeeded, killed, or canceled)
- the creation time (https://github.com/y-scope/clp/blob/8bbec00d1b0a4c4e779376ee36f736959f42b567/components/clp-py-utils/clp_py_utils/initialize-orchestration-db.py#L137) is smaller than the cutoff?

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
```diff
@@ -38,46 +81,59 @@ def _delete_result_metadata(
 
 
 def _collect_and_sweep_expired_search_results(
-    result_cache_config: ResultsCache, results_metadata_collection_name: str
-):
-    expiry_epoch = get_expiry_epoch_secs(result_cache_config.retention_period)
```
LinZhihao-723, line 43: The method `get_expiry_epoch_secs` is no longer used. We should remove it.
https://github.com/y-scope/clp/blob/8bbec00d1b0a4c4e779376ee36f736959f42b567/components/job-orchestration/job_orchestration/garbage_collector/utils.py#L30

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -160,12 +160,19 @@ async def read_results(self, query_id: str) -> list[dict]:
...
 
-        async for doc in collection.find({}, limit=SEARCH_MAX_NUM_RESULTS):
+        async for raw_doc in collection.find({}, limit=SEARCH_MAX_NUM_RESULTS):
+            doc = dict(raw_doc)
+            result_id = doc["_id"]
+            if isinstance(result_id, dict):
+                doc["archive_id"] = result_id["archive_id"]
+                doc["log_event_idx"] = result_id["log_event_idx"]
+            else:
+                doc["log_event_idx"] = doc.pop("log_event_ix")
```
LinZhihao-723: I'm not sure if it's worth providing backward compatibility, considering this is a breaking change that will not provide backward compatibility in other components in the package.

LinZhihao-723: That said, this PR is a breaking change. We should update the PR title accordingly to reflect that.

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
LinZhihao-723: tbh I don't like how the current search scheduler sets the start time and the duration. The current behavior:
- Sets the start time using the scheduler's time.
- Computes the duration using the scheduler's time.

Instead, it's probably better to:
- Set the start time when the job starts, using the database's time.
- Set the end time when the job reaches a terminal state, using the database's time.
- When duration is needed, query the job and take the diff of `end_time - start_time`.

In this way:
- The database is the single source of truth for the timestamp.
- The end time is always set when a job terminates.

I'm ok to stick with the current behavior for now, but we should plan this down the road of the query coordinator redesign.

cc @kirkrodrigues @Bill-hbrhbr

LinZhihao-723 (changes requested):

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
kirkrodrigues: As discussed offline, the proposal makes sense to me as long as we can ensure the timestamp is always generated on the server.

LinZhihao-723: `CURRENT_TIMESTAMP` is evaluated by `mysqld` on the server host. Let's use `CURRENT_TIMESTAMP`.
@coderabbitai Create an issue and assign it to me and @Bill-hbrhbr.
@sitaowang1998 In this PR, let's do what I suggested: we add an OR to assert terminated jobs that don't have `duration` set.

components/core/src/clp_s/MongoDBUtils.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+namespace clp_s {
+/**
+ * Returns whether the bulk write failed only because some documents already exist.
+ *
+ * Command and write-concern errors are rejected since they mean MongoDB did not confirm the
+ * outcome of the entire batch. At least one write error must be present, and every write error
+ * must be a duplicate-key error.
+ * @param exception The exception containing the raw MongoDB bulk-write reply.
+ * @return true if the reply contains only duplicate-key write errors, false otherwise.
```
LinZhihao-723:
```suggestion
 * @return Whether the reply contains only duplicate-key write errors.
```
Do it for the rest of the docstrings in other C++ headers/sources.

components/clp-mcp-server/clp_mcp_server/clp_connector.py
```diff
@@ -160,12 +160,16 @@ async def read_results(self, query_id: str) -> list[dict]:
         collection = self._results_cache[str(query_id)]
         results = []
 
-        async for doc in collection.find({}, limit=SEARCH_MAX_NUM_RESULTS):
+        async for raw_doc in collection.find({}, limit=SEARCH_MAX_NUM_RESULTS):
+            doc = dict(raw_doc)
```
LinZhihao-723: Can you justify why do we need this?

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
```diff
@@ -38,46 +95,59 @@ def _delete_result_metadata(
...
-                clp_config.results_cache, clp_config.webui.results_metadata_collection_name
+                clp_config.results_cache,
+                clp_config.database,
+                clp_config.webui.results_metadata_collection_name,
             )
             await asyncio.sleep(sweep_interval_secs)
     except Exception:
-        logger.exception(f"{SEARCH_RESULT_GARBAGE_COLLECTOR_NAME} exited with failure.")
+        logger.exception("%s exited with failure.", SEARCH_RESULT_GARBAGE_COLLECTOR_NAME)
         raise
```
LinZhihao-723, line 153: Prefer not to do this refactoring in this PR: the scope of this PR is already wide. Let's try to reduce the scope of the diff to make this PR more readable.

components/webui/packages/client/src/pages/SearchPage/SearchResults/SearchResultsTable/Native/SearchResultsVirtualTable/useSearchResults.ts
```diff
@@ -33,15 +57,24 @@ const useSearchResults = () => {
...
+            }
 
             return {
                 ...doc,
-                _id: "object" === typeof doc._id ?
-                    doc._id.$oid :
-                    doc._id,
+                _id: JSON.stringify(doc._id),
+                archive_id: doc._id.archive_id,
+                log_event_idx: doc._id.log_event_idx,
```
LinZhihao-723, line 76: What is `_id` used for? It looks like we duplicate the archive ID and the log event index in both the flattened fields and the ID. Might need WebUI owners to take a look.

LinZhihao-723 (changes requested):
Part of my review comments for this round. Still need to check `clp-s` and `clo` changes with more details.
On the other hand, I think the `Validation performed` section for this PR is poorly written. This PR changes multiple components in the system, and as a reviewer I'd expect to see a compreshensive report on what's been tested and what behaviors were checked to be correct. For example, you should probably ask your coding agent to set up a mock environment to test the GC behavior. Please make sure you've done enough testing and report it in the PR description.

LinZhihao-723 (commented):
Updated my performance evaluation results in the PR description.

components/core/src/clp_s/MongoDBUtils.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+#include <mongocxx/exception/bulk_write_exception.hpp>
+
+namespace clp_s {
+/**
+ * Returns whether the bulk write failed only because some documents already exist.
+ *
+ * Command and write-concern errors are rejected since they mean MongoDB did not confirm the
+ * outcome of the entire batch. At least one write error must be present, and every write error
+ * must be a duplicate-key error.
```
LinZhihao-723:
```suggestion
 * Command and write-concern errors are rejected since they mean MongoDB did not confirm the outcome
 * of the entire batch. At least one write error must be present, and every write error must be a
 * duplicate-key error.
```
Can you check your tooling to figure out why this is missed?

components/core/src/clp_s/MongoDBUtils.cpp
```diff
@@ -0,0 +1,106 @@
...
+[[nodiscard]] auto has_command_errors(bsoncxx::document::view const& reply) -> bool {
+    auto const errors_element = reply["errorReplies"];
+    if (false == static_cast<bool>(errors_element)) {
+        return false;
+    }
+    if (bsoncxx::type::k_array != errors_element.type()) {
+        return true;
+    }
+    auto const errors = errors_element.get_array().value;
+    return errors.begin() != errors.end();
```
LinZhihao-723: Why not using `false == errors.empty()`?
```suggestion
    return false == errors.empty();
```
If the iterator is selected on purpose, please reply to this comment for why.

components/core/src/clp_s/MongoDBUtils.hpp
```diff
@@ -0,0 +1,21 @@
...
+ * @param exception The exception containing the raw MongoDB bulk-write reply.
+ * @return true if the reply contains only duplicate-key write errors, false otherwise.
+ */
+[[nodiscard]] auto contains_only_duplicate_key_write_errors(
+        mongocxx::bulk_write_exception const& exception
+) -> bool;
```
LinZhihao-723, line 23: Quote my agent's review comment, which seems worth for a closer look:

> `contains_only_duplicate_key_write_errors` classifies the batch purely from the reply's *contents*, never from whether every submitted document was accounted for. libmongoc splits one `insert_many` into several server batches past `maxMessageSizeBytes` or 100k documents and merges the results in unordered mode (`mongoc-bulk-operation.c:761-802`). If an early batch produced only 11000s and a later batch fails at the transport level, `result->failed` is set (`mongoc-write-command.c:747`) but the empty network reply is not appended to `rawErrorReplies` — `_mongoc_cmd_check_ok` returns `true` for a document with no `ok`/`code`/`errmsg` (`mongoc-rpc.c:173-198`) — so `errorReplies` stays absent. The aggregated reply then carries `writeErrors: [11000, …]` and nothing else, the helper returns `true`, and the search task exits success having silently dropped a batch.
> Reachability is low (default `--batch-size` is 1000, typical messages are sub-kilobyte), but the failure mode is precisely the one this design exists to prevent, and it is invisible. The fix is cheap: the reply carries `nInserted` on this path (`mongoc-write-command.c:1127-1131`), so require `nInserted + |writeErrors| == m_results.size()` alongside the all-11000 test. That needs the batch size threaded into the helper — a second parameter, or returning `std::optional<size_t>` (the duplicate count) as the design report's own sketch does.

Can you do some research to make sure that transport errors are not handled silently in this case?

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -178,6 +185,24 @@ void ResultsCacheOutputHandler::write(
     }
 }
 
+auto ResultsCacheOutputHandler::insert_results() -> bool {
+    try {
+        mongocxx::options::insert options;
+        options.ordered(false);
```
LinZhihao-723: It looks like the option should be resuable. How about making it a member variable, `m_insert_options`, and initialize it in the constructor?

components/core/src/clp_s/MongoDBUtils.cpp
```diff
@@ -0,0 +1,106 @@
+#include "MongoDBUtils.hpp"
+
+#include <algorithm>
```
LinZhihao-723: `cstdint` include is missing. Please ask ur coding agent to run clang-tidy on this newly created file.

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -96,6 +99,21 @@ ErrorCode ResultsCacheOutputHandler::finish() {
             m_results.emplace_back(
                     std::move(
                             bsoncxx::builder::basic::make_document(
+                                    bsoncxx::builder::basic::kvp(
+                                            constants::results_cache::search::cId,
+                                            bsoncxx::builder::basic::make_document(
+                                                    bsoncxx::builder::basic::kvp(
+                                                            constants::results_cache::search::
+                                                                    cArchiveId,
+                                                            result.archive_id
```
LinZhihao-723: We're using `std::move` before this rewrite for submitting the archive ID. We should probably do it even though there's a chance that compiler will do this for u.
```suggestion
                                                            std::move(result.archive_id)
```

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
```diff
@@ -1,33 +1,90 @@
...
+
+    MariaDB computes each query's completion time as `start_time + duration`. A query-job ID is
+    included when the time since completion is greater than `retention_period_minutes`. For a
+    terminated query without a completion time, `creation_time` is used instead.
+
+    :param database_config: Configuration for the orchestration database.
+    :param job_ids: Query-job IDs to filter.
+    :param retention_period_minutes: Length of the retention period following query completion, in
+        minutes.
+    :return: Query-job IDs completed more than `retention_period_minutes` ago.
```
LinZhihao-723: Did you proof-read the docstring generated by your agent?
- It should start with `Filters`, not `Filter`.
- The repo's convention is to use `query job`, not `query-job`, right?

Please check the docstring for the other method.

components/core/src/clp_s/OutputHandlerImpl.hpp
```diff
@@ -200,6 +200,13 @@ class ResultsCacheOutputHandler : public ::clp_s::search::OutputHandler {
     void write(std::string_view message) override { write(message, 0, {}, 0); }
```
LinZhihao-723, line 201: With the current implementation, we should probably stop supporting this method. This method is never enabled in the current `clp-s` anyways. Shall we just  through an exception indicating this is not supported? @gibber9809

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
LinZhihao-723, line 1: The implementation lgtm.

LinZhihao-723 (changes requested):
Please review the comments carefully as some may require you to repeat on other changes.

components/core/src/clp_s/MongoDBUtils.cpp
```diff
@@ -0,0 +1,107 @@
...
+        || bsoncxx::type::k_array != write_errors_element.type())
+    {
+        return false;
+    }
+
+    auto const write_errors = write_errors_element.get_array().value;
+    if (write_errors.empty()) {
+        return false;
+    }
+    return std::all_of(write_errors.begin(), write_errors.end(), is_duplicate_key_write_error);
```
junhaoliao: blocking: don't clear the batch unless this reply accounts for every attempted write. The pinned driver can retain earlier 11000 errors while a later transport failure contributes no `errorReplies`, so this `all_of` can report success for an incomplete batch. Thread `m_results.size()` into the helper and require:

```cpp
num_attempted_writes == nInserted + num_duplicate_errors
```

Apply the signature/caller change to both `clp-s` and `clo`, and test incomplete accounting.

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
```diff
@@ -1,33 +1,90 @@
...
+            query = f"""
+                SELECT id
+                FROM `{QUERY_JOBS_TABLE_NAME}`
+                WHERE id IN ({job_id_placeholders})
+                AND (
+                    TIMESTAMPADD(
+                        MICROSECOND,
+                        CAST(duration * 1000000 AS SIGNED),
+                        start_time
+                    ) < TIMESTAMPADD(MINUTE, %s, CURRENT_TIMESTAMP(3))
```
junhaoliao: blocking: use one clock domain. Normal completion is reconstructed from the scheduler host's naive `datetime.now()`, but this cutoff uses the database session timezone. A supported external DB with an offset can drop fresh results immediately. Coordinate the schema and scheduler (see #2516), for example:

```sql
UPDATE query_jobs SET end_time = CURRENT_TIMESTAMP(3) WHERE ...;
-- expire against the database-owned end_time
```

Alternatively, explicitly enforce UTC for every writer and SQL session.

components/core/src/clp_s/OutputHandlerImpl.cpp
```diff
@@ -178,6 +185,22 @@ void ResultsCacheOutputHandler::write(
     }
 }
 
+auto ResultsCacheOutputHandler::insert_results() -> bool {
+    try {
+        m_collection.insert_many(m_results, m_insert_options);
```
junhaoliao, line 190: this isn't quite what i expected from our offline discussion — i thought we were going to use upsert here. is there a reason we went with insert and special handling for duplicate-key errors instead?

with `update_one` filtered by the complete `_id`, `upsert(true)`, and `$setOnInsert` for the remaining fields, retrying an existing event would be a no-op. MongoDB's [official documentation](https://www.mongodb.com/docs/manual/reference/operator/update/setOnInsert/) says:

> If the update operation does not result in an insert, `$setOnInsert` does nothing.

the equivalent operation is:

```js
collection.updateOne(
    { _id: eventId },
    { $setOnInsert: resultFields }, // excludes _id
    { upsert: true }
)
```

we can batch these with the [C++ driver's bulk-write API](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/crud/bulk-write/). this should avoid treating already-stored results as duplicate-insert errors on retries. we'd still need to propagate actual write failures.

that said, i wouldn't assume either approach is faster without comparing them; the current benchmark doesn't test upserts. unless there's another reason to keep inserts, can we use bulk upserts with `$setOnInsert` for both `clp-s` and `clo` as discussed?

junhaoliao (commented):
the webui changes lgtm

components/core/src/clp_s/OutputHandlerImpl.cpp
LinZhihao-723: Sorry for the misunderstanding. I will benchmark and compare before we proceed.

LinZhihao-723: After further investigation, I think we should stick to `insert` instead of switching to `upsert` for the following reasons:
- The happy-path performance is bad in the benchmark results: `upsert` has to perform an indexed match for every document. We've done the same benchmark as the one in PR description for `insert` vs. `upsert` and it confirms this overhead as shown in the table:
    | N | INSERT wall (s) | UPSERT wall (s) | Δ wall | INSERT e2e mean (s) | UPSERT e2e mean (s) | Δ e2e | per-rep ranges |
    | --- | --- | --- | --- | --- | --- | --- | --- |
    | 1 | 16.5227 | 22.0372 | **+33.4%** | 0.1291 | 0.1721 | +33.4% | **disjoint** |
    | 2 | 8.2495 | 12.5577 | **+52.2%** | 0.1289 | 0.1962 | +52.2% | **disjoint** |
    | 4 | 4.2669 | 7.8855 | **+84.8%** | 0.1333 | 0.2430 | +82.3% | **disjoint** |
    | 8 | 2.4910 | 6.6335 | **+166.3%** | 0.1550 | 0.4079 | +163.2% | **disjoint** |
    | 16 | 1.7453 | 7.8710 | **+351.0%** | 0.2147 | 0.9603 | +347.2% | **disjoint** |
- Using `upsert` would not simplify the error-handling path: `upsert` does not eliminate duplicate-key errors under concurrency; if two operations both find no matching document, both attempt the insert, one succeeds, and the other fails with E11000. The same logic we've implemented so far would still be necessary even for `upsert`.

I think the conclusion is that we should stick to the current approach.

NOTE: The `upsert` benchmark uses bulk-write API to batch writes.

LinZhihao-723: @sitaowang1998 did a similar micro-benchmark to test the performance of `insert` vs. `upsert` under different document duplication levels. This micro-benchmark doesn't involve any clp-s time, and it proves that using `upsert` would hurt performance in the happy path (where all documents create a new _id).
image

components/core/src/clp_s/MongoDBUtils.cpp
```diff
@@ -0,0 +1,128 @@
+#include "MongoDBUtils.hpp"
+
+#include <cstdint>
+
+#include <bsoncxx/types.hpp>
+#include <mongocxx/exception/bulk_write_exception.hpp>
+
+namespace clp_s {
+namespace {
+constexpr int32_t cDuplicateKeyErrorCode{11'000};
```
LinZhihao-723, line 10: This constant is only used inside `is_duplicate_key_write_error`, how about we move it inside the method?

components/job-orchestration/job_orchestration/garbage_collector/search_result_garbage_collector.py
```diff
@@ -67,6 +134,7 @@ def _collect_and_sweep_expired_search_results(
 
 
 async def search_result_garbage_collector(clp_config: ClpConfig) -> None:
+    """Run search-result collection and sweeping at the configured interval."""
```
LinZhihao-723:
```suggestion
    """Runs search-result collection and sweeping at the configured interval."""
```

LinZhihao-723 (changes requested):
lgtm. Two more nit comments.

components/core/src/clp_s/OutputHandlerImpl.hpp
gibber9809: I think it's still used for some of the output handlers -- originally I think this interface was intended to reduce overhead by skipping retrieving some of this metadata + giving output handlers a direct interface for forwarding just the message to their output sink to save on bandwidth, but:

1. Skipping retrieving the timestamp/archive ID/log event index for matching results shouldn't add any meaningful overhead anyway (though we should double-check this)
2. If some output handlers want to not send all of this metadata over the network they can just implement it behind the other interface, so this interface just makes our code more complex for no gain

I'll write up a github issue for removing this interface.

LinZhihao-723: Sorry, I didn't mean we should remove this method from the interface. I mean that for this particular output handler (result-cache writer), this method is not used, and the current implementation won't be meaningful. Shall we just throw an exception for this implementation instead?

LinZhihao-723: @coderabbitai Can you create an issue to keep track of this for entirely removing this method from the interface?
@sitaowang1998 For this PR, let's throw an exception to guard it, making sure no one can actually use this method with the latest index scheme.

gibber9809: Already wrote #2519, so @coderabbit do not create an issue.

gibber9809 (approved):
Reviewed all of c++ code, LGTM. PR title is fine as well.

20001020ycx:
Unfortunately I cannot reproduce the bug that you have reported for the clp_mcp_server. Please see my conversation history with Claude code: https://claude.ai/share/687a9263-68f0-44a0-b93b-40fac9c6366d

20001020ycx:
> Unfortunately I cannot reproduce the bug that you have reported for the clp_mcp_server. Please see my conversation history with Claude code: https://claude.ai/share/687a9263-68f0-44a0-b93b-40fac9c6366d

Given that Claude code does not show the content for the tool call in the exported conversation history, I will paste the input and output for first tool call made by AI here:
```
Request

{
  "kql_query": "*error*"
}
Response

{"items":["timestamp: 1970-01-01T00:00:00.000Z, message: {\"ts\":1756700000118,\"level\":\"ERROR\",\"service\":\"svc-4\",\"txn\":4,\"op\":\"select\",\"rows\":53,\"elapsed_ms\":777,\"msg\":\"row batch flushed\"}\n, link: http://localhost:4000/streamFile?type=json&streamId=07268a67-6972-49fb-b7ca-5f97c71d0a2b&dataset=default&logEventIdx=4","timestamp: 1970-01-01T00:00:00.000Z, message: {\"ts\":1756700000795,\"level\":\"ERROR\",\"service\":\"svc-1\",\"txn\":25,\"op\":\"select\",\"rows\":2786,\"elapsed_ms\":114,\"msg\":\"replica lag detected\"}\n, link: http://localhost:4000/streamFile?type=json&streamId=07268a67-6972-49fb-b7ca-5f97c71d0a2b&dataset=default&logEventIdx=25","timestamp: 1970-01-01T00:00:00.000Z, message: {\"ts\":1756700001585,\"level\":\"ERROR\",\"service\":\"svc-5\",\"txn\":53,\"op\":\"rollback\",\"rows\":465,\"elapsed_ms\":520,\"msg\":\"lock wait timeout\"}\n, link: http://localhost:4000/streamFile?type=json&streamId=07268a67-6972-49fb-b7ca-5f97c71d0a2b&dataset=default&logEventIdx=53","timestamp: 1970-01-01T00:00:00.000Z, message: {\"ts\":1756700002364,\"level\":\"ERROR\",\"service\":\"svc-3\",\"txn\":83,\"op\":\"rollback\",\"rows\":2086,\"elapsed_ms\":45,\"msg\":\"index scan completed\"}\n, link: http://localhost:4000/streamFile?type=json&streamId=07268a67-6972-49fb-b7ca-5f97c71d0a2b&dataset=default&logEventIdx=83","timestamp: 1970-01-01T00:00:00.000Z, message: {\"ts\":1756700004236,\"level\":\"ERROR\",\"service\":\"svc-2\",\"txn\":154,\"op\":\"commit\",\"rows\":4994,\"elapsed_ms\":757,\"msg\":\"lock wait timeout\"}\n, link: http://localhost:4000/streamFile?type=json&streamId=07268a67-6972-49fb-b7ca-5f97c71d0a2b&dataset=default&logEventIdx=154","timestamp: 1970-01-01T00:00:00.000Z, message: {\"ts\":1756700009659,\"level\":\"ERROR\",\"service\":\"svc-1\",\"txn\":361,\"op\":\"delete\",\"rows\":4815,\"elapsed_ms\":186,\"msg\":\"row batch flushed\"}\n, link: http://localhost:4000/streamFile?type=json&streamId=07268a67-6972-49fb-b7ca-5f97c71d0a2b&dataset=default&logEventIdx=361","timestamp: 1970-01-01T00:00:00.000Z, message: {\"ts\":1756700020655,\"level\":\"ERROR\",\"service\":\"svc-1\",\"txn\":785,\"op\":\"delete\",\"rows\":4047,\"elapsed_ms\":152,\"msg\":\"lock wait timeout\"}\n, link: http://localhost:4000/streamFile?type=json&streamId=07268a67-6972-49fb-b7ca-5f97c71d0a2b&dataset=default&logEventIdx=785","timestamp: 1970-01-01T00:00:00.000Z, message: {\"ts\":1756700040071,\"level\":\"ERROR\",\"service\":\"svc-1\",\"txn\":1553,\"op\":\"commit\",\"rows\":47,\"elapsed_ms\":404,\"msg\":\"lock wait timeout\"}\n, link: http://localhost:4000/streamFile?type=json&streamId=07268a67-6972-49fb-b7ca-5f97c71d0a2b&dataset=default&logEventIdx=1553","timestamp: 1970-01-01T00:00:00.000Z, message: {\"ts\":1756700082590,\"level\":\"ERROR\",\"service\":\"svc-7\",\"txn\":3231,\"op\":\"rollback\",\"rows\":1729,\"elapsed_ms\":151,\"msg\":\"checkpoint written\"}\n, link: http://localhost:4000/streamFile?type=json&streamId=07268a67-6972-49fb-b7ca-5f97c71d0a2b&dataset=default&logEventIdx=3231","timestamp: 1970-01-01T00:00:00.000Z, message: {\"ts\":1756700153477,\"level\":\"ERROR\",\"service\":\"svc-5\",\"txn\":6061,\"op\":\"commit\",\"rows\":3882,\"elapsed_ms\":682,\"msg\":\"index scan completed\"}\n, link: http://localhost:4000/streamFile?type=json&streamId=07268a67-6972-49fb-b7ca-5f97c71d0a2b&dataset=default&logEventIdx=6061"],
"num_total_pages":100,"num_total_items":1000,"num_items_per_page":10,"has_next":true,"has_previous":false}
```

PR #2512, opened 2026-09-01
feat(clp-tdl-package): Implement the `query::clp_s_search` task.

LinZhihao-723:
# Description

This PR depends on #2503 and #2508.

This PR implements `query::clp_s_search` task. Given a query job ID, `clp-s` query options, an optional dataset, one archive ID, and an output handle, the task:

* Resolves how `clp-s` should address the archive — either a local dataset archive directory plus `--archive-id`, or an S3 object URL read with `--auth s3` — and resolves the AWS credentials that second form needs.
* Builds the `clp-s s ... results-cache` argument vector.
* Runs `clp-s` to completion, letting it write matches straight into MongoDB. Results never travel back through Spider.

Only `OutputHandle::ResultsCache` is supported; `OutputHandle::File` and any storage engine other than `clp-s` are rejected before `clp-s` is spawned.

Alongside the task itself:

* Fills in `OutputHandle`'s variants, which #2503 deliberately left empty.
* Adds `ArchiveOutput::dataset_archive_object_key` so the read side builds byte-identical S3 keys to the ones compression writes.
* Make use of #2508: Hoists `clp_binary_path` and `s3_credential_env` out of the compression task into a shared `task::utils` module.

# Checklist

  breaking change.

# Validation performed

## Unit tests added

16 new unit tests, all committed and run with no external services.

## Filesystem- and S3-backed validation

The task was exercised end to end against a real `clp-s` binary, a `mongo:8.0.21` container, and a MinIO container, driven through the package's real FFI entry points (`__spider_tdl_package_init` / `__spider_tdl_package_execute`) with parameters encoded exactly as the coordinator will encode them. This harness is **not** included in the PR — it is one-time verification ahead of the end-to-end system, so only the unit tests above are committed.

**Filesystem-backed archives** (multi-file directory archives, as an FS deployment produces) — 17 cases:

* Happy path with an implicit dataset: results land in the `<query_job_id>` collection, and every document carries the expected `archive_id`, `dataset: "default"`, and message.
* Explicit dataset: results carry the dataset label, **and** an archive that exists only under `default` correctly fails to resolve under another dataset name — so the path join is doing real work.
* `--tge` alone, `--tle` alone, and a bracketed range: each narrows the result set against the fixture's real millisecond timestamps, with exact expected counts. This is what proves no microsecond/millisecond conversion crept in — a conversion bug returns zero rows here.
* `ignore_case`: a case-mismatched query returns hits when set and none when unset.
* `max_num_results: Some(1)`: exactly one document, and it is the newest by timestamp. Against a >1000-event archive, an explicit cap of 1000 truncates to exactly 1000, confirming the fixture is large enough to be meaningful.
* A query matching nothing succeeds with an empty collection, rather than being reported as a failure.
* Two tasks sharing one `query_job_id` accumulate into the same collection without clobbering each other, mirroring a multi-archive job.
* Failure paths: a nonexistent archive ID, a malformed KQL query, and an unreachable results-cache URI each surface as an error.
* `OutputHandle::File` is rejected before `clp-s` is spawned and writes nothing.
* Wire contract: parameters round-trip over the real msgpack encoding; a malformed `OutputHandle` payload is rejected as a deserialization error without running the task; an unknown task name returns `TaskNotFound`; and `__spider_tdl_package_init` returns an error rather than panicking when `CLP_CONFIG_PATH` is unset.

**S3-backed archives** (single-file archives in MinIO, since `clp-s` reads any network archive as single-file) — 5 cases:

* Happy path: the archive is fetched over a path-style URL with `--auth s3`, and every result document's `archive_id` matches — which on this path `clp-s` derives from the last component of the object key, so this also pins the key builder.
* `staging_directory` pointed at an **empty** directory: results still come back in full, and the directory is still empty afterwards. Without this the S3 branch could silently search the local staging path and return zero rows with exit code 0.
* Explicit dataset: the object resolves under `<key_prefix><dataset>/<archive_id>` and the documents carry the dataset label.
* Credentials from config reach the `clp-s` child process, overriding a deliberately poisoned inherited environment.
* The generated URL is path-style (`http://<host>:<port>/<bucket>/<key>`), which is what an S3-compatible endpoint requires.

## Summary by CodeRabbit

* **New Features**
  * Added single-archive CLP search tasks with support for query limits, time ranges, and case-insensitive searches.
  * Added results-cache and file-based output options.
  * Added dataset-aware archive path generation for default and named datasets.

* **Documentation**
  * Documented the new `query::clp_s_search` task and expanded package task descriptions.

* **Testing**
  * Added coverage for query serialization, archive paths, search argument handling, and storage resolution.


components/clp-tdl-package/src/task/query/search.rs
```diff
@@ -0,0 +1,704 @@
...
+            args.push(OsString::from(archive_id));
+        }
+        ArchiveSelector::ObjectUrl(url) => {
+            args.push(OsString::from(url));
+            args.push(OsString::from("--auth"));
+            args.push(OsString::from("s3"));
+        }
+    }
+
+    args.push(OsString::from(clp_s_query_option.query_string.as_str()));
```
Bill-hbrhbr, line 194:
```suggestion
    args.push(OsString::from(format!(
        "({})",
        clp_s_query_option.query_string.as_str()
    )));
```
Wrap the query in KQL parentheses so queries aren’t interpreted as CLI options. Shell quotes alone prevent splitting and expansion, not option parsing.

components/clp-tdl-package/README.md
```diff
@@ -12,3 +12,7 @@ documented below.
 
 * `compression::clp_s_s3_compress`: Compress inputs from S3 using `clp-s`.
 * `compression::commit`: Commit compression task outcomes to the CLP metadata database.
+
+### Query
+
+* `query::clp_s_search`: Query a single `clp-s` archive.
```
Bill-hbrhbr:
```suggestion

### Query

* `query::clp_s_search`: Search a single archive using the `clp-s` engine.
```

Bill-hbrhbr (changes requested):
Nits

components/clp-tdl-package/src/task/query/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+mod search;
+
+#[task(name = "query::clp_s_search")]
+pub(crate) fn clp_s_search_task(
+    ctx: TaskContext,
+    query_job_id: QueryJobId,
+    clp_s_query_option: ClpSQueryOption,
+    dataset: Option<NonEmptyString>,
+    archive_id: NonEmptyString,
+    output_handle: OutputHandle,
```
Bill-hbrhbr, line 20:
```suggestion
    clp_s_query_option: ClpSQueryOption,
    output_handle: OutputHandle,
    dataset: Option<NonEmptyString>,
    archive_id: NonEmptyString,
```
Let's match the signature argument order.

components/clp-tdl-package/src/task/query/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+) -> Result<(), TdlError> {
+    search::search(
+        &ctx,
+        crate::common::spider_task_executor_config(),
+        query_job_id,
+        &clp_s_query_option,
+        &output_handle,
+        dataset.as_ref().map(NonEmptyString::as_str),
```
Bill-hbrhbr:
```suggestion
        &dataset,
```
I assume this is to indicate that `search::search` will only require a read-only view of the `dataset` variable?
Would the suggestion be simpler?

components/clp-tdl-package/src/task/query/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+        archive_id.into_inner(),
```
Bill-hbrhbr, line 27:
```suggestion
        &archive_id,
```
Similarly wonder if this would just work.

components/clp-tdl-package/src/task/query/search.rs
```diff
@@ -0,0 +1,704 @@
...
+/// * Forwards [`resolve_archive_input`]'s return values on failure.
+/// * Forwards [`run_clp_s_search`]'s return values on failure.
+pub(super) fn search(
+    ctx: &spider_tdl::TaskContext,
+    config: &SpiderTaskExecutorConfig,
+    query_job_id: QueryJobId,
+    clp_s_query_option: &ClpSQueryOption,
+    output_handle: &OutputHandle,
+    dataset: Option<&str>,
+    archive_id: String,
```
Bill-hbrhbr:
```suggestion
    dataset: &Option<NonEmptyString>,
    archive_id: &NonEmptyString,
```
As discussed previously, is there a strong preference of one versus another here? Do we want to avoid `NonEmptyString` for inner helper functions?
I feel like according to the guildines we should specify non-emptiness whenever it can be guaranteed.

components/clp-rust-utils/src/clp_config/package/config.rs
```diff
@@ -381,6 +381,20 @@ impl ArchiveOutput {
...
     }
+
+    /// Derives the S3 object key of an archive in a dataset.
+    ///
+    /// # Returns
+    ///
+    /// The dataset's archive storage directory joined with `archive_id`, where a `None` dataset
+    /// resolves to `default`.
+    #[must_use]
+    pub fn dataset_archive_object_key(&self, dataset: Option<&str>, archive_id: &str) -> String {
```
Bill-hbrhbr, line 392: Would it make sense to put this migration from `create_archive_s3_key` to a member function of the `ArchiveOutput` in a separate PR just for cleaniness?

components/clp-tdl-package/src/task/query/search.rs
```diff
@@ -0,0 +1,704 @@
...
+        OsString::from("--uri"),
+        OsString::from(result_cache_uri),
+        OsString::from("--collection"),
+        OsString::from(query_job_id.to_string()),
+    ]);
+    if let Some(max_num_results) = clp_s_query_option.max_num_results {
+        args.push(OsString::from("--max-num-results"));
+        args.push(OsString::from(max_num_results.to_string()));
+    }
+    args.extend([OsString::from("--dataset"), OsString::from(dataset)]);
```
Bill-hbrhbr, line 218: Should we move this before the results-cache-specific arguments? Just for readability.

components/clp-tdl-package/src/task/query/search.rs
```diff
@@ -0,0 +1,704 @@
...
+        OsString::from("results-cache"),
+        OsString::from("--uri"),
+        OsString::from(result_cache_uri),
+        OsString::from("--collection"),
+        OsString::from(query_job_id.to_string()),
+    ]);
+    if let Some(max_num_results) = clp_s_query_option.max_num_results {
+        args.push(OsString::from("--max-num-results"));
+        args.push(OsString::from(max_num_results.to_string()));
+    }
```
Bill-hbrhbr, line 217: I understand that this function specifically constructs the results-cache output-handler arguments. 

`max_num_results` belongs in `ClpSQueryOption` rather than `OutputHandle::ResultsCache`, since it defines query behavior, not the output destination.

Nevertheless, is there a way to make it clearer that clp-s supports `--max-num-results` only for the results-cache output handler in our code?

components/clp-tdl-package/src/task/query/search.rs
```diff
@@ -0,0 +1,704 @@
...
+///
+/// # Returns
+///
+/// The ordered clp-s arguments.
+fn build_clp_s_search_args_for_result_cache(
+    archive_selector: &ArchiveSelector,
+    clp_s_query_option: &ClpSQueryOption,
+    result_cache_uri: &str,
+    query_job_id: QueryJobId,
+    dataset: &str,
```
Bill-hbrhbr, line 178:
```suggestion
    archive_selector: &ArchiveSelector,
    clp_s_query_option: &ClpSQueryOption,
    result_cache_uri: &NonEmptyString,
    query_job_id: QueryJobId,
    dataset: &NonEmptyString,
```
Should we enforce non-empty strings as stronger invariants for bullet-proof arg construction?

Bill-hbrhbr (commented):
More comments

components/clp-tdl-package/src/task/query/search.rs
```diff
@@ -0,0 +1,704 @@
...
+    };
+
+    let object_key = config
+        .archive_output
+        .dataset_archive_object_key(Some(dataset), &archive_id);
+    let object_key = NonEmptyString::try_from(object_key)
+        .map_err(|_| anyhow::anyhow!("archive object key must not be empty"))?;
+    let url = generate_s3_url(
+        s3_config.endpoint_url.as_ref().map(NonEmptyString::as_str),
+        s3_config.region_code.as_ref().map(NonEmptyString::as_str),
```
Bill-hbrhbr, line 156: **AI-agent review suggestion (Codex):** When both `endpoint_url` and `region_code` are configured, `generate_s3_url` prepends the region to the custom endpoint’s hostname. For example, `http://minio:9000` becomes `http://us-east-1.minio:9000`. Since compression’s S3 client uses the configured endpoint unchanged, could this cause compression and search to address different hosts?

components/clp-tdl-package/src/task/query/search.rs
```diff
@@ -0,0 +1,704 @@
...
+        &s3_config.bucket,
+        &object_key,
```
Bill-hbrhbr, line 158: **AI-agent review suggestion (Codex):** The S3 object key is inserted into the generated URL without percent-encoding. Since `key_prefix` may contain URL-significant characters such as spaces, `#`, or `?`, could search resolve a different URL from the raw key used by compression? Should URL construction encode the bucket/key path components?

Bill-hbrhbr (commented):
The two inline notes below are review suggestions generated by Codex, an AI agent, and posted on behalf of Bill Hu.

components/clp-rust-utils/src/clp_config/package/config.rs
LinZhihao-723: This method is required and used directly by the changes in this PR, so it should be ok to be included as a part of this PR.

components/clp-tdl-package/src/task/query/mod.rs
LinZhihao-723: This actually indicates a different type:
- `dataset.as_ref().map(NonEmptyString::as_str)` means `Option<&str>`
- `&dataset` means `&Option<NonEmptyString>`.

In Rust standard, `Option<&T>` (in our case, `&T` of `NonEmptyString` is `&str`) is usually preferred over `&Option<T>` for function parameters. There are two reasons:
- Flexibility. `&Option<T>` means you need to construct an optional variable to use the call, while `Option<&T>` allows u to pass `T` directly by wrapping its reference.
- Efficiency. To access the underlying string for `&Option<NonEmptyString>`, you need to first access the address of the option variable, and then the string pointer underneath `NonEmptyString`. But `Option<&str>` allows u to access the string pointer directly without an extra layer of indirection.
  - Similarly, if you have `Option<Vec<T>>` and wanna pass it as a function param, `Option<&[T]>` is preferred over `&Option<Vec<T>>`.

components/clp-tdl-package/src/task/query/search.rs
LinZhihao-723: As local helpers, `&str` is preferred for the same reason as explained in https://github.com/y-scope/clp/pull/2512/changes/BASE..672af8be42c13bda97399e57ce0ffa6ce9762276#r3982795813

components/clp-tdl-package/src/task/query/search.rs
LinZhihao-723: We can't 'cuz this is actually a result-cache-sepcific argument, lol

components/clp-tdl-package/src/task/query/search.rs
LinZhihao-723: This should be fine. The existing scheme won't lead to any special characters in the URL.

components/clp-tdl-package/src/task/query/search.rs
LinZhihao-723: tbh I didn't get this...

components/clp-tdl-package/src/task/query/search.rs
LinZhihao-723: As discussed offline, these options are only used by the result handler at this point, but it can still be needed by other output handles. Before we have a clear idea of how to structure them, let's keep the current shape for now.

components/clp-tdl-package/src/task/query/search.rs
LinZhihao-723: Alright, I got it now. Tracked by #2523. I don't think this is a problem we should solve in this PR, but it's worth documenting.

components/clp-tdl-package/src/task/query/search.rs
LinZhihao-723: Hmmm, this is indeed a problem. The Python implementation should have the same issue.
However, I don't think the proposed solution is correct: suppose you have a query `a) AND (b`. Without this proposed fix, it would be considered a syntax error; with this fix, it becomes `(a) AND (b)` and it would pass silently.
Since we're planning to make a formal Rust library for CLP (either through a C-ffi layer or through a real Rust library) that can naturally resolve this issue, I think it might be ok to leave it as is to replicate Python/Celery's behavior. @gibber9809 What do you think?

Bill-hbrhbr: Ok seems like this issue will not be within the scope of this PR. Once there is a consensus on what to do, an issue should be opened.

components/clp-tdl-package/src/task/query/mod.rs
LinZhihao-723: I actually prefer the order in this task definition. I changed the implementation signature to match this order instead.

components/clp-tdl-package/src/task/query/search.rs
gibber9809: I'd say we should avoid modifying the query for now, and just allow this to be addressed when this becomes a library call. Without parsing the query before passing this argument to the binary there isn't really a clear fix that avoids introducing a new edge case, so it probably makes sense to keep behaviour consistent with what we were doing before.

LinZhihao-723: sgtm. I will leave the behavior untouched.
@coderabbitai Can you create an issue to keep track of this issue?

Bill-hbrhbr (approved):

components/clp-tdl-package/src/task/query/search.rs
```diff
@@ -0,0 +1,704 @@
...
+            ));
+        }
+        ArchiveOutputStorage::S3 { s3_config, .. } => s3_config,
+    };
+
+    let object_key = config
+        .archive_output
+        .dataset_archive_object_key(Some(dataset), &archive_id);
+    let object_key = NonEmptyString::try_from(object_key)
+        .map_err(|_| anyhow::anyhow!("archive object key must not be empty"))?;
```
Bill-hbrhbr: Should we just make `dataset_archive_object_key` return `NonEmptyString`?

components/clp-tdl-package/src/task/query/search.rs
```diff
@@ -0,0 +1,704 @@
...
+
+    /// The URL of an S3-hosted archive, read with `--auth s3`.
+    ObjectUrl(String),
+}
+
+/// Resolves how clp-s addresses the archive, and the credential env vars it needs to read it.
+///
+/// # Returns
+///
+/// A tuple containing:
```
Bill-hbrhbr:
```suggestion
/// A tuple on success, containing:
```

components/clp-tdl-package/src/task/query/search.rs
```diff
@@ -0,0 +1,704 @@
...
+/// * `output_handle` is not [`OutputHandle::ResultsCache`]. The current implementation only
+///   supports result cache output streaming.
+/// * Forwards [`resolve_archive_input`]'s return values on failure.
+/// * Forwards [`run_clp_s_search`]'s return values on failure.
+pub(super) fn search(
+    ctx: &spider_tdl::TaskContext,
+    config: &SpiderTaskExecutorConfig,
+    query_job_id: QueryJobId,
+    clp_s_query_option: &ClpSQueryOption,
+    archive_id: String,
```
Bill-hbrhbr:
```suggestion
    archive_id: ArchiveId,
```
@LinZhihao-723 
Should we enforce the type here? If so, let's open an issue and fix it in a separate PR.

Bill-hbrhbr (changes requested):

components/clp-tdl-package/src/task/query/search.rs
Bill-hbrhbr: The joined string from the archive staging directory and archive_id should always be non-empty. If we want to throw an error here, it should indicate that either the directory path is invalid or the archive ID is empty.

components/clp-tdl-package/src/task/query/search.rs
LinZhihao-723: Shouldn't we do it after #2531 is merged? I.e., we should do this as a part of #2531.

PR #2513, opened 2026-09-02
feat(query-coordinator): Add query job lifecycle handling

Bill-hbrhbr:
# Description

Add the query-job lifecycle foundation to the query coordinator:

* Introduce `QueryJobHandle` and a coordinator-prepared `QueryPlan`.
* Submit new jobs, persist their Spider ID and running state, and recover submitted jobs.
* Start and poll Spider jobs with bounded exponential backoff.
* Persist successful, failed, and unexpectedly cancelled outcomes in the query-jobs table.
* Add the `spider_id` and `status_msg` query-job columns, including idempotent upgrades for existing tables.

Spider task-graph construction remains intentionally deferred; `submit_query_job` retains its existing skeleton.

# Checklist

  breaking change.

# Validation performed

* [X] Conform to the design and works in e2e branches.

## Summary by CodeRabbit

* **New Features**
  * Added query job coordination, including submission, recovery, polling, and completion tracking.
  * Added single-archive search with query text, result limits, time ranges, and case-insensitive matching.
  * Added result-cache output support for archives stored in supported locations.
  * Query tasks can now run alongside existing compression tasks.

* **Improvements**
  * Query job records now retain status messages and associated job identifiers.
  * Improved handling for failed jobs, invalid configurations, and unavailable task inputs.
  * Added clearer task documentation for query capabilities.


components/clp-py-utils/clp_py_utils/initialize-orchestration-db.py
```diff
@@ -134,19 +134,53 @@ def main(argv):
...
+                    """,
+                    ER_DUP_KEYNAME,
+                ),
+            )
+            for upgrade_query, duplicate_error_code in query_jobs_table_upgrades:
+                try:
+                    scheduling_db_cursor.execute(upgrade_query)
+                except Exception as err:
+                    if not (hasattr(err, "errno") and err.errno == duplicate_error_code):
+                        raise
```
LinZhihao-723: Similar to compression job table: we do not need to support backward compatibility.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,258 @@
...
+pub struct SpiderOption {
+    /// Delay before the first Spider job-state poll.
+    pub initial_poll_backoff: Duration,
+
+    /// Maximum delay between Spider job-state polls.
+    pub max_poll_backoff: Duration,
+}
+
+/// Drives one already-planned query job through submission and terminal persistence.
+pub struct QueryJobHandle<SubmitterType: QueryJobSubmitter> {
```
LinZhihao-723, line 47: Type parameter docstring is missing.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,258 @@
...
+    db_pool: MySqlPool,
+    query_job_id: QueryJobId,
+    job_submitter: SubmitterType,
+    resource_group_id: ResourceGroupId,
+    query_plan: QueryPlan,
+    spider_option: Arc<SpiderOption>,
+}
+
+impl<SubmitterType: QueryJobSubmitter> QueryJobHandle<SubmitterType> {
+    /// Constructs a handle for an already-planned query job.
```
LinZhihao-723: This docstring doesn't match the convention for factories.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,258 @@
...
+        );
+
+        let result = self.to_completion(spider_job_id).await;
+        if let Err(error) = &result {
+            self.report_failure(error).await;
+        }
+        result
+    }
+
+    async fn submit_and_wait(&self) -> Result<(), Error> {
```
LinZhihao-723: I will stop reading since the docstrings are missing for all private methods.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,258 @@
...
+
+    /// Submits the prepared graph and drives the query job to a terminal state.
+    ///
+    /// On an orchestration failure, this method makes a best-effort attempt to mark the CLP query
+    /// job as failed before returning the original error.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if submission, submission persistence, polling, or terminal persistence
+    /// fails.
```
LinZhihao-723: The docstring does not match the coding guideline.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,258 @@
...
+        result
+    }
+
+    /// Resumes a query job that was already submitted to Spider.
+    ///
+    /// The caller must ensure `spider_job_id` belongs to this CLP query job.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if polling or terminal persistence fails.
```
LinZhihao-723: The docstring does not match the coding guideline.

LinZhihao-723 (changes requested):
You should ask codex/claude to do an adversay code review: instead of telling it to code by following the guideline from https://github.com/LinZhihao-723/claude-instruction/tree/main/rust, ask it to:
- The coding subagent programs by following the guideline.
- The review subagent reviews the code by checking against the guideline.

That should bring the delivered code much closer to the desirable shape.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,308 @@
...
+        spider_job_id: SpiderJobId,
+        num_tasks: i32,
+    ) -> Result<(), Error> {
+        let query = format!(
+            "UPDATE `{QUERY_JOBS_TABLE_NAME}` SET `spider_id` = ?, `status` = ?, `num_tasks` = ?, \
+             `start_time` = CURRENT_TIMESTAMP(3) WHERE `id` = ? AND `status` = ?"
+        );
+        let result = sqlx::query(&query)
+            .bind(spider_job_id.get())
+            .bind(i32::from(QueryJobStatus::Running))
```
LinZhihao-723: We should avoid implicit enum-to-int conversion. Use `sqlx::Type` derive instead like compression job status:
- Definitions: https://github.com/y-scope/clp/blob/main/components/clp-rust-utils/src/job_config/compression.rs
- Use: https://github.com/y-scope/clp/blob/2a5fbee5d6c6635aa3a4c354317b57a865b96bd2/components/compression-coordinator/src/job_handle.rs#L468

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,308 @@
...
+
+        tracing::info!(
+            query_job_id = % self.query_job_id,
+            spider_job_id = % spider_job_id,
+            outcome = ? outcome,
+            "Query job reached a terminal Spider state.",
+        );
+
+        let (status, status_message) = match outcome {
+            QueryJobOutcome::Succeeded => (QueryJobStatus::Succeeded, String::new()),
```
LinZhihao-723, line 233: This is a very bad design pattern: on success, a status message should not be required. Instead of constructing an empty string, we should pass an optional string and mark it `None`. Never use an empty container as an indicator of a null value.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,308 @@
...
+            QueryJobOutcome::Failed { error_message } => (
+                QueryJobStatus::Failed,
+                format!("The Spider query job failed: {error_message}"),
+            ),
+            QueryJobOutcome::UnexpectedlyCancelled => (
+                QueryJobStatus::Failed,
+                "The Spider query job was unexpectedly cancelled.".to_string(),
+            ),
```
LinZhihao-723, line 241: Why do we need this to be `UnexpectedCancelled`? As we discussed previously, the plan to support query job cancellation would be to cancel the spider job and handle the `canceled` status in the job handler, so this should be where the cancellation gets finalized, right?
Please make sure we're on the same page for the future cancellation implementation before you make any further changes.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,308 @@
...
+    async fn update_terminal_status(
+        &self,
+        status: QueryJobStatus,
+        status_message: &str,
+        expected_status: QueryJobStatus,
+    ) -> Result<(), sqlx::Error> {
+        let query = format!(
+            "UPDATE `{QUERY_JOBS_TABLE_NAME}` SET `status` = ?, `status_msg` = LEFT(?, 512), \
+             `duration` = CASE WHEN `start_time` IS NULL THEN 0 ELSE TIMESTAMPDIFF(MICROSECOND, \
+             `start_time`, CURRENT_TIMESTAMP(3)) / 1000000.0 END WHERE `id` = ? AND `status` = ?"
```
LinZhihao-723, line 296: Note to myself: review this query against Celery task implementation.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,308 @@
...
+    /// Returns an error if:
+    ///
+    /// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
+    async fn update_terminal_status(
+        &self,
+        status: QueryJobStatus,
+        status_message: &str,
+        expected_status: QueryJobStatus,
+    ) -> Result<(), sqlx::Error> {
+        let query = format!(
```
LinZhihao-723: Sall we use `formatcp` as the compression coordinator?
https://github.com/y-scope/clp/blob/2a5fbee5d6c6635aa3a4c354317b57a865b96bd2/components/compression-coordinator/src/job_handle.rs#L462

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,308 @@
...
+                QueryJobStatus::Failed,
+                &format!("Query-job orchestration failed: {error}"),
+                QueryJobStatus::Pending,
+            )
+            .await
+        {
+            tracing::error!(
+                query_job_id = % self.query_job_id,
+                error = % status_error,
+                "Failed to persist the query-job failure.",
```
LinZhihao-723: I might miss this in the previous batch: we should use the term "query job", not "query-job". Please drop `-` for all occurrences.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,308 @@
...
+            );
+        }
```
LinZhihao-723: It looks like this can be re-written using `.inspect_err`.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,308 @@
...
+    ///
+    /// A newly created [`QueryJobHandle`] for the given already-planned query job.
+    pub const fn new(
+        db_pool: MySqlPool,
+        query_job_id: QueryJobId,
+        job_submitter: SubmitterType,
+        resource_group_id: ResourceGroupId,
+        clp_s_query_option: ClpSQueryOption,
+        output_handle: OutputHandle,
+        archives_to_search: Vec<(ArchiveMetadata, ExecutionPolicy)>,
```
LinZhihao-723, line 57: This signature is wrong: we shouldn't have `archives_to_serach` ready when creating a job handle: the query planning is a part of the job handle's job.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,308 @@
...
+        self.job_submitter
+            .submit_query_job(
+                self.query_job_id,
+                self.resource_group_id,
+                self.clp_s_query_option.clone(),
+                self.output_handle.clone(),
+                self.archives_to_search.clone(),
+            )
+            .await
+    }
```
LinZhihao-723, line 173: What's the purpose of this function if it's just a wrapper of `submit_query_job`?

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,308 @@
...
+use crate::query_job_submitter::QueryJobSubmitter;
+
+/// Spider polling options shared by query-job handles.
+pub struct SpiderOption {
+    /// Initial delay after a non-terminal Spider job-state poll.
+    pub initial_poll_backoff: Duration,
+
+    /// Maximum delay between Spider job-state polls.
+    pub max_poll_backoff: Duration,
+}
```
LinZhihao-723: Nothing really wrong here, but since it's a mirror of the compression job handle's option, we should try to make the docstring style consistent. Please check https://github.com/y-scope/clp/blob/2a5fbee5d6c6635aa3a4c354317b57a865b96bd2/components/compression-coordinator/src/job_handle.rs#L29

LinZhihao-723 (changes requested):
Still need another round to go through the actual details in the code level. There're some high-level issues, we might want to get aligned on the design level first before we dive more into the details.

components/query-coordinator/src/job_handle.rs
Bill-hbrhbr: fix

components/query-coordinator/src/job_handle.rs
Bill-hbrhbr: fix

components/query-coordinator/src/job_handle.rs
Bill-hbrhbr: fix all occurrences

components/query-coordinator/src/job_handle.rs
Bill-hbrhbr: fix

components/query-coordinator/src/job_handle.rs
Bill-hbrhbr: fix

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,326 @@
...
+use crate::query_job_submitter::QueryJobOutcome;
+use crate::query_job_submitter::QueryJobSubmitter;
+
+/// Options for a query job running in Spider.
+pub struct SpiderOption {
+    pub initial_poll_backoff: Duration,
+    pub max_poll_backoff: Duration,
+}
+
+/// Drives one already-planned query job through submission and terminal persistence.
```
LinZhihao-723:
```suggestion
/// Handles the asynchronous submission of a query job and the retrieval of its result.
```
For this type of docstrings, please check compression coordinator to make them consistent: https://github.com/y-scope/clp/blob/325bcc078f434a1a2a8edc47a72ac9286c211367/components/compression-coordinator/src/job_handle.rs#L37

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,326 @@
...
+    pub fn new(
+        db_pool: MySqlPool,
+        db_config: Database,
+        query_job_id: QueryJobId,
+        job_submitter: SubmitterType,
+        resource_group_id: ResourceGroupId,
+        search_job_config: SearchJobConfig,
+        output_handle: OutputHandle,
+        spider_option: Arc<SpiderOption>,
+    ) -> Result<Self, Error> {
```
LinZhihao-723: This is already getting verbose and not adequate enough to cover the parameters needed...
I think it's worth to define an arc of job-handle-context like the prototype branch: https://github.com/LinZhihao-723/clp/blob/12c19973a25728522ba52c20067b843008c72536/components/query-coordinator/src/job_handle.rs#L101

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,326 @@
...
+    pub async fn run(self) -> Result<(), Error> {
+        tracing::info!(query_job_id = % self.query_job_id, "Starting query job.");
+
+        let spider_job_id = match self.submit().await {
+            Ok(spider_job_id) => spider_job_id,
+            Err(error) => {
+                if !matches!(error, Error::JobNotPending(_)) {
+                    self.report_failure(&error).await;
+                }
+                return Err(error);
```
LinZhihao-723: I don't think I get it. Why do we need to handle this error here? Is it to handle the case where the job is already cancelled or sth?

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,326 @@
...
+            }
+        };
+        self.to_completion(spider_job_id).await
```
LinZhihao-723:
```suggestion
        match self.submit().await {
            Ok(spider_job_id) => self.to_completion(spider_job_id).await,
            Err(error) => {
                if !matches!(error, Error::JobNotPending(_)) {
                    self.report_failure(&error).await;
                }
                Err(error)
            }
        }
```
Style-wise: why not rewriting in this way?

components/query-coordinator/src/job_handle.rs
LinZhihao-723: Also, "already-planned" is wrong, right?

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// * [`Error::NoTaskInputs`] if no query task inputs are produced.
+    /// * [`Error::TooManyQueryTasks`] if the number of query tasks exceeds `i32`'s range.
+    /// * Forwards [`Self::prepare_task_inputs`]'s return values on failure.
+    /// * Forwards [`QueryJobSubmitter::submit_query_job`]'s return values on failure.
+    /// * Forwards [`Self::persist_spider_job_id`]'s return values on failure.
+    async fn submit(&self) -> Result<SpiderJobId, Error> {
+        let archives_to_search = self.prepare_task_inputs().await?;
+        let num_tasks = archives_to_search.len();
+        if num_tasks == 0 {
+            return Err(Error::NoTaskInputs);
```
LinZhihao-723: I didn't make this clearly in the offline discussion. This should not be considered as an error. Instead, we just early terminate the query as success. It could be the top-level filtering finds out there're nothing to search, meaning that the query will forever return no results.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// as failed before returning the original error. After the job is durably running, monitoring
+    /// and terminal-persistence failures leave it running so recovery can reattach to Spider.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`Self::submit`]'s return values on failure.
+    /// * Forwards [`Self::to_completion`]'s return values on failure.
+    pub async fn run(self) -> Result<(), Error> {
```
LinZhihao-723, line 108: As a part of https://github.com/y-scope/clp/pull/2513/changes#r4021775305:

We should split this function into more detailed steps:

1. Plan: returns the archives to search. When empty, terminate the query and mark it success without creating spider job.
2. Submit: create a spider job, persist its ID, and then start.
3. Wait until the job completes.

You may find the similar flow in the prototype branch: https://github.com/LinZhihao-723/clp/blob/12c19973a25728522ba52c20067b843008c72536/components/query-coordinator/src/job_handle.rs#L181. But we should move everything in `start` into this `run` function to make the steps explicit.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,326 @@
...
+                    error = % status_error,
+                    "Failed to persist the query job failure.",
+                );
+            });
+    }
+
+    /// Updates a query job only when it has the expected non-terminal status.
+    /// Leaves the status message unchanged when `status_message` is `None`.
+    /// A zero-row update is treated as success so an ineligible or missing job row is left
+    /// unchanged.
```
LinZhihao-723: I doubt this is a good design. If you look at how this function is used in the current implementation, it swallows all the cases where the job status can't be updated due to an invalid precondition. We should at least be able to log it, for example, in the case where we need to mark a job as terminated but failed.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,326 @@
...
+
+        let (status, status_message) = match outcome {
+            QueryJobOutcome::Succeeded => (QueryJobStatus::Succeeded, None),
+            QueryJobOutcome::Failed { error_message } => (
+                QueryJobStatus::Failed,
+                Some(format!("The Spider query job failed: {error_message}")),
+            ),
+        };
+        self.update_job_status(status, status_message.as_deref(), QueryJobStatus::Running)
+            .await?;
```
LinZhihao-723: The only case in which this can fail with an invalid precondition is when the job is canceled, right? I don't think we should allow a best-effort status update and treat failed preconditions as a success though.

components/query-coordinator/src/query_job_submitter/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+pub enum QueryJobOutcome {
+    /// Every archive query completed successfully.
+    Succeeded,
+
+    /// At least one archive query failed.
+    Failed {
+        /// The error reported by Spider.
+        error_message: String,
+    },
+}
```
LinZhihao-723, line 45: Missing "cancelled" state.

components/query-coordinator/src/query_job_submitter/spider.rs
```diff
@@ -0,0 +1,90 @@
...
+                            spider_job_id = % spider_job_id,
+                            error = % error,
+                            "Failed to fetch the Spider job error.",
+                        );
+                        format!("<failed to fetch job error: {error}>")
+                    }
+                };
+                QueryJobOutcome::Failed { error_message }
+            }
+            JobState::Cancelled => todo!("query job cancellation is not implemented"),
```
LinZhihao-723: Add the state in the job outcome enum. The way to handle cancellation should be added in the future but I don't see a reason why this piece of handler should be marked todo.

components/query-coordinator/src/query_job_submitter/spider.rs
```diff
@@ -0,0 +1,90 @@
...
+        let terminal_state = loop {
+            let state = self.get_job_state(spider_job_id).await?;
+            if state.is_terminal() {
+                break state;
+            }
+            tokio::time::sleep(backoff).await;
+            backoff = backoff
+                .saturating_mul(POLL_BACKOFF_FACTOR)
+                .min(max_poll_backoff);
+        };
```
LinZhihao-723: Query jobs are latency-sensitive. I don't think we should use exponential backoff; a long waiting time could waste time when the actual Spider job has terminated, but the job handle is still waiting for it.
We are trying to solve this problem with native Spider support. Before that, let's use a fixed polling time for query jobs.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,326 @@
...
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
+    async fn update_job_status(
+        &self,
+        status: QueryJobStatus,
+        status_message: Option<&str>,
+        expected_status: QueryJobStatus,
+    ) -> Result<(), sqlx::Error> {
```
LinZhihao-723: How about:
```
    async fn update_job_status(
        &self,
        from: QueryJobStatus,
        to: QueryJobStatus,
        msg: Option<&str>,
    ) -> Result<bool, sqlx::Error> {
```
- `status` -> `to`, `expected_status` -> `from`: The method is already named "update status". Using `from` and `to` with the type annotation should be clear.
- `status_message` -> `msg`: Drop the `status_` prefix.
- **Return a boolean instead**: This can be used to indicate whether the update actually went through.

LinZhihao-723 (changes requested):
Dive into more details.

components/query-coordinator/src/job_handle.rs
Bill-hbrhbr: Yes. That would make sense. I updated the docstring for you to review.

components/query-coordinator/src/job_handle.rs
Bill-hbrhbr: Please check if the current design is good enough.

components/query-coordinator/src/job_handle.rs
Bill-hbrhbr: Please check if the current design is good enough.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,365 @@
...
+/// Options for a query job running in Spider.
+pub struct SpiderOption {
+    pub poll_interval: Duration,
+}
+
+/// Resources shared by query job handles created by the coordinator.
+pub struct QueryJobHandleContext {
+    pub db_pool: MySqlPool,
+    pub db_config: Database,
+    pub output_handle: OutputHandle,
```
LinZhihao-723: I understand this is what the prototype branch does, but we should probably not share this in the context: everything in the context should be shared globally, but `OutputHandle` should be a per-query-job option.

components/query-coordinator/src/job_handle.rs
LinZhihao-723: No. My original comments:
- "we should move everything in start into this run function to make the steps explicit."
- "we should split this function into more detailed steps..."

You should break `plan_and_submit` into `plan` and `submit`. The API is just designed in a bad way: returning an optional Spider ID to differentiate whether there're archives to search is too ad-hoc.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,365 @@
...
+            {
+                return Err(Error::SqlxNoRowsAffected(format!(
+                    "no pending query job row found for query job {}",
+                    self.query_job_id
+                )));
+            }
+            return Ok(None);
+        }
+        self.submit(archives_to_search).await.map(Some)
+    }
```
LinZhihao-723: Move this into `run` as discussed.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,365 @@
...
+        ))
+        .bind(to)
+        .bind(msg.unwrap_or_default())
+        .bind(self.query_job_id)
+        .bind(from);
+        execute_update(query, &self.context.db_pool).await
+    }
+}
+
+/// Executes an SQL update and reports whether any row was affected.
```
LinZhihao-723: Missing docstring for return/error annotations.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// * Forwards [`Self::plan_and_submit`]'s return values on failure.
+    /// * Forwards [`Self::to_completion`]'s return values on failure.
+    pub async fn run(self) -> Result<(), Error> {
+        tracing::info!(query_job_id = % self.query_job_id, "Starting query job.");
+
+        match self.plan_and_submit().await {
+            Ok(Some(spider_job_id)) => self.to_completion(spider_job_id).await,
+            Ok(None) => Ok(()),
+            Err(error) => {
+                if !matches!(error, Error::SqlxNoRowsAffected(_)) {
```
LinZhihao-723: This error code is too general: imagine if I read this error code in this particular function without any context of what's happening under the hood, how could I understand why this error should be skipped?
`SqlxNoRowsAffected` here indicates either the job's precondition test failed, or the job doesn't even exist. It's worth to be explicit.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,365 @@
...
+                .await?
+            {
+                return Err(Error::SqlxNoRowsAffected(format!(
+                    "no pending query job row found for query job {}",
+                    self.query_job_id
+                )));
+            }
+            return Ok(None);
+        }
+        self.submit(archives_to_search).await.map(Some)
```
LinZhihao-723: Swap this to `!archives_to_search.is_empty()` and move the current if branch outside. This lead to less indentation:

```suggestion
        if !archives_to_search.is_empty() {
            return self.submit(archives_to_search).await.map(Some);
        }
        if !self
            .update_job_status(QueryJobStatus::Pending, QueryJobStatus::Succeeded, None)
            .await?
        {
            return Err(Error::SqlxNoRowsAffected(format!(
                "no pending query job row found for query job {}",
                self.query_job_id
            )));
        }
        Ok(None)
```

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,365 @@
...
+            spider_job_id = % spider_job_id,
+            outcome = ? outcome,
+            "Query job reached a terminal Spider state.",
+        );
+
+        let (status, status_message) = match outcome {
+            QueryJobOutcome::Succeeded => (QueryJobStatus::Succeeded, None),
+            QueryJobOutcome::Failed { error_message } => (
+                QueryJobStatus::Failed,
+                Some(format!("The Spider query job failed: {error_message}")),
```
LinZhihao-723:
```suggestion
                Some(format!("The Spider job failed: {error_message}")),
```
"Spider query job" is a wrong term.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,365 @@
...
+            ),
+            QueryJobOutcome::Cancelled => (
+                QueryJobStatus::Cancelled,
+                Some("The Spider query job was cancelled.".to_owned()),
```
LinZhihao-723:
```suggestion
                Some("The Spider job was cancelled.".to_owned()),
```

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,365 @@
...
+        let query = sqlx::query(formatcp!(
+            "UPDATE `{QUERY_JOBS_TABLE_NAME}` SET `status` = ?, `status_msg` = ? WHERE `id` = ? \
+             AND `status` = ?"
+        ))
+        .bind(to)
+        .bind(msg.unwrap_or_default())
+        .bind(self.query_job_id)
+        .bind(from);
+        execute_update(query, &self.context.db_pool).await
+    }
```
LinZhihao-723: Looking at now how this function is actually used in your implementation, combined with the issue of `SqlxNoRowsAffected`, you should:
- Drop `execute_update`: I don't really think reusing this piece of code logic is worth a helper.
- Instead of returning a `bool`, just return an error if no rows get updated:
  - In the current use case, you either inspect the error to log it or check the error on the top level. Keep it returning an error makes the error-handling path clean.

components/query-coordinator/src/error.rs
```diff
@@ -0,0 +1,26 @@
+//! The crate-level error type for the query coordinator.
+
+/// Errors returned by the query coordinator.
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("invalid query job configuration: {0}")]
+    InvalidQueryJobConfig(String),
+
+    #[error("failed to update SQL database: {0}")]
+    SqlxNoRowsAffected(String),
```
LinZhihao-723: As part of the discussion, here's what I propose:

```suggestion
    #[error(
        "metadata corrupted for query job {0}: its status changed or its row no longer exists"
    )]
    QueryJobMetadataCorrupted(QueryJobId),
```

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,365 @@
...
+            });
+    }
+
+    /// Updates the query job status in the CLP database.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
```
LinZhihao-723, line 369: By the way, this docstring is outdated after you switched to using `execute_update`. You should ask your coding agent to go through all the docstrings against the coding guidelines. I think they're particularly good at catching these types of errors.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,365 @@
...
+
+    /// Submits the prepared graph and drives the query job to a terminal state.
+    ///
+    /// On a submission failure, this method makes a best-effort attempt to mark the CLP query job
+    /// as failed before returning the original error. After the job is durably running, monitoring
+    /// and terminal-persistence failures leave it running so recovery can reattach to Spider.
+    ///
+    /// If no matching row is found when persisting the Spider ID, the job may have been cancelled,
+    /// deleted, or claimed by another coordinator job handler. Anyhow, this handle no longer owns
+    /// it, so it skips trying to report a job failure.
```
LinZhihao-723: "or claimed by another coordinator job handler" Why?? Shouldn't this be a bug?

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,365 @@
...
+impl<SubmitterType: QueryJobSubmitter> QueryJobHandle<SubmitterType> {
+    /// Factory function.
+    ///
+    /// # Returns
+    ///
+    /// A newly created [`QueryJobHandle`] for the given query job configuration.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the query string is empty.
```
LinZhihao-723: What error? Doesn't match the error-type docstring standard.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,365 @@
...
+        if !execute_update(query, &self.context.db_pool).await? {
+            return Err(Error::SqlxNoRowsAffected(format!(
+                "no pending query job row found for query job {} (Spider job ID {})",
+                self.query_job_id, spider_job_id
+            )));
+        }
+        Ok(())
+    }
+
+    /// Starts the Spider job if needed, waits for completion, and finalizes the query job.
```
LinZhihao-723: "starts the Spider job if needed" is not a direction operation running by this method, right? It's abstracted inside `run_query_job_to_completion`.
Please try to match the docstring of the compression coordinator's `to_completion`: https://github.com/y-scope/clp/blob/d7c53055b8ce97243244f9782c2824d907307294/components/compression-coordinator/src/job_handle.rs#L490

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// The archives to search and their execution policies on success.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if archive input preparation fails.
+    async fn prepare_task_inputs(&self) -> Result<Vec<(ArchiveMetadata, ExecutionPolicy)>, Error> {
+        todo!("prepare query task inputs")
+    }
+
+    /// Persists the Spider job ID and marks the query job as running.
```
LinZhihao-723, line 223: We should sync this docstring with compression coordinator side: https://github.com/y-scope/clp/blob/d7c53055b8ce97243244f9782c2824d907307294/components/compression-coordinator/src/job_handle.rs#L455

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// * [`Error::SqlxNoRowsAffected`] if no pending query job row was updated.
+    /// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
+    async fn persist_spider_job_id(
+        &self,
+        spider_job_id: SpiderJobId,
+        num_tasks: i32,
+    ) -> Result<(), Error> {
+        let query = formatcp!(
+            "UPDATE `{QUERY_JOBS_TABLE_NAME}` SET `spider_id` = ?, `status` = ?, `num_tasks` = ?, \
+             `start_time` = CURRENT_TIMESTAMP(3) WHERE `id` = ? AND `status` = ?"
```
LinZhihao-723: What is your plan for updating the dispatch time? Are you planning to do it in a future PR? If so, we should add a TODO.

components/query-coordinator/src/query_job_submitter/spider.rs
```diff
@@ -0,0 +1,175 @@
...
+
+        Ok(spider_job_id)
+    }
+
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`SpiderClient::start_job`]'s return values on failure, except
+    ///   [`ClientError::InvalidJobState`].
```
LinZhihao-723:
```suggestion
    ///   [`ClientError::InvalidJobState`], which indicates the job has already been started.
```
To match https://github.com/y-scope/clp/blob/d7c53055b8ce97243244f9782c2824d907307294/components/compression-coordinator/src/compression_job_submitter/spider.rs#L109

components/query-coordinator/src/query_job_submitter/spider.rs
```diff
@@ -0,0 +1,175 @@
...
+    /// * Forwards [`SpiderClient::get_job_state`]'s return values on failure.
+    ///
+    /// # Panics
+    ///
+    /// Panics if Spider returns a terminal state without a corresponding [`QueryJobOutcome`].
```
LinZhihao-723: We don't document `unreachable` panics.
```suggestion
```

components/query-coordinator/src/query_job_submitter/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+    async fn submit_query_job(
+        &self,
+        query_job_id: QueryJobId,
+        resource_group_id: ResourceGroupId,
+        clp_s_query_option: ClpSQueryOption,
+        output_handle: OutputHandle,
+        archives_to_search: Vec<(ArchiveMetadata, ExecutionPolicy)>,
+    ) -> Result<JobId, Error>;
+
+    /// Idempotently starts `spider_job_id` and waits for it to reach a terminal state.
```
LinZhihao-723: Please match the compression coordinator side docstring: https://github.com/y-scope/clp/blob/d7c53055b8ce97243244f9782c2824d907307294/components/compression-coordinator/src/compression_job_submitter/mod.rs#L67

LinZhihao-723 (changes requested):
- Please cross-reference to the compression coordinator's job handle implementation: there're a lot of functions/methods share the same semantics, but the docstrings diverge on both sides.
- Please ask coding agent to review the docstring's return values annotations after you've done all code changes.

components/query-coordinator/src/job_handle.rs
LinZhihao-723: A bigger issue is we don't have a formal definition of what ownership means in this context. This docstring should be rephrased.

components/query-coordinator/src/job_handle.rs
Bill-hbrhbr: Renamed class to `QueryCoordinatorContext` for better clarity

components/query-coordinator/src/job_handle.rs
Bill-hbrhbr: Decided to include it in this PR.

components/query-coordinator/src/job_handle.rs
Bill-hbrhbr: Refactored into `run()` for error handling and `run_job()` for the actual multi-stage execution.

components/query-coordinator/src/job_handle.rs
LinZhihao-723: Why?? This is not the context used by the query coordinator; if it is, it should not be located in this `job_handle` module. This is making it worth it in terms of clarity and scope. In the future, please:
- Do not do any random extra work on top of the comment I had. The comment never asked for a rename.
- If you think that's necessary, talk to me offline to confirm.

Bill-hbrhbr: sorry, reverted.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,409 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`Self::plan`]'s return values on failure.
+    /// * Forwards [`Self::update_job_status`]'s return values on failure for an empty plan.
+    /// * Forwards [`Self::submit`]'s return values on failure.
+    /// * Forwards [`Self::to_completion`]'s return values on failure.
+    async fn run_job(&self) -> Result<(), Error> {
```
LinZhihao-723: This is a private symbol, and it should not be placed here.

LinZhihao-723: This function is poorly named.

Bill-hbrhbr:
@LinZhihao-723 Cross-checked all the docstrings to keep the wording as close to the compression side as possible. A few places have to differ because of the new `Error::QueryJobMetadataCorrupted`:
  1. `update_job_status` now takes both a `from` and a `to` status, where `from` is an `Option`. `report_failure` passes `None` so it marks the job `Failed` regardless of its current status; the other call sites pin the status they expect to transition from.
  2. `report_failure` has to handle this error up front. If we already got it, re-attempting the row update is pointless. We just log it instead.

`report_failure` may not have the prettiest logic flow, but it avoids unnecessary SQL executions. Please advise on a cleaner logic flow if you can see it.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,409 @@
...
+    /// Returns an error if:
+    ///
+    /// * [`Error::QueryJobMetadataCorrupted`] if no pending query job row was updated.
+    /// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
+    async fn persist_spider_job_id(
+        &self,
+        spider_job_id: SpiderJobId,
+        num_tasks: i32,
+    ) -> Result<(), Error> {
+        let query = formatcp!(
```
LinZhihao-723: This query string is a constant since it's using `formatcp!`, but the variable itself is not.

components/query-coordinator/src/job_handle.rs
```diff
@@ -0,0 +1,409 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`Error::QueryJobMetadataCorrupted`] if no matching query job row can be found.
+    /// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
+    async fn update_job_status(
+        &self,
+        from: Option<QueryJobStatus>,
```
LinZhihao-723: I didn't expect this: the precondition should always be verified even if you want to set it to fail. Unlike compression jobs, an external service (API server) can set the job status to `Canceled` at any time, which is the only way to communicate a cancellation to the query coordinator.
I might have missed this detail, but if you're making this change now, it may indicate you're not aware of this issue. I will re-review the status handling for the entire flow on my own.

LinZhihao-723 (changes requested):
There are couple more things need to be fix for the implementation. And I will take over this PR from now on for the following reasons:
- It just took too long to iterate on my batch of comments.
- Your action on my comments sometimes bring extra "surprises" (for example, making the precondition none in `update_job_status`) without notifying me offline. Putting more time trying to understand what you've changed without a proper communication is just a waste of my time.

I believe doing it my own is the most efficient way to make this PR ready.

In the meantime, please review the planning implementation on your own before I move on to that PR.

LinZhihao-723 (commented):
This is part of my problem: I found it's hard to design the job-handler's programming model without seriously considering how cancellation should be handled. Before I make further code changes, I'm going to finalize the design for the formal definition of the job state transition, plus how these edges are handled. Please review the following. I will add them into the Notion doc as well.

# Query job state machine

Scope: query jobs driven by `query-coordinator`. `Killed` is out of scope — it is no longer used for coordinator-driven jobs.

## Transitions

```mermaid
stateDiagram-v2
    [*] --> Pending

    Pending --> Running: handle submits to Spider
    Pending --> Cancelling: API server
    Running --> Cancelling: API server

    Running --> Succeeded: handle, Spider succeeded
    Running --> Failed: handle, Spider failed

    Cancelling --> Cancelled: handle, Spider cancelled or succeeded
    Cancelling --> Cancelled: coordinator restart sweep
    Cancelling --> Failed: handle, job failed

    Pending --> Succeeded: handle, no archives to search
    Pending --> Failed: handle, planning failed

    Succeeded --> [*]
    Failed --> [*]
    Cancelled --> [*]
```

## State classes

* **Cancellable**: `Pending`, `Running`. Enforced in the API server — already the case today, `cancel_search_job` updates only `WHERE status IN (Pending, Running)`.
* **Terminal**: `Succeeded`, `Failed`, `Cancelled`.
* **Live**: `Pending`, `Running`, `Cancelling`. `Cancelling` is *not* terminal; it is a state the job handle must still drive to a terminal one.

## Outcome priority

When a cancellation races a Spider outcome: **failed > cancelled > succeeded**.

A job in `Cancelling` whose Spider job succeeded still terminates as `Cancelled` — the cancellation wins over the success. A job in `Cancelling` whose Spider job failed terminates as `Failed` — the failure wins over the cancellation.

## Cancellation flow

* A background coroutine monitors jobs in `Cancelling`. If the job has an associated Spider job, it cancels that Spider job; otherwise it does nothing. **It never writes job status** — only the job handle and the restart sweep do.
* The job handle observes the cancellation one of two ways:
  * Before a Spider job was submitted: it sees the status transition and terminates the job as `Cancelled` directly.
  * After a Spider job was submitted: it sees the Spider job return `cancelled`, which terminates the job as `Cancelled`.
* On coordinator restart, jobs found in `Cancelling` are swept to `Cancelled`.

LinZhihao-723 (commented):
To implement the status transition, and also to solve the problem where `update_job_status(from: Option<QueryJobStatus>, ...)` conflates too much: the `from: None` case blindly overwrites any status, including a terminal one, and every failure collapses into `rows_affected == 0`, so "already terminated", "invalid transition" and "row is gone" are indistinguishable. Here's what I'm going to change on top of the current code:

## Error variants

| Variant | Meaning |
|---|---|
| `QueryJobMetadataCorrupted` | The job's row is missing. |
| `QueryJobCancelled` | The job was cancelled before it was submitted to Spider. Returned by `start`. |
| `InvalidQueryJobStatusTransition(from, to)` | The transition isn't in the state machine. Records the observed `from`, so a terminal `from` is how "the job already terminated" is reported. |

## `TerminalQueryJobStatus`

A private enum holding the statuses a handle can produce — `Succeeded`, `Failed`, `Cancelled` — with a method mapping to `QueryJobStatus`. This makes "must be terminal" unrepresentable rather than asserted at runtime.

## `persist_spider_job_id` → `start`

Runs in a transaction that locks the row:

- Row missing → `QueryJobMetadataCorrupted`.
- Status is `Cancelling` → `QueryJobCancelled`.
- Status is not `Pending` → `InvalidQueryJobStatusTransition`. This covers a terminal status too; the recorded `from` says which.
- Otherwise → persist `spider_id`, `num_tasks`, `start_time`, `dispatch_time`, and move to `Running`.

## `update_job_status` → `terminate`

Takes a `TerminalQueryJobStatus` and an optional status message. Runs in a transaction that locks the row:

- Row missing → `QueryJobMetadataCorrupted`.

Then, by target status. Each rule below rejects a disallowed current status — including any terminal one — with `InvalidQueryJobStatusTransition`, so no separate already-terminated check is needed:

- **`Succeeded`** — permitted from `Running` (Spider succeeded) and from `Pending` (empty query plan, no archives to search). If the status is `Cancelling`, write `Cancelled` instead, with a status message recording that the job had in fact completed, log it, and return `Ok`. Any other status → `InvalidQueryJobStatusTransition`.
- **`Cancelled`** — permitted only from `Cancelling`. Any other status → `InvalidQueryJobStatusTransition`.
- **`Failed`** — permitted from any live status. Failure takes priority over cancellation.

The transaction exists for **error diagnosis, not correctness**: a single guarded `UPDATE ... WHERE status NOT IN (<terminal>)` is already atomic and cannot lose a terminal status. It is what lets us report which of the three error variants above applies, and which status was actually observed, instead of collapsing everything into "no rows updated". Keep this note in the docstring.

## `report_failure` → `finalize_on_error`

Logs the error as a job-handle failure, then dispatches on the variant:

- `QueryJobMetadataCorrupted` → nothing to persist.
- `QueryJobCancelled` → attempt `terminate(Cancelled)`.
- `InvalidQueryJobStatusTransition(from, _)` → if `from.is_terminal()`, the job already reached a terminal status through another path, so there is nothing to persist. Otherwise attempt `terminate(Failed)`.
- Any other error → attempt `terminate(Failed)`.

A failed attempt is logged and swallowed; it must not re-enter this path. An attempt that comes back `InvalidQueryJobStatusTransition` with a terminal `from` lost a race to another path rather than failing, so it is logged at `debug`/`warn`, not `error`.

LinZhihao-723 (commented):
Implemented what's described in https://github.com/y-scope/clp/pull/2513#pullrequestreview-5286171167 and https://github.com/y-scope/clp/pull/2513#pullrequestreview-5286744460. One difference I made was to reuse `QueryJobOutcome` as the only parameter passing to `terminate`. Please check @Bill-hbrhbr.

PR #2498, opened 2026-08-24
docs: Document GitHub workflows and jobs missed in previous PRs.

kirkrodrigues:
# Description

Several GH workflows and jobs that were added in previous PRs haven't been documented in our dev docs. This PR documents those workflows and jobs. Note that the docs could use some refactoring to make the information easier to digest, but in the interest of keeping the PR small, I've deferred those changes to a future PR.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* `task docs:serve`
* Validated that the Developer Docs -> Tooling -> GitHub workflows page rendered as expected.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **Documentation**
  * Expanded workflow documentation for architecture-specific dependency images, Python wheels, CLP and Spider-worker images, linting, unit tests, and integration tests.
  * Updated build workflow guidance for trigger calculation, multi-architecture manifests, image publishing, and published-image selection.
  * Added guidance for macOS linting, documentation validation, generated-code checks, language-specific linting, Helm publishing, pull request title validation, and Rust checks.
  * Updated workflow job descriptions and dependency relationships.

PR #2518, opened 2026-09-08
ci: Consolidate calculation of container image build and usage conditions into a single workflow job.

kirkrodrigues:
# Description

The current set of conditions used for building spread out all over the clp-artifacts-build workflow, making it hard to understand how and why a certain container image will be built.

Similarly, the current set of conditions for determining when a published container image should be used were also spread out.

This PR:

* updates `filter-relevant-changes` job to also compute flags for when a container image should be built and when a published image should be used.
* renames `filter-relevant-changes` to `calc-build-triggers` to reflect its updated purpose.
* removes some redundant comments that are more confusing than reading the code (imo).
* updates the docs.
  * NOTE: The docs are missing details about some jobs like the `spider-worker-image` job. These updates are being made in #2498.

Consolidating the conditions into a single job will be useful in a future PR where we need to publish container images on semver branches, in addition to main.

# Checklist

  breaking change.

# Validation performed

* Validated that all jobs in the clp-build-artifact that should run on PRs ran.
* Merged the PR in my fork and validated that all jobs in clp-build-artifact that should run on main ran.
* Edited a file in CLP's core and validated that the core dependency images weren't rebuilt.

## Summary by CodeRabbit

- **Chores**
  - Improved build automation to determine which platform images and package or worker artifacts should be published or reused.
  - Updated build dependencies, image selection, publishing, and manifest generation based on calculated build triggers.
  - Ensured Ubuntu Jammy aarch64 dependencies are built whenever publishing is enabled, including scheduled and eligible main-branch builds.

- **Documentation**
  - Updated workflow documentation and diagrams to reflect the revised build-trigger process, publishing decisions, and image-selection behaviour.

### DIFF
--- a/.github/workflows/clp-artifact-build.yaml
+++ b/.github/workflows/clp-artifact-build.yaml
@@ -1008,7 +1008,10 @@ jobs:
           image_registry_username: "${{github.actor}}"
           image_registry_password: "${{secrets.GITHUB_TOKEN}}"
           arch: "${{matrix.arch}}"
+          dockerfile_path: "./tools/docker-images/clp-package/Dockerfile"
+          image_name: "clp-package"
           platform_version_codename: "jammy"
+          push: "${{github.event_name != 'pull_request' && github.ref == 'refs/heads/main'}}"
 
   package-image-multiarch-manifest:
     name: "package-image-multiarch-manifest"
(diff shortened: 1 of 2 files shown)


.github/workflows/clp-artifact-build.yaml
20001020ycx: should this be `needs.calc-build-triggers.outputs.package_publish_image == 'true'`? I tried a grep on `${{github.event_name != 'pull_request' && github.ref == 'refs/heads/main'` and found two dangling instances.

20001020ycx: Ditto as comment in https://github.com/y-scope/clp/pull/2518/changes#r3964494857

.github/workflows/clp-artifact-build.yaml
```diff
@@ -32,19 +32,46 @@ concurrency:
...
       ${{
         github.repository_owner == 'y-scope'
         && fromJSON('["self-hosted", "x64", "ubuntu-noble"]')
         || 'ubuntu-24.04'
       }}
     outputs:
+      # Container-image outputs
       centos_stream_9_image_changed: "${{steps.filter.outputs.centos_stream_9_image}}"
+      centos_stream_9_publish_image: >-
+        ${{github.event_name != 'pull_request' && github.ref == 'refs/heads/main'}}
```
20001020ycx: Should we hoist this expression into a variable as this is used 6 times, is this doable?

.github/workflows/clp-artifact-build.yaml
20001020ycx: actually, ignore this comment, the semantic here is difference. Here it means if we need to build ARM image or not, but the vairable was intended for whether we shall publish the image or not. we should keep it as `${{github.event_name != 'pull_request' && github.ref == 'refs/heads/main'`. 

But this confusion might support my claim in https://github.com/y-scope/clp/pull/2518/changes#r3964511856, as I only aware its semantic when actually carefully reading the code.

.github/workflows/clp-artifact-build.yaml
kirkrodrigues: How about an output called `is_push_or_run_on_main`?

.github/workflows/clp-artifact-build.yaml
kirkrodrigues: I ended up going with your initial suggestion. I felt like creating an output to replace `github.event_name != 'pull_request' && github.ref == 'refs/heads/main'` ended up being just as confusing, and then your initial suggestion made sense again, lol.

20001020ycx (approved):

